PR #45, opened 2014-07-28
added dependency to javafx maven plugin

notmandatory:
This is to get around javafx-maven-plugin issue with latest maven
version. see:
http://stackoverflow.com/questions/19407959/javafx-maven-plugin-and-api-
incompatibility


bisq-network:
Ah great! Thanks!

PR #251, opened 2014-11-03
Support code coverage using JaCoCo and Coveralls

cbeams:
The Gradle build now supports generating code coverage reports using
JaCoCo. The reports can be run with:

```
./gradle jacocoTestReport
```

and then view the HTML output at:

```
build/reports/jacoco/test/html/index.html
```

The bitsquare repository has now also been registered with a (free)
account at http://coveralls.io. Our Travis CI configuration has been
updated to publish JaCoCo XML test report data to coveralls.io, which
results in the reports found at:
- https://coveralls.io/r/bitsquare/bitsquare

For more information on the JaCoCo and Coveralls Gradle plugins, see:
- http://www.gradle.org/docs/current/userguide/jacoco_plugin.html
- https://github.com/kt3k/coveralls-gradle-plugin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,3 +18,9 @@ notifications:
     on_failure: always
     use_notice: true
     skip_join: true
+
+env:
+  - TERM=dumb
+
+after_success:
+  - ./gradlew jacocoTestReport coveralls
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 <img src="https://bitsquare.io/images/logo.png" width="240"/>
 
 [![Build Status](https://travis-ci.org/bitsquare/bitsquare.svg?branch=master)](https://travis-ci.org/bitsquare/bitsquare)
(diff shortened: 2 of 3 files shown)


ManfredKarrer:
Ah cool! Our coverage is not so cool, but did nto expect more for the current state...

cbeams:
yeah. will be good to watch it ratchet up from here :)

ManfredKarrer:
Even when we will have much more tests the UI code will drive the value down. For UI tests are IMO overhead and make agile development harder. At least in the phase where the UI is not completely defined and finalized.

cbeams:
yeah. we can set up exclusions for the `gui` package, etc as appropriate.

ManfredKarrer:
Yes that might make sense. The UI has 20k loc, the rest 12k loc. So it would distort the overall coverage pretty  much.

Issue #291, opened 2014-11-23
Add --clean param and button in settings to clean data directory


ManfredKarrer:

ManfredKarrer:
Only param impl. atm. as after startup a dir clean would cause problems. The removal is handled before the startup process, so all files will be created form scratch.

cbeams:
Per aa3fc92, `--clean=true` is now accepted as a convenient shortcut for `--app.data.dir.clean=true` (just to make it as easy as possible for folks trying out the alpha, etc.)

PR #296, opened 2014-11-27
Add support for $HOME/.bitsquare/bitsquare.properties

cbeams:
This change allows users to place a bitsquare.properties file within
a `.bitsquare` directory under their home directory. This makes it
possible to specify property values (such as
node.useManualPortForwarding=true) that will survive cleaning the
application data dir (e.g. with `--clean=true` flag as introduced
in #291).

The precedence of this new "homeDirProperties" property source is lower
than the "appDirProperties" property source.

This means that if a user has a `bitsquare.properties` file both in his
home directory and in his application data directory, both files will be
processed, but for any properties that exist in both files, the property
specified in the application data directory file will take precedence.
This arrangement allows users to customize settings on an app-by-app
basis, which is especially useful in local testing scenarios where more
than one Bitsquare instance is running.


cbeams:
FYI @bitsquare/core-team

PR #326, opened 2014-12-20
Make addresses selectable so that they can be copied to the clipboard

alexbirkett:
I can't find my own wallet address anywhere in the app other than in the transactions list so being able to copy / paste addresses is a big deal. 

This is intended as a temporary change to make addresses copy / pasteable until the details dialog is ready.

### DIFF
--- a/src/main/java/io/bitsquare/gui/main/funds/transactions/TransactionsView.java
+++ b/src/main/java/io/bitsquare/gui/main/funds/transactions/TransactionsView.java
@@ -18,7 +18,6 @@
 package io.bitsquare.gui.main.funds.transactions;
 
 import io.bitsquare.btc.WalletService;
-import io.bitsquare.gui.components.Popups;
 import io.bitsquare.gui.util.BSFormatter;
 
 import org.bitcoinj.core.Transaction;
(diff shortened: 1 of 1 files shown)


cbeams:
Thanks, Alex! @ManfredKarrer, assigned to you for review and merging.

ManfredKarrer:
@alexbirkett Thanks a lot for the PR.

Bitsquare uses dedicated trade wallets for every single trade instead of a global application wallet. The reason behind that design is to protect the traders privacy by avoiding coin merge. So there is no concept to pay in bitcoin in advance, but just in time when you want to create an offer or take an offer.
Thats why there is no deposit screen under funds.

But it might be useful to select the address field. In the tx history it is intended to open later a popup with details information regarding that tx but for now a way to copy the address would be good as well.
Instead of making the label selectable I prefer to open directly the blockexlorer (will be later customizeable in the preferences which block explorer is preferred). I think the primary use case if you want to copy the address is to check it on the blockexplorer, and even if not then you can easily copy the address from there.
Here is the actual commit.
https://github.com/bitsquare/bitsquare/commit/f5282fc3d6ca1e51a71837fcbbbb58f741bb7958

PR #330, opened 2014-12-21
Add copy column to transaction view

alexbirkett:
Adds copy button to the transaction view. 

I missed the fact that the other views have a button that copies the address to the clipboard! In my opinion the copy button is not very intuitive. It's not immediately obvious what it does: Copy the address to the clipboard? copy the balance /  amount or something else? It has a mouse over explanation but the text is invisible when the row is selected.

After 15 years of using web apps, it's strange that everything is not selectable / copy pastable. So if I could choose, I would do away with all the copy buttons and make all the addresses / amounts selectable. Would you accept a pull request that did that to all the views?

### DIFF
--- a/src/main/java/io/bitsquare/gui/main/funds/transactions/TransactionsView.fxml
+++ b/src/main/java/io/bitsquare/gui/main/funds/transactions/TransactionsView.fxml
@@ -44,7 +44,7 @@
                     <PropertyValueFactory property="amount"/>
                 </cellValueFactory>
             </TableColumn>
-
+            <TableColumn text="Copy" fx:id="copyColumn" minWidth="30" sortable="false"/>
             <TableColumn text="Status" fx:id="confidenceColumn" minWidth="30" sortable="false"/>
         </columns>
     </TableView>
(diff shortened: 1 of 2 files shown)


ManfredKarrer:
Hi Alex,
thanks for your PR!

I use copy/selectable only in cases when there is a clear use-case for the user. So informational data are not selectable/copyable by default. The copy button from the withdrawal screen should be removed, as the below form get filled automatically by selecting the row (which might be not so obvious for the user- thats why I added the copy button here, but thats not a final solution, probably a select button would be more appropiate).
In the tx history the only use case for copy I can see is to check the tx in a block explorer.
Click on the address in the tx histroy should open a popup displaying details info. An external link button (to block explorer) might be useful. If the address use OP_RETURN (used in deposit multisig and registration - which is deactivated atm) the address is not displayed at the moment, that should be fixed as well.
So the current screens are not representing the optimal state. I would prefer instead merging your PR to change directly to open a popup displaying all the important details data of that tx (similar to btc core or other wallets). If you are interested to work on that would be very welcome. 
So to change all that would include:
- to change the copy button in widthdrawal to a select button.
- add a external link button next to the address hyperlink field (in the same column) in tx history (opening block explorer) - there is no icon yet for that
- open a popup with all tx details
- display the address also form those tx using OP_RETURN in the tx history (that might be not so trivial and can be postponed as well - we need just to handle those cases - atm causes errors when clicked...) 

I am half on holiday until mid of january so i am not so much available but will try to answer if there are questions....

PR #331, opened 2014-12-22
Use https URL for artifactoryonline.com

msgilligan:
It's a best practice to use `https://` URLs for all maven repositories. (I was just trying a `./gradlew run` and noticed an `http://` URL in the output.)


ManfredKarrer:
Thanks! Good point.

PR #334, opened 2015-02-07
Updated project to use 5.0-Beta2 of tomp2p

joeschmid:
I don't have a lot of background on the project yet so I might be missing some history, but I've updated build.gradle to use the latest release of tomp2p (5.0 Beta 2 which came out on Jan 8, 2015) and switched to use the tomp2p maven repository. I do see that bitsquare has its own fork of tomp2p, but it looked to me like there weren't substantial changes in that fork. (I did see some changes to examples, etc.) Again, I might be missing some important history so feel free to reject this PR if it's not helpful. Even if accepted, we should probably stick this on a branch so that it can be tested before being merged into master.

Note that this has not been thoroughly tested and, in particular, my router does not support UPnP so I've only been able to test with --node.useManualPortForwarding=true  The project does compile and Bitsquare does run successfully though. With manual port forwarding I'm able to connect to the p2p network and view open offers successfully.

This will obviously need thorough testing before being incorporated into an actual release!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,12 +36,12 @@ processResources {
 
 repositories {
     jcenter()
-    maven { url 'https://partnerdemo.artifactoryonline.com/partnerdemo/libs-snapshots-local' }
+    maven { url 'http://tomp2p.net/dev/mvn/' }
 }
 
 dependencies {
     compile 'org.bitcoinj:bitcoinj-core:0.12.2'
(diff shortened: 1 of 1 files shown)


ManfredKarrer:
Thanks! I will give it a bit more testing the next week. Caught a cold :-( so might take a bit...

ManfredKarrer:
Tested trades with 2 clients and all works like expected. Thanks for the PR.

PR #353, opened 2016-01-24
changed the offline message

rohythakur:
I have changed the msg shown to the user when he looses his internet connection , I have changed it to 
"You lost your connection, Unable to connect to the internet.\n Please check your internet connection or try to restart the application."


ManfredKarrer:
The message:
Maybe you lost your internet connection or your computer was in hibernate/sleep mode.
describes it better IMO. Asking for restart is not needed as the network re-establish connections if possible. The main case will when the computer gets into sleep mode.

rohythakur:
@ManfredKarrer  ok got it

PR #372, opened 2016-04-05
Escape vertical bar when splitting seed node addresses in command line

ivilata:
Otherwise it is interpreted as a special regular expression character and all characters are splitted apart, causing a `NullPointerException` after further splitting on `:`.

The fix allows using `NODE1_ADDR.onion:PORT1|NODE2_ADDR.onion:PORT2|…` to indicate nodes to connect to when running `SeedNode.jar`.


ManfredKarrer:
Ah thanks!

PR #389, opened 2016-04-20
Update build.md

JoshLambda:
When building package, step #7 is needed before step #3 or the later would fail. I suggest moving step #7 to prerequisites chapter since that's what it seems to be.


ManfredKarrer:
Thanks!

PR #433, opened 2016-05-05
New timer race conditions

ivilata:
I found some race conditions in message handlers. Basically, they do a `networkNode.sendMessage()`, then add callbacks to the resulting future, and then create a timeout timer.

The problem is that the message exchange may actually run, cleanup and set stop before timer creation (e.g. when testing in a single machine), which goes on to create the timer anyway.  The timer does trigger of course (since the operation completed before) and when it runs some cleanup method it uses to check `stopped` (which is already true) and complain about that.

This PR fixes the issue by creating the timer before sending the message.  I ran the network stress test on this and some warnings have gone away.

I also added some missing checks for `stopped` in `GetDataRequestHandler.handle()` and `BroadcastHandler.broadcast()`.


ManfredKarrer:
Good find!

PR #501, opened 2016-06-21
Second try for the jdk bug fix

mrosseel:
- jdkfix maven module has been added to create its own jdkfix.jar ; this mechanism can be used for all jdk bugs.
- in package directory, change creation script to use the -Xbootclasspath option which puts jdkfix.jar before the jdk classes.

TODO:
- change arch script at: https://aur.archlinux.org/cgit/aur.git/tree/bitsquare.sh?h=bitsquare-git
- check if this is an issue:

```
Building DMG package for Bitsquare
Did not find a key matching 'Developer ID Application: '
Result DMG installer for Bitsquare: /Users/mike/dev/bitcoin_projects/bitsquare/gui/deploy/bundles/Bitsquare-0.4.9.dmg

```
- test whether everything works by using -verbose:class


ManfredKarrer:
The "Did not find a key matching" warning is probably because you don't have a Apple developer license.

Still testing but so far I cannot get it working.

ManfredKarrer:
Get a java.lang.NoClassDefFoundError: javafx/collections/transformation/TransformationList

ManfredKarrer:
Ah now it works with following:
-Xbootclasspath/a:"jdkfix/target/jdkfix-0.4.9.jar:/Library/Java/JavaVirtualMachines/jdk1.8.0_92.jdk/Contents/Home/jre/lib/ext/jfxrt.jar"

PR #505, opened 2016-06-23
Update Myriad ticker, MYR->XMY

cryptapus:
Since MYR conflicts with Malaysian Ringgit, Myriad will be transitioning to XMY


ManfredKarrer:
Ok I changed from MYRC to XMY

PR #506, opened 2016-06-25
Add verbose apt-get install instruction for debian

Z819:
I had to search the web to find the correct packages. And at the end wasn't sure if it was correct. A more verbose instruction would have had helped.
I'm not sure, if the packages are correct, or if anything is missing, please review.


ManfredKarrer:
Hm, I hvae nto tried those so I cannot confirm.
I used that to install the Oracle JDK on Ubuntu:

$ add-apt-repository ppa:webupd8team/java
$ apt-get update
$ apt-get -y oracle-java8-installer install git maven

I will merge yours and add mine as well.

PR #508, opened 2016-06-28
Second version of SortedList fix

mrosseel:
This is the cleaner version of the fix, please test to see if it still works in your environment.


ManfredKarrer:
Does not work. The jdkfix jar is on the same level as the bitsquare.jar, but it would need to be inside the bitcoin jar, what si probably not possible. ../ reference also did not work :-(

ManfredKarrer:
Ah sorry my test was not correct, need to build again...

ManfredKarrer:
Yeah it works! Thanks a lot!

Issue #488, opened 2016-06-11
Bounty: Route BitcoinJ over Bitsquare's JTorProxy


ManfredKarrer:
Requirements for taking that bounty:
- Expertise with Java
- Expertise with Tor
- Expertise with BitcoinJ
- Expertise with the usage of JTorProxy and the related libraries/network module in Bitsquare 

We used to run BitcoinJ over Tor with its built in Tor support (Orchid) but that was not very stable, so we deactivated that. As we use a native Tor instance for the P2P network it would be better to route BitcoinJ over our JTorProxy.
BitcoinJ uses nio (netty) which does not provide an API for sockets, so that might be difficult to get working with Socks5Proxy. But it also supports a BlockingClientManager (no nio) which is also used when BitcoinJ runs over Orchid.
We would need to pass our socket (or socketFactory) to the BlockingClientManager.
We maintain our own fork (https://github.com/bitsquare/bitcoinj) of BitcoinJ so changes need to be added there.

If you want to work on that bounty it requires close communication. I need to be sure that you fulfill the above requirements! Also the final scope of the bounty will be decided during work on it. The bounty amount will be paid out for a working implementation where BitcoinJ is routed over our Tor instance and it does not conflict with the Bitsquare P2P network. The work might require more effort as it is covered by the bounty, if work relationship turns out positive we can arrange alternative reward schemes.

Bounty address (initial funding 1 BTC):
https://blockchain.info/address/1LkUscnxVAeNsuRMu9Veuo6MLaoDEG9JXw

Related issues: https://github.com/bitsquare/bitsquare/issues/357

dan-da:
I was surprised to notice that bitsquare is connecting to mainnet directly and not through tor, since the p2p operates through tor.   I came to github to find if there is an open issue or else to create one, so I'm glad to find this writeup.

one question I have.  It is easy to route bitcoin-core through a tor proxy simply by setting a couple config settings (proxy, onlynet).   Does bitcoinJ not support proxies?  And if it does support a proxy, then couldn't users at least configure it to route through their O/S tor proxy?   Or alternatively through bitsquare's tor if it exposes an open proxy port?     I'm just thinking that could be a way to use existing plumbing rather than integrate at a lower level.

ManfredKarrer:
I don't know how easy it is to route it from the outside but from the BitcoinJ side it should not be too hard, but needs some dev effort I don't have at the moment. 
[...]

PR #534, opened 2016-07-21
Route bitcoinj over jtoryproxy

dan-da:
This pull request is in response to #488.    At this point, the code is generally working as intended but needs review by others.

**Summary of Changes**
- Routes bitcoinj traffic over internal Tor proxy (or not) based on user pref.
- Defers walletService init until internal tor is started, but only when useTor pref is on.
- Changes SafeSocks setting in torrc from 1 to 0 to allow bare IPs over Tor.
- Enables remote DNS lookup when using socks5 proxy, eg Tor.
- parses and makes use of --btcSeedNodes arg.
- code is in place but commented out for user to specify other socks5 proxy, eg i2p, ssh, vpn.  it just needs GUI or CLI prefs added. See MainViewModel.

**Testing**

I have run the code with the following permutations:
- Tor Disabled, no seedpeers.
- Tor Enabled, no seedpeers.
- Tor Enabled, using seedpeers.
- Tor Disabled using seedpeers.

In all cases, the wallet service was able to get connected and function seemingly normally.

I have NOT yet been able to verify that bitcoinj is not somehow leaking connections outside Tor, via DNS or otherwise.   If anyone knows a good tool for logging all outbound connections made by a process during its lifetime on linux, please let me know.   wireshark and tcpdump are possibilities, but tcpdump is snapshot oriented and wireshark doesn't know anything about PID.

Suggestions for verification methodology appreciated.  We should make sure it is not leaking before making a release.

**Notes**
- The code has been written in such a way that bitcoinj only sees a Socks5 proxy.  There is nothing Tor specific about it.  This means that we can (and I think should) add a mechanism for user to specify socks5 host/port.  This way, the wallet could be used over i2p, ssh, vpn, etc if desired.
- I still get various types of connection errors when going over Tor, but still connections get made.
- I still see occasional "Connection not allowed by ruleset" messages and always when this happens the IP is in ipV6 format, so apparently DNS is handing these back also.  research required.
- I don't always get the green icon indicating that the wallet is ready at startup.  This may just be a timing issue, due to walletService init being deferred.  I'm not sure.
- I've tried to comment the code to highlight issues or areas for improvement.
- At present, we are only using the hostname seeds from bitcoinj NetworkParameters.  I considered making use of the IP seeds also, but decided that is outside the scope of this issue.

**Future Improvement**
- .onion addresses do not work, eg as seedNodes.  The PeerAddress class in bitcoinj throws null pointer exception if fed an unresolved hostname.  It might be as simple as fixing that, or might require significant changes in bitcoinj.
- bitcoinj does not include .onion seeds in the parameters for each network, so those classes would need to be added to support it.

**Bounty**

@ManfredKarrer  At such time as you think this code is worthy of the bounty, you can make payment to the address I've sent you privately.  Here's my public key fingerprint to prove I am the same person.

```
5C0E F833 79CB 892F 3FC9 1519 744D D69B 94E9 CC1A
```


dan-da:
oops, I made this against master instead of Development.  closing and will make a new one.

PR #535, opened 2016-07-22
Route bitcoinj over jtoryproxy

dan-da:
This pull request is in response to #488.    At this point, the code is generally working as intended but needs review by others.

**Summary of Changes**
- Routes bitcoinj traffic over internal Tor proxy (or not) based on user pref.
- Defers walletService init until internal tor is started, but only when useTor pref is on.
- Changes SafeSocks setting in torrc from 1 to 0 to allow bare IPs over Tor.
- Enables remote DNS lookup when using socks5 proxy, eg Tor.
- parses and makes use of --btcSeedNodes arg.
- code is in place but commented out for user to specify other socks5 proxy, eg i2p, ssh, vpn.  it just needs GUI or CLI prefs added. See MainViewModel.

**Testing**

I have run the code with the following permutations:
- Tor Disabled, no seedpeers.
- Tor Enabled, no seedpeers.
- Tor Enabled, using seedpeers.
- Tor Disabled using seedpeers.

In all cases, the wallet service was able to get connected and function seemingly normally.

I have NOT yet been able to verify that bitcoinj is not somehow leaking connections outside Tor, via DNS or otherwise.   If anyone knows a good tool for logging all outbound connections made by a process during its lifetime on linux, please let me know.   wireshark and tcpdump are possibilities, but tcpdump is snapshot oriented and wireshark doesn't know anything about PID.

Suggestions for verification methodology appreciated.  We should make sure it is not leaking before making a release.

**Notes**
- The code has been written in such a way that bitcoinj only sees a Socks5 proxy.  There is nothing Tor specific about it.  This means that we can (and I think should) add a mechanism for user to specify socks5 host/port.  This way, the wallet could be used over i2p, ssh, vpn, etc if desired.
- I still get various types of connection errors when going over Tor, but still connections get made.
- I still see occasional "Connection not allowed by ruleset" messages and always when this happens the IP is in ipV6 format, so apparently DNS is handing these back also.  research required.
- I don't always get the green icon indicating that the wallet is ready at startup.  This may just be a timing issue, due to walletService init being deferred.  I'm not sure.
- I've tried to comment the code to highlight issues or areas for improvement.
- At present, we are only using the hostname seeds from bitcoinj NetworkParameters.  I considered making use of the IP seeds also, but decided that is outside the scope of this issue.

**Future Improvement**
- .onion addresses do not work, eg as seedNodes.  The PeerAddress class in bitcoinj throws null pointer exception if fed an unresolved hostname.  It might be as simple as fixing that, or might require significant changes in bitcoinj.
- bitcoinj does not include .onion seeds in the parameters for each network, so those classes would need to be added to support it.

**Bounty**

@ManfredKarrer  At such time as you think this code is worthy of the bounty, you can make payment to the address I've sent you privately.  Here's my public key fingerprint to prove I am the same person.

```
5C0E F833 79CB 892F 3FC9 1519 744D D69B 94E9 CC1A
```


core/src/main/java/io/bitsquare/btc/SeedPeersSocks5Dns.java
```diff
@@ -0,0 +1,178 @@
...
+        this.seedAddrs = convertAddrsString( params.getDnsSeeds(), params.getPort() );
+
+        if( false ) {
+            // This is an example of how .onion servers could be used.  Unfortunately there is presently no way
+            // to hand the onion address (or a connected socket) back to bitcoinj without it crashing in PeerAddress.
+            // note:  the onion addresses should be added into bitcoinj NetworkParameters classes, eg for mainnet, testnet
+            //        not here!
+            this.seedAddrs = new InetSocketAddress[] { InetSocketAddress.createUnresolved( "cajrifqkvalh2ooa.onion", 8333 ),
+                                                       InetSocketAddress.createUnresolved( "bk7yp6epnmcllq72.onion", 8333 )
+            };
```
ManfredKarrer, line 69: From where did u get the seed onion addresses? Are there more or just 2?

core/src/main/java/io/bitsquare/btc/WalletService.java
```diff
@@ -267,9 +267,35 @@ public void onTransaction(Peer peer, Transaction t) {
...
+        log.error( "seedNodes: " + seedNodes.toString() );
+        
         // Pass custom seed nodes if set in options
         if (seedNodes != null && !seedNodes.isEmpty()) {
-            //TODO Check how to pass seed nodes to the wallet kit. Probably via walletAppKit.setPeerNodes
+            
+            // todo: this parsing should be more robust,
+            // give validation error if needed.
+            // also: it seems the peer nodes can be overridden in the case
+            // of regtest mode below.  is that wanted?
```
ManfredKarrer, line 278: Yes, because Regtest uses localhost. I will add a comment there.

core/src/main/java/io/bitsquare/btc/WalletService.java
```diff
@@ -272,10 +272,11 @@ public void onTransaction(Peer peer, Transaction t) {
...
 
-        if (useTor && params.getId().equals(NetworkParameters.ID_MAINNET))
-            walletAppKit.useTor();
-            
-        
+        // We do not call walletAppKit.useTor() anymore because that would turn
+        // on orchid Tor, which we do not want.  Instead, we create a Tor proxy
+        // later.
+        // if (useTor && params.getId().equals(NetworkParameters.ID_MAINNET))
+        //    walletAppKit.useTor();
```
ManfredKarrer, line 315: Yes, I just left it to have the prog arg an effect. I will delete it.

core/src/main/java/io/bitsquare/btc/SeedPeersSocks5Dns.java
dan-da, line 69: Two sources for .onion addresses:
1. bitcoin-core source code.  See previous comment in #488 where I list them and give a pointer to source.
2. A running bitcoin-core node with onlynet=tor.  I used this to obtain "known good right now" peers for testing a few times.

PR #541, opened 2016-07-24
Use Seedpeers for default DNS Discovery when Socks proxy active. ( Tor )

dan-da:
This is a temporary measure until  SeedPeersSocks5Dns can be made to work more reliably.

This pull request relies upon bitcoinj PR for up-to-date seeds:
    https://github.com/bitsquare/bitcoinj/pull/2

Code comments are pretty explanatory.

```
        // If operating over a proxy and we haven't set any peer nodes, then
        // we want to use SeedPeers for discovery instead of the default DnsDiscovery.
        // This is only because we do not yet have a Dns discovery class that works
        // reliably over proxy/tor.
        //
        // todo: There should be a user pref called "Use Local DNS for Proxy/Tor"
        // that disables this.  In that case, the default DnsDiscovery class will
        // be used which should work, but is less private.  The aim here is to
        // be private by default when using proxy/tor.  However, the seedpeers
        // could become outdated, so it is important that the user be able to
        // disable it, but should be made aware of the reduced privacy.
        // SeedPeersSocks5Dns should replace SeedPeers once working reliably.
```

### DIFF
--- a/common/src/main/java/io/bitsquare/app/Log.java
+++ b/common/src/main/java/io/bitsquare/app/Log.java
@@ -114,6 +114,6 @@ public class Log {
         if (DevFlags.STRESS_TEST_MODE)
             System.err.println(new SimpleDateFormat("HH:mm:ss.SSS").format(new Date()) +
                     " - " + msg +
-                    " / Memory(MB): " + Profiler.getUsedMemory());
+                    " / Memory(MB): " + Profiler.getUsedMemoryInMB());
     }
 }
--- a/core/src/main/java/io/bitsquare/btc/WalletAppKitBitSquare.java
+++ b/core/src/main/java/io/bitsquare/btc/WalletAppKitBitSquare.java
@@ -64,11 +64,6 @@ public class WalletAppKitBitSquare extends WalletAppKit {
         mgr.setConnectTimeoutMillis(CONNECT_TIMEOUT_MSEC);
         peerGroup.setConnectTimeoutMillis(CONNECT_TIMEOUT_MSEC);
 
-        // This enables remote DNS lookup of peers over socks5 proxy.
-        // It is slower, but more private.
-        // This could be turned into a user option.
-        this.setDiscovery( new SeedPeersSocks5Dns(socks5Proxy, params) );
-        
         return peerGroup;
     }    
 }
--- a/core/src/main/java/io/bitsquare/btc/WalletService.java
+++ b/core/src/main/java/io/bitsquare/btc/WalletService.java
@@ -39,6 +39,8 @@ import javafx.beans.property.*;
 import org.bitcoinj.core.*;
 import org.bitcoinj.crypto.DeterministicKey;
 import org.bitcoinj.crypto.KeyCrypterScrypt;
+import org.bitcoinj.kits.WalletAppKit;
+import org.bitcoinj.net.discovery.SeedPeers;
 import org.bitcoinj.params.MainNetParams;
 import org.bitcoinj.params.RegTestParams;
 import org.bitcoinj.params.TestNet3Params;
@@ -268,6 +270,7 @@ public class WalletService {
         walletAppKit.setBloomFilterFalsePositiveRate(0.00005);
 
         log.debug( "seedNodes: " + seedNodes.toString() );
+        boolean setPeerNodes = false;
         
         // Pass custom seed nodes if set in options
         if (seedNodes != null && !seedNodes.isEmpty()) {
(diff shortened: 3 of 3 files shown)


ManfredKarrer:
Thanks for the PR!
One question: In the BitcoinJ MainNetParams class it a comment that the addresses are taken from the Bitcoin src code. I tried to find that but did not (but did not search much). I assume you were aware if that and have looked up yourself. It is probably not anymore there... So just wanted to cross check...

ManfredKarrer:
And another:
I will add a prog arg: useDnsDiscovery which if set to true will ignore the 
 walletAppKit.setDiscovery( new SeedPeers( params) );
call. I will add a warnign popup with more info regarding reduced privacy as well.
I assume the hard coded seed should work for a good while. We should keep it updated each 1-2 months to see if there are many changes, but I assume there are at least 10-20 stable nodes. So I think that solution shoudl be good enough for now.

I will do soon a new release with trade statistics and will pack the tor part in as well, though have nto tested yet, but if no problems occure it will get out soon.

dan-da:
> One question: In the BitcoinJ MainNetParams class it a comment that the addresses are taken from the Bitcoin src code. I tried to find that but did not (but did not search much). I assume you were aware if that and have looked up yourself. It is probably not anymore there... So just wanted to cross check...

Yeah I found those when I went looking for onion seed peers once.  Here's the comment I made:

```
        // this list comes from
        //    https://github.com/bitcoin/bitcoin/blob/master/contrib/seeds/nodes_main.txt
        // which is used by:
        //    https://github.com/bitcoin/bitcoin/blob/master/contrib/seeds/generate-seeds.py
        // which generates output included in:
        //    https://github.com/bitcoin/bitcoin/blob/master/src/chainparamsseeds.h 
```

Somehow I totally forgot about that though when adding the ipv4 addresses.  I was just thinking that DNS should be the most up-to-date source.   Checking bitcoin source code now, I see that the list hasn't been updated since January.   We could add those in if you want.

https://github.com/bitcoin/bitcoin/blob/master/contrib/seeds/nodes_main.txt

dan-da:
> I will do soon a new release with trade statistics and will pack the tor part in as well, though have nto tested yet, but if no problems occure it will get out soon.

Remember there is still sometimes the shutdown exception where bitcoinj tries to write to already-closed socket.

ManfredKarrer:
Yes maybe its nto a mistake to add those as well. I assume the access is random so we cannot prioritize entries (your entries are more up to date).

Is that shutdown exception happing only in the tor mode or are u referring to an error u saw in the normal (live) mode as well?

dan-da:
only in Tor mode.

ManfredKarrer:
I added prog arg for btc proxy: --btcProxyAddress=localhost:8999
I override the proxy in the WalletService:
https://github.com/bitsquare/bitsquare/commit/423cb141f4d49b4df8eabb70eda9b700c43dd84e

ManfredKarrer:
I have not tested it though....

ManfredKarrer:
Seems the PeerGroup.handlePeerDeath does not check if it was shut down already before calling  startBlockChainDownloadFromPeer(newDownloadPeer);. 
Have not investigated further, though...

Jul-25 22:15:02.043 [BlockingClient network thread for /208.66.68.130:8333] ERROR o.b.n.BlockingClient: Error writing message to connection, closing connection java.net.SocketException: Socket is closed
    at java.net.Socket.getOutputStream(Socket.java:943) ~[na:1.8.0_92]
    at org.bitcoinj.net.BlockingClient.writeBytes(BlockingClient.java:141) ~[bitcoinj-core-0.13.1.5.jar:na]
    at org.bitcoinj.core.PeerSocketHandler.sendMessage(PeerSocketHandler.java:88) [bitcoinj-core-0.13.1.5.jar:na]
    at org.bitcoinj.core.Peer.blockChainDownloadLocked(Peer.java:1328) [bitcoinj-core-0.13.1.5.jar:na]
    at org.bitcoinj.core.Peer.startBlockChainDownload(Peer.java:1357) [bitcoinj-core-0.13.1.5.jar:na]
    at org.bitcoinj.core.PeerGroup.startBlockChainDownloadFromPeer(PeerGroup.java:1672) [bitcoinj-core-0.13.1.5.jar:na]
    at org.bitcoinj.core.PeerGroup.handlePeerDeath(PeerGroup.java:1474) [bitcoinj-core-0.13.1.5.jar:na]
    at org.bitcoinj.core.PeerGroup$PeerStartupListener.onPeerDisconnected(PeerGroup.java:234) [bitcoinj-core-0.13.1.5.jar:na]
    at org.bitcoinj.core.Peer$1.run(Peer.java:296) [bitcoinj-core-0.13.1.5.jar:na]
    at org.bitcoinj.utils.Threading$2.execute(Threading.java:141) [bitcoinj-core-0.13.1.5.jar:na]
    at org.bitcoinj.core.Peer.connectionClosed(Peer.java:293) [bitcoinj-core-0.13.1.5.jar:na]
    at org.bitcoinj.net.BlockingClient$1.run(BlockingClient.java:114) [bitcoinj-core-0.13.1.5.jar:na]

PR #574, opened 2016-08-09
parent directory might not exist: causes exception

arichnad:
Files.createDirectories fixes this.

exception:
Application data directory '.../.local/share/Bitsquare' could not be created
Caused by: java.nio.file.NoSuchFileException


ManfredKarrer:
Thanks.

PR #576, opened 2016-08-10
Hashperformance

mrosseel:
Did not help my pi problem but it's cleaner like this. Was not thread safe and now even less, noted in docs.

==> what's up with the spongycastle vs bouncycastle import, something on my side? Spongy is for android I read...


ManfredKarrer:
I used for a while similar code to re-use the MessageDigest but got issues, probably because of threading issues. So we would need to make it thread safe.

PR #590, opened 2016-08-27
Sort currencies in combobox.  offerbook tab and trades tab.

dan-da:
Looks better this way to me....

### DIFF
--- a/gui/src/main/java/io/bitsquare/gui/main/market/offerbook/OfferBookChartViewModel.java
+++ b/gui/src/main/java/io/bitsquare/gui/main/market/offerbook/OfferBookChartViewModel.java
@@ -194,6 +194,8 @@ class OfferBookChartViewModel extends ActivatableViewModel {
 
     public ObservableList<TradeCurrency> getTradeCurrencies() {
         final ObservableList<TradeCurrency> list = FXCollections.observableArrayList(preferences.getTradeCurrenciesAsObservable());
+
+        FXCollections.sort(list);
         list.add(new CryptoCurrency(GUIUtil.EDIT_FLAG, GUIUtil.EDIT_FLAG));
         return list;
     }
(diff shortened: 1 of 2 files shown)


ManfredKarrer:
The currency list can be edited and in future versions the user should be able to rearrange the position in the list as well. We have already more the 70 altcoins and with fiat currencies the list might grow to >100 items in future. I think the user will be only interested in a subset of currencies which he can define in the settings. 
An option to fill all currencies might be added as well there.

Also I use those 2 different icons ont he left to separate fiat with altcoins (fiat on top).
So a sorting would not work well with that.

dan-da:
1. At present, user cannot change order, and the ordering is (seemingly) random.   This strongly argues that it is better to have sorting until such time as that functionality is present.  And even then, I would argue it should be sorted until/unless user explicitly changes it.  That is something I bet few users would bother to do.  Even for a subset of currencies, sorted by default makes sense.
2. Personally I prefer having all currencies sorted and intermixed fiat/crypto.    However, it could also be fiat (sorted) then crypto (sorted).   Either methodology is better than the present unsorted/random display.

ManfredKarrer:
No it is not random, it is sorted in code by relevance (top fiat currencies, then top crypto currencies). The user currency (derived form OS) is first entry.

I don't want that the not-relevant altcoins filling up the list (some are name with A just to be at the first place in sorted lists...), when we have 70 altcoins in the list even if sorted it is cumbersome to scroll to the right one.
I think we should use a search/filter box like I use at the altcoin account setup. But that is too much effort for that release so I will keep that for later UI improvements...

There is another difference between the markets page and the app. You only display the currencies where are data (which is good), I display the user defined currency list (which gets pre filled by default selection based on relevance).

So I agree that it should be improved but I am not convinced that just adding sorting the current list is enough (also would require more changes as the list is used in several places). 
I think it would need a bit more effort to get the user exerience better, but atm I don't have resources for that.

dan-da:
I stated "seemingly random".  There is no usage/relevance information displayed per currency, thus the sorting basis is obscured from the user.    If I had stared at it long enough, I probably could have deduced the underlying logic, but I wasn't interested in that, and I doubt most people would be.   Instead, it simply looks random.

This is something reasonable people can disagree on.  Sometimes that means a user-pref is called for.

For my own usage, I will keep using the sort patch, as it is a simple change that makes the list much more pleasing to my eye, at least.

I agree a search/filter box would be nice.

If I get really motivated, I might post in the forums to see what people prefer.  My hunch is sorted by type then alpha.

ManfredKarrer:
Ok, after further consideration I got convinced and checked how to add it the easiest way to cover all usages.
https://github.com/bitsquare/bitsquare/commit/993b6d2d07fa4554f0aef203f578375f60807ea4

Sorry for my renunciative attitude, just did not wanted to get more work for the release, but was a quicky at the end... So thanks for your input!

PR #629, opened 2016-10-06
Updating build notes with info from Linux Mint 17.3

jonathancross:
Sorry if this got a little more complicated.

Unfortunately JDK 8 and  OpenJFX are a bit of a mess in Ubuntu 14.04 and Linux Mint 17.3 -- both are LTS ("Long Term Support") releases till 2019, yet it seems nobody wants to backport these packages.

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -21,43 +21,63 @@ If `java` is not found, or your version is anything less than `1.8.0_66`, then f
 
 #### 1.1 Debian based systems (Ubuntu)
 
-To install OpenJDK use:
+You can use either OpenJDK or Oracle JDK.
+
+**To install OpenJDK use:**
 
     $ sudo apt-get install openjdk-8-jdk maven libopenjfx-java
 
-To install the Oracle JDK use:
- 
+Unfortunately, Ubuntu 14.04 & Linux Mint 17.3 are missing OpenJdk 8 and OpenJFX, so this might be useful:
+
+If `openjdk-8-jdk` is not found you can add this ppa, update, then try again:
+
+    $ sudo apt-add-repository ppa:openjdk-r/ppa && sudo apt-get install openjdk-8-jdk
+
+If `libopenjfx-java` is not found you can build & install it yourself:
+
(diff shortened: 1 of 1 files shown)


doc/build.md
```diff
@@ -21,88 +21,108 @@ If `java` is not found, or your version is anything less than `1.8.0_66`, then f
...
 -----------------
 
 ### 3. Get Bitsquare source code and build a preliminary Bitsquare version
 
-You need to get the Bitsquare dependencies first as we need to copy the BountyCastle jar to the JRE directory as well as the jdkfix jar. 
+You need to get the Bitsquare dependencies first as we need to copy the BountyCastle jar to the JRE directory as well as the jdkfix jar.
 
     $ git clone https://github.com/bitsquare/bitsquare.git
     $ cd bitsquare
     $ mvn clean package -DskipTests -Dmaven.javadoc.skip=true
```
jonathancross, line 92: The same command is called later on line 142 (as part of section 8).  Are both needed?

ManfredKarrer: Yes, because initially we need to copy the BountyCastle jar to the JRE directory as well as the jdkfix jar. Then after the JDK is setup correctly we need to build the app.

ManfredKarrer: We will update soon to the new JDK (112) once it is out as it has 2 important bugfixes and we can remove then the jdkfix.jar, but I am not sure how long it takes until OpenJDK will have that included as well. But will update build file once that is out (prob. 19.10)...

jonathancross: Ah, okay, makes sense.  Cheers!

ManfredKarrer:
New jdk is out: http://www.oracle.com/technetwork/java/javase/downloads/index-jsp-138363.html

jonathancross:
Cool, I'll give it a try!

jonathancross:
Suggest we move discussion to #650

ManfredKarrer:
No jdk fix will be removed (pushed a commit to dev branch already)

PR #635, opened 2016-10-08
Fix "BountyCastle" typos to "BouncyCastle"

bitsgh:
I noticed this typo appearing a few times in the build.md walk-through. "BountyCastle" sounds like an excellent name for a bug hunting service though!


ManfredKarrer:
Upps. thanks!

PR #649, opened 2016-10-18
Development - German translation i18n

haiqu:


ManfredKarrer:
Thanks. I just adopted a small isuess in german... please update from dev branch

Issue #651, opened 2016-10-18
Warning when building on Mac OSX


jonathancross:
```
Platform: Mac OSX 10.11.6
JDK: Oracle JDK 8 version 8u102 (1.8.0_102)
```

I am trying to build from master on Mac OSX but am getting errors / warnings on startup / shutdown.
I have [tested that strong crypto support is enabled](https://github.com/jonathancross/jc-docs/tree/master/java-strong-crypto-test), checked Bouncy Castle was added in `java.security`, etc.

The app loads and **_appears_** to function properly, but I still get this warning on startup:
- WARN  i.b.c.u.Utilities: Failed to remove cryptography restrictions java.lang.IllegalAccessException: Can not set static final boolean field javax.crypto.JceSecurity.isRestricted to java.lang.Boolean

I get 10 bitcoin peers + 8 Tor peers despite the warning (and errors mentioned below), so maybe these can be ignored?

Related: #650

See complete console output attached to this issue: [console-log-errors.txt](https://github.com/bitsquare/bitsquare/files/536604/console-log-errors.txt)

ManfredKarrer:
Did you follow exactly instructions at the build.md file? Maybe you have different jdks and copied/edited on the wrong jdk? 
The cryptography restrictions need to be changes as we use strong keys. If done like in the instructions in the build file that code path should not be executed. It is just an extra care if people dont follow and then we try via reflection, but I think that does not work reliable on all systems.

jonathancross:
> Did you follow exactly instructions at the build.md file?

I can double check the instructions and try again from scratch,  will report back tomorrow.

> different jdks and copied/edited on the wrong jdk

`JAVA_HOME` does seem to work in a different way on the Mac.  This is why I created this simple test: https://github.com/jonathancross/jc-docs/tree/master/java-strong-crypto-test to be sure that larger keys were supported.

ManfredKarrer:
Hav you checked if java-version reports your jdk and if echo JAVA_HOME points to your right jdk?

jonathancross:
Yes, I checked those:

```
java -version
java version "1.8.0_102"
Java(TM) SE Runtime Environment (build 1.8.0_102-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.102-b14, mixed mode)

which java
/usr/bin/java

readlink `which java`
/System/Library/Frameworks/JavaVM.framework/Versions/Current/Commands/java

[...]

Issue #650, opened 2016-10-18
Is jdkfix still required?


jonathancross:
From what I can see, [Oracle JDK version 8u102](http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html#jdk-8u102-oth-JPR) (1.8.0_102) fixes the [SortedList bug](https://bugs.openjdk.java.net/browse/JDK-8134655).  Can we confirm if `jdkfix-0.4.9.6.jar` is still needed?

I am working on updates to the build instructions for Mac OSx and Bitsquare seems to be working fine without `jdkfix-0.4.9.6.jar`.  However, I am still getting errors in my console and would like to get these resolved to be sure they are not related.

Has anyone else built Bitsquare using the latest JDK and tried this?

Here is the console output (1 unique warning and 1 error on shutdown): [console-log-errors.txt](https://github.com/bitsquare/bitsquare/files/536735/console-log-errors.txt)

ManfredKarrer:
I think with the new version 8_u_112 announced for tomorrow it should be fixed. But had not checked yet if the 1.8.0_102 fixed it already.

" kcr Kevin Rushforth added a comment - 2016-06-03 07:55
Approved to backport to 8u-dev (8u112). "
https://bugs.openjdk.java.net/browse/JDK-8134655

The SortedList bug did not appear consistently, so it was hard to test.

The error in the log is related to the security restriction issue (https://github.com/bitsquare/bitsquare/issues/651).

jonathancross:
Thanks, I missed the `102` vs `112` detail.
Surprising that Bitsquare still worked without the jdkfix jar though...

Feel free to close if you like until we test `112`.

ManfredKarrer:
Yes the SortedList throws only in certain situations, I think mostly at updates of the offer book...

ManfredKarrer:
Hope 112 will be released soon, otherwise I will release without that.

jonathancross:
Per #629

> New jdk is out: http://www.oracle.com/technetwork/java/javase/downloads/index-jsp-138363.html

Cool, got it installed and everything seems to work (no difference from `102` that is).
Do you have a unit test for the SortedList class bug?
Or how would you suggest testing?

ManfredKarrer:
No there is no for the unit test for the SortedList class bug, but I compared the class from our jdkfix module and the new class in the jdk and they are exactly the same.

jonathancross:
Okay, great!
[...]

PR #652, opened 2016-10-18
Updates to Build notes

jonathancross:
Related: #650 & #651 

I understand that some of these changes might be a little controversial (eg: external links), so please feel free to suggest changes or edit directly if you want.
1. Mac OSX information
2. Move info on **[enable strong cryptographic cyphers](https://github.com/jonathancross/jc-docs/blob/master/java-strong-crypto-test/README.md)** to an external link + a test and some minor improvements.  Apparently this will no longer be needed in Java 9, [only a properties setting will be needed](http://stackoverflow.com/questions/39097058/jce-zip-file-for-jdk-9/39872144#39872144).
3. Some info on installing bitcoinj from fresh directory.

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -64,7 +64,13 @@ Reload the file in your shell:
 
 #### 1.2 Other systems
 
-[Download and install the latest Oracle JDK]( http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html) for your platform.
+[Download and install the latest Oracle JDK](http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html) for your platform.
+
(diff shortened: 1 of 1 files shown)


jonathancross:
I've updated this to reflect that the jdkfix step can't be removed yet.

ManfredKarrer:
Thanks a lot!
Good summary and test tool at: https://github.com/jonathancross/jc-docs/tree/master/java-strong-crypto-test
We might add the test u used to Bitsquare as well. 
Though I do a test routine for encrypting and decrypting a msg at startup (in MainViewModel) to check if all is setup correctly. There had been issues when the bouncycastle security provider was missing, so that check should cover any possibly issues.
As some people run from source code with incomplete setup we need those checks to avoid issues in the trade process.

Issue #521, opened 2016-07-07
Popup notifications


bitcoinuser:
Hi,

I would like to see a popup notification when someone add an offer to sell or buy bitcoin for Brazilian Real.

Thanks

haiqu:
I'd like to extend this to using a system sound rather than a popup, and allowing the user to choose one or the other.

Should also add notifications for an online offer taken, although I think this already has a popup.

ManfredKarrer:
If anyone can research Java support for all supported OS regarding system sound we would get a good base fro adding that.

PR #669, opened 2016-10-23
Add sound file player

haiqu:
For #521

### DIFF
--- a/gui/src/main/java/io/bitsquare/gui/util/MakeSound.java
+++ b/gui/src/main/java/io/bitsquare/gui/util/MakeSound.java
@@ -1,0 +1,73 @@
+import java.io.File;
+import java.io.IOException;
+
+import javax.sound.sampled.AudioFormat;
+import javax.sound.sampled.AudioInputStream;
+import javax.sound.sampled.AudioSystem;
+import javax.sound.sampled.DataLine;
+import javax.sound.sampled.LineUnavailableException;
+import javax.sound.sampled.SourceDataLine;
+
+public class MakeSound {
+
+    private final int BUFFER_SIZE = 128000;
+    private File soundFile;
+    private AudioInputStream audioStream;
+    private AudioFormat audioFormat;
+    private SourceDataLine sourceLine;
+
+    /**
+     * @param filename the name of the file that is going to be played
+     */
+    public void playSound(String filename){
+
+        String strFilename = filename;
+
+        try {
+            soundFile = new File(strFilename);
+        } catch (Exception e) {
+            e.printStackTrace();
+            System.exit(1);
+        }
+
+        try {
+            audioStream = AudioSystem.getAudioInputStream(soundFile);
+        } catch (Exception e){
+            e.printStackTrace();
+            System.exit(1);
+        }
+
+        audioFormat = audioStream.getFormat();
+
+        DataLine.Info info = new DataLine.Info(SourceDataLine.class, audioFormat);
+        try {
+            sourceLine = (SourceDataLine) AudioSystem.getLine(info);
+            sourceLine.open(audioFormat);
+        } catch (LineUnavailableException e) {
+            e.printStackTrace();
+            System.exit(1);
+        } catch (Exception e) {
+            e.printStackTrace();
+            System.exit(1);
+        }
+
+        sourceLine.start();
+
+        int nBytesRead = 0;
+        byte[] abData = new byte[BUFFER_SIZE];
+        while (nBytesRead != -1) {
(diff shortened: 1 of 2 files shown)


gui/src/main/java/io/bitsquare/gui/util/MakeSound.java
```diff
@@ -0,0 +1,75 @@
+package io.bitsquare.gui.util;
```
mrosseel, line 18: since there's nothing gui specific here maybe this should be added to the common module? It's not unthinkable that a headless user might appreciate a beep in certain situations?

gui/src/main/java/io/bitsquare/gui/util/MakeSound.java
```diff
@@ -0,0 +1,75 @@
...
+import javax.sound.sampled.AudioFormat;
+import javax.sound.sampled.AudioInputStream;
+import javax.sound.sampled.AudioSystem;
+import javax.sound.sampled.DataLine;
+import javax.sound.sampled.LineUnavailableException;
+import javax.sound.sampled.SourceDataLine;
+
+public class MakeSound {
+
+    private final int BUFFER_SIZE = 128000;
```
mrosseel, line 26: where does this number come from? You don't have to change it but maybe a comment would be nice to show how it was calculated.

gui/src/main/java/io/bitsquare/gui/util/MakeSound.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public void playSound(String filename){
+
+        String strFilename = filename;
+
+        try {
+            soundFile = new File(strFilename);
+        } catch (Exception e) {
+            e.printStackTrace();
+            System.exit(1);
```
mrosseel: instead of a stacktrace and an exit, just log the error. Bitsquare shouldn't crash because a sound is not playing!

gui/src/main/java/io/bitsquare/gui/util/MakeSound.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        try {
+            audioStream = AudioSystem.getAudioInputStream(soundFile);
+        } catch (Exception e){
+            e.printStackTrace();
```
mrosseel, line 46: see previous comment, just log the error

gui/src/main/java/io/bitsquare/gui/util/MakeSound.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        audioFormat = audioStream.getFormat();
+
+        DataLine.Info info = new DataLine.Info(SourceDataLine.class, audioFormat);
+        try {
+            sourceLine = (SourceDataLine) AudioSystem.getLine(info);
+            sourceLine.open(audioFormat);
+        } catch (LineUnavailableException e) {
+            e.printStackTrace();
```
mrosseel, line 56: just log the error

gui/src/main/java/io/bitsquare/gui/util/MakeSound.java
```diff
@@ -0,0 +1,75 @@
...
+            System.exit(1);
+        } catch (Exception e) {
+            e.printStackTrace();
```
mrosseel, line 58: just log the error

mrosseel:
thanks for the commit Rob!
this is just the util to be used later in the gui I suppose?

haiqu:
All comments noted, they make good sense. The file was scrounged from an internet search, it was actually a forum reply to someone who wanted to produce sounds from file. So, I have no idea how the buffer size was arrived at.

It will be utilized mainly to alert the user when the status of a trade changes. Some people may be away from the computer but desire to know when the next stage is reached. Any sound file can be used. I'm looking around for a Klaxon Horn. :)

And I hadn't even considered headless use, it would be easy to move it to the common tree. However I wonder if there is any point since the headless user is likely to be remote from the server where the sound is playing. Does VNC pull audio as well?

haiqu:
I've removed the system exits and tidied it up a bit but adding a logger is a bit beyond my current Java skills. I'll leave that to the experts. :)

Stack trace could be removed too since it's excessive information for most purposes.

If this one isn't merged soon I'll close it, I didn't create a new branch for it and I want to get on with other tasks.

mrosseel:
Hi, please wait for Manfred before closing, your comment was during the night for us :)

haiqu:
Already done. I have four other branches that need work and it was holding me up.

I can add it again later.

Issue #630, opened 2016-10-06
Address field in "Fund your wallet" should only launch external wallet via left click


jonathancross:
Unexpected mouse behavior in this Address field:
Funds -> Deposit funds -> Fund your wallet -> Address
Platform: Linux Mint 17 / Cinnamon

The following actions currently launch the external bitcoin wallet:
- Clicking "Address" field's empty white space (left and right mouse click)
- Selecting text of the bitcoin address
- Right clicking on text of the bitcoin address (to open context menu)

Expected behavior:
- Only a left-click on the address text should launch the external wallet.

---

Additional "Nice to have" (low priority) improvements:
- We might reduce the width of the Address and Amount fields so the "Copy" icon is closer to the address.
- We could then move the QR code to the right or left side to reduce the overall vertical space needed for this panel (important for laptops or smaller monitors).
- ~~Remove the label "QR-Code:"~~ - Addressed in f6b2d77d12ca18988a8e7ae15bc8f62ec5d9c522

ManfredKarrer:
Thanks for the input!

Seems JavaFX does not distinguish between left and right click with standard handlers. I use setOnMouseClicked. With more specific code it should be possible, but have no time yet to fix that now.
- Yes I agree that the copy icon is too far away, but that is used in many places and would require a bit of graphic re-design to get it right.
- QR code has only a few good resolutions. Next smaller size would be like in the create offer view. But might be re-layouted like there.
- I removed the "QR-Code:"

ManfredKarrer:
Fixed here: https://github.com/bitsquare/bitsquare/commit/f51bf75003daf36a9d47a598d75155ecd325f5e5

PR #674, opened 2016-10-25
Fix unexpected mouse behavior in AddressTextField

jonathancross:
This fixes all major items in #630
- Address text is still clickable with primary (left) button.
- Address text is now selectable.
- In empty white space, secondary mouse button (right) opens context menu.
- In empty white space, `mouseReleased` event (after click) does not trigger external bitcoin wallet URI.

Testing:
I tested `DepositView` and confirmed that both Mac and Linux worked identically before the change, then confirmed the fix on Mac (no access to Linux at the moment, but fairly confident it will also be fixed there).

I looked for other views using AddressTextField, but was not able to visually confirm because I didn't want to create bogus trades, deposits, etc.  I am happy to get some testnet coins and test all instances if needed, but I think this mouse behavioral removal is pretty safe.

### DIFF
--- a/gui/src/main/java/io/bitsquare/gui/components/AddressTextField.java
+++ b/gui/src/main/java/io/bitsquare/gui/components/AddressTextField.java
@@ -56,7 +56,6 @@ public class AddressTextField extends AnchorPane {
         textField.textProperty().bind(address);
         String tooltipText = "Open your default bitcoin wallet";
         Tooltip.install(textField, new Tooltip(tooltipText));
-        textField.setOnMouseClicked(mouseEvent -> GUIUtil.showFeeInfoBeforeExecute(this::openWallet));
         textField.focusTraversableProperty().set(focusTraversableProperty().get());
         //TODO app wide focus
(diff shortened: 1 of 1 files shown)


ManfredKarrer:
For me the click on the address text field does not open the local wallet anymore if the handler is removed. 

I checked again how to deactivate right click and found a solution:
isPrimaryButtonDown is only set onPress so we need to store it loacally to have it setOnMouseReleased....

textField.setOnMousePressed(event -> wasPrimaryButtonDown = event.isPrimaryButtonDown());
        textField.setOnMouseReleased(event -> {
            if (wasPrimaryButtonDown)
                GUIUtil.showFeeInfoBeforeExecute(AddressTextField.this::openWallet);

```
        wasPrimaryButtonDown = false;
    });
```

ManfredKarrer:
Fixed here: https://github.com/bitsquare/bitsquare/commit/f51bf75003daf36a9d47a598d75155ecd325f5e5

jonathancross:
You are right, I should have tested better.  Thanks for the fix!

PR #675, opened 2016-10-25
Simplifying DepositView popup message

jonathancross:
No code changes in this PR, however I wonder if we might do better.
Why do we show an active button to "Generate new address" and then complain if the user clicks on it?  Might be better to disable the button, then show a message on hover, eg:
"Please select an unused address from the above table."

Thoughts?


ManfredKarrer:
Yes, I agree. I just noticed that many users don't see the tooltips so the popup is more safe to cover all...

jonathancross:
I was thinking a little more about this and wonder why we have a button at all?  Can't the software generate a new address whenever it sees that all addresses have been used?  This is the way modern bitcoin wallets generally work.

ManfredKarrer:
Good idea! If you want to have a look to that it should nto be too hard. In the activate method u could check if there is no address created and if so create one. When an address gets an event that it was used we need to check again.
I should focus on other work, btu if you like to have a look to that I am happy to take a PR.

jonathancross:
> I should focus on other work...

Yes, agreed  :-)

I'll try to take a look after finishing up some other projects.

PR #683, opened 2016-10-28
Spanish translation for i18n

haiqu:
One more translated page


ManfredKarrer:
Can you please check the conflicts and remove that file causing the commit? Before committing and opening a PR you should always update so you get the latest version of the branch.

haiqu:
Sorry about the conflicting file. Once Vim gets hold of the pull it usually turns into a mess. I refuse to try learning how to use such a crude piece of software but working around it can cause me some grief.

PR #684, opened 2016-10-31
Reactivate travis builds for future automatic testing

sgeisler:
* reactivate travis
* add bitcoinj as submodule

I'm not completely shure if adding bitcoinj as a maven moule is a good idea. But it was the cleanest way to reenable travis builds.

[See the working travis CI build here.](https://travis-ci.org/sgeisler/bitsquare/builds/171997049)


mrosseel:
Did you use the correct branch of bitcoinj as detailed in the build instruction page?

$ git clone -b FixBloomFilters https://github.com/bitsquare/bitcoinj.git

sgeisler:
I will re-check, but I thought is was the default branch of the bitsquare/bitcoinj repo.

EDIT: yes, the submodule has version 206b5b, which is the newest on FixBloomFilters.

ManfredKarrer:
I would prefer to do that merge to the branch https://github.com/bitsquare/bitsquare/tree/TravisAndSubModules.
Master is used by some folks to run the release version so I don't want to break anything here or force devs to change their setup before it is not all completed.

jonathancross:
Is this now obsolete per #855 ?

ManfredKarrer:
I think so.

PR #694, opened 2016-11-07
replace instanceof with getClass on equals method

riclas:
I plan to do a full cleanup of the instanceofs in the project.
There should be some where i will need some guidance with the testing to make sure i am not breaking anything.
These two are simple enough.

### DIFF
--- a/common/src/main/java/io/bitsquare/common/crypto/DecryptedDataTuple.java
+++ b/common/src/main/java/io/bitsquare/common/crypto/DecryptedDataTuple.java
@@ -32,7 +32,7 @@ public final class DecryptedDataTuple {
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
-        if (!(o instanceof DecryptedDataTuple)) return false;
+        if (this.getClass() != o.getClass()) return false;
 
         DecryptedDataTuple that = (DecryptedDataTuple) o;
 
--- a/common/src/main/java/io/bitsquare/common/crypto/KeyRing.java
+++ b/common/src/main/java/io/bitsquare/common/crypto/KeyRing.java
@@ -55,7 +55,7 @@ public class KeyRing {
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
-        if (!(o instanceof KeyRing)) return false;
+        if (this.getClass() != o.getClass()) return false;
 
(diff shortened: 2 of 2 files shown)


riclas:
It needs an additional null check though. Manfred can i clump up all equals changes in one pr or do you prefer some split per module?

ManfredKarrer:
Whats wrong with instanceof? IntelliJ auto generates that version and I think they had good reasons why to use it, though i did not explore....

riclas:
http://stackoverflow.com/questions/596462/any-reason-to-prefer-getclass-over-instanceof-when-generating-equals

Using instanceof is a bad OOP practice in general.
Using instanceof on equals specifically can break the Object symmetry.
You can actually change the intellij default generation of equals.
http://stackoverflow.com/questions/18362721/is-it-possible-to-change-intellijs-code-generation-template-for-equals-and-ha

You can see there other reasons for changing the default (e.g. using
getters if they make checks).

In other instances it will even save code because you don't need to do
error checking.
This is just the first step and an easy replace all on the project, if you
approve.

2016-11-07 12:48 GMT+00:00 Manfred Karrer notifications@github.com:

mrosseel:
For generated methods such as equals/hash, Lombok is another solution to consider.

ManfredKarrer:
@riclas To change that globally need extensive testing to not break anything. There might be subtle cases which are hard to test. I don't have time for such a task as too much else is on my table and I assume you don't know the project that well to be able to do a profound testing. Though after a quick look into the links it sounds reasonable to use getClass. Might consider that for new code and slowly update old not-critical parts.

riclas:
Most IDEs and Lombok use getClass.
You can actually see why this subject sparked some debate ten years ago, and why the problem is not in the use of getClass, but on the Hibernate project.
http://www.jroller.com/sjivan/entry/intellij_generated_equals_breaks_hibernate

This is quite a straightforward change. The behaviour is the same since we are not using hibernate.

Indeed i don't know the project well enough to do full testing on this.

Other changes regarding instanceof should be easier to test.

riclas:
I'll try to keep an eye for any new uses of instanceof that you create or you can ping me if you remember this talk.
I'll start doing some more interesting changes regarding this in the future.

PR #698, opened 2016-11-12
replace instanceof with getClass on equals methods

riclas:
This patch simply replaces the instanceof usage on equals() by getClass().
This was done with a replace all: !(o instanceof *)  -->  o == null || this.getClass() != o.getClass()

The code is equivalent, except for the instanceof version breaking the symmetry contract. Even though I have not checked if there are places where bitsquare relies on breaking object symmetry, I highly doubt that behaviour exists. It has actually only been a demonstrable issue on the Hibernate Java library.

One of the very old discussions over this issue on the eclipse forums: http://www.eclipsezone.com/eclipse/forums/t92613.html

### DIFF
--- a/common/src/main/java/io/bitsquare/common/FrameRateTimer.java
+++ b/common/src/main/java/io/bitsquare/common/FrameRateTimer.java
@@ -73,7 +73,7 @@ public class FrameRateTimer implements Timer, Runnable {
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
-        if (!(o instanceof FrameRateTimer)) return false;
+        if (o == null || this.getClass() != o.getClass()) return false;
 
         FrameRateTimer that = (FrameRateTimer) o;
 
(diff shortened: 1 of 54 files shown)


riclas:
I looked over all calls of equals() in bitsquare and didn't find any where this could be an issue. For instance, on Hibernate it was an issue because it serialized objects into proxies for storage/communication. I didn't find such a usage on bitsquare.

Also, there are a lot of classes on bitsquare that already use the getClass comparison.

ManfredKarrer:
@riclas There can be issues for instance when used in HashMaps or HashSets. Also the serialized objects or the verification of the p2p network objects might cause issues. Without a sound testing I cannot merge that yet. I don't have time for that testing. So we can leave it here for later but atm I need to focus on other things.

EDIT:
Also such a testing should be done first in a local network as there might be some issues with p2p network data and we don't want to pollute the network with invalid objects.

riclas:
I understand. fine by me, keep it up.
The idea of the comments was to let you know where the critical points might be, if any.

PR #701, opened 2016-11-13
refactor dispatching of dispute messages

riclas:
It should be possible to refactor the whole message system into the visitor programming pattern. For now I show how it is done with dispute messages.

### DIFF
--- a/core/src/main/java/io/bitsquare/arbitration/DisputeManager.java
+++ b/core/src/main/java/io/bitsquare/arbitration/DisputeManager.java
@@ -66,7 +66,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public class DisputeManager {
-    private static final Logger log = LoggerFactory.getLogger(DisputeManager.class);
+    public static final Logger log = LoggerFactory.getLogger(DisputeManager.class);
 
     private final TradeWalletService tradeWalletService;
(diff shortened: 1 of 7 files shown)


ManfredKarrer:
I am not a fan of using the visitor pattern here. I think it makes the code less clear and the message object gets a dependency to the DisputeManager and the relevant methods to be called there. I find it easier to follow what is going on if the dispatch is done in the DisputeManager only.
Also the ((DisputeMessage)message).dispatch(this); sounds not logical to me from the intention. 
It would need to be renamed at least to something like
((DisputeMessage)message).resolveHandler(this);

riclas:
My initial idea was to implement the dispatchers in their own classes. However, that would require a lot of variables moving from the DisputeManager to the DisputeMessages, which might not make sense for several of them. Hence i went with refactoring with the visitor pattern. This pattern is always less clear to read IMHO, but improves on the coding practices.
I'm not quite sure what the best approach is other than that... I don't like the dispatcher as it is now, but i don't think the processing should be moved out of the DisputeManager.
Maybe wait for someone else to cast their opinion on the subject. Or for a major refactoring on message handling.

riclas:
There's actually a stupidly huge bug on this (casting to DisputeMessage), so i will close the PR.
Maybe I'll try to rework the whole messaging someday.

PR #706, opened 2016-11-26
Adding italian language translation

HostFat:


ManfredKarrer:
FYI: We have now all display strings supported for translation.
https://forum.bitsquare.io/t/any-cunning-linguists-among-us/
https://www.transifex.com/bitsquare/bitsquare/

PR #710, opened 2016-12-06
Rerquest add IoP to bitsquare

hendry19901990:
Hello again,
I am from fermat, Please we would be very happy to be able to exchange on Bitsquare. 

Ticker symbol: IOP
Official coin name: Internet of People
webpage: http://www.fermat.org/
URL of the block explorer: http://explorer.fermat.community
Example address: pKbz7iRUSiUaTgh4UuwQCnc6pWZnyCGWxM
Confirmation that you have checked that the ticker does not conflict with ISO_421730: Confirmed
Social media accounts like Twitter account, subreddit, etc.: Twitter @fermat_org
GitHub : https://github.com/fermat-org

Regards.


ManfredKarrer:
Thanks a lot for your fast PR! I will be pleased to add IOP. Though there are some issues with that PR. 

We cannot use a newer version of BitcoinJ because we use Java serialisation which was removed after v 0.13.
So I cannot merge the maven dependency to bitcoinj 0.14.3.

If you want to do the address check with NetworkParameters please remove all what is not needed so the Fermat NetworkParameters class so it is a small as needed for the address check.
Please use the switch case in  AltCoinAddressValidator and not a custom Validator class. See BTC is an example how it should be used.

We cannot add for each altcoin a custom Validator. We have 120 altcoins....

PR #714, opened 2016-12-19
+ Byteball address validation

DevAleksei:


ManfredKarrer:
Thanks for the PR. I added the AGPL and move the class to another package.

PR #725, opened 2017-01-17
Address validator for XTO

bryceweiner:


ManfredKarrer:
I just saw that the test failed. Also the testclass was not in the test directory but in the src directory. I think u cannot use the check with new Address() as that only fits for Bitcoin. I commented out the switch case and test and await a new PR.
Please be sure that the next PR is correct!

PR #730, opened 2017-01-21
DNS lookups over Tor and connecting bitcoinj to .onion addresses.

dan-da:
Changes to support DNS lookups over Tor:

+ Add DnsTorLookup class that uses Tor Socks5 RESOLVE extension.

+ Add Socks5DnsDiscovery class that runs on separate threads for each lookup.

A limitation of performing DNS lookups over Tor is that only a single IP can be returned per lookup.  Also, the lookups are sometimes slow and timeout.  ( Bitcoinj sets the timeout to 5 secs in PeerGroup ).

For these reasons, DNS lookup is performed in addition to (not instead of) ipv4 seeds and .onion seeds.

Changes to support connecting to .onion addresses:

+ Add Socks5SeedOnionDiscovery class for discovery of .onion seeds.

+ Add Socks5MultiDiscovery wrapper class with modes for using ipv4 seeds, .onion seeds, dns, or all/mixed.

I tried to add a CLI option socks5DiscoveryMethods=ipv4addr,onion,dns to allow user to choose any combination of the 3 methods.  However it seemed more complicated than expected to get this information to the WalletService class where I need it.   @ManfredKarrer maybe you can assist with that, or point me in right direction.


ManfredKarrer:
After testing I did not get connection to the BTC network, so I reverted the merge. Here is the branch with the merge and some changes: https://github.com/bitsquare/bitsquare/tree/dns

dan-da:
Sorry I forgot to state  that this patch has a dependency on [this bitcoinj pr](https://github.com/bitsquare/bitcoinj/pull/5).

I built the dns merged branch, ran twice without args, and both times it connects to 10 peers, mixture of .onion and ipv4.

I'm not sure what would happen if running against an unpatched bitcoinj.  But quite probably that is the cause.

ManfredKarrer:
Ah ok. Yes not it is working. Just committed all to Dev.

PR #735, opened 2017-01-30
IBAN and BIC validation

SimonTuberlin:
Added IBAN and BIC verification according to German wikipedia on IBAN and ISO 9362.

Known issues: 
-used method Character.isLetter() does not detect decorated letters like Ä, Å, ... as invalid
-according to wikipedia some few IBANs from some banks don't yet comply with the checksum rules
-letter-to-digit transformation (A->10, B->11, ...) in IBAN validation might not work with different charsets. It should, however, work as long as the latin alphabet is contained subsequently

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,5 @@ build
 desktop.ini
 *.bat
 !package/windows/*bit*.bat
+*/target/*
+*.class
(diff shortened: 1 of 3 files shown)


ManfredKarrer:
Great thanks! Very appreciated! Do you have some references to docs on what the specs are based?

SimonTuberlin:
BIC: https://de.wikipedia.org/w/index.php?title=ISO_9362&oldid=162024671
IBAN: https://de.wikipedia.org/w/index.php?title=IBAN&oldid=161966386

They are based on ISO-9362 and ISO-13616-1, respectively, which are unfortunately not publicly available.

PR #738, opened 2017-01-31
PIVX Address Validation

ghost:


ManfredKarrer:
There is no unittests with example addresses. Please verify the that the tests work. I assume the current code is not correct.

ManfredKarrer:
As expected the test failed as u use BTC address validation (new Address(MainNetParams.get(), input);).

PR #740, opened 2017-02-02
PIVX Address Validation Test

ghost:


ManfredKarrer:
As expected the test failed. You did not include the test as the @Test annotation was missing.
I will not accept another PR, as the quality of the PR was below our requirements.

PR #743, opened 2017-02-04
Added some German translations, one Bugfix

ghost:
According to the Post of user coinstash: https://forum.bitsquare.io/t/any-cunning-linguists-among-us/721/5
I would be delighted if I could help out with German translations.

There is also one typo bugfix.


ManfredKarrer:
Thanks!

PR #744, opened 2017-02-06
Email validator

SimonTuberlin:
ChaseQuickPayValidator calls EmailValidator

EmailValidator:
Added basic rules for email validation
- only one '@' sign allowed (unless between double quotes in local part)
- local part cannot start or end with '.'
- top level domain in domain part is letters only and at least two chars long (http://www.iana.org/domains/root/db)

Rules are taken from German Wikipedia (https://de.wikipedia.org/w/index.php?title=E-Mail-Adresse&stableid=160390365) and are mainly specified in RFC 5321 and RFC 5322


ManfredKarrer:
Thanks!

ManfredKarrer:
Btw. a few unit tests would be good for the email validator.

SimonTuberlin:
Sure.

Issue #755, opened 2017-02-11
Scale GUI to fit in 1024x700 pixel


ghost:
Bitsquare 0.4.9.8 on Debian 8 x86_64, XFce WM

Having Bitsquare running in a Virtual Machine with a display of 1024x768 pixel.
1024x768 is small nowadays, but the real monitor to display the VM display (via VNC) is not much bigger, it is 1280x1024, so I cannot make the VM display bigger for visual comfort reasons, because even if the VM display has a size of 1280x800 it does not fit, because it has the same width and leads to ugly scroll bars that take away display place.

Maybe Bitsquare is fitting right now in 1024x768, but when having a window manager like XFce, the height of the display is less, because of two windows manager bars:

- one on the top for showing an application menue, open application window titles etc., which has a height of 30 pixel in my example
- one on the bottom for displaying time, logout button, memory usage graphs, application starter buttons etc., which has a height of 48 pixel in my example

So for me the GUI would fit into the display if it had a size of 1024x690 pixel (768 - 30 - 48 = 690).

But to have a clean round number 700 would be ok also, I think.

For an example, please see attached screenshot: The "Display options" is cut off and one cannot see the lowest button.

![matrica](https://cloud.githubusercontent.com/assets/8447873/22856815/b7066f7e-f099-11e6-8b0b-6e4d69874580.png)

ManfredKarrer:
There are layout problems to make it smaller as many views have no scrolling yet. It is on the todo list to add scrollbars everywhere though...

ManfredKarrer:
Fixed with PR from @undso

PR #766, opened 2017-03-05
Scale GUI to fit in 1024x700 pixel #755

undso:
Scale GUI to fit in 1024x700 pixel should now be possible for the three settings tabs.
Please try it out.


ManfredKarrer:
Great thanks again!

undso:
No problem. Are there some more screens that should get a scrollpane? Havn't found more open issues yet.

ManfredKarrer:
Yes there are some others:
Market/Offer book
Market/Trades
Portfolio/Open trades

I think that should be all.
Thanks for your support!

PR #789, opened 2017-03-29
complete the german translation from transifex

riclas:
updated the german translation file with all the strings from transifex


ManfredKarrer:
Thanks!

PR #828, opened 2017-06-04
Update Bitsquare.iss

SMH17:
https://github.com/bitsquare/bitsquare/issues/572


ManfredKarrer:
AH, thanks! Hope that fixes the missing icon issue on windows

PR #831, opened 2017-06-10
Gui scaling

sqrrm:
This adds a -forceDPI n option to scale the gui.

I've only tested on a couple of linux window managers but they seem to behave ok using the command line option. Using automatic detection didn't seem as reliable and might break completely on other systems. OS X apparently automatically scales with a factor 2 for retina displays, that's why it 'just works'.


ManfredKarrer:
Thanks for the PR and all the work you put into it!!!

Though I cannot merge that atm for various reasons:
1. We don't merge into Master non-trivial changes but use the Development branch for PRs. Master is used by people who run from source code so that would be too risky before a PR has been extensively tested.
2. We work currently on the DAO branch and are close to the release. There are tons of changes and to merge now such a big code change would postpone our release plans.
3. I would prefer to find an easier solution to fix those scaling issues. It should be handled by the JavaFX framework. Also: Does Screen.getPrimary().getDpi() not deliver correct values on the problematic linux windows managers? I did not had time to investigate further and have no VM with a problematic setup. Can you provide some setups where scaling does not work? We should first be 100% sure that there is no framework side solution before going the manual way. Getting in touch with JavaFX bug tracker/forum would be good as well. One of the core devs of JavaFX is aware of Bitsquare and might be helpful.
4. I tested quickly the patch on OSX and the main menu icons have been too small (images are using the @2x convention) as well the TAC popup got very high, exceeding the screens so being impossible to accept. 

With all that said I hope I don't demotivate you. I am aware that this was a lot of work and appreciate that a lot!
If you don't mind I would suggest following:
1. Get in touch with JavaFX devs to see if there is a solution on the framework side, that would reduce a lot the affected changes.
2. If there is no solution from there side, investigate if the Screen.getPrimary().getDpi() works reliable (prob. u did that already). Requiring users to add manually the dpi is not a great solution (though I know its better to have that option than nothing).
3. To have small test app with a few major use cases and a variety of problematic VMs would be great for testing.
4. Apply your changes to the dev branch or even better to the DAO branch.
5. I cannot guarantee that I will be able to merge and test it for the next release as it would require a lot of test effort and opens new risks on a release already full with 100s of changes. 

Follow the forum. There will be soon an announcement regarding the DAO and your work should be recognized for receiving tokens from the initial genesis distribution.

sqrrm:
Thanks for the feedback.

I agree this is a rather ugly fix so I'll take another look at the javafx side of things. My main concern was to make sure nothing that already works was broken. If this breaks on OSX running as normal that's strange, but if you set the forceDPI I expect it to be odd.

Will rebase on DAO branch and continue work from there.

ManfredKarrer:
I just created a branch for it: https://github.com/bitsquare/bitsquare/tree/UI_scaling_issue_686 Could u use that for the PR? So we are more flexible which merge to the DAO branch....

sqrrm:
Created new PR in https://github.com/bitsquare/bitsquare/pull/861 of basically the same feature set (+ option to use getDpi() from OS).

Issue #346, opened 2015-11-03
Re-activate travis


ManfredKarrer:
Travis has infrastructure updates (sudo: false) and beside that the support for javaFX seems broken 
([ERROR] location: class com.sun.javafx.tk.quantum.QuantumToolkit)
Need investigation how to fix that.

jonathancross:
Is this of use?

https://docs.travis-ci.com/user/languages/java/#Testing-Against-Multiple-JDKs

> Of note: OracleJDK 8 and JavaFX  projects may need to update to the latest available version from a repository. This can be accomplished by adding the following lines from [this issue comment](https://github.com/travis-ci/travis-ci/issues/3259#issuecomment-130860338) to your .travis.yml:

```yaml
sudo: false
addons:
  apt:
    packages:
      - oracle-java8-installer
 ```

jonathancross:
Ping @cedricwalter who expressed interest in possibly helping with this  :-)

cedricwalter:
i will look at it , which branch to create PR? is it still 'Development'?

ManfredKarrer:
Yes Development is right one. On Master is the code from the latest release (at the moment it has already more, but thats the theory ;-)).

cedricwalter:
@ManfredKarrer I need to build all dependent project first btcd-cli4j / libdohj / bitcoinj but they have no 'Development branches....and we need to build them in travis as well. I will therefore  create new branches from master except if you have any objections?

Travis do not support dependent build projects.... so i can maybe try to build All in one build. This is not pretty but it seems there is not any other way

cedricwalter:
Ok done it builds now again in travis (mvn install), but unit test are still switch off and need to be fixed. I am looking at this now

Tests run: 15, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 1.198 sec <<< FAILURE! - in org.bitcoinj.wallet.DeterministicKeyChainTest
watchingChain(org.bitcoinj.wallet.DeterministicKeyChainTest)  Time elapsed: 0.073 sec  <<< FAILURE!
org.junit.ComparisonFailure: 
expected:<...233j\3437\317q-\241![J \323\'\264s\203\314\321\v\346"
  path: 2147483648
}

Nothing bad but since tests are not running since date ???  some tests may need to be fixed/removed/... will look at each test but may need help from you. I don't see any infrastructure issue left open for now.

ManfredKarrer:
[...]

PR #855, opened 2017-07-06
Fix travis build

cedricwalter:
fixing travis build, only using maven 'install' phase for now without test running. real work is now to fix unit tests


ManfredKarrer:
Thanks a lot! There are a few other things needed at the build (see build.md) but for unit tests they are probably irrelevant.

jonathancross:
Thanks @cedricwalter!

(Issue being fixed: #346)

PR #858, opened 2017-07-08
Installing protobuf automagically

cedricwalter:
Protobuf compiler get extracted as dependency under linux, windows, macos removing the need to install it using sudo and documenting it. This is more the maven way of doing: all in pom, mvn package and it is done without more explanation except what is in pom. Build is now more portable

Middle term objective in this branch is to be able to build with mvn package without any documentation,  in intellij windows/linux/macos

this is lowering the barrier entry to do proper TDD, and avoid complicated howto run

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -11,18 +11,90 @@
 
     <artifactId>common</artifactId>
 
+    <properties>
+        <protobuf.version>3.3.0</protobuf.version>
+    </properties>
+
+    <profiles>
+        <profile>
+            <id>windows</id>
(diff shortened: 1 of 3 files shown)


ManfredKarrer:
Are you using JDK 9? I assume that "enable-strong-cipher-jdk9" only works with JDK 9.
Great you found a way to auto-install PB!

cedricwalter:
enable-strong-cipher-jdk9 is only a profile setting a -D variable so harmful :-)  i use jdk 8

PR #859, opened 2017-07-08
Remove bouncy castle copy

cedricwalter:
This branch show how to remove the need to tell the user to copy bouncy-castle in its local jre lib folder.

It consist of 2 commits which can be/are independent:
1. bouncy-castle jar are signed and can not be placed in shaded jar, maven will copy them in a lib directory that will be use by shaded.jar class-loader. This remove the need to copy the lib ib jre at the cost of shipping a lib folder at the same level of shaded.jar :-)

2. Setting unlimited cipher length: this commit can be controversial as it use reflection to hack the internal api of java BUT the code is only active for JDK 1.7/1.8 and nothing else. It was tested on windows, jdk 1.8.131. This avoid to copy the file jre policy zip file

Was is nice now is that you can git clone, mvn package and just java -jar shaded.jar to run without any effort on any os/ide


ManfredKarrer:
Great thanks! The reflection hack u use seems a bit different to what I used int the past (which did not work reliably). So lets try it until we find problems.

ManfredKarrer:
Btw: Better stick with Dev branch for PRs....

PR #861, opened 2017-07-09
GUI scaling using forceDPI flag

sqrrm:
Rebased forceDPI scaling PR on DAO branch, continued from https://github.com/bitsquare/bitsquare/pull/831

Using manual scaling for all measures is wrong in principle, it's something that should be handled by the framework. It seems javafx is not there yet but being actively worked on. For OSX it's done by just doubling the size of everything and using image@2x.png. For windows it's supposed to work but I had some trouble getting it running on my windows vm, complains about LimitedKeyStrengthException, will look at that again, guess it's the oracle jdk that needed some extra steps in the security setup. Not much is said about linux.

In this PR only the images for the navigation buttons are scaled, every other image keeps it's original size. To make it work properly I think the @2x version of images would have to be used and scaled to avoid scaling up low res images. I'll look into that if this PR is actually merged.

Normal behavior should not change if forceDPI flag is not set or set to 0. All measures are scaled by 1 and current behavior would remain. For forceDPI = 1 the dpi setting from Screen.getPrimary().getDpi() is taken if available. Using forceDPI = xx will use the a dpi setting of xx.


ManfredKarrer:
Thanks. I will need a bit of time to review it, please have patience.

sqrrm:
No rush, I imagine there are other more pressing issues with the DAO and all these days. I consider this a learning experience to get to know the bisq code. If you think it's worth merging something like this I'll go through the image scaling more in detail as well.

I still intend to look a bit more at the javafx handling of scaling on different systems but I'm not that hopeful.

sqrrm:
I've compiled bisq with JDK9 and it does seem to handle scaling with the GDK_SCALE environment variable. It only works for integer multiples though, but it's probably a better solution still. Much like how mac retina only handles a scale factor of 2x.

It's not a clean move to JDK9 however. Some stuff breaks as internal APIs are removed. To get it to compile I just replaced some offending classes but I don't know how much that breaks but it might make it impossible to make bisq compile for both java 8 and 9 without maintaining a separate branch for each.

For now, just avoid merging this PR.

ManfredKarrer:
Thanks a lot! I will look into it in the next weeks. Busy with other stuff atm...

ManfredKarrer:
@sqrrm Seems JDK 9 get released soon: http://www.java9countdown.xyz/
There is no release planned form our side soon, so there will be some time for testing JDK9 in dev environment and then we can see also if that solution works well. I will close that PR by now as it will not require a code change. Thanks for your support on it!

PR #874, opened 2017-07-17
Fix security policy

cedricwalter:
## Run Bisq with Java policy
With **Minimal change version/no java code change/no packaging changes**, do not require any java code change to `io.bisq.core.app.BisqEnvironment`

* create the shaded jar (`mvn install -DskipTests`)
* run Bisq with java security policy enabled:

`java -jar shaded -Djava.security.manager -Djava.security.debug=failure -Djava.security.policy=./bisq.policy -Dbisq.home=. -Dbisq.conf=C:\Users\cedric\Bisq`

${bisq.home} is used in Bisq.policy to located the shaded.jar,
${bisq.conf} is used in Bisq.policy to located the Bisq/bisq configuration and writable folder (tmp,backup,tor).

Which can be simplified if shaded.jar & bisq.policy are in same folder to 

for windows 
`java -Dbisq.home=. -Dbisq.conf=C:\Users\<username>\AppData\Roaming -Djava.security.manager -Djava.security.debug=failure -Djava.security.policy=bisq.policy -jar shaded.jar`

for linux:
`java -Dbisq.home=.
-Dbisq.conf=/home/<username>/.local/share
-Djava.security.manager
-Djava.security.debug=failure
-Djava.security.policy=./bisq.policy -jar shaded.jar`

for macos
`java -Dbisq.home=.
-Dbisq.conf=/User/<username>/Library/Application\ Support
-Djava.security.manager
-Djava.security.debug=failure
-Djava.security.policy=bisq.policy -jar shaded.jar`

## Creating java security policy from scratch
The good news is that i found a way to create it in less than 20 min, reviews and some manual changes are still needed, but they are trivial. 

To be done by core developer only, from a trusted source of code. How to create the java policy on the fly for an OS, let's say OSx:
Add to parent pom the following dependency

```
<dependency>
            <groupId>net.sourceforge.pro-grade</groupId>
            <artifactId>pro-grade</artifactId>
            <version>1.1.1</version>
        </dependency>
```
* create the shaded jar (`mvn install -DskipTests`)
then run 
`java -jar shaded.jar -Djava.security.manager=net.sourceforge.prograde.sm.PolicyFileGeneratorJSM     -Djava.security.policy==/path/to/initial.policy    -Dprograde.generated.policy=C:\Users\cedric\IdeaProjects\bitsquare\bitsquare\core\src\main\resources\generated-OSx.policy`

play with Bisq app and go through all processes, ensuring that most code path are executed. The policy file will grow. Stop Bisq. This create a huge file that need to be refactored a bit:

* All fixed path  `C:\Users\cedric\Bisq`   -> `${user.home}/Bisq`
* all fixed IP:8333 if valid -> *:8333
turn these
```
  permission java.io.FilePermission "C:\Users\cedric\Bisq\btc_mainnet\wallet\backup\backups_bisq_BSQ_wallet", "read";
  permission java.io.FilePermission "C:\Users\cedric\Bisq\btc_mainnet\wallet\backup\backups_bisq_BSQ_wallet\1500318964153_bisq_BSQ.wallet", "write";
```
into this 
```
  permission java.io.FilePermission "${user.home}/Bisq/btc_mainnet/wallet/backup/backups_bisq_BSQ_wallet", "read";
  permission java.io.FilePermission "${user.home}/Bisq/btc_mainnet/wallet/backup/backups_bisq_BSQ_wallet/-", "read,write,delete";
```
of course every single rule has to be reviewed
this
```
permission java.io.FilePermission "C:\Users\cedric\Bisq\btc_mainnet\wallet\bisq_BTC.wallet", "delete";
  permission java.io.FilePermission "C:\Users\cedric\Bisq\btc_mainnet\wallet\bisq_BTC.wallet", "read";
  permission java.io.FilePermission "C:\Users\cedric\Bisq\btc_mainnet\wallet\bisq_BTC.wallet", "write";
```
can be compressed to 
`permission java.io.FilePermission "C:\Users\cedric\Bisq\btc_mainnet\wallet\bisq_BTC.wallet", "read,write,delete";`

Use the excellent "String manipulations" plugin in Intellij to remove duplicate and sort line carefully


cedricwalter:
note that i lost 4 hours trying to debug why there is no valid combination in java policy if we save data of bisq in ${user.home} /AppData/Roaming/Bisq

there is not a single combination that did work, to iron differences between OS and since in java and windows 10 ${user.home} resolve properly to in my case c:\users\cedric i did change the code to not use appdata anymore (see in BisqEnvironment::defaultUserDataDir)

this code is not in any way finished nor can be merged since changing this location has many impacts: migration of data, i will still try to understand why ${Apphome} can not be used in java.policy (windows specifics)

cedricwalter:
the file bisq.policy contains a nearly working java policy when running in intellij bisq without packaging -> just maven compile is enough. For development this is a lot faster and more efficient. Note that there is still many open issues in this file :-(

cedricwalter:
Bisq.policy is now running in windows AND linux, same policy file. But in order to achieve this i found that we need to alter the file structure for a portable policy. 

i was forced to define temporary -Dbisq.home and -Dbisq.conf

The user home should be defined as a -D (recommended) or the Bisq code should set it (i dont like code setting system variable but i am open to discussion).  

I need  in policy file the location of shaded.jar also, see the  Grant clause. 
`grant codeBase "file:${bisq.home}/shaded.jar" {`
Could be also replaced with ./shaded.jar

I think it would be great if we use only one variable bisq.home with below some folder, a bit like
Apache Tomcat or Maven, we can still let the user set the conf in its user.home (again TOMCAT is a a mature example where we can get ideas.  e.g. CATALINA_BASE, CATALINA_HOME)
```
|-conf/
 |-lib/
 |-log/
 |- other /
```

The shaded jar is a bad idea, policy could be tighten a lot more if we have all jar in /lib exploded,
This would allow fine granular permissions, if you look at the policy you can see a huge grant block for the shaded.jar -> BAD, a 3rd party library with a bug/issue could use more permissions than it should

Moreover there is one rule that break everything and that need to be removed (Tor)
```
    // TODO ALARM required by linux only: not acceptable
    permission java.io.FilePermission "<<ALL FILES>>", "execute";
```

cedricwalter:
Again this PR is blocked if we continue using a unique file shaded.jar that get installed by the installer.

having an exploded file structure allow to fine tune the java policy

|-conf/
 |-lib/
 |-tor/
 |-log/
 |- other /

See how to apache tomcat did it https://github.com/apache/tomcat/blob/trunk/conf/catalina.policy

Does not mean that wallet and user file are not saved elsewhere

cbeams:
Closing as out of date (our fault, not @cedricwalter's). It would be good to revisit this stuff, though. A lot has changed since this PR was originally submitted, and moving away from the single shaded jar is a possibility for sure. @cedricwalter if you're interested in looking back into this, let us know. Thanks.

PR #885, opened 2017-07-25
add plugin to automatically verify PGP signature of all project dependencies

cedricwalter:
at build time all 3rd party dependency PGP signature will be validated, this may avoid poisoned local maven repository. In travis since vm is garbaged, this is unlikely. Verifying  signature is still a good move.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,6 +74,13 @@
                         </goals>
                     </execution>
                 </executions>
+                <configuration>
+                    <pgpverify.keyserver>hkps://hkps.pool.sks-keyservers.net</pgpverify.keyserver>
+                    <pgpverify.failNoSignature>true</pgpverify.failNoSignature>
+                    <pgpgverify.failWeakSignature>true</pgpgverify.failWeakSignature>
+                    <pgpverify.verifyPomFiles>true</pgpverify.verifyPomFiles>
+                </configuration>
+
             </plugin>
             <!--
             <plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -62,6 +62,19 @@
                     <target>1.8</target>
                 </configuration>
             </plugin>
+            <plugin>
+                <!--This plugin allow you to automatically verify PGP signature of all project dependency.-->
+                <groupId>com.github.s4u.plugins</groupId>
+                <artifactId>pgpverify-maven-plugin</artifactId>
+                <version>1.0.0</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
             <!--
             <plugin>
                 <groupId>org.xolstice.maven.plugins</groupId>


ManfredKarrer:
Ah great. How does it get the pub keys? How does it deal with libs which don't provide sigs/keys?

cedricwalter:
it get public key from 
- pgpverify.keyserver   public key from hkps://hkps.pool.sks-keyservers.net, can be configured
- do not fail if signature is missing: pgpverify.failNoSignature = false
- do not fail if any dependency has a weak signature. pgpgverify.failWeakSignature = false
- Verify pom files also. pgpverify.verifyPomFiles = true

see also https://github.com/s4u/pgpverify-maven-plugin/blob/master/src/main/java/com/github/s4u/plugins/PGPVerifyMojo.java

cedricwalter:
it get public key from
* pgpverify.keyserver public key from hkps://hkps.pool.sks-keyservers.net, can be configured
* do not fail if signature is missing: pgpverify.failNoSignature = false
* do not fail if any dependency has a weak signature. pgpgverify.failWeakSignature = false
* Verify pom files also. pgpverify.verifyPomFiles = true

see https://github.com/s4u/pgpverify-maven-plugin/blob/master/src/main/java/com/github/s4u/plugins/PGPVerifyMojo.java

ManfredKarrer:
Ah ok. Lets try it with strict rules and see which are failing.

cedricwalter:
default value added, set to failed if no or weak signature found. No issues detected

```
            <configuration>
                    <pgpverify.keyserver>hkps://hkps.pool.sks-keyservers.net</pgpverify.keyserver>
                    <pgpverify.failNoSignature>true</pgpverify.failNoSignature>
                    <pgpgverify.failWeakSignature>true</pgpgverify.failWeakSignature>
                    <pgpverify.verifyPomFiles>true</pgpverify.verifyPomFiles>
                </configuration>
```

mrosseel:
As discussed, let's have a look if the enforcer-plugin can be removed as a result of this PR.
Pros/cons of both solutions, testing it for various failure scenarios.

cedricwalter:
**enforcer-plugin:**
Do not protect against Cross Build Injection (XBI). Bascially it is a fancy name for smuggling bad code into an application at build time via automated build systems such as ant, maven or ivy.
Maven provide hashes on remote repository, but those hashes are stored in the same location as the artifacts. So once a malicious hacker compromises the repository and can replace the artifact they can replace the hashes as well. Bitcoin4j uses this plugin https://github.com/gary-rowe/BitcoinjEnforcerRules, 

-> For every update of dependencies, you need to update the hashes in enforcer plugin configuration

**pgpverify-maven-plugin:**
For each dependency the plugin retrieves the corresponding PGP signature file (.asc), extracts the public key ID from the signature file, contacts the key server (default is hkp://pool.sks-keyservers.net) to pull the public key, and finally uses the signature and public key to verify the downloaded artifact's integrity. –

Possible attack a MITM attacker could sign the modified jar with their private key and upload their public key to the key server (may not be possible), causing the modified signature to verify successfully.

-> For every update of dependencies, you do not need to update anything as long as your new artifacts if PGP signed by author.

In facts both solution can be combined, a hash check do not tell anything about who produced the binary, while the pgp signature if a proof of who produce that binary.

jonathancross:
What are the next steps on this PR?

cbeams:
Closing as out of date (our fault, not @cedricwalter's). We've since moved the build from Maven to Gradle and made other significant changes to the way the build works. Note that we're now using the gradle-witness plugin to verify dependency checksums (see https://github.com/bisq-network/bisq-desktop/blob/master/build.gradle#L65-L130), but further ideas and improvements are welcome.

PR #890, opened 2017-07-26
Update Bisq.iss

ManfredKarrer:
the lines
SetupIconFile={localappdata}\Bisq.ico
UninstallDisplayIcon={localappdata}\Bisq.ico
are wrong, Bisq folder is located in {localappdata} and the icons are inside Bisq\ path
so a more coherent change would be
SetupIconFile={localappdata}\Bisq\Bisq.ico
UninstallDisplayIcon={localappdata}\Bisq\Bisq.ico
but even this would be wrong since SetupIconFile directive as it is used on compile time only and should contain the absolute path to the icon


ManfredKarrer:
Thanks, will try in next build!

PR #912, opened 2017-08-25
Update displayStrings.properties

stevenkain:
"should be happen very rare" -> "should happen very rarely"


stevenkain (commented):
Some minor reformulations

stevenkain:
Some minor reformulations in order to make better sense and be better understood

PR #914, opened 2017-08-31
update build.md with Windows downloads for git and maven

riclas:


ManfredKarrer:
The build process will be changed for the next release. So much of that file will become obsolete. For that reason I will close the PR without merging.

PR #929, opened 2017-09-15
Adds support for IOTA  altcoin: https://iota.org/

cj4c0b1:
Adds support for IOTA  altcoin: https://iota.org/


ManfredKarrer:
This is not a correct PR.

tr37ion:
Will you add it to the next version? Would be nice.

PR #942, opened 2017-09-24
Chinese translation 30%

kdaye:


ManfredKarrer:
Ah thanks. Could you add that to https://www.transifex.com/bitsquare/bitsquare? We handle the translations there.

kdaye:
Ok,I have send an application to join it just now

PR #958, opened 2017-10-06
Add a paddedId and use it for String concatenation

### LINKED ISSUE #957: Spacing issue displaying Trade ID in Buyer's "Start Payment" prompt
Sorry, I didn't capture a screenshot when I saw the problem, and that I'm too lazy to fire up the client on the test network to reproduce the steps to show the mangled text in the UI popup. But, here's the problem:

When you are displaying the instructions to the buyer of bitcoin who is using the ClearX/Zelle payment account. The strings are being concatenated without any whitespace padding around the "id" of the transaction.

So, you get text like
```
Please don't forget to add the trade IDFYLDYJas "reason for payment" so the receiver can assign your payment to this trade.
```
instead of
```
Please don't forget to add the trade ID FYLDYJ as "reason for payment" so the receiver can assign your payment to this trade.
```

The problem is the interplay between the English i18n file:
https://github.com/bisq-network/exchange/blob/f5479eca2c061ff2825d1a351d512fe8b4364636/common/src/main/resources/i18n/displayStrings.properties#L456-L458

and the the class that makes the larger message from that file:
https://github.com/bisq-network/exchange/blob/99dc2f710a2e46cb001ade883a22292a5fa4c764/gui/src/main/java/io/bisq/gui/main/portfolio/pendingtrades/steps/buyer/BuyerStep2View.java#L346-L376

I'm not sure how or if you are testing these GUI classes. I couldn't find any test which referenced `BuyerStep2View` or `TradeStepView` specifically.

A quick look at the german translations also appears that it would have the same problem. https://github.com/bisq-network/exchange/blob/f5479eca2c061ff2825d1a351d512fe8b4364636/common/src/main/resources/i18n/displayStrings_de.properties#L452-L454

So, I would suggest that we simply wrap the id with single spaces before using it in the string concatenation in `BuyerStep2View`

If you want to assign this to me, I'll be happy to send you a pull request. If you have any hints on running the client against the test network to confirm my changes, or other ideas around visually confirming it works, I'm all ears.

eljobe:
See #957 for why this is needed.

Fixes #957

### DIFF
--- a/gui/src/main/java/io/bisq/gui/main/portfolio/pendingtrades/steps/buyer/BuyerStep2View.java
+++ b/gui/src/main/java/io/bisq/gui/main/portfolio/pendingtrades/steps/buyer/BuyerStep2View.java
@@ -334,6 +334,7 @@ public class BuyerStep2View extends TradeStepView {
             String assign = Res.get("portfolio.pending.step2_buyer.assign");
             String fees = Res.get("portfolio.pending.step2_buyer.fees");
             String id = trade.getShortId();
+            String paddedId = " " + id + " ";
             String amount = model.btcFormatter.formatVolumeWithCode(trade.getTradeVolume());
             if (paymentAccountPayload instanceof CryptoCurrencyAccountPayload)
                 //noinspection UnusedAssignment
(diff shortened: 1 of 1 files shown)


ManfredKarrer:
Thanks!

PR #959, opened 2017-10-07
Development

ManfredKarrer:
This branch contains the latest DAO phase 1 changes. Because it is not much tested yet I prefer to not merge it into master, but would like to merge it to the Bisq repo so it is more accessible if people want to run the DAO version.


cbeams:
@ManfredKarrer, it looks like there's a conflict on this branch now. I won't be able to do any sort of proper review on this, so please just go ahead and merge the branch when the conflict is resolved. That way, @mrosseel will be able to sync up to it and submit his own PR for the HTTP API work.

ManfredKarrer:
@cbeams Done. 
@mrosseel Can you update to that branch and make a PR based on that?

mrosseel:
sure, will do. One question though, cbeams had a config file to tell
intellij and other editors about whitespace rules etc, can you point me to
the file or the doc about it? Can't seem to find it anywhere and want to
make sure it's correct before the PR

never mind it's '.editorconfig' and it's only used in bisq-proposals. For some reason my intellij with the editorconfig plugin applies it also to bisq itself. So either I commit this or do a bunch of reverts :/

ManfredKarrer:
https://github.com/bisq-network/docs/blob/master/CONTRIBUTING.adoc
https://github.com/bisq-network/docs/blob/master/.editorconfig

cbeams:
> never mind it's '.editorconfig' and it's only used in bisq-proposals.

We should have a similar `.editorconfig` file in every repository, including here in bisq-network/exchange; I just haven't added one yet. For the purposes of this PR, please just make sure your code follows the conventions of whatever code it is modifying. We can introduce editorconfig (and possibly reformat stuff) here in the exchange repo at some later point. Thanks for looking after this, though!

PR #961, opened 2017-10-08
Change "until" to "by" in three places

eljobe:
This common grammatical error for German-speakers can sound
particularly strange to native English speakers. The rule
is subtle, but these two sentences mean the same thing:
  1. You have until 12am to place your drink order.
  2. You must place your drink order by 12am.


ManfredKarrer:
Thanks!

PR #964, opened 2017-10-12
Deterministic/Reproducible build of shaded.jar

cedricwalter:
It is now possible to reproduce, byte for byte, every build of Bisq. More information about reproducible builds in general are available at https://reproducible-builds.org Reproducible builds are a set of software development practices that create a verifiable path from human readable source code to the binary code used by computers

Why we want that?
* To prove that a given binary file is the result of the compilation of a given source hash
* to increase security

## How to test:
checkout this branch, and run 
`mvn clean package`
then check hashes of shaded jar
```
sha256sum gui\target\shaded.jar seednode\target\SeedNode.jar statistics/target/Statistics.jar provider/target/provider.jar
```
will return

```
0ab3c7d95df4db9af20c0c07e468220348a8e13dd0d1f2d99e4915ef0dd8318b *gui\\target\\shaded.jar
ae80419acdf208a2eae4022af710703b4848659eeca9a7d015f968a176ba87c1 seednode\\target\\SeedNode.jar
4dad338913b99bf93214d5bdcd60dc31c7bce8c7feb0b1ac6e82165a9393c6c2 statistics/target/Statistics.jar
df0310e0f2bef2bf36371f4d39865654df2e81d161c3051ff9a59cdef0c57209 *provider/target/provider.jar
```

run again 
`mvn clean package`
and you will get the same hashes again and again....

# Notes
* I use maven-enforcer-plugin  in order to compile Bisq on a given compiler version (java 1.8 and Maven 3.5.0) and to ensure all plugins dependencies have an imposed version
* a 3rd party plugin is required for now strip-jar: it removes un-reproducible bits inside the JAR (timestamp), sets ZIP format time-stamps to 0 and re-orders the files inside the ZIP so they are always in a deterministic order. 
* Apache Maven has recognized the issue, and some work is being done see https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=74682318 but it no ready yet.


ManfredKarrer:
Wow, great thanks! 
Do you have an idea if there is a solution for the binaries (dmg, exe,...) as well? I think the code signing (only on OSX atm but for Windows we would like to get also a code signing certificate) will be a problem as the signature is part of the binary package. Any idea how do deal with that?

cedricwalter:
ah yes wait deterministic is only for shaded jar files. I forgot all installers :-( let me check if this can be done :-)

cedricwalter:
up to date with development branch, PR renamed to what it does. Deterministic installed will/may be done in another PR

ManfredKarrer:
Hi @cedricwalter I locally merged with dev branch and tested on OSX but the resulting binary from the dmg file does not run because bouncycastle is not found (java.lang.NoClassDefFoundError: org/bouncycastle/openpgp/PGPKeyPair).
I created a new branch to have the merge included: https://github.com/bisq-network/exchange/tree/cedric_maven_exec
Seems the linking of the local bouncycastle lib is not applied.

EDIT:
Starting the app shows a blank screen. 
Started as well with terminal:  java -jar /Applications/Bisq.app/Contents/Java/gui-0.6.0-app.jar  there I get the errors in the console.
When I run it from Intellij it works (but I have bouncycastle in the ext folder of the jdk).

cedricwalter:
hi Manfred, i will have a look at the issue, could be a bad merge, i will digg into it.

ManfredKarrer:
Yes maybe I messed something up at merge...

mrosseel:
When the merge problems are resolved, maybe also have a look to replace some hard-coded version with maven properties. Most interesting would be maven version and java version, but also the reproducible-build-maven-plugin version is a good candidate.

I'll test as well once the merges are done.

ManfredKarrer:
@cedricwalter Do you have a mac already for testing?

cedricwalter:
@ManfredKarrer nope i was having a sierra vm running great on intel but i did move to amd 1950x and its now broken... I will try to make a hackintosh with an old intel notebook again

jonathancross:
@cedricwalter Any luck testing on a Mac?
Needs a rebase. :-)

PR #968, opened 2017-10-13
API PR part 1/2, necessary changes to bisq-exchange for the API to work

mrosseel:
- updated dependencies are so they are consistent with dropwizard dependencies on those libraries. Not 100% necessary but always better.
- Extracting certain methods from GUI classes to other classes so they're available to the API.


ManfredKarrer:
That method is wrong:

 public Coin getMakerFee(boolean isCurrencyForMakerFeeBtc) {
        return OfferUtil.getMakerFee(bsqWalletService, preferences, amount.get(), marketPriceAvailable, marketPriceMargin);
    }

Need to be:
 public Coin getMakerFee(boolean isCurrencyForMakerFeeBtc) {
        return OfferUtil.getMakerFee(isCurrencyForMakerFeeBtc, amount.get(), marketPriceAvailable, marketPriceMargin);
    }

Otherwise the isCurrencyForMakerFeeBtc param gets ignored.

PR #976, opened 2017-10-18
API PR part 2/2, adding the enableApi switch

mrosseel:
- adds --enableApi switch to the GUI
- adds dependency on bisq-api

the bisq-api dependency can be found here, on the master branch: https://github.com/mrosseel/bisq-api


ManfredKarrer:
As discussed in a voice chat we add the UI sopport to the API project to not introduce dependencies to the API into the exchange project. I will close that and looking forward for the API project to get aded.

PR #984, opened 2017-10-25
Development

ManfredKarrer:


cbeams:
Manfred and I have discussed this off-thread and I'm going ahead with the merge, though I have not actually reviewed anything here. Ideally we'll work in smaller chunks in the future.

PR #986, opened 2017-10-29
Issue559 tradeschart

sqrrm:


ManfredKarrer:
Thanks a lot. I tested a bit with live data and seems that all works as it should be!

Issue #971, opened 2017-10-14
Integrate new tor libraty to support pluggable transports


ManfredKarrer:
There is a new version [1] of the tor proxy library from the developer who has worked on the initial version [2] which supports pluggable transports.

It will require minor API adoption and thoroughly testing. With that library the firewall problems for Chinese users should be solved (Obfusc4, Meek should work to get over the GFW).
That library uses 2 other sub libs which might also require an update. They should be extracted to separate projects as well (currently they are modules inside Bisq).

[1] https://github.com/JesusMcCloud/netlayer
[2] https://github.com/bisq-network/exchange/tree/master/jtorproxy

mrosseel:
current work can be tracked in this PR: https://github.com/bisq-network/exchange/pull/987

mrosseel:
I’m having one remaining problem with the netlayer integration (all the rest is working): I can connect to onion addresses of peers, but the bitcoin seed nodes are unreachable. 
These are all clearnet addresses (haven’t tried with onion nodes). The error happens inside bitcoinj, this is the error: 
```
Nov-03 10:37:14.598 [BlockingClient network thread for dnsseed.bluematt.me/75.40.139.204:8333] ERROR o.b.n.BlockingClient: Error trying to open/read from connection: dnsseed.bluematt.me/75.40.139.204:8333: SOCKS: Connection not allowed by ruleset
```

Anyone has an idea?

ManfredKarrer:
Can you check at the API call to BitcoinJ what is passed in the current version and what is passed in the new version?

JesusMcCloud:
This error can have various reasons. It essentially means that tor cannot connect to the destination. Unfortunately this is all the semantics you get from this error message. In could even be a local firewall problem (after all, this is a new binary). It could be that the seed node is blocking connections from the exit node your current tor circuit is using.
Unfortunately I don't have any time to investigate.
However, this issue did [pop up](https://github.com/bisq-network/exchange/issues/488#issuecomment-231610686) some time in 2016, and @dan-da found a solution back then.

mrosseel:
I had already reviewed dan-da's bitcoinj commits but your linked issue explains everything in detail, should be able to resolve it using that. Thanks!

mrosseel:
after a quick read, it seems the main thing is to set the torrc option SafeSocks to 0. I found a torrc in the netlayer lib so will change it there and test.

JesusMcCloud:
[...]

PR #987, opened 2017-10-30
Integrate new netlayer library

mrosseel:
This is the first part of the integration of the new netlayer lib. (see issue #971)
These are the 3 parts:
- integrate netlayer, adapt TorNetworkNode
- add bridges UI
- clean up

Note: feel free to ignore this pull request, it's just open in case someone wants to test the new library.


ManfredKarrer:
Just tested but localhost mode does not work. The relevant code in LocalhostNetworkNode is commented out.
With the tor version Bitcoin nodes are not discovered.
There are also a few data which seems not set: "hiddenservicedir_changeme", streamID ("Foo) -  I assume that should be a random string.
Please clean up also unused code, and make sure there are no complaints from code inspection (nullable problems,...)

mrosseel:
I clearly stated that cleanup was not yet done.
Do you have any idea why the seednodes do work and the bitcoin nodes not?

ManfredKarrer:
Ah, sorry did not read the comment above... No idea yet but can look into it.

ManfredKarrer:
Merged into https://github.com/bisq-network/exchange/tree/netlayer

PR #993, opened 2017-11-04
Install package openjfx instead of libopenjfx-java

olibre:
On Ubuntu 17.10, installing `libopenjfx-java` was not sufficient
producing a build error because could not find `openjfx`.
Build succeeded after install of package `openjfx`.
Package `openjfx` depends on `libopenjfx-java` (and `openjdk-8-jre`)


ManfredKarrer:
Thanks!

PR #997, opened 2017-11-08
Make BisqAppModule public so bisq-api can use it

mrosseel:
The GUI startup executable of bisq-api reuses BisqAppModule, for this the class needs to be public.

### DIFF
--- a/gui/src/main/java/io/bisq/gui/app/BisqAppModule.java
+++ b/gui/src/main/java/io/bisq/gui/app/BisqAppModule.java
@@ -52,7 +52,7 @@ import java.io.File;
 
 import static com.google.inject.name.Names.named;
 
-class BisqAppModule extends AppModule {
+public class BisqAppModule extends AppModule {
 
     private final Stage primaryStage;
 


cbeams:
It was intentional that we made this class (and some other Guice modules like it) package-private. It really is intended to be the module specific to the Bisq GUI Application. For example, on [line 96](https://github.com/mrosseel/bisq-exchange/blob/9614fd5ac425f3652ad9f535419e0422c248b685/gui/src/main/java/io/bisq/gui/app/BisqAppModule.java#L96), it installs `guiModule`. Is that really something that the Trading API should blindly re-use? I'll let Manfred make the call here, but just want to weigh in that I'd prefer to see some duplication rather than potentially inappropriate / inefficient or otherwise problematic re-use. Maybe there is an abstract / common module base class that can be factored out here. But in any case, changes to the GUI app's Guice module should be able to be done in isolation from the Trading API or any other separate application. Basically, changes to this module shouldn't risk breaking other stuff. That's too much coupling.

Again, please forgive me if I'm missing anything here. Just saw this fly by and wanted to weigh in. I'll leave it to you and Manfred from here.

mrosseel:
hi @cbeams, this is for people who want to start up both the API and the GUI. It was decided not to put  that capability in the bisq GUI itself, so Manfred whipped up a GUI startup class which depends on this Module.  So the GUI dependency is necessary, and staying in sync with bisq changes should be as painless as possible. I see what you mean though, and maybe refactoring is the solution. 

In any case this is the only thing stopping me from releasing/promoting the API so I hope we can resolve this quickly.

ManfredKarrer:
@cbeams The Trading API project supports also a GUI app which has some additional initialization code in the BisqApp class. I did not want to add API dependencies to the Bisq project itself so we decided to duplicate the BisqApp and BiqAppMain in the API project. An alternative to not make the BisqAppModule public would be to also duplicate that class. But I don't see it that critical. The API maintainer is responsible to keep track if changes in those classes. 
Some reflection based approach might be some other alternative to avoid that code duplication but I prefer to not over-use reflection if not absolutely needed. 
I will merge but if anyone come up with better alternatives we can revert it back any time.

PR #1003, opened 2017-11-10
Add docker files + scripts to easily start a price provider with tor

mrosseel:
Contains a Dockerfile + docker-compose.yml to quickly start a new priceprovider on any host operating system where Docker is available.

The README.md details which info is needed before you can start, concretely this is a hostname+private_key for tor, and the btcaverage API keys.


cbeams:
Hey @mrosseel, having a Docker image for the pricenode would be great, but I'd prefer not to add new stuff to the provider directory under the exchange repository if we can avoid it. Not sure if you're aware, but I've already extracted that code to bisq-network/pricenode. Would be better to have it there. Would you mind if we wait on merging this (or moving it over to the pricenode repo) until @ManfredKarrer and I have a chance to chat about next steps with that new repository? I've put this on our list of stuff to discuss next time we chat.

mrosseel:
Ok, I made a PR on the other repo. 
Some remarks:
* please remove the module asap from bisq-exchange to prevent confusion
* the 21.com fix was not yet applied to the pricenode repo, see first point ;)
* my new PR still points to the legacy provider, I need to rework it to work with the gradle stuff.
Let me know when you will remove the module, I'll have it done by then.
* no issue with gradle as such, but cedric is doing a lot of maven work so I'm assuming both will be 'allowed' build systems which will exist side-by-side.

PR #1005, opened 2017-11-11
Western union integration

Nolaan:
Can create offer but cannot finish trade because contracts doesn't match


ManfredKarrer:
I added the missing parts. Also requested feedback if all is correct at https://forum.bisq.io/t/wu-added-or-bisq-fork/3194/60.

PR #1012, opened 2017-11-14
National account email mismatch

sqrrm:
Payment accounts mismatched after parsing from protopuf. So newly created account would not work, but after closing down and starting up it worked since saving includes going through protobuf.


ManfredKarrer:
Thanks for spotting that!

PR #1025, opened 2017-11-23
Fix ever increasing height of arbitrator selection headline

sqrrm:
It seems a bit weird to bind to height and width. For the height it
causes an ever increasing pane height of the headline, at least for the
arbitrator selection.

Under account arbitrator selection, try clicking one of your languages and then click outside that list, the "Which arbitrators do you accept" label pane will grow a little every time. Not relying on the current height will avoid resizing the pane.

### DIFF
--- a/gui/src/main/java/io/bisq/gui/components/TableGroupHeadline.java
+++ b/gui/src/main/java/io/bisq/gui/components/TableGroupHeadline.java
@@ -47,7 +47,6 @@ public class TableGroupHeadline extends Pane {
         Pane bg = new StackPane();
         bg.setId("table-group-headline");
         bg.prefWidthProperty().bind(widthProperty());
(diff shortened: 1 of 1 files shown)


ManfredKarrer:
The blue border would exceed the bottom of the list if the binding is removed.
The problem is the label.setLayoutY(-8); if the 8 px offset is moved to the padding it works.
label.setPadding(new Insets(-8, 7, 0, 5)); 
Fixed in: https://github.com/bisq-network/exchange/commit/706bc19afd766ae6c0a97420ea4f3640bb3b7a48

sqrrm:
Indeed, that makes sense, thanks for fixing. Got quite annoyed with this issue while testing stuff even though it didn't really matter much.

ManfredKarrer:
Good catch, have not seen it myself ;-)

PR #1026, opened 2017-11-25
Add Bitcoin Cash, Bitcoin Gold and Bitcoin Clashic

mrosseel:
Adding missing 'airdrop' coins, so they can be transacted in a more privacy-preserving way on the Bisq exchange.

### DIFF
--- a/common/src/main/java/io/bisq/common/locale/CurrencyUtil.java
+++ b/common/src/main/java/io/bisq/common/locale/CurrencyUtil.java
@@ -87,8 +87,12 @@ public class CurrencyUtil {
         final List<CryptoCurrency> result = new ArrayList<>();
 
         // result.add(new CryptoCurrency("BSQ", "Bisq Token"));
+
         if (!baseCurrencyCode.equals("BTC"))
             result.add(new CryptoCurrency("BTC", "Bitcoin"));
+        result.add(new CryptoCurrency("BCH", "Bitcoin Cash"));
+        result.add(new CryptoCurrency("BCHC", "Bitcoin Clashic"));
+        result.add(new CryptoCurrency("BTG", "Bitcoin Gold"));
         result.add(new CryptoCurrency("BURST", "Burstcoin"));
         result.add(new CryptoCurrency("GBYTE", "Byte"));
         result.add(new CryptoCurrency("XCP", "Counterparty"));
--- a/core/src/main/java/io/bisq/core/trade/statistics/TradeStatisticsManager.java
+++ b/core/src/main/java/io/bisq/core/trade/statistics/TradeStatisticsManager.java
@@ -258,6 +258,9 @@ public class TradeStatisticsManager {
         newlyAdded.add("PART");
         // v0.6.1
         newlyAdded.add("MAD");
+        newlyAdded.add("BCH");
+        newlyAdded.add("BCHC");
+        newlyAdded.add("BTG");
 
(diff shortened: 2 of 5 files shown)


sqrrm:
Builds fine and tests don't complain.

nothingmuch:
What approach will be taken for BCH/BCHC replay protection? Bclashic has a large backlog but it appears to be processing most bcash txns

mrosseel:
@nothingmuch since the Bisq exchange never touches either BCH or BCHC, replay protection cannot be done by Bisq. I do agree that selling BCHC might have the problem that someone steals your more valuable BCH. What is your suggestion, not listing BCHC ?

nothingmuch:
I'm biased (I like Clashic, and full disclosure I've invested quite a bit of meme magic into it ;-), so I'd like to see it listed... However as far as I know doing this without special protection, the BCHC will be sent as BCH by default, as most txns are being replayed right now. BCH txns are a bit better off, since there is considerable amounts of spendable coinbase on that chain already, and btc.com has [a service for purchasing BCH only tainted coins](https://btc.com/tools/replay).

Furthermore, it will never be easy in principle, given the way that bcash hard forked away without replay protection, I think the onus would always be on the users' operating the clients, and given block times as they are on clashic right now, this might present problems for Bisq's deposit/arbitration process.

As part of trying to contribute to the https://chainspl.it project I've been trying to scope out some minimal APIs that might help in this situation, but from the point of view of Bisq users what would probably be in order is basic tools/services that people can use to check in advance which chains a txn is valid on, and the state of the UTXO sets on the two chains? Admittedly I haven't gotten very far, and I've only really been thinking about low level primitives.

At any rate, I think a noticeable warning that says that bitcoin clashic is not very actively mined, and that transacting on it is not safe if the UTXOs are also valid on bcash would probably be a good idea.

Finally, I'd like to add that I think that this is a very worthwhile opportunity for the entire ecosystem as a whole to learn how to handle these problems, as they may very well happen to more valuable chains

ManfredKarrer:
@nothingmuch We can add a warning popup that the users need to inform themselves and need to be aware of the risks. Do you know a good source where those issues are discussed?

nothingmuch:
Good technical resource, yes (chainsplit slack & associated google docs), but not a good resource for regular users, unfortunately (maybe Jimmy Song's replay protection explainer for 2x?).

However, having that kind of material is part of the chainspl.it project goals though, so I think it'd be a fruitful discussion to see how Bisq's users can be better accommodated?

ManfredKarrer:
I added here some basic info which I linked in the warning popups:
https://forum.bisq.io/t/airdrop-coins-information-thread
Anyone is invited to add more info there...

nothingmuch:
the link above now 404s for me, but this one works: https://forum.bisq.io/t/airdrop-coins-information-thread-bch-btg-bchc/3532

ManfredKarrer:
For me it works. Maybe it was temp. not available?

nothingmuch:
Hmm, I'm getting "Oops! That page doesn’t exist or is private." even when I disable all of my tin foil hat browser extensions

ManfredKarrer:
Hm, weird... Can you access the Forum main page? And browse there?

nothingmuch:
yep, that's how I found the post again and pasted the alternate link (also note that your original link *did* work for me yesterday)

ManfredKarrer:
Ah sorry, I renamed title, did not know that the forum software does not support the link with the old title anymore...

Issue #1042, opened 2017-12-04
Blurred GUI after pop-up interrupted


r8d8:
Steps to reproduce:
* Open Bisq and make any action to spawn pop-up window
* Close windows with window control (black arrow) insted of `Close` button
![bisq_popup](https://user-images.githubusercontent.com/8018746/33549379-eb53f430-d8f2-11e7-9b05-def185ff1f7f.png)

Received behaviour:
* Pop-up window closed but blured layer resist, app doesn't responds to mouse actions
![bisq_blured](https://user-images.githubusercontent.com/8018746/33549382-ee80d394-d8f2-11e7-9418-e6375e8943b3.png)

Required behaviour: 
* Closed pop-up window and blured layer should dissapear

Enviroment:
ArchLInux (kernel 4.13) + Awesome window manager
Bisq v0.6.1

ManfredKarrer:
The outer frame of the popup should not be visible. Seems that some JavaFX features are not supported on that window manager.

r8d8:
@ManfredKarrer, may I try to fix it?

ManfredKarrer:
Sure would be great!

ghost:
ArchLinux, @r8d8 please check in your settings `wmname LG3D` (`.xinitrc`) [1].

[1] https://wiki.archlinux.org/index.php/java#Impersonate_another_window_manager

r8d8:
@dulanov, `wname` doesn't helped

r8d8:
Same error on Ubuntu + AwesomeWM. Will try to tune OpenJDK.

ghost:
on vanilla GNOME is it work?

r8d8:
It works on XFCE (I haven't gnome), so looks like  trouble in awesomeWM.

ghost:
It's all interesting, because I couldn't find any issues either in original JDK and in awesome window manager issue trackers, only very old one issue:

> Issue RT-4797: <code>StageStyle.TRANSPARENT</code> is not supported on the OpenSolaris and Linux platforms.
Description: On the OpenSolaris and Linux platforms, a stage style of <code>TRANSPARENT</code> is treated as <code>UNDECORATED</code>. 
Workaround: None.

In code it simply:

```java
                stage = new Stage();
                stage.setScene(scene);
                Window window = rootScene.getWindow();
                setModality();
>>>                stage.initStyle(StageStyle.TRANSPARENT);
                stage.show();
```

[...]

PR #1070, opened 2017-12-07
Fix for handling of modal window closing

r8d8:
Related to #1042

### DIFF
--- a/gui/src/main/java/io/bisq/gui/main/overlays/Overlay.java
+++ b/gui/src/main/java/io/bisq/gui/main/overlays/Overlay.java
@@ -459,6 +459,10 @@ public abstract class Overlay<T extends Overlay> {
                 Window window = rootScene.getWindow();
                 setModality();
                 stage.initStyle(StageStyle.TRANSPARENT);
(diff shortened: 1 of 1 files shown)


ManfredKarrer:
Thanks for the PR!
I assume you have tested it with some popups that it does not break anything and that it solves the issue on those Linux Desktops, right? Usually I prefer such PR to Development branch because master is used by users who run from source but as it is a small isolated change I will merge it, once you confirm me the above.

r8d8:
@ManfredKarrer, 
tested on Arch and Ubuntu both using AwesomeWM. 
Let's make it into development, not to risk with master branch.
Will close this PR and reopen it into development, ok?

ManfredKarrer:
Yes better make it to dev. Not a high urgent bug fix anyway. thanks

ManfredKarrer:
@r8d8 No need for a new PR I will merge via console to dev

ManfredKarrer:
Merged: https://github.com/bisq-network/exchange/commit/41cf028f7ea2404d572cb3941c365295c9cac284

Issue #588, opened 2016-08-25
Incorrect ordering of Available offers


ligi:
![bitsquare_order_bug](https://cloud.githubusercontent.com/assets/111600/17970792/5024a6c8-6ad8-11e6-8007-0a09dcf8cc9b.png)

happens sometimes after startup - after clicking the header the problem corrects it self

haiqu:
There are a number of bugs associated with the initialization of the program, which correct themselves once it has been running for a couple of minutes and establishes price feeds. Seems to me if one is fixed it should resolve the whole lot.

ManfredKarrer:
Just looked into it. The reason is that the price feed comes a bit later when the list is already sorted and the percentage prices need the market price to get calculated.
An update on the list when the price response is completed should fix it, but not sure if it will make it to the next release, wanted to keep it as minimal as possible to not requiring much time for testing and to move on the work on the Bitsquare foundation...

ghost:
I tried to reproduce it and saw strange sort behavior by market column. Is it ok, or I can try to fix it?

![image](https://user-images.githubusercontent.com/39793/33529811-0631fd2e-d876-11e7-8e10-38ad0117ee67.png)

ManfredKarrer:
Ah good spot! Hm no idea why that is so strange. Maybe because the Coin name is used not the ticker? But should be alphabetically to that what the user sees... PR welcome :-)

ghost:
Unlike other columns the market column doesn't have custom comparator and JavaFX sort it using `OfferBookListItem.toString()` - also doesn't defined and provided by JDK, as a result - pseudo-random sort.

ghost:
The reported error is reproducible on the latest version (v0.6.1), I will try to fix it too:

![image](https://user-images.githubusercontent.com/39793/33727537-a182dd66-db78-11e7-9a86-b0a299bed6a8.png)

ghost:
This problem occurs for me not every start up, but only together with `N/A` values. After values were actualized - they order doesn't changed. In code (`OfferViewBook::getPriceColumn`):

```java
if (priceChangedListener == null) {
    priceChangedListener = (observable, oldValue, newValue) -> {
        if (offerBookListItem != null && offerBookListItem.getOffer().getPrice() != null) {
(2) >>>            setText(model.getPrice(offerBookListItem));
        }
    };
    model.priceFeedService.updateCounterProperty().addListener(priceChangedListener);
}
[...]

PR #1072, opened 2017-12-07
Absent of market column custom comparator in `OfferBookView`

ghost:
Related to #588.


ManfredKarrer:
Thanks for the PR!
If the changes are bit more risky than that one I prefer PR to Dev branch. But that one is fine in Master... Just mention it for upcoming PRs...

PR #1075, opened 2017-12-08
After price feed actualization re-ordering doesn't perform

### LINKED ISSUE #588: Incorrect ordering of Available offers

ghost:
Fixes #588.


ghost:
Or as alternative I can register in `OfferBookView.activate()` (remove in `deactivate()`) a listener for `model.priceFeedService.updateCounterProperty()` which will call `tableView.sort()`. May be it's more general variant instead of run it every time from scattered `setCellFactory`,

ManfredKarrer:
Yes I agree would be better to not have it triggered from inside the cell factory. I will close and happy to merge the other variant. Thanks!

PR #1091, opened 2017-12-15
Clean up CSS and extract inline css styles where applicable

ripcurlx:


ManfredKarrer:
Merged to https://github.com/bisq-network/exchange/tree/UI-improvements

ManfredKarrer:
@ripcurlx I just saw that some address or tx text fields don't have the blue color anymore.  Saw it at the trade process the deposit tx field and when creating an offer the address textfield for funding.

PR #1112, opened 2017-12-21
Fix InsufficientMoneyException when taking an offer

### LINKED ISSUE #1111: InsufficientMoneyException when taking an offer
## Description
When a user tries to take an offer with insufficient funds in her local Bisq wallet the balance is not updated properly. After a transfer of the necessary funds from an external wallet the user is able to take the offer, but the balance is not properly updated in all places which causes an InsufficientMoneyException. This prevents the user from taking the offer.

## Steps to reproduce
1. Bob creates an offer
2. Make sure that Alice has too little coins in her wallet to take the offer
3. Alice takes the offer and clicks on "Transfer funds from Bisq wallet". An warning popup is shown to update the funds. If you cancel this popup, you'll see that the "Trade wallet balance" was still updated with the too little funds.
4. Click on the Trade wallet address and transfer the funds from your external wallet
5. Click on "Transfer funds form Bisq wallet" again to fund the trade.
6. Review and confirm the trade

## Expected Result
Offer is successfully taken.

## Actual Result
BitcoinJ InsufficientMoneyException is shown.

Severity: No coins are lost and the user can try to take the offer again with the available funds in its local Bisq wallet.

ripcurlx:
Fixes #1111.

### DIFF
--- a/gui/src/main/java/io/bisq/gui/main/offer/takeoffer/TakeOfferDataModel.java
+++ b/gui/src/main/java/io/bisq/gui/main/offer/takeoffer/TakeOfferDataModel.java
@@ -372,8 +372,10 @@ class TakeOfferDataModel extends ActivatableDataModel {
     void fundFromSavingsWallet() {
         useSavingsWallet = true;
         updateBalance();
-        if (!isWalletFunded.get())
+        if (!isWalletFunded.get()) {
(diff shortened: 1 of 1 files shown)


ManfredKarrer:
Grate thanks!

PR #1119, opened 2017-12-26
Branding change on README.md for @Bisq_Network Twitter account.

sgornick:


cbeams:
Thanks, @sgornick. See the additional commits I added on top of yours.

PR #1196, opened 2018-01-15
Unify tooltip styling

ripcurlx:
This pull request should unify the tooltip handling within the client. Unfortunately I had to add at one place a manual setting of a text color style as I wasn't able to change the standard style inheritance in this case.


cbeams:
I have slated this for v0.6.5, because, upon talking with @ripcurlx, it sounds like we'll merge the `UI-improvements` branch (into which this PR was merged) soon.

PR #1256, opened 2018-01-23
Fix tooltip issues

ripcurlx:
This should fix all issues, except the extended height in transactions rows mentioned in https://github.com/bisq-network/exchange/issues/1253

### DIFF
--- a/gui/src/main/java/io/bisq/gui/bisq.css
+++ b/gui/src/main/java/io/bisq/gui/bisq.css
@@ -398,6 +398,10 @@ textfield */
     -fx-alignment: center;
 }
 
+#addressColumn {
+    -fx-alignment: left;
+}
+
 .table-view .column-header .label {
     -fx-alignment: center;
 }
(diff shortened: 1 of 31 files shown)


cbeams:
@ripcurlx, this PR is `In Progress` but unassigned. Are you in fact still working on it? If so, please assign to yourself. If it's waiting for Manfred to take action, then I'd recommend putting it in `Priorities` (if that's indeed true), and assigning him as a reviewer.

ripcurlx:
From my side everything should be fixed so far. I'll do a you suggested.

cbeams:
Slated for `v0.6.5`. This change lives only on `Development` at the moment, but we'll get that branch merged into `master` (and hopefully removed altogether) by the time we ship `v0.6.5` (if not much sooner).

ManfredKarrer:
I don't plan to merge it with 0.6.5 as it would require much more testing effort. Once merged into master all devs use it and we have longer time for testing until the next release.

PR #1258, opened 2018-01-23
dark menu bar icons prop

charlesrocket:
Dark tray icon set suggestion

bisq1
bisq2

outlined cat version doesn't look acceptable with non-retina displays, so i deiced to test with these instead


ripcurlx:
I think inverting it like that is the the better way, as it would get too fine grained on dark background using just the outline. I think I will have to look at the tray icon for light mode as well, as it has this grey shadow underneath the cat. So it doesn't look perfect when selected in light theme atm.
tray icon

ManfredKarrer:
That would require code changes also to support dark themes. Easier fix is to use a white background.

ManfredKarrer:
I will close the PR for now. Feel free to reopen it after the icon support would get full support and has been tested.

cbeams:
I've added the `was:incorrect` label, indicating that this PR was not a 'correct' PR in the sense that it was incomplete, untested, didn't have code actually to make the change that was suggested.

Nothing wrong here, by the way @charlesrocket—thanks for submitting. We're just in the process here of getting really good at letting folks know what's required for a PR to be "correct" and thus to get merged, so this is helpful in the sense that it provides an example of an "incorrect" PR.

PR #1269, opened 2018-01-24
Bsqparser test

sqrrm:
Added a simple test and did some refactoring #1268

Question on naming of non thread safe functions, is there a convention for naming functions or some kind of language construct to handle this? (see, isBsqTxUnsafe meaning the non thread safe version of isBsqTx)


cbeams:
I've slated this for the `v0.6.5` milestone, as (hopefully) we'll have the long-lived `dao` branch merged into `master` (and removed altogether) by then.

PR #1287, opened 2018-01-27
Start trading period after deposit transaction confirmation

ManfredKarrer:
Hi Chris. Could you make a review?


gui/src/main/java/io/bisq/gui/main/offer/createoffer/CreateOfferViewModel.java
```diff
@@ -198,7 +198,7 @@ protected void activate() {
                 switch (BisqEnvironment.getBaseCurrencyNetwork().getCurrencyCode()) {
                     case "BTC":
                         amount.set("0.0001");
-                        price.set("14029");
+                        price.set("11029");
```
cbeams, line 201: Why?

cbeams (approved):
Untested ACK.

Nit: The logic in the new `getTradeStartTime` would be simpler and clearer without the local `startTime` var, using `return` statements everywhere it was assigned.

Note: I've created bisq-network/docs#28 to capture style tips like the one above, and so am referencing it here.

cbeams:
Sanity-check: the review comment I left above contained a reference to bisq-network/docs#28, just as this comment does, but no reference showed up in the timeline for that review comment. Does a reference show up now?

cbeams:
It does show up now. That seems like a @GitHub bug.

gui/src/main/java/io/bisq/gui/main/offer/createoffer/CreateOfferViewModel.java
ManfredKarrer: I just adjusted to current btc price for dev testing with pre-filled values. The old value created a price> 10% distance thus require a manual entering of the price.

ManfredKarrer:
The reasons why I used the local var was because it is used in the logging. 
I personally prefer only one return at the end of a method rather then several inside the methods branches. In such simple cases like here it might be irrelevant...

PR #1290, opened 2018-01-27
Fix zero balance after restore

### LINKED ISSUE #1288: Balance is zero after restoring wallet
There are some reports that the wallet balance is not visible after restoring from seed words.
After testing it I saw that the AddressEntryList does not contain the old addresses used for receiving funds, thus the balance in the UI shows 0.

To fix that we will backup (rename) the AddressEntryList and let it recreate new at next startup. If the wallet has positive balance we add the receiving addresses to the AddressEntryList.

ManfredKarrer:
Fixes #1288.

### DIFF
--- a/common/src/main/java/io/bisq/common/storage/FileUtil.java
+++ b/common/src/main/java/io/bisq/common/storage/FileUtil.java
@@ -6,7 +6,10 @@ import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.*;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.Date;
@@ -96,8 +99,20 @@ public class FileUtil {
     }
 
     public static void deleteFileIfExists(File file) throws IOException {
-        if (file.exists() && !file.delete())
-            throw new FileNotFoundException("Failed to delete file: " + file);
+        try {
+            if (Utilities.isWindows()) {
+                final File canonical = file.getCanonicalFile();
+                if (canonical.exists() && !canonical.delete()) {
+                    throw new IOException("Failed to delete canonical file");
+                }
+            } else if (file.exists() && !file.delete()) {
+                throw new IOException("Failed to delete file");
+            }
+        } catch (Throwable t) {
+            log.error(t.toString());
+            t.printStackTrace();
+            throw new IOException(t);
+        }
     }
 
     public static void resourceToFile(String resourcePath, File destinationFile) throws ResourceNotFoundException, IOException {
--- a/common/src/main/java/io/bisq/common/storage/JsonFileManager.java
+++ b/common/src/main/java/io/bisq/common/storage/JsonFileManager.java
@@ -75,18 +75,7 @@ public class JsonFileManager {
                 printWriter = new PrintWriter(tempFile);
                 printWriter.println(json);
 
-                if (Utilities.isWindows()) {
-                    // Work around an issue on Windows whereby you can't rename over existing files.
-                    final File canonical = jsonFile.getCanonicalFile();
-                    if (canonical.exists() && !canonical.delete()) {
-                        throw new IOException("Failed to delete canonical file for replacement with save");
-                    }
-                    if (!tempFile.renameTo(canonical)) {
-                        throw new IOException("Failed to rename " + tempFile + " to " + canonical);
-                    }
-                } else if (!tempFile.renameTo(jsonFile)) {
-                    throw new IOException("Failed to rename " + tempFile + " to " + jsonFile);
-                }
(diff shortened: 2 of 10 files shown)


cbeams:
Will review asap, thanks @ManfredKarrer.

cbeams:
Here's what I did to test this:

First I tried to reproduce the issue by doing the following:

1. Created a backup of my production Bisq client (just to be safe)
2. Ran Bisq with `--appName=Bisq-Test1` from HEAD of `master` (1b4e0c3e8), i.e. without the changes in this PR
3. Went to `Account->Wallet seed`, entered my seed words and clicked `Restore wallets`

Expected behavior was that my Bisq wallet should be restored with Available balance reflecting the correct amount.

Actual behavior was that Bisq was still in the process of synchronizing blocks and while I could see activity in the log, including the balance getting updated, I could not see that information reflected in the app's "Available balance" field. I never actually saw what happens to the balance when Bisq finished syncing the chain, because as I type it is still synchronizing.

To verify that the fix works as expected, I repeated steps 1–3 above, this time with `--appName=Bisq-Test2` and from the HEAD of this PR's branch (b220c1a).

Actual behavior was that my `Available balance` in the Bisq GUI started updating dynamically along with the chain synchronizing, which is certainly a bit better experience than seeing a zero balance for an extended period of time, but is probably still confusing to an uninitiated user. I can imagine people being quite confused while their balance goes up and down during chain synchronization.

It might be better to just put the text `updating` in the Available balance window while synchronization is happening.

In both testing scenarios above, Bisq became **extremely unresponsive** during the sync process. I'm not sure if this is normal or expected, as this is actually the first time I've gone through a seed word restore. I would click on menu items and Bisq would not respond at all, sometimes for many seconds or minutes, and would then eventually respond.

This unresponsiveness makes it all the more important, I guess, that the `Available balance` readout gets updated like it does with this fix, to show the user that indeed something is happening. They also have the chain synchronization progress meter in the bottom, but it too moves very slowly.

Not an ideal experience overall, but it does seem to work, and the zero balance problem seems to be handled. I still haven't finished synchronizing in either of my test clients, so I'm going to wait until they're both done and will then merge if everything looks good.

cbeams:
On returning to my machine just now, I see both `Bisq-Test1` and `Bisq-Test2` clients frozen, with red P2P network disconnected errors in the status bar and the following `Error` being thrown in an endless loop:

```
Jan-28 12:31:30.930 [JavaFX Application Thread] ERROR i.b.g.a.BisqApp: Uncaught Exception from thread JavaFX Application Thread 
java.lang.OutOfMemoryError: GC overhead limit exceeded
Jan-28 12:31:48.954 [JavaFX Application Thread] ERROR i.b.g.a.BisqApp: throwableMessage= GC overhead limit exceeded 
Jan-28 12:32:01.962 [JavaFX Application Thread] ERROR i.b.g.a.BisqApp: throwableClass= class java.lang.OutOfMemoryError 
Jan-28 12:32:13.859 [JavaFX Application Thread] ERROR i.b.g.a.BisqApp: Stack trace:
java.lang.OutOfMemoryError: GC overhead limit exceeded
```

So for the record, I still have not completed the basic use case of restoring my Bisq wallet from seed words. I think the reason for that has nothing to do with the change proposed by this PR, however, and so I say go ahead with it. I can confirm that the behavior in `Bisq-Test2` was different. I could see its `Available balance` field update live as the underlying SPV wallet synchronized with block headers.

Incidentally, it's fun to watch the `Funds->Transactions` screen while restoring a wallet. It's a high-speed replay of your transaction history. Even the 6-part confirmation wheels spin in fast motion as the block headers whizz by.

cbeams (approved):
ACK

ManfredKarrer:
@cbeams As your arbitration app has a few 1000 txs, that is expected. BitcoinJ downloads nearly the whole blockchain from the time when your wallet was created in that case as nearly in each block there is likely a tx for you. Normal users which have maybe 100 tx wil not have such a heavy load to download and thus the resync is much faster. 

Regarding the updating of the balance:
That is similar to the tx list where you basically run fast forward the whole history. I agree it might be confusion for users but I don't see it as a major issue and prefer to not fix that for now (by deactivating the UI display at resync).

Did you had a look to the code changes if it all makes sense to you?

cbeams:
I did review the code as well. Should have mentioned that. In the future, I'd like to generally adopt Bitcoin Core's language around this stuff. We can tweak as we need, but this a good baseline, I think:

From https://bitcoincore.org/en/faq/contributing-code/:

> - `ACK` means “I have tested the code and I agree it should be merged”;
> - `NACK` means “I disagree this should be merged”, and must be accompanied by sound technical justification. NACKs without accompanying reasoning may be disregarded;
> - `utACK` means “I have not tested the code, but I have reviewed it and it looks OK, I agree it can be merged”;
> - `Concept ACK` means “I agree in the general principle of this pull request”;
> - `Nit` refers to trivial, often non-blocking issues.

With regard to `ACK` above, I would add not only that "I've tested this code", but that "I've reviewed and tested this code and agree it should be merged". That is indeed what I meant by by own `ACK` above.

PR #1294, opened 2018-01-28
Fix issue with custom btc nodes

### LINKED ISSUE #1291: If user pass btc node per prg argument nr of min nodes is not correct
See:
https://bisq.community/t/bisq-not-connecting-to-my-full-node-and-connecting-to-a-remote-node-instead/4190/8

ManfredKarrer:
Fixes #1291

### DIFF
--- a/core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
+++ b/core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
@@ -299,22 +299,34 @@ public class WalletsSetup {
     }
 
     private void configPeerNodes(Socks5Proxy socks5Proxy) {
+        boolean useTorForBitcoinJ = socks5Proxy != null;
         List<BitcoinNodes.BtcNode> btcNodeList = new ArrayList<>();
-        walletConfig.setMinBroadcastConnections((int) Math.floor(DEFAULT_CONNECTIONS * 0.8));
         switch (BitcoinNodes.BitcoinNodesOption.values()[preferences.getBitcoinNodesOptionOrdinal()]) {
             case CUSTOM:
-                String bitcoinNodesString = preferences.getBitcoinNodes();
-                if (bitcoinNodesString != null) {
+                final String bitcoinNodesString = preferences.getBitcoinNodes();
+                if (bitcoinNodesString != null && !bitcoinNodesString.isEmpty()) {
                     btcNodeList = Splitter.on(",")
                             .splitToList(StringUtils.deleteWhitespace(bitcoinNodesString))
                             .stream()
                             .filter(e -> !e.isEmpty())
                             .map(BitcoinNodes.BtcNode::fromFullAddress)
                             .collect(Collectors.toList());
+                    walletConfig.setMinBroadcastConnections((int) Math.ceil(btcNodeList.size() * 0.5));
+                    long onionAddresses = btcNodeList.stream()
+                            .filter(BitcoinNodes.BtcNode::hasOnionAddress)
+                            .count();
+                    if (onionAddresses != btcNodeList.size()) {
+                        log.warn("We do not support mixed onion and clear net addresses. We use those which represent the majority of nodes, either all clear-net or all onion addresses.");
+                        useTorForBitcoinJ = onionAddresses >= (btcNodeList.size() - onionAddresses);
+                    }
+                } else {
+                    log.warn("Custom nodes is set but no nodes are provided.");
+                    walletConfig.setMinBroadcastConnections((int) Math.floor(DEFAULT_CONNECTIONS * 0.8));
                 }
                 break;
             case PUBLIC:
                 // we keep the empty list
+                walletConfig.setMinBroadcastConnections((int) Math.floor(DEFAULT_CONNECTIONS * 0.8));
                 break;
             default:
             case PROVIDED:
@@ -326,7 +338,6 @@ public class WalletsSetup {
                 break;
         }
 
-        final boolean useTorForBitcoinJ = socks5Proxy != null;
         List<PeerAddress> peerAddressList = new ArrayList<>();
 
         // We connect to onion nodes only in case we use Tor for BitcoinJ (default) to avoid privacy leaks at
--- a/core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
+++ b/core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
@@ -86,7 +86,7 @@ public class WalletsSetup {
     private final BisqEnvironment bisqEnvironment;
     private final BitcoinNodes bitcoinNodes;
     private final int numConnectionForBtc;
-    private final boolean useAllProvidedNodes;
+    private boolean useAllProvidedNodes;
     private final String userAgent;
     private final NetworkParameters params;
     private final File walletDir;
@@ -299,7 +299,6 @@ public class WalletsSetup {
     }
 
     private void configPeerNodes(Socks5Proxy socks5Proxy) {
-        boolean useTorForBitcoinJ = socks5Proxy != null;
         List<BitcoinNodes.BtcNode> btcNodeList = new ArrayList<>();
         switch (BitcoinNodes.BitcoinNodesOption.values()[preferences.getBitcoinNodesOptionOrdinal()]) {
             case CUSTOM:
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
```diff
@@ -86,7 +86,7 @@
     private final BisqEnvironment bisqEnvironment;
     private final BitcoinNodes bitcoinNodes;
     private final int numConnectionForBtc;
-    private final boolean useAllProvidedNodes;
+    private boolean useAllProvidedNodes;
```
cbeams: Sort mutable (non-final) fields after final fields so that the reader can more easily identify what's stateful and thus likely to be problematic.

core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
```diff
@@ -300,35 +300,41 @@ private void configPeerNodesForLocalHostBitcoinNode() {
 
     private void configPeerNodes(Socks5Proxy socks5Proxy) {
         List<BitcoinNodes.BtcNode> btcNodeList = new ArrayList<>();
-        walletConfig.setMinBroadcastConnections((int) Math.floor(DEFAULT_CONNECTIONS * 0.8));
         switch (BitcoinNodes.BitcoinNodesOption.values()[preferences.getBitcoinNodesOptionOrdinal()]) {
             case CUSTOM:
-                String bitcoinNodesString = preferences.getBitcoinNodes();
-                if (bitcoinNodesString != null) {
+                final String bitcoinNodesString = preferences.getBitcoinNodes();
```
cbeams: Not sure why the final modifier was added here. Generally, final should be omitted for local variables unless there's some special need (like use within a subsequent lambda expression). It's too noisy otherwise, and IIRC we haven't made a practice of using final on local variables elsewhere in Bisq.

core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
```diff
@@ -300,35 +300,41 @@ private void configPeerNodesForLocalHostBitcoinNode() {
...
+                if (bitcoinNodesString != null && !bitcoinNodesString.isEmpty()) {
```
cbeams: As far as I can tell, it is impossible for `bitcoinNodesString` to be empty. If the user provides `--btcNodes=""` at the command line, the empty string value gets rejected at command line option parsing time.

What the user *can* do is provide `--btcNodes=","`, which is effectively an empty list of nodes, but this logic doesn't capture it, because the raw `bitcoinNodesString` is literally `,`, and this is not an empty string. This leads to a bug where the list of custom nodes in the app is literally `,`, but Bisq still ends up connecting to the public network. I didn't track down exactly why that happens, because clearly this guard logic should be reworked anyway.

core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
```diff
@@ -300,35 +300,41 @@ private void configPeerNodesForLocalHostBitcoinNode() {
...
                     btcNodeList = Splitter.on(",")
                             .splitToList(StringUtils.deleteWhitespace(bitcoinNodesString))
                             .stream()
                             .filter(e -> !e.isEmpty())
                             .map(BitcoinNodes.BtcNode::fromFullAddress)
                             .collect(Collectors.toList());
+                    walletConfig.setMinBroadcastConnections((int) Math.ceil(btcNodeList.size() * 0.5));
+                    // If Tor is set we usually only use onion nodes, but if user provides mixed clear net and onion nodes we want to use both
+                    useAllProvidedNodes = true;
```
cbeams: It looks like the semantics of `useAllProvidedNodes` is getting overloaded here, because in this context, it really means "use all _custom_ nodes" that have been specified by the user at the command line. That'll be pretty confusing for anyone who doesn't have the context of this pull request.

It's also generally unfortunate to introduce this kind of mutability, but for a quick fix, at least consider changing the name to something like `useAllSpecifiedNodes`, so it at least catches the "provided" and "custom" cases equally.

core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
```diff
@@ -300,35 +300,41 @@ private void configPeerNodesForLocalHostBitcoinNode() {
...
+                } else {
+                    log.warn("Custom nodes is set but no nodes are provided. We fall back to provided nodes option.");
+                    preferences.setBitcoinNodesOptionOrdinal(BitcoinNodes.BitcoinNodesOption.PROVIDED.ordinal());
+                    btcNodeList = bitcoinNodes.getProvidedBtcNodes();
+                    walletConfig.setMinBroadcastConnections(4);
                 }
                 break;
```
cbeams: The `else` clause says we're going to "fall back" to the provided nodes option, and it then duplicates the code in the actual `PROVIDED` case block below. Why not remove the `break` statement here, move the `PUBLIC` case above this case, and let the logic here _actually_ fall though to the `PROVIDED` case below?

You can keep the `log.warn` statement, and `setBitcoinNodesOptionOrdinal` call, but the duplicate assignment of `btcNodeList` and call to `walletConfig.setMinBroadcastConnections` could be eliminated.

core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
```diff
@@ -300,35 +300,41 @@ private void configPeerNodesForLocalHostBitcoinNode() {
...
                 // We require only 4 nodes instead of 7 (for 9 max connections) because our provided nodes
                 // are more reliable than random public nodes.
                 walletConfig.setMinBroadcastConnections(4);
                 break;
         }
 
-        final boolean useTorForBitcoinJ = socks5Proxy != null;
         List<PeerAddress> peerAddressList = new ArrayList<>();
-
+        boolean useTorForBitcoinJ = socks5Proxy != null;
```
cbeams: Not sure why the `final` modifier was removed here or why the line was shifted around. Leaving it off is fine for reasons mentioned elsewhere, but this change threw me off, made me wonder if I was missing something.

cbeams (changes requested):
NACK, see review comments.

core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
ManfredKarrer: Yes was leftover from earlier code where it was used in a lambda.

core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
ManfredKarrer: I preferred to not relay on the default value from the program arguments as this class does not know really about that.
The .filter(e -> !e.isEmpty()) should handle the case if we get a list of empty strings.
I improved the ocde via commit 9ac49bd91a2789e0db56d91a64b567a18bf9e951

ManfredKarrer: Added a test as well.

core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
ManfredKarrer: I added a new flag.

core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
ManfredKarrer: Yes, another point I felt guilty as well ;-).
I refactored it in commit 83f7cfbd8b199a0006f4ea042fe3edb52e4b9cdb

I preferred to not alter the cases because I follow the convention to use the cases as they are defined in the enum and to use the default case at the end. Also IMO using the break for logical flows make the code harder to read/understand (its for me like return statements in the middle of the code which I try to prevent as well usually).

core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
ManfredKarrer: Was from a earlier code version where I altered the flag. I reverted it to final.

ManfredKarrer:
@cbeams I changed all and it would be ready for another review round.

cbeams (approved):

cbeams:
ACK, merging now.

PR #1301, opened 2018-01-31
Add cleanup tor files button to tor network settings

### LINKED ISSUE #1300: Add button for deleting Tor directory except hidden service folder
I observed an issue with one seed node and after deleting the whole tor directory in the app dir folder (except the hidden service folder!) it seems to have resolved the issue. I assume some files there get invalidated and cause issues. 
I will add a button in settings/network to do that deletion in a safe way (not deleting the hidden service folder) to give users an easy option to fix similar issues.

ManfredKarrer:
Update (by @cbeams):

This PR fixes #1300. The following is from @ManfredKarrer's original comment there:

> I observed an issue with one seed node and after deleting the whole tor directory in the app dir folder (except the hidden service folder!) it seems to have resolved the issue. I assume some files there get invalidated and cause issues.
> I will add a button in settings/network to do that deletion in a safe way (not deleting the hidden service folder) to give users an easy option to fix similar issues.
>

----

@ManfredKarrer's original text in this PR:

Could you also review the deleteDirectory method in FileUtils. We have to be sure to not delete the hidden service folder. As far I tested it it worked... 

```
   public static void deleteDirectory(File file, File exclude) throws IOException {
        boolean excludeFileFound = false;
        if (file.isDirectory()) {
            File[] files = file.listFiles();
            if (files != null)
                for (File f : files) {
                    if (!excludeFileFound)
                        excludeFileFound = f.equals(exclude);
                    if (!f.equals(exclude))
                        deleteDirectory(f);
                }
        }
        // Finally delete main file/dir if exclude file was not found in directory
        if (!excludeFileFound && !file.equals(exclude))
            deleteFileIfExists(file);
    }
```

### DIFF
--- a/common/src/main/resources/i18n/displayStrings.properties
+++ b/common/src/main/resources/i18n/displayStrings.properties
@@ -1250,8 +1250,8 @@ torNetworkSettingWindow.restartInfo=You need to restart to apply the changes
 torNetworkSettingWindow.openTorWebPage=Open Tor project web page
 torNetworkSettingWindow.deleteFiles.header=Connection problems?
 torNetworkSettingWindow.deleteFiles.info=If you have repeated connection problems at start up, deleting outdated Tor files might help. To do that click the button below and restart afterwards.
-torNetworkSettingWindow.deleteFiles.button=Deleting outdated Tor files and shut down
-torNetworkSettingWindow.deleteFiles.success=Deleting outdated Tor files wa successful. Please restart.
+torNetworkSettingWindow.deleteFiles.button=Delete outdated Tor files and shut down
+torNetworkSettingWindow.deleteFiles.success=Outdated Tor files deleted successfully. Please restart.
 torNetworkSettingWindow.bridges.header=Is Tor blocked?
 torNetworkSettingWindow.bridges.info=If Tor is blocked by your internet provider or in your country you can try to use Tor bridges.\n\
   Visit the Tor web page at: https://bridges.torproject.org/bridges to learn more about \
--- a/common/src/main/resources/i18n/displayStrings.properties
+++ b/common/src/main/resources/i18n/displayStrings.properties
@@ -811,7 +811,7 @@ settings.net.roundTripTimeColumn=Roundtrip
 settings.net.sentBytesColumn=Sent
 settings.net.receivedBytesColumn=Received
 settings.net.peerTypeColumn=Peer type
-settings.net.openTorSettingsButton=Open Tor network settings
+settings.net.openTorSettingsButton=Open Tor settings
 
 settings.net.needRestart=You need to restart the application to apply that change.\nDo you want to do that now?
 settings.net.notKnownYet=Not known yet...
@@ -1248,7 +1248,12 @@ torNetworkSettingWindow.enterBridge=Enter one or more bridge relays (one per lin
 torNetworkSettingWindow.enterBridgePrompt=type address:port
 torNetworkSettingWindow.restartInfo=You need to restart to apply the changes
 torNetworkSettingWindow.openTorWebPage=Open Tor project web page
-torNetworkSettingWindow.info=If Tor is blocked by your internet provider or in your country you can try to use Tor bridges.\n\
+torNetworkSettingWindow.deleteFiles.header=Connection problems?
+torNetworkSettingWindow.deleteFiles.info=If you have repeated connection problems at start up, deleting outdated Tor files might help. To do that click the button below and restart afterwards.
+torNetworkSettingWindow.deleteFiles.button=Deleting outdated Tor files and shut down
+torNetworkSettingWindow.deleteFiles.success=Deleting outdated Tor files wa successful. Please restart.
+torNetworkSettingWindow.bridges.header=Is Tor blocked?
+torNetworkSettingWindow.bridges.info=If Tor is blocked by your internet provider or in your country you can try to use Tor bridges.\n\
   Visit the Tor web page at: https://bridges.torproject.org/bridges to learn more about \
   bridges and pluggable transports.
 
--- a/gui/src/main/java/io/bisq/gui/main/MainView.java
+++ b/gui/src/main/java/io/bisq/gui/main/MainView.java
@@ -37,7 +37,6 @@ import io.bisq.gui.main.market.MarketView;
 import io.bisq.gui.main.offer.BuyOfferView;
 import io.bisq.gui.main.offer.SellOfferView;
 import io.bisq.gui.main.overlays.popups.Popup;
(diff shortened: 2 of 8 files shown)


cbeams:
I'm reviewing this now, @ManfredKarrer.

common/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1248,7 +1248,12 @@ torNetworkSettingWindow.enterBridge=Enter one or more bridge relays (one per lin
 torNetworkSettingWindow.enterBridgePrompt=type address:port
 torNetworkSettingWindow.restartInfo=You need to restart to apply the changes
 torNetworkSettingWindow.openTorWebPage=Open Tor project web page
-torNetworkSettingWindow.info=If Tor is blocked by your internet provider or in your country you can try to use Tor bridges.\n\
+torNetworkSettingWindow.deleteFiles.header=Connection problems?
+torNetworkSettingWindow.deleteFiles.info=If you have repeated connection problems at start up, deleting outdated Tor files might help. To do that click the button below and restart afterwards.
+torNetworkSettingWindow.deleteFiles.button=Deleting outdated Tor files and shut down
```
cbeams: s/Deleting/Delete/

common/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1248,7 +1248,12 @@ torNetworkSettingWindow.enterBridge=Enter one or more bridge relays (one per lin
...
+torNetworkSettingWindow.deleteFiles.success=Deleting outdated Tor files wa successful. Please restart.
```
cbeams: s/wa/was/

gui/src/main/java/io/bisq/gui/main/settings/network/NetworkSettingsView.java
```diff
@@ -109,15 +110,23 @@
...
-                               FilterManager filterManager, BisqEnvironment bisqEnvironment, Clock clock, BSFormatter formatter) {
+    public NetworkSettingsView(WalletsSetup walletsSetup,
+                               P2PService p2PService,
+                               Preferences preferences,
+                               BitcoinNodes bitcoinNodes,
+                               FilterManager filterManager,
+                               BisqEnvironment bisqEnvironment,
+                               TorNetworkSettingsWindow torNetworkSettingsWindow,
+                               Clock clock,
+                               BSFormatter formatter) {
```
cbeams, line 121: I don't think we want to move to one-parameter per line parameter stacking like this. Creates a lot of vertical noise, and constructors with a zillion parameters are usually a signal of something needing to be refactored anyway. I'd prefer that we keep parameters aligned with the opening paren, but not stack one per line. Just go to the 120 margin, and create a new line as necessary.

gui/src/main/java/io/bisq/gui/main/MainViewModel.java
```diff
@@ -219,7 +219,7 @@ public MainViewModel(WalletsManager walletsManager, WalletsSetup walletsSetup,
                          DaoManager daoManager, EncryptionService encryptionService,
                          KeyRing keyRing, BisqEnvironment bisqEnvironment, FailedTradesManager failedTradesManager,
                          ClosedTradableManager closedTradableManager, AccountAgeWitnessService accountAgeWitnessService,
-                         BSFormatter formatter) {
+                         TorNetworkSettingsWindow torNetworkSettingsWindow, BSFormatter formatter) {
```
cbeams, line 222: This is a good example of adding a constructor parameter (see other comments)

gui/src/main/java/io/bisq/gui/main/MainView.java
```diff
@@ -117,11 +119,13 @@ public static void removeEffect() {
...
     private Popup<?> p2PNetworkWarnMsgPopup, btcNetworkWarnMsgPopup;
-    private static StackPane rootContainer;
 
     @SuppressWarnings("WeakerAccess")
     @Inject
-    public MainView(MainViewModel model, CachingViewLoader viewLoader, Navigation navigation, Transitions transitions,
+    public MainView(MainViewModel model,
+                    CachingViewLoader viewLoader,
+                    Navigation navigation,
+                    Transitions transitions,
```
cbeams, line 128: Avoid one per line ctor parameter stacking (see other, more detailed comment)

cbeams (changes requested):
NACK. See comments for details. In short:

 - A couple typos in user-facing messages that need to be fixed
 - A couple style nits around ctor param stacking

I tested the new functionality and confirmed that it left my `tor/hiddenservice` dir intact as expected.

As a more general note, it seems that we should have a dedicated `Tor network` view, as opposed to this sort of overloaded use of a dialog. Not saying that needs to be fixed in order to merge this, just mentioning it. Perhaps you think the same, and this is just more expedient right now, I don't know.

Also, if people are having trouble connecting in the first place, and are never getting past the splash screen, how are they going to get to these network settings at all in order to delete the Tor directory? I must be missing something.

And another thing: I have no idea why deleting the files in this directory would make a difference at all. Based on your original comments in the description of this PR, it sounds like this change is all based on a pragmatic observation that when you delete this directory, everything seems to start working. Not particularly satisfying, but I can't really argue with it either. Would be great to know what is actually making a difference here though.

gui/src/main/java/io/bisq/gui/main/settings/network/NetworkSettingsView.java
ManfredKarrer: Re typos: 
Fixed in next commit.

Re Popup: 
The popup is used when Tor cannot start up after 4 minutes at splash screens. So that is a reason why a dedicated Tor view will not work here. It is also considered as "exceptional" screen which should not be needed for normal use case. 

Re why it changes anything:
I agree it is not satisfying not knowing more details. Will be a task for our new P2P network dev. There are some consensus files which might get outdated. I just have seen it yesterday with a seed node which kept failing after repeated restarts and after deleting the Tor dir it was fine.

Re ctor params: 
I personally prefer it for readability to have one per line if its more then 3 or 4 params...
But not super strong opinion here. we can discuss... I prefer to keep it atm as I usually use it that way in most of the newer code parts.

Re deleteDirectory:
Do you see any risks in that method?

cbeams: Makes sense, re popup. I didn't realize that before. Thanks.

On `deleteDirectory` I didn't analyze it line by line, but gave it a good look and then tested it end-to-end on my machine. This is the only place it's used in the codebase, so I'd say the risk is contained for now. Would be good to have tests around it for sure, but I'm ok with it as-is now.

cbeams (approved):
ACK.

PR #1302, opened 2018-01-31
Support quick access to top nav items with Alt 1–9 keyboard shortcuts

### LINKED ISSUE #1286: Support quick access to top nav items with ⌘1–⌘9 keyboard shortcuts
This feature is about increasing speed and productivity for Bisq power users (especially @bisq-network/arbitrators!) by reducing the need to use the mouse for navigation.

This, like another recent bounty we did in #1144, is just one of many such keyboard shortcut implementations that can add up to better productivity and enjoyment when using Bisq. Another would be supporting alt+letter key combinations for navigating to submenu items (but that would be managed as a separate issue; I'm just mentioning it here for context).

image

Of course, it wouldn't be ⌘ keys on Windows and Linux, it would be whatever is idiomatic there for this kind of navigation.

Note that while there are usually only 8 top nav items, the DAO tab can be enabled with ⌘D, making a 9th.

blabno:
Fixes #1286

### DIFF
--- a/gui/src/main/java/io/bisq/gui/main/MainView.java
+++ b/gui/src/main/java/io/bisq/gui/main/MainView.java
@@ -151,17 +151,37 @@ public class MainView extends InitializableView<StackPane, MainViewModel> {
             daoButton.setManaged(false);
         }
 
-        // TODO can be removed once DAO is released
-        if (BisqEnvironment.getBaseCurrencyNetwork().isBitcoin()) {
-            UserThread.runAfter(() -> {
-                root.getScene().addEventHandler(KeyEvent.KEY_RELEASED, keyEvent -> {
-                    if (Utilities.isAltOrCtrlPressed(KeyCode.D, keyEvent)) {
+        UserThread.runAfter(() -> {
+            root.getScene().addEventHandler(KeyEvent.KEY_RELEASED, keyEvent -> {
+                // TODO can be removed once DAO is released
+                if (Utilities.isAltOrCtrlPressed(KeyCode.D, keyEvent)) {
+                    if (BisqEnvironment.getBaseCurrencyNetwork().isBitcoin()) {
                         daoButton.setVisible(true);
                         daoButton.setManaged(true);
                     }
-                });
-            }, 1);
-        }
+                } else if (Utilities.isAltOrCtrlPressed(KeyCode.DIGIT1, keyEvent)) {
+                    marketButton.fire();
+                } else if (Utilities.isAltOrCtrlPressed(KeyCode.DIGIT2, keyEvent)) {
+                    buyButton.fire();
+                } else if (Utilities.isAltOrCtrlPressed(KeyCode.DIGIT3, keyEvent)) {
+                    sellButton.fire();
+                } else if (Utilities.isAltOrCtrlPressed(KeyCode.DIGIT4, keyEvent)) {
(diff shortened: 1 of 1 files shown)


blabno:
Supports left Alt on linux. I need to obtain Mac to test that rounted sharp key.

cbeams:
I just quickly tested on mac, none of the keys work.

cbeams:
Also, @blabno, it's probably best to implement this in a way that follows the existing approach for supporting keyboard shortcuts, which you can find at https://github.com/bisq-network/exchange/blob/master/gui/src/main/java/io/bisq/gui/app/BisqApp.java#L249-L284. If you follow that convention, these keyboard shortcuts should "just work" across all platforms, or at the very least, they will work as well as other, existing, keyboard shortcuts do across all platforms.

For example, line https://github.com/bisq-network/exchange/blob/master/gui/src/main/java/io/bisq/gui/app/BisqApp.java#L261, is where the `CMD-J` shortcut is configured, which opens the emergency wallet tool.

blabno:
@cbeams I've pushed new commit using suggested approach. You can give it a try.

gui/src/main/java/io/bisq/gui/main/MainView.java
```diff
@@ -27,7 +27,11 @@
 import io.bisq.core.app.BisqEnvironment;
 import io.bisq.core.exceptions.BisqException;
 import io.bisq.gui.Navigation;
-import io.bisq.gui.common.view.*;
+import io.bisq.gui.common.view.CachingViewLoader;
+import io.bisq.gui.common.view.FxmlView;
+import io.bisq.gui.common.view.InitializableView;
+import io.bisq.gui.common.view.View;
+import io.bisq.gui.common.view.ViewLoader;
```
cbeams: Please don't expand these imports or the others that have been expanded below. The built-in IDEA settings should handle this correctly when running `Optimize Imports`. You may not be sync'd up correctly with those settings, or perhaps you're using some other IDE, in which case you'll need to take greater care to get it right manually.

In any case, please keep PRs atomic and avoid mixing in unrelated changes.

cbeams (changes requested):
NACK. The substance of the PR looks good, and I've tested it locally on my Mac, where everything works as expected. Just need to clean up the imports and we're good.

gui/src/main/java/io/bisq/gui/main/MainView.java
blabno: Sorry, I didn't see `.gitignore` ignores `.idea` but actually keeps code style.
I've optimized imports using Bisq code style.

cbeams (approved):

cbeams:
ACK, thanks @blabno!

Note that the next compensation request voting period starts tomorrow, and you have until end of day today to submit a request for this work if you'd like to. Otherwise you can always submit for a subsequent voting period, no problem.

Instructions are at https://github.com/bisq-network/docs/blob/master/dao/phase-zero.adoc#submit-a-compensation-request.

With regard to how much you should ask for, please gauge that based on how much value you believe this change adds to Bisq, the Bisq network, and its users. Not "how much time you spent on it". You can look at other compensation requests to get a gauge of things, and of course, feel free to ask around in Slack or elsewhere to see if the amount you're requesting seems reasonable to others.

PR #1306, opened 2018-01-31
Refactoring: TransactionsView

tau3:
refactoring and unit tests

### DIFF
--- a/core/src/main/java/io/bisq/core/trade/Trade.java
+++ b/core/src/main/java/io/bisq/core/trade/Trade.java
@@ -225,6 +225,7 @@ public abstract class Trade implements Tradable, Model {
 
     // Persistable
     // Immutable
+    @Nullable
     @Getter
     private final Offer offer;
     @Getter
--- a/gui/src/main/java/io/bisq/gui/GuiModule.java
+++ b/gui/src/main/java/io/bisq/gui/GuiModule.java
@@ -28,6 +28,7 @@ import io.bisq.gui.common.view.ViewFactory;
 import io.bisq.gui.common.view.ViewLoader;
 import io.bisq.gui.common.view.guice.InjectorViewFactory;
 import io.bisq.gui.main.funds.transactions.TradableRepository;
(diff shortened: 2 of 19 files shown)


cbeams:
@tau3, did everything pass for you locally? We just enabled Travis CI for pull requests earlier today, so yours is the first to actually go through the motions there. I only have a moment right now, but if you can spot what's gone wrong, it would be appreciated.

Also, please see our `CONTRIBUTING.adoc` file for details on commit message style.

cbeams:
At a very brief glance, these changes look good (assuming they actually pass and we get the Travis issues worked out as mentioned above).

@tau3, one nit: please add the AGPL license header to all new files, as seen elsewhere throughout the codebase.

cbeams:
> We just enabled Travis CI for pull requests earlier today, so yours is the first to actually go through the motions there.

We've since seen another PR branch go through CI checks successfully, FYI. The issue is likely local to these changes.

mrosseel:
Travis fails because it detected that your addition of powermock in the pom.xml has dependencies on a newer mockito-core than is already used in Bisq. Just click on the Travis details and scroll down to see the errors. This can also be checked locally by using the same command as Travis (I believe it's 'mvn install')

tau3:
@cbeams @mrosseel is it ok if i update version of mockito?

mrosseel:
since it's only used in testing I see no problem with updating the version. Do make sure all unit tests pass and no new conflicts are introduced. Best to test locally with 'mvn install'

tau3:
fixed build problem and added headers.

mrosseel:
some general remarks: I'm working on the GUI-less API version of Bisq. For this I'm always interested in GUI-agnostic refactoring of the code, meaning that some core methods are available without the javafx dependencies. GUI stuff should stay in the GUI, logic should be separated (and could move to core eventually). Of course this might require some thinking, and other people might want to chime in. My message is: be aware that there are other consumers of these classes besides the GUI, and maybe think if you can improve the refactorings to cater to that use case. 
The maintainers could also decide to merge this PR and do the above mentioned work in another PR.

PR #1311, opened 2018-02-01
Use auto focus on the first input field on all screens

### LINKED ISSUE #1208: Use auto focus on the first input field on all screens
If the user is required to enter some input on a screen we could put the focus on the first input field. As we added the functionality recently that the prompt text only is hidden as soon as the user enters the first input we wouldn't loose any information.

blabno:
Fixes #1208

### DIFF
--- a/gui/src/main/java/io/bisq/gui/main/funds/deposit/DepositView.java
+++ b/gui/src/main/java/io/bisq/gui/main/funds/deposit/DepositView.java
@@ -124,8 +124,10 @@ public class DepositView extends ActivatableView<VBox, Void> {
         tableView.setColumnResizePolicy(TableView.CONSTRAINED_RESIZE_POLICY);
         tableView.setPlaceholder(new Label(Res.get("funds.deposit.noAddresses")));
         tableViewSelectionListener = (observableValue, oldValue, newValue) -> {
-            if (newValue != null)
+            if (newValue != null) {
                 fillForm(newValue.getAddressString());
+                GUIUtil.requestFocus(amountTextField);
+            }
         };
 
         setSelectColumnCellFactory();
@@ -205,6 +207,8 @@ public class DepositView extends ActivatableView<VBox, Void> {
                 updateList();
             }
         };
+
+        GUIUtil.focusWhenAddedToScene(amountTextField);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


gui/src/main/java/io/bisq/gui/main/funds/deposit/DepositView.java
```diff
@@ -124,8 +124,10 @@ public void initialize() {
         tableView.setColumnResizePolicy(TableView.CONSTRAINED_RESIZE_POLICY);
         tableView.setPlaceholder(new Label(Res.get("funds.deposit.noAddresses")));
         tableViewSelectionListener = (observableValue, oldValue, newValue) -> {
-            if (newValue != null)
+            if (newValue != null) {
                 fillForm(newValue.getAddressString());
+                UserThread.execute(() -> amountTextField.requestFocus());
```
blabno: Although `UserThread.execute` is short I think it's hiding the meaning.
What do you think about creating method in `GUIUtil` called `executeOnNextRenderFrame`?

blabno (commented):
Question

gui/src/main/java/io/bisq/gui/main/funds/deposit/DepositView.java
ManfredKarrer: Hm... Maybe better if we rename "execute". But that is used in many places so I would prefer to do that myself to check all and have an atomic commit with only that refactoring.

blabno: I think it's fine to have an additional method in proper place (GUIUtil) that simply delegates to UserThread if that method's name tells us what's the purpose. Just in case impl changes later.

ManfredKarrer: Ok. We should use that only then for the render in next frame use case as that is GUI specific. The UserThread.execute is GUI agnostic and works for headless versions as well.

ManfredKarrer: Actually I am not sure if the UserThread.execute is needed at all cases. Can you test the different views if it works without the UserThread.execute as well? I will wait for merging the PR until that is resolved.

blabno:
I've tried all screens with and without execute and it was needed.

1 lut 2018 21:59 "Manfred Karrer" <notifications@github.com> napisał(a):

ManfredKarrer:
Ok, then I would suggest to use GUIUtil.requestFocus(Node node).

blabno:
Introduced GUIUtil.requestFocus.

ManfredKarrer (approved):
Seems all ok to me.

cbeams:
@blabno, @ManfredKarrer, there was a minor bug introduced with these changes. I've just fixed it in 8d44f75.

PR #1313, opened 2018-02-01
Use environment locale to select right locale and currency

### LINKED ISSUE #1204: Use environment locale to select right locale and currency
At the moment the default environment locale is not used/set properly which causes every user to be defaulted to `en_US`. Every new user will experience the app in English with USD selected.

**Steps to reproduce:**
- Have system settings set to e.g. `de_AT`
- Start a clean Bisq installation

**Expected Behavior:**
On first startup of Bisq I expect the app to be in German and EUR selected as currency.

**Actual Behavior:**
The app starts in English and USD is selected.

blabno:
Fixes #1204

### DIFF
--- a/common/src/main/java/io/bisq/common/GlobalSettings.java
+++ b/common/src/main/java/io/bisq/common/GlobalSettings.java
@@ -26,7 +26,7 @@ import java.util.Locale;
 
 public class GlobalSettings {
     private static boolean useAnimations = true;
-    private static Locale locale = new Locale("en", "US");
+    private static Locale locale = Locale.getDefault();
     private static final ObjectProperty<Locale> localeProperty = new SimpleObjectProperty<>(locale);
     private static TradeCurrency defaultTradeCurrency;
     private static String btcDenomination;
--- a/gui/src/main/java/io/bisq/gui/app/BisqAppMain.java
+++ b/gui/src/main/java/io/bisq/gui/app/BisqAppMain.java
@@ -25,17 +25,12 @@ import joptsimple.OptionException;
 import joptsimple.OptionParser;
 import joptsimple.OptionSet;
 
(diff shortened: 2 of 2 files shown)


ManfredKarrer:
Have you tested that with non en_US locales? I don't remember what was the issue but I think there have been an issue with not using en US here.

blabno:
Yes, I've tested it with `en_US`. It selects:
* MainView currency price dropbox: BTC/USD
* Settings->Base currency: BTC
* Settings->Language: English
* Settings->Country: United States
* Settings->Preferred Currency: USD

I don't see any errors in logs.

blabno:
Hm, although stuff works for this ticket, I've just realized that language is always overwritten with EN due to this:
https://github.com/bisq-network/exchange/blob/master/gui/src/main/java/io/bisq/gui/app/BisqAppMain.java#L37

ManfredKarrer:
Yes there have been issues with NON English locales. Can you try to deactivate the code in https://github.com/bisq-network/exchange/blob/master/gui/src/main/java/io/bisq/gui/app/BisqAppMain.java#L37 and test with non English locales?

blabno:
I did that with German and got German translation from the beginning.

1 lut 2018 22:01 "Manfred Karrer" <notifications@github.com> napisał(a):

ManfredKarrer:
As far I remember it was some more subtle issue... damn I should have commented it more in details....
But we can change it, the issue will pop up during dev testing or at least at release testing. The next release will require a profound test cycle anyway.

blabno:
Should we merge it then?

ManfredKarrer:
It is missing the removal of the english locale setting at startup. Can you add that?

blabno:
Removed setting default language to English at startup.

PR #1316, opened 2018-02-02
Finalize info icon for funds needed

### LINKED ISSUE #1219: Use consistent (i) icon for funds needed and re-position it
As the popup that is opened by clicking on the (?) shows the concrete details of the funds needed, it should be placed within the input box after the BTC value. Also we should use the (i) like everywhere else.

Create offer to sell - step 7

ripcurlx:
This will fix https://github.com/bisq-network/exchange/issues/1219

### DIFF
--- a/common/src/main/resources/i18n/displayStrings.properties
+++ b/common/src/main/resources/i18n/displayStrings.properties
@@ -353,7 +353,7 @@ createOffer.fundsBox.offerFee=Trade fee:
 createOffer.fundsBox.networkFee=Mining fee:
 createOffer.fundsBox.placeOfferSpinnerInfo=Offer publishing is in progress ...
 createOffer.fundsBox.paymentLabel=Bisq trade with ID {0}
-createOffer.fundsBox.fundsStructure=({0} deposit, {1} trade fee, {2} mining fee)
+createOffer.fundsBox.fundsStructure=({0} security deposit, {1} trade fee, {2} mining fee)
 createOffer.success.headline=Your offer has been published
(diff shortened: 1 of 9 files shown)


ManfredKarrer (changes requested):
When doing a rollover over the icon it does not always work. Sometimes it flickers up quickly but then gets removed. I assume it is an issue with competing rollover handlers (maybe the underlaying text field). I am not so sure if the layout over the textfield is required. We will have the info icon in more places and it will complicate things to lay it over existing components. I would suggest to use the layout where the copy icons are right and the info left and in the next row the open icon under the info icon.
I also preferred the more light green, but leave that to you...
I think we uses "security deposit" instead of "deposit" in all placed. No strong opinion about it but might be good to keep it the same wording.

ripcurlx:
I also tried already to move it outside the text field box as I recognized the flickering as well. Unfortunately the flickering also occurred there (maybe a little bit less). I'll give it a try by extending the hit area either by some javafx native functionality or by underlying it with a solid color circle that matches the background color and adding the roll over handler for this container.

ripcurlx:
I've used the old-school timeout trick to prevent the flickering, decided in the end to use the brighter green again (it is true that it pops more into the eye) and added german translations on the way.

ripcurlx:
Still there is a flickering on hiding of the component that occurs no matter if the icon is positioned on (with timeout trick) or outside of the text field (without timeout trick). Haven't found out yet why this happening at all.

ripcurlx:
@ManfredKarrer did you see my changes on this PR, in the case that you're already back from flu hell 😉?

ManfredKarrer (approved):

Issue #1315, opened 2018-02-01
IllegalMonitorStateException in Connection#doShutDown


tau3:
Hello! 

I've found a comment about `IllegalMonitorStateException` in `io.bisq.network.p2p.network.Connection#doShutDown`. 

I think that the problem is:
1. Some code invokes the method with `CloseConnectionReason` which requires to send message about closing;
2. The message is sent in a separate thread (see `Connection#shutDown(CloseConnectionReason, Runnable`)) in current implementation;
3. `shutDown()` method invokes `doShutDown()` through `UserThread`. But the current executor of `UserThread` (`MoreExecutors#directExecutor`) just runs all jobs in the same thread that has called `UserThread`;
4. So `doShutdown()` method is actualy invoked in a thread that was created in step 2 (a _different_ thread);
5. The thread is definitiely not the owner of `Connection#protoOutputStreamLock` which is attempted to be unlocked. That is why the exception occurs IMO.

tl;dr ReentrantLock is locked in one thread and unlocked in another.

ManfredKarrer:
Good analysis! Thanks!

The shutdown is called from user threads as well as from the connection itself. We might need to refactor that to handle those 2 different thread context differently. Feel free to make a suggestion!

tau3:
TBH, i'm not actually sure that i completely understand locks flow in `Connection` class, so my suggestion is quite straightforward - can't we just execute shutdown in the same thread in case of sending close message? Would it cause any problems?

ManfredKarrer:
Yes that should work. Just the shutdown calls from outside need to be considered. Feel free to refactor the locking handling there. The Connection class could need some love ;-).

ManfredKarrer:
Merged in https://github.com/bisq-network/bisq/pull/1318

Btw: @tau3 Feel free to make a compensation request for all your past contributions!

PR #1318, opened 2018-02-02
bugfix/1315/connection-lock-problem

tau3:
Hello!

Here is my implementation for #1315. I've moved all locking logic from `Connection` to `SynchronizedProtoOutputStream` class.

What is attempted to be achieved:
1. Only one thread can write into socket's `OutputStream` each time;
2. Only currently writing thread would finish message sending on connection's shutdown; other locked threads would be immediately unlocked, but wouldn't attempt to write anything.

I've used a single thread executor, so only one thread performs actions upon socket's `OutputStream`, all other writing threads would wait on `Future#get()` (to simulate synchronous execution)

I understand that i've modified a very sensitive code, so please, double check after me 😃

### DIFF
--- a/network/src/main/java/io/bisq/network/p2p/network/SynchronizedProtoOutputStream.java
+++ b/network/src/main/java/io/bisq/network/p2p/network/SynchronizedProtoOutputStream.java
@@ -1,14 +1,37 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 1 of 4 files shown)


network/src/main/java/io/bisq/network/p2p/network/SynchronizedProtoOutputStream.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    void writeEnvelope(NetworkEnvelope envelope) {
+        Future<?> future = executorService.submit(() -> super.writeEnvelope(envelope));
+        try {
+            future.get();
+        } catch (InterruptedException e) {
+            Thread currentThread = Thread.currentThread();
+            currentThread.interrupt();
+            log.error("Thread " + currentThread + " was interrupted", e);
+            throw new BisqRuntimeException("Failed to write envelope", e);
```
tau3, line 51: @ManfredKarrer not sure if an exception should be thrown in this case. An Interrupted exception may occur on shutdown of executor service - maybe just returning from `writeEnvelope` is better.

ManfredKarrer (approved):
Looks good to me. But we need to test quite a bit with long running live nodes to see if there are unexpected issues. Thanks again for your great improvements!

PR #1323, opened 2018-02-02
Update displayStrings_ro.properties

stevenkain:
New translation


ManfredKarrer (approved):
I cannot review much here as I don't speak romanian ;-)

PR #1326, opened 2018-02-05
Make arbitration chat messages selectable

### LINKED ISSUE #1058: Make arbitration chat messages selectable
Currently, the text in Bisq's arbitration chat messages is not selectable, i.e. it is not possible to click and drag over the text in order to copy a portion to the clipboard.

image

Notice in the screenshot above the double-document icon in the upper right of each message. This copies the entire contents of the chat message to the clipboard, but this is cumbersome when all one wants is to copy a particular passage of text, e.g. a URL.

With regard to the related topic of URLs actually being clickable in these messages, see #1059.

blabno:
Fixes #1058

### DIFF
--- a/gui/src/main/java/io/bisq/gui/util/GUIUtil.java
+++ b/gui/src/main/java/io/bisq/gui/util/GUIUtil.java
@@ -46,16 +46,15 @@ import io.bisq.gui.components.indicator.TxConfidenceIndicator;
 import io.bisq.gui.main.overlays.popups.Popup;
 import io.bisq.network.p2p.P2PService;
 import javafx.beans.property.DoubleProperty;
-import javafx.collections.ObservableList;
 import javafx.geometry.Orientation;
 import javafx.scene.Node;
+import javafx.scene.Scene;
 import javafx.scene.control.ScrollBar;
+import javafx.scene.control.TextArea;
 import javafx.scene.control.Tooltip;
 import javafx.scene.layout.Pane;
 import javafx.scene.layout.Region;
-import javafx.stage.DirectoryChooser;
-import javafx.stage.FileChooser;
-import javafx.stage.Stage;
+import javafx.stage.*;
 import javafx.util.StringConverter;
 import lombok.extern.slf4j.Slf4j;
 import org.bitcoinj.core.Address;
@@ -64,7 +63,6 @@ import org.bitcoinj.core.TransactionConfidence;
 import org.bitcoinj.uri.BitcoinURI;
 import org.bitcoinj.wallet.DeterministicSeed;
 
-import javax.annotation.Nullable;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


gui/src/main/java/io/bisq/gui/util/GUIUtil.java
```diff
@@ -446,4 +444,22 @@ public static void restoreSeedWords(DeterministicSeed seed, WalletsManager walle
...
+
+    public static void showSelectableTextModal(String title, String text) {
+        TextArea textArea = new TextArea();
+        textArea.setText(text);
+        textArea.setEditable(false);
+        textArea.setWrapText(true);
+        textArea.setPrefSize(800, 600);
+
+        Scene scene = new Scene(textArea);
+        Stage stage = new Stage();
```
blabno, line 456: Most of this code comes from ContractWindow.java.
That class also sets owner (I'm not sure what that is for).
https://github.com/bisq-network/exchange/blob/master/gui/src/main/java/io/bisq/gui/main/overlays/windows/ContractWindow.java#L205

blabno (commented):
Just a comment

cbeams (approved):
ACK.

I've tested these changes and they work well. @keo-bisq, heads-up: you can now double-click on arbitration chat messages and a window like the one below will pop up, where you can select and copy any portion of the text.

@bisq-network/docs-maintainers, this is also good for users to know, and is something we should add to our (currently non-existent) documentation.

image

Thanks, @blabno!

cbeams:
> @bisq-network/docs-maintainers, this is also good for users to know, and is something we should add to our (currently non-existent) documentation.

Actually, we shouldn't rely on documentation to reveal these sort of "hidden features". We should add an affordance to the UI that makes it obvious to the user that this feature is there.

We could (1) add a second icon next to the "copy to clipboard" icon that's already there, or (2) just change the behavior of the existing button to open this dialog instead. I'm open to either. Thoughts, @ripcurlx, @ManfredKarrer, @keo-bisq?

@blabno, feel free to open a follow-on issue about this.

ManfredKarrer:
I would prefer to make the text in the chat selectable. That would not require any user instruction. I will not have time to look into it but it was not trivial as far I remember from the text component which is used. But with a bit of effort I am sure it can be made.

ManfredKarrer:
I had another look....
Seems there is no ready made component out there which does what we want.
Label does not support text selection. Textfield is not multiline and Textarea does not adjust to size as we need it.

The hacks i found are using an invisible Label for layout and put a Textarea with the correct size over it (taken from the label) and use styles to look like a label (transparent background).
here are some examples: 
https://community.oracle.com/thread/2319231
https://stackoverflow.com/questions/22534067/copiable-label-textfield-labeledtext-in-javafx
https://stackoverflow.com/questions/46869986/javafx-selectable-label-textarea-without-padding-or-border?rq=1

Would be good to make a component which integrates that hack or even better merge the functionalities of Label and Textarea, thought that might be a bit more tricky I assume.

ManfredKarrer:
@ripcurlx Maybe you have an idea?

ripcurlx:
Maybe we could do something like this https://stackoverflow.com/questions/44173811/how-to-make-a-javafx-label-selectable on single mouse click. Besides that I also didn't find a perfect solution that is ready to use.

ManfredKarrer:
Did not like that so much to add a new behavior. It should just behave like expected to select a text.

cbeams:
@ManfredKarrer, do you want to revert this change after all?

ManfredKarrer:
I dont mind. It might be an improvement to current copy icon and as its for arbitrators only i dont see a big usability issue. A proper selectable label component seems to be not trivial and for now maybe not worth the effort.

PR #1329, opened 2018-02-06
Fix wrong css alignment.

ripcurlx:
Bug wasn't recognized as the visual outcome is the same. But the current statement causes errors in the debug log.


ManfredKarrer (approved):
uACK

PR #1330, opened 2018-02-06
Configurable hostname in localhost p2p mode.

blabno:
This is necessary to put each peer in separate docker container (for integration tests).


ManfredKarrer (approved):
uACK

PR #1337, opened 2018-02-08
Add $BOGUS altcoin with failing test

cbeams:
To test the changes made to Travis CI config in ed0148b.

### DIFF
--- a/gui/src/test/java/io/bisq/gui/util/validation/AltCoinAddressValidatorTest.java
+++ b/gui/src/test/java/io/bisq/gui/util/validation/AltCoinAddressValidatorTest.java
@@ -585,6 +585,6 @@ public class AltCoinAddressValidatorTest {
 
     @Test
     public void testBOGUS() {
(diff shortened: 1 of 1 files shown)


cbeams:
Closing as complete. Not merging because this was just a test to verify that pull request branches are in fact running tests, and fail / succeed as expected, especially with regard to status information here in GitHub.

@ManfredKarrer, note the status indicators above, e.g.:

image

If you follow that red X, you'll see the Travis CI build that failed, and specifically, you'll see that **the build failed because a unit test failed.**

I then fixed that failing test in the next commit, and as you can see, the status indicator turned green for that commit:

image

And when the last build was a failing build, you get this BIG red X in the main status overview:

image

Altogether, this means that you can now trust this information to know if a pull request is utterly broken (as they all too often are). The PR still needs to be reviewed, of course, for overall correctness and sanity, but at least we have an early warning system now against the worst offenders.

/cc @mrosseel

PR #1355, opened 2018-02-13
Comp req refactor

sqrrm:
Some refactoring of compensation request code, moving domain specifics out of gui. Should make it easier to write tests, but should finish the compensation req somewhat first. Relates to #1268

Not sure about the handling of CompensationRequest vs CompensationRequestPayload. I feel the CompensationRequest should handle the related functionality and the payload just be the data send on the wire basically.

### DIFF
--- a/core/src/main/java/io/bisq/core/dao/compensation/CompensationRequest.java
+++ b/core/src/main/java/io/bisq/core/dao/compensation/CompensationRequest.java
@@ -18,16 +18,22 @@
 package io.bisq.core.dao.compensation;
 
 import io.bisq.common.proto.persistable.PersistablePayload;
+import io.bisq.core.btc.wallet.BsqWalletService;
 import io.bisq.generated.protobuffer.PB;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
+import org.bitcoinj.core.Address;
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/bisq/core/dao/compensation/CompensationRequest.java
```diff
@@ -108,4 +122,17 @@ public static CompensationRequest fromProto(PB.CompensationRequest proto) {
...
+    /// API
+    public Coin getRequestedBsq() {
+        checkNotNull(payload);
+        return payload.getRequestedBsq();
+    }
+
+    public Address getIssuanceAddress (BsqWalletService bsqWalletService) {
+        checkNotNull(payload);
+        // Remove leading 'B'
+        String underlying_btc_address = payload.getBsqAddress().substring(1, payload.getBsqAddress().length());
```
ManfredKarrer, line 135: We should stick with camelCase style: underlying_btc_address -> underlyingBtcAddress

ManfredKarrer (approved):
uACK

core/src/main/java/io/bisq/core/dao/compensation/CompensationRequest.java
sqrrm: Thanks for fixing, it's sometimes hard to kill those habits.

PR #1360, opened 2018-02-14
Shorten labels in call-to-action buttons in offer book

### LINKED ISSUE #1199: Shorten text in call to action buttons
Here again, it is as explicit as it can be what will happen when you click on one of the call-to-action buttons in the offer book, but a slight simplification and a currency selection dialogue upfront for first time users could make this screen less overwhelming. This will also make it easier for new users to connect the buttons on the bottom with the icon buttons on the top. Also it would help people to be 100% sure that BTC means Bitcoin 😉 .
offer book

ripcurlx:
This will fix https://github.com/bisq-network/exchange/issues/1199.


ManfredKarrer (approved):
uACK

PR #1362, opened 2018-02-15
Refactoring of CoreSeedNodesRepository + unit tests

tau3:


core/src/main/java/io/bisq/core/network/CoreSeedNodesRepository.java
```diff
@@ -1,58 +1,37 @@
...
-import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-@Slf4j
+import static io.bisq.core.network.DefaultNodeAddresses.DEFAULT_LOCALHOST_SEED_NODE_ADDRESSES;
+import static io.bisq.core.network.DefaultNodeAddresses.DEFAULT_TOR_SEED_NODE_ADDRESSES;
+
+@NotThreadSafe
```
tau3: @ManfredKarrer is it ok that i marked the class as not thread safe? Actually it had been safe to access `seedNodeAddresses` from multiple threads before; but it hadn't been so for any other field of the class

ManfredKarrer: Yes you can mark it. 

In general most classes are living only in the context of a single user thread. 
GUI app:
```UserThread.setExecutor(Platform::runLater);```
or if not UI app (seed nodes):
```UserThread.setExecutor(Executors.newSingleThreadExecutor());```

There are only a few areas where multi threading is used: 
- Connection (P2P network)
- Tor startup
- Wallet encryption
- Local persistence
- Price request
- Fee request
- DAO/BSQ

From BitcoinJ we map to user thread as well.

Usual strategy is to map the result from the threaded operation to the user thread (via ``` UserThread.execute(() -> ....);``` so the clients can be sure that the callbacks are living in the same thread context. That model removes a lot of the threading complexity and isolates threading only to those areas where it is required.

**Comment by cbeams** [comment]
@tau3, I'd like to push some changes as part of my review. Could you please check the following box on the right side of this PR? Thanks.

image

File: core/src/main/java/io/bisq/core/network/CoreSeedNodesRepository.java [hunk 2]
```diff
@@ -1,58 +1,37 @@
...
 
-        if (bannedNodes == null)
-            log.info("seedNodeAddresses={}", seedNodeAddresses);
-        else
-            log.warn("We received banned seed nodes={}, seedNodeAddresses={}", bannedNodes, seedNodeAddresses);
+    @Override
+    public Set<NodeAddress> getSeedNodeAddresses() {
+        if (seedNodeAddresses == null) {
+            seedNodeAddresses = lookup.resolveNodeAddresses();
+        }
```
cbeams: Why the lazy initialization here in the setter? Is this an attempt to avoid an expensive operation? Is this actually an optional code path? If it is usually invoked, or if the operation is not terribly expensive, prefer assigning `seedNodeAddresses` directly in the constructor, such that the field can be marked `final`.

tau3: `NodeAddressLookup` contains some non-trivial logic, so i thought that invoking it in constructor would not be a good idea. 

> prefer assigning seedNodeAddresses directly in the constructor, such that the field can be marked `final`

totally agree, that way is better. But it would lead us to creation of factory of `SeedNodesRepository`. Is it ok?

cbeams (changes requested):
NACK.

@tau3, I have a number of commits I'd like to push to this PR branch as part of my review. Please check the box that allows maintainers to push commits, thanks.

Once I'm able to push those changes, I'll add a few more comments.

In the meantime, there is a bug somewhere in commits 3e13a0f2262a2f1ca4182a299099edcee4c4f83e and 9c2f0d5c00fdeac377c3a1ac41cca95f1978fd82 that causes the following error on attempting to start Bisq:

```
java.lang.IllegalArgumentException: fullAddress must contain ':'
```

Just try to run `BisqAppMain` from within IDEA, or run `mvn package` followed by `java -jar gui/target/shaded.jar` and you'll see the error get raised immediately as Guice tries and fails to wire up the application.

tau3:
@cbeams yep, there was a silly mistake 😕 
fixed and added unit test.

cbeams:
> @cbeams yep, there was a silly mistake 😕
> fixed and added unit test.

Thanks. I've confirmed locally. **Can you please check the "Allow edits from maintainers" option** as requested above in https://github.com/bisq-network/exchange/pull/1362#issuecomment-366744626

image

I have 11 additional commits as part of my review, and I cannot push them until you've checked this box.

```
~/Work/bisq-network/exchange[tau3-refactoring/core-seed-nodes-repository]
$ git push tau3 tau3-refactoring/core-seed-nodes-repository
fatal: remote error:
  You can't push to git://github.com/tau3/exchange.git
  Use https://github.com/tau3/exchange.git
```

tau3:
@cbeams it is already checked ¯\\_(ツ)_/¯

cbeams (approved):

cbeams (approved):
ACK.

@tau3, please review each of my additional commits, including the extended commit comments where they exist.

There is more I would have done here, or requested that you do, but I've held off on it, because I have a larger refactoring around seed nodes in mind anyway. Would you be up for having a conversation about that?

In general, your changes are great. Thank you for putting this PR together!

In the future, please tweak the grammar of your commit comments so they adhere to the guidelines at https://chris.beams.io/posts/git-commit#seven-rules. You're already doing most of these; it's just the "imperative mood" one I'm focusing on here.

There are some other nits, like the fact that I don't think we need immutable sets here, and therefore don't need the `ImmutableSetDecorator` class you've introduced, but again, I'm happy to leave this as-is in service of moving on to larger, higher-impact changes. Don't get me wrong here, though: immutable sets are great, and ideally, all collections would be immutable and persistent by default in Java, but they're not, and therefore introducing them always carries certain costs, like the extra class we see here. Generally, if there's not a requirement for collection immutability, just go with the standard collections; it's the convention throughout most of the rest of Bisq anyway.

cbeams:
> @cbeams it is already checked ¯\(ツ)/¯

Regarding this: it was an issue with the way I was attempting to push to your remote. My bad; thanks for taking care of this when you originally did. That checkbox will, I believe, now stay checked for future PRs in this repository (thanks to a cookie), but please double-check the next couple PRs to be sure. Thanks.

cbeams:
My apologies, @tau3, I never actually merged this PR after our last interactions above. Some merge conflicts had cropped up in the meantime, and I've fixed them. I'll be merging shortly, as soon as Travis gives the OK.

PR #1365, opened 2018-02-16
Only show range when necessary

### LINKED ISSUE #1215: Don't show amount range if there is no amount range
In some screens (e.g. take offer, my open offers,...) we display the amount range, even if there is no amount range. In the cases like my open offers it should be displayed e.g 0.10 instead of 0.10-0.10 or in cases like the take offer screen hide this field at all.

*Take Offer*
Taker Offer
*My Open Offers*
My Open Offers

ripcurlx:
This will fix https://github.com/bisq-network/exchange/issues/1215.
It only shows an range (amount, volume), if there is one and only will allow the user select an amount (take offer) if there is one to select from. In the case of take offer, it will directly jump to the fund offer step if no amount needs to be selected


ManfredKarrer (approved):
uACK

Issue #1265, opened 2018-01-24
Implement compensation requests


cbeams:

ripcurlx:
@ManfredKarrer I think this can be closed as complete, correct?

PR #1367, opened 2018-02-17
Compensation req checks

sqrrm:
Some further improvements on the compensation requests. It seems not trivial to write tests so will wait with that until it's a bit more solidified.

On style, I added CompensationAmountException to make sure it's not possible to create comp reqs of less bsq than min comp req fee. I'm not a big fan of using exceptions as a regular workflow, this case being on the edge of what I think is reasonable. Any thoughts?

### DIFF
--- a/common/src/main/resources/i18n/displayStrings.properties
+++ b/common/src/main/resources/i18n/displayStrings.properties
@@ -1032,7 +1032,7 @@ dao.compensation.display.link=Link to detail info:
 dao.compensation.display.link.prompt=Provide a link to an issue at \"https://github.com/bisq-network/compensation/issues/\" or the Bisq Forum.
 dao.compensation.display.requestedBsq=Requested amount in BSQ:
 dao.compensation.display.bsqAddress=BSQ address:
-dao.compensation.display.txId=BTC TxID:
+dao.compensation.display.txId=BTC Transaction ID:
 
 dao.voting.item.title=Compensation request
 dao.voting.item.stage.title=Compensation request with ID: {0}
--- a/common/src/main/resources/i18n/displayStrings.properties
+++ b/common/src/main/resources/i18n/displayStrings.properties
@@ -1102,7 +1102,7 @@ dao.compensation.create.confirm.info=Requested amount: {0}\n\
   Are you sure you want to publish the compensation request?
 dao.compensation.create.missingFunds=You don''t have sufficient funds for creating the compensation request.\n\
   Missing: {0}
-
+dao.compensation.create.amountTooLow=You need to request to issue more than {0}
 
 ####################################################################
 # Windows
--- a/common/src/main/resources/i18n/displayStrings.properties
+++ b/common/src/main/resources/i18n/displayStrings.properties
@@ -1032,6 +1032,7 @@ dao.compensation.display.link=Link to detail info:
 dao.compensation.display.link.prompt=Provide a link to an issue at \"https://github.com/bisq-network/compensation/issues/\" or the Bisq Forum.
 dao.compensation.display.requestedBsq=Requested amount in BSQ:
 dao.compensation.display.bsqAddress=BSQ address:
+dao.compensation.display.txId=BTC TxID:
 
 dao.voting.item.title=Compensation request
 dao.voting.item.stage.title=Compensation request with ID: {0}
--- a/core/src/main/java/io/bisq/core/dao/blockchain/parse/OpReturnVerification.java
+++ b/core/src/main/java/io/bisq/core/dao/blockchain/parse/OpReturnVerification.java
@@ -56,6 +56,7 @@ public class OpReturnVerification {
                         return compensationRequestVerification.processOpReturnData(tx, opReturnData, txOutput,
                                 bsqFee, blockHeight, btcOutput);
                     case DaoConstants.OP_RETURN_TYPE_VOTE:
+                        // TODO: Handle missing bsqOutput, is it considered an invalid vote?
                         if (bsqOutput == null) {
                             log.warn("Voting tx is missing bsqOutput for vote base txid={}", tx.getId());
                         }
--- a/core/src/main/java/io/bisq/core/dao/blockchain/parse/OpReturnVerification.java
+++ b/core/src/main/java/io/bisq/core/dao/blockchain/parse/OpReturnVerification.java
@@ -46,7 +46,7 @@ public class OpReturnVerification {
         TxOutput txOutput = txOutputs.get(index);
         final long txOutputValue = txOutput.getValue();
         // If we get an OP_RETURN it has to be the last output and the txOutputValue has to be 0 as well there have be at least one bsqOutput
-        if (txOutputValue == 0 && index == txOutputs.size() - 1 && bsqOutput != null && bsqFee > 0) {
+        if (txOutputValue == 0 && index == txOutputs.size() - 1 && bsqFee > 0) {
             byte[] opReturnData = txOutput.getOpReturnData();
             // We expect at least the type byte
             if (opReturnData != null && opReturnData.length > 1) {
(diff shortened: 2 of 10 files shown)


sqrrm:
Relates to #1265

common/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1032,6 +1032,7 @@ dao.compensation.display.link=Link to detail info:
 dao.compensation.display.link.prompt=Provide a link to an issue at \"https://github.com/bisq-network/compensation/issues/\" or the Bisq Forum.
 dao.compensation.display.requestedBsq=Requested amount in BSQ:
 dao.compensation.display.bsqAddress=BSQ address:
+dao.compensation.display.txId=BTC TxID:
```
ManfredKarrer: I would prefer: "BTC transaction ID:" if space in UI allows, otherwise tx ID

core/src/main/java/io/bisq/core/dao/blockchain/parse/OpReturnVerification.java
```diff
@@ -56,6 +56,9 @@ boolean processDaoOpReturnData(Tx tx, int index, long bsqFee,
                         return compensationRequestVerification.processOpReturnData(tx, opReturnData, txOutput,
                                 bsqFee, blockHeight, btcOutput);
                     case DaoConstants.OP_RETURN_TYPE_VOTE:
+                        if (bsqOutput == null) {
```
ManfredKarrer: Maybe we add a TODO here as I think it was not allowed to have no bsqOutput here. But voting is WIP anyway...

core/src/main/java/io/bisq/core/dao/compensation/CompensationAmountException.java
```diff
@@ -0,0 +1,12 @@
+package io.bisq.core.dao.compensation;
+
+import org.bitcoinj.core.Coin;
+
+public class CompensationAmountException extends Throwable {
```
ManfredKarrer: I think its better to extend from Exception here.

core/src/main/java/io/bisq/core/dao/compensation/CompensationRequestManager.java
```diff
@@ -141,15 +141,14 @@ public void addToP2PNetwork(CompensationRequestPayload compensationRequestPayloa
     }
 
     public CompensationRequest prepareCompensationRequest(CompensationRequestPayload compensationRequestPayload)
-            throws InsufficientMoneyException, ChangeBelowDustException, TransactionVerificationException, WalletException, IOException {
+            throws InsufficientMoneyException, ChangeBelowDustException, TransactionVerificationException, WalletException, IOException, CompensationAmountException {
+        if (compensationRequestPayload.getRequestedBsq().compareTo(feeService.getCreateCompensationRequestFee()) <= 0) {
+            throw new CompensationAmountException(feeService.getCreateCompensationRequestFee(), compensationRequestPayload.getRequestedBsq());
```
ManfredKarrer, line 146: I think that is not correct. The requested BSQ are not the BSQ used for the fee payment and it would be a < not a <= comparison.

gui/src/main/java/io/bisq/gui/main/dao/compensation/CompensationRequestDisplay.java
```diff
@@ -62,12 +69,19 @@ public void createAllFields(String title, double top) {
         linkHyperlinkWithIcon.setManaged(false);
         linkInputTextField.setPromptText(Res.get("dao.compensation.display.link.prompt"));
         requestedBsqTextField = addLabelInputTextField(gridPane, ++gridRow, Res.get("dao.compensation.display.requestedBsq")).second;
-
+        if (feeService != null) {
+            BsqValidator bsqValidator = new BsqValidator(bsqFormatter);
+            bsqValidator.setMinCompensationRequest(feeService.getCreateCompensationRequestFee());
+            requestedBsqTextField.setValidator(bsqValidator);
+        }
```
ManfredKarrer, line 76: I think here it is again wrong: The requested BSQ are not the BSQ fee.

gui/src/main/java/io/bisq/gui/util/validation/BsqValidator.java
```diff
@@ -118,4 +125,17 @@ protected ValidationResult validateIfSufficientAvailableBalance(String input) {
             return new ValidationResult(false, Res.get("validation.invalidInput", t.getMessage()));
         }
     }
+
+    protected ValidationResult validateIfMoreThanMinCompensationRequest(String input) {
+        try {
+            final Coin coin = bsqFormatter.parseToCoin(input);
+            if (minCompensationRequest != null && coin.compareTo(minCompensationRequest) <= 0)
+                return new ValidationResult(false, Res.get("dao.compensation.create.amountTooLow",
```
ManfredKarrer, line 133: We would allow coin == minCompensationRequest, so the <= 0 should be < 0.

Btw in the validateIfSufficientAvailableBalance method above there is wrong comaparison as well:
if (availableBalance != null && coin.compareTo(availableBalance) > 0)
should be the opposite:
if (availableBalance != null && coin.compareTo(availableBalance) < 0)

ManfredKarrer (changes requested):

core/src/main/java/io/bisq/core/dao/compensation/CompensationRequestManager.java
sqrrm: As discussed on slack: if the minimum compensation amount is less than or equal to the bsq fee then issuance output will be treated as a bsq output. Example with bsq fee = 1 BSQ and issuance request for 1 BSQ
Input 0: 100 BSQ (0.001 BTC)
Input 1: 0.00001 BTC 
Output 0: 99 BSQ (0.00099 BTC)
Output 1: issuance of 1 BSQ (0.00001 BTC)

Output 1 is supposed to be the issuance output using the BTC from input 1 but will actually be treated as a BSQ output using the remaining 1 BSQ from input 0. No BSQ will be burnt and it would create a malformed compensation request. By requiring the issuance to be more than the BSQ fee (which makes sense anyway, who would want to pay more than the earn to issue new BSQ) this is no longer an issue.

gui/src/main/java/io/bisq/gui/main/dao/compensation/CompensationRequestDisplay.java
sqrrm: Same check as the one done in CompensationRequestManager. It's better to make sure to have the check here in the UI, but for correctness I think it should also be handled in CompensationRequestManager.

gui/src/main/java/io/bisq/gui/util/validation/BsqValidator.java
sqrrm: This is the same issue with BSQ fee as above.

For the validateIfSufficientAvailableBalance that seems correct. If coin > availableBalance it fails.

common/src/main/resources/i18n/displayStrings.properties
sqrrm: Fixed.

core/src/main/java/io/bisq/core/dao/blockchain/parse/OpReturnVerification.java
sqrrm: TODO added.

core/src/main/java/io/bisq/core/dao/compensation/CompensationAmountException.java
sqrrm: Fixed.

gui/src/main/java/io/bisq/gui/util/validation/BsqValidator.java
ManfredKarrer: I will refactor the validator to make it more generic with a minAmount. Yes the validateIfSufficientAvailableBalance was right but I will change the order of comparison so its better readable.

core/src/main/java/io/bisq/core/dao/compensation/CompensationRequestManager.java
ManfredKarrer: I will introduce a minCompRequestAmount to cover that.

PR #1372, opened 2018-02-19
Reflect pricenode operator change in ProvidersRepository

cbeams:
See bisq-network/roles#14 and bisq-network/pricenode#7


ManfredKarrer:
Ah I just committed the same a minute before ;-)

cbeams:
> Ah I just committed the same a minute before ;-)

Saw that, was wondering about it. Do you want to just back that out of the 0.6.6 release branch, then? Otherwise, we get into cherry-picking commits, doesn't seem worth it for this "surgical" release.

PR #1374, opened 2018-02-20
Rename Bicoin Clashic ticker symbol from BCHC to BCL

cbeams:
From https://twitter.com/bisq_network/status/963508497328549889:

image

@ManfredKarrer, please note that this PR does *not* update `network/src/main/resources/PersistableNetworkPayloadMap_BTC_MAINNET`, which I believe you usually do when making these kinds of changes. I'm not sure what process you follow there, please update it as you see fit, and perhaps provide details about how and when that resource should be updated in the future. I only considered changing it because it matched a `git grep BCHC` search of the codebase.

Another caveat is that I don't know what effect this change will have on calculating the market price of BCL. We don't get Bitcoin Clashic exchange rate data from any of our providers, so we're displaying market price based on the last Bisq trade. I imagine that we'll revert to a state where there is no last price information, since there will (after this change) be no last BCL trade to base it on. Again, any details you can provide here would be appreciated.

### DIFF
--- a/common/src/main/java/io/bisq/common/locale/CurrencyUtil.java
+++ b/common/src/main/java/io/bisq/common/locale/CurrencyUtil.java
@@ -94,7 +94,7 @@ public class CurrencyUtil {
         if (!baseCurrencyCode.equals("BTC"))
             result.add(new CryptoCurrency("BTC", "Bitcoin"));
         result.add(new CryptoCurrency("BCH", "Bitcoin Cash"));
-        result.add(new CryptoCurrency("BCHC", "Bitcoin Clashic"));
+        result.add(new CryptoCurrency("BCL", "Bitcoin Clashic"));
         result.add(new CryptoCurrency("BTG", "Bitcoin Gold"));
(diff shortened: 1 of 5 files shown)


ManfredKarrer:
Renaming an existing currency has issues with historical data and open offers. I personally don't have time to spend on that.

cbeams:
I can sympathize, and of course you're right. That creates ugly incompatibilities for historical data. The only thing I can think of is to add a new BCL and "deprecate" BCHC, i.e. make it such that the symbol is still there for historical data, but does not show up as an option in the UI for viewing / making / taking new offers. Not sure if we have such a deprecation mechanism right now. Not sure if we've ever removed an altcoin that's had more than zero trades.

In any case, feel free to just leave this as-is for now. We can leave it out of 0.6.6, which is more pressing and time-driven. Thanks.

ManfredKarrer:
Yes there have been once the case but as far I remember there have been not trades. The ticker symbol is a unique ID and altcoiners need to learn to take it as that. Anyone who wants to put work on that to support it is welcome, it is just that I have tons of other hi-prio work.

PR #1373, opened 2018-02-19
Proposed fix for Issue 1186: Resize "Offers to buy/sell with X" list to Bisq's vertical window size.

ghubstan:

PR #1375, opened 2018-02-20
Resize "Offers to buy/sell with X"; list to Bisq's vertical window size

ghubstan:
Replaces closed PR 1373 (had dirty commit history).

### DIFF
--- a/gui/src/main/java/io/bisq/gui/main/market/offerbook/OfferBookChartView.java
+++ b/gui/src/main/java/io/bisq/gui/main/market/offerbook/OfferBookChartView.java
@@ -63,6 +63,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.inject.Inject;
 import java.util.Collections;
+import java.util.function.Function;
 
 @FxmlView
 public class OfferBookChartView extends ActivatableViewAndModel<VBox, OfferBookChartViewModel> {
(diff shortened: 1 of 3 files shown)


cbeams (approved):
utACK.

@ManfredKarrer please review as well, I've only looked at this for PR correctness, not its substance.

@ghubstan, in future commits, please:

Follow the commit comment guidelines at https://chris.beams.io/posts/git-commit.

And In GitHub PRs and issues, please link to issues with hashmarks, e.g. #1373 (your description in this PR mentioned the issue number without the `#`). Doing this creates linkage and discoverability between the issues in the GitHub UI.

Thanks!

gui/src/main/java/io/bisq/gui/main/market/offerbook/OfferBookChartView.java
```diff
@@ -88,7 +89,13 @@
     private HBox bottomHBox;
     private ListChangeListener<OfferBookListItem> changeListener;
     private ListChangeListener<CurrencyListItem> currencyListItemsListener;
-
+    private ChangeListener<Number> bisqWindowVerticalSizeListener;
+    private final double pixelsPerOfferTableRow = (109.0 / 4.0) + 10.0;
+    private final Function<Double, Double> offerTableViewHeight = (screenSize) -> {
+        // startup defaults: scene height = 710   tbl view height = 109   visible row count = 4
+        int extraRows = screenSize <= 710 ? 0 : (int) ((screenSize - 710) / pixelsPerOfferTableRow);
```
ripcurlx: Maybe you could move the initial layout height (710) as a field to the [Layout class](https://github.com/bisq-network/exchange/blob/master/gui/src/main/java/io/bisq/gui/util/Layout.java) and also replace it in the [BisqApp class](https://github.com/bisq-network/exchange/blob/1d0591084d6d2cc8204357204ca0f54b887ab765/gui/src/main/java/io/bisq/gui/app/BisqApp.java#L235) so the code will work as expected if we decide to change the default height at some point. Also please move the table view height (109) in a field and re-use it in the table view initialization https://github.com/bisq-network/exchange/blob/1d0591084d6d2cc8204357204ca0f54b887ab765/gui/src/main/java/io/bisq/gui/main/market/offerbook/OfferBookChartView.java#L302

ripcurlx (changes requested):
Besides moving the numbers into fields everything else works as expected 👍

gui/src/main/java/io/bisq/gui/main/market/offerbook/OfferBookChartView.java
ghubstan: I pushed the changes you asked for in two commits that should have been one.

ghubstan (commented):
Review request for changes made as per

"Maybe you could move the initial layout height (710) as a field to the Layout class and also replace it in the BisqApp class so the code will work as expected if we decide to change the default height at some point. Also please move the table view height (109) in a field and re-use it in the table view initialization "

ripcurlx (approved):

ripcurlx:
ACK

PR #1378, opened 2018-02-20
Update displayStrings.properties

stevenkain:
"both trades" makes no sense. Shouldn't it be "both traders"?


ripcurlx:
I think so too. @ManfredKarrer was there a specific reason why you chose **trades** instead of **traders**?

ManfredKarrer:
No was type...

PR #1379, opened 2018-02-20
Update displayStrings_hu.properties

stevenkain:
New translations


ripcurlx:
utACK - same goes for these changes. Thanks @stevenkain for cleaning and filling up missing translations!

PR #1380, opened 2018-02-20
Update displayStrings_ro.properties

stevenkain:
New Romanian translations


ripcurlx:
utACK - as I don't speak Romanian 😉

PR #1383, opened 2018-02-21
Make USE_DEV_PRIVILEGE_KEYS flag configurable

### LINKED ISSUE #1359: Regression: No longer possible to register arbitrator in REGTEST.
At:
https://github.com/bisq-network/exchange/blob/ff629542e7e83c70202078c20edec6a37f5f5878/common/src/main/java/io/bisq/common/app/DevEnv.java#L13

`USE_DEV_PRIVILEGE_KEYS` is  `final` and has been changed to `false`, thus preventing arbitrator registration in REGTEST.

If we change it to:
```java
public static boolean USE_DEV_PRIVILEGE_KEYS = false;
```
and then, immediately inside the main method at:
https://github.com/bisq-network/exchange/blob/ff629542e7e83c70202078c20edec6a37f5f5878/gui/src/main/java/io/bisq/gui/app/BisqAppMain.java#L37
insert:
```java
for (String arg: args){
    if(arg.contains("REGTEST")) DevEnv.USE_DEV_PRIVILEGE_KEYS = true;
}
```
the behaviour returns to expected and possible production issues which may have caused the change are avoided.

blabno:
Fixes #1359

### DIFF
--- a/common/src/main/java/io/bisq/common/app/DevEnv.java
+++ b/common/src/main/java/io/bisq/common/app/DevEnv.java
@@ -4,13 +4,11 @@ public class DevEnv {
     // Was used for P2P network stress test to adjust several setting for the tests (e.g. use lower btc fees for offers,..)
     public static final boolean STRESS_TEST_MODE = false;
 
-    // If that is true all the privileged features which requires a private key to enable it are overridden by a dev ey pair.
     // The UI got set the private dev key so the developer does not need to do anything and can test those features.
     // Features: Arbitration registration (alt+R at account), Alert/Update (alt+m), private message to a
     // peer (click user icon and alt+r), filter/block offers by various data like offer ID (cmd + f).
     // The user can set a program argument to ignore all of those privileged network_messages. They are intended for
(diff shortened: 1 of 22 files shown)


ManfredKarrer (changes requested):
Why not use a boolean and leave the key pair as it is? The way you did it requires that you create a EC keypair and pass both keys. Is there a use case where you need keys different to the hard coded one in the app? I assume what you want is to flip the flag by a prog arg, right?

blabno:
I don't understand, I just extracted that Boolean and did not touch keys.

And yes, I want to steer if hardcoded keys should be used or not, using prog arg.

ManfredKarrer:
Ah sorry, ignore my previous comment. I just skimmed over it and got it wrong. I will merge as soon I am better (still recovering a bronchitis) so that might still take a bit, as I want to test it also.

ManfredKarrer:
@blabno I moved the useDevPrivilegeKeys to BisqAppModule as it is used at several domains not only in arbitration.

PR #1416, opened 2018-03-02
Use btcNodes argument in regtest mode.

blabno:
This is needed for bisq-api and bitcoin docker containers to see each other during dockerized regtest.

### DIFF
--- a/core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
+++ b/core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
@@ -196,7 +196,7 @@ public class WalletsSetup {
         };
 
         if (params == RegTestParams.get()) {
-            configPeerNodesForRegTest();
+            configPeerNodesForRegTest(socks5Proxy);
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/bisq/core/btc/wallet/WalletsSetup.java
```diff
@@ -196,7 +196,7 @@ protected void onSetupCompleted() {
         };
 
         if (params == RegTestParams.get()) {
-            configPeerNodesForRegTest();
+            configPeerNodesForRegTest(socks5Proxy);
```
ManfredKarrer, line 199: I would prefer to have additional checks here and not call  configPeerNodes(proxy); from configPeerNodesForRegTest. What is the usecase that you need?

blabno: When each node runs in separate docker container they have different IPs and are not reachable by localhost. Regtest mode currently forces to use specific bitcoin node IP or tries to use localhost.
I need ability to point bisq to specific IP in regtest mode.

PR #1426, opened 2018-03-06
Fix height of blockconfirmation icon and reduces row height

### LINKED ISSUE #1421: Blockconfirmation indicator is too large
The block confirmation indicator is now larger as it it in the current release version. IMO it is too large in relation to other icons.

screen shot 2018-03-05 at 10 55 52

ripcurlx:
This will fix #1421


ManfredKarrer:
utACK

PR #1428, opened 2018-03-06
Re-calculate values when tabbing into percentage market distance field

### LINKED ISSUE #1411: Market price margin set to zero when tabbing through
When tabbing through the create offer interface. Distance is set to zero when field loses focus, but amount to receive and price is not updated.
Margin set to zero
Focus leaves margin field
Either we don't default it to zero when tabbing through or we update the corresponding fields.

I personally would keep the zero default and just update the corresponding values correctly, as it would help the user to finalize the offer without the need to think too much.

What do you think @ManfredKarrer?

ripcurlx:
Fixes #1411.


ManfredKarrer:
utACK

PR #1431, opened 2018-03-07
Use Amount term for Volume

ripcurlx:
This PR will add the last changes discussed with @ManfredKarrer recently. It will remove the redundant coin code in colume "Amount in {COIN_CODE}" and rename Volume to Amount in Failed trades, Open trades and History.


ManfredKarrer:
utACK

PR #1432, opened 2018-03-07
Change wording from distance to below/above

ripcurlx:
As discussed with @ManfredKarrer this PR will change the general wording "distance" to "below" or "above" where applicable. So it is easier for the user to understand if positive percentage means above or below the market price. See the following screenshots on the new state.
bildschirmfoto 2018-03-07 um 13 15 15
bildschirmfoto 2018-03-07 um 13 15 09
bildschirmfoto 2018-03-07 um 13 14 55
bildschirmfoto 2018-03-07 um 13 14 37


ManfredKarrer:
utACK

PR #1437, opened 2018-03-08
Extract seednode module to new repository

cbeams:
Problem
-------

Bisq seednodes are versioned and deployed independently from Bisq desktop clients, but have been managed here in the bisq-exchange repository as "just another submodule". This has led to out-of-sync versioning, and an unclear set of rules as to what defines a new seednode release and when those releases should be deployed by @bisq-network/seednode-operators.

Solution
-------
This pull request removes the seednode module entirely. The removed sources, along with related docs have been moved into a new repository at https://github.com/bisq-network/bisq-seednode. The seednode sources in the new repository are largely unchanged, and have only been modified as necessary to deal with the structural change, e.g. the Maven pom has been updated accordingly, docs have been updated accordingly, a README has been added, etc.

Notes
-----
The `package/stresstest/init.sh` file here in the exchange repository still has references to the SeedNode jar as if it were still buildable within this repository. As I am unfamiliar with this script and whether it is still actively used in any way, I have simply left it untouched.

**Please reveiew the newly extracted https://github.com/bisq-network/bisq-seednode repository as part of the review of this pull request,** and give this an ACK and merge if you are happy with both. Once this has been merged, I'll update the new bisq-seednode repository to officially release v0.7.0 and we can have all @bisq-network/seednode-operators upgrade accordingly.


ManfredKarrer:
Sounds good, will look into it.
All related to stresstest can be removed. I doubt that much will still work it was last used 2 years ago...

ManfredKarrer:
ACK

PR #1442, opened 2018-03-09
Fix vertical alignment

### LINKED ISSUE #1441: Layout of transaction list items is not vertically centered
Layout of transaction list items is not vertically centered in row ("Sent to:, Received with:"). 

screen shot 2018-03-08 at 20 19 40

ripcurlx:
This will fix #1441.


ManfredKarrer (approved):
utACK

PR #1443, opened 2018-03-09
Handle change output from genesis transaction

sqrrm:
Make sure any change from the genesis transaction is not treated as valid BSQ.

### DIFF
--- a/core/src/main/java/io/bisq/core/dao/node/consensus/GenesisTxVerification.java
+++ b/core/src/main/java/io/bisq/core/dao/node/consensus/GenesisTxVerification.java
@@ -20,6 +20,7 @@ package io.bisq.core.dao.node.consensus;
 import io.bisq.core.dao.DaoOptionKeys;
 import io.bisq.core.dao.blockchain.BsqBlockChain;
 import io.bisq.core.dao.blockchain.vo.Tx;
+import io.bisq.core.dao.blockchain.vo.TxOutput;
 import io.bisq.core.dao.blockchain.vo.TxType;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/bisq/core/dao/node/consensus/GenesisTxVerification.java
```diff
@@ -48,11 +49,17 @@ public boolean isGenesisTx(Tx tx, int blockHeight) {
...
-            txOutput.setUnspent(true);
-            txOutput.setVerified(true);
-            bsqBlockChain.addUnspentTxOutput(txOutput);
-        });
+        long totalOutput = 0;
+        for (int i = 0; i < tx.getOutputs().size(); ++i) {
+            TxOutput txOutput = tx.getOutputs().get(i);
+            totalOutput += txOutput.getValue();
+            // Handle change from genesis transaction, only count outputs below the issuance amount as BSQ outputs
+            if (totalOutput <= bsqBlockChain.getIssuedAmount().getValue()) {
```
ManfredKarrer, line 57: I would suggest to use the new paradign to do any logic/verification in dedicated classes. Might be a bit of overhead here but will help to isolate the rules from the specs from other code. So to create  GenesisTxOuputsController class where the check is done and if the check return true the state gets updated.

ManfredKarrer (changes requested):

ManfredKarrer:
Due the repo split I will merge the current PR as is. U can make a new one later from the new repo once moved.

ManfredKarrer (approved):

PR #1448, opened 2018-03-10
Move statistics and monitor modules to separate repositories

cbeams:
See:
 - https://github.com/bisq-network/bisq-monitor
 - https://github.com/bisq-network/bisq-statsnode


ManfredKarrer (approved):
utACK

PR #1449, opened 2018-03-10
Isolate 'gui' module and rename to 'bisq-desktop'

cbeams:
As part of the larger Bisq modularization and repository splitting effort, this patch:

 - Removes core, network and common modules
 - Hoists the gui module up to the root of the repository
 - Adapts the root pom, packaging scripts and docs to reflect the change

bisq-network/exchange is dead; long live bisq-network/bisq-desktop!


cbeams:
Note that I've already renamed bisq-network/exchange => bisq-network/bisq-desktop as well.

ManfredKarrer (changes requested):
Those classes are deleted but are required:
WMCCAddressValidator
AlcParams
StrayaParams
WMCCParams

cbeams:
Good catch! Looks like I missed properly sync'ing up to the latest HEAD of master. Will fix, thanks.

cbeams:
@ManfredKarrer, take another look now, should be good to go.

This PR is a great example of why it's good to do reviews even when you're sure you don't need it!

ManfredKarrer (approved):

ManfredKarrer:
Looks good now. utACK.

ManfredKarrer:
@cbeams  I will leave the merge to you, so you can sync with the other repos.

Issue #1214, opened 2018-01-18
Offer additional information/hints whereever price distance in % (dynamic pricing) is used


ripcurlx:
There are multiple places in the app where the a percentage number is shown next to the price or can be entered. Although we display e.g. in the create offer screen that you can enter the "Distance in % from the market price" it is not explained if this will adapt if the bitcoin price is changing. Also the user might not be 100% sure if the BitcoinAverage price will be used or another price feed. One solution could be to add a (i) icon next to the input field displaying this information. The same can be done in tables where we display the price e.g. 8020 (4%) --> add an (i) icon next to the percentage explaining the reasoning behind this number.

ripcurlx:
If we render a bar underneath the price itself for the price distance/spread it could look like this. Red would mean price distance is negative for the offer taker and green means it is positive for the offer taker. 
![sell screen](https://user-images.githubusercontent.com/170962/35928468-83e87136-0c2d-11e8-8b47-3ab3c79c6179.png)

ripcurlx:
Of course most of the time it will be red underlays and no underlay if the price distance is 0%

ripcurlx:
See https://github.com/bisq-network/exchange/issues/1203#issuecomment-368877577 how this could be done for the create offer screen.

GuiSim:
What would the content of the (i) "popup" be? How would you explain this difference.

ripcurlx:
@GuiSim Did you see my comment from the other issue https://github.com/bisq-network/exchange/issues/1203#issuecomment-368877577? It wouldn't be an (i) icon, but just the sync icon and on roll over we have enough space to explain the exact details on this dynamic price distance offer.
![sell screen no flags](https://user-images.githubusercontent.com/170962/36736195-4b6ff120-1bd8-11e8-8366-f5a4fd580d6c.png)

GuiSim:
I had missed that. Would you prefer to have comments on this tooltip here?

I still feel like it's not super clear what "below the market price" really entails. I had written some comments in https://github.com/bisq-network/exchange/issues/1191#issue-288180241 but essentially I wonder if the text should be different if you're buying VS selling.

I would also suggest that instead of saying "and updated every minute" it could say "This offer is 2% below the market price as of `X seconds ago`." or similar. This would make it more obvious when was the last update and also would make it more apparent that the sync button worked.

"Updated every minute" could be a setting or explained elsewhere, it doesn't have to be on every line.

[...]

Issue #1203, opened 2018-01-16
Make sell/buy BTC screen more self-explanatory


ripcurlx:
Add information inline for price distance and volume range and add hint line before the create offer call-to-action button. Here's a rough sketch how it could look like:
![sell bitcoin](https://user-images.githubusercontent.com/170962/34993334-4cddd640-fad0-11e7-9ffc-177aa4286e95.png)

tr37ion:
Please don't forget to dot-align the figures #1028

ripcurlx:
Good point! I'll have a look how to do this in a consistent way. Sometimes it is a little bit tricky when we also have to display ranges.

tr37ion:
For now you could just switch to `Fira Mono` font, which looks like a sans-serif font  and then just fill up the values with zeros to the right, so that values align nicely to the right. It is important to have a mono-space font, because then values are vertically comparable.

Sure a dot-aligned solution would be the best solution in the end.

ripcurlx:
@fira mono: I'll have a look how it looks like mixing these two fonts within the app and the size of the font, as we have to bundle it within the app.

tr37ion:
Fira Mono is one of the few fonts which scales very well. It is perfectly readable even with very small sizes. 

![image](https://user-images.githubusercontent.com/1538617/35221974-878d0a8a-ff7c-11e7-816b-7122e3236895.png)

ripcurlx:
An updated version of the buy/sell screen up for discussion. @ManfredKarrer & @tr37ion let the discussion begin 😉 
![sell screen](https://user-images.githubusercontent.com/170962/35856944-d214a934-0b37-11e8-8e34-9c4f03bebeef.png)

ripcurlx:
Or the same with the call-to-action button below.
![sell screen](https://user-images.githubusercontent.com/170962/35859847-f6678d56-0b41-11e8-9ba7-a23529ca81cc.png)

tr37ion:
I got this layout. Positioning the filters and the `Create` button is not easy.

![image](https://user-images.githubusercontent.com/1538617/35861388-6b1d0494-0b3f-11e8-91f5-29485386e4de.png)

ripcurlx:
[...]

PR #1463, opened 2018-03-16
Making Buy/Sell BTC screen more user friendly (Part 1)

ripcurlx:
This will implement most changes of #1203. As the suggested changes in the price column are also dependent on #1214, they will be implemented as soon as a final proposal is ready for it as well.

Dependent on PR https://github.com/bisq-network/bisq-common/pull/5


ManfredKarrer:
concept ACK

PR #1473, opened 2018-03-21
Migrate build from Maven to Gradle

cbeams:
Per bisq-network/dao#41


package/osx/create_app.sh
```diff
@@ -51,8 +43,6 @@ $JAVA_HOME/bin/javapackager \
     -vendor Bisq \
     -outdir deploy \
     -srcfiles "deploy/Bisq-$version.jar" \
-    -srcfiles "target/lib/bcpg-jdk15on.jar" \
-    -srcfiles "target/lib/bcprov-jdk15on.jar" \
     -appclass bisq.desktop.app.BisqAppMain \
```
ManfredKarrer, line 46: How are the BC jars now included? They have been earlier in the jdk dir under ext but Cedric moved them into the maven build script and the inclusion in the package script was needed.

ManfredKarrer (approved):
ACK

PR #1478, opened 2018-03-27
Voting

ManfredKarrer:


ripcurlx:
utACK.
Just a heads up: I had some issues with my plan to slowly migrate all fontawesome icons to the new version, so it would be great if everyone from now on uses the `getIcon` method in Formbuilder to create the label icons. Will make it easier in the end to do the migration. You'll see it as soon as https://github.com/bisq-network/bisq-desktop/pull/1479 is merged.

ManfredKarrer:
@ripcurlx Thanks will take care in future of those.

sqrrm (approved):
utACK

PR #1486, opened 2018-03-30
Fix bugs in issuance process and tx view

ManfredKarrer:


sqrrm (approved):
utACK

Once the minimum requirements for requests, issuance, voting and bonding is done I think we should open for more feedback on how to present this in the UI. I currently find it a bit confusing and I'm familiar with how the transactions work.

PR #1497, opened 2018-04-05
Add new domain for handling parameter changes

ManfredKarrer:
- Add DaoParam, ParamChangeEvent, ParamChangeEventList and DaoParamService
- Move all default values to DaoParam
- Access fees, quorum and voteThreshold from DaoParamService
- Remove handling those param data from BsqBlockChain


sqrrm (approved):
utACK

PR #1500, opened 2018-04-06
Enable distZip flag in "build.gradle".

blabno:
This flag is required to run `./gradlew install` which in turn is required by jitpack.


ripcurlx:
utACK

PR #1506, opened 2018-04-10
Remove offer duplicates in spread view

### LINKED ISSUE #1202: Offer counts in Market->Spread tab temporarily doubles
Here are two screenshots, one taken right after the other.

screen shot 2018-01-16 at 3 05 00 pm

image

Note how the offer counts in the first are exactly double the counts in the second.

The only thing I did in between these two screenshots was click on "Buy BTC" and "Sell BTC" in order to examine whether there was a single market maker responsible for many of these trades. I noticed immediately on those screens that the numbers were off (there weren't ~180 trades there), and when I came back to the Spread tab, I saw the view in the second screenshot.

Whatever is causing this behavior is probably what was going on when I tweeted this the other day: https://twitter.com/bisq_network/status/952485704994639872. So unfortunately, that was probably a false "record" being set.

ripcurlx:
Fixes #1202.

I was only able to reproduce it in a test in the SpreadModel so I added an additional safeguard there.

### DIFF
--- a/src/main/java/bisq/desktop/main/market/spread/SpreadViewModel.java
+++ b/src/main/java/bisq/desktop/main/market/spread/SpreadViewModel.java
@@ -51,6 +51,10 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 class SpreadViewModel extends ActivatableViewModel {
@@ -90,6 +94,12 @@ class SpreadViewModel extends ActivatableViewModel {
         offerBookListItems.removeListener(listChangeListener);
     }
 
+    private static <T> Predicate<T> distinctByKey(Function<? super T, ?> keyExtractor)
+    {
+        Set<Object> seen = ConcurrentHashMap.newKeySet();
+        return t -> seen.add(keyExtractor.apply(t));
+    }
+
     private void update(ObservableList<OfferBookListItem> offerBookListItems) {
         Map<String, List<Offer>> offersByCurrencyMap = new HashMap<>();
         for (OfferBookListItem offerBookListItem : offerBookListItems) {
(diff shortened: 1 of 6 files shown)


src/main/java/bisq/desktop/main/market/spread/SpreadViewModel.java
```diff
@@ -90,6 +94,12 @@ protected void deactivate() {
         offerBookListItems.removeListener(listChangeListener);
     }
 
+    private static <T> Predicate<T> distinctByKey(Function<? super T, ?> keyExtractor)
+    {
```
cbeams: Nit: place opening brace on same line

src/test/java/bisq/desktop/main/market/spread/SpreadViewModelTest.java
```diff
@@ -58,14 +62,38 @@ public void testMaxCharactersForAmountWithNoOffers() {
...
-        offerBookListItems.addAll(make(btcItem.but(with(OfferBookListItemMaker.amount, 1403000000L))));
+        offerBookListItems.addAll(make(btcBuyItem.but(with(OfferBookListItemMaker.amount, 1403000000L))));
         assertEquals(7, model.maxPlacesForAmount.intValue()); //14.0300
     }
+
+    @Test
+    public void testfilterSpreadItemsForUniqueOffers() {
+        OfferBook offerBook = mock(OfferBook.class);
+        PriceFeedService priceFeedService = mock(PriceFeedService.class);
+        final ObservableList<OfferBookListItem> offerBookListItems = FXCollections.observableArrayList();
```
cbeams: Nit: don't use `final` modifier on local variables unless strictly necessary (same for line 85 below.

cbeams (approved):
utACK. Great that you were able to repro this at all, @ripcurlx! You'll see I called out a couple style nits, but looks good otherwise. Feel free to merge.

src/test/java/bisq/desktop/main/market/spread/SpreadViewModelTest.java
ripcurlx: I use it out of a habit. Does it have a disadvantage to use final everywhere by default and only not use it where necessary?

PR #1509, opened 2018-04-14
 Move out startup code from MainViewModel

ManfredKarrer:
- I extracted all startup code to AppSetupFullApp
- The MainViewModel implements StartupHandler where all relevant callbacks are executed

Was done in 2 hours after a long day, so it's all not polished at all...
Is not tested much but at least the headless starts up as expected and the Desktop looks also OK without deeper testing.

I would not recommend to merge that, I just wanted to show that it is really not that hard to get a headless Bisq core app running. If we want to merge then a more detailed comparision with the old version and testing is required. It can easily be that some of the bindings are missing and some of not so obvious use cases and error paths are not covered correctly.

Of course the startup code should be refactored and split in smaller chunks, but nothing prevents anyone to just start on that.
The handler class would be also better to be separated by diff. use cases (e.g. warnings, ...).

### DIFF
--- a/src/main/java/bisq/desktop/main/MainViewModel.java
+++ b/src/main/java/bisq/desktop/main/MainViewModel.java
@@ -31,74 +31,35 @@ import bisq.desktop.util.BSFormatter;
 import bisq.desktop.util.GUIUtil;
 
 import bisq.core.alert.Alert;
-import bisq.core.alert.AlertManager;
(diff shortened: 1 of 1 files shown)


ManfredKarrer:
@cbeams Feel free to close it once the basic idea has arrived. Was not intended to get merged.

PR #1510, opened 2018-04-14
Update CreateOfferDataModel.java

marco4444:
Round Fiat values to significantly reduce the likelihood of identity disclosure


ManfredKarrer:
Can you just make a normal issue for that? PRs are for code or document changes.

cbeams:
Closing as invalid (i.e. not a valid PR). The change itself is worth considering, and is something I've thought about as well. Should make it an option, and not force it though. Possibly even make it the default, and then allow people to opt out of it. Consider writing a proposal issue for this, though (in https://github.com/bisq-network/proposals). It will require discussion and thought, and just adding an issue (to https://github.com/bisq-network/bisq-desktop) will probably go nowhere.

PR #1513, opened 2018-04-17
Feature/startup refactoring 2

blabno:
This is just a proposition of startup refactoring.

Have a look at the context of complimentary PR https://github.com/bisq-network/bisq-core/pull/76.


blabno:
Since the approach has been rejected I'm closing this proposal.

PR #1515, opened 2018-04-17
This will be a series of small scale and low risk changes for improving startup

ManfredKarrer:


blabno (approved):
tACK

cbeams (approved):
utACK

cbeams (approved):
ACK (tested against related merges in common and core)..

PR #1516, opened 2018-04-17
Remove dependency on stage from bisq module

blabno:
This is preparation for moving DesktopModule instantiation out of BisqApp which is controlled by JavaFX.

Merge #1515 first.


ManfredKarrer:
This approach would have worked as well. But I found that it was not needed at all to use Stage via Guice. The views which used it could access it. It is already removed from the module as well from view client classes as part of my PRs.

PR #1519, opened 2018-04-24
Change startup order to prevent Bouncy Castle exception

ripcurlx:
After talking it through with @ManfredKarrer I changed the startup order based on the suggestion by @blabno


ManfredKarrer (approved):
utACK

PR #1520, opened 2018-04-24
Prevent editing of amount

ripcurlx:
Prevent editing of min amount and max amount for now, as it would need a change in reserved funds for higher amounts.


ManfredKarrer (approved):
utACK

PR #1521, opened 2018-04-26
Improvements for edit offer

ripcurlx:
Dependent on https://github.com/bisq-network/bisq-core/pull/102


ManfredKarrer (changes requested):
Looks good as far I tested. Only small thing is that the title in the box "Set amount and price" is not correct anymore as we don't allow editing amount.

ripcurlx:
@ManfredKarrer please merge bisq-network/bisq-core#102 for new translations

ManfredKarrer (approved):
utACK

ManfredKarrer:
Actually it was tested, just last commit was untested....

PR #1528, opened 2018-05-01
[WIP] Release preparations for v0.7.0

ripcurlx:
We need to adapt the current release script to work with the new gradle setup. At the moment it fails during packaging with
```
Bundler DMG Installer skipped because of a configuration problem: Main application jar is missing.  
Advice to fix: Make sure to use fx:jar task to create main application jar.
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ build
 desktop.ini
 */target/*
 *.class
+deploy
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.github.jengelman.gradle.plugins:shadow:2.0.2'
+        classpath 'com.github.jengelman.gradle.plugins:shadow:2.0.4'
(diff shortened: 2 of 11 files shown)


XuezProject:
Hey there - Been keeping my Eye on this, Do you have an ETA for the next release?

cbeams:
@XuezProject as soon as possible.

cbeams:
@ManfredKarrer, with commit 4a78010, running `./package/osx/create_app.sh` now works as expected.

There are some other issues, though, with the new qualified approach to the bundle name, e.g. using `CAT` to signify **C**hristoph **At**teneder. For example, this shows up as the application name in the macOS menu bar. I guess this was all about devising a way that different maintainers can release the application, and I haven't fully caught up on Christoph's plan here. Perhaps we can work through it together / figure out what to do next.

I'll need to sit down with you to discuss this and to work through the other build steps and make sure everything else works as expected. I'll ping you offline to set up some time for this.

ManfredKarrer:
Ah cool. I will try it out soon. The bundle name I will revert anyway as I will do the build now.

PR #1535, opened 2018-05-08
Incorporate http api

blabno:


cbeams (changes requested):
NACK. There are many things that need to be dealt with before anything like this PR can be merged.

As we discussed in our call about this (https://youtu.be/RkN2YbFXBsU?t=1h8m9s), we're not going to ship the HTTP API integration until it's ready for primetime. Here is a probably incomplete list of what that means:

 - Project packaging should change from `io.bisq` => `network.bisq`.

 - Project name should change. `bisq-api` is ambiguous. `bisq-http-api` would be workable, or perhaps just `bisq-http`. This can be discussed elsewhere.

 - The build should be ported from Maven to Gradle, so that the project can participate in the larger composite build with all other Bisq software components. I'm happy to help with that porting effort, so long as someone from the team shadows me while I do it, in order to transfer knowledge.

 - The project needs to be thoroughly code reviewed.

 - User-focused documentation needs to be written. This is HUGELY IMPORTANT. I talked with @blabno about this at the end of the call linked above, after @mrosseel and @ManfredKarrer had left, and I'll repeat it here below...

Let's assume that all the issues above were resolved and we did merge this PR right now. What would be the next steps? How would potential users know how to use the HTTP API? They could dig around and find the project README, they might find the swagger documentation, the might find the outdated YouTube video. If they really dig, they might find the (AFAICT undocumented) `support` directory and the scripts within, but all of this is a far cry from delivering first-class software. If it's not documented, it doesn't exist. Especially when what we're talking about is an API. Bisq has gotten away without real documentation for most of its existence because it's a GUI and people can more or less click their way around to figuring stuff out. But even that must change now, and [is](https://github.com/bisq-network/bisq-docs/issues/37) [changing](https://github.com/bisq-network/bisq-docs/pull/45) as we speak.

With an API, we absolutely must provide high-level documentation that helps take the uninitiated user from zero to Bisq-scripting hero.

At a minimum, I'd say this means a "Getting Started with the Bisq HTTP API" doc, in which you walk the user through, step by explicit step, what they must do to build a fully-functional, albeit very simple trading bot. Indeed it should be the simplest bot that could possibly work. The code for that guide should be published along with it, so that people can check themselves against code that definitely works in case they get stuck. There is much to discuss putting together such a guide, and we obviously don't want to do that here in the context of this PR, but my point is that such documentation must be written in order to consider this anything like delivered software.

If we have proper getting started resources like the one detailed above, then it's easy to announce and promote the new functionality. We can tweet about it and communicate through other channels to turn as as many users' heads as possible, and to in turn get as much early feedback as possible. If we don't have such resources available, then we can announce and promote the effort all we want, but it will fail to get significant traction, because people are busy, and anyone other than the most motivated user will not take the time to dig through our materials to figure it all out for themselves.

If it's not obvious, this whole bit about documentation goes back to the definition of delivered software as presented in https://github.com/bisq-network/proposals/issues/19:

> **delivered work is work that:**
> 
> - users can use immediately because it **has been released**
> - users can figure out how to use because it **has been documented**
> - users know exists in the first place because it **has been announced and promoted**

There are a number of smaller issues with this PR, like the fact that the `dependencyVerification` block was blown away, that imports are not properly organized, etc, but they don't matter, given that this PR isn't going to be merged any time soon.

While we could prefix this PR with `[WIP]` and iterate on it, I'd prefer to close it and then open a new PR when we're actually getting close to being ready with all the stuff mentioned above.

If any of this is unclear, or if there is disagreement about it, I would do another call. **What I would really like to see, though, is a proposal that writes up exactly what the plan is for integrating the HTTP API, i.e. the steps I listed above and anything else I missed, who's going to be on point for what, etc, such that we are aligned about the criteria for shipping this work.** That proposal and its comments are a place where we can iron out the specifics of the naming and packaging and other decisions that need to get made.

To set expectations, and this is consistent with what we discussed on the call, I would not expect this functionality to get shipped in a production Bisq Desktop release until at least 0.7.1 or 0.7.2. Assuming we start shipping on a monthly cadence after 0.7.0, it would be ambitious indeed to check every box in the list above to get things in by 0.7.1, but I'm happy to work with you guys to move as quickly as possible to get this done. Just need to make sure everybody is aligned about what _done_ actually means.

blabno:
This PR is to show how small changes need to be done to incorporate the API.
This is not compensation request so I don't see any point of not merging the code without documentation.
From our conversation I understood that we should include the API in bisq ASAP so that core team can be aware what needs to be changed in core alone to make Bisq work in headless mode with new interfaces (http, rpc).
Manfred has already mentioned that he's not reviewing the code outside of the bisq-network.

PR #1539, opened 2018-05-14
[WIP] Redesign - Market section

ghost:
Following the conversation with @pedromvpg  and @ripcurlx  on how to tackle the redesign of the bisq client to achieve Pedro’s initial proposal, we agreed that the ideal next steps would be to start introducing small ux improvements and at the same time change the ui to the new style to help bridge the gap before a full redesign. 

With that in mind I created some mockups for the Market section, that includes changes to Offer Book, Spread, Trades and to some of the main components of the client (Navigation and Tickers).

**Notes:**
Offer book was very confusing to me when I first saw it. The chart displays Buy/Sell, then the offer tables switch those around and then CTA’s are switched again. This cross referencing is confusing for a first time user.

I tried to improve it and keep everything consistent to what users might be used to or expecting from other exchanges. 

Buy Offers Chart | Sell Offers Chart
--- | --- 
Buy Offers Table | Sell Offers Table
Create Buy Offer | Create Sell Offer

This way everything is aligned properly, easier to understand and we eliminated the need to cross reference in the UI. Left is all about **buys**, right is all about **sells**.

To make it easier for users to take offers from the tables I added a “Take Offer” button. Then the main CTA is all about creating an offer for that specific table. 

Feedback, please.


ghost:
**Offer book:**
market-offer-book

**Spread**
market-spread

**Trades**
market-trades

m52go:
This is great! Overall I love the clean approach to layout, colors, and icons. 

One design element I miss from the current build is seeing the market price front and center. I wonder if there's a way to emphasize it in a way that fits this new aesthetic.

diogorsergio:
Yeah, definitely!
Thats what I’m looking for the more feedback the best.

PR #1553, opened 2018-05-23
Remove validation for Argentinian account numbers and allow manual edit

### LINKED ISSUE #1543: Account number should be editable (Argentina)
We need you to change the field "Número de cuenta" / "Account number" from read-only to editable for recently supported for Argentina:
- Cash Deposit
- National bank transfer
- Transfer with same bank
- Transfers with specific banks

Looks like that number doesn't follow a rigid rule, so banks are able to change the way they create it. It would be great if users could edit that field after it is auto-filled based on the CBU number. Additionally, special characters  '-' and '/' should be allowed.

In summary: 
- please don't remove the currently auto-filled
- allow users to change that value (at least '-' and '/' should be allowed)
- The change should be applied to Cash Deposit | National bank transfer | Transfer with same bank | Transfers with specific banks

https://user-images.githubusercontent.com/4233784/38172661-d7b47862-35b0-11e8-8438-cea521a1ac2c.png

Thank you!

ripcurlx:
Fixes #1543.

### DIFF
--- a/src/main/java/bisq/desktop/components/paymentmethods/BankForm.java
+++ b/src/main/java/bisq/desktop/components/paymentmethods/BankForm.java
@@ -52,6 +52,9 @@ import javafx.scene.control.Label;
 import javafx.scene.control.TextField;
 import javafx.scene.layout.GridPane;
 
+import javafx.beans.value.ChangeListener;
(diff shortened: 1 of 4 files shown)


ManfredKarrer (approved):
utACK

PR #1556, opened 2018-05-25
Fix reuse of change addresses

ManfredKarrer:
Depends on https://github.com/bisq-network/bisq-core/pull/113


ripcurlx (approved):
utACK - Although it is a little commit mix of different PRs I ACK the changes as we discussed on Slack.

PR #1557, opened 2018-05-25
Remove venmo cashapp

ManfredKarrer:
Depends on https://github.com/bisq-network/bisq-core/pull/115


ripcurlx (approved):
utACK with the same comment as on the other PRs that we discussed the commit mix of the recent PRs on Slack

PR #1558, opened 2018-05-26
[WIP] Round volume of fiat amounts

ManfredKarrer:
- See: https://github.com/bisq-network/bisq-desktop/issues/1511

Depends on https://github.com/bisq-network/bisq-core/pull/116

@ripcurlx Not tested much yet but seems it covers all appearances of volume. Main question is how we avoid confusion of users if the volume amount does not change when the user changes the price.


ManfredKarrer:
We just merged another PR which implements that feature.

PR #1578, opened 2018-06-21
Fix select currency issues after edit offer

### LINKED ISSUE #1565: Bug with fixed currency at edit offer
After  a user has clicked edit offer the currency will stick in create offer views as well as in the edit offer view.
To reproduce create 2 offers with fiat and an altcoin. Then go to edit offer at both and back to create offer. Switch payment method at create offer. The currency does not change anymore. At the edit offer it shows the same currency for both offers. 

See also: https://bisq.community/t/create-buy-offer-wrong-currency/5573/11

ripcurlx:
Fixes #1565.


src/main/java/bisq/desktop/main/portfolio/editoffer/EditOpenOfferDataModel.java
```diff
@@ -49,14 +49,21 @@
...
+        super(openOfferManager, btcWalletService, bsqWalletService, preferences, user, keyRing, p2PService,
+                priceFeedService, filterManager, accountAgeWitnessService, tradeWalletService, feeService, formatter);
     }
 
     public void initWithData(OpenOffer openOffer) {
         this.openOffer = openOffer;
         this.initialState = openOffer.getState();
         this.paymentAccount = user.getPaymentAccount(openOffer.getOffer().getMakerPaymentAccountId());
+        // select the current offer payment account as default payment account
+        preferences.setSelectedPaymentAccountForCreateOffer(paymentAccount);
```
ManfredKarrer: Is that needed in the base class or should that be only called in the createOffer sub class? If so, maybe better to overwrite initWithData there.

src/main/java/bisq/desktop/main/portfolio/editoffer/EditOpenOfferView.java
```diff
@@ -131,9 +131,9 @@ protected void deactivate() {
     ///////////////////////////////////////////////////////////////////////////////////////////
 
     public void initWithData(OpenOffer openOffer) {
+        model.initWithData(openOffer);
```
ManfredKarrer: I would prefer if we don't have that call before the super call. Carries risky dependencies of order of calls (see my comment in other commit - maybe that solves the issue without changing here the order?)

ManfredKarrer (changes requested):

src/main/java/bisq/desktop/main/portfolio/editoffer/EditOpenOfferDataModel.java
ripcurlx: Actually this is already the edit offer sub class. The naming of the method is misleading, I'll rename it to a generic one.

ripcurlx:
@ManfredKarrer: I think I somehow f** up this PR. I'll delete it an re-open a new one with a different strategy that makes it even easier to solve this issue 😄

PR #1581, opened 2018-06-22
Handle unknown PB data from encrypted messages

ManfredKarrer:
https://github.com/bisq-network/bisq-common/pull/25 , https://github.com/bisq-network/bisq-p2p/pull/11 and https://github.com/bisq-network/bisq-core/pull/122

- Use ProtobufferException for
CoreNetworkProtoResolver.fromProto(PB.NetworkEnvelope proto)
- Add raw PB data in case if an exception
- log error if the encrypted data contains unknown PB data

Note: We should refactor the resolvers with checked exceptions instead
of the ProtobufferRuntimeException, but for the moment we leave that to
not add too much complexity/risk for that release.


sqrrm (approved):
utACK

PR #1583, opened 2018-06-25
Set default payment account for an edited offer to null

### LINKED ISSUE #1565: Bug with fixed currency at edit offer

ripcurlx:
Fixes #1565.

### DIFF
--- a/src/main/java/bisq/desktop/main/offer/EditableOfferDataModel.java
+++ b/src/main/java/bisq/desktop/main/offer/EditableOfferDataModel.java
@@ -89,6 +89,8 @@ import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
 
+import javax.annotation.Nullable;
+
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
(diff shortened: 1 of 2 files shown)


src/main/java/bisq/desktop/main/offer/EditableOfferDataModel.java
```diff
@@ -245,7 +245,7 @@ public boolean initWithData(OfferPayload.Direction direction, TradeCurrency trad
         fillPaymentAccounts();
 
         PaymentAccount account;
-        PaymentAccount lastSelectedPaymentAccount = preferences.getSelectedPaymentAccountForCreateOffer();
+        PaymentAccount lastSelectedPaymentAccount = getPreselectedPaymentAccount();
```
ManfredKarrer: Can we add @Nullable annotations for all nullable fields?

ManfredKarrer (approved):
utACK

I don't have enough context to reason about the change. Will checkout the code once its complete and review all again.

src/main/java/bisq/desktop/main/offer/EditableOfferDataModel.java
ripcurlx: Done

ManfredKarrer (approved):
utACK

PR #1588, opened 2018-06-26
Add popup for feedback after trade completion

ManfredKarrer:
See: https://github.com/bisq-network/bisq-desktop/issues/1584

Depends on: https://github.com/bisq-network/bisq-core/pull/125


sqrrm (approved):
utACK

PR #1590, opened 2018-06-26
Add support for referrer IDs at offers and trades

ManfredKarrer:
See: https://github.com/bisq-network/proposals/issues/28

Depends on: bisq-network/bisq-common#28 and https://github.com/bisq-network/bisq-core/pull/127

### DIFF
--- a/src/main/java/bisq/desktop/main/settings/preferences/PreferencesView.java
+++ b/src/main/java/bisq/desktop/main/settings/preferences/PreferencesView.java
@@ -40,6 +40,7 @@ import bisq.core.locale.LanguageUtil;
 import bisq.core.locale.Res;
 import bisq.core.locale.TradeCurrency;
 import bisq.core.provider.fee.FeeService;
+import bisq.core.trade.statistics.ReferralIdService;
 import bisq.core.user.BlockChainExplorer;
 import bisq.core.user.Preferences;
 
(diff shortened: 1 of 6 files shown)


sqrrm (approved):
utACK

src/main/java/bisq/desktop/main/settings/preferences/PreferencesView.java
```diff
@@ -298,6 +299,13 @@ public BaseCurrencyNetwork fromString(String string) {
                 preferences.setIgnoreTradersList(Arrays.asList(StringUtils.deleteWhitespace(newValue)
                         .replace(":9999", "").replace(".onion", "")
                         .split(",")));
+
+        // referralId
+        referralIdInputTextField = addLabelInputTextField(root, ++gridRow, Res.get("setting.preferences.refererId")).second;
```
sqrrm, line 304: Perhaps this field could get a grayed out text that disappears when focusing saying something like "optional referral code" to make it clear it's just optional and generate less confusion.

ghost: I think also it would be good to make it clear that it's just optional.

ManfredKarrer: Yes, good point. I also though it might require some extra info. I will add a prompt text when not filled.

PR #1591, opened 2018-06-27
Fix sorting of currencies in market/trades view

ManfredKarrer:
Currencies in the markets/trades view are sorted by number of trades (if selected that option in
preferences). The PriorityQueue used for the list do not apply the
sorting of the list (only sorts when taking an element out). We use a
ArrayList instead to apply sorting.


sqrrm (approved):
utACK

PR #1607, opened 2018-07-20
Implements Face to Face payment method

ManfredKarrer:
Implements Face to Face payment method

This is a very basic version without any of the advanced features discussed at https://github.com/bisq-network/proposals/issues/24
It assumes that both traders carry their laptop to the meeting.
We could start with that and see how much interest it creates.
If positive we can add support for the mobile support as discussed here:
https://github.com/bisq-network/proposals/issues/24#issuecomment-406693836

Screenshots:
https://github.com/bisq-network/proposals/issues/24#issuecomment-406749665

Depends on:
https://github.com/bisq-network/bisq-common/pull/30
https://github.com/bisq-network/bisq-core/pull/138

### DIFF
--- a/src/main/java/bisq/desktop/bisq.css
+++ b/src/main/java/bisq/desktop/bisq.css
@@ -286,6 +286,10 @@ bg color of non edit textFields: fafafa
     -fx-background-color: -bs-bg-light;
 }
 
+.text-area {
+    -fx-prompt-text-fill: derive(-fx-control-inner-background, -30%);
+}
+
 #label-url {
     -fx-cursor: hand;
     -fx-text-fill: -bs-blue;
--- a/src/main/java/bisq/desktop/components/paymentmethods/F2FForm.java
+++ b/src/main/java/bisq/desktop/components/paymentmethods/F2FForm.java
@@ -1,0 +1,261 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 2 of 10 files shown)


sqrrm (approved):
utACK

ripcurlx (approved):
ACK. Did a test trade and everything looks fine. Not sure if the link in the docs will be a top level anchor in https://docs.bisq.network or if it will be just a link to an additional document under the https://docs.bisq.network/#features group. @cbeams what's your take on the link structure?

ManfredKarrer:
@ripcurlx I think the text for the F2F trade rules, suggestions, guide will be quite a bit, so I think a dedicated page might be worth for it. But leave it to @cbeams where he thinks it fits best. We can also use a global referrer link (e.g. f2f.bisq.network or bisq.network/f2f) and then forward where ever it should be....

cbeams:
At https://docs.bisq.network/, there's an entry in the index that reads:

> Trading and Arbitration Rules — _Know what’s expected of you and others when trading_

This doc has not yet been written, but would take what's in the FAQ and elsewhere about the rules of trading and arbitration, and would locate them in a single, first-class doc with the URL https://docs.bisq.network/rules.html.

By default, I would want to keep the F2F rules in the same doc (one stop shopping), at e.g. https://docs.bisq.network/rules.html#face-to-face-trading.

If the F2F content is too long, and swamps the rest of the document, then we could do a separate rules-f2f.html doc. Hopefully not necessary, though.

In any case, it would be a good time to write the rules doc. IIRC, @ManfredKarrer, you and @m52go discussed this elsewhere recently. Steve, are you interested in writing that? I can help direct you to the stuff that is already documented, and can be available for Q&A / review along the way.

m52go:
Yup. Agree that it makes sense to keep them together in 1 doc. I'll take a closer look when I get back near a computer in the next 4-5 days.

ghost:
In this intended F2F version,
will it always be mandatory to have BTC on one side of a trade ?
or will it be possible to trade eg fiat for XMR ?
(per reddit question)

ManfredKarrer:
@HarryMacfinned In the current version we just use the normal Bisq trade protocol, so yes BTC is required. If F2F gets some traction we might evolve and add a off-chain trade protocol and then there are no restrictions on BTC anymore (but likely on BSQ).

PR #1610, opened 2018-07-25
Refact generics

ManfredKarrer:
Use generics for type safety at ComboBoxes and ListViews in Formbuilder.

Depends on:
https://github.com/bisq-network/bisq-common/pull/31
https://github.com/bisq-network/bisq-core/pull/139


sqrrm (approved):
utACK

Would be good if @ripcurlx could have a proper look as well though.

ripcurlx (approved):
utACK - Looks fine from my point of view

PR #1619, opened 2018-08-10
Add HalCash

ManfredKarrer:
Depends on:
https://github.com/bisq-network/bisq-common/pull/32
https://github.com/bisq-network/bisq-core/pull/144

### DIFF
--- a/src/main/java/bisq/desktop/components/paymentmethods/HalCashForm.java
+++ b/src/main/java/bisq/desktop/components/paymentmethods/HalCashForm.java
@@ -1,0 +1,112 @@
+/*
+ * This file is part of Bisq.
+ *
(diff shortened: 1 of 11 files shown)


ripcurlx:
bildschirmfoto 2018-08-14 um 17 59 36
Editing of HalCash offer doesn't work properly atm. The price is updated, but the amount stays the same.

ripcurlx:
I just recognized that the min amount is updated. Just looked like an error having a min-max range of 50-50.

ManfredKarrer:
Hm... yes thats not good. I will have a look

ripcurlx (approved):
ACK - Tested it locally and everything worked except some edge cases when editing an offer. When the edit offer change discussed in Zoom is implemented everything is OK from my side.

PR #1620, opened 2018-08-13
Properly reset cell in amount column

### LINKED ISSUE #1613: Amount table cell in trade statistics not reset
When the trade statistics table got filled and you switch to another currency which has just a few entries the empty rows still have the amount field set from the previous data.

ripcurlx:
Fixes #1613.


ManfredKarrer (commented):
utACK

Thanks for fixing it!

ManfredKarrer (approved):

PR #1621, opened 2018-08-14
Add warning for languages not natively supported by arbitration

### LINKED ISSUE #1612: Add popup if user has non-english language
If the user has a non-english language selected (or default selected because the OS is in another language) we should display a popup to tell the user that the arbitrators are english-only atm.

That popup can be displayed in 3 situations:
- When selecting another language in settings
- When making an offer (only if no language match)
- When taking an offer (only if no language match)

The arbitrator selection is language based, so if we have one day other languages supported by arbitrators we can detect that by checking the supported arbitrator field in the offer and only display the popup if there is no match of arbitrator's language and user's language.

We should add that to the next release and translate the popup text to all yet supported languages!
We need to push the translators to translate that with high priority to get it into the release.

Suggested text in context of offers:
Please be aware that in case of a dispute the arbitrator(s) selected for that offer are not capable of {0}. The language(s) used in case of a dispute is: {1}

Suggested text when switching language in settings:
Please be aware that in case of a dispute the available arbitrator(s) are not capable of {0}. The language(s) used in case of a dispute would be: {1}

{0} -> users language. E.g. "French"
{1} -> arbitrators language(s). E.g. "English, German"

@m52go @ripcurlx 
Please feel free to improve the suggested text...

ripcurlx:
Fixes #1612.

Dependent on https://github.com/bisq-network/bisq-core/pull/145.

### DIFF
--- a/src/main/java/bisq/desktop/main/offer/BuyOfferView.java
+++ b/src/main/java/bisq/desktop/main/offer/BuyOfferView.java
@@ -21,6 +21,7 @@ import bisq.desktop.Navigation;
 import bisq.desktop.common.view.FxmlView;
 import bisq.desktop.common.view.ViewLoader;
 
+import bisq.core.arbitration.ArbitratorManager;
 import bisq.core.user.Preferences;
 
 import javax.inject.Inject;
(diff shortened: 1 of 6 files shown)


ManfredKarrer (approved):
utACK

PR #1624, opened 2018-08-15
Add mobile notifications

ManfredKarrer:
Depends on: 
https://github.com/bisq-network/bisq-common/pull/33
https://github.com/bisq-network/bisq-core/pull/148


build.gradle
```diff
@@ -15,7 +16,7 @@ apply plugin: 'witness'
 apply plugin: 'com.github.johnrengelman.shadow'
 
 group = 'network.bisq'
-version = '0.7.1-SNAPSHOT'
+version = '0.8.0'
```
ripcurlx: Shouldn't we only do this during the release?

build.gradle
```diff
@@ -33,25 +34,29 @@ tasks.withType(JavaCompile) {
...
     maven { url 'https://dl.bintray.com/jerady/maven' }
 }
 
 dependencies {
-    compile 'network.bisq:bisq-p2p:-SNAPSHOT'
-    compile 'network.bisq:bisq-core:-SNAPSHOT'
-    compile 'network.bisq:bisq-common:-SNAPSHOT'
+    compile 'network.bisq:bisq-p2p:0.8.0'
+    compile 'network.bisq:bisq-core:0.8.0'
+    compile 'network.bisq:bisq-common:0.8.0'
```
ripcurlx: Same here: I think this should be -SNAPSHOT when merging

build.gradle
```diff
@@ -33,25 +34,29 @@ tasks.withType(JavaCompile) {
 sourceSets.main.resources.srcDirs += ['src/main/java'] // to copy fxml and css files
 
 repositories {
+    mavenLocal()
```
ripcurlx: Was this only for testing purposes locally?

package/osx/create_app.sh
```diff
@@ -82,18 +82,18 @@ $JAVA_HOME/bin/javapackager \
     -Bicon=package/osx/Bisq.icns \
     -Bruntime="$JAVA_HOME/jre" \
     -native dmg \
-    -name Bisq \
-    -title Bisq \
+    -name Bisq_notifications \
+    -title Bisq_notifications \
```
ripcurlx: Also shouldn't be changed here

ripcurlx (changes requested):
Please reset the version and naming used for the local test build. Besides that utACK

ripcurlx (approved):
utACK

PR #1629, opened 2018-08-16
Add rounded fiat value info popup

### LINKED ISSUE #1511: Option to have fiat values rounded to full euros, 0 cents
Proposal to change the precision of the price in the trade statistics
 
Setup of the fiat trade values. In order to increase the privacy of fiat2BTC trades all fiat amounts 
should be optional rounded to a full amount $xx.00, €xx.00. Background is that a decimal place .00 occurs 39 
times more often in all fiat transfers than the average .xy and even 390 times more often than bad 
choices like .93 or .74. If rounded by value = Math.round(value) it would be much more difficult to 
reveal the identity of a bisq trader on the fiat side by time/amount attacks while beeing no big loss 
for the traders. Because the history of all trade amounts+times are public and all unhacked bank data have a high 
chance to be hacked and sold in future it's wise to use only Euro amounts which have many clones in the total currency area 
at the trading time + 24h.

P.S: Important is that the fiat value transferred via bank account has 0 cents, not only the visible history in the bisq client.

EDIT: Clean up formatting and made title more clear

ripcurlx:
Fixes #1511.

Besides the 👁 icon that is added already, I've added an additional popup for create and take offer screens.

Dependent on: https://github.com/bisq-network/bisq-core/pull/153

bildschirmfoto 2018-08-16 um 11 25 37
bildschirmfoto 2018-08-16 um 11 25 29


ManfredKarrer (approved):
utACK

PR #1635, opened 2018-08-20
Mirror webcam image at QR code scan view

ManfredKarrer:
For better usability we mirror the scanned image from the webcam so that
users move the mobile phone to the direction to the webcam by intuition.

@ripcurlx If it is too late for that release no worry, then we keep it for next release...


ripcurlx (approved):
utACK

PR #1638, opened 2018-08-21
Fix broadcast issues

ManfredKarrer:
Depends on:
https://github.com/bisq-network/bisq-common/pull/37
https://github.com/bisq-network/bisq-core/pull/165


chirhonul (approved):
utACK 9a9d5745.

In commit 9a9d57452 we are removing all the SHA256 checksum verifications for the dependencies we have.

This seems less than ideal, and I understand after discussion with @ManfredKarrer that it's meant to be temporary until the current build issues can be addressed. If so, I'd suggest adding a comment to `build.gradle` to note this.

PR #1639, opened 2018-08-21
Fix tx broadcast issue

ManfredKarrer:
The issue was still not fixed with the latest changes as it has been shown in more tests. We reverted now completely the changes introduced in January 2018 as they did not improve the broadcast reliability but decreased it. It is still not understood why... 

Depends on:
https://github.com/bisq-network/bisq-common/pull/38
https://github.com/bisq-network/bisq-core/pull/166


ripcurlx (approved):
ACK - trading test on mainnet worked.

PR #1640, opened 2018-08-23
Use correct version number

ripcurlx:
Didn't have an effect on the release, but should be now 0.8.0-SNAPSHOT


ManfredKarrer (approved):
utACK

PR #1642, opened 2018-08-25
Verify all checksums in prepare-system.sh

chirhonul:
Also improve style in general.


ripcurlx (approved):
utACK - Thanks for adding the additional checks!

PR #1662, opened 2018-09-07
List Bitcoin 2 (BTC2)

Bitc2org:
Resubmission. Original one: https://github.com/bisq-network/bisq-assets/pull/105
- Already ACKed.

Official project URL: https://www.bitc2.org/
Official block explorer URL: https://www.bitc2.org/block-explorer/

Suggestion: In case of a dispute, the seller of BTC2 should tell the transaction ID (hash). He/she can obtain that easily from the block explorer or from the Bitcoin 2 Core wallet by clicking the Transactions Tab, then clicking the transaction and copy-pasting the transaction ID.

### DIFF
--- a/assets/src/main/java/bisq/asset/coins/Bitcoin2.java
+++ b/assets/src/main/java/bisq/asset/coins/Bitcoin2.java
@@ -1,0 +1,28 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.asset.coins;
+
+import bisq.asset.Base58BitcoinAddressValidator;
+import bisq.asset.Coin;
+
+public class Bitcoin2 extends Coin {
+
+    public Bitcoin2() {
+        super("Bitcoin 2", "BTC2", new Base58BitcoinAddressValidator());
+    }
+}
--- a/assets/src/main/resources/META-INF/services/bisq.asset.Asset
+++ b/assets/src/main/resources/META-INF/services/bisq.asset.Asset
@@ -7,6 +7,7 @@ bisq.asset.coins.Angelcoin
 bisq.asset.coins.Aquachain
 bisq.asset.coins.Arto
 bisq.asset.coins.BitCloud
+bisq.asset.coins.Bitcoin2
 bisq.asset.coins.BitcoinCash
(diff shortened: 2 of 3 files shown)


blabno (approved):
ACK

ManfredKarrer:
I really get sick with those name squatter coins. Why do you think you have the right to call your self "Bitcoin 2"? Suggest like you are the successor of Bitcoin which of course you are not.
Once the DAO is out I will request removal of that coin.

Bitc2org:
Bitcoin 2, just like many other Bitcoin forks rewards the holders of Bitcoin and most of its code is Based on that of Bitcoin. It was always going to be Bitcoin something, because of that.

Yes it is an ambitious project, that's why we chose an ambitious name. Bitcoin is not a trademark. When someone mentions Bitcoin 2, mainstream will have a clue what it could be. When someone mentions something like PIVX, most people will have no clue what it is even about, from just the name.

ManfredKarrer:
@BTC2CoreDevs Bad and poor arguments. But did not expect anything else...
Hope there will be a fork of Bitcoin 2 called Bitcoin 2* so you might learn from experience...

Names are common goods needed for communication. Just because Bitcoin does not use a trademark protection does not mean that anyone is welcome to abuse a well established and understood brand.

Anyway, I will take the chance to test out the DAO voting feature, so thanks ;-)

joachimneumann:
I don't understand why is this obvious scam managed to get merged into master. Is there anything I can do to keep Bisq clean from these people?

cbeams:
> I don't understand why is this obvious scam managed to get merged into master.

I'm not sure it's any more obvious than many other assets that get submitted for listing. Our stated policy is that we list any correct PR (where correctness is defined at https://bisq.network/list-asset).

As painful as this policy is, there isn't any other feasible option at this point. Either we (a) sit in judgment of each asset on a PR-by-PR basis, deciding what's a scam and what isn't, wasting our time with what is sure to be endless arguments (with asset listers), (b) we shut down listing altogether, or (c) we let them all in. For better or worse, we've gone with (c). As Manfred mentioned above, the DAO will have a mechanism to vote assets out, but we're not there yet. Most assets never get traded anyway, and we already have a mechanism to de-list those. Thanks for your concern, but I'm not sure we have any great options at the moment.

joachimneumann:
@cbeams thanks for the explanation, that makes total sense.

aejontargaryen:
all subject matter aside, I must say this is just another testament to Bisq's awesomeness-  this is amazing to see the asset still being merged regardless of a strong opinion otherwise, but to stay true to (current) method.  just cool

cbeams:
Thanks for saying so, @aejontargaryen.

Bitc2org:
Agreed that it is nice to see you truly sticking to the decentralized principles despite one with the power to merge disagreeing. We have every intention to promote our listing on Bisq and tell users how to trade BTC2 on Bisq once you release the new version with us listed. By all means if that fails to bring any trades to your platform, a de-listing would be understandable.

Since we can't influence the development of Bitcoin, we could only do our own version of it with a hard fork, which we did. We brought a lot more changes to a Bitcoin fork than any of the other Bitcoin forks that we know about that have Bitcoin in their name.

If someone makes Bitcoin2* or such, that's their right. There is already "Bitcoin2network" that was made after we made Bitcoin 2. It didn't really harm us just like we can't harm Bitcoin. We can only try to innovate alongside it, and be one of the projects to pick up where they left off if something bad happens to Bitcoin Core or Bitcoin.

ManfredKarrer:
Got removed following proposal https://github.com/bisq-network/proposals/issues/60

PR #1665, opened 2018-09-08
List Ryo (RYO)

Technohacker:
- Official project URL: https://ryo-currency.com/
- Official block explorer URL: http://explorer.ryo-currency.com/
- Arbitation: Ryo wallets support transaction keys that enable 3rd party audits in case of a dispute similar to Monero

This is a resubmission of [List Ryo (RYO) bisq-assets#64](https://github.com/bisq-network/bisq-assets/pull/64) and has received ACK.


Technohacker:
Force Push Note: Regex was changed to avoid invalid Base58 characters

blabno (approved):
ACK

PR #1673, opened 2018-09-11
Add installing Lombok to build instructions

cbeams:
Also, separate one-time IDEA configuration items from details of
importing the project.


ripcurlx:
ACK

PR #1674, opened 2018-09-11
Improve grammar of seed info

arunasurya:
I only updated the English version. Let me know if we should update the other languages too.


ripcurlx:
ACK

ripcurlx:
@arunasurya All translations are done from within [transifex](https://www.transifex.com/bisq/bisq-desktop/dashboard/), so they will be updated automatically with the English translation if there was no translation in the specific language available. Only the English seed is changed directly in the repository. Thanks for the fix!

PR #1675, opened 2018-09-11
added some spanish translations

grunch:


ripcurlx:
@grunch Thanks for the translation PR, but translations for Bisq are handled only over [transifex](http://transifex.com/bisq/bisq-desktop/). The translation files are synchronized before every release. For more informations and any further question please join the #transifex channel on Slack.

PR #1682, opened 2018-09-13
Check bisq.asset.Asset sort order during CI

cbeams:
This check had been set up for some time in the original
bisq-network/bisq-assets repository.


ripcurlx:
utACK

Issue #1652, opened 2018-09-06
Bisq doesn't validate XMR addresses


bitlamas:
I'm using Bisq on Linux (Debian), version 0.8.0.
When creating an "Altcoin account" and selecting Monero (XMR), I can write whatever in the field "Altcoin address". It doesn't validate anything. I can write an actual XMR address or I can write "banana" and it will go through, letting me start a trade on it.

ManfredKarrer:
We can offer a 100 BSQ bounty for any dev who implements an XMR address validation. Feel free to promote that in the Monero communities!

Gingeropolous:
@ManfredKarrer , please include, as part of your bounty criteria, that the validation includes the new subaddress types that start with an 8. I don't know if bisq uses integrated payment IDs, but subaddresses should be favored over integrated PIDs.

Technohacker:
Done :sweat_smile: 
Standard addresses and Subaddresses are accepted :D

ManfredKarrer:
@Gingeropolous Thanks for the information. Bisq does not use payment IDs, just regular addresses. In case of a dispute the XMR sender needs to provide the tx key to the arbitrator.

ManfredKarrer:
@Technohacker  That was fast! ;-)

@Gingeropolous Could you help to review the PR?

Technohacker:
@ManfredKarrer If you don't mind the meme, _"The fastest PR in the west"_ :sweat_smile:

Gingeropolous:
unfortunately im just a hacker groupy

ManfredKarrer:
@Technohacker  Haha ;-) 

Don't forget to file a compensation request end of the month (PR at https://github.com/bisq-network/compensation)

Technohacker:
@ManfredKarrer I'll have a look, thanks!
So should I do it as soon as the PR is merged?

ManfredKarrer:
Yes once its merged you can file the request. I will be not much online next days, so might take a bit for review and testing...

fireice-uk:
@Technohacker Well done :+1:

sneurlax:
Weak validation, doesn't ensure that addresses are actually valid, just that they're prefixed with a 4 or an 8 and the correct number of correct characters--not that the checksum validates or anything like that.

Does anybody know of a Java implementation of cryptonote address validation?

Technohacker:
[...]

PR #1683, opened 2018-09-13
Fix Monero address validation (#1652)

Technohacker:
Add a RegexAddressValidator and tests

### DIFF
--- a/assets/src/main/java/bisq/asset/CryptonoteAddressValidator.java
+++ b/assets/src/main/java/bisq/asset/CryptonoteAddressValidator.java
@@ -1,0 +1,55 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.asset;
+
+/**
+ * {@link AddressValidator} for Base58-encoded Cryptonote addresses.
+ *
+ * @author Chris Beams
+ * @since 0.7.0
+ */
+public class CryptonoteAddressValidator implements AddressValidator {
+
+    private final String prefix;
+    private final String subAddressPrefix;
+    private final String validCharactersRegex = "^[1-9A-HJ-NP-Za-km-z]+$";
+
+    public CryptonoteAddressValidator(String prefix, String subAddressPrefix) {
+        this.prefix = prefix;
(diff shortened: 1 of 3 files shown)


matt2134711 (changes requested):
[Monero addresses are encoded in base58](https://github.com/monero-project/monero/blob/master/src/common/base58.cpp), so using `[0-9A-Za-z]` would still let some bad characters through (`0`, `I`, `O`, and `l` are never in an address).

Also the last few bytes are a checksum so it would be nice if that was validated too.

Technohacker:
This might take me a while, but I'll try getting it done :)
If anyone gets it done, I'd highly suggest notifying all CryptoNote coins to follow suit as well

matt2134711:
We already have Bouncy Castle as a dependency, which provides all of the cryptographic primitives needed (just keccak-256 IIRC), but I'm not sure if Bitcoin's base58 decoding class can be used as it is slightly different.

I can have a go at implementing this today, if you don't mind.

Technohacker:
Sure, I'll see if I can get an implementation done as well :smile:

Technohacker:
I've made changes to avoid the invalid characters (and made it a utility class for other CryptoNotes). What remains is the checksum. CryptoNote doesn't do decoding like Bitcoin (in blocks of bytes rather than as a whole), and I'm not sure if adding the decoder only for CryptoNote addresses is warrantable. Of course on the other hand there's a risk of putting a wrong (but valid) address.

matt2134711:
The decoder shouldn't be too complicated, the JavaScript on [luigi1111's site](https://xmr.llcoins.net/js/base58.js) is a good reference.

ManfredKarrer:
@matt2134711 @Technohacker We are in the process to remove Bouncy Castle as security provider as with Java 10 they removed support for the extension mechanism (which was used to add the BC signed jar) and we have not found a way yet to include it in the fat jar for the binary otherwise. So we will use the default provider which have less features but covers the features we use atm. I doubt that keccak-256 will be supported there.

matt2134711:
@ManfredKarrer Would it be okay to depend on something like [keccakj](https://github.com/aelstad/keccakj) instead?

ManfredKarrer:
@matt2134711 Is there a jar on a public repo? If so, yes we add that dependency.

Technohacker:
I've (force-)pushed my changes for the new `CryptonoteAddressValidator`. Checksum verification is still pending

ManfredKarrer (approved):
ACK for first part of validation. 
Validation including the check sum will be delegated to a new PR.

assets/src/main/java/bisq/asset/CryptonoteAddressValidator.java
```diff
@@ -0,0 +1,55 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.asset;
+
+/**
+ * {@link AddressValidator} for Base58-encoded Cryptonote addresses.
+ *
+ * @author Chris Beams
```
Technohacker, line 23: Damn it, I forgot to change this. Could I post a follow-up PR to fix this?

ManfredKarrer: What did u forget? Yes feel free for another PR.

Technohacker: The author :sweat_smile:

Issue #1693, opened 2018-09-18
Add info for Faster Payment that the receivers name is not required


ManfredKarrer:
As discussed here we add info for Faster Payment that the receivers name is not required.

"Some banks might require the receivers name. The UK sort code and account number is sufficient for a Faster Payment transfer and the receivers name is not verified by any of the banks."

<img width="1196" alt="screen shot 2018-09-18 at 17 30 51" src="https://user-images.githubusercontent.com/1449498/45720171-d8050280-bb68-11e8-9e98-02ce4e5928ec.png">

ripcurlx:
Closing as complete.

PR #1694, opened 2018-09-18
Add info for Faster Payment that the receivers name is not required

ManfredKarrer:
See:
https://github.com/bisq-network/bisq/issues/1693


ripcurlx (approved):
utACK

cbeams:
The text doesn't really tell the user what to do in the case of the bank asking for first and last name. Should they write "N/A"? Should they make up a name? I think I saw someone advocating to make up a name on the forum the other day, and I think that's a pretty bad idea, as it could be seen as some kind of fraud. I would rather recommend that users repeat the sort code and account nummber in the first and last name field, or write N/A in both.

Also, "receivers" should be "receiver's".

ManfredKarrer:
@cbeams Yes suggestion a "made up name" is problematic. I assume most users find a solution on their own. If it is true that all banks require that as the Forum poster said, we can see that nobody had any problem with that so far as it was never reported (at least I am not aware of). 
Of course we could add more suggestion here, I have no strong opinion and not sure if it could cause more problems (e.g. using N/A the "/" might be not accepted in some bank forms or number are not accpeted in the name field,....). I will fix the typo in a next PR. 
If you have a suggestion or a better text, please add it here or as PR...

PR #1660, opened 2018-09-07
Java 10 support

ripcurlx:
I've started to adapt the current mono repository to support Java 10.
The current state of this branch is, that `gradle build` runs and you can launch the app from IDEA.
Still there are a couple of things left, where help is wanted (see check list)

- [x] Upgrade JavaFX code 
- [x] Implement alternative for progress indicator (block chain confirmations) as current implementation won't work for java 10
- [x] Update lombok
- [x] Update source compatibility settings
- [x] Update travis setting
- [x] Implement power mock support
- [x] Implement JMockit support
- [x] Merge with master (and update everything new to Java 10)
- [x] Solve issue with library that includes Java 9 module configuration

## Release Build
- [x] Java strong cyphers supported by default in Java 10
- [x] Configure or replace BouncyCastle provide ([Removed Extension Mechanism](https://docs.oracle.com/javase/10/migrate/toc.htm#JSMIG-GUID-2C896CA8-927C-4381-A737-B1D81D964B7B))

PR #1696, opened 2018-09-19
Fix gradle build files, update docs

ManfredKarrer:


ripcurlx (approved):
utACK

cbeams:
It's a little strange to see this merged before #1660, but assuming that happens soon, I guess it's no big deal. People may not actually be able to build against JDK 10 if they follow the updated instructions in the README. Hopefully that just doesn't happen in the meantime.

ManfredKarrer:
@cbeams #1660 is already merged

cbeams:
Hah, so it has. I was looking at a stale browser tab. 🤦‍♂️ Thanks.

Issue #1701, opened 2018-09-19
OSX App Nap mode causes Bisq to lose network connections


ManfredKarrer:
It seems that Apple changed the behavior of standby mode so that network connections get lost while in standby mode (even if only display is set to sleep).

This is pretty bad as if you have offers open those offers will get removed when your Bisq app is running but has no network access. I am not 100% sure how it is handled as it seems that only the P2P nodes get disconnected but not the Bitcoin nodes. So it might be some special handling for limiting the resources and not completely preventing network activity.

What is really annoying that there is zero information about those hidden behaviors (experienced similar issues with the iOS notification app). It is really a shame how Apple developed over the last years. Time to move to Linux....

Here is one of the few discussion about that issue: 
https://forums.macrumors.com/threads/mac-looses-connection-when-sleeping.1414285/

Here are the settings I use.
<img width="665" alt="screen shot 2018-09-19 at 14 39 59" src="https://user-images.githubusercontent.com/1449498/45776897-eb23db00-bc19-11e8-93a6-58d8d6f9686b.png">

Does anyone has an idea how to solve that?

Seems that app "Caffeinate" is build so solve that issue, testing it out now (costs 2 USD).

Maybe putting the display into lock screen by ctrl-shift-eject and leave the standby setting to never helps as well as one user posted in the link above:
"You can put just the display to sleep (ctrl-shift-eject does this). This will turn off the display and conserve power whilst still keeping the Mac turned on and thus the network connection(s)."

ManfredKarrer:
Seems both approaches (Caffeinate, ctrl-shift-eject) did not help. Also I observed that when I have Bisq open but not as active window network connections drop, at least when I bring it in foreground more offers get quickly displayed in the spread screen.

ManfredKarrer:
App Nap causes the issues:
https://www.lifewire.com/control-how-mac-applications-use-app-nap-2260788
https://apple.stackexchange.com/questions/121386/missing-prevent-app-nap-button-on-app

Run that command in the terminal and restart Bisq. It should never end up in App Nap mode with that and therefore not lose network connectivity when not visible (hidden by other windows).
```
defaults write io.bisq.CAT NSAppSleepDisabled -bool YES
```

It would be good if we find a way that App Nap is disabled by default for Bisq. Will search more for the java packager options.

ManfredKarrer:
[...]

PR #1712, opened 2018-09-21
Play silent sound to avoid standby mode

ManfredKarrer:
Apple disabled options do avoid App Nap in recent OSX versions.
Playing an inaudible sound marks the application for preventing to get
set to standby mode as well as App Nap mode.
The alternative to that "hack" would be to add OSX native code, but
even then it is likely only possible to prevent App Nap but not
sleep mode.
If Bisq is in App Nap mode offers cannot be taken as well the user loses
network connection and offers.
See: https://github.com/bisq-network/bisq/issues/1701


ripcurlx (approved):
utACK

Issue #1707, opened 2018-09-20
An offer maker's contact information is not easily seen in an F2F trade


arunasurya:
When taking an offer for an F2F trade, I was not able to see the offer maker's contact information easily. I could only see it in the "locked funds" window.

Suggestion: It would be great to have the offer maker's and taker's contact info displayed in the same place.

ManfredKarrer:
You mean the trade details popup which you open when clicking the trade ID? That can be clicked also in the pending trade screen.
The contact info is also displayed in the trade state screen.

<img width="1201" alt="screen shot 2018-09-20 at 22 42 55" src="https://user-images.githubusercontent.com/1449498/45859176-8ace8f80-bd26-11e8-9cee-58d17c4599a0.png">

arunasurya:
OK, thanks.

arunasurya:
As a taker, I don't see contact info as easily. Please see the screenshot below.

![contactinfoissuesfortakers](https://user-images.githubusercontent.com/43150241/45914811-d45bcf00-be39-11e8-9f76-2fba1c81548d.png)

ManfredKarrer:
Ah ok, right. That is a bit more effort as it changes behaviour in relation to other payments.

ManfredKarrer:
Implemented in PR https://github.com/bisq-network/bisq/pull/1713

<img width="1197" alt="screen shot 2018-09-22 at 18 56 50" src="https://user-images.githubusercontent.com/1449498/45922697-50d1ca80-be99-11e8-9e04-3e3c393b61cc.png">

PR #1713, opened 2018-09-22
Add contact info for seller in case of aa F2F trade

ManfredKarrer:
Fixes: https://github.com/bisq-network/bisq/issues/1707#issuecomment-423724283


ripcurlx (approved):
utACK

PR #1719, opened 2018-09-24
Refactoring of Btc package

ManfredKarrer:
This PR was part of the fix of the duplicated btc nodes. While investigating I saw that the btc package was very unclearly organized. i tried to rearrange and rename classes so it provides a better structure.

It is all pure technical refactoring except https://github.com/ManfredKarrer/bisq/commit/c4d45ceecab6b8add9ff84114cc64d06ed3b8fac where we removed LTC support for seed nodes.


ripcurlx (approved):
utACK

PR #1722, opened 2018-09-25
Fix bugs with voting

ManfredKarrer:


core/src/main/java/bisq/core/dao/state/governance/Param.java
```diff
@@ -74,9 +74,9 @@
     PHASE_BREAK1(1),        // 10 blocks
     PHASE_BLIND_VOTE(2),    // 4 days
     PHASE_BREAK2(1),        // 10 blocks
-    PHASE_VOTE_REVEAL(1),   // 2 days
+    PHASE_VOTE_REVEAL(2),   // 2 days
```
sqrrm, line 77: Do you think we should add a minimum allowed for these params? We don't want a vote to change a param into something that would break the DAO.

sqrrm (approved):
utACK

core/src/main/java/bisq/core/dao/state/governance/Param.java
ManfredKarrer: The real number will be much higher. In voting we still have human control of main stakeholders, but yes would be good to have some sanity checks...

sqrrm: My thinking is that it's good to add hard limits that can't be exceeded now since it makes no sense to allow breaking them, and we now have it clear in our minds.

PR #1730, opened 2018-09-27
Fix last request timestamp in price feed service

devinbileck:
I noticed the last update time shown in the market price popup was incorrect. It was updating/refreshing, but not accurate.
image

Upon debugging, I identified the timestamps were incorrectly being multiplied by 1000 in the price feed service. Removing this caused it to return the correct date, as shown below.
image


ManfredKarrer (approved):
Ups we were ahead of time ;-) in year 50710. A bit too far ;-)

ACK.

PR #1735, opened 2018-09-28
Fixes and improvements

ManfredKarrer:


sqrrm (approved):
utACK

I would prefer smaller more contextual PRs. That would make it easier to review.

ManfredKarrer:
@sqrrm Yes agree. Was started with a small DAO only PR, but then added up more. Though to have several parallel branches where I need to cross merge until the PRs are merge in master is also a bit cumbersome. Some of the changes I needed immediately in my running dev tests (like the btcd-cli4j fixes). But sure I agree that is not as it should be....

PR #1738, opened 2018-10-01
Dao fixes

ManfredKarrer:


core/src/main/java/bisq/core/dao/governance/proposal/ProposalService.java
```diff
@@ -243,14 +259,9 @@ private void onAppendOnlyDataAdded(PersistableNetworkPayload persistableNetworkP
...
                 if (proposalValidator.areDataFieldsValid(proposal)) {
-                    if (proposalValidator.isTxTypeValid(proposal)) {
-                        proposalPayloads.add(proposalPayload);
-                        log.info("We received a ProposalPayload and store it to our appendOnlyStoreList. proposalTxId={}",
-                                proposal.getTxId());
-                    } else {
-                        log.warn("We received a proposal with an invalid txId. Proposal.txId={}",
-                                proposal.getTxId());
-                    }
+                    proposalPayloads.add(proposalPayload);
```
sqrrm, line 262: Could this be a potential attack vector, spamming invalid proposals that persist indefinitely?

sqrrm (approved):
utACK

core/src/main/java/bisq/core/dao/governance/proposal/ProposalService.java
ManfredKarrer: Yes, I will look into it more if I find a way to verify the relevant tx.

PR #1739, opened 2018-10-01
List BTC/ODIN

Manbearpixel:
**Work In Progress**

### DIFF
--- a/assets/src/main/java/bisq/asset/coins/ODIN.java
+++ b/assets/src/main/java/bisq/asset/coins/ODIN.java
@@ -1,0 +1,38 @@
+/*
+ * This file is part of Bisq.
+ *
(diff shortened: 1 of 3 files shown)


blabno (changes requested):
NACK

Do not submit WIP. It looks like you were not even compiling it on your own machine but editing through github instead as leveraging Travis as a compiler (facepalm).

Manbearpixel:
@blabno Sorry mate, didn't know I couldn't create a WIP PR. I'll rework :)

We have a machine that has run Gitian builds for supported platforms and have released our GUI a few weeks ago. I had this marked as a WIP as we were just releasing the GUI and an Explorer. We've also added Travis-ci integration and will be working to improve some test runs.

PR #1741, opened 2018-10-02
List QMCoin

project-qmc:
- Official project URL: https://github.com/project-qmc/QMC
- Official block explorer URL: http://136.243.26.230/

### DIFF
--- a/assets/src/main/java/bisq/asset/coins/QMCoin.java
+++ b/assets/src/main/java/bisq/asset/coins/QMCoin.java
@@ -1,0 +1,56 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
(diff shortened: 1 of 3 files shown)


project-qmc:
Was trying to figure an rough ETA for possible listing but noticed no releases have been made recently (and also a note about reaching consensus regarding listing fees - which was settled as still no fee I think?) ... So my question is, how does it work now days?

blabno (approved):
ACK

project-qmc:
Thanks, I don't suppose there is any new release ETA, is there? seeing as the last one was in August and the one before that was in July?

I've got a feeling that a lot of coins get listed after a very long wait and therefor they end up without volume (as they seek alternatives) - plus it is possible their devs forget to check for a new release after a month. We'll make sure this doesn't happen with QMC - but even a rough ETA would be nice !

ghost:
Next release is _planned_ for mid-end october.

project-qmc:
Awesome, looking forward to it.
Thanks again!

project-qmc:
@HarryMacfinned Any idea for a new ETA? :)

ghost:
There are some issues with the last java version to use.
And since v0.9.0 will have many novelties, it will require strong testing.
We'll probably have to wait at least one more week I guess.

project-qmc:
Sounds good. Not much of a java guy myself but do know a bit of everything - so if you guys ever need a helping hand just ping ;)

project-qmc:
Soooo this isn't in the new release ?

ghost:
I guess new assets listing will be done with v0.9.0.
And there are delays.

ripcurlx:
@project-qmc 0.8.1 was just about temporarily fixing a startup issue. Release 0.9.0 is in its final development phase and it will need in-depth testing as we nearly touched every class and introduced lots of new stuff. So fingers crossed that we'll start testing soon.

PR #1751, opened 2018-10-05
Dao ui improvements

ManfredKarrer:


core/src/main/java/bisq/core/dao/governance/proposal/compensation/CompensationConsensus.java
```diff
@@ -18,18 +18,21 @@
...
 
 import org.bitcoinj.core.Coin;
 
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class CompensationConsensus extends ProposalConsensus {
-    public static Coin getMinCompensationRequestAmount() {
-        return Coin.valueOf(1_000); // 10 BSQ
+    public static Coin getMinCompensationRequestAmount(BsqStateService bsqStateService, int chainHeadHeight) {
```
sqrrm: Why call the argument chainHeadHeight instead of just chainHeight?

core/src/main/java/bisq/core/dao/governance/proposal/param/ChangeParamValidator.java
```diff
@@ -41,9 +47,109 @@ public void validateDataFields(Proposal proposal) throws ValidationException {
...
+                break;
+            case BTC_MAKER_FEE_IN_PERCENT:
+                break;
+            case BTC_TAKER_FEE_IN_PERCENT:
+                break;
+            case PROPOSAL_FEE:
+                break;
+            case BLIND_VOTE_FEE:
+                break;
+            case COMPENSATION_REQUEST_MIN_AMOUNT:
```
sqrrm, line 85: There should also be a check that `COMPENSATION_REQUEST_MIN_AMOUNT < COMPENSATION_REQUEST_MAX_AMOUNT`, both current and proposed.

core/src/main/java/bisq/core/dao/governance/proposal/param/ChangeParamValidator.java
```diff
@@ -41,9 +47,109 @@ public void validateDataFields(Proposal proposal) throws ValidationException {
...
+     * @param param         The param to change
+     * @param percentChange 100 means 100% more than current value -> 2 times current value. -50 means half of the current value
+     * @return The new value.
+     */
+    //TODO add test
+    // TODO use multiplier to make it more intuitive? (4,4) means 4 times current value for max and divided by 4 to get min value)
+    @VisibleForTesting
+    long getNewValueByPercentChange(Param param, long percentChange) {
+        checkArgument(percentChange > -100, "percentChange must be bigger than -100");
+        return (getCurrentValue(param) * 100 * (100 + percentChange)) / 100;
```
sqrrm: This multiplies by 100 one time too many.

core/src/main/java/bisq/core/util/BsqFormatter.java
```diff
@@ -114,4 +116,101 @@ public Coin parseSatoshiToBtc(String satoshi) {
...
+    public String formatParamValue(Param param, long value) {
+        switch (param) {
+            case UNDEFINED:
+                return Res.get("shared.na");
+
+            case BSQ_MAKER_FEE_IN_PERCENT:
+            case BSQ_TAKER_FEE_IN_PERCENT:
+            case BTC_MAKER_FEE_IN_PERCENT:
+            case BTC_TAKER_FEE_IN_PERCENT:
+                return formatToPercentWithSymbol(value / 10000d);
```
sqrrm, line 130: Why is the value divided by 10000 rather than 100 to get percent?

sqrrm (commented):
I would like some clarification on my comments on how calculations are done. Something seems strange about it, either it's correct and some code comments are needed or something isn't working as intended.

core/src/main/java/bisq/core/dao/governance/proposal/compensation/CompensationConsensus.java
ManfredKarrer: Yes should be chainHeight, was mixing it up a bit. will change...

core/src/main/java/bisq/core/util/BsqFormatter.java
ManfredKarrer: Because we want to support 2 decimal places (e.g. 12.34%) so its 100 * 100 = 10 000

core/src/main/java/bisq/core/dao/governance/proposal/param/ChangeParamValidator.java
ManfredKarrer: Yes the validations are very much WIP...

ManfredKarrer:
The validation code for params is WIP and I am not sure if the current implementation is intuitive (range set by % of current value...). I will keep that for a bit later and try to get other high prio stuff done first. But feel free to work on that if you want.

core/src/main/java/bisq/core/util/BsqFormatter.java
sqrrm: Ah, now I get it. Probably good to add a comment about that to make it clear

core/src/main/java/bisq/core/dao/node/full/network/FullNodeNetworkService.java
```diff
@@ -173,6 +189,24 @@ public void onFault(String errorMessage, @Nullable Connection connection) {
             } else {
                 log.warn("We have stopped already. We ignore that onMessage call.");
             }
+        } else if (networkEnvelope instanceof RepublishBlindVotesRequest) {
+            ObservableList<BlindVotePayload> appendOnlyStoreList = blindVoteListService.getAppendOnlyStoreList();
+            appendOnlyStoreList
+                    .forEach(blindVotePayload -> {
+                        // We want a random delay between 0.1 and 30 sec. depending on the number of items
```
sqrrm, line 201: What's the reason to depend on the number of items?

core/src/main/java/bisq/core/provider/fee/FeeService.java
```diff
@@ -51,42 +54,72 @@
...
+        return bsqStateService != null && periodService != null ? bsqStateService.getParamValue(parm, periodService.getChainHeight()) : 0;
+    }
 
-    // 0.05 BSQ (5 satoshi) for a 1 BTC trade -> 0.005%. 0.05 USD if 1 BSQ = 1 USD, 10 % of BTC fee
-    private static final long MIN_MAKER_FEE_BSQ = 5;
-    private static final long MIN_TAKER_FEE_BSQ = 5;
-    // 2 BSQ or 200 BSQ-satoshi. About 2 USD if 1 BSQ = 1 USD for a 1 BTC trade which is about 10% of a normal BTC fee.
-    private static final long DEFAULT_TAKER_FEE_BSQ = 200;
-    private static final long DEFAULT_MAKER_FEE_BSQ = 200;
+    public static Coin getMakerFeePerBtc(boolean currencyForFeeIsBtc) {
```
sqrrm, line 73: This is a bit confusing. Probably better to have separate methods for the two coins. If necessary to have just one method it should be clear what the Coin returned might be.

sqrrm (approved):
utACK

Some comments on parts that might need more documentation but please go ahead and merge.

core/src/main/java/bisq/core/dao/node/full/network/FullNodeNetworkService.java
ManfredKarrer: We don't want to stress the network with many broadcasts, so if there are many objects it will be delayed, the more objects the longer the dealy, but not more then 30 sec.

core/src/main/java/bisq/core/provider/fee/FeeService.java
ManfredKarrer: Hm.. the handling of the BTC or BSQ fee is not very nice atm... would require a bit more refactoring at the clients side... Atm I don't want to touch more as it is needed so prefer to keep it but yes, its an open TODO...

PR #1669, opened 2018-09-09
List Bitmark

bitmark-protocol:
If you need extra parts to the commit (like test cases), let me know. I just copied and pasted what I saw for Namecoin, which is a similar coin (both based on Bitcoin).

The version of Bitmark I want to use may different than what will soon be listed on other exchanges. The "official team" wants to add a controversial fork that is obviously a scam and departs from the honest protocol that Bitmark maintained throughout the years. You can se the discussion here: https://bitcointalk.org/index.php?topic=3169983.msg43299842#msg43299842

If you want call it "Bitmark Classic", but the ticker name BTM shouldn't conflict because the team of investors controlling the alternate repo (project-bitmark) want the ticker name to be MARKS.

The repo I want to be used is the one created by this account (bitmark-protocol/bitmark). The explorer: https://explorer.bitmark.cc

PR #1755, opened 2018-10-05
List Bitmark (BTM)

bitmark-protocol:
I am resubmitting a pull request for Bitmark. Previously it was submitted with the master branch, now I am using "list-bitmark" as the branch.

Last submission: https://github.com/bisq-network/bisq/pull/1669
Block explorer: https://explorer.bitmark.cc

### DIFF
--- a/assets/src/main/java/bisq/asset/coins/Bitmark.java
+++ b/assets/src/main/java/bisq/asset/coins/Bitmark.java
@@ -1,0 +1,28 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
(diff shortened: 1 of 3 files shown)


assets/src/main/java/bisq/asset/coins/Bitmark.java
```diff
@@ -0,0 +1,28 @@
...
+
+package bisq.asset.coins;
+
+import bisq.asset.Coin;
+import bisq.asset.DefaultAddressValidator;
+
+public class Bitmark extends Coin {
+
+    public Bitmark() {
+        super("Bitmark", "BTM", new DefaultAddressValidator());
```
blabno, line 26: This validator is deprecated:
>it exists for legacy purposes and is deprecated to indicate that new Asset implementations should NOT use it

blabno (changes requested):
NACK per comments

ManfredKarrer:
Closed as not updated after 2 weeks. Open again after it is fixed.

PR #1764, opened 2018-10-08
Deactivate OKPay as it does not exist anymore

ManfredKarrer:
OkPay seems to have rebranded to Weezzo.
The OkPay webpage is offline but some links go to Weezzo.
Seems pretty unprofessional hwo they did the move...
Seems they have been involved with BTC-e…
https://bitcointalk.to/index.php?topic=2094721.0


sqrrm (approved):
utACK

Agree, best to remove, not much info on this.

PR #1765, opened 2018-10-08
Show cancel button in wallet password prompt

### LINKED ISSUE #1181: Login password dialog doesn't provide 'exit' button
Please add an option to exit Bisq from the login password dialog.

devinbileck:
Fixes https://github.com/bisq-network/bisq/issues/1181

When launching the application, there was no way to cancel the wallet password prompt and quit the application. Attempting to close the application window did not quit the application (at least not in Windows 10).

So a cancel button has been added:
image

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/MainViewModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/MainViewModel.java
@@ -17,6 +17,7 @@
 
 package bisq.desktop.main;
 
+import bisq.desktop.app.BisqApp;
 import bisq.desktop.common.model.ViewModel;
 import bisq.desktop.components.BalanceWithConfirmationTextField;
 import bisq.desktop.components.TxIdTextField;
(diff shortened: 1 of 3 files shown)


ManfredKarrer (changes requested):
NACK
I agree that it is good to add a cancel button here but it must be done differently.

The WalletPasswordWindow is used in several places and the close button must not trigger a shitdonw in the other use cases. 
The hideCloseButton call can be removed in MainViewModel at line294.
To add the shut down functionality need to be added there as well so it does not affect the other use cases.

Following code should do it:
```
 bisqSetup.setRequestWalletPasswordHandler(aesKeyHandler -> walletPasswordWindow
                .onAesKey(aesKeyHandler::accept)
                .onClose(()-> BisqApp.getShutDownHandler().run())
                .show());
```

devinbileck:
Thanks @ManfredKarrer. I completely missed that. I have made the suggested changes.

ManfredKarrer (approved):
ACK.
Looks good now. Thanks!

PR #1767, opened 2018-10-11
Workaround for JDK-8199592; text being truncated at certain display scaling levels

devinbileck:
Using display scaling (e.g. 150% in Windows 10), text content within popups were being truncated.
A couple examples:
image

image

This is a result of a known Java bug (https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8199592) which mentions a workaround:
> explicitly calling sizeToScene appears to suppress the bug at DPI levels other than 175%

I confirmed that this workaround fixes the truncation at 125% and 150% scaling, but noticed it still truncates the checkbox label at 175% (as stated in the workaround):
image

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/overlays/Overlay.java
+++ b/desktop/src/main/java/bisq/desktop/main/overlays/Overlay.java
@@ -488,6 +488,7 @@ public abstract class Overlay<T extends Overlay> {
                     event.consume();
                     doClose();
                 });
+                stage.sizeToScene();
                 stage.show();
 
                 layout();


ManfredKarrer:
Ah cool! The checkbox issue might be because the method is called after the checkbox has been rendered. can u try to apply that as early as possible and check with breakpoints if the checkbox gets created before or ater that?

ManfredKarrer:
Ah just saw in the code that this is not possible. The  display(); is claled after the  addContent(); where the checkbox gets created. But still wondering why it then applies on the other text. Maybe its because we use AutoTooltipCheckBox? @ripcurlx Do you have any idea?

devinbileck:
There may not be anything we can do. As stated in the bug report, the workaround doesn’t work at all scaling levels. Or at least not at 175%.

ManfredKarrer:
I will leave merge to @ripcurlx, maybe he wants to check out why the checkbox is not working. From my side we can merge that as it is and postpone further improvement for later...

ripcurlx:
Hi @devinbileck! I'm just back from vacation and will have a look at your PRs today. Thanks for your investigation into this issue so far!

ripcurlx:
@devinbileck Unfortunately I somehow wasn't able to reproduce the bug on my Window 10 VM, but that could also be an issue with my local setup. Still I'll merge the PR as the changed code shouldn't do any harm, besides fixing this display scaling issue.

Issue #354, opened 2016-01-29
Arbitration selection>languages insertion on <cursor down>


esques:
Right now in the Arbitration selection>languages combo box 
if you press cursor down
you automatically get "Albanian" as a language. 
I suggest this to classify as an bug, independent from #112 as it is rather confusing and remove keyboard events until #112 is implemented

ManfredKarrer:
Ah ok, thanks. It the arrow down, right whcih causes selection.

esques:
exactly

ripcurlx:
Closing this issue as there is no manual arbitrator selection available anymore in 0.9+.

PR #1771, opened 2018-10-13
Fix arbitrator language selection

devinbileck:
- Fix arbitrator language selection where using arrow keys to navigate the combobox would add the first language. (https://github.com/bisq-network/bisq/issues/354)
- Fix arbitrator language selection prompt text no longer appearing after selecting a language.


devinbileck:
I just realized the same issue exists with the comboboxes for adding currencies and altcoins in the preferences view. Not sure yet if there are others. I will enter a separate PR to address those.

ManfredKarrer:
I assume same issue will happen at comboboxes as they use internally a popup (new stage). Will require probably to overwrite base classes...

devinbileck:
It only affects comboboxes that intentionally clear their value once you make a selection. So far, I believe it only affects just these 3 comboboxes mentioned.

ripcurlx:
ACK - looking good. Thanks for the fix @devinbileck! It is kind of tricky with prompt text and comboboxes. I also ran into some issues during the re-design. But your workaround helps to fix this issue at least.

PR #1774, opened 2018-10-14
Fix preferences view combobox selection

devinbileck:
Relates to https://github.com/bisq-network/bisq/pull/1771.

Regarding the ComboBoxes for adding currencies and altcoins in the preferences view:
- Fix for using arrow keys to navigate the ComboBox adding the first option.
- Fix prompt text no longer appearing after selecting an option.
- Fix option not sorted when added back to the ComboBox.


ripcurlx:
ACK - Thanks for the fix @devinbileck!

Issue #1742, opened 2018-10-02
Take offer screen layout issue


ManfredKarrer:
As visible in screenshot the "Check if offer is available..." text is not layouted correctly and the spinner is missing. That happens only if there is no amount range, so the next button get skipped and we show directly the funding screen.

<img width="1195" alt="screen shot 2018-10-01 at 21 55 16" src="https://user-images.githubusercontent.com/1449498/46326921-b3774480-c5c4-11e8-9d8e-98a2b09670a2.png">

When the BTC amount can be set by the user then the layout is correct and the spinner is there.

<img width="960" alt="screen shot 2018-10-01 at 21 57 05" src="https://user-images.githubusercontent.com/1449498/46326963-f507ef80-c5c4-11e8-9a02-9c36f06c2185.png">
 

Not sure where the text and spinner should be in the first case. At the current position it is easy to overlook and it is not clear why it is there.

devinbileck:
@ManfredKarrer @ripcurlx As far as I can tell, the "Check if offer is available" text is only applicable if there is an amount range for which the taker needs to specify how much they want to buy. If so, I believe I have a fix to just hide it.

devinbileck:
Upon further checking, looks like it is already supposed to be hiding it but isn’t. I will keep digging.

devinbileck:
I have opened PR https://github.com/bisq-network/bisq/pull/1775.

It was explicitly hiding the busy animation, but not the label. So my fix also hides the label.

However, I saw isOfferAvailableSubscription is supposed to be hiding it if the offer is available. And after debugging, I saw the offer state was unknown when taking any offer without a range. So that explains why the label was not being hidden. But I am a starting to dive deeper and getting a little lost trying to understand if that is to be expected (unknown state). Plus it is late and time for bed. Any clarification would be appreciated.

ManfredKarrer:
It is a bit tricky.
If there is an offer with range we show the edit amount screen with a next button and then the animation and label text for checking if the peer is online is all fine as the funding screen is not visible before the user clicks the next button and that is only possible after the offer availibility check has succeeded (the label and spinner animation is hidden).
[...]

PR #1775, opened 2018-10-15
Hide text that is not applicable in take offer view

devinbileck:
When taking an offer that does not specify an amount range, "Check if offer is available" text was being displayed when it shouldn't be.

It was explicitly hiding the offerAvailabilityBusyAnimation, but not the offerAvailabilityLabel.

Fixes https://github.com/bisq-network/bisq/issues/1742

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferView.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferView.java
@@ -418,6 +418,7 @@ public class TakeOfferView extends ActivatableViewAndModel<AnchorPane, TakeOffer
         cancelButton1.setVisible(false);
         cancelButton1.setManaged(false);
         cancelButton1.setOnAction(null);
+        offerAvailabilityBusyAnimation.stop();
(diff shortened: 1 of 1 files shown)


ManfredKarrer (changes requested):
NACK.
I added more background info in #1742. We must not hide is but change the layout so that it is better visible to the user and update the button states so that the user does not fund his wallet before the offer availibility check is successfully completed.

devinbileck:
@ManfredKarrer A fixed amount offer now waits until the offer has been confirmed to be available, as shown below:
image

Once confirmed, it will show the fund trade fields:
image

ManfredKarrer:
A cool! Looks good to me! @ripcurlx - What do you think? I leave it to you to review/merge.

ManfredKarrer (approved):
utACK - I would prefer if @ripcurlx does the review/merge

ripcurlx (approved):
ACK - tested it on Regtest with a range and a single amount offer. Everything looking good!

ripcurlx:
I'll merge it although Travis fails as it is related to the current Java 10 JDK not loaded issue.

PR #1776, opened 2018-10-16
[WIP] Bonded reputation

sqrrm:
Planning to merge the role and reputation service, a lot of copy paste there.

The trade protocol needs to be figured but I think what was discussed about a hashed salt should be enough together with tracking own bonded reputation and that of counter parties since those will be the only ones with known salt. A reputation bond without known salt is useless so no need to track those.


ManfredKarrer:
Close because it is merged into that PR: https://github.com/bisq-network/bisq/pull/1851

PR #1778, opened 2018-10-17
Persitence of DAO state 

ManfredKarrer:


common/src/main/proto/pb.proto
```diff
@@ -916,7 +916,7 @@ message PersistableEnvelope {
         UserPayload user_payload = 10;
         PaymentAccountList payment_account_list = 11;
 
-        BsqState bsq_state = 12;
+        // BsqState bsq_state = 12; // not used but as other non-dao data have a higher index number we leave it to make clear that we cannot change following indexes
```
sqrrm, line 920: Just comment with `deprecated` or some such to make it clearer I think.

sqrrm (approved):
utACK

common/src/main/proto/pb.proto
ManfredKarrer: If its commented out PB will not generate the code for it, so I think thats better, but sure "deprecated" will be good as well.

Issue #1663, opened 2018-09-08
Warn user when closing bisq with open offers.


headblock:
Some users may not know that bisq must remain open for offers to be seen by others.
Give popup warning if user tries to close bisq with open offers.

ManfredKarrer:
Good idea! @ripcurlx We could add that with a "dont show again" for the next release.

ManfredKarrer:
Implemented with https://github.com/bisq-network/bisq/pull/1716/commits/f0471bccfd02e7349d4a8adecf6dd83505e66102

See:
<img width="1199" alt="screen shot 2018-09-23 at 13 43 16" src="https://user-images.githubusercontent.com/1449498/45931695-25052200-bf37-11e8-89c8-ea76bd5d74dd.png">

Here is the text:
```
popup.info.shutDownWithOpenOffers=You try to shut down Bisq with open offers. \n\n\
  When you shut down Bisq your offers are not available anymore in the P2P network. \
  The next time you start up Bisq again your offers will get re-published to the P2P network.\n\n\
  If you want to keep your offers online you need to leave Bisq running. \
  Be sure that your computer does not switch to standby mode when not active. \
  Standby mode of the monitor is not a problem.
```
@m52go 
Any improvement of the text is welcome.

PR #1781, opened 2018-10-17
Refine shutdown message to warn about open offers

m52go:
Per #1663.


ManfredKarrer (approved):
utACK

Thanks!

m52go:
> Double quotes are only needed when a placeholder ({0}) is used in the text.

Ah -- alright, will remember that next time.

PR #1794, opened 2018-10-19
Use OpenJDK10

devinbileck:
Oracle JDK10 is EOL and is no longer available for download. So switch to OpenJDK10 in Gradle builds.


ManfredKarrer:
Great thanks for working on that. I will wait for more review from @cbeams and @ripcurlx ...

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: java
-jdk: oraclejdk10
+jdk: openjdk10
```
cbeams, line 2: Glad to see this works without the `install-jdk.sh` bit.

common/build.gradle
```diff
@@ -1,12 +1,27 @@
...
+    }
+    dependencies {
+        classpath 'com.google.gradle:osdetector-gradle-plugin:1.6.0'
+    }
+}
+
 plugins {
     id 'java'
     id 'maven'
     id 'com.google.protobuf' version '0.8.5'
```
cbeams, line 16: Add the following line here, and delete the `buildscript` block above:

    id 'com.google.osdetector' version '1.6.0'

(Per https://github.com/google/osdetector-gradle-plugin#for-gradle-21-and-higher)

common/build.gradle
```diff
@@ -1,12 +1,27 @@
...
 }
 
+apply plugin: 'application'
```
cbeams: Why?

common/build.gradle
```diff
@@ -1,12 +1,27 @@
...
+apply plugin: 'com.google.osdetector'
```
cbeams, line 19: Remove along with the `buildscript` block.

common/build.gradle
```diff
@@ -1,12 +1,27 @@
...
+ext.platform = osdetector.os == 'osx' ? 'mac' : osdetector.os == 'windows' ? 'win' : osdetector.os
+
 group = 'network.bisq'
 version = '-SNAPSHOT'
 
+mainClassName = 'bisq.common'
```
cbeams: This isn't the name of a class. Assuming you remove the `application` plugin as suggested above, this should be removed too.

common/build.gradle
```diff
@@ -19,6 +34,7 @@ tasks.withType(JavaCompile) {
 
 repositories {
     jcenter()
+    mavenCentral()
```
cbeams: Everything resolvable from Maven Central should also be resolvable from JCenter. Did that not prove true for you? Please remove `mavenCentral()` references here and everywhere else if possible.

cbeams (changes requested):
NACK per inline comments; all fairly minor stuff. Thanks for putting this together, @devinbileck.

common/build.gradle
devinbileck: I had to do it this way because I ran into the issue mentioned here: https://github.com/google/osdetector-gradle-plugin/issues/15

common/build.gradle
devinbileck: Whoops. That shouldn't have been added.

common/build.gradle
cbeams: Ok, that's a shame. Would be worth a comment about this in the `buildscript` block; otherwise it just looks like it got added due to ignorance of how `plugins {}` works.

cbeams: And you should be able to use `jcenter()` in the buildscript block vs `mavenCentral()`.

cbeams (approved):

cbeams:
LGTM from my side now, and I've approved the changes.

ripcurlx:
Thanks for the PR @devinbileck! I’ll test the client and release build on Monday if we need to do some additional changes because of the jar with modules config, but I guess our current workaround might fix this problem already.

cyphersphinx86:
Nice footwork @devinbileck
I'm still a little puzzled on how the refered javafx version 11 seamlessly work with jdk version 10, but cool!!!

devinbileck:
> Nice footwork @devinbileck

Thanks! :)

> I'm still a little puzzled on how the refered javafx version 11 seamlessly work with jdk version 10, but cool!!!

As per https://github.com/javafxports/openjdk-jfx/blob/jfx-11/doc-files/release-notes-11.md#release-notes-for-javafx-11:
> JavaFX 11 requires either JDK 10 (which must be an OpenJDK build) or JDK 11.

ripcurlx:
ACK - I tested a release build and everything seems to work. I pushed the verification checksum of the new libraries to this branch. As GitHub seems to be having issues right now I'll wait until [my commit](https://github.com/devinbileck/bisq/commit/c231187daa5a828198268836a2ee68163b5d1d56) will also show up here in the PR. @devinbileck Just one question: Why was it necessary to add the module-path to the configuration as we don't load libraries as java 9 modules within the app?

ripcurlx (approved):

devinbileck:
> Why was it necessary to add the module-path to the configuration as we don't load libraries as java 9 modules within the app?

@ripcurlx Good point, that may not be necessary. I was following the example from https://openjfx.io/openjfx-docs/#gradle and left that in. Shall I remove it?

ManfredKarrer:
@devinbileck Just tested the changes with installed openjdk. All works fine! Thanks a lot for your help!!!!

ripcurlx:
@devinbileck I'll do some checks tomorrow also try to get the checksum verification working again. Thanks for all your work!

PR #1801, opened 2018-10-22
Remove openjfx checksum until implementation for platform specific libraries is finished

ripcurlx:
I needed to remove the checksum check atm until I have properly implemented platform dependent checksums. Just works on Mac with that configuration.


ManfredKarrer (approved):
utACK as discussed on Slack

PR #1802, opened 2018-10-22
Move javafx dependencies to core

ManfredKarrer:
We use some javafx (Properties,...) classes in core as well, so we need
to have the dependency resolved there already.


ripcurlx:
utACK

PR #1803, opened 2018-10-22
Move javafx dependencies to common

ManfredKarrer:
For the upcoming API work we use java fx classes also in common (PR
is pending), so that is one reason to move that. But there was also an
issue on Linux (travis) as the platform property was not set in the
core gradle build file, but it is already defined in common.

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -46,6 +46,12 @@ protobuf {
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


ManfredKarrer:
I will merge myself without review as currently travis build fails...

ripcurlx:
utACK

PR #1805, opened 2018-10-23
Remove java 9 module configuration

ripcurlx:
Remove java 9 module configuration for openjfx libraries as they are used regularily during compile time.

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -85,12 +85,3 @@ dependencies {
     compile "org.openjfx:javafx-graphics:11:$platform"
     compile "org.openjfx:javafx-controls:11:$platform"
 }
-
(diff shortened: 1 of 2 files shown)


ripcurlx:
@devinbileck I've also ran it with OpenJDK now locally and everything looks fine!
@ManfredKarrer The only thing I need to test is a trade, but atm my local Regtest setup is not working anymore as the client is not able to connect to the local Bitcoin Core node over Regtest.

devinbileck (approved):
ACK

ManfredKarrer:
@ripcurlx Re regtest: just delete all data dirs also from bitcoin core regtest (exclude bitcoin.conf from delete).

Issue #1791, opened 2018-10-19
Not running on Ubuntu 18.04 LTS


5andr0:
I was trying to run Bisq on a remote 18.04 LTS server instance with xrdp (xfce4)

same error for v0.8 and v0.7 with java-8-openjdk-amd64 and java-8-oracle (log attached below)

Then i tried to compile from master with java-10.0.2-openjdk-amd64 and java-11-openjdk-amd64, but ran into javafx issues:
```
> Task :common:compileJava
/root/bisq-desktop/common/src/main/java/bisq/common/util/Utilities.java:39: error: package javafx.scene.input does not exist
import javafx.scene.input.Clipboard;
.
.
/root/bisq-desktop/common/src/main/java/bisq/common/util/Utilities.java:423: error: cannot find symbol
    public static boolean isAltOrCtrlPressed(KeyCode keyCode, KeyEvent keyEvent) {
                                             ^
  symbol:   class KeyCode
  location: class Utilities
```

This type of exchange is the future (great work) but it totally lacks a serverside/gui-less daemon, especially if it requires you to be online all the time! You should fund or join forces with this guy: https://github.com/citkane/bisq-front / https://github.com/citkane/bisq-engine

```
java -jar bisq-desktop-0.8.0-all.jar
Successfully removed cryptography restrictions
Oct-19 16:07:50.135 [main] INFO  b.c.s.CoreSetup: 


Log files under: /root/.local/share/Bisq/bisq 
Oct-19 16:07:50.141 [main] INFO  b.c.u.Utilities: System info: os.name=Linux; os.version=4.15.0-1021-gcp; os.arch=amd64; sun.arch.data.model=64; JRE=1.8.0_191-b12 (Oracle Corporation); JVM=25.191-b12 (Java HotSpot(TM) 64-Bit Server VM) 
Oct-19 16:07:50.801 [main] INFO  b.c.app.Version: Version{VERSION=0.8.0, P2P_NETWORK_VERSION=1, LOCAL_DB_VERSION=1, TRADE_PROTOCOL_VERSION=1, BASE_CURRENCY_NETWORK=0, getP2PNetworkId()=10} 
Oct-19 16:07:50.809 [main] INFO  b.c.s.CoreSetup: Path to Bisq jar file: /root/bisq-desktop/build/libs/bisq-desktop-0.8.0-all.jar 
[...]

PR #1807, opened 2018-10-23
Update install scripts

ManfredKarrer:
Add not tested script based on what @devinbileck posted in
https://github.com/bisq-network/bisq/issues/1791


devinbileck:
That was a rough draft and made a few assumptions (e.g. curl, git already installed). I can refine it and then test it :)

ManfredKarrer:
The old script was completely out dated, so even if there are issues its much better ;-). But sure we can update if its better tested.

ripcurlx:
utACK - looks good to me and it is more up-to-date than the BC/Java 8 setup we had in the old script.

ripcurlx (approved):

Issue #1769, opened 2018-10-12
Verification for DAO parameter default values


ManfredKarrer:
As mentioned in https://github.com/bisq-network/proposals/issues/47 we need verification of DAO parameter values which are different on parameter type.
The class `bisq.core.dao.governance.proposal.param.ChangeParamValidator` is doing already a very basic validtion and need to be extended for all params. The current solution for a resrtiction of a max. change is not optimal and a more intuitive solution would be preferred. Double values (specially division) should be avoided to not risk consensus failured due rounding issues on different OS.

As the open PR is not merged yet you need to fork from https://github.com/ManfredKarrer/bisq/tree/Dao-UI-improvements. After the PR is merged please redo your work on a new fork from Bisq repo to make later merge of the PR easier.

ManfredKarrer:
@Cypherspace86 will work on that task

ManfredKarrer:
Close as I have implemented it already

PR #1820, opened 2018-10-26
Verifi dao param default vals #1769

cyphersphinx86:
Surrounding code for DAO param validation

New, more intuitive methods for testing min and max values of proposed DAO parameters.
Unit tests also provided.

Next up will be a discussion of what limits each parameter should have.

First part of issue: #1769


cyphersphinx86:
a little too fast...

PR #1824, opened 2018-10-27
Verify dao param default vals #1769

cyphersphinx86:
Basis code for DAO param validation
    
    New, more intuitive methods for testing min and max values of proposed DAO parameters.
    Futhermore default limits set for all current DAO parameter tyeps. (Suggestions for other default values are very welcome)
    Unit tests also provided.
    
    Solves issue: #1769


ManfredKarrer (changes requested):
I could not add the comment to the code directly...

I would prefer to not have class fieds in the test as that makes the test more complex. Would prefer also to keep the formatter out completely as its an injected object. The test cannot be run atm in the IDEA without extra setup (jvm args) as the mockit framework has issues with the recent java 10 update.... I know we have other tests as well where that is the case but I prefer that new tests are as functional style as possible (e.g. simple input simple output). The formatting can be done in the caller and that does not need to be tested.

In the test there are some unused fields, please remove those.

ManfredKarrer:
As discussed on Slack there have been bigger changes of the parameter handling so that is outdated.

PR #1827, opened 2018-10-27
Add fiat amount to fee

ManfredKarrer:
Please see the screenshot in the slack #design channel. Not 100% sure if that solution is really good...

### DIFF
--- a/common/src/main/proto/pb.proto
+++ b/common/src/main/proto/pb.proto
@@ -722,6 +722,7 @@ message PaymentAccountPayload {
         WeChatPayAccountPayload we_chat_pay_account_payload = 22;
         MoneyGramAccountPayload money_gram_account_payload = 23;
         HalCashAccountPayload hal_cash_account_payload = 24;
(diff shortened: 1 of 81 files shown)


ripcurlx:
I'm not so sure if it is a good idea to put the values into an InputField style box if you can't change them. What about putting it together with the security deposit in an advanced options section that can be hidden/shown based on a preferences flag?
trade fees
This on the other hand makes it still confusing for the first time user. Maybe we could add another step where you have to select the trade fee currency once and your setting is stored. The only way to change it is by activating advanced options in the preferences and/or being able to set the trade fee currency in the preferences as well.

ManfredKarrer:
Great thanks, yes much better. But also agree that it still is confusing. We still have time as it will be only relevant for the DAO mainnet launch. Before that there would be only displayed the fee in BTC and Fiat and %. 
As you observed that PR was already part in another merged PR (was not intended) so I will close that. Feel free to add your changes to your follow up PRs!

ManfredKarrer:
Close as it was already merged as part of another PR

PR #1829, opened 2018-10-28
Improve grammar of Background Info

arunasurya:


core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -792,21 +792,19 @@ support.buyerOfferer=BTC buyer/Maker
...
-support.backgroundInfo=Bisq is not a company and not operating any kind of customer support.\n\n\
+support.backgroundInfo=Bisq is not a company and does not provide any kind of customer support.\n\n\
 If there are disputes in the trade process (e.g. one trader does not follow the trade protocol) \
-the application will display a \"Open dispute\" button after the trade period is over \
+the application will display an \"Open dispute\" button after the trade period is over \
 for contacting the arbitrator.\n\
-In cases of software bugs or other problems, which are detected by the application there will \
-be displayed a \"Open support ticket\" button to contact the arbitrator who will forward the issue \
+If there is an issue with the application, the software will detect this and display \
+an \"Open support ticket\" button to contact the arbitrator who will forward the issue \
```
ManfredKarrer, line 800: Hm... that is not really the case. Not all bugs are detected, only if the app detects an issue we show the button.

ManfredKarrer (changes requested):

core/src/main/resources/i18n/displayStrings.properties
arunasurya: I tried to preserve the meaning of the old text and make it clearer. The old text said:
> which are detected by the application

which implies that all bugs are detected. 

I realize that it is not always possible to detect bugs in the software. I would be happy to edit the text to make this more clear. How about:
> If there is an issue with the application, the software will try to detect this and, if possible, display...

Let me know if you have another suggestion to edit this or if you would be OK to merge this rewording that I believe preserves the meaning.

ManfredKarrer: Yes that was probably cunfusion due my bad english... 
Yes your suggestion sounds good. Please push tha change to that PR.

arunasurya:
Please take another look. I included the changes as discussed.

ManfredKarrer (approved):
utACK

Great thanks!

cbeams:
Really appreciate fixes like this. Thanks, @arunasurya.

PR #1831, opened 2018-10-28
Add PromptPay (Thailand)

leshik:
Closes https://github.com/bisq-network/payment-methods/issues/6.

### DIFF
--- a/core/src/main/java/bisq/core/payment/payload/PaymentMethod.java
+++ b/core/src/main/java/bisq/core/payment/payload/PaymentMethod.java
@@ -81,6 +81,7 @@ public final class PaymentMethod implements PersistablePayload, Comparable {
     public static final String HAL_CASH_ID = "HAL_CASH";
     public static final String F2F_ID = "F2F";
     public static final String BLOCK_CHAINS_ID = "BLOCK_CHAINS";
+    public static final String PROMPT_PAY_ID = "PROMPT_PAY";
 
     @Deprecated
     public static PaymentMethod OK_PAY;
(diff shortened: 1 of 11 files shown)


core/src/main/java/bisq/core/payment/payload/PaymentMethod.java
```diff
@@ -227,6 +229,9 @@ public PaymentMethod(String id, long maxTradePeriod, @NotNull Coin maxTradeLimit
                     ALI_PAY = new PaymentMethod(ALI_PAY_ID, DAY, maxTradeLimitLowRisk),
                     WECHAT_PAY = new PaymentMethod(WECHAT_PAY_ID, DAY, maxTradeLimitLowRisk),
```
ManfredKarrer, line 230: Is that really a low risk payment method? Those are the ones where chargeback risk is super low or even stated that it is not possible to do chargeback. If so great, otherwise maybe better to use the mid risk parameter.

ManfredKarrer: Ah I just read the comment in the issue. Seems its really low risk. So lets keep it.

ManfredKarrer (approved):
ACK

Thanks a lot for addint it! Dont forget to file a compensation request at https://github.com/bisq-network/compensation/issues

ManfredKarrer:
@leshik Btw if you would be interested to work more on Bisq, please get in touch (Slack would be easiest)! We have plenty of open tasks, not all are in Github... If you are not familiar with the Bisq DAO please check out https://docs.bisq.network/dao/phase-zero.html . Partizipation in Bisq gets rewarded. If the DAO is too risky for you, lets talk there are other options as well...

leshik:
@ManfredKarrer oh thanks, I should learn more about DAO, will read & then join Slack.

ManfredKarrer:
We are growing very fast now and we need more devs! So you are very welcome!

yebii:
Hello, may I know where do you find the related documents? I have searched for several days and did not find the official intergration guidelines and API……！Thanks!

PR #1841, opened 2018-10-30
Add BTC address param, use String for param value instead of long

ManfredKarrer:


ripcurlx (approved):
utACK - @ManfredKarrer I guess this is mainly to have more flexibility within the DAO?

ManfredKarrer:
Yes, it make it easier to add new param types (like in that case the btc donation address, which would not have been possible with long data type for the value. In a string u can pack all even byte arrays...

core/src/main/java/bisq/core/dao/state/DaoStateService.java
```diff
@@ -770,7 +773,22 @@ public void setNewParam(int blockHeight, Param param, long paramValue) {
                 });
     }
 
-    public long getParamValue(Param param, int blockHeight) {
+    public Coin getParamValueAsCoin(Param param, int blockHeight) {
```
sqrrm, line 776: These getParamValueAs...() functions could use a check that they're used correctly.

core/src/main/java/bisq/core/util/BsqFormatter.java
```diff
@@ -111,32 +114,56 @@ public String formatBTCWithCode(long satoshi) {
...
+
+    private void validateCoinInput(String input, MonetaryFormat coinFormat) throws ValidationException {
+        try {
+            coinFormat.parse(cleanDoubleInput(input));
+        } catch (Throwable t) {
+            throw new ValidationException("Invalid format for a " + coinFormat.code() + " value");
+        }
+    }
+
+    public String formatParamValue(Param param, String value) {
```
sqrrm, line 145: Would be nice if these functions could avoid the large switch blocks and use the ParamType now that it's been added. I think it would make sense, might take a look at that later, I think it would be safer.

core/src/main/java/bisq/core/dao/governance/proposal/param/ChangeParamValidator.java
```diff
@@ -59,120 +62,118 @@ public void validateDataFields(Proposal proposal) throws ValidationException {
         }
     }
 
-    // TODO
-    public void validateParamValue(Param param, long paramValue) throws ChangeParamValidationException {
-        // max 4 times the current value. min 25% of current value as general boundaries
-        checkMinMax(param, paramValue, 300, -75);
+    public void validateParamValue(Param param, String inputValue) throws ValidationException {
```
sqrrm, line 65: All these values would be nice to have in Param.java to be able to see all the relevant info for each param. Perhaps ParamType could also store the restrictions?

sqrrm (commented):
utACK

Had a look at the String param changes and added some comments.

core/src/main/java/bisq/core/dao/state/DaoStateService.java
ManfredKarrer: you mean to check the type? as they are called only from switch cases by type i am not sure if its needed. would throw exceptions otherwise...

core/src/main/java/bisq/core/util/BsqFormatter.java
ManfredKarrer: ah yes, good idea...

core/src/main/java/bisq/core/dao/governance/proposal/param/ChangeParamValidator.java
ManfredKarrer: Yes we could add it the params

PR #1842, opened 2018-10-30
Adjust layout with new design

ManfredKarrer:
- on css there are some changes marked with MK. @Christoph - please
check if those are ok for you, otherwise lets discuss...
- in Formbuilder there is removed the automatic setting of the column
span and index
- Add support for columnIndex as parameter
- setMouseTransparent for addTopLabelTextField is removed (we want that
user can copy text)
- Adjustments in the DAO code are WIP and can be ignored by @Christoph

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -1185,10 +1185,6 @@ dao.cycle.voteReveal=Vote reveal phase
 dao.cycle.voteResult=Vote result
 dao.cycle.phaseDuration={0} blocks (≈{1}); Block {2} - {3} (≈{4} - ≈{5})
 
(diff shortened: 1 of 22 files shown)


ripcurlx:
I'm not sure if we should get rid of the focus indicator (line) on fields that are not editable. If they are disabled I agree not to have this line. Maybe something like that?
bildschirmfoto 2018-10-30 um 15 06 25

If we have values that are just for display maybe we should just use a Label? Limitation is atm a Textfield that is set manually not selectable. Setting it as disabled isn't the right state imho.
bildschirmfoto 2018-10-30 um 15 26 19

ManfredKarrer:
As discussed on phone we will go for that for now but might change if we find better ideas.

ripcurlx (approved):
ACK

PR #1846, opened 2018-10-30
Adjust with redesign

ManfredKarrer:


ManfredKarrer:
Was created to master instead of release branch...

PR #1847, opened 2018-10-31
Adjust in DAO views with redesign

ManfredKarrer:


desktop/src/main/java/bisq/desktop/util/FormBuilder.java
```diff
@@ -962,23 +1001,36 @@ private static Label getTopLabel(String title) {
...
 
 
+    public static Tuple3<Label, TxIdTextField, VBox> addTopLabelTxIdTextField(GridPane gridPane, int rowIndex, String title, double top) {
+        TxIdTextField textField = new TxIdTextField();
+        textField.setFocusTraversable(false);
+
+        final Tuple2<Label, VBox> topLabelWithVBox = addTopLabelWithVBox(gridPane, rowIndex, title, textField, top);
+
+        // TOD not 100% sure if that is a good idea....
+        //topLabelWithVBox.first.getStyleClass().add("jfx-text-field-top-label");
```
ripcurlx, line 1028: I'll look into this at a later point, when the most critical parts are finished, as it should be a fix once work everywhere.

ripcurlx (approved):
utACK

PR #1851, opened 2018-10-31
Bonded reputation

ManfredKarrer:
PR https://github.com/bisq-network/bisq/pull/1776 is hard to merge dorectly as there have been many cahnges. So I merged it locally and added also a bit more work in progress implementation.


ripcurlx (approved):
utACK

core/src/main/java/bisq/core/dao/state/DaoStateService.java
```diff
@@ -646,6 +646,10 @@ public boolean isLockupOutput(TxOutput txOutput) {
                 .findFirst());
     }
 
+    public Optional<TxOutput> getLockupOpReturnTxOutput(String txId) {
```
sqrrm, line 649: Perhaps a check to make there is actually an opreturn output would be good

desktop/src/main/java/bisq/desktop/main/dao/bonding/unlock/LockupTxListItem.java
```diff
@@ -88,7 +99,18 @@
     }
 
     public boolean isLockupAndUnspent() {
-        return !isSpent() && getTxType() == TxType.LOCKUP;
+        boolean isLocked;
```
sqrrm, line 102: I've tried to avoid using 'locked' as a key word, it's not entirely clear. Better to use 'lockup', 'unlocking', 'unlocked' or more specific words for clarity.

sqrrm (commented):
ACK

core/src/main/java/bisq/core/dao/state/DaoStateService.java
ManfredKarrer: like: 
```  
public Optional<TxOutput> getLockupOpReturnTxOutput(String txId) {
        return getTx(txId).map(Tx::getLastTxOutput).filter(txOutput -> txOutput.getOpReturnData() != null);
    }
```

??

desktop/src/main/java/bisq/desktop/main/dao/bonding/unlock/LockupTxListItem.java
ManfredKarrer: ok

ManfredKarrer: i will try to use a state enum, the booleans are not covering the states well as u need to combine a few...

core/src/main/java/bisq/core/dao/state/DaoStateService.java
sqrrm: Yes, perfect

desktop/src/main/java/bisq/desktop/main/dao/bonding/unlock/LockupTxListItem.java
sqrrm: Yeah, if you can think of a better way to convey the different states. Sometimes we want to know if it's spendable, sometimes only if it's being unlocked and so on.

PR #1860, opened 2018-11-03
Add CODEOWNERS file

cbeams:
Per documentation at https://help.github.com/articles/about-codeowners/, this file specifies who should get requested to review what in the repository. I was prompted to add this by the desire to auto-request @blabno for all `/assets` PRs, and added other obvious or established code owners as well.

@ManfredKarrer, take note that while you're the default code owner, and will therefore be requested to review all PRs by default, you will *not* be requested to review any PRs for subsequent entries in the file that you are not the code owner for. So for example, you would be requested to review anything under `/core` because no more specific code owner is specified for that, but you will not be requested to review `*.gradle` files, because I'm the code owner for those.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,3 +2,4 @@
 # See https://help.github.com/articles/about-codeowners/.
 
 * @ManfredKarrer
+/assets/ @blabno
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -3,3 +3,4 @@
 
 * @ManfredKarrer
 /assets/ @blabno
+/desktop/ @ripcurlx, @ManfredKarrer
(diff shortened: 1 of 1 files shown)


ManfredKarrer (approved):
utACK

Looks good to me

devinbileck:
@cbeams As you may have noticed, the latest PRs for assets do not have @blabno auto-assigned as reviewer. It appears that auto-assigning code reviewers only works if the owner has push access to the repo. As per https://help.github.com/articles/repository-permission-levels-for-an-organization/, to `Act as a designated code owner for a repository` requires write permissions.

cbeams:
Good point, I hadn't realized that. Even though it wasn't the intended result, the outcome is actually OK, because while @blabno doesn't get assigned as a reviewer, neither does anybody else (usually @ManfredKarrer is usually the fallback as he's the catch-all `*` code owner). @blabno can still review these as he has time, but it's basically a good thing that no one is explicitly "on the hook" for these reviews. We just get to them as we have time.

PR #1808, opened 2018-10-23
Cleanup gradle config

devinbileck:
I made a few changes at an attempt to cleanup and reorganize some of the gradle config.

- Remove duplicate openjfx dependencies in common/build.gradle.
- Use latest Gradle (4.10.2).
- Move desktop/gradle files to root gradle folder. Rather than having a gradle folder underneath the desktop folder, put its content in the root gradle folder to reduce duplication and have a single location for gradle content.
- Define some of the common buildscript dependencies in the root build.gradle so as
to reduce duplication and easier maintenance (i.e. updating version numbers). Gradle build files within sub folders should be able to apply these plugins without also having to define them as a dependency first. **I am not sure if this is a desired change. Feedback appreciated!**

PR #1862, opened 2018-11-04
Refactor Gradle build

cbeams:
This PR supersedes @devinbileck's similar work at #1808. It makes a number of similar changes, does a few things differently, and then goes a good deal further toward simplifying the build arrangement. This is work that I've been meaning to do since reconsolidating into a single repository, but am just now getting to (thanks Devin for spurring that on).

There is now just a single `build.gradle` file in the root directory that uses `configure` blocks to configure each project in turn. It's important to understand that these blocks are additive. That is, it's possible to configure several projects at once in a single `configure` block, and then further configure each individual project in a dedicated `configure` block. You can see this being done in the way that all subprojects with a `main` method are configured with the `application` plugin in a single `configure` block, and then how each of those projects are further configured individually with their specific dependencies, etc, in subsequent blocks.

The bottom line is that we now have one-stop shopping for all our build needs, while still keeping concerns separated. Much cruft and duplication has been removed, and we're not doing anything really fancy with the build anywhere. It should be possible, with a minimum of ramp-up time for contributors not too familiar with Gradle to understand what's going on and to be able to make what changes they need. If you are one of those contributors, and find this new arrangement otherwise, please let me know.


mrosseel:
hi Chris thx for the effort, could you have a quick look at the http-api gradle config to assess if it fits in the new structure? 

https://github.com/mrosseel/bisq/blob/http-api/http-api/build.gradle

cbeams:
Looks like it’ll be straightforward to integrate. Happy to help when needed.

mrosseel:
> Looks like it’ll be straightforward to integrate. Happy to help when needed.

great, once it's merged in master we'll have a go at it and call for help if necessary :)

build.gradle
```diff
@@ -1,3 +1,320 @@
...
+    }
+
+    repositories {
+        jcenter()
+        maven { url 'https://jitpack.io' }
+        maven { url 'https://raw.githubusercontent.com/JesusMcCloud/tor-binary/master/release/' }
+    }
+
+    dependencies {
+        testCompile 'junit:junit:4.12'
```
devinbileck, line 46: Since you have defined the junit version above, I assume this can be changed to the following?
`testCompile "junit:junit:$junitVersion"`

build.gradle
```diff
@@ -1,3 +1,320 @@
...
+
+    dependencies {
+        compile "org.openjfx:javafx-base:11:$os"
+        compile "org.openjfx:javafx-graphics:11:$os"
+        compile "com.google.protobuf:protobuf-java:$protobufVersion"
+        compile 'com.google.code.gson:gson:2.7'
+        compile('com.googlecode.json-simple:json-simple:1.1.1') {
+            exclude(module: 'junit')
+        }
+        compile 'org.springframework:spring-core:4.3.6.RELEASE'
```
devinbileck: Since you have the spring version defined above, I assume this can be changed to the following?
`compile "org.springframework:spring-core:$springVersion"`

build.gradle
```diff
@@ -1,3 +1,320 @@
...
+    }
+}
+
+
+configure(project(':desktop')) {
+    apply plugin: 'com.github.johnrengelman.shadow'
+    apply plugin: 'witness'
+    apply from: '../gradle/witness/gradle-witness.gradle'
+
+    version = '0.8.0-SNAPSHOT'
```
devinbileck, line 202: A comment that is out of scope for this PR. But perhaps this should be taken out of the gradle file and put into a resource file, similar to how it is done for pricenode?
`version = file("src/main/resources/version.txt").text`

devinbileck (approved):
utACK. Looks good to me. Thanks for putting this together!
Just have a few minor comments.

build.gradle
cbeams: Good catch. I extracted that variable when there were still multiple declarations of the same junit dependency across the projects. Now, however, there is just one declaration applied to all subprojects in the `configure(subprojects)` block, so I'll simply remove the variable as it's no longer needed. The practice here is to extract these variables only as needed to eliminate duplication. If there's just one declaration, the version can remain inline.

build.gradle
cbeams: Yes, an oversight on my part, thanks. Fixed in 978fd2d4a.

build.gradle
cbeams: Yeah, I cleaned up a number of unnecessary, unused, or just wrong instances of version metadata elsewhere in this PR, but in this case, the assignment of the `$project.version` variable affects the name of the resulting jar and shadow jar, so it's not quite as simple as extracting it to a resource file. There is also a lot of duplication of the current version throughout the codebase and supporting shell scripts that are used during releases. As you mention, it's a topic for another, more comprehensive PR that really cleans this stuff up.

ripcurlx (approved):
ACK - works for me. Thanks for cleaning up the build scripts @cbeams and @devinbileck!

PR #1889, opened 2018-11-07
Complete bond domain and views

ManfredKarrer:


ripcurlx (approved):
utACK

core/src/main/java/bisq/core/dao/state/DaoStateService.java
```diff
@@ -743,29 +743,75 @@ public long getTotalAmountOfUnLockedTxOutputs() {
...
-        // TODO SQ TxOutputType is immutable after parsing
-        // We need to add new checks if a txo is not confiscated by using the map similar like utxo map
-        // txOutput.setTxOutputType(TxOutputType.BTC_OUTPUT);
+    public boolean isConfiscatedLockupTxOutput(String lockupTxId) {
+        return getLockupTxOutput(lockupTxId)
+                .map(lockupTxIdxOutput -> isConfiscated(lockupTxIdxOutput.getKey()))
+                .orElse(false);
     }
 
+    public boolean isConfiscatedUnlockTxOutput(String unlockTxId) {
```
sqrrm, line 791: What is the motivation to do confiscation on both lockupTxOutput and unlockTxOutput? Wouldn't it be clearer to only use the lockupTxOutput for confiscation?

sqrrm (commented):
utAck

core/src/main/java/bisq/core/dao/state/DaoStateService.java
ManfredKarrer: If the user has created already the unlock tx confiscating the lockup txo will not have any effect. For identifying we use the lockup tx.
But after a second thought, you might be right. If the lockup txo is confiscated all follow up are invalid anyway. Need to check again and see if we are filtering by unspent state...
Would need more testing anyway...

sqrrm: That's what I was thinking, no point in targeting the individual unlock tx since the bsq is already confiscated in the lockup. But perhaps it's needed to mark the unlock txouput as confiscated at the point of parsing the vote result if it's already being unlocked.

PR #1893, opened 2018-11-07
Torrc

### LINKED ISSUE #413: Can't set up proxy connection to connect to Tor network
Hi all,

When launching Bitsquare in Windows 7, the app tries to connect to Tor network automatically. This fails because my connection MUST pass through a proxy. I cannot set it up anywhere. I tried to edit the torrc file manually (C:\Users\myUsername\AppData\Roaming\Bitsquare\mainnet\tor\torrc), where I can add my proxy settings with the "HTTPSProxy" setting, but this torrc file gets overwritten every time I run bitsquare, thereby losing the HTTPSProxy setting.

Any easy way around this?

Cheers!

### LINKED ISSUE #483: Bounty: Investigate solution how to solve blocked Tor connection in China
Bitsquare cannot be used inside China mainland because Tor is blocked by the (not so) great Firewall.

We want to set up a bounty for finding the best technical solution to overcome that problem.
It might be possible to solve by using a VPN, though it would still be a big hurdle for many users (https://forum.bitsquare.io/t/using-bitsquare-in-mainland-china). If we find a way to integrate a VPN inside Bitsquare so the user has no setup costs, that might be an interesting approach.  
If VPN does not solve that issue, Tor Bridges or Tors pluggable transports might be solutions.
Please note that we use Tor Hidden Services, so Bitsquare Tor traffic is never leaving the Tor network to the clearnet (via exit nodes), so it is a different use case as usual.

### LINKED ISSUE #829: Allow Manual Proxy Specification At Startup
Upon first starting the application (running an x86_64 GNU/Linux machine, running BitSquare v0.4.9.9.3) I noticed that it started trying to connect to the TOR network (and gets stuck there, but that's beside the point). However I already have an instance of TOR running, and would prefer to be able to use the system one instead in order to avoid double-work and it would also lower startup time (since the system TOR is already connected).

I have read in other issues (and seen the `--help` option) that there are the `--socks5ProxyHttpAddress` and `--socks5ProxyBtcAddress` options, however I have also read that these do not redirect all traffic through system TOR and will still try to initialize BitSquare's own TOR instance.

My suggestion is to either have the application ask the user if they would like to specify a system proxy address upon startup or simply (like the other options) have a command line argument that can be used to direct all TOR traffic normally sent through BitSquare's instance through the system TOR. This may also help so that people can use other networks, such as I2P.

### LINKED ISSUE #965: Make torrc file editable
Some people want to edit the torrc file to add proxies. The file gets replaced at each startup and local path is written to it. The handling of that is questionable and subject for improvement, but it would be better to invest the effort in the new Tor library instead to modify the existing.

An alternative approach would be to let the user append entries to the torrc file via a program argument. 

Easiest way would be to add a list of optional key/value pairs to OnionProxyManager.installAndConfigureFiles(). The list get passed there and is fed by a new prog argument.

See also: 
https://github.com/bisq-network/exchange/issues/829
https://github.com/bisq-network/exchange/issues/483
https://github.com/bisq-network/exchange/issues/413

freimair:
Makes torrc "editable" via system properties.

### Usage

- use `-Dtorrcfile=/path/to/torrc` to source a valid torrc file from you system
- use `-Dtorrc:key1=value1 -Dtorrc:key2=value2` to append the tor configuration
- use both for a more complex setup

### Examples

- use `-Dtorrc:Socks5Proxy=host[:port]` to configure a Socks5 proxy
- use `-Dtorrcfile=/path/to/customtorrc` with `customtorrc` containing

    ```
    Log info file /tmp/bisq.tor.log
    TruncateLogFile 1
    ```
    to configure raw tor logging for troubleshooting.

### Notes
note that torrc-entries which are critical to the software cannot be overwritten 

bumps [netlayer](https://github.com/JesusMcCloud/netlayer) to 0.4.7  in order to profit from its bugfixes and addresses fixes #965, fixes #829 and thus fixes #413 and fixes #483


p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
```diff
@@ -242,8 +244,21 @@ private void createTorAndHiddenService(File torDir, int localPort, int servicePo
         ListenableFuture<Void> future = executorService.submit(() -> {
             try {
                 long ts1 = new Date().getTime();
+
+                Torrc override = null;
+
+                // check if the user wants to provide his own torrc file
+                String torrcfile = System.getProperty("torrcfile");
```
ManfredKarrer: I would prefer camelCase as its more readable... torrcfile -> torrcFile. As Torrc comes from a lib it cannot be changed TorRc which I would prefer as well... We use basically camelCase when 2 words are connected. If its a abbreviation liek RC or BTC we use first letter uppercae only -> Rc or Btc, so btcAddress not BTCAddress.

p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
```diff
@@ -257,6 +258,20 @@ private void createTorAndHiddenService(File torDir, int localPort, int servicePo
...
 
+                // check if the user wants to temporarily add to the default torrc file
+                LinkedHashMap<String, String> tmp = new LinkedHashMap<>();
+                System.getProperties().forEach((k, v) -> {
+                    if(((String) k).startsWith("torrc:"))
+                        tmp.put(((String) k).substring(6), (String) v);
+                });
+                if(!tmp.isEmpty())
+                    // check for custom torrcfile
+                    if(null != override)
```
ManfredKarrer: Why not if(override != null)? I know some devs use that style but not sure the reason. For me its better readable when null is right...

p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
```diff
@@ -249,7 +253,7 @@ private void createTorAndHiddenService(File torDir, int localPort, int servicePo
                 Torrc override = null;
 
                 // check if the user wants to provide his own torrc file
-                String torrcfile = System.getProperty("torrcfile");
+                String torrcfile = System.getProperty(CONFIG_TORRCFILE);
```
ManfredKarrer: Should we use our program arguments instead of system properties? I am not sure whats better here or if both make sense. Just want to avoid that we mix it as we never use system properties for application specific options so far. But maybe it integrates better with certain environments like with Tails?

ManfredKarrer:
Beside that nits in the comments looks good and I tested it with your example and it works. Very helpful or debugging tor issues to add that logs... At the user would need to change the log level to get full tor logging included.
Thanks for getting that implemented! Was often requested feature.

p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
freimair: Well, an old habit of mine. [For your reference.](https://en.wikipedia.org/wiki/Yoda_conditions). This style of coding saved myself countless hours of debugging. Shall I swap it nonetheless?

p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
freimair: I thought about that but decided that the implementation is much faster when done this way. I have not been aware of the historical ways of doing it.
I will look into it.

p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
freimair: Old habit of mine. Just wait a little.

p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
ManfredKarrer: You can find it in NetworkOptionKeys. It is a bit cumbersome to define all but if you follow one existing example its not hard. I would prefer to stick with program arguments as they also come with the help option.

p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
ManfredKarrer: Ah thanks for the link. 
With the case of NullPointerException I agree and also use it (e.g. "BTC".equals(code)) but with the asignment argument I think with a good IDE like IntelliJ mistakes should be easy to spot. 
Would prefer readiblity here.

core/src/main/java/bisq/core/app/BisqExecutable.java
```diff
@@ -330,6 +333,14 @@ protected void customizeOptionParsing(OptionParser parser) {
...
                 description("A proxy address to be used for Http requests (should be non-Tor). [host:port]", ""))
                 .withRequiredArg();
+        parser.accepts(NetworkOptionKeys.TORRC_FILE,
+                description("An existing torrc-file to be sourced for Tor. Note that torrc-entries, which are critical to Bisqs flawless operation, cannot be overwritten.", ""))
+                .withRequiredArg();
+//                .withValuesConvertedBy(new PathConverter(PathProperties.FILE_EXISTING, PathProperties.READABLE));
+        parser.accepts(NetworkOptionKeys.TORRC_OPTIONS,
+                description("A list of torrc-entries to amend to Bisqs torrc. Note that torrc-entries, which are critical to Bisqs flawless operation, cannot be overwritten. [torrc options line, torrc option, ...]", ""))
+                .withRequiredArg();
+//                .withValuesConvertedBy(RegexMatcher.regex("^([^\\s,]+\\s[^,]+,?\\s*)+$"));
```
freimair, line 343: JOptSimple offers input parsing and therefore allows for instant feedback (i.e. showing --help). However, JOptSimple only engages these checks when the value of interest is accessed via `option.valueOf`. So as of now, these lines are disabled.

In order to make Bisq more user friendly, we can do something like:
1) do not use this feature. Errors on command line inputs show up some time later (if manually checked), probably hidden under tons of logs.
2) enable line 339 and 343 above and receive an exception stating the cause
3) change BisqExecutable.java: move line 151: `setupEnvironment(options);` up after line 134, i.e. into the try-catch-block, that houses the initial parsing
4) enable line 339 and 343 and change BisqExecutable::setupEnvironment to something like that:
     ```
    protected void setupEnvironment(OptionSet options) {
        try {
            bisqEnvironment = getBisqEnvironment(options);
        }catch (OptionException e) {
            System.err.println(e.getMessage() + ": " + e.getCause().getMessage());
            // we only tried to load some config, so no graceful shutdown is required
            System.exit(1);
        }
    }
    ```
    resulting in a final log line such as
    ```
    Cannot parse argument '/tmp/torr' of option torrcFile: File [/tmp/torr] does not exist
    ```
5) somehow influence JOptSimple to do the check on parse

IMHO, 4 seems like a good solution. 1 and 2 are not user friendly at all, 3 only states the help index without any detailled information on the error and 5 might probably never happen

ManfredKarrer: I am not very experienced with the option parsing part and also don't have strong opinions there. Most options are used anyway only by devs so far, but of course that torrc option might be more used by users. I leave it to you what you think makes most sense and causes least effort.

ManfredKarrer (approved):
utACK

PR #1895, opened 2018-11-07
Support handling of removed assets.

ManfredKarrer:
- cross check the isCryptoCurrency method if the symbol matches any
fiat currency and of both not match consider it still a CC. In case of
a removed asset it was returning false before which caused an issue in
the trade currency pair showing both sides as BTC.
- Show N/A for name in case the asset is not available.

The code mostly operates with the ticker symbol which gets stored in
the offer, so even an asset is removed traders who have that asset in
their account list can trade without problems.
Tested create offer, take offer and executing a trade.


ripcurlx (approved):
utACK

cbeams (approved):
utACK

ManfredKarrer:
Fixing tests and add some changes...

ManfredKarrer:
As the additional changes have been non trivial I leave it for another ACK.

ripcurlx:
utACK

Issue #1885, opened 2018-11-05
Update instructions for Monero transaction proof


ManfredKarrer:
Seems that Monero has changed the way how a sender can proof his payment. In the latest GUI wallet there is the tool to do that. We need to update our popup text where we explain what is needed.

ManfredKarrer:
It is a small task but would be nice if one experienced with Monero can do that.
I suggest a bounty of 30 BSQ for it.

ManfredKarrer:
<img width="816" alt="screen shot 2018-11-05 at 09 05 25" src="https://user-images.githubusercontent.com/1449498/48002594-fc497e00-e0d9-11e8-82cd-fea221dc88d5.png">

KanoczTomas:
It is fairly easy with the gui, one has to go to history and click on `P`, it will show the payment proof for the tx. Double click on the generated proof and it gets to the clipboard.

KanoczTomas:
@ManfredKarrer if some screenshots would help I am setting up a testnet node for monero just now and can do some screenshots with some steps. Either the text will have a link to a github howto file with steps (which could be updated separately) or the release will have the images in it. The steps are super easy though, not sure if it is really needed. Let me know.

ManfredKarrer:
We need written (simple formatted) text as otherwise its more effort to add it the popup in the app...

KanoczTomas:
"Go to `History` tab, click on the sent transaction. Make sure details are shown for it (press `down arrow` on the right, if details are not shown). Click on the `P` button above the `?` button to generate the proof. Double click on the generated proof to get it to clipboard. Note that the proof as well as the txid is needed for correct verification."

ManfredKarrer:
Implemented with #1902

KanoczTomas:
@ManfredKarrer what does this mean from the bounty standpoint?

PR #1762, opened 2018-10-07
List Dragonglass (DRGL)

aejontargaryen:
This is a resubmission of [(DRGL) bisq-assets#86](https://github.com/bisq-network/bisq-assets/pull/86)
awaiting ACK/approval.

Official project URL: https://zirtysperzys.org
Official block explorer URL: http://zirtysperzys.info
Additional: DRGL wallets support transaction keys that enable 3rd party audits in case of a dispute similar to Monero

PR #1902, opened 2018-11-09
Update altcoin accounts info popups

aejontargaryen:
reference issue https://github.com/bisq-network/bisq/issues/1885

as well as Arbitration instructions  +   version requirement for DRGL wallets. 

Asset reference PR https://github.com/bisq-network/bisq/pull/1762

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -1027,10 +1027,11 @@ If you do not save the transaction secret key immediately after transferring CCX
 If you do not understand these requirements, seek help at the Conceal discord (http://discord.conceal.network).
 account.altcoin.popup.drgl.msg=Trading Dragonglass on Bisq requires that you understand and fulfill \
 the following requirements:\n\n\
-Dragonglass arms you with the power to prove payments while remaining anonymous thru use of your TXN-Private-Key.\n\
+Because of the privacy Dragonglass provides a transaction is not verifyable on the public blockchain. If required you \
+can prove your payment thru use of your TXN-Private-Key.\n\
 The TXN-Private Key is a one-time key automatically generated for every transaction that can \
 only be accessed from within your DRGL wallet.\n\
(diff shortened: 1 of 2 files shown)


ripcurlx (commented):
utACK - as I never used the Monero wallet I can't say if the mentioned commands in the text are correct. @ManfredKarrer Do you have any concerns regarding the textual changes?

ManfredKarrer (commented):
Did you check if those old commands on the terminal are still supported?

aejontargaryen:
@ManfredKarrer yes same commands, just different cli-wallet name.
I included this additional link in the info box as well. 
https://www.getmonero.org/resources/user-guides/prove-payment.html

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1021,6 +1025,22 @@ Because Conceal is a privacy coin, block explorers cannot verify transfers.\n\n\
 If you cannot provide the required data to the arbitrator, you will lose the dispute case.\n\
 If you do not save the transaction secret key immediately after transferring CCX, it cannot be recovered later.\n\
 If you do not understand these requirements, seek help at the Conceal discord (http://discord.conceal.network).
+account.altcoin.popup.drgl.msg=Trading Dragonglass on Bisq requires that you understand and fulfill \
+the following requirements:\n\n\
+Dragonglass arms you with the power to prove payments while remaining anonymous thru use of your TXN-Private-Key.\n\
```
ManfredKarrer: I would prefer a more neutral language here. We are not promoting any coin (even it sounds interesting one). I will chang it myself to not postpone the merge...

ManfredKarrer (approved):
utACK

core/src/main/resources/i18n/displayStrings.properties
aejontargaryen: @ManfredKarrer yes sir. thats best, as i do have a tendency to get carried away. :smile: Thanks!

PR #1905, opened 2018-11-10
Optimize initial data load

ManfredKarrer:
We changed the earlier behaviour with delayed execution of chunks of the list as it caused 
worse results as if it is processed in one go.
Main reason is probably that listeners trigger more code and if that is called early at 
startup we have better chances that the user has not already navigated to a screen where the
trade statistics are used for UI rendering.
We need to take care that the update period between releases stay short as with the current
situation before 0.9 release we receive 4000 objects with a newly installed client, which 
causes the application to stay stuck for quite a while at startup.

### DIFF
--- a/core/src/main/java/bisq/core/payment/AccountAgeWitnessService.java
+++ b/core/src/main/java/bisq/core/payment/AccountAgeWitnessService.java
@@ -136,9 +136,7 @@ public class AccountAgeWitnessService {
     }
 
     private void addToMap(AccountAgeWitness accountAgeWitness) {
-        log.debug("addToMap hash=" + Utilities.bytesAsHexString(accountAgeWitness.getHash()));
-        if (!accountAgeWitnessMap.containsKey(accountAgeWitness.getHashAsByteArray()))
-            accountAgeWitnessMap.put(accountAgeWitness.getHashAsByteArray(), accountAgeWitness);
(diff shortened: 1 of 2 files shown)


ManfredKarrer:
@ripcurlx I was wondering if we should shoot out a quick new release based on the old release but with the actual trade statistic and witness data. I tested with new data directory and at first startup the app get frozen. After that it works, but pretty bad user experience... Surprised that we did not get more complaints... Do you think a new release could be done without a lot of effort (do you have still the old java setup on the VMs)?

cbeams:
Glad you brought this up, @ManfredKarrer. I too have experienced this delay on starting up new instances. I meant to report it earlier myself. A must-fix, IMO. The delay is long enough that many if not most users would think something is fatally wrong with their client.

ripcurlx:
@ManfredKarrer I have installed a new SDK, but I should have the old configured one still in the VM. I could create a 0.8.1 release on Monday based on the last tag and just update the store dbs.

ripcurlx:
We could do it without sending out update messages and so on. Just to have a version for download that doesn't freezes that long on start-up until we have fixed this issue properly.

ManfredKarrer:
@ripcurlx Yes would be great! 
Only the TradeStatistics2Store_BTC_MAINNET and AccountAgeWitnessStore_BTC_MAINNET need to be updated with the a file from a live version. 
And yes just make a silent update without sending a notification.

ripcurlx (approved):
utACK

PR #1912, opened 2018-11-11
Add proof of burn feature

ManfredKarrer:


ManfredKarrer:
This is based on the asset listing fee branch, so that should be merged before...

See: https://github.com/bisq-network/proposals/issues/55

core/src/main/java/bisq/core/dao/governance/proofofburn/ProofOfBurnService.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+
+    ///////////////////////////////////////////////////////////////////////////////////////////
+    // API
+    ///////////////////////////////////////////////////////////////////////////////////////////
+
+    public Transaction burn(String preImageAsString, long amount) throws InsufficientMoneyException, TxException {
+        try {
+            // We create a prepared Bsq Tx for the listing fee.
```
sqrrm:
```suggestion
            // We create a prepared Bsq Tx for the burn amount
```

core/src/main/java/bisq/core/dao/governance/proofofburn/ProofOfBurnService.java
```diff
@@ -0,0 +1,245 @@
...
+    @Override
+    public void onParseBlockChainComplete() {
+    }
+
+
+    ///////////////////////////////////////////////////////////////////////////////////////////
+    // API
+    ///////////////////////////////////////////////////////////////////////////////////////////
+
+    public Transaction burn(String preImageAsString, long amount) throws InsufficientMoneyException, TxException {
```
sqrrm, line 141: Would be nice if this could reuse the code from AssetService.payFee()

core/src/main/java/bisq/core/dao/node/parser/TxOutputParser.java
```diff
@@ -115,6 +115,8 @@ void processTxOutput(TempTxOutput tempTxOutput) {
                 // The LOCKUP BSQ is burnt unless the output exactly matches the input, that would cause the
                 // output to not be BSQ output at all
                 handleUnlockBondTx(tempTxOutput);
+            } else if (isBurnFeeTxOutput(tempTxOutput)) {
```
sqrrm: I found this a bit confusing. Seems correct but might help with extra comment or moving this into the next else if, eg.
`else if (!isBurnFeeTxOutput(tempTxOutput) && availableInputValue > 0 && availableInputValue >= txOutputValue) {`

Maybe change function name to `isBurnFeeBtcTxOutput`

sqrrm (approved):
utACK

core/src/main/java/bisq/core/dao/governance/proofofburn/ProofOfBurnService.java
ManfredKarrer: Not sure if we want marry both to close by using a common super class. Prefer a partial bit code duplication here. But no too strong opinion either...

core/src/main/java/bisq/core/dao/node/parser/TxOutputParser.java
ManfredKarrer: Yes, true. Will try to make it more clear.

core/src/main/java/bisq/core/dao/governance/proofofburn/ProofOfBurnService.java
sqrrm: No, I don't think sharing this kind of code via inheritance is right, but putting it in a utility class would work.

ManfredKarrer: Yes, agree, but think that this method is too trivial/small to justify an extra util class. But of course subject of opinion...

Issue #1909, opened 2018-11-11
BTC/USD and BTC/EUR trades showing up as BTC/BTC


cbeams:
I'm running my arbitration client against current `HEAD` of the `master` branch (fa911e96ed18bfb0214a441be8c00434fd3cd68f) and seeing what are actually BTC/EUR or BTC/USD trades showing up as BTC/BTC trades.

When I build and run the client against the `v0.8.0` tag, these trades show up correctly, and I know they showed up correctly even a few days ago as well (I've been rebuilding and running against master for the last week or so).

![image](https://user-images.githubusercontent.com/301810/48309977-86379200-e585-11e8-9a08-d47370957108.png)

![image](https://user-images.githubusercontent.com/301810/48309979-8f286380-e585-11e8-991f-5dac21360a7f.png)

Perhaps this has something to do with the recent changes around asset listing?

cbeams:
This behavior is also showing up in the Market->Trades UI:

![image](https://user-images.githubusercontent.com/301810/48310704-85a5f800-e593-11e8-9254-fcc1b3ce6333.png)

ManfredKarrer:
It is probably related to the change in chang ein isCuryptoCurrency. I will have a look.

PR #1913, opened 2018-11-11
Fix issues with wrong base currency

ManfredKarrer:
From the changes in isCryptoCurrency which return true in case of BTC
there have been some bugs. We changed the check to distinguish between
altcoin and fiat to check for BTC as base currency now.
That area would require a bigger refactoring...

Fixes https://github.com/bisq-network/bisq/issues/1909


cbeams (approved):
ACK. I've just tested this locally and currency pairs are being reported as expected once again.

PR #1917, opened 2018-11-13
Fixed netlayer and tor binary dependency issues

freimair:
netlayer had and has a well-hidden build configuration issue since version 0.4.2. It always fetches the "master" binaries instead of the binaries of the respective release. although until recently, bisq wanted to use 0.4.2, has actually used 0.4.9pre-draft already. Due to changes in netlayer and its upcoming release, the netlayer build configuration has changed and thus, the issue showed itself. (at least we now know that Netlayer 0.4.9pre still works with bisq :) )

This fixes the issue temporarily.

I am currently working on getting the binaries of TorBrowser 8.0.3 (out for a month or so) into bisq. so IMHO this temporary fix is GEFN


ManfredKarrer (approved):
utACK

PR #1924, opened 2018-11-14
Revise contributor / developer docs

cbeams:
This PR simplifies and restructures our developer documentation for ease of discovery and use. Major changes include:

 - Breaking up the root-level README into docs/build.md and docs/idea-import.md
 - A complete rewrite of CONTRIBUTING.md, loosely based on the structure of Bitcoin Core's own CONTRIBUTING.md.
 - Separating out `dev-setup.md` and `dao-setup.md`, as having the two sets of instructions intermixed was a bit overwhelming and confusing.
 - Removal of older files under scripts and misc. Each of these removals have been done in a dedicated commit for ease of reverting in case one is deemed important enough to keep around.

The motivation for putting these changes together was that I followed the instructions in dev-setup.md (formerly DEV_SETUP.md), and found them useful and accurate, but a bit too much with all the DAO information intermixed. That led to separating the two docs out as mentioned above, which in turn led to this more general overhaul. Altogether, I think these docs are now in good shape for guiding any new contributor in the right direction, giving them just what they need to know at each step, and hopefully keeping the instructions simple and enjoyable to follow along the way.

Note that there is now a bit of overlap between the docs here and the "contributor docs" section of https://docs.bisq.network. I think it may make sense to make docs.bisq.network our user-facing and non-dev contributor documentation portal, while keeping all things dev here. Particularly where class names, command line args and the like are mentioned, it's better to keep such docs in the repository, so as to keep things in sync with the code that's being described. Where higher-level contributor documentation, e.g. roles, proposals and compensation are concerned, it probably makes sense to keep documenting that at docs.bisq.network. /cc @m52go


CONTRIBUTING.md
```diff
@@ -1,3 +1,85 @@
...
+ - Create a topic branch from the `master` branch
+ - Commit patches
+ - Submit a pull request from your topic branch back to the `master` branch of the main repository
+
+Pull requests should be focused on a single change. Do not mix, for example, refactorings with a bug fix or implementation of a new feature. This practice makes it easier for fellow contributors to review each pull request on its merits and and to give a clear ACK/NACK (see below).
+
+
+## Reviewing Pull Requests
+
+Bisq follows the review workflow established by the Bitcoin Core project. The following is adapted from the Bitcoin Core contributor documentation](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#peer-review):
```
devinbileck: Missing an opening square bracket for the link text.

CONTRIBUTING.md
```diff
@@ -1,3 +1,85 @@
...
+### Sign your commits with GPG
+
+_Not strictly required, but recommended for those who commit frequently._
+
+See https://github.com/blog/2144-gpg-signature-verification for background and
+https://help.github.com/articles/signing-commits-with-gpg/ for instructions.
+
+### Use an editor that supports Editorconfig
+
+The [.editorconfig](.editorconfig) settings in this repository ensure consistent management of whitespace, line endings and more. Most modern editors support it natively or with plugin. See http://editorconfig.org for details. ([bisq-network/style#9](https://github.com/bisq-network/style/issues/9))
```
devinbileck: I assume the link to bisq-network/style#9 should not be here?

devinbileck: I guess it should be bisq-network/style#10?

docs/dev-setup.md
```diff
@@ -0,0 +1,92 @@
...
+
+The regtest mode operates a local Bitcoin network on your computer. This environment is ideally suited for testing because you are able to create blocks on demand (no need to wait for confirmations) and you don't need to download the blockchain. By creating blocks you act like a miner and you can generate new Bitcoin.
+
+You can find more information about the Bitcoin regtest mode [here](https://bitcoin.org/en/developer-examples#regtest-mode).
+
+Navigate to the [bitcoin.conf](https://en.bitcoin.it/wiki/Running_Bitcoin#Bitcoin.conf_Configuration_File) file and set `regtest=1`, or add `-regtest` as a program argument when starting Bitcoin Core.
+
+At first startup you need to create 101 blocks using the command `generate 101` from the terminal inside Bitcoin Core. 101 blocks are required because of the coin maturity (100 blocks) so you need one more to have at least 50 BTC available for spending.
+
+Later you can create a new blocks with `generate 1`.
```
devinbileck: Later you can create <s>a</s> new blocks with `generate 1`.

docs/dev-setup.md
```diff
@@ -0,0 +1,92 @@
...
+ - localhost + testnet
+ - localhost + mainnet
+ - Tor + regtest
+ - Tor + testnet
+ - Tor + mainnet
+
+
+## Understand Bisq program arguments
+
+There are several program arguments required to run in development mode.
```
devinbileck, line 50: Not required, but a useful argument potentially worth mentioning here is --useDevMode

From https://github.com/bisq-network/bisq/blob/master/common/src/main/java/bisq/common/app/DevEnv.java#L42:

> If set to true we ignore several UI behavior like confirmation popups as well dummy accounts are created and offers are filled with default values. Intended to make dev testing faster.

docs/dao-setup.md
```diff
@@ -0,0 +1,83 @@
...
+    #!/bin/bash
+    echo $1 | nc -w 1 127.0.0.1 5120
+    echo $1 | nc -w 1 127.0.0.1 5121
+    echo $1 | nc -w 1 127.0.0.1 5122
+    echo $1 | nc -w 1 127.0.0.1 5123
+
+It defines the ports where a new block event gets forwarded. Bisq will listen on that port and each Bisq node need to use a different port. You can add or remove ports from the list inside the file if needed.
+
+
+## Program arguments for DAO mode
```
devinbileck, line 35: The arguments are mashed together when the markdown is rendered:
image

Perhaps should make them into list items.

devinbileck:
Thanks @cbeams. This was much needed.
I have reviewed the content and it looks good to me, with just a few minor comments.
I am unable to comment on dao-setup as I am not familiar with it nor have I ran through the process yet.

CONTRIBUTING.md
cbeams: Fixed in 7985ff7e0.

docs/dev-setup.md
cbeams: Fixed in 7985ff7e0.

docs/dao-setup.md
cbeams: Fixed in 7985ff7e0.

docs/dev-setup.md
cbeams: Good call. I'll address this in a subsequent PR, though.

cbeams:
Thanks for the review, @devinbileck. Good catches, all.

CONTRIBUTING.md
cbeams: Fixed in 7985ff7e0.

ManfredKarrer (approved):
utACK

Thanks for cleaning that all up!!!

PR #1927, opened 2018-11-16
Add gitattributes

devinbileck:
Adding a .gitattributes file will ensure that text files that any contributor introduces to the repository have their line endings normalized. This is in reaction to the issue discussed at https://github.com/bisq-network/gradle-witness/pull/1#discussion_r233669786 while trying to generate reproducible builds.


ManfredKarrer:
I leave that for @cbeams ...

cbeams (approved):
ACK. I've added an additional commit to make our .editorconfig match these .gitattribute settings with regard to .bat files.

Issue #1761, opened 2018-10-06
Use system Tor if possible


qubenix:
Currently the software uses its bundled Tor regardless if Tor is already installed on the users system. This results in a [Tor over Tor](https://tor.stackexchange.com/questions/427/is-running-tor-over-tor-dangerous) situation on systems like Whonix which transparently proxy all traffic.

Please see:

+ Tor friendly applications best practices: https://trac.torproject.org/projects/tor/wiki/doc/Tor_friendly_applications_best_practices.

+ Whonix friendly application best practices: https://www.whonix.org/wiki/Dev/Whonix_friendly_applications_best_practices

---

Thank you for all your hard work on this project.

ManfredKarrer:
There is an open issue out for that but dev resources found so far who work on that. If you are a dev you are welcome.

qubenix:
I made a post on the Whonix forum's Bisq thread about needing Java coders: https://forums.whonix.org/t/bisq-the-p2p-exchange-network/4953/17?u=qubenix.

Feel free to close this issue if you feel it is a duplicate. I apologize, I did try to search for this exact issue and didn't see it (although it was mentioned in other issues).

ManfredKarrer:
Thanks. I think it is posted at the netlayer library GH issues. I leave it open here to keep attention to it.

ManfredKarrer:
See https://github.com/JesusMcCloud/netlayer/issues/7

ManfredKarrer:
In anyone is interested to work on that please check with @freimair if he has not started to work on that already....

freimair:
Manfred, you can assign the issue to me, so there is no doubt.

PR #1928, opened 2018-11-16
Use tor binaries from TorBrowser 8.0.3

freimair:
Updates netlayer dependency to version 0.4.7.2 which includes the tor-binaries extracted from the TorBrowser version 8.0.3 release a month or so ago.

Lets see if jitpack works...


freimair:
jitpack does some weird things... onto it

build.gradle
```diff
@@ -125,7 +125,7 @@ configure(project(':common')) {
 configure(project(':p2p')) {
     dependencies {
         compile project(':common')
-        compile('com.github.JesusMcCloud.netlayer:tor.native:0.4.7.2') {
+        compile('com.github.JesusMcCloud.netlayer:tor.native:0.4.7.1.1') {
```
ManfredKarrer, line 128: Why it was reverted to 4.7.1.1?

ManfredKarrer: Ah I saw https://github.com/JesusMcCloud/netlayer/releases/tag/0.4.7.1.1 was the latest release.

ManfredKarrer (approved):
utACK

build.gradle
freimair: Yes, jitpack refused to build the correct commit  behind the 0.4.7.2 tag. It is a quick workaround until https://github.com/jitpack/jitpack.io/issues/3572 is resolved (if it ever gets resolved).

Furthermore, I do have a solution for #1761 almost ready. Hence, I am going to bump the netlayer-version to 0.6 anyways. And yes, the jitpack behaviour does bite us again - hence, 0.6. (All in agreement with Bernd)...

PR #1935, opened 2018-11-17
Bisq can now use an external Tor service

### LINKED ISSUE #492: Using tor binary from o/s instead of distributing it
As part of #484 I noticed Bitsquare distributes it's own tor binary. I think this might be a nice fallback, but I also believe Bitsquare should use the o/s binary in /usr/bin when it it available instead.

This also saves you from needing to update the tor-client whenever a new version comes out. My qnd POC can be found here: https://github.com/SjonHortensius/bitsquare/commit/338f7f117939ecc2fc302c3d57079f6c81c851a7

I'd be happy to rewrite this to something like `torPath = exists(/usr/bin/tor) ? '/usr/bin/tor' : local.path();`. It seems the best location to change this is [the OnionProxyContext constructor](https://github.com/bitsquare/bitsquare/blob/32e1a5a3accbb8ca98a0632f360ee95f1f844360/jtorproxy/src/main/java/com/msopentech/thali/toronionproxy/OnionProxyContext.java#L40)

### LINKED ISSUE #811: Bitsquare not working in Tails OS
I'm trying to run Bitsquare on Tails - I've installed the necessary dependencies and it starts but hangs at startup. I think Bitsquare starts a tor session on its own and doesn't use the preinstalled tor already running. It seems that there's no way to disable the internal tor client and just pass a parameter with proxydetails like "localhost:9150" or something like that.

### LINKED ISSUE #1761: Use system Tor if possible

### LINKED ISSUE #1887: tor blocked in china,  i can use tor browser with proxy in china, but bisq does not has such configuration
tor blocked in china,  i can use tor browser with proxy in china, but bisq does not has such configuration

freimair:
Bisq can now use an already running Tor service (TorBrowser for example). The feature can be used by supplying the required command line parameters on Bisq start, being:
```
--torControlCookieFile <Path>           The cookie file for authenticating    
                                          against the already running Tor     
                                          service. Use in conjunction with -- 
                                          torControlUseSafeCookieAuth (default: )
--torControlPassword <String>           The password for controlling the      
                                          already running Tor service. (default: )
--torControlPort <Integer>              The control port of an already running
                                          Tor service to be used by Bisq      
                                          [port]. (default: )
--torControlUseSafeCookieAuth           Use the SafeCookie method when        
                                          authenticating to the already       
                                          running Tor service. (default: )    
```
In order to get this functionality, the Netlayer dependency and the jtorctl dependency have been bumped to version [0.6](https://github.com/JesusMcCloud/netlayer/releases/tag/0.6) and [1](https://github.com/JesusMcCloud/jtorctl/releases/tag/1), respectively.

closes #1761, closes #811, closes #492, closes #1887


ManfredKarrer (approved):
utACK

PR #1937, opened 2018-11-17
Persist selected account at takeoffer

ManfredKarrer:


ManfredKarrer:
Ups, was to master instead of release branch

PR #1945, opened 2018-11-21
Limit initial window size on application launch

devinbileck:
**Problem:** When launching Bisq on my Windows 10 laptop, the application window is larger than the usable screen bounds causing the title bar of the application to be mostly off the screen. This is an annoyance and requires me to move and resize the window so that it is within the screen bounds.

**Cause:** My laptop has a 13-inch 1920x1080 display and uses 150% display scaling. As a result, it has a usable resolution (maximum screen bounds) of 1280x660 (accounting for the height of the Windows task bar). When launching the application, it sets the initial window height to 710px, which exceeds my usable screen height.

**Fix:** When launching the application, prevent the initial window size from exceeding the maximum available bounds. These bounds account for objects in the native windowing system such as task bars and menu bars.

<s>**Note: This is a work in progress as I have yet to test it on anything other than Windows (I plan to verify against Ubuntu). Nor have I verified a multi-display setup. But I still wanted to put this PR together in case anyone is able to verify it especially on Mac which I do not have available to test.**</s>

### DIFF
--- a/desktop/src/main/java/bisq/desktop/app/BisqApp.java
+++ b/desktop/src/main/java/bisq/desktop/app/BisqApp.java
@@ -71,6 +71,9 @@ import javafx.scene.input.KeyEvent;
 import javafx.scene.layout.Pane;
 import javafx.scene.layout.StackPane;
 
+import java.awt.GraphicsEnvironment;
+import java.awt.Rectangle;
+
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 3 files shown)


devinbileck:
I have tested this on Ubuntu 18.04 at high and low resolutions and no issues encountered. I have also tested it on a multi-display setup on Windows with no issues encountered.

ManfredKarrer (approved):
utACK

ManfredKarrer:
Great thanks!

PR #1950, opened 2018-11-22
Redesign: Improve mobile notification, take offer and add account views

### LINKED ISSUE #1949: Show info about single altcoin account when taking offer
When I create a new offer to but altcoins, e.g., MXR (sell BTC), Bisq shows the first 6 characters of the Monero address to which the funds will be transferred:
screenshot 2018-11-22 at 11 49 33

If a user takes an offer and has more than one altcoin trading account, he will be presented with a dropdown list to select the altcoin account:
screenshot 2018-11-22 at 11 58 48

However, when a user takes an existing offer and has only set up a single altcoin account, Bisq shows only general Payment info without specifying the e.g., Monero address to which the funds will be transferred:
screenshot 2018-11-22 at 11 49 19

This can be confusing if the user wrongly assumes that he will be able to specify the Monero address at a later point of time. Especially if the user has created a Monero altcoin account just for testing without having saved the seed words. 
Showing the first 6 characters of the Monero address also in this case (single account, taking offer) would prevent users from making wrong assumptions about the point of time at which the altcoin address is selected.

ripcurlx:
Fixes #1949.


ManfredKarrer (approved):
utACK

PR #1953, opened 2018-11-23
Remove Utilities.removeCryptographyRestrictions

cbeams:
This code is no longer necessary after the recent move to JDK10, and
removing it entirely avoids spending unnecessary cycles and eliminates
the confusing "Cryptography restrictions removal not needed" message in
console output.


ManfredKarrer (approved):
utACK

PR #1954, opened 2018-11-23
Remove Logback config xml from core module

cbeams:
Usually Bisq is run from its 'shadow jar', i.e. the executable fat jar
that collapses all dependencies and resources into a single file. In that
scenario, only a single 'logback.xml' is included in the fat jar and all
is well.

However, when Bisq is run with a normal external classpath, e.g. from
within IDEA, when using `gradle :desktop:run` or when running one of the
Gradle-generated shell scripts under desktop/build/scripts, multiple
'logback.xml' files are present on the classpath, which causes Logback
to complain with messages like the following:

    Found resource [logback.xml] at [file:.../desktop/build/resources/main/logback.xml]
    Resource [logback.xml] occurs multiple times on the classpath.
    Resource [logback.xml] occurs at [file:.../desktop/build/resources/main/logback.xml]
    Resource [logback.xml] occurs at [jar:file:.../core/build/libs/core.jar!/logback.xml]

Note the last line above. The presence of 'logback.xml' in the bisq-core
jar is the 'duplicate'. This is Logback's way of warning the user that
it doesn't know which file should be considered canonical. The only way
to avoid this error message (and the many other lines that get output
along with it) is to remove the offending duplicate file.

This commit removes logback.xml from core to solve the problem described
above. The file is unnecessary in the context of running bisq-desktop,
bisq-seednode, etc, and the only context in which it would actually be
used is when running BisqHeadlessAppMain in core. With that said, the
configuration in the core logback.xml file is identical to the default
Logback configuration, so removing the file has no effect on the logging
output from BisqHeadlessAppMain, and furthermore, it appears that
BisqHeadlessAppMain is not actually in use anywhere right now (it was
added some months ago in anticipation of integrating the new HTTP API,
but this has yet to occur).

In a future change, we should probably return the core module to being a
pure library, without any main methods, and introduce a different
module, e.g. 'daemon' that contains the equivalent of
BisqHeadlessAppMain and whatever logback.xml configuration is
appropriate for it. Doing things this way will avoid the 'duplicate'
errors shown above, because desktop, seednode, etc will not depend on
core module, but will not depend on the daemon module.


ManfredKarrer (commented):
utACK

ManfredKarrer:
Thanks for cleaning that up!

PR #1955, opened 2018-11-23
Quell help output on option parsing error

cbeams:
This change suppresses help output when option parsing fails, e.g. due
to an unrecognized option being specified. This is in keeping with *nix
utility idioms; for example, notice what happens when running `git
--bogus` or `ls --bogus`: they output an error message, and in some
cases a usage message, but they do not print complete help text.
This approach is especially important in the case of an application like
Bisq, where there are many options and the help text is longer than a
typical screen, making it easy to miss the error message altogether, as
it is printed at the top of the screen.

This change also prints the error message to stderr vs stdout, which is
again in keeping with *nix utility idoms.


ManfredKarrer (approved):
utACK

PR #1956, opened 2018-11-23
Generate scripts for Bisq executables in root dir

cbeams:
This change configures the Gradle build to generate "start scripts" for
each Bisq executable (e.g. Bisq Desktop, Bisq Seednode, etc) in the root
project directory, such that after invoking `./gradle build`, the
following executable scripts become available:

    ~/Work/bisq-network/bisq
    $ ls -1 | egrep '(bisq*|lib)'
    bisq-desktop
    bisq-desktop.bat
    bisq-monitor
    bisq-monitor.bat
    bisq-relay
    bisq-relay.bat
    bisq-seednode
    bisq-seednode.bat
    bisq-statsnode
    bisq-statsnode.bat
    lib

This makes it possible for users (developers) to easily discover and use
these scripts in an idiomatic and platform-agnostic way as opposed to
the previous situation where we would advise users to run e.g.

    java -jar desktop/build/libs/desktop-0.8.0-SNAPSHOT-all.jar

This approach works, but is cumbersome and focuses unnecessarily on the
Java-based nature of the project. Now, with the changes in this commit,
the user would simply run:

    ./bisq-desktop

The 'lib' directory shown above contains all the jar files necessary to
construct classpaths for these various scripts. The 'cleanInstallDist'
task deletes the 'bisq-*' files and the 'lib' directory, and the default
'clean' task has been configured to depend on the 'cleanInstallDist'
task to ensure this cleanup happens automatically when most users would
expect it.

In the future, these same scripts can be used when installing Bisq
executables properly on users' systems via package managers like Brew
and Apt. The goal is to have the user experience around running
`bisq-desktop` (and more importantly, the forthcoming `bisqd`) be
similar in every way to installing and using `bitcoind`, `lnd` and other
idiomatic *nix-style utilities, be they Bitcoin-related or not.

See the changes in docs/build.md and docs/dev-setup.md for a further
sense of the how this change impacts the developer experience.


ManfredKarrer (approved):
utACK

PR #1959, opened 2018-11-24
Remove .bat scripts unless user OS is Windows

cbeams:
Previously, start scripts were generated for both *nix and Windows
platforms, resulting in an unnecessarily cluttered root directory.

With this change, both types of script are still generated, but Windows
.bat scripts are deleted immediately afterward if the user is running a
non-Windows OS (unfortunately, there was no clean way to suppress the
generation of these scripts in the Gradle StartScripts API).

See #1956


ManfredKarrer (approved):
utACK

PR #1964, opened 2018-11-25
Remove *nix scripts on Windows

devinbileck:
Related to https://github.com/bisq-network/bisq/pull/1959 which removed .bat files on a non-Windows OS to prevent a cluttered root directory, this change will delete *nix scripts on a Windows OS.


ManfredKarrer (approved):
utACK

PR #1974, opened 2018-11-26
Update store db files

ManfredKarrer:


ripcurlx (approved):
utACK - I'll update it again when I build the release

PR #1975, opened 2018-11-26
Remove color from Tor icon

ManfredKarrer:
The purple of the tor icon clashes hard with out green.
I converted to greyscale.
@pedro, @christoph: Feel free to change to an alternative if you don't
like it...


ripcurlx (approved):
utACK

PR #1979, opened 2018-11-27
Fix broken startScripts in $destinationDir/bin dirs

freimair:
A recent change in the build procedure produces broken startScripts in the respective `$destinationDir/bin`-directories.

### Issue
startScripts located in `$destinationDir/bin`-directories yield
```
Error: Could not find or load main class bisq.desktop.app.BisqAppMain
Caused by: java.lang.ClassNotFoundException: bisq.desktop.app.BisqAppMain
```
when executed.

### Cause
The startScripts get changed in order to be executed in the project root directory. However, they get changed before they are copied and thus, the original startScripts (still located in the `$destinationDir/bin`-directories) have a wrong path prefix for the classpath declaration (i.e. `.../bin/lib/...` instead of `.../bin/../lib/...`)

### Fix
First copy the startScripts to `rootProject.projectDir` and only then change them.

### Notes
- I believe that every startScript should work, regardless where it resides (even when its use is somewhat deprecated)
- I stumbled over the issue while fixing the [AUR bisq-git package](https://aur.archlinux.org/packages/bisq-git/) (doing my duty https://github.com/bisq-network/roles/issues/78)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -106,7 +106,7 @@ configure([project(':desktop'),
 
             // edit generated shell scripts such that they expect to be executed in the
             // project root dir as opposed to a 'bin' subdirectory
-            def windowsScriptFile = file("${rootProject.projectDir}/bisq-$applicationName.bat")
+            def windowsScriptFile = file("${rootProject.projectDir}/bisq-${applicationName}.bat")
             windowsScriptFile.text = windowsScriptFile.text.replace(
                 'set APP_HOME=%DIRNAME%..', 'set APP_HOME=%DIRNAME%')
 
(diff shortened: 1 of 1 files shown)


ripcurlx:
utACK - I'll leave it to @cbeams to ACK and merge.

cbeams (approved):
ACK. Thanks for the fix, @freimair. Please see the additional commit I pushed at 5c9eac9.

Nit: In the future, please favor the Git idiom of branch names separated with dashes (lisp case), e.g. `fix-broken-executables` vs. lowerCamelCase, e.g. `fixBrokenExecutables`.

freimair:
hm. I did not get this error when I tested fix locally. And Travis did not complain either. 

regarding lisp case: old habits. I will try :)

PR #1983, opened 2018-11-27
Send mobile notification msg in a thread.

ManfredKarrer:
The http request can take a few seconds to return so we use a thread to
send it and return the result in a handler.


ripcurlx (approved):
utACK

PR #1984, opened 2018-11-27
Remove block explorer request classes

ManfredKarrer:
Those have never been used. They were in place or requesting non-wallet
transactions like the trade fee tx of the peer.


ripcurlx (approved):
utACK

PR #1995, opened 2018-11-28
Blur: Add instructions for GUI Wallet in pop-up

ghost:
* Clarify existing instructions for CLI Wallet
* Provide instruction to retrieve information from GUI Wallet
* Ensure responsibilities are non-ambiguous with regard to burden of responsibility.
* Format line breaks properly


ManfredKarrer (approved):
utACK

PR #2003, opened 2018-11-30
Add install_java.sh

devinbileck:
Add a bash script for installing and configuring Java, based on [this file](https://github.com/bisq-network/bisq/blob/a762a6cc7c7eab6423b43a60f11e7ef698d7c236/doc/scripts/install_on_unix.sh), which was removed as part of restructuring done in https://github.com/bisq-network/bisq/pull/1924.

I have put it under a new scripts folder for lack of a better location.


ManfredKarrer (approved):
utACK

PR #2012, opened 2018-11-30
Make role details popup readable

sqrrm:
The info popup had alignment issues. This way it's readable but perhaps not conforming to the remaining UI components, not sure since I haven't followed that development closely.


ManfredKarrer (approved):
Quickly tested it and looks more or less ok. Layout has to be fixed in many places in the DAO. Will try to go over it after all the rest is completed with testing.
ACK

PR #2014, opened 2018-11-30
List IdaPay (IDA)

CaveSpectre11:
Official project URL: https://idapay.org
Official block explorer URL: http://explorer.idapay.org/


blabno (changes requested):
NACK unknown repository

Remove double newlines and squash commits into single one.

ManfredKarrer:
Open again once resolved

PR #2016, opened 2018-11-30
Bug fixes dao

ManfredKarrer:


desktop/src/main/java/bisq/desktop/main/dao/burnbsq/assetfee/AssetFeeView.java
```diff
@@ -164,20 +164,29 @@ protected void activate() {
...
-
-                if (!DevEnv.isDevMode()) {
-                    GUIUtil.showBsqFeeInfoPopup(listingFee, miningFee, txSize, bsqFormatter, btcFormatter,
-                            Res.get("dao.burnBsq.assetFee"), () -> doPublishFeeTx(transaction));
-                } else {
-                    doPublishFeeTx(transaction);
+            long days = getDays();
+            // We don't allow shorter periods as it would allow an attacker to try to deactivate other coins by making a
+            // small fee payment to reduce the trial period and look back period.
+            // Still not a perfect solution but should be good enough for now.
```
sqrrm, line 170: Good point about an attacked reducing the listing period of a coin. Perhaps we should consider not allowing a shorter period than the one initially chosen when doing a reactivation? It solves this problem and doesn't really reduce usability.

core/src/main/java/bisq/core/dao/governance/param/Param.java
```diff
@@ -110,6 +110,16 @@
     // Min required trade volume to not get de-listed. Check starts after trial period and use trial period afterwards to look back for trade activity.
     ASSET_MIN_VOLUME("0.01", ParamType.BTC, 10, 10),
 
+  /*  // TODO for testnet we want to have a short cycle of about a week
```
sqrrm, line 113: This only adds up to 5+ days. Still ok, but proposal phase could be extended to 600 blocks to make it 7 days.

sqrrm (approved):
utACK

desktop/src/main/java/bisq/desktop/main/dao/burnbsq/assetfee/AssetFeeView.java
ManfredKarrer: So then they only can use same or larger amounts in fee for future payments. We can change those parts anyway later as they are not really consensus part.

core/src/main/java/bisq/core/dao/governance/param/Param.java
ManfredKarrer: Prefer to keep it as short as possible. I think 2 days for important phases should be enough.

PR #2019, opened 2018-11-30
Fix log path

ManfredKarrer:
Resolving the property from the bisqEnvironment via getProperty returns
the default value instead of the program argument value. Not sure why,
seems to be related to recent changes in the property handling.

@freimair Could you have a look. Maybe you find out what is the reason why the getProperty call returns the incorrect value. You can reproduce it when starting a seed node. The log path is the the default Bisq directory. Feel free to revert my change if you found the cause...

### DIFF
--- a/core/src/main/java/bisq/core/setup/CoreSetup.java
+++ b/core/src/main/java/bisq/core/setup/CoreSetup.java
@@ -17,7 +17,6 @@
 
 package bisq.core.setup;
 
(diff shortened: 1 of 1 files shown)


ManfredKarrer:
I just saw that this happens also with the normal app, so I will merge that myself to get it in fast to avoid provblems with log mess up.... @freimair Feel free to revert it in a follow up PR if you find a better fix.

PR #2020, opened 2018-11-30
Use date of 0.5 release if no date for seed was specified

ManfredKarrer:
We had set the date to 0 which would lead to set the wallet date to the
genesis block. We had a "hard fork" with v0.5. so we can use that date
to ensure that the wallet birth date cannot be older.


ripcurlx (approved):
utACK

PR #2022, opened 2018-12-01
Use en_US if default locale has no country defined.

ManfredKarrer:
On some systems there is not country defined, in that case we use en_US


ripcurlx (approved):
utACK

PR #2024, opened 2018-12-01
Added script to install Java SE 10 target platform on OS rpm based

albpal:
Easy script to install the required java and perform ./gradlew build after that.

### DIFF
--- a/scripts/install_java_rpm-os_based.sh
+++ b/scripts/install_java_rpm-os_based.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+JAVA_HOME=/usr/lib/jvm/openjdk-10.0.2
+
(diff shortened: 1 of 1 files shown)


ManfredKarrer (approved):
utACK

ManfredKarrer:
@devinbileck Maybe you can have a look as well?

devinbileck:
utACK
Looks fine to me. However, with only one line that differs from the other script, it should really be handled within a single script - less duplication and maintenance required.

ManfredKarrer:
@devinbileck Agree. Would you be able to make that  merged scipt (after release)?

devinbileck:
@ManfredKarrer yep I can clean that up later.

ManfredKarrer:
@devinbileck Thanks. I will merge it in and later we consolidate those 2 files.

PR #2025, opened 2018-12-01
Issue 2021 - Asset validation override issues in many coin extensions

CaveSpectre11:
- Fix length issue in regular expression of validate() override in main asset coins.
    PIVX most likely started the 1+{25,34} that was then proliferated through the copies.
    While it likely should be 1+33, I could only confirm that 35 is too long, so the
    code is changed to be 1+{24,33} in order to reduce it by the presumed forgotten first
    character.
  - Fixup testInvalidAddresses() to better test regex rules in test asset coins.  The PIVX
    base code was testing against 3 addresses, that all had a different first character,
    therefore none of the rest of the regular expression was being tested.  Changed those
    addresses (and others for other susceptable coins) to accurately test first character
    match, length, and invalid characters in the 24 to 33 series.

### DIFF
--- a/assets/src/main/java/bisq/asset/coins/BitCloud.java
+++ b/assets/src/main/java/bisq/asset/coins/BitCloud.java
@@ -37,7 +37,7 @@ public class BitCloud extends Coin {
 
         @Override
         public AddressValidationResult validate(String address) {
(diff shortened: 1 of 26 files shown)


ManfredKarrer (approved):
utACK

ManfredKarrer:
Thanks, much better!

PR #2028, opened 2018-12-02
Do not use tor for btcj testnet

ManfredKarrer:


ManfredKarrer:
I merge it myself as i want to get the master merge done and that PR is rather trivial. As its sunday I dont expect other reviewers will review soon...

PR #2031, opened 2018-12-03
Normalize line endings

devinbileck:
Change line endings for desktop/src/main/resources/fonts/OFL.txt to LF
rather than CRLF for consistency.


ManfredKarrer (approved):
utACK

PR #2032, opened 2018-12-03
Handle cli-output_windows.txt with CRLF line endings

devinbileck:
On Windows, it is likely that cli-output_windows.txt will have CRLF line
endings since core.autocrlf is normally true which converts LF endings
to CRLF on checkout. This causes BisqHelpFormatterTest to fail since
the actual content it is compared against uses LF endings.

As a result, when loading the cli-output_windows.txt file normalize
line endings to LF.


ManfredKarrer (approved):
utACK

PR #2034, opened 2018-12-03
UI improvements for release

ripcurlx:


cbeams:
Before 083856a:

image


After 083856a:

image

@ripcurlx, can you explain why 083856a was committed? I would much rather revert it and keep the display of this information tight, not just in `Market->Spreads`, but in `Funds->Transactions` and probably other screens as well. IMO there is too much vertical whitespace now, forcing vertical scrolling where it wouldn't have been necessary in the past.

----

Also, I understand the motivation for centering address hyperlinks in 002f8f3 was to be consistent with the centered nature of the 'Address' column title, but the net effect of this change looks poor, IMO, particularly when there are multiple addresses in the list. I'd rather revert this change too, and leave the addresses left-aligned (cc @devinbileck who reported the issue originally).

ripcurlx:
@cbeams At row height: Because @pedromvpg wanted to increase the row height like that as he finds it better like that.

Regarding alignment: In the final design of Pedro everything will be left aligned anyways. I just didn't pushed it, as there were weird issues where the left alignment wasn't always exactly left aligned but rather one or to pixel left or right. I have to look into more detail what was going wrong.

PR #2036, opened 2018-12-03
Remove 32bit versions and update for Java 9+ support

ripcurlx:


desktop/package/macosx/finalize.sh
```diff
@@ -34,43 +33,26 @@ cp "deploy/Bisq-$version.jar.txt" "$target_dir/"
...
 echo Verify signatures
 gpg --digest-algo SHA256 --verify $dmg{.asc*,}
 gpg --digest-algo SHA256 --verify $deb64{.asc*,}
-gpg --digest-algo SHA256 --verify $deb32{.asc*,}
 gpg --digest-algo SHA256 --verify $exe64{.asc*,}
-gpg --digest-algo SHA256 --verify $exe32{.asc*,}
 
 #mkdir ../../build/vm/vm_shared_windows_32bit/$version
-cp -r . $win32/$version
+cp -r . $win64/$version
```
ManfredKarrer, line 56: Is that correct?

ManfredKarrer (approved):
utACK

Small comment with prob. unnecessary line...

desktop/package/macosx/finalize.sh
ripcurlx: As we don't have the win32 version anymore I copy everything to the win64 version where we'll scan it with an anti-virus.

PR #2039, opened 2018-12-04
Add script for easy installation and configuration of the necessary JDK on Windows

devinbileck:


ManfredKarrer (approved):
Is Java 10 accessible still without registering at the (buggy) Oracle page?

I am not experiences with Windows scripting, so maybe someone else can review as well...
Otherwise utACK from me.

ripcurlx:
utACK - @ManfredKarrer It is just for OpenJDK, so there shouldn't be an issue with Oracle.

PR #2055, opened 2018-12-04
Consolidate install_java.sh scripts

devinbileck:
Rather than having a separate install_java.sh script for distributions
that use a different package manger, I updated install_java.sh to
handle package managers used by various Linux distributions.


ManfredKarrer (approved):
utACK

PR #2056, opened 2018-12-05
Dao fix reorg issues

ManfredKarrer:


core/src/main/java/bisq/core/dao/node/BsqNode.java
```diff
@@ -179,19 +189,87 @@ protected void onParseBlockChainComplete() {
...
+                pendingBlocks.remove(rawBlock);
+
+            // After parsing we check if we have pending blocks we might have received earlier but which have been
+            // not connecting from the latest height we had. The list is sorted by height
+            if (!pendingBlocks.isEmpty()) {
+                // To avoid ConcurrentModificationException we copy the list. It might be altered in the method call
+                ArrayList<RawBlock> tempPendingBlocks = new ArrayList<>(pendingBlocks);
+                for (RawBlock tempPendingBlock : tempPendingBlocks) {
+                    try {
+                        doParseBlock(tempPendingBlock);
```
sqrrm, line 222: This recursion seems a bit strange, it has to copy pendingBlocks into a temporary array for each recursion level, probably ok but a bit wasteful.

core/src/main/java/bisq/core/dao/node/BsqNode.java
```diff
@@ -179,19 +189,87 @@ protected void onParseBlockChainComplete() {
...
+                // We received an older block. We compare if we have it in our chain.
+                Optional<Block> optionalBlock = daoStateService.getBlockAtHeight(rawBlock.getHeight());
+                if (optionalBlock.isPresent()) {
+                    if (optionalBlock.get().getHash().equals(rawBlock.getPreviousBlockHash())) {
+                        log.info("We received an old block we have already parsed and added. We ignore it.");
+                    } else {
+                        log.info("We received an old block with a different hash. We ignore it. Hash={}", rawBlock.getHash());
+                    }
+                } else {
+                    log.info("In case we have reset from genesis height we would not find the block");
```
sqrrm, line 252:
```suggestion
                    log.info("Older block received but nothing found at that height, this is normal if there was a reset from genesis height");
```

core/src/main/java/bisq/core/dao/node/BsqNode.java
```diff
@@ -179,19 +189,87 @@ protected void onParseBlockChainComplete() {
...
+
+            // After parsing we check if we have pending blocks we might have received earlier but which have been
+            // not connecting from the latest height we had. The list is sorted by height
+            if (!pendingBlocks.isEmpty()) {
+                // To avoid ConcurrentModificationException we copy the list. It might be altered in the method call
+                ArrayList<RawBlock> tempPendingBlocks = new ArrayList<>(pendingBlocks);
+                for (RawBlock tempPendingBlock : tempPendingBlocks) {
+                    try {
+                        doParseBlock(tempPendingBlock);
+                    } catch (RequiredReorgFromSnapshotException e1) {
```
sqrrm, line 223: Catching the RequiredReorgFromSnapshotException here and braking will stop this loop but if there was a loop in a previous level of recursion that would continue running and parsing in a now known bad state. I think the exception has to be rethrown here.

core/src/main/java/bisq/core/dao/node/full/FullNode.java
```diff
@@ -245,16 +240,13 @@ private void parseBlock(int blockHeight, int chainHeight,
...
         log.error(errorMessage);
-
-        if (throwable instanceof BlockNotConnectingException) {
-            startReOrgFromLastSnapshot();
-        } else if (throwable instanceof RpcException &&
+        if (throwable instanceof RpcException &&
                 throwable.getCause() != null &&
                 throwable.getCause() instanceof HttpLayerException &&
                 ((HttpLayerException) throwable.getCause()).getCode() == 1004004) {
             errorMessage = "You have configured Bisq to run as DAO full node but there is not " +
```
sqrrm, line 247:
```suggestion
            errorMessage = "You have configured Bisq to run as DAO full node but there is no " +
```

ripcurlx:
utACK

sqrrm (commented):
Please have a look at my comments on recursion

core/src/main/java/bisq/core/dao/node/BsqNode.java
sqrrm: Tested a bit and did a 6 block reorg by generating 10 blocks, invalidating the fourth of those blocks and generating 10 more blocks. Results in almost 8200 useless iterations.

It also seems to have broken the phase/cycle handling. Created an issue about it, shouldn't have since it's probably due to the reorg handling, #2064

core/src/main/java/bisq/core/dao/node/BsqNode.java
ManfredKarrer: Yes right I think that is wrong. It would iterate multiple times with lists which have been already applied. E.g. if you have 5 items and first recursive call copies the list then removes the current block from the source list and then copy the list again and start another loop with 4 itmes. Once that recursion is done the parent recursion would still iterate the remaining elements but those have been already processed. So either we add a check there if the block is still in the source list or we dont use a loop but only call with the first element (list is sorted) and so we avoid the remaining iterations from the current loop. 
What do you think?

core/src/main/java/bisq/core/dao/node/BsqNode.java
ManfredKarrer: Good point about the RequiredReorgFromSnapshotException. I think if we use the first item instead of loop it should be ok. I will look into the other issue. Have not tested how it effect other aspects. We should make a client with all features used (requests, voting, bonds,...) and then apply  a reorg to see if it break anything. Might be valid result if stuff become invalid by deep reorgs - e.g. < break, with < break it should not change results.

ManfredKarrer:
I will merge as it is at least much better as current state. But will start a new PR with the recursion issue and we need to test more all other aspects with reorgs.

PR #2062, opened 2018-12-05
Left align columns, except last one

ripcurlx:
Adapted the left align style for all columns except the last one based on the design of @pedromvpg. I saw that for avatars and icons that are in the last column they were centered, so I implemented it like that. @pedromvpg Could you please have a look if it is as you wanted it to have?

Maybe we could discuss in this PR also the line height of rows in certain screens. It came up during the release that some are not very happy with the line height as it will show less information now than it did before. 

The question is f the user has the requirement to get everything in one screen or if the extended height of the rows is more pleasant to watch.

As this is an design issue @pedromvpg has the last call, but it seems there should be some public discussion on this.

bildschirmfoto 2018-12-05 um 14 45 02
bildschirmfoto 2018-12-05 um 14 44 58
bildschirmfoto 2018-12-05 um 14 44 50
bildschirmfoto 2018-12-05 um 14 43 36
bildschirmfoto 2018-12-05 um 14 43 32
bildschirmfoto 2018-12-05 um 14 43 27
bildschirmfoto 2018-12-05 um 14 42 49
bildschirmfoto 2018-12-05 um 14 41 36
bildschirmfoto 2018-12-05 um 14 36 09
bildschirmfoto 2018-12-05 um 14 36 01

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -193,6 +193,9 @@ shared.all=All
 shared.edit=Edit
 shared.advancedOptions=Advanced options
 shared.interval=Interval
+shared.actions=Actions
+shared.buyerUpperCase=Buyer
+shared.sellerUpperCase=Seller
 
 ####################################################################
 # UI views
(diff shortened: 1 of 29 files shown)


ripcurlx:
☝️ @cbeams @ManfredKarrer

ManfredKarrer:
I think those colums with buttons (Select) or checkboxes should be centered still or the header alignment aligned with the button text. 
In the first screen there is still one centered.
The confirmation column could be centered as well IMO. 
Those with the info icon on the right might be more clean if that is right aligned.

But no strong opinions here...

ManfredKarrer (approved):
utACK

PR #1961, opened 2018-11-24
Redesign CLI help output

cbeams:
Prior to this pull request, Bisq's `--help` output looked like this:

    $ bisq-desktop --help
    Option                                  Description
    ------                                  -----------
    --appDataDir <String>                   Application data directory (default:
                                              /Users/cbeams/Library/Application
                                              Support/Bisq)
    --appName <String>                      Application name (default: Bisq)
    --banList <String>                      Nodes to exclude from network
                                              connections. (default: )
    --baseCurrencyNetwork <String>          Base currency network (default:
                                              BTC_MAINNET)
    --bitcoinRegtestHost <RegTestHost>      (default: LOCALHOST)
    --btcNodes <String>                     Custom nodes used for BitcoinJ as
                                              comma separated IP addresses.
                                              (default: )
    [...]

Now it reads as follows:

    $ bisq-desktop --help
    Bisq Desktop version 0.8.0

    Usage: bisq-desktop [options]

    Options:

      --help
            This help text

      --logLevel=<OFF|ALL|ERROR|WARN|INFO|DEBUG|TRACE> (default: INFO)
            Log level

      --seedNodes=<host:port[,...]>
            Override hard coded seed nodes as comma separated list e.g.
            'rxdkppp3vicnbgqt.onion:8002,mfla72c4igh5ta2t.onion:8002'

      --myAddress=<host:port>
            My own onion address (used for bootstrap nodes to exclude itself)

      --banList=<host:port[,...]>
            Nodes to exclude from network connections.

      --nodePort=<Integer> (default: 9999)
            Port to listen on

      --useLocalhostForP2P=<Boolean> (default: false)
            Use localhost P2P network for development

      [...]

This new help format is modeled after bitcoind's own help output, which reads as follows:

    $ bitcoind --help | head -20
    Bitcoin Core Daemon version v0.17.0.0-ge1ed37edaedc85b8c3468bd9a726046344036243

    Usage:  bitcoind [options]                     Start Bitcoin Core Daemon

    Options:

      -?
           Print this help message and exit

      -alertnotify=<cmd>
           Execute command when a relevant alert is received or we see a really
           long fork (%s in cmd is replaced by message)

      -assumevalid=<hex>
           If this block is in the chain assume that it and its ancestors are valid
           and potentially skip their script verification (0 to verify all,
           default:
           0000000000000000002e63058c023a9a1de233554f28c7b21380b6c9003f36a8,
           testnet:
[...]

PR #2063, opened 2018-12-05
Fix base currency option handling

### LINKED ISSUE #2048: Changing BTC network does not have any effect
If user changes the BTC network in settings and restarts MAINNET will be selected again even if the user has changed to testnet. There is probably some bug in the option parsing. @freimair will look into it.
For now users who want to try out the testnet DAO need to add a program argument: 
'--baseCurrencyNetwork=BTC_TESTNET ' to enable testnet.

cbeams:
This change fixes #2048 by removing the assignment of a default value
for the `baseCurrencyNetwork` option at the level of the command line
option parser. The assignment of this default was an oversight in #1961
(specifically commit 83e1dd3) that did not account for the fact that
users can change the `baseCurrencyNetwork` value via the Settings screen
in the application. When users change the setting in the application, the new
value is persisted to <appDataDir>/bisq.properties, which is handled at
runtime as a PropertySource with lower precedence than the command line
property source, which means that the changed value is never picked up
because the higher-precedence command line PropertySource always has a
default value.

This fix is surgical in that it addresses only this specific option. A
subsequent change should address the more general issue that setting
defaults in the command line option parser always precludes the
possibility of overriding them in bisq.properties. Basically, we should
revert to the previous strategy of reporting what the default value will
be in the help text without actually assigning a default value in the
option parser using the `defaultsTo` method.

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqExecutable.java
+++ b/core/src/main/java/bisq/core/app/BisqExecutable.java
@@ -68,7 +68,11 @@ import lombok.extern.slf4j.Slf4j;
 
 import static bisq.core.app.BisqEnvironment.DEFAULT_APP_NAME;
 import static bisq.core.app.BisqEnvironment.DEFAULT_USER_DATA_DIR;
+import static bisq.core.btc.BaseCurrencyNetwork.BTC_MAINNET;
+import static bisq.core.btc.BaseCurrencyNetwork.BTC_REGTEST;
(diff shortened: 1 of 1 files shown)


ManfredKarrer (approved):
utACK

ManfredKarrer:
Thanks @cbeams for fixing it so fast! @freimair I merge it already so that users who run from source can this fixed. But please have a look if we need to change that for the other options as well. At least with the path for log files there was an issue as well (which is fixed by not using getProperty).

PR #2067, opened 2018-12-05
Use only first block in pendingBlocks instead of iteration

ManfredKarrer:
Do not merge yet as more will be added.

@sqrrm Just added that PR for sharing progress on improvements. Needs more testings, but would be good if you could base your testing on that PR.


ManfredKarrer:
I will merge that now as I detected a problem in the current master which this PR should avoid.

sqrrm:
Sorry for slow review.

This looks much better but it still breaks the phase handling. I don't see the big recursions in the log, naturally, so clearly better than before.

ManfredKarrer:
Yes I have not looked into the phases yet...

PR #2070, opened 2018-12-06
Update checkpoint files

ManfredKarrer:
We used checkpoint files from BitcoinJ 0.14.4 and those caused a very
slow initial download. I tested with the WalletAppKit and different
release version sof BitcoinJ and could reproduce the behaviour there.
WIth v0.14.5 it was much better. But also that caused slow startup in
Bisq. After creating a current checkpoint file and using that the
download is now very fast.

Also added logging for versions of BitcoinJ and libDohJ so that we have
better control which version is used.


gradle/witness/gradle-witness.gradle
```diff
@@ -12,6 +12,9 @@
 
 dependencyVerification {
     verify = [
+        'bisq:core:4d0e437026e53968298710acd8bc2e4b1b0e1d37859c69359cf63e073ae02b86',
+        'bisq:p2p:c36dde32292ca7335ea594e0cf36f8f766cd3689cd3886b095def3ff946ed60b',
+        'bisq:common:d9fef12428771154ca70d9299a36feb1d855b07242822a96a16f43260af03149',
         'org.controlsfx:controlsfx:b98f1c9507c05600f80323674b33d15674926c71b0116f70085b62bdacf1e573',
```
ripcurlx, line 15: I think the checksums for core, p2p and common were introduced by mistake. They are unnecessary and cause the build to fail.

gradle/witness/gradle-witness.gradle
```diff
@@ -20,28 +23,36 @@ dependencyVerification {
...
+        'org.openjfx:javafx-fxml:ef5b10aa0b985e3bece34629bcaf4f462d3a60e1139a83edd116f69c1a36e0e8',
+        'org.openjfx:javafx-controls:689626346d7b390a46c7acd7741af19f91567110a6611cad50c60d8c611ac59f',
+        'org.openjfx:javafx-swing:885a5ac37b82913dbfe2ed0742315b6275e8137f19b1135893cae42c4483591b',
+        'com.jfoenix:jfoenix:4739e37a05e67c3bc9d5b391a1b93717b5a48fa872992616b0964d3f827f8fe6',
         'com.github.JesusMcCloud.netlayer:tor.native:f1bf0096f9eb6020645a65d91aa530d15aef97e69cc5a79d7b2405421f74700a',
         'com.github.JesusMcCloud.netlayer:tor.external:cfba681398c191a1906d6d023a3be28a8fa9b1f4eee52e966daf7b1ae630414f',
         'org.apache.httpcomponents:httpclient:db3d1b6c2d6a5e5ad47577ad61854e2f0e0936199b8e05eb541ed52349263135',
         'net.sf.jopt-simple:jopt-simple:6f45c00908265947c39221035250024f2caec9a15c1c8cf553ebeecee289f342',
         'org.fxmisc.easybind:easybind:666af296dda6de68751668a62661571b5238ac6f1c07c8a204fc6f902b222aaf',
+        'bisq:assets:121fd110d20b451203ec3cdfcb23682c9e1a3aa3d1e7f0158c581ad20b865b6c',
```
ripcurlx: Also this needs to be removed

ripcurlx (changes requested):
Please see my comments.

ripcurlx:
It seems to be there is still a problem with the platform specific openjfx libs. It works for me locally as I also use macOS, but it fails on our linux build server. I guess we have to do a platform dependent checksum test to cover also the openjfx libs.

gradle/witness/gradle-witness.gradle
```diff
@@ -20,6 +20,10 @@ dependencyVerification {
         'de.jensd:fontawesomefx-commons:5539bb3335ecb822dbf928546f57766eeb9f1516cc1417a064b5709629612149',
         'com.googlecode.jcsv:jcsv:73ca7d715e90c8d2c2635cc284543b038245a34f70790660ed590e157b8714a2',
         'com.github.sarxos:webcam-capture:d960b7ea8ec3ddf2df0725ef214c3fccc9699ea7772df37f544e1f8e4fd665f6',
+        'org.openjfx:javafx-fxml:ef5b10aa0b985e3bece34629bcaf4f462d3a60e1139a83edd116f69c1a36e0e8',
```
devinbileck: I recall encountering issues before when we added the openjfx dependencies since they are OS specific. It may work for you on Mac, but will likely fail on Windows (and probably Linux).

gradle/witness/gradle-witness.gradle
cbeams: Indeed. https://github.com/bisq-network/bisq/blob/9389a8e1fa5005d16771116599da9b5da13ad9ce/gradle/witness/gradle-witness.gradle#L6 should be changed to exclude `bisq:` instead of `network.bisq` libraries.

So it should be change to `... | grep -v 'bisq:'`

ManfredKarrer: I use now:
`./gradlew -q calculateChecksums | grep -v 'bisq:\|org.openjfx:' >> gradle/witness/gradle-witness.gradle`

That filters only out the local bisq projectes but not bitcoinj and others as well org.openjfx libs which cause issues.

sqrrm:
utACK

PR #2074, opened 2018-12-06
Don't validate display fields in proposal display

sqrrm:
When displaying fields they shouldn't run validation as they're non-editable and populated from external data.


ManfredKarrer (approved):
utACK

Issue #2068, opened 2018-12-05
Wrong wording in dropdown in markets trades


ManfredKarrer:
Shows "offers" instead of "trades"
<img width="415" alt="screen shot 2018-12-05 at 20 38 39" src="https://user-images.githubusercontent.com/1449498/49539405-dcba9680-f8cd-11e8-9c2c-050a8334ca4a.png">

PR #2077, opened 2018-12-06
Fix wording and close window without action

### LINKED ISSUE #2051: Cancel of a dialog box during setup password handled incorrectly
As a new user, I want to setup a wallet password to make my wallet encrypted

Clicking on “Set Password” button I get the following dialog box:
image3

After pressing ESC, the dialog accepts a NOT recommended choice and the new password set up

EXPECTED BEHAVIOR:  
Pressing ESC should just close a dialog window without any action applied

ripcurlx:
Fixes #2051 and #2068.


ManfredKarrer (approved):
utACK

PR #2080, opened 2018-12-06
Take offer refactoring

blabno:


ManfredKarrer:
Close as discussed by phone. Will wait for a PR for core and desktop only code.

PR #2081, opened 2018-12-06
Stream isolation for Tor

### LINKED ISSUE #731: Tor stream isolation
As far as I can tell, Bitsquare is not using the stream isolation feature of Tor.  Stream isolation separates different identities' TCP connections into separate Tor circuits, which makes it more difficult for wiretappers to correlate identities.  It would be great to see Bitsquare implement support for stream isolation to improve privacy.

For some context, please see [the IRC conversation from today](https://botbot.me/freenode/bitsquare/2017-01-23/?msg=79797161&page=1).  The brief upshot is that to implement stream isolation, Bitsquare would need to supply a SOCKS5 username/password that uniquely corresponds to each identity that should stay separated to maintain privacy.  For example, maybe the SOCKS5 username/password could correspond to a given trade order.  (But please do read the IRC conversation for the full context.)

Cheers.

freimair:
closes #731 

Adds a command line parameter `--torStreamIsolation`. If set, Bisq uses a new stream for every request to the Tor network.

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqEnvironment.java
+++ b/core/src/main/java/bisq/core/app/BisqEnvironment.java
@@ -206,7 +206,7 @@ public class BisqEnvironment extends StandardEnvironment {
             torRcFile, torRcOptions, externalTorControlPort, externalTorPassword, externalTorCookieFile,
             socks5ProxyHttpAddress, useAllProvidedNodes, numConnectionForBtc, genesisTxId, genesisBlockHeight, referralId, daoActivated;
 
-    protected final boolean externalTorUseSafeCookieAuthentication;
+    protected final boolean externalTorUseSafeCookieAuthentication, torStreamIsolation;
 
     public BisqEnvironment(OptionSet options) {
         this(new JOptCommandLinePropertySource(BISQ_COMMANDLINE_PROPERTY_SOURCE_NAME, checkNotNull(
@@ -295,6 +295,9 @@ public class BisqEnvironment extends StandardEnvironment {
                 (String) commandLineProperties.getProperty(NetworkOptionKeys.EXTERNAL_TOR_COOKIE_FILE) :
                 "";
         externalTorUseSafeCookieAuthentication = commandLineProperties.containsProperty(NetworkOptionKeys.EXTERNAL_TOR_USE_SAFECOOKIE) ?
+                true :
+                false;
+        torStreamIsolation = commandLineProperties.containsProperty(NetworkOptionKeys.TOR_STREAM_ISOLATION) ?
                 true :
                 false;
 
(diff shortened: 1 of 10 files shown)


p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
```diff
@@ -115,7 +117,17 @@ protected Socket createSocket(NodeAddress peerNodeAddress) throws IOException {
...
         try {
             tor = Tor.getDefault();
-            return tor != null ? tor.getProxy() : null;
+
+            String stream = "";
+            if (streamIsolation) {
+                // create a random string
+                byte[] bytes = new byte[512]; // note that getProxy does Sha256 that string anyways
+                new SecureRandom().nextBytes(bytes);
+                stream = new String(bytes);
```
ManfredKarrer: Could there be any issues to use plain bytes here? Maybe more safe to use hex?

core/src/main/java/bisq/core/app/BisqEnvironment.java
```diff
@@ -206,7 +206,7 @@ public static boolean isDaoActivated(Environment environment) {
             torRcFile, torRcOptions, externalTorControlPort, externalTorPassword, externalTorCookieFile,
             socks5ProxyHttpAddress, useAllProvidedNodes, numConnectionForBtc, genesisTxId, genesisBlockHeight, referralId, daoActivated;
 
-    protected final boolean externalTorUseSafeCookieAuthentication;
+    protected final boolean externalTorUseSafeCookieAuthentication, torStreamIsolation;
```
ManfredKarrer, line 209: Should torStreamIsolation be by default turned on (true)?

p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
freimair: Netlayer does a `stream.getBytes()` immediately, so there should be no issue. Anyhow, the `stream` now gets a Base64-encoded `bytes`.

core/src/main/java/bisq/core/app/BisqEnvironment.java
freimair: I have been thinking about that as well.
- However, in the past, we had some issues with `SecureRandom` on Linux taking an extraordinary amount of time to produce its first output (> 2 minutes) which could only be sped up by changing something in the OS.
- Furthermore, creating a random string and a new Socket every time takes time.
- Last but not least, I am not entirely sure if changing the socket everytime does effect network performance to the worse.

I added an "experimental!" to the option so we can life the `--streamIsolation` for a little while and see if we encounter any unforeseen issues. If, one day, we are convinced that everything is fine, we can use the feature by default.

Or should we do it the other way around and provide an off-switch to the feature?

ManfredKarrer (approved):
utACK

core/src/main/java/bisq/core/app/BisqEnvironment.java
ManfredKarrer: Yes, good points. Is SecureRandom really needed at all here? I have not followed closely the context and motivation behind it but to keep it turned off by default is definitely better.

PR #2085, opened 2018-12-06
Update install_java.bat

devinbileck:
If the JDK is already installed, there is no need to download,
extract, and install it again. Just need to ensure the environment
variables are set.


ManfredKarrer (approved):
utACK

PR #2089, opened 2018-12-07
HTTP API with take offer refactoring

blabno:
Same as https://github.com/bisq-network/bisq/pull/2080 but in a single commit.


ManfredKarrer:
Close as discussed by phone. Will wait for a PR for core and desktop only code.

PR #2096, opened 2018-12-08
Remove BCH

ManfredKarrer:
Applying the proposal "Remove BCH from Bisq"
(https://github.com/bisq-network/proposals/issues/61)
There have been 44 upvotes and 0 downvotes so a clear signal that the
Bisq community does not want to have BCH listed on Bisq.


sqrrm (approved):
utACK

Issue #2037, opened 2018-12-03
Update from 0.8.0 to 0.9.0 new bisq doesn't start at all


Onefox:
Well, after I downloaded and updated to v0.9.0 the program doesn't even start any more...
Mouse cursor starts spinning and nothing happens, no process is running.
I have Windows 8.1 Pro and v0.8.0 was working fine...
No AV running.
Are there any log files anywhere that I can check?

I uninstalled and reinstalled it and rebooted windows.

ManfredKarrer:
There are some know issues with windows.

Can you try to backup your data directory and then delete  the app data directory or uninstall the Bisq app. Then re-install again. If you uninstall the data directory gets deleted as well, so if you used Bisq before you need to move over your backup directoy.

Another reason can be AV software. Some are deleting files from the installation or data directory.

Another is that at installation time the AV software stops Tor and then there is a pending locked file which cannot be deleted by another process. An OS restart resolves that.

All those issues are known since a while but happen not often. If you are a developer and experienced with Windows, maybe you can help to get those fixed.

ManfredKarrer:
Yes log file is in data directory.

Onefox:
I had no AV running.

The log file only shows the download process and than a normal looking shutdown. (the bisq.log file in the roaming bisq folder).

After I cleared the roaming folder and started bisq It worked again. Than I stopped it and removed the new generated btc_mainnet folder and copied my backup into place and got all data back. So its running again now.

Now its working again...
Is there anything that could help analyse the behavior?

ps.: sadly I don't have experience with developing for windows :/

ManfredKarrer:
Ah good! We did not had time to dive deeper to investigate.... Most Bisq devs don't use Windows...

sethforprivacy:
I'm also having this issue -- cannot open 0.9.0 at all after updating. Here is the output of the log file after downloading the new version and attempting to install it:

https://paste.fedoraproject.org/paste/vTCwMuRlyZVy2kLjRfi~ig

ManfredKarrer:
@GoodEn0ugh Whcih OS? Can u post full logs?

sethforprivacy:
W10 -- I backed up C:\Users\<username>\AppData\Roaming\Bisq, uninstalled Bisq, and reinstalled, and it started right up.

sethforprivacy:
[bisq.log](https://github.com/bisq-network/bisq/files/2641766/bisq.log)

izzytdi:
[...]

PR #2103, opened 2018-12-10
Fix Windows installer issues

### LINKED ISSUE #2066: Improve Windows installer so it deletes old app directory
We got a lot of reports with that release that users cannot run the new version. After deleting the app directory (.../Local/Bisq) it works. We got those issues also in the past but less often. I assume as with Java 10 we have a different runtime that this cause imcompatibility with almost all Windows users old installation.

Is there a way to configure in the java packager that the installer runs a pre-install script where we rename the app directory to Bisq_0.8 or the like to get rid of that issue? Or we simply set the property which is used to name the Bisq directory to Bisq_0.9.0, so with every version that directory gets newly created. Would avoid future issues as well. That might be the easiest way to deal with it.

@devinbileck Any idea?

ripcurlx:
Fixes #2066, #2037


desktop/package/windows/Bisq.iss
```diff
@@ -64,11 +65,24 @@ begin
...
+end;
+
+
 function InitializeSetup(): Boolean;
 begin
 // Possible future improvements:
 //   if version less or same => just launch app
 //   if upgrade => check if same app is running and wait for it to exit
 //   Add pack200/unpack200 support?
+  DeleteOldAppDataDirectory;
```
devinbileck: Slight problem with putting it here. It removes the folder prior to starting the install. So if you cancel the installer, your Bisq app folder will be gone. I think it might need to be done in `PrepareToInstall()` method?

devinbileck (changes requested):

desktop/package/windows/Bisq.iss
devinbileck: Or you may want to take a look at using an [InstallDelete] section: http://www.jrsoftware.org/ishelp/topic_installdeletesection.htm

ripcurlx: Good point - I've switched to the ```PrepareToInstall()``` trigger to prevent the deletion before the installer is actually run.

ripcurlx:
@devinbileck I've created a new installer (including tor cleanup) If you want to give it a try again: https://www.dropbox.com/s/rdecl0zyla7mr3q/Bisq-0.9.0-installer-with-tor-cleanup.exe?dl=0

desktop/package/windows/Bisq.iss
```diff
@@ -64,6 +65,96 @@ begin
...
+  hiddenServiceBackupDir := mainnetDir + '\hiddenservice_backup';
+  if DirExists(hiddenServiceDir) then begin
+   if DirExists(hiddenServiceBackupDir) then begin
+     DelTree(hiddenServiceBackupDir, true, true, true);
+   end
+   CreateDir(hiddenServiceBackupDir);
+   DirectoryCopy(hiddenServiceDir, hiddenServiceBackupDir);
+   DelTree(torDir, false, true, true);
+   CreateDir(hiddenServiceDir);
+   DirectoryCopy(hiddenServiceBackupDir, hiddenServiceDir);
```
devinbileck, line 147: Should we remove hiddenServiceBackupDir once we are done with it?

desktop/package/windows/Bisq.iss
```diff
@@ -64,6 +65,96 @@ begin
...
+    Log(Format('Failed to list %s', [SourcePath]));
+  end;
+end;
+
+//Delete old app directory to prevent issues during update
+procedure DeleteOldAppDataDirectory;
+var
+  entry: String;
+begin
+  entry := ExpandConstant('{localappdata}') + '\Bisq\';
```
devinbileck, line 122: Is {localappdata} correct? I don't see it purging the app folder anymore with the latest exe, like it did in the previous exe. Did you use {app} previously?

desktop/package/windows/Bisq.iss
ripcurlx: I left it there by intent, just in case there is an issue and someone has to recover their hiddenservices manually.

devinbileck (changes requested):
NACK
Tor cleanup works, but might want to remove the backup once done with it. And this time I don't see the app folder being purged.

desktop/package/windows/Bisq.iss
ripcurlx: hmm - worked for me. I didn't change anything there. I put a test file in the app folder in it was gone afterwards. I can have a look again tomorrow. How did you test it?

devinbileck: My apologies, it is deleting it. I confirmed using process monitor.
Previously there was a noticeable delay between the purge (when done in InitializeSetup) and install, that I could see the folder being removed. But now that the purge is done in PrepareToInstall, it is not noticeable.

devinbileck (approved):
ACK

PR #2104, opened 2018-12-10
Refine various ui strings

m52go:
When I installed a fresh copy of Bisq to test the DAO, some of the text in the UI looked like it could use some edits.

This PR includes edits to a variety of text, beyond what I saw, but I only made significant changes (e.g., adding a line break) to elements that I came across myself which I knew for sure could handle them.

There were a couple places I wanted to add a new line break, but wasn't sure if the UI could handle it, so I passed (in order to avoid creating any other issues).

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -807,7 +807,7 @@ support.buyerOfferer=BTC buyer/Maker
 support.sellerOfferer=BTC seller/Maker
 support.buyerTaker=BTC buyer/Taker
 support.sellerTaker=BTC seller/Taker
-support.backgroundInfo=Bisq is not a company and does not guarantee customer support.\n\n\
+support.backgroundInfo=Bisq is not a company, so it handles disputes differently.\n\n\
 If there are disputes in the trade process (e.g. one trader does not follow the trade protocol) \
 the application will display an \"Open dispute\" button after the trade period is over \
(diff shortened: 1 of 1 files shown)


core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -807,26 +807,26 @@ support.buyerOfferer=BTC buyer/Maker
 support.sellerOfferer=BTC seller/Maker
 support.buyerTaker=BTC buyer/Taker
 support.sellerTaker=BTC seller/Taker
-support.backgroundInfo=Bisq is not a company and does not provide any kind of customer support.\n\n\
+support.backgroundInfo=Bisq is not a company and does not guarantee customer support.\n\n\
```
ManfredKarrer: Maybe we can rewrite that. It sounds a bit harsh to say there is no support. In fact our support is better than in many companies. its just nto company provided but community driven....

ManfredKarrer (approved):
utACK

Maybe you can push another small change to that issue with "support" as mentioned?

core/src/main/resources/i18n/displayStrings.properties
m52go: Ha yes, I thought the same thing, but assumed it was on purpose for legal reasons so I didn't change it. Pushing an update now.

ripcurlx (approved):
utACK

PR #2108, opened 2018-12-11
Fix missing render update issue

ManfredKarrer:


ripcurlx (approved):
ACK - wasn't able to reproduce it anymore locally with this PR 👍

PR #2109, opened 2018-12-11
Minor DAO ui fixes

ripcurlx:


desktop/src/main/java/bisq/desktop/components/MenuItem.java
```diff
@@ -58,7 +58,7 @@ public MenuItem(Navigation navigation,
         this.viewClass = viewClass;
         this.baseNavPath = baseNavPath;
 
-        setLabelText(title);
+        setLabelText(title.toUpperCase());
```
sqrrm: This doesn't handle the 'Vote on proposals'/'Browse open proposals' text that changes depending on phase

sqrrm (changes requested):
NACK

Need to change all labels to upper case

desktop/src/main/java/bisq/desktop/components/MenuItem.java
sqrrm: Could be fixed in GovernanceView.java:90/92

ripcurlx: Thanks for pointing it out. Just fixed it by converting to uppercase also on label update.

ripcurlx: And I removed now an unnecessary uppercasing. Should have looked through the code more carefully 🙄

sqrrm (approved):
ACK

Of course, much better to handle it there :)

desktop/src/main/java/bisq/desktop/components/MenuItem.java
sqrrm: Yes, better to only set it inside setLabelText as you do now

Issue #2053, opened 2018-12-04
Dialog message is shown twice when switching from Password/Seed states


Hashedhash:
The following scenario is used: 
- Input a password and click “Set Password” button

![image3](https://user-images.githubusercontent.com/8158351/49452069-2a110800-f7e9-11e8-8f4f-fe6a3adba41e.png)

- In the dialog window (“It is highly recommended to write down your seed words before setting a password” click “Go To Account/Wallet Seed”
- When the page is loaded, a dialog window appears and says: “You have not setup a wallet password, which would protect the display of the seed words”

![image1](https://user-images.githubusercontent.com/8158351/49452107-37c68d80-f7e9-11e8-959e-3e106fb93557.png)

- Click NO
- The dialog appears once again
 
EXPECTED BEHAVIOR
The dialog window should be shown only once

Issue #2042, opened 2018-12-04
Fiat amount is not rounded in dispute details


devinbileck:
The fiat amount is rounded in the trade details for maker/taker, but not in dispute details for arbitrator.
![image](https://user-images.githubusercontent.com/603793/49424214-cafeb500-f74e-11e8-92f4-9791b9256016.png)

PR #2111, opened 2018-12-11
Different ui bug fixes

### LINKED ISSUE #2044: Issue with input validation in create offer view
When creating an offer, after entering amount of BTC to buy I put the cursor in the minimum amount of BTC box and the USD amount box showed error text but not below % box. If I then click in the below % box the values get filled in and the error text remains.
image

ripcurlx:
Fixes #2044, #2042 and #2053.


ManfredKarrer (approved):
utACK

PR #2117, opened 2018-12-12
Fix reorg issues with phases

ManfredKarrer:


core/src/main/java/bisq/core/dao/governance/period/CycleService.java
```diff
@@ -96,6 +96,28 @@ public void onParseBlockChainComplete() {
     // API
     ///////////////////////////////////////////////////////////////////////////////////////////
 
+    public void addFirstCycle() {
+        daoStateService.getCycles().add(getFirstCycle());
+    }
+
+    public int getCycleIndex(Cycle cycle) {
+        return (cycle.getHeightOfFirstBlock() - genesisBlockHeight) / cycle.getDuration();
```
sqrrm, line 104: This is old code, but wouldn't this return the wrong index if the cycle length has been modified at some point.

sqrrm (approved):
utACK

core/src/main/java/bisq/core/dao/governance/period/CycleService.java
ManfredKarrer: Would need to look closer to the caller. Might be that the caller has an old cycle and that cause issues. Anyway that reorg topic will require more testing and effort to cover all edge cases. So I leave that for upcoming PRs...

PR #2121, opened 2018-12-12
Fix reorg issues with phases

ManfredKarrer:


ManfredKarrer:
Sorry was wrong branch....

PR #2122, opened 2018-12-12
[WIP] Split bsq changeoutputs

ManfredKarrer:
Add multiple change outputs in case there are not sufficient utxos

To avoid that a user cannot make another BSQ tx before a pending tx is
confirmed in case he used up all utxo for that tx we add multiple change
outputs to following tx types:
- ~~Trade fee tx~~ (Trade protocol assumes 1 change output so we cannut use it here)
- Send BSQ tx
- Proposal Tx (but not comp. request and reimbursement)
- Blind vote tx
- Lockup tx

All others have more strict consensus rule regarding the output
structure.

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
@@ -60,6 +60,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Random;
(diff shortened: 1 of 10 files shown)


ManfredKarrer:
Please do not merge yet. Not sure if we should get it into the release last minute as it seems more complex as expected.

ManfredKarrer:
Too many issues with that approach and not solving the main problem (traders using BSQ as fee). Will close for now.

PR #2126, opened 2018-12-13
Add print tool for all markets

ripcurlx:
Used in https://github.com/bisq-network/bisq-website/blob/master/_includes/market_currency_selector.html


ManfredKarrer (approved):
utACK

PR #2131, opened 2018-12-15
Fix install_java.bat when setting long PATH environment variable

devinbileck:
Avoid using setx to append to the PATH environment variable as it will
truncate to 1024 characters. Instead, use SetEnvironmentVariable
in powershell.


ManfredKarrer (commented):
utACK

ManfredKarrer (approved):

PR #2156, opened 2018-12-21
Prepare for http-api incorporation

blabno:


ManfredKarrer:
Will review when back from holiday in January....

ManfredKarrer:
I will make a series of small PRs with specific changes grouped together. Currently there is too much in that and there is no Git history. The non risky changes I will do quickly the more complex one after the next release.

Issue #1952, opened 2018-11-23
Investigate signing Windows and macOS binary executables


devinbileck:
Launching the installation exe on Windows 10 triggers Windows Defender SmartScreen (see below). To install you have to select `More info` and then click `Run anyway`.

In order to prevent this, we will need to investigate signing the executable.

![image](https://user-images.githubusercontent.com/603793/48925312-3acf8e80-ee77-11e8-819d-c79ac338a2cc.png)

ManfredKarrer:
Signing for windows is long time on our todo list. That would be definitely good.

devinbileck:
@ripcurlx feel free to assign this one to me to investigate. Unless you have some ideas.

devinbileck:
Since Bisq is a decentralized project and team, there is no official business entity to register and sign the code signatures as required by certificate authorities.

Potentially a single trusted contributor could register and manage the code signing, likely as a bonded role.

However, Bitcoin Core has implemented code signing by having some core developers create an association with the single purpose of registering for code signing certificates. To quote their website (https://bitcoincorecodesigning.org/):

> Bitcoin Core signs its Windows and macOS binaries with code signing certificates. Those certificates either need to be bound to an individual or a company organization. Bitcoin Core is purely an open source project and has no legal entity, thus, some of the Core Developers have founded an association with the single purpose of registering for code signing certificates.

So Bisq could potentially follow suit and create an association just for the purpose of registering for code signing certificates. This approach would likely still require a bonded role for all those involved.

@ManfredKarrer @cbeams @ripcurlx Do you have any comments or suggestions on which approach and whether to proceed?

ManfredKarrer:
Still a priority!

devinbileck:
Still relevant

devinbileck:
Still relevant.

ripcurlx:
To reduce unnecessary OS warnings and AntiVirus issues on Windows we need to sign our Windows builds. 

- https://docs.microsoft.com/de-de/previous-versions/windows/internet-explorer/ie-developer/platform-apis/ms537361(v=vs.85)
- https://docs.microsoft.com/en-us/windows/win32/appxpkg/how-to-create-a-package-signing-certificate

I'll apply for a certificate to have everything ready for v1.2.1

ripcurlx:
@devinbileck It would be great if you could help out if I'm stuck on the way. Or maybe even prepare the build file for Windows to make the signing of the executable as easy as possible 😄

[...]

PR #2163, opened 2018-12-22
Update windows packaging process

devinbileck:
The 64bitBuild.bat script has been renamed to package.bat and has
been updated so that it is capable of performing the complete packaging
process without having to rely on the jar first being built and prepped
from the MacOS scripts. However, it does support having the jar
previously built and prepped and will look for a prepped jar in the
desktop/package folder. If not found, it will build and prep it
prior to packaging the executable.

Additionally, some unnecessary options that were being passed to
javapackager.exe have been eliminated such as BappVersion and Bruntime.
AppVersion is now being passed via environment variables and the
Bruntime option is valid only when the -native option is set to jnlp.

The Bisq.iss file was changed so it no longer needs to be updated with
AppVersion every time as it is being passed from package.bat via
environment variables. Also, AppCopyrightYear does not need to be
updated as it is determined automatically. A few other options were
added or tweaked as well.

Finally, a release.bat script has been added that will perform the
release process of copying necessary files to a versioned release folder
and generating/verifying signatures. Linux and MacOS packaged installers
should be copied to their appropriate package folders prior to
executing this script if they are to be included in this release
process, otherwise only the Windows files will be included.

The MacOS and Linux packaging scripts should be reviewed and updated
accordingly.


devinbileck:
I will review and update the Linux packaging scripts next.

devinbileck:
Note, signing the Windows installer (https://github.com/bisq-network/bisq/issues/1952) is still on my TODO list.

ManfredKarrer:
@ripcurlx Can you review that PR? I am on holiday untill January.

ripcurlx (changes requested):
@devinbileck Sorry for my late review.  I was also off most of the time and had a cold on top. 
Similar to my remarks [here](https://github.com/bisq-network/bisq/pull/2190#pullrequestreview-189302359), could you please update https://github.com/bisq-network/bisq/blob/db439ce4e2386ad7258f9481480751bcfc5f9f24/desktop/package/macosx/create_app.sh#L53 as well, so it works within the current build process?

ripcurlx:
Requested changes merged in https://github.com/bisq-network/bisq/pull/2190

ripcurlx:
@devinbileck Could you please check if my merge in https://github.com/bisq-network/bisq/pull/2163/commits/3d2817e884e5876b97ac6d9f8485c00a4c76657d is correct?

ripcurlx (approved):

devinbileck:
@ripcurlx Sure, I will review and test your changes.

devinbileck:
@ripcurlx The merge looks fine. However, a few comments to other commits:
1. The iss file requires the BOM since it contains unicode messages. Otherwise they will not be displayed properly within the installer for those languages. See https://stackoverflow.com/a/49923030/8871018. Not sure why you are encountering issues? (worked for me with Windows 10)
1. My intention with overriding JAVA_HOME was to simplify the process, rather than the user having to set JAVA_HOME before to OracleJDK, and then change it back afterwards to OpenJDK (its a pity packagemanager is not available in OpenJDK). I agree hard coding the path in the script makes it inflexible in case the user has it installed elsewhere, and they may need to edit it. But, I was considering investigating installing OracleJDK as part of the install_java script and we can have a standardized path (again, a pity that archived Java versions require a user account to download the JDK). I suppose an alternative solution is to use a different variable name within the script and check if the path exists first, else use JAVA_HOME environment variable.

ripcurlx:
@devinbileck 
1) I just removed the BOM as Windows complained during the build that it doesn't like this character in the beginning of the file. You don't see this error when running the script?
2) This was the reason why I removed it as I had a different installation directory on Linux. I removed it from Windows just because of consistency in that case. But yes, we could do some workaround with checking for the javapackager.

devinbileck:
1. No, worked fine for me. I will give it a try again.
2. I will see what I can do.

devinbileck:
1. @ripcurlx What is the error that you encounter?

As a result of it now missing the BOM, I just tried the v0.9.3 release installer on Windows 7 French and the custom messages are not displayed properly.

Launch the installer with the application running:
image

Comment text within the control panel:
image

Note, this issue will only affect the custom messages contained within the Bisq.iss file, not the general translations within the installer.

Once I added the BOM, the translations above were shown properly.

PR #2164, opened 2018-12-22
Update windows install experience

devinbileck:
- Increase the minimum supported OS version to Windows 7 with SP1.
Previously it was Windows XP which has been EOL for some time now.
Hopefully no one is still using Windows XP.

- Show a final dialog once the install has completed. This will inform
the user that the install has completed with an option to launch the
application. Previously it would skip this final dialog and immediately
launch the application, but the user may not want to launch it
immediately.

- Create an application shortcut on the user's desktop.

- Check if the application is running prior to install or uninstall and
abort with a prompt to the user that the application must first be
closed.

- German, French, and Spanish translations. The installer should
auto-detect the OS language and display all text in the appropriate
language. For now I have just added a couple languages, but it is very
easy to add any of the other languages supported by Inno Setup.

- Additionally, I removed a few unnecessary items that weren't being
utilized such as installing a service.


ripcurlx (approved):
utACK - Thanks, looking good! I'll give everything an in-depth testing round during our next release on Tuesday.

PR #2165, opened 2018-12-22
Update install_java.bat

devinbileck:
Do not download 7zip if it had previously been downloaded, for example
if this script had previously been run. And don't delete it at the end
of this script in case this script is run again. This is an attempt to
save time and bandwidth.


ManfredKarrer:
@ripcurlx Can you review that PR? I am on holiday untill January.

ManfredKarrer:
@ripcurlx Can you review that PR? I am on holiday untill January.

ripcurlx (approved):
utACK

PR #2171, opened 2018-12-25
[WIP] Spend pending BSQ

### LINKED ISSUE #2107: DAO: Submitting a proposal (2 BSQ burnt) labels entire UTXO "unverified" until block is processed, preventing submission of other proposals
See below:

used 2 bsq for vote

I had 9998 BSQ and burnt 2 to submit the proposal, causing 9996 to be "unverified" until the block was processed.

Since my entire BSQ balance is one UTXO that got spent as an input in a proposal request 
(tx: https://explorer.bisq.network/testnet/tx.html?tx=f86258a7ddfbfba163eaa54855ec400342f5e1d51dbedc93205f5e4b34ecff3c),
the remainder (9996) became "unverified" in the DAO gui, preventing other proposal-making until the next block.

warning

Should Bisq be able to recognize the output address is controlled by the same wallet?

Perhaps this is working as intended/there is no workaround, in which case perhaps the warning notice should be updated to make the users aware they simply must wait for the next block? Maybe with a link to the explorer? In general perhaps it makes sense to provide the transaction hash for a proposal made in the confirmation notice.

Thank you

sqrrm:
Introduces parsing of unconfirmed transactions to allow for spending of pending BSQ transactions.

I'm not really clear how the UI should handle the pending amounts, but the accounting needs an overhaul anyway so I plan to do that in a different PR.

Also fixes some other issues found along the way while testing.

Fixes #2107

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
@@ -23,6 +23,9 @@ import bisq.core.btc.exceptions.TransactionVerificationException;
 import bisq.core.btc.exceptions.WalletException;
 import bisq.core.btc.listeners.BsqBalanceListener;
 import bisq.core.btc.setup.WalletsSetup;
+import bisq.core.dao.exceptions.IncompleteBitcoinjTransactionException;
+import bisq.core.dao.node.full.RawTx;
+import bisq.core.dao.node.parser.TxParser;
 import bisq.core.dao.state.DaoStateListener;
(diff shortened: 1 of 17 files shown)


sqrrm:
I'm splitting this into separate PRs to not mix features in one PR

ManfredKarrer:
@sqrrm Should we close that one then?

sqrrm:
@ManfredKarrer I will close it when I've split the commits.

sqrrm:
Split this into #2242 and #2244

PR #2185, opened 2018-12-31
Update displayStrings.properties (error & bug report message)

BravoIndia:
Clarifying and correcting the grammar of the bug report message. Please review for your discretion, I've simplified the language to be as clear as possible to non-technical users, which i believe is appropriate and important.


ManfredKarrer (commented):
utACK

ManfredKarrer:
Thanks!

Issue #2134, opened 2018-12-18
Focus color of buttons should be more saturated/darker not lighter


ripcurlx:
Atm giving a button focus by tabbing through or by default focus shows the button with a lighter color as focus indicator.

Focus color should always be a more saturated color or darker to prevent seeing it as disabled by mistake.

**No focus**
![Bildschirmfoto 2018-12-18 um 10.29.29.png](https://images.zenhubusercontent.com/5a0bffe08a75884b90877eda/87b489a9-ce78-4f24-ba3a-e9bba7b0debb)
**Focus**
![Bildschirmfoto 2018-12-18 um 10.29.36.png](https://images.zenhubusercontent.com/5a0bffe08a75884b90877eda/e14cd87a-1745-41f3-b730-4bc04ead22a5)

CaveSpectre11:
I know where the change needs to be made; setup one of my old laptops to run linux to setup a development environment.  Going to try and get this one banged out.

CaveSpectre11:
Sorry it took a while, I was battling with Idea; it seemed to be building fine but it didn't update bisq-desktop.  I assumed, obviously, that I was changing the wrong thing, rather than it not getting built into image.

I didn't notice the part about tabs at first, but I got that taken care of to.  However, I'm not sure I like the action that occurs when you tab over to it.  It's darker, but instead of staying vivid, it fades... and I've not figured out what is causing that yet.   I'm going to PR as it is, and if the action fade is needed to be resolved, I'll look further into that.

CaveSpectre11:
This has been updated with PR #2187

ripcurlx:
Closing as complete.

PR #2187, opened 2018-12-31
Focus color of buttons should be more saturated/darker (Issue #2134)

CaveSpectre11:
Issue #2134 reports an issue with hovering or tabbing focus onto a button lightens the button rather than darkens it, making it more apt to appear disabled.

This PR corrects this behavior by changing the derived color on hover and focused to -30% rather than +30%.

One potential remaining issue is the action that occurs when the button is focused upon tabbing; there is an animated effect that dims the button.  While it is now darker than previous, it still becomes somewhat washed out and not as vivid as a mouse hover.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/bisq.css
+++ b/desktop/src/main/java/bisq/desktop/bisq.css
@@ -298,12 +298,12 @@ bg color of non edit textFields: fafafa
     -fx-cursor: hand;
 }
 
-.jfx-button:hover {
-    -fx-background-color: derive(-bs-rd-grey-light, 30%);
+.jfx-button:hover, .jfx-button:focused {
+    -fx-background-color: derive(-bs-rd-grey-light, -30%);
 }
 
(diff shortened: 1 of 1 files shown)


CaveSpectre11:
Hover (right side is hovered):
screenshot from 2018-12-31 12-40-10

screenshot from 2018-12-31 12-40-45

Focused (right side is focused)
screenshot from 2018-12-31 12-45-06

screenshot from 2018-12-31 12-44-28

ManfredKarrer:
@ripcurlx Leave it to you. Maybe @pedromvpg could add his opinion as well!

ripcurlx (approved):
utACK from a code perspective, but I'd love to get an ACK from @pedromvpg as well, if he's fine with these UI changes.

ripcurlx:
ACK - just tested it locally.  @pedromvpg If you want to change the color/saturation, we can adapt it in the next release again.

PR #2190, opened 2019-01-03
Update linux packaging process

devinbileck:
The 64bitBuild.sh script has been renamed to package.sh and has
been updated so that it is capable of performing the complete packaging
process without having to rely on the jar first being built and prepped
from the MacOS scripts. However, it does support having the jar
previously built and prepped and will look for a prepped jar in the
desktop/package folder. If not found, it will build and prep it
prior to packaging.

Additionally, the Bruntime option passed to javapackager is unnecessary and
has been removed. The Bruntime option is valid only when the -native
option is set to jnlp.

Also, the prepare-system.sh script was removed as it did not appear to be
doing anything useful.

Finally, a release.sh script has been added that will perform the
release process of copying necessary files to a versioned release folder
and generating/verifying signatures. Windows and MacOS packaged installers
should be copied to their appropriate package folders prior to
executing this script if they are to be included in this release
process, otherwise only the Linux files will be included.


ManfredKarrer:
@ripcurlx I leave the review to you.

ripcurlx (changes requested):
@devinbileck Thanks for putting this PR together! I had a quick look over the changed scripts and I have following questions/remarks:

I want to prevent to create the release builds from source on every platform, as it is a time consuming process already and this would make it even slower. To be able to re-use an existing jar file it has to be placed in a similar structure as if I would have the full repository checked out. So we would also need to adapt https://github.com/bisq-network/bisq/blob/db439ce4e2386ad7258f9481480751bcfc5f9f24/desktop/package/macosx/create_app.sh#L51 Correct? Could you add this changes as well, so my release process is still working with this PR?

The advantage of having the release process now also for Linux is that it is easier for Linux dev machines to create releases. The disadvantage is, that we have to maintain both release scripts if anything changes.

devinbileck:
Correct. I can attempt to modify the macosx script. However, I think it requires a little more cleanup. I did not touch it because I am unable to test and verify any changes I make (no access to a Mac). But I agree about the duplication and perhaps the Linux and Mac build/prep portions can be combined into a single script, with only the packaging portion separated.

ripcurlx:
@devinbileck No worries. I'll test it now on my VMs and will make the necessary changes.

desktop/package/linux/package.sh
```diff
@@ -0,0 +1,105 @@
...
+
+    echo Making deterministic jar by stripping out parameters and comments that contain dates
+    # Jar file created from https://github.com/ManfredKarrer/tools
+    # TODO Is this step still necessary? Since we are using preserveFileTimestamps and reproducibleFileOrder in build.gradle
+    java -jar $base_dir/desktop/package/tools-1.0.jar $jar_file
+
+    echo SHA256 after stripping jar file:
+    shasum -a256 $jar_file | awk '{print $1}' | tee $base_dir/desktop/package/desktop-$version-all.jar.txt
+
+    chmod o+rx "$base_dir/desktop/package/desktop-$version-all.jar"
```
ripcurlx: This only works when building from source. When using an existing jar file in a virtual box shared folder, we need to copy it locally and change the permissions and reference it from there.

ripcurlx (approved):

PR #2192, opened 2019-01-03
Revert to netlayer 0.6 because of an issue with seednodes

ManfredKarrer:
If a seednode is using netwlayer version 0.6.2 (or 0.6.1) it is not
reachable from other nodes. Still unclear what causes that and if it
would affect also normal Bisq nodes, but to be careful we revert to
latest stable netlayer version which fixes the issue.


freimair (approved):
utACK

sqrrm (approved):
utACK

PR #2195, opened 2019-01-03
Dao: fix vote reveal tx publishing

ManfredKarrer:
@sqrrm Opened the PR for review but need more testing before I merge it.


ManfredKarrer:
@sqrrm It is built on top of #2191 . You can ignore those commits....

core/src/main/java/bisq/core/dao/governance/votereveal/VoteRevealService.java
```diff
@@ -172,7 +175,13 @@ public void onParseBlockChainComplete() {
     // the blind vote was created in case we have not done it already.
     // The voter need to be at least once online in the reveal phase when he has a blind vote created,
     // otherwise his vote becomes invalid and his locked stake will get unlocked
-    private void maybeRevealVotes(int chainHeight) {
+    private void maybeRevealVotes() {
+        // We must not use daoStateService.getChainHeight() because that gets updated with each parsed block but we
+        // only want to publish the vote reveal tx if our current real chain height is matching the cycle and phase and
+        // not at any intermediate height during parsing all blocks. The bsqNode knows the latest height from either
+        // Bitcoin Core or from the seed node.
+        int chainHeight = bsqNode.getChainTipHeight();
```
sqrrm, line 183: I have a feeling that we might want to wait to publish until we've parsed the chain up until chainTipHeight instead of publishing before we're done parsing. I don't have a concrete example, but publishing before knowing the actual current state seems like it could lead to more trouble. Maybe it would invalidate some BSQ, or maybe it would just burn some to fees.

core/src/main/java/bisq/core/dao/governance/votereveal/VoteRevealService.java
```diff
@@ -183,75 +184,87 @@ private void maybeRevealVotes() {
...
+
+                        boolean isAfterVoteRevealPhase = periodService.getPhaseForHeight(chainHeight).ordinal() > DaoPhase.Phase.VOTE_REVEAL.ordinal();
+
+                        // We missed the reveal phase but we are in the correct cycle
+                        boolean missedPhaseSameCycle = isAfterVoteRevealPhase && isBlindVoteTxInCorrectPhaseAndCycle;
+
+                        // If we missed the cycle we don't care about the phase anymore.
+                        boolean isBlindVoteTxInPastCycle = periodService.isTxInPastCycle(myVote.getTxId(), chainHeight);
+
+                        if (missedPhaseSameCycle || isBlindVoteTxInPastCycle) {
```
sqrrm, line 212: It seems all paths lead here except for when the time between the vote is cast and the reveal phase. Wouldn't it be easier to check at which block height the vote reveal should be sent and do a simple check if the block height is more than the reveal block height to send the vote reveal?

sqrrm (approved):
utACK

It might be worth simplifying the vote reveal publishing but not necessary.

I think this PR will also make sure to not burn any BSQ associated with a vote reveal, even if it's published before the vote reveal phase, but the vote calculation will still be correct.

core/src/main/java/bisq/core/dao/governance/votereveal/VoteRevealService.java
ManfredKarrer: The onNewBlockHeight is called before the parsing of the block. So at the first block of the reveal phase we create and publish the tx. In the next block it will be validated and added to to the DaoState. So we could not wait for parsing until it is not published.

core/src/main/java/bisq/core/dao/governance/votereveal/VoteRevealService.java
ManfredKarrer: Yes the current checks are a bit verbose. I preferred to make it very explicit which cases are handled and not optimize on logical code flow (which could be optimized). But feel free to make a suggestion for a change.

PR #2197, opened 2019-01-03
Remove "pricenode" direct dependency on "assets"

oscarguindzberg:
It depends on "assets" already because it is a depencendy of "core"


ManfredKarrer (approved):
utACK

PR #2198, opened 2019-01-03
Be able to upgrade bitcoinj without upgrading libdohj

oscarguindzberg:


ripcurlx (approved):
~utACK~ ACK - Builds and runs locally for me. As @cbeams won't be available for additional review, I'll merge it myself.

PR #2205, opened 2019-01-04
Update offer button state on account change

### LINKED ISSUE #2204: Unable to take a fixed amount offer with multiple accounts
To reproduce:

1. Have an account aged 2 months
2. Create an account of the same payment method aged less than two months
3. Try and take a fixed amount offer that only the 2-month account should be able to accept

On the take offer screen that pops up, changing the account does not enable the “Next Step” button even when the 2-month account is selected.

j3ko:
fixes #2204 

Add updateButtonDisabledState() to onPaymentAccountSelected() in
TakeOfferViewModel since changing the account dropdown in the take
offer screen does not trigger an update of the "Next Step" button.

The default amount of an offer should not change to less than what is
actually offered, change onPaymentAccountSelected() in
TakeOfferDataModel to take into account the min. amount offered.

In TakeOfferView, the payment account defaults to the last one
selected.  Make sure TakeOfferViewModel and TakeOfferDataModel take
this into account in a couple locations.


ripcurlx:
ACK - @j3ko Thanks for the PR!

ripcurlx (approved):

PR #2211, opened 2019-01-07
Configure default system JDK

devinbileck:
The install_java.sh script was not setting the appropriate java and
javac as default. This is not an issue if no other JDK is installed, but
if so it could cause the wrong JDK to be used. So it now sets the
appropriate default system JDK.

Also, added a comment at the top of the install_java.sh and
install_java.bat files indicating it configures the default system
JDK and informs the user how to go about selecting a different
default JDK if necessary.


ManfredKarrer (approved):
utACK

PR #2212, opened 2019-01-07
Simplify fee model

ManfredKarrer:


ripcurlx (changes requested):
Please fix ```CreateOfferViewModelTest.java:115:```. Besides that utACK

PR #2220, opened 2019-01-08
Don't show filtered currencies

ManfredKarrer:
- For supporting Grin we want to add it to the next release but Grin
is not launched yet (15.1.2018 is launch date). We deactivate the asset
by using the filter message and enable it at launch date.
To support that at the altcoin account screen this commit implements
the required code.


ripcurlx (approved):
utACK

PR #1970, opened 2018-11-26
List SpaceCash (XSC)

tlightsky:
Official project URL: https://hspace.app
Official block explorer URL: https://hyperstats.info/

This is a resubmission of List SpaceCash (XSC) #1766 
have done the rebase of current master and conflict resolve

PR #2232, opened 2019-01-09
fix the wrong ticker name from SPACE to XSC in Coin second param

tlightsky:
we use a wrong param in the pr before, fix that in this pr

the pr before:
https://github.com/bisq-network/bisq/pull/1970

### DIFF
--- a/assets/src/main/java/bisq/asset/coins/SpaceCash.java
+++ b/assets/src/main/java/bisq/asset/coins/SpaceCash.java
@@ -23,6 +23,6 @@ import bisq.asset.RegexAddressValidator;
 public class SpaceCash extends Coin {
 
     public SpaceCash() {
-        super("SpaceCash", "SPACE", new RegexAddressValidator("^([0-9a-z]{76})$"));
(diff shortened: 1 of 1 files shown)


ripcurlx:
@ManfredKarrer This change would cause issues if we have open offers/trades already, wouldn't it?

ManfredKarrer:
Ticker symbol is used as ID and should not be changed. Requires extra testing. Don't want to invest that effort before we know that it is traded. >90% of added assets never are traded anyway. So I will not do the testing for that and others likely don't have enough context to test it correctly.

tlightsky:
@ripcurlx @ManfredKarrer 
https://bisq.network/markets/?currency=space_btc
this is the trade page, seems no history till now, 
hope we can fix it before too late

tlightsky:
@ManfredKarrer ok, so we should do it maybe after some trade happen?

tlightsky:
@blabno hi, our ticker name need a small fix, 
can we merge this fix to prevent someone can't find our coin to trade

PR #2237, opened 2019-01-10
Clear and re-assign series to fix rendering issues.

### LINKED ISSUE #2136: Changing The Currency in the Offer Book Results in Faulty Chart
When the currency in the _Offer Book_ is changed with the combo box from a currency that has both sell and buy offer e.g. EURO to a currency that only has buy offers e.g. NOK, then an artifact of the previous sell graph remains in the chart.

**Correct Euro Chart**
euro

**Incorrect NOK Chart** - There are no sell offers
nok

ripcurlx:
Fixes #2136.


ManfredKarrer (approved):
utACK

PR #2242, opened 2019-01-11
Dao avoid bsq burn

sqrrm:


core/src/main/java/bisq/core/dao/node/parser/TxParser.java
```diff
@@ -248,10 +248,15 @@ private void processIssuance(int blockHeight, TempTx tempTx, long bsqFee) {
                 log.warn("It can be that we have a opReturn which is correct from its structure but the whole tx " +
                         "in not valid as the issuanceCandidate in not there. " +
                         "As the BSQ fee is set it must be either a buggy tx or an manually crafted invalid tx.");
-                tempTx.setTxType(TxType.INVALID);
+                // Even though the request part if invalid the BSQ transfer and change output should still be valid
+                // as long as the BSQ change <= BSQ inputs.
+                tempTx.setTxType(TxType.UNDEFINED_TX_TYPE);
```
ManfredKarrer, line 253: Not sure if we should use UNDEFINED_TX_TYPE as it does not describe it correctly. Maybe some weaker term than "invalid"? What about NON_STANDARD or "NON_RULED_COMFORM"?

core/src/main/java/bisq/core/dao/governance/bond/Bond.java
```diff
@@ -53,8 +53,10 @@ protected Bond(T bondedAsset) {
     }
 
     public boolean isActive() {
-        return bondState != BondState.READY_FOR_LOCKUP &&
-                bondState != BondState.UNLOCKED;
+        return bondState == BondState.LOCKUP_TX_CONFIRMED ||
```
ManfredKarrer, line 56: I think UNLOCK_TX_PENDING is missing

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1635,7 +1635,8 @@ dao.wallet.dashboard.price=Latest BSQ/BTC trade price (in Bisq)
 dao.wallet.dashboard.marketCap=Market capitalisation (based on trade price)
 
 dao.wallet.receive.fundYourWallet=Fund your BSQ wallet
-dao.wallet.receive.bsqAddress=BSQ wallet address
+dao.wallet.receive.bsqAddress=BSQ wallet address (This is a new unused address to avoid address reuse, \
```
ManfredKarrer: That is quite verbose. What about:
Fresh unused BSQ wallet address

sqrrm:
Regarding Bond.isActive(), I think it's already correct the way it is now.

ManfredKarrer (changes requested):

core/src/main/java/bisq/core/dao/node/parser/TxParser.java
sqrrm: If we add another type we can, this was an attempt to avoid adding more txTypes.

core/src/main/java/bisq/core/dao/governance/bond/Bond.java
sqrrm: It's here, next line down.

core/src/main/resources/i18n/displayStrings.properties
sqrrm: I shortened in a later commit

core/src/main/java/bisq/core/dao/node/parser/TxParser.java
ManfredKarrer: I think IRREGULAR might be a good type for not-rule-conform but still valid txs.

sqrrm: Yeah, let's do that to show it's parsed and something is not right but it's still a valid BSQ tx in some way.

PR #2244, opened 2019-01-11
[WIP] Spend unacked bsq

sqrrm:
This PR allows to spend pending BSQ. It doesn't check where the pending txs came from however so any incoming BSQ will be spendable immediately. It might be better to only allow to spend own unacked txs rather than all, not sure.

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
@@ -23,6 +23,8 @@ import bisq.core.btc.exceptions.TransactionVerificationException;
 import bisq.core.btc.exceptions.WalletException;
 import bisq.core.btc.listeners.BsqBalanceListener;
 import bisq.core.btc.setup.WalletsSetup;
+import bisq.core.dao.node.full.RawTx;
+import bisq.core.dao.node.parser.TxParser;
 import bisq.core.dao.state.DaoStateListener;
 import bisq.core.dao.state.DaoStateService;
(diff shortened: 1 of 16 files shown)


sqrrm:
Depends on #2242, that should be merged first

ManfredKarrer:
Here a few general comments:
- I would prefer to not add too much dependency from DAO to BitcoinJ. The btc package should ideally does not know about the DAO (I know there are some violations of that goal, but I prefer to not add new ones). So the TxParser should not be injected there. We could invert that so that BsqNode listens on BitcoinJ for new blocks and then passes those transactions to the parser.
- The UnconfirmedState is not persisted. Could that mess up the state if you close the app and restart? 
- There is no guarantee of the order we receive the block from Bitcoin Core and the tx from BitcoinJ. It can be that we complete parsing but BitcoinJ has received the tx still as unconfirmed as well the other way round that BitcoinJ gets a tx which is condisdered confirmed but we have not got it from Bitcoin Core. Not sure if that can cause issues but I fear it can and it is difficult to test and simulate. 
Another approach might be to use Bitcoin Core to request the tx by `getmempoolentry <txId>` but of course that would only work for a full DAO node.

Lets discuss all that directly, it feels quite a complex and risky thing we want to achieve here.

ManfredKarrer:
Btw was looking into the other dao dependencies there. I think we should extract those to a custom class which is responsible to deal with both domains: bitcoinj and DAO.

ManfredKarrer:
Superseded by https://github.com/bisq-network/bisq/pull/2482

PR #2247, opened 2019-01-11
Fix windows install custom translations

devinbileck:
The Bisq.iss file must be saved using UTF-8 encoding with BOM for the unicode custom messages to be displayed properly.

For example, the v0.9.3 release on Windows 7 French has the following issues:

- Launching the installer with the application running shows the following incorrect message text:
image

- Observing the application within the control panel shows the following incorrect message text:
image

Note, this issue only affects the custom messages contained within the Bisq.iss file, not the general translations within the installer.

### DIFF
--- a/desktop/package/windows/Bisq.iss
+++ b/desktop/package/windows/Bisq.iss
@@ -1,6 +1,6 @@
-;This file will be executed next to the application bundle image
+﻿;This file will be executed next to the application bundle image
 ;I.e. current directory will contain folder Bisq with application files
-;Note: This file must use UTF-8 encoding
+;Note: This file must use UTF-8 encoding with BOM for the unicode custom messages to be displayed properly
 
 #define SourceDir GetEnv('package_dir') + '\windows'
 #define AppVersion GetEnv('version')


devinbileck:
@ripcurlx let me know if you are still encountering issues with the BOM.

ripcurlx:
I'll check it out again, but I fear it will fail during the build again. I'll let you know about the outcome

ripcurlx:
@devinbileck Unfortunately I always get following error (Windows 10 Home):
```
Error on line 1 in C:\Users\chris\AppData\Local\Temp\fxbundler2867454632442430686\images\win-exe.image\Bisq.iss: Text is not inside a section.
Compile aborted.
java.io.IOException: Exec failed with code 2 command [[C:\Program Files (x86)\Inno Setup 5\iscc.exe, /oE:\desktop\package\windows\..\windows, C:\Users\chris\AppData\Local\Temp\fxbundler2867454632442430686\images\win-exe.image\Bisq.iss] in C:\Users\chris\AppData\Local\Temp\fxbundler2867454632442430686\images\win-exe.image
```

If I run the Bisq.iss file within the Inno Setup Compiler I get following error:
screenshot

Any idea why it fails on my side?

ripcurlx:
ACK - I searched quickly around and found out, that the problem is the non-unicode installation of InnoSetup. I now installed a unicode version manually and it builds now without any problem.

PR #2248, opened 2019-01-11
Improve JAVA_HOME override

devinbileck:
Check to see if javapackager exists prior to overriding JAVA_HOME
in package scripts.


ripcurlx (approved):
utACK - looking good

PR #2249, opened 2019-01-11
Refactor BalanceModel to Balances

ManfredKarrer:
- Remove BalanceModel
- Add Balances (pure domain model)
- Add BalancePresentation (shared presentation model in core)
- Move domain listener callbacks which can trigger balance update
to Balances


ManfredKarrer:
Close as included in https://github.com/bisq-network/bisq/pull/2250

PR #2251, opened 2019-01-11
Refactor fee estimation

ManfredKarrer:
Danger zone ;-)
This needs some good testing and careful review.

### DIFF
--- a/core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
+++ b/core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
@@ -127,17 +127,18 @@ public class TxFeeEstimationService {
     }
 
     // We start with the initialEstimatedTxSize for a tx with 1 input (260) bytes and get from BitcoinJ a tx back which
-    // contains the required inputs to fund that tx (out outputs + miner fee). The miner fee is that case is based on
+    // contains the required inputs to fund that tx (outputs + miner fee). The miner fee in that case is based on
     // the assumption that we only need 1 input. Once we receive back the real tx size from the tx BitcoinJ has created
-    // with the required inputs we compare that if it is in the tolerance of 20% from our assumed tx size. If we inside
-    // that tolerance we use that tx size for our fee estimation. If not (if there has been more then 1 inputs) we
-    // apply the new fee based on the reported tx size and request again from BitcoinJ to full that tx with the inputs
+    // with the required inputs we compare if the size is not more then 20% different to our assumed tx size. If we are inside
+    // that tolerance we use that tx size for our fee estimation, if not (if there has been more then 1 inputs) we
+    // apply the new fee based on the reported tx size and request again from BitcoinJ to fill that tx with the inputs
     // to be sufficiently funded. The algorithm how BitcoinJ selects utxos is complex and contains several aspects
     // (minimize fee, don't create too many tiny utxos,...). We treat that algorithm as an unknown and it is not
     // guaranteed that there are more inputs required if we increase the fee (it could be that there is a better
-    // selection of inputs chosen if we have increased the fee and therefor less inputs). As the increased fee might
+    // selection of inputs chosen if we have increased the fee and therefore less inputs and smaller tx size). As the increased fee might
     // change the number of inputs we need to repeat that process until we are inside of a certain tolerance. To avoid
-    // potential endless loops we add a counter. Worst case would be that the last size we got reported is > 20% off to
+    // potential endless loops we add a counter (we use 10, usually it takes just very few iterations).
+    // Worst case would be that the last size we got reported is > 20% off to
     // the real tx size but as fee estimation is anyway a educated guess in the best case we don't worry too much.
     // If we have underpaid the tx might take longer to get confirmed.
     @VisibleForTesting
--- a/core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
+++ b/core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
@@ -42,6 +42,10 @@ import static com.google.common.base.Preconditions.checkArgument;
 @Slf4j
 public class TxFeeEstimationService {
     public static int TYPICAL_TX_WITH_1_INPUT_SIZE = 260;
+    private static int DEPOSIT_TX_SIZE = 320;
+    private static int PAYOUT_TX_SIZE = 380;
+    private static int BSQ_INPUT_INCREASE = 150;
+    private static int MAX_ITERATIONS = 10;
     private final FeeService feeService;
     private final BtcWalletService btcWalletService;
     private final Preferences preferences;
@@ -83,13 +87,13 @@ public class TxFeeEstimationService {
                                                            Preferences preferences) {
         Coin txFeePerByte = feeService.getTxFeePerByte();
         // We start with min taker fee size of 260
-        int estimatedTxSize = 260;
+        int estimatedTxSize = TYPICAL_TX_WITH_1_INPUT_SIZE;
         try {
             estimatedTxSize = getEstimatedTxSize(List.of(tradeFee, amount), estimatedTxSize, txFeePerByte, btcWalletService);
         } catch (InsufficientMoneyException e) {
             if (isTaker) {
                 // if we cannot do the estimation we use the payout tx size
-                estimatedTxSize = 380;
+                estimatedTxSize = PAYOUT_TX_SIZE;
             }
             log.info("We cannot do the fee estimation because there are not enough funds in the wallet. This is expected " +
                     "if the user pays from an external wallet. In that case we use an estimated tx size of {} bytes.", estimatedTxSize);
@@ -97,15 +101,15 @@ public class TxFeeEstimationService {
 
         if (!preferences.isPayFeeInBtc()) {
             // If we pay the fee in BSQ we have one input more which adds about 150 bytes
-            estimatedTxSize += 150;
+            estimatedTxSize += BSQ_INPUT_INCREASE;
         }
 
         Coin txFee;
         int size;
         if (isTaker) {
-            int averageSize = (estimatedTxSize + 320) / 2;  // deposit tx has about 320 bytes
+            int averageSize = (estimatedTxSize + DEPOSIT_TX_SIZE) / 2;  // deposit tx has about 320 bytes
             // We use at least the size of the payout tx to not underpay at payout.
-            size = Math.max(380, averageSize);
+            size = Math.max(PAYOUT_TX_SIZE, averageSize);
             txFee = txFeePerByte.multiply(size);
             log.info("Fee estimation resulted in a tx size of {} bytes.\n" +
                     "We use an average between the taker fee tx and the deposit tx (320 bytes) which results in {} bytes.\n" +
@@ -139,7 +143,7 @@ public class TxFeeEstimationService {
             }
             counter++;
         }
-        while (!isInTolerance && counter < 10);
+        while (!isInTolerance && counter < MAX_ITERATIONS);
         if (!isInTolerance) {
             log.warn("We could not find a tx which satisfies our tolerance requirement of 20%. " +
                             "realTxSize={}, estimatedTxSize={}",
--- a/core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
+++ b/core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
@@ -1,0 +1,159 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.core.btc;
+
+import bisq.core.btc.wallet.BtcWalletService;
+import bisq.core.provider.fee.FeeService;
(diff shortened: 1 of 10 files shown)


core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
```diff
@@ -0,0 +1,159 @@
...
+                // if we cannot do the estimation we use the payout tx size
+                estimatedTxSize = 380;
+            }
+            log.info("We cannot do the fee estimation because there are not enough funds in the wallet. This is expected " +
+                    "if the user pays from an external wallet. In that case we use an estimated tx size of {} bytes.", estimatedTxSize);
+        }
+
+        if (!preferences.isPayFeeInBtc()) {
+            // If we pay the fee in BSQ we have one input more which adds about 150 bytes
+            estimatedTxSize += 150;
```
blabno: Can we avoid magic numbers? https://en.wikipedia.org/wiki/Magic_number_(programming)

core/src/test/java/bisq/core/btc/TxFeeEstimationServiceTest.java
```diff
@@ -0,0 +1,123 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@RunWith(PowerMockRunner.class)
```
blabno: There is no need for PowerMock in this test class.

core/src/test/java/bisq/core/btc/TxFeeEstimationServiceTest.java
```diff
@@ -0,0 +1,123 @@
...
+        txFeePerByte = Coin.valueOf(10);
+        realTxSize = 260;
+
+        txFee = txFeePerByte.multiply(initialEstimatedTxSize);
+        when(btcWalletService.getEstimatedFeeTxSize(outputValues, txFee)).thenReturn(realTxSize);
+        result = TxFeeEstimationService.getEstimatedTxSize(outputValues, initialEstimatedTxSize, txFeePerByte, btcWalletService);
+        assertEquals(260, result);
+
+
+        // TODO check how to use the mocking framework for repeated calls
```
blabno: This is separate test scenario and should be placed in separate method.
Good naming convention for test methods is `testMethodName_scenario_expectedCondition`
i.e. `getEstimatedTxSize_nullOutputs_throwsException`

core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
```diff
@@ -0,0 +1,159 @@
...
+        } else {
+            size = estimatedTxSize;
+            txFee = txFeePerByte.multiply(size);
+            log.info("Fee estimation resulted in a tx size of {} bytes and a tx fee of {} Sat.", size, txFee.value);
+        }
+
+        return new Tuple2<>(txFee, size);
+    }
+
+    @VisibleForTesting
```
blabno, line 144: I've been always told that this is a bad idea, as it's an attempt to test implementation details.
Either extract it to an utility class that could be tested separately or do not test it individually.

core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
ManfredKarrer: Yes we can use a constant. It is just used once so thats why I did not use one and the class is small so easty to overlook all as well there is a comment decribing the value. But no problem to make a constant for it.

core/src/test/java/bisq/core/btc/TxFeeEstimationServiceTest.java
ManfredKarrer: OK, I will remove it

core/src/test/java/bisq/core/btc/TxFeeEstimationServiceTest.java
ManfredKarrer: You refer to the commented out lines, right? Would you mind to have a hands on the test class? More your domain, you know my love for tests is limited ;-)

core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
ManfredKarrer: Not clear what you mean. Lets discuss on a call in a bit...

core/src/test/java/bisq/core/btc/TxFeeEstimationServiceTest.java
ManfredKarrer: @blabno I broke them up in https://github.com/bisq-network/bisq/pull/2251/commits/877d0ae3acb2a1ea94e389f99cbd0194ebae6aee and looked again into it. See commit comments... Could u have a look to try to fix it?

core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * Util class for getting the estimated tx fee for maker or taker fee tx.
+ */
+@Slf4j
+public class TxFeeEstimationService {
+    public static int TYPICAL_TX_WITH_1_INPUT_SIZE = 260;
+    private static int DEPOSIT_TX_SIZE = 320;
+    private static int PAYOUT_TX_SIZE = 380;
+    private static int BSQ_INPUT_INCREASE = 150;
+    private static int MAX_ITERATIONS = 10;
```
ripcurlx, line 48: Is there a specific reason why you've put payout_tx_size, bsq_input_increase and max_iterations here and not as a local variable? To get a better overview on the values used within this class?

core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
ripcurlx: I think @blabno is talking about the fact, that if you need to make something visible for testing that wouldn't be accessible otherwise, is a sign that something should be structured or tested differently 😉

core/src/test/java/bisq/core/btc/TxFeeEstimationServiceTest.java
```diff
@@ -0,0 +1,146 @@
...
+        int result;
+        int realTxSize;
+        Coin txFee;
+
+        initialEstimatedTxSize = 260;
+        txFeePerByte = Coin.valueOf(10);
+        realTxSize = 2600;
+
+        txFee = txFeePerByte.multiply(initialEstimatedTxSize);
+        when(btcWalletService.getEstimatedFeeTxSize(outputValues, txFee)).thenReturn(realTxSize);
```
blabno, line 80: do/while loop inside `getEstimatedFeeTxSize` has 2 cycles because first invocation of `isInTolerance` returns `false`.
So for the first iteration we have to mock with initial `txFee`:
        
    when(btcWalletService.getEstimatedFeeTxSize(outputValues, txFee)).thenReturn(realTxSize);

And for the second iteration we have to mock with `txFee` multiplied by `txFeePerByte`.

    when(btcWalletService.getEstimatedFeeTxSize(outputValues, txFeePerByte.multiply(realTxSize))).thenReturn(realTxSize)

Alternatively we can mock `getEstimatedFeeTxSize` to return the same result for any `txFee`:

    when(btcWalletService.getEstimatedFeeTxSize(eq(outputValues), any(Coin.class))).thenReturn(realTxSize);

core/src/test/java/bisq/core/btc/TxFeeEstimationServiceTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+        // repeated calls to getEstimatedFeeTxSize do not work (returns 0 at second call in loop which cause test to fail)
+        result = TxFeeEstimationService.getEstimatedTxSize(outputValues, initialEstimatedTxSize, txFeePerByte, btcWalletService);
+        assertEquals(2600, result);
+    }
+
+    // FIXME @Bernard could you have a look?
```
blabno, line 87: Same as above

ManfredKarrer:
@ripcurlx Could you ACK if you have completed review?

core/src/test/java/bisq/core/btc/TxFeeEstimationServiceTest.java
ManfredKarrer: @blabno Thanks. I would prefer to leave that for another PR, too buys with other high prio stuff. If you have time to take that would be great as well!

core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
```diff
@@ -0,0 +1,179 @@
...
+    static int getEstimatedTxSize(List<Coin> outputValues,
+                                  int initialEstimatedTxSize,
+                                  Coin txFeePerByte,
+                                  BtcWalletService btcWalletService)
+            throws InsufficientMoneyException {
+        boolean isInTolerance;
+        int estimatedTxSize = initialEstimatedTxSize;
+        int realTxSize;
+        int counter = 0;
+        do {
```
ripcurlx, line 154: As discussed it would be good to add a note why there is the need to run this part of the code up to 10 times to find the best transaction setup.

desktop/src/main/java/bisq/desktop/main/offer/MutableOfferDataModel.java
```diff
@@ -135,10 +134,9 @@
     protected PaymentAccount paymentAccount;
     protected boolean isTabSelected;
     protected double marketPriceMargin = 0;
-    private Coin txFeeFromFeeService = Coin.ZERO;
-    private boolean marketPriceAvailable;
-    private int feeTxSize = 260; // size of typical tx with 1 input
-    private int feeTxSizeEstimationRecursionCounter;
+    protected Coin txFeeFromFeeService = Coin.ZERO;
+    protected boolean marketPriceAvailable;
+    protected int feeTxSize = TxFeeEstimationService.TYPICAL_TX_WITH_1_INPUT_SIZE;
```
ripcurlx: Is there a reason why these three fields are protected and not private?

desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferDataModel.java
```diff
@@ -328,81 +333,37 @@ void onTakeOffer(TradeResultHandler tradeResultHandler) {
...
+            // as they are dependent txs the miner will pick both if the fee in total is good enough.
+            // We make sure that the fee is sufficient to meet our intended fee/byte for the larger payout tx with 380 bytes.
+            Tuple2<Coin, Integer> estimatedFeeAndTxSize = txFeeEstimationService.getEstimatedFeeAndTxSizeForTaker(fundsNeededForTrade,
+                    getTakerFee());
+            txFeeFromFeeService = estimatedFeeAndTxSize.first;
+            feeTxSize = estimatedFeeAndTxSize.second;
         } else {
-            feeTxSize = 320;
-            txFeeFromFeeService = getTxFeeBySize(feeTxSize);
+            feeTxSize = 380;
```
ripcurlx, line 360: Why did this change from 320?

desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferDataModel.java
```diff
@@ -620,10 +581,32 @@ public String getCurrencyNameAndCode() {
     }
 
     public Coin getTotalTxFee() {
+        Coin totalTxTees = txFeeFromFeeService.add(getTxFeeForDepositTx()).add(getTxFeeForPayoutTx());
```
ripcurlx: Typo: Should be totalTxFees not ..Tees

ripcurlx (approved):
ACK - Just a couple of questions and minor changes.

core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
ManfredKarrer: I have not a very strong opinion here, @blabno critiziced that they should be statics and not magic numbers in the code. I usually only make statics if they are used repeatedly and/or if the class is longer and it helps to see quickly important parameters. I would prefer to leave it now.

core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
ManfredKarrer: OK, I will add more comment.

desktop/src/main/java/bisq/desktop/main/offer/MutableOfferDataModel.java
ManfredKarrer: No, I will make them private.

desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferDataModel.java
ManfredKarrer: In the TxFeeEstimationService we use as min. the payout tx size (380), so that reflects the same logic. It is to be more on the safe side and prefer to overpay the other txs and not underpay the payout tx.

From: TxFeeEstimationService
   int averageSize = (estimatedTxSize + DEPOSIT_TX_SIZE) / 2;  // deposit tx has about 320 bytes
            // We use at least the size of the payout tx to not underpay at payout.
            size = Math.max(PAYOUT_TX_SIZE, averageSize);

desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferDataModel.java
ManfredKarrer: Fixed

core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
ManfredKarrer: Ah ok. Yes basically agree. In that case I would not like to expose getEstimatedTxSize as it is only interesting for that class. The more high level methods carry much more special case handling that they would be harder to test then the core functionality in getEstimatedTxSize.

ManfredKarrer:
No idea why tests fail, the EditOfferDataModelTest does not exist. Seems a caching issue.

PR #2252, opened 2019-01-11
Move util code from UI to core

ManfredKarrer:
Move code to OfferUtil and PaymentAccountUtil to avoid code duplication
once http-API gets merged


ManfredKarrer:
@blabno I added some methods of which I could not find a code duplication in the UI code but commented it out. Maybe you can have a look where it should be used. Otherwise if its API only consider to keep it there.

ripcurlx (approved):
ACK

Issue #2259, opened 2019-01-12
Test fails if language is not set to en_US


ManfredKarrer:
CurrencyListTest.testUpdateWhenNotSortNumerically fails for some users.

See https://github.com/bisq-network/bisq/issues/1562#issuecomment-453754849

"My system's locale is lt_LT.UTF-8. Using LC_ALL=en_US fixes the problem."

PR #2264, opened 2019-01-14
Fix CurrencyListTest when locale is not en_US

devinbileck:
When declaring a fiat currency and only providing the currency code,
it uses the system locale to get the currency display name.
This potentially causes different sorting results in
testUpdateWhenNotSortNumerically when locale is not en_US.

Now, it specifies the en_US locale when defining the fiat currencies
for this test.

Fixes https://github.com/bisq-network/bisq/issues/2259


ripcurlx (approved):
utACK - Thanks for fixing it. Weird that it didn't fail on my local machine which has a German locale.

PR #2275, opened 2019-01-16
Start api before wallet started

ManfredKarrer:


blabno (commented):
utACK Code looks very good. I'll try merging it into full version of api and see how it runs.

ManfredKarrer:
@blabno u can give a utACK. ACK if you have tested it.

ripcurlx (approved):
ACK - Everything seems to behave as expected with the desktop client

PR #2277, opened 2019-01-16
Adjust length of DAO phases depending on network

devinbileck:
Adjust the length of DAO phases depending if running on mainnet,
testnet, or regtest.


ManfredKarrer (approved):
utACK

ManfredKarrer:
Great thanks!

PR #2281, opened 2019-01-17
List Cash2 (CASH2)

cash2org:
- Official project URL: https://cash2.org
- Official block explorer URL: https://blocks.cash2.org


blabno (commented):
This asset uses Cryptonote validator. Shouldn't it include similar messages as Monero?

```
Trading XMR on Bisq requires that you understand and fulfill \
the following requirements:\n\n\
For sending XMR, you need to use either the official Monero GUI wallet or Monero CLI wallet with the \
store-tx-info flag enabled (default in new versions). Please be sure you can access the tx key as \
that would be required in case of a dispute.\n\
monero-wallet-cli (use the command get_tx_key)\n\
monero-wallet-gui (go to history tab and click on the (P) button for payment proof)\n\n\
In addition to XMR checktx tool (https://xmr.llcoins.net/checktx.html) verification can also be accomplished in-wallet.\n\
monero-wallet-cli : using command (check_tx_key).\n\
monero-wallet-gui : on the Advanced > Prove/Check page.\n\
At normal block explorers the transfer is not verifiable.\n\n\
You need to provide the arbitrator the following data in case of a dispute:\n\
- The tx private key\n\
- The transaction hash\n\
- The recipient's public address\n\n\
Failure to provide the above data, or if you used an incompatible wallet, will result in losing the \
dispute case. The XMR sender is responsible for providing verification of the XMR transfer to the \
arbitrator in case of a dispute.\n\n\
There is no payment ID required, just the normal public address.\n\
If you are not sure about that process visit (https://www.getmonero.org/resources/user-guides/prove-payment.html) \
or the Monero forum (https://forum.getmonero.org) to find more information.
```

cash2org:
Hello and thank you for looking into my pull request. Yes, I agree that a similar message should be included. Do you know where I can include it and how it should be done?

blabno:
Have a look at Blur: https://github.com/bisq-network/bisq/commit/7e9bc1fee88887615b3460cdeb0153848e4c09b7#diff-ef294f386c27fc762baf4c55e8b33e53

cash2org:
Hello, I am sorry but it looks like things got kind of messed up in this commit. I was getting merge conflict errors, so I updated my branch from the master branch. It seems now that I am getting a build error that is not related to the code I am trying to add. Do you have suggestions on how I can fix this?

PR #2291, opened 2019-01-19
List Mile (MILE)

vlad-khramov:
- Official project URL: http://mile.global/
- Official block explorer URL: https://explorer.mile.global/

### DIFF
--- a/assets/src/main/java/bisq/asset/coins/Mile.java
+++ b/assets/src/main/java/bisq/asset/coins/Mile.java
@@ -1,0 +1,75 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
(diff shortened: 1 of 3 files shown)


assets/src/main/java/bisq/asset/MileAddressValidator.java
```diff
@@ -0,0 +1,64 @@
...
+import org.bitcoinj.core.Base58;
+
+import java.util.Arrays;
+import java.util.zip.CRC32;
+import java.util.zip.Checksum;
+
+/**
+ * Mile address - base58(32 bytes of public key + 4 bytes of crc32)
+ */
+public class MileAddressValidator implements AddressValidator {
```
blabno: Make this inner class of Mile as it will not be used anywhere else.

blabno (changes requested):
NACK per comments

assets/src/main/java/bisq/asset/MileAddressValidator.java
vlad-khramov: Fixed

blabno (approved):
ACK

PR #2310, opened 2019-01-23
Fix trade currency handling for multiple currency payment accounts

### LINKED ISSUE #1753: Default Currency and Trading account error
Issued reproduced with two o more accounts with different currencies.

Steps to reproduce:
1. "Buy bitcoin" tab
2. "Create new offer to buy BTC with USD"
3. Select a Trading account that is not USD
4. Cancel the offer by clicking the x or "Cancel"
5. "Create new offer to buy BTC with USD"
At this point, the "Trading account" and "Currency" shoud be USD in the "Create offer" tab, but is not. This could produce a mistake we create a new offer.

Same issue in "Sell bitcoint" tab.

ripcurlx:
Fixes #1753.


ManfredKarrer (approved):
ACK

PR #2313, opened 2019-01-24
Fix incorrect time shown in phase duration

devinbileck:
The phase duration was showing the hour in am/pm (1-12) rather than the
hour in the day (0-23).


ripcurlx (approved):
utACK

PR #2315, opened 2019-01-24
Not disable the payment account selection in take offer screen

### LINKED ISSUE #2276: Multiple SEPA accounts, but cannot choose, when buying!
I would like to refer to this problem:

https://bisq.community/t/multiple-sepa-accounts-but-cannot-choose-when-buying/6925/2

I have multiple SEPA Options added which were on the same day and use the same bank, and it worked fine up until recently. Maybe someupdate caused the Issue.

Now I cannot chose which SEPA Option I want to buy with as it is greyed out and only one SEPA is prechosen.

Thanks

ripcurlx:
Fixes #2276.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferView.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferView.java
@@ -484,9 +484,11 @@ public class TakeOfferView extends ActivatableViewAndModel<AnchorPane, TakeOffer
 
         model.onShowPayFundsScreen();
 
-        paymentAccountsComboBox.setMouseTransparent(true);
-        paymentAccountsComboBox.setDisable(true);
-        paymentAccountsComboBox.setFocusTraversable(false);
+        if (DevEnv.isDaoActivated()) {
+            paymentAccountsComboBox.setMouseTransparent(true);
+            paymentAccountsComboBox.setDisable(true);
+            paymentAccountsComboBox.setFocusTraversable(false);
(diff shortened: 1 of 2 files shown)


desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferView.java
```diff
@@ -484,10 +484,6 @@ private void onShowPayFundsScreen() {
 
         model.onShowPayFundsScreen();
```
ManfredKarrer, line 486: Doesn't that this keep the combobox enabled in the DAO mode when at payment screen? Maybe it requires an if check ro the dao mode?

ManfredKarrer (approved):

desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferView.java
ripcurlx: If the DAO is activated there is the additional step where you select the trade fee currency (BTC or BSQ), so there is no need to be able to switch the payment account also in the show funds screen. Before the DAO, you switch directly to the Pay Funds screen if there is no range to choose from, so it wouldn't be possible to switch to another payment account. That's why I needed to add this check.
```
if (DevEnv.isDaoActivated()) {
            paymentAccountsComboBox.setMouseTransparent(true);
            paymentAccountsComboBox.setDisable(true);
            paymentAccountsComboBox.setFocusTraversable(false);
}
```

PR #2316, opened 2019-01-24
Not show warning popup for imported payment accounts

### LINKED ISSUE #1981: Security recommendation prompt is shown multiple times when importing multiple accounts
When importing accounts (either national currency or altcoins), the security recommendation prompt will be shown once for each imported account (multiple times if importing multiple accounts):
image

before it shows the final result prompt indicating which accounts were imported.

Also, the security recommendation prompt should be shown after the result prompt.

ripcurlx:
Fixes #1981.


ManfredKarrer (approved):
utACK

PR #2317, opened 2019-01-24
Improve resizing of buy and sell tables

### LINKED ISSUE #1587: Offer Book screen does not reflow contents as expected
## Repro steps

1. Open Bisq
2. Navigate to a screen other than Offer Book
3. Maximize the window
4. Navigate back to the Offer Book screen
5. Notice that contents do not reflow, as seen in the screenshot below:

image

## Workaround

Explicitly resize the window once again, while remaining on the Offer Book screen. Notice that contents do reflow as expected:

image

## Notes

There probably needs to be a check in the Offer Book's activation logic to see whether the window dimensions have changed, and to force a reflow. Not sure if it can be that simple, but seems logical.

There still seems to be a bit of wasted real estate even after the reflow occurs. Not strictly part of this bug, but worth mentioning it in context.

The ability to reflow at all on this screen is new in v0.7.0, so not too surprising that there are some rough edges to smooth out.

ripcurlx:
Fixes #1587.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java
+++ b/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java
@@ -90,6 +90,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.Optional;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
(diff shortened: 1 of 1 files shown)


ManfredKarrer (approved):
utACK

PR #2319, opened 2019-01-24
List Cash2 (CASH2)

cash2org:
Official project URL: https://cash2.org
Official block explorer URL: https://blocks.cash2.org

Cash2 is based on CryptoNote. Included is a pop up message to let CASH2 senders know what is required of them in the event a trade goes to arbitration.


blabno (approved):
ACK

PR #2336, opened 2019-01-29
Add LOCK_TIME_TRADE_PAYOUT and ARBITRATOR_FEE param

ManfredKarrer:
@sqrrm This requires careful testing to guarantee that we are backward compatible.

@ripcurlx Please leve merge to myself, so we have time to double check to not break anything....


sqrrm (approved):
ACK

This will make old clients unhappy with the new parameter, but since we're on testnet it's better to just launch and have DAO users upgrade quickly. Others shouldn't be affected anyway.

ManfredKarrer:
Will wait with merge after a bit more testing

PR #2344, opened 2019-01-30
Delete tor files at startup

ManfredKarrer:
We get sometimes Tor startup problems which is related to some tor
files in the tor directory. It happens more often if the application
got killed (not graceful shutdown).
Creating all tor files newly takes about 3-4 sec. longer and it does
not benefit from cache files.
TODO: We should fix those startup problems in the netlayer library,
once fixed there we can remove that call at the
Bisq startup again.


ripcurlx (approved):
utACK

freimair (approved):
utAck

ghost:
We just got one more Tor issue on the forum.

I wonder if there are not some companies working for govs, who use IPs addresses, do voluntarily bad things with those IPs over the Tor network, in order to have those IPs banned.
And so we have more and more IPs unusable.
... but maybe I'm paranoid.

PR #2345, opened 2019-01-30
List Augmint Euro (AEUR)

phraktle:
- Official project URL: https://www.augmint.org
- Official block explorer URL: N/A since it's an ERC-20 token, not a coin. The token contract is at: https://etherscan.io/token/0xc994a2dEb02543Db1f48688438b9903c4b305ce3

### DIFF
--- a/assets/src/main/java/bisq/asset/tokens/AugmintEuro.java
+++ b/assets/src/main/java/bisq/asset/tokens/AugmintEuro.java
@@ -1,0 +1,27 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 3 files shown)


ripcurlx (changes requested):
Being a little bit picky, but please rename your commit message to `List Augmint Euro (AEUR)` instead of `List Augmint Euro ([AEUR])` (see https://docs.bisq.network/exchange/howto/list-asset.html#step-5-create-a-well-formed-git-commit).
Besides that ACK from my side.

phraktle:
Commit message fixed, thanks @ripcurlx!

phraktle:
Btw, shouldn't ERC-20 tokens specify the token contract's address somewhere? I guess there's no usage of this in Bisq currently, but it could be useful on the UI.

ripcurlx (approved):

ripcurlx:
It is not a requirement IMHO if you just want to trade it. Just merged your PR.

phraktle:
Sure, it's just potentially very helpful for all parties involved (the recipient and in case of a dispute, the arbitrator needs to know definitively what the token address was supposed to be, so they are receiving the appropriate token, not something else named the same).

ripcurlx:
> Sure, it's just potentially very helpful for all parties involved (the recipient and in case of a dispute, the arbitrator needs to know definitively what the token address was supposed to be, so they are receiving the appropriate token, not something else named the same).

true - @ManfredKarrer Would it be a problem if we add the token address as another property to the ERC20Token class? This is not serialized as a ProtoBuffer message anyways, correct? Of course this would in the end also needed to be added in the trade process, in trade details,... as well. Could be a bounty that we put out to be taken.

ManfredKarrer:
@ripcurlx I don't see a problem to add that to the ERC20Token class

phraktle:
If you do decide to add this, the token address for AEUR is `0xc994a2dEb02543Db1f48688438b9903c4b305ce3`

PR #2346, opened 2019-01-30
Allow tor to be used with regtest

devinbileck:
In order to simulate a realistic environment for testing purposes, this will allow tor to be used with regtest. Use the bitcoinRegtestHost parameter and specify a tor .onion address as well as include the --useTorForBtc=true parameter.

### DIFF
--- a/core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
+++ b/core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
@@ -236,7 +236,6 @@ public class WalletsSetup {
             if (regTestHost == RegTestHost.LOCALHOST) {
                 walletConfig.setPeerNodesForLocalHost();
             } else if (regTestHost == RegTestHost.REMOTE_HOST) {
-                walletConfig.setMinBroadcastConnections(1);
                 configPeerNodesForRegTestServer();
             } else {
                 configPeerNodes(socks5Proxy);
@@ -315,7 +314,11 @@ public class WalletsSetup {
 
     private void configPeerNodesForRegTestServer() {
         try {
-            walletConfig.setPeerNodes(new PeerAddress(InetAddress.getByName(RegTestHost.HOST), params.getPort()));
+            if (RegTestHost.HOST.endsWith(".onion")) {
+                walletConfig.setPeerNodes(new PeerAddress(RegTestHost.HOST, params.getPort()));
+            } else {
+                walletConfig.setPeerNodes(new PeerAddress(InetAddress.getByName(RegTestHost.HOST), params.getPort()));
+            }
         } catch (UnknownHostException e) {
             log.error(e.toString());
             e.printStackTrace();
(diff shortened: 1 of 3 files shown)


core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
```diff
@@ -236,7 +236,6 @@ protected void onSetupCompleted() {
             if (regTestHost == RegTestHost.LOCALHOST) {
                 walletConfig.setPeerNodesForLocalHost();
             } else if (regTestHost == RegTestHost.REMOTE_HOST) {
-                walletConfig.setMinBroadcastConnections(1);
```
ManfredKarrer, line 239: Is that correct to remove that line?

core/src/main/java/bisq/core/user/Preferences.java
```diff
@@ -692,10 +692,12 @@ public boolean showAgain(String key) {
...
     public boolean getUseTorForBitcoinJ() {
-        // We override the useTorForBitcoinJ and set it to false if we detected a localhost node or if we are not on mainnet.
-        // At testnet there are very few Bitcoin tor nodes and we don't provide tor nodes.
-        if (!BisqEnvironment.getBaseCurrencyNetwork().isMainnet()
+        // We override the useTorForBitcoinJ and set it to false if we detected a localhost node or if we are not on mainnet,
+        // unless the useTorForBtc parameter is explicitly provided.
+        // On testnet there are very few Bitcoin tor nodes and we don't provide tor nodes.
+        if ((!BisqEnvironment.getBaseCurrencyNetwork().isMainnet()
                 || bisqEnvironment.isBitcoinLocalhostNodeRunning())
+                && bisqEnvironment.getProperty(BtcOptionKeys.USE_TOR_FOR_BTC).isEmpty())
```
ManfredKarrer: Maybe better to check against a boolean as USE_TOR_FOR_BTC can be set to false. A default values should be set anyway so an empty property should be not be the case if the property setup is done correctly.

ManfredKarrer:
@devinbileck In case you have not seen my comments, could you have a look?

devinbileck:
@ManfredKarrer I dont see any comments within this PR.

ManfredKarrer (changes requested):

core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
devinbileck: This is already being done on line 235 so I removed it.

core/src/main/java/bisq/core/user/Preferences.java
devinbileck: Good point. I will rework that.

devinbileck: After reviewing the code again, my initial implementation appears to be technically correct and attempts to determine if the USE_TOR_FOR_BTC parameter was explicitly provided or not by checking whether the property is empty (which it will be empty if not provided). And if it is provided, do not override it to false and instead use whatever value was provided. It handles each scenario where (1) the parameter is not provided, (2) provided and set to false, and (3) provided and set to true. I did not change the implementation/type of the USE_TOR_FOR_BTC as a string parameter to boolean in order to keep it consistent with other similar parameters. If that needs to change, that can be done separately. The only thing I identified was that I should have instead used the useTorFlagFromOptions parameter rather than bisqEnvironment.getProperty. I will commit a fix for that.

ManfredKarrer (approved):
utACK

PR #2348, opened 2019-01-31
Monitor release: Showing off

freimair:
see https://github.com/bisq-network/proposals/issues/62 for details

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -322,6 +322,8 @@ configure(project(':monitor')) {
     }
 
     dependencies {
+        compile project(':p2p')
+        compile project(':core')
         compile "org.slf4j:slf4j-api:$slf4jVersion"
         compile "ch.qos.logback:logback-core:$logbackVersion"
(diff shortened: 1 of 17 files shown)


ripcurlx (approved):
utACK

ManfredKarrer (commented):
In P2PNetworkLoad.java:

Why not use 
`buckets.putIfAbsent(className, new Counter());
buckets.get(className).increment();`

instead of 
`try {
                    buckets.get(className).increment();
                } catch (NullPointerException nullPointerException) {
                    buckets.put(className, new Counter());
                }`

ManfredKarrer:
In P2PNetworkLoad.java:
Re:
`bucketsPerHost.put(connection.peersNodeAddressProperty().getValue(), buckets);`
It can be that connection.peersNodeAddressProperty() is not set yet. Maybe more safe if you add a check if it is nul null.

ManfredKarrer (approved):
utACK

PR #2354, opened 2019-02-02
Navigate to bond

sqrrm:
I added the option to set extra data when navigating between tabs as I've seen the need before. I had a look at PortfolioView to see how the data was shared there as suggested by @ripcurlx but that way didn't seem to work so nicely here.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/Navigation.java
+++ b/desktop/src/main/java/bisq/desktop/Navigation.java
@@ -38,12 +38,14 @@ import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 
+import javax.annotation.Nullable;
+
 @Slf4j
 public final class Navigation implements PersistedDataHost {
     private static final ViewPath DEFAULT_VIEW_PATH = ViewPath.to(MainView.class, MarketView.class);
 
     public interface Listener {
-        void onNavigationRequested(ViewPath path);
+        void onNavigationRequested(ViewPath path, @Nullable Object data);
     }
 
     // New listeners can be added during iteration so we use CopyOnWriteArrayList to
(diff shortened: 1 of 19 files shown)


desktop/src/main/java/bisq/desktop/Navigation.java
```diff
@@ -38,12 +38,14 @@
...
 
+import javax.annotation.Nullable;
+
 @Slf4j
 public final class Navigation implements PersistedDataHost {
     private static final ViewPath DEFAULT_VIEW_PATH = ViewPath.to(MainView.class, MarketView.class);
 
     public interface Listener {
-        void onNavigationRequested(ViewPath path);
+        void onNavigationRequested(ViewPath path, @Nullable Object data);
```
ManfredKarrer, line 48: I think its better to make a second method with the additional param.

```  
public interface Listener {
        default void onNavigationRequested(ViewPath path){
            onNavigationRequested(path, null);
        }
        default void onNavigationRequested(ViewPath path, @Nullable Object data){
            onNavigationRequested(path);
        }
    }
```

Then you can avoid all those changed in clients where you pass null. Using `default` does not force implementing clients to define both methods.

desktop/src/main/java/bisq/desktop/Navigation.java
```diff
@@ -93,10 +95,15 @@ public void readPersisted() {
...
+        navigateTo(ViewPath.to(viewClasses), null);
+    }
+
+    @SafeVarargs
+    public final void navigateToWithData(Object data, Class<? extends View>... viewClasses) {
+        navigateTo(ViewPath.to(viewClasses), data);
     }
 
-    public void navigateTo(ViewPath newPath) {
+    public void navigateTo(ViewPath newPath, Object data) {
```
ManfredKarrer, line 106: Can you add @Nullable

ManfredKarrer (changes requested):

ManfredKarrer:
I added a now PR (https://github.com/bisq-network/bisq/pull/2366) based on yours.

PR #2356, opened 2019-02-03
Add new btc node from KanoczTomas

ManfredKarrer:
onion address: mbm6ffx6j5ygi2ck.onion
ipv4 address: 193.58.196.212
dns: btc.ispol.sk


KanoczTomas (approved):
Double checked info, it is correct!

PR #2360, opened 2019-02-03
Dao fix issues with majority hash

ManfredKarrer:


core/src/main/java/bisq/core/dao/node/BsqNode.java
```diff
@@ -63,7 +63,11 @@
     @Nullable
     protected Consumer<String> warnMessageHandler;
     protected List<RawBlock> pendingBlocks = new ArrayList<>();
+
     // The chain height of the latest Block we either get reported by Bitcoin Core or from the seed node
+    // This property should not be used in consensus code but only or retrieving blocks as it is not in sync with the
```
sqrrm:
```suggestion
    // This property should not be used in consensus code but only for retrieving blocks as it is not in sync with the
```

core/src/main/java/bisq/core/dao/governance/votereveal/VoteRevealService.java
```diff
@@ -161,40 +161,43 @@ public void addVoteRevealTxPublishedListener(VoteRevealTxPublishedListener voteR
...
-    // otherwise his vote becomes invalid and his locked stake will get unlocked
-    private void maybeRevealVotes() {
-        // We must not use daoStateService.getChainHeight() because that gets updated with each parsed block but we
-        // only want to publish the vote reveal tx if our current real chain height is matching the cycle and phase and
-        // not at any intermediate height during parsing all blocks. The bsqNode knows the latest height from either
-        // Bitcoin Core or from the seed node.
-        int chainHeight = bsqNode.getChainTipHeight();
+    // otherwise his vote becomes invalid.
+    // In case the user miss the vote reveal phase an (invalid) vote reveal tx will be created the next time the user is
+    // online. That tx only serves the purpose to unlock the stake from teh blind vote but it will be ignored for voting.
```
sqrrm:
```suggestion
    // online. That tx only serves the purpose to unlock the stake from the blind vote but it will be ignored for voting.
```

core/src/main/java/bisq/core/dao/governance/votereveal/VoteRevealService.java
```diff
@@ -161,40 +161,43 @@ public void addVoteRevealTxPublishedListener(VoteRevealTxPublishedListener voteR
...
                 .forEach(myVote -> {
                     boolean isInVoteRevealPhase = periodService.getPhaseForHeight(chainHeight) == DaoPhase.Phase.VOTE_REVEAL;
                     boolean isBlindVoteTxInCorrectPhaseAndCycle = periodService.isTxInPhaseAndCycle(myVote.getTxId(), DaoPhase.Phase.BLIND_VOTE, chainHeight);
                     if (isInVoteRevealPhase && isBlindVoteTxInCorrectPhaseAndCycle) {
+                        log.info("We call revealVote at blockHeight {} for blindVoteTxId {}", chainHeight, myVote.getTxId());
                         // Standard case that we are in the correct phase and cycle and create the reveal tx.
-                        revealVote(myVote);
+                        revealVote(myVote, true);
                     } else {
+                        // We missed the vote reveal phase but publish a vote reveal tx to unlock he blind vote stake.
```
sqrrm:
```suggestion
                        // We missed the vote reveal phase but publish a vote reveal tx to unlock the blind vote stake.
```

core/src/main/java/bisq/core/dao/governance/votereveal/VoteRevealService.java
```diff
@@ -213,31 +216,29 @@ private void maybeRevealVotes() {
...
     }
 
-    private void revealVote(MyVote myVote) {
+    private void revealVote(MyVote myVote, boolean inBlindVotePhase) {
         try {
             // We collect all valid blind vote items we received via the p2p network.
             // It might be that different nodes have a different collection of those items.
             // To ensure we get a consensus of the data for later calculating the result we will put a hash of each
-            // voters  blind vote collection into the opReturn data and check for a majority at issuance time.
+            // voters  blind vote collection into the opReturn data and check for a majority at the vote result phase.
```
sqrrm:
```suggestion
            // voter's blind vote collection into the opReturn data and check for a majority in the vote result phase.
```

core/src/main/java/bisq/core/dao/governance/voteresult/VoteResultException.java
```diff
@@ -28,17 +28,17 @@
 
 public class VoteResultException extends Exception {
     @Getter
-    private final Cycle cycle;
+    private final int heightOfFirstBlock;
```
sqrrm, line 31: The name `heightOfFirstBlock` is a bit confusing without the context that it's the first block in the cycle

sqrrm (approved):
utACK

Added some fixes to spelling mistakes and variable naming.

Otherwise this looks better, waiting for the end of batch parsing and checking the current phase at that height before publishing vote. I wonder if our usage of chainHeight, blocHeight, height and what not is lacking since we're getting confused. Perhaps better naming would help us understand better what we're doing.

core/src/main/java/bisq/core/dao/governance/voteresult/VoteResultException.java
ManfredKarrer: True. Will update wit next PR....

ManfredKarrer:
Yes we should rename once blockHeight params to be always the same.

sqrrm:
> Yes we should rename once blockHeight params to be always the same.

I think we might need to differentiate between parsed height, known last block height, height for a certain block and perhaps even more and use different but consistent names.

PR #2363, opened 2019-02-03
Dao: Add handler for republishing all dao governance data, improve vote result handling 

ManfredKarrer:


ManfredKarrer:
@sqrrm I will rebase after i merged master with the other 2 PRs...

core/src/main/java/bisq/core/dao/governance/voteresult/MissingDataRequestService.java
```diff
@@ -51,4 +76,39 @@ public void start() {
...
+    public void reRepublishAllGovernanceData() {
+        ObservableList<ProposalPayload> proposalPayloads = proposalService.getProposalPayloads();
+        proposalPayloads.forEach(proposalPayload -> {
+            // We want a random delay between 0.1 and 30 sec. depending on the number of items
+            int delay = Math.max(100, Math.min(30_000, new Random().nextInt(proposalPayloads.size() * 500)));
+            UserThread.runAfter(() -> {
+                boolean success = p2PService.addPersistableNetworkPayload(proposalPayload, true);
+                String txId = proposalPayload.getProposal().getTxId();
+                if (success) {
+                    log.warn("We received a RepublishGovernanceDataRequest and re-published a proposalPayload to " +
```
sqrrm, line 89: Maybe this shouldn't be warning level log

core/src/main/java/bisq/core/dao/governance/voteresult/MissingDataRequestService.java
```diff
@@ -51,4 +76,39 @@ public void start() {
...
+        ObservableList<BlindVotePayload> blindVotePayloads = blindVoteListService.getBlindVotePayloads();
+        blindVotePayloads
+                .forEach(blindVotePayload -> {
+                    // We want a random delay between 0.1 and 30 sec. depending on the number of items
+                    int delay = Math.max(100, Math.min(30_000, new Random().nextInt(blindVotePayloads.size() * 500)));
+                    UserThread.runAfter(() -> {
+                        boolean success = p2PService.addPersistableNetworkPayload(blindVotePayload, true);
+                        String txId = blindVotePayload.getBlindVote().getTxId();
+                        if (success) {
+                            log.warn("We received a RepublishGovernanceDataRequest and re-published a blindVotePayload to " +
```
sqrrm, line 106: Same here, I don't think warning level is right

sqrrm (approved):
utACK

The comments on log level stand but it's just refactoring so can be tended to in another PR

core/src/main/java/bisq/core/dao/governance/voteresult/MissingDataRequestService.java
ManfredKarrer: Yes was also wondering. But it make it more visible if that occurres so that anyone can see if any peer is missing data. Basically the P2P network should be reliable enough to not cause that.

core/src/main/java/bisq/core/dao/governance/voteresult/MissingDataRequestService.java
ManfredKarrer: Ah, I see now, its at each success case, thats definitely too much. Will fix it in next PR.

PR #2364, opened 2019-02-03
[WIP] Dao avoid bsq burn

ManfredKarrer:


ManfredKarrer:
@sqrrm I changed the tx to IRREGULAR and added some more changes. Tested a bit with proposals, blind vote and vote reveal by delaying the tx broadcast and increasing the blocks in between so that the phase in not valid anymore at the tx confirmation block. Blind votes show irregular state, the others not but that might be ok as long they are not burned. A blind vote did not get unlocked though if no vote reveal followed (maybe the vote reveal tx which we make to unlock did not gor created as the txType is now irregular instead of blind vote?). Was also wondering if we should better keep the original type and add a new boolean flag which marks a tx invalid for dao rules but the BSQ still spendable....
I leave the PR WIP as I am not sure if the additional type is really the best way (and once we add it we cannot remove it anymore).

PR #2368, opened 2019-02-04
Remove libdohj dependency

oscarguindzberg:
Remove libdohj dependency and the code to use LTC, DASH or DOGE as base currencies.


core/src/main/java/bisq/core/app/WalletAppSetup.java
```diff
@@ -90,8 +88,8 @@ void init(@Nullable Consumer<String> chainFileLockedExceptionHandler,
               Runnable walletPasswordHandler,
               Runnable downloadCompleteHandler,
               Runnable walletInitializedHandler) {
-        log.info("Initialize WalletAppSetup with BitcoinJ version {} and LibDohJ version {} with hash of BitcoinJ commit {}",
-                VersionMessage.BITCOINJ_VERSION, Version.VERSION, Version.BITCOINJ_VERSION);
+        log.info("Initialize WalletAppSetup with BitcoinJ version and hash of BitcoinJ commit {}",
```
ManfredKarrer: I think here is missing the {} -> BitcoinJ version {} and

ManfredKarrer (changes requested):

core/src/main/java/bisq/core/app/WalletAppSetup.java
ManfredKarrer: Beside that, ACK (just tested).

ripcurlx (changes requested):
ACK - besides the small change requested by @ManfredKarrer

core/src/main/java/bisq/core/app/WalletAppSetup.java
oscarguindzberg: Done. just force pushed the fix.

ManfredKarrer (approved):
utACK

PR #2370, opened 2019-02-05
DAO:Use highlight color without link behavior

### LINKED ISSUE #2139: DAO: Some table headings act like links (color change on hover) but do not load the browser when clicked
image

image

ripcurlx:
Fixes #2139.


ManfredKarrer (approved):
utACK

PR #2372, opened 2019-02-05
Show info if not in proposal phase

### LINKED ISSUE #2087: DAO: Inform user that proposals cannot be submitted while not in the proposal phase
When not in the proposal phase, you are unable to make a proposal (the proposal type dropdown is disabled).
image

As a usability improvement, it would be nice to show a message informing the user that they have to wait for the next proposal phase to submit a new proposal.

ripcurlx:
Fixes #2087.

**In make-proposal-phase**
bildschirmfoto 2019-02-05 um 14 55 25
**Not in make-proposal-phase**
bildschirmfoto 2019-02-05 um 14 55 17


desktop/src/main/java/bisq/desktop/main/dao/governance/make/MakeProposalView.java
```diff
@@ -178,19 +205,44 @@ protected void deactivate() {
...
-        boolean isProposalPhase = daoFacade.isInPhaseButNotLastBlock(DaoPhase.Phase.PROPOSAL);
-        proposalTypeComboBox.setDisable(!isProposalPhase);
-        if (!isProposalPhase)
+        isProposalPhase.set(daoFacade.isInPhaseButNotLastBlock(DaoPhase.Phase.PROPOSAL));
+        if (isProposalPhase.get()) {
+            proposalGroupTitle.set(Res.get("dao.proposal.create.selectProposalType"));
+        } else {
+            proposalGroupTitle.set(Res.get("dao.proposal.create.phase.inactive"));
             proposalTypeComboBox.getSelectionModel().clearSelection();
+            updateTimeUntilNextProposalPhase(height);
```
ManfredKarrer, line 240: Here u get the current cycle so it will show the past proposal phase.
screen shot 2019-02-05 at 15 07 44
Either we add the duration of a full cycle, or we show just a diff. text and pointing to the dashboard to see when phases start and end (i think thats better as there its easier to understand if u see all phases).

ManfredKarrer (changes requested):

core/src/main/java/bisq/core/dao/DaoUtil.java
```diff
@@ -0,0 +1,43 @@
...
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.core.dao;
```
ManfredKarrer, line 18: Lets move that in a `presentation` package. We use that aleady in other domains for UI sepefic stuff which is in core only for sharing with API. Actaully in that case its not clear if we will need that for the API so it could be left in the desktop module as well.

desktop/src/main/java/bisq/desktop/main/dao/governance/make/MakeProposalView.java
ripcurlx: Ah - missed that. Didn't read the content of the proposal phase field properly 😬. I just wanted to give the user directly the information when the next proposal phase will be active, so he/she can make a calendar entry or similar. I'm personally more for adding a full cycle duration and show the correct date (without duration in blocks in the front). I'll give it a quick shot how it would look like.

ripcurlx:
I've added the current cycle time to calculate the next phase. Of course this might be not always 100% correct if we change the phase duration for the next cycle, but at least the starting block should always be correct.
bildschirmfoto 2019-02-05 um 15 52 46

ManfredKarrer (approved):
ACK

core/src/main/java/bisq/core/dao/DaoUtil.java
ripcurlx: I wasn't sure if it will be needed in the API as well. But I'll move it into the desktop module then.

ManfredKarrer:
Pushed https://github.com/bisq-network/bisq/commit/a4428ffc065f8e88c81f7dc02e814d055b03d53a

PR #2377, opened 2019-02-05
Fix missing removal of old or invalid proposals

ManfredKarrer:
- If a proposal had a tx which is not in our BSQ state it was not removed
- If a tx was outdated (> 60 days old) it was only removed if we are
in proposal phase. We added check for past cycle to cover those as well.


ripcurlx (approved):
utACK

ManfredKarrer:
@sqrrm Could u have a look as well?

sqrrm (approved):
utACK

Issue #568, opened 2016-08-05
Snap Package


ksthiele:
Offering a snap package of bitsquare would be very nice and would make it compatible to the most important Linux distros.

dmp1ce:
I am working on a Snap package here: https://github.com/dmp1ce/bitsquare-snap

Right now it gets the error `Segmentation fault (core dumped)` but it does build. I'm not sure how to debug the `Segmentation fault (core dumped)` error to see what is wrong with the package.

Here is more information on debugging a Snap package. It may be that the application needs to be modified to work as a Snap. https://developer.ubuntu.com/en/snappy/build-apps/debug/

The core dump on my machine doesn't reveal much information. Here is the core dump:

```
> coredumpctl info -1
           PID: 23710 (java)
           UID: 1000 (dmp1ce)
           GID: 100 (users)
        Signal: 11 (SEGV)
     Timestamp: Thu 2016-09-08 10:38:56 EDT (12min ago)
  Command Line: /snap/bitsquare/x1/usr/lib/jvm/default-java/bin/java -Dsecurity.provider.11=org.bouncycastle.jce.provider.BouncyCastleProvider -jar /snap/bitsquare/x1/gui/shaded.jar
    Executable: /snap/bitsquare/x1/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
 Control Group: /
         Slice: -.slice
       Boot ID: dbc7e03954374d25adc0810dce5ddaa5
    Machine ID: 1e4f9ee982834cb9a8fd87bbe38278eb
      Hostname: samus
      Coredump: /var/lib/systemd/coredump/core.java.1000.dbc7e03954374d25adc0810dce5ddaa5.23710.1473345536000000000000.lz4
       Message: Process 23710 (java) of user 1000 dumped core.

                Stack trace of thread 23731:
                #0  0x00007ff356413754 n/a (/lib/x86_64-linux-gnu/libc-2.23.so)
```

dmp1ce:
I published a Snap which should work. Please send bugs for this Snap my way.

https://github.com/dmp1ce/bisq-snap
https://snapcraft.io/bisq-desktop

ghost:
Hello,
Did you test the install yourself ?
How does it deal with the signatures ?

dmp1ce:
Yes, I tested the install myself.
[...]

PR #2200, opened 2019-01-04
Generate RPM package

devinbileck:
An RPM package is now generated as part of the Linux packaging script.
The packaging script can be run on any Debian or Redhat based
system and will generate both a deb and rpm package.

Resolves https://github.com/bisq-network/bisq/issues/401

PR #2378, opened 2019-02-06
Initial commit of Snap package

dmp1ce:
Adds snap package code which will allow Snap packages to be automatically built and hosted on the Snapcraft store.

#568


snap/snapcraft.yaml
```diff
@@ -0,0 +1,81 @@
...
+      - network
+      - network-bind
+      - desktop
+      - desktop-legacy
+      - x11
+      - wayland
+      - home
+    desktop: usr/share/applications/bisq.desktop
+    environment:
+      JAVA_HOME: "$SNAP/usr/lib/jvm/java-11-openjdk-amd64"
```
ripcurlx, line 24: Bisq should be built with Java 10. Java 11 isn't officially supported yet.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,81 @@
...
+      # Install executable.
+      install -d "${SNAPCRAFT_PART_INSTALL}/opt/bisq"
+      cp -r desktop/build/app/* "${SNAPCRAFT_PART_INSTALL}/opt/bisq"
+
+      # Install desktop launcher icon
+      install -Dm644 desktop/package/linux/icon.png "${SNAPCRAFT_PART_INSTALL}/usr/share/pixmaps/bisq.png"
+
+    source: https://github.com/bisq-network/bisq-desktop.git
+    source-tag: v0.9.3
+    build-packages: [openjdk-11-jdk]
```
ripcurlx, line 42: Bisq should be built with Java 10. Java 11 isn't officially supported yet.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,81 @@
...
+    stage-packages: [openjdk-11-jre]
```
ripcurlx, line 43: Bisq should be built with Java 10. Java 11 isn't officially supported yet.

ripcurlx (changes requested):
NACK - see my comments below. Unfortunately I don't have time to evaluate if we want to "offically" support Snap packages. @devinbileck do you have time to look into this if we want to add this kind of support? Thanks!

snap/snapcraft.yaml
dmp1ce: I think `java-11-jdk` is Java 10 on Ubuntu. I have also learned that a Snap package can be created from the `.deb` if that would be better.

snap/snapcraft.yaml
dmp1ce: I think `java-11-jdk` is Java 10 on Ubuntu. I have also learned that a Snap package can be created from the `.deb` if that would be better.

snap/snapcraft.yaml
dmp1ce: I think `java-11-jdk` is Java 10 on Ubuntu. I have also learned that a Snap package can be created from the `.deb` if that would be better.

snap/snapcraft.yaml
dmp1ce: Ubuntu 18.04 is still on Java 10 it seems. https://askubuntu.com/a/1037655/25776

snap/snapcraft.yaml
dmp1ce: Ubuntu 18.04 is still on Java 10 it seems. https://askubuntu.com/a/1037655/25776

snap/snapcraft.yaml
dmp1ce: Ubuntu 18.04 is still on Java 10 it seems. https://askubuntu.com/a/1037655/25776

devinbileck:
To be honest, I am unfamiliar with snap packages and this is my first time looking into it. The more I look into it, the less inclined I am to suggest the snapcraft store as a way to distribute Bisq (or any Bitcoin application for that matter) as I have not found a way to verify the application against our signature files (and ideally it should be verified before installing via snap in the first place). @HarryMacfinned highlighted these same concerns in https://github.com/bisq-network/bisq/issues/568.

Up until this point, we have only provided DEB packages with our releases which is serving only a subset of the Linux community, and with our next release we will be able to provide an RPM package (see https://github.com/bisq-network/bisq/pull/2200), which will help cover a wider audience.

We could potentially provide a snap file with our releases (along with signature file) as it can be installed locally. For example (as per https://docs.snapcraft.io/java-applications/7819):
> sudo snap install bisq-desktop.snap --devmode --dangerous

But this kind of defeats the purpose of using snaps as an easy install/update app-store.

However, since snaps can be used on all major Linux distributions (according to https://docs.snapcraft.io/installing-snapd/6735), providing a snap package would allow us to provide an install package for the majority of the Linux community.

But to be honest, this may not warrant the added effort as it is likely the rest of the Linux community is skilled enough to just build the application from source.

ripcurlx:
NACK - Based on the concerns by @devinbileck and @HarryMacfinned I won't merge it into master. @dmp1ce Sorry for the efforts you put into this. Maybe you can use this on your own fork, but we won't add this as an official distribution channel yet.

dmp1ce:
Snaps installed from the Snapcraft store can be verified by an end user with a signature or checksum easily because the `.snap` files are all stored in `/var/lib/snapd/snaps` and mounted as read-only for the system to use.

The `.snap` and signature could be added to the releases page on Github and `.snap` file could be distributed on the Snapcraft store. The validity of the `.snap` can be verified if one so chooses either way the user gets the `.snap`.

devinbileck:
I suppose that may be a viable solution. But I guess it all depends if building the snap package is deterministic and yields the exact same file between a locally generated package and one published to the snapcraft store. Do you know if that is the case?

@freimair how is bisq distributed via the Arch platform? I am not familiar with the platform or process. Is a compiled package provided and if so how do you address these concerns?

dmp1ce:
I don't know if a snap build is deterministic but I'll try to find out.

I help @freimair maintain the Arch Linux Bisq packages.

There are scripts for installing both the binary and the source compiled version of bisq. https://aur.archlinux.org/packages/?O=0&K=bisq Arch Linux users can also install snapd and install the Bisq snap if they choose.

The typical way Arch Linux users address these concerns is either relying on the Arch Linux official maintainers to package the application as a binary or use a package on the AUR. If using a package on the AUR it is recommended to look over the PKGBUILD, which is a bash script, before using it to install the application. Many PKGBUILD scripts install from source or official binaries.

dmp1ce:
I just built the Bisq snap again and got a different sha256 hash. So, I'm going to say no. The build is not deterministic, at least not the way I am building the snap.

devinbileck:
When you upload a locally generated snap file to the snapcraft store, does the hash of the snap that actually gets installed match? If so, that may be a viable approach as opposed to having it automatically built and published out of our control.

devinbileck:
Another item to consider is the update check mechanism within the app. If installed via snap, does it still make sense to download the install files? Should it instead show a prompt to upgrade via snap? If so, that means it would bypass the automatic signature verification portion after download and would be up to the user to self verify it.

dmp1ce:
Great questions. I think the hashes do match after going through the Snapcraft store but I haven't verified yet.

I wasn't aware that Bisq had an automatic signature process. Snaps update automatically which has some people in the community concerned. I need to learn more about how Bisq and Snaps update to see if they are compatible.

For Arch Linux packages I usually ignore update messages from apps and just install the latest package from the AUR. The Bisq AUR package does the verification every time.

dmp1ce:
I don't know if it is of interest here but the Snapcraft community has been discussing how "refreshing" apps should work for over a year. https://forum.snapcraft.io/t/bug-saves-are-blocked-to-snap-user-data-if-snap-updates-when-it-is-already-running/3226/16

Issues around data consistency, user feedback and you bring up good points about security. It might be that the Bisq application could ask snapd to refresh Bisq and the validated current version of Bisq could verify the hash from the Snapcraft store. I'll keep an eye on it.

devinbileck:
Thanks. I have added comments to the proposal and we can continue discussion there.
https://github.com/bisq-network/proposals/issues/70

PR #2379, opened 2019-02-06
Show BSQ trading fee more explicitly when funding your offer

### LINKED ISSUE #2075: DAO: Point out BSQ separately in "Fund your offer" popup
We do show the trading fee when paid in BSQ already in BSQ, but we show on top only the total BTC equivalent. This was a little bit confusing for me before reading through the whole popup. It should say "You need to deposit 0.01011325 BTC and 0.99 BSQ to this offer."

ripcurlx:
- Fixes #2075.
- Fixes wrong total funds value in info popup and funds tooltip
- Fixes missing info icon in Funds needed field
- Fixes positioning of info icons

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -373,6 +373,7 @@ createOffer.fundsBox.networkFee=Mining fee
 createOffer.fundsBox.placeOfferSpinnerInfo=Offer publishing is in progress ...
 createOffer.fundsBox.paymentLabel=Bisq trade with ID {0}
 createOffer.fundsBox.fundsStructure=({0} security deposit, {1} trade fee, {2} mining fee)
+createOffer.fundsBox.fundsStructure.BSQ=({0} security deposit, {1} mining fee) + {2} trade fee
 createOffer.success.headline=Your offer has been published
 createOffer.success.info=You can manage your open offers at \"Portfolio/My open offers\".
 createOffer.info.sellAtMarketPrice=You will always sell at market price as the price of your offer will be continuously updated.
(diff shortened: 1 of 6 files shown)


core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -373,6 +373,7 @@ createOffer.fundsBox.networkFee=Mining fee
 createOffer.fundsBox.placeOfferSpinnerInfo=Offer publishing is in progress ...
 createOffer.fundsBox.paymentLabel=Bisq trade with ID {0}
 createOffer.fundsBox.fundsStructure=({0} security deposit, {1} trade fee, {2} mining fee)
+createOffer.fundsBox.fundsStructure.BSQ=({0} security deposit, {1} mining fee) + {2} trade fee
```
sqrrm, line 376: Maybe this should keep the same order (deposit, trade fee, mining fee) as `createOffer.fundsBox.fundsStructure` for less confusion?

sqrrm (approved):
utACK

core/src/main/resources/i18n/displayStrings.properties
ripcurlx: The reason why I did it like that is, that it would otherwise look like as if it is part of the total btc funds needed.
bildschirmfoto 2019-02-11 um 09 54 54
So I implemented it as follows:
bildschirmfoto 2019-02-11 um 09 56 30

sqrrm: Ah, perhaps the ordering should be deposit, mining, trade fee for both?

ripcurlx: On first sight yes, but we have this order on a couple of places and it makes sense from my point of view to have it as it is right now:
bildschirmfoto 2019-02-11 um 15 12 05

bildschirmfoto 2019-02-11 um 15 11 08

bildschirmfoto 2019-02-11 um 15 14 44

The only problem is to mix it in the total funds field with the BSQ value.
I personally would  go with this little inconsistency and have the order in this field different between BSQ and BTC trading fees.

sqrrm: Yeah, that's a lot to change, then just go with your original suggestion unless we have an obviously better way to do it.

PR #2384, opened 2019-02-07
Wait until proposal is published before resetting the form

### LINKED ISSUE #2367: DAO: Add a pop up asking users to wait for the proposal confirmation
I noticed that there is a delay of 10-15 sec between making a proposal and getting its confirmation in DAO on testnet. It is not clear if a user should wait or go ahead and do something else. 

Suggestion: add a pop-up window that says something like "Please wait for the confirmation of your proposal" while the proposal transaction is being confirmed.

ripcurlx:
Fixes #2367.


desktop/src/main/java/bisq/desktop/main/dao/governance/make/MakeProposalView.java
```diff
@@ -300,19 +305,29 @@ private void publishMyProposal(ProposalType type) {
...
+                        proposalDisplay.clearForm();
+                    proposalTypeComboBox.getSelectionModel().clearSelection();
+                    busyAnimation.stop();
+                    busyLabel.setVisible(false);
+                    makeProposalButton.setDisable(false);
                 },
-                errorMessage -> new Popup<>().warning(errorMessage).show());
+                errorMessage -> {
+                    new Popup<>().warning(errorMessage).show();
+                    makeProposalButton.setDisable(false);
```
sqrrm, line 351: By not disabling the busyAnimation/button here it will keep spinning after an error was reported, I don't think that's correct.

sqrrm (commented):
See comment on animation on error.  Apart from that it looks good.

desktop/src/main/java/bisq/desktop/main/dao/governance/make/MakeProposalView.java
ripcurlx: Yes, it also should be reset in case of an error. Just pushed the commit.

PR #2386, opened 2019-02-07
Show warning for bonded role if user has insufficient funds

### LINKED ISSUE #2307: Validate for sufficient BSQ balance when proposing bonded role
I just made a proposal for a bonded role (rocket.chat admin, 20 000 BSQ). I only had ~400 BSQ in my wallet.

The transaction went through without any errors. This shouldn't be possible, right?

[Here is the transaction](https://explorer.bisq.network/tx.html?tx=aff3b311cc2fdc923fcec65a0737d77c1bd3577d85a93961323b5b1a1be65a02).

ripcurlx:
Fixes #2307.


sqrrm (approved):
ACK

This is better than not showing a warning but since it uses getAvailableBalance from bsqWalletService it will only know about acked funds and would warn if some BSQ are pending. I don't have a solution for that yet though, it turned out to be not trivial.

PR #2359, opened 2019-02-03
List CRowdCLassic (CRCL)

cryptocrowdclub:
- Official project URL: https://github.com/CRowdCLassic/crowdclassic-core
- Official block explorer URL: http://explorer.crcl.cryptocrowdclub.com:3001

PR #2387, opened 2019-02-07
Revert "List CRowdCLassic (CRCL)"

cryptocrowdclub:
Reverts bisq-network/bisq#2359

Wrong address validator.


cryptocrowdclub:
There`s new commit with right one within cryptocrowdclub:master repository:
https://github.com/cryptocrowdclub/bisq/commit/90871ea13b8faf57e66c4d5df6480fa1176fe568
How shall we proceed?

blabno:
Instead of adding "revert" code, please add code that fixes the existing validator.

cryptocrowdclub:
> Instead of adding "revert" code, please add code that fixes the existing validator.
Is this enough: https://github.com/cryptocrowdclub/bisq/commit/90871ea13b8faf57e66c4d5df6480fa1176fe568

PR #2388, opened 2019-02-07
Add indication that BSQ transactions are not up to date

### LINKED ISSUE #2303: Add indication that BSQ transactions are not up to date
When BSQ transactions are not up to date, the only indication of this is in the BSQ Wallet > Transactions tab. It may be beneficial to indicate this in the status bar at the bottom of the UI somehow. Perhaps just an icon that when clicked takes you to the BSQ Wallet > Transactions tab to see the status.
image

ripcurlx:
Fixes #2303.


core/src/main/java/bisq/core/dao/DaoSetup.java
```diff
@@ -34,22 +36,44 @@
...
 import java.util.function.Consumer;
 
+import lombok.Getter;
+
 /**
  * High level entry point for Dao domain.
  * We initialize all main service classes here to be sure they are started.
  */
-public class DaoSetup {
+public class DaoSetup implements DaoStateListener {
```
sqrrm: I think it would be better to make a separate BSQSyncService or something like that, it feels wrong to add this to the DaoSetup.

sqrrm (approved):
ACK

I have tested and it works, but I still think it would be better with a separate service for the sync handling to not pollute the setup class.

core/src/main/java/bisq/core/dao/DaoSetup.java
ManfredKarrer: I would prefer to not use DaoSetup for that but DaoFacade (or some custom class). I will have a look soon into it to make a more concrete sugestion.

ManfredKarrer: NACK

ripcurlx: Yes, I also wasn't quite happy to put it in this place, I just though to make it consistent how it is handled with the btcInfo WalletSetup right now:
https://github.com/bisq-network/bisq/blob/8cdc7599031a159b5c311904299136ba14b8a0a0/core/src/main/java/bisq/core/app/BisqSetup.java#L341-L348
https://github.com/bisq-network/bisq/blob/8cdc7599031a159b5c311904299136ba14b8a0a0/core/src/main/java/bisq/core/app/WalletAppSetup.java#L68

How do we differentiation between a Facade and a Service atm in Bisq?
I'll have a look to find a better place for this presentation logic. I might change the implementation of the btcInfo/btcSyncProgress on the way as well.

ripcurlx: I moved the code into the DaoPresentation class which I created already for another DAO related PR.

Issue #2398, opened 2019-02-08
Application dies shortly after launch while attempting to connect with Tor with error "Auth cookie not created"


devinbileck:
When launching Bisq on my Windows 10 VM, it dies several seconds later and doesn't get past this on the splash screen:
![image](https://user-images.githubusercontent.com/603793/52500223-53a43200-2b92-11e9-8bf9-158e1b611974.png)

In the log:
`Feb-08 10:59:37.065 [NetworkNode-9999] ERROR b.n.p.n.TorNetworkNode: Could not connect to running Tor: java.io.IOException: Auth cookie not created`

I am consistently encountering it on my VM, but not on my primary Windows 10 machine. I have investigated/eliminated possible causes such as firewall, antivirus, file permissions and as of yet been unable to determine the root cause.

I see it is creating an empty tor/.tor/control_auth_cookie file, but not writing to it. While looking at the netlayer module, it throws this exception if the file is not written to within 3 seconds.
https://github.com/JesusMcCloud/netlayer/blob/master/tor.native/src/main/kotlin/org/berndpruenster/netlayer/tor/TorContext.kt#L346

At this point the only thing I can think of is it perhaps could be a timing issue with my VM running slower than my primary machine and it taking longer than 3 seconds to write to this file. But I am not familiar with the netlayer module nor with tor in general.

devinbileck:
I have a fix that will show an error message rather than just exiting the application when this is encountered.
![image](https://user-images.githubusercontent.com/603793/52509382-cb7f5600-2bac-11e9-82d4-1d08c3017a2a.png)

This may also address some of the other recent reported issues with the application exiting shortly after launch.

scottydk:
I just downloaded and installed Bisq for the first time, and I am getting the same error. Do you know how I can fix this issue?

ghost:
If you could give a bit more informations about your configuration (what OS) and connection, it may help.

scottydk:
I installed it on my windows laptop. Not sure about connection.

devinbileck:
At this point I have been unable to determine the cause or a possible workaround. However, this is on the list of things to do for @freimair.

freimair:
seems to be closely related to #1299

devinbileck:
On my VM that consistently exhibits this issue, increasing the COOKIE_TIMEOUT resolves it. I just doubled it to 6 seconds.
[...]

PR #2399, opened 2019-02-08
Improve handling when unable to connect to tor

devinbileck:
Issue: If an IOException is raised when attempting to create
tor and the hidden service, the application will just quit without
any indication to the user. One particular scenario where this occurs
is mentioned in https://github.com/bisq-network/bisq/issues/2398.

Cause: There is an explicit statement to exit the application when an
IOException is raised.

Fix: Rather than just exit the application, show an error message
and inform the user what went wrong.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
+++ b/p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
@@ -288,13 +288,13 @@ public class TorNetworkNode extends NetworkNode {
                 log.error("Tor node creation failed: " + (e.getCause() != null ? e.getCause().toString() : e.toString()));
                 restartTor(e.getMessage());
             } catch (IOException e) {
-                log.error("Could not connect to running Tor: "
-                        + e.getMessage());
-
-                // Seems a bit harsh, but since we cannot connect to Tor, we cannot do nothing.
+                log.error("Could not connect to running Tor: " + e.getMessage());
+                // Since we cannot connect to Tor, we cannot do nothing.
                 // Furthermore, we have no hidden services started yet, so there is no graceful
                 // shutdown needed either
-                System.exit(1);
+                UserThread.execute(() -> {
+                    setupListeners.stream().forEach(s -> s.onSetupFailed(new RuntimeException(e.getMessage())));
+                });
             } catch (Throwable ignore) {
             }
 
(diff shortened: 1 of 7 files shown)


ghost:
Thanks @devinbileck for intercepting this exception early enough.
This should help with the support !

freimair (approved):
Has been on my list for quite some time now. Thanks! utAck

lichuan:
Yes, this happened in China, the GFW prevent us connecting TOR, and the Bisq application quit without any dialogs.

ManfredKarrer:
@freimair If there is a tor connection problem the tor settings windows should get displayed. There the user can choose a plugable transport or other custom settings.

devinbileck:
@ManfredKarrer I agree that it should appear in the case of a connection/communication problem. But in this case it is a setup problem (i.e. cant setup/launch tor in the first place) so there is no need since changing connection settings would have no effect.

freimair:
@ManfredKarrer yes, @devinbileck is right - this code path is mainly used when trying to use system tor. when the connection fails there is nothing we can do.
maybe we want to integrate configuration for system tor into the settings page some time...

p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
```diff
@@ -288,13 +288,13 @@ public void run() {
...
+                log.error("Could not connect to running Tor: " + e.getMessage());
+                // Since we cannot connect to Tor, we cannot do nothing.
                 // Furthermore, we have no hidden services started yet, so there is no graceful
                 // shutdown needed either
-                System.exit(1);
+                UserThread.execute(() -> {
+                    setupListeners.stream().forEach(s -> s.onSetupFailed(new RuntimeException(e.getMessage())));
+                });
             } catch (Throwable ignore) {
             }
```
devinbileck, line 299: @freimair Since I removed `restartTor(throwable.getMessage());` from onFailure below since I didn't want it to be called after handling IOException above, perhaps I should add it to this ignored exception?

devinbileck:
@freimair Integrating configuration for system tor into the settings page sounds like a good idea to me. Perhaps you can open an issue?

ripcurlx:
@devinbileck Is there anything you want to add to this PR, otherwise I would give it an utACK as well and merge it to master for the upcoming release.

ManfredKarrer:
@ripcurlx Please wait with merge for another 1-2 days to get feedback from @freimair .

p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
freimair: Maybe we can force `onSuccess` by `return null;` after triggering the `onSetupFailed` and let the `onFailure` be?

And yes, this class cries for being refactored :)

devinbileck: In fact we are already doing that. Let me review this again. I should be able to add restartTor back to onFailure.

ManfredKarrer (approved):
ACK

PR #2409, opened 2019-02-12
Update view on activation with block at current chain height

### LINKED ISSUE #2408: Attempting to make a proposal incorrectly indicated to wait until next proposal phase
I went to submit a proposal and the `Next proposal phase` was empty:
image

Looking at the voting cycle overview, I should have been able to submit a proposal as the block height was still within the proposal phase:
image

ripcurlx:
Fixes #2408.

### DIFF
--- a/core/src/main/java/bisq/core/dao/DaoFacade.java
+++ b/core/src/main/java/bisq/core/dao/DaoFacade.java
@@ -55,6 +55,7 @@ import bisq.core.dao.governance.proposal.role.RoleProposalFactory;
 import bisq.core.dao.state.DaoStateListener;
 import bisq.core.dao.state.DaoStateService;
 import bisq.core.dao.state.DaoStateStorageService;
+import bisq.core.dao.state.model.blockchain.Block;
 import bisq.core.dao.state.model.blockchain.Tx;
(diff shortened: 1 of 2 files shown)


core/src/main/java/bisq/core/dao/DaoFacade.java
```diff
@@ -484,6 +485,15 @@ public int getChainHeight() {
         return daoStateService.getChainHeight();
     }
 
+    public Block getBlockAtChainHeight() {
+        return getBlockAtHeight(getChainHeight());
+    }
+
+    public Block getBlockAtHeight(int chainHeight) {
```
ManfredKarrer: Please add @Nullable

ManfredKarrer: Or stick with Optional...

ManfredKarrer (approved):
Beside the @Nullable/Optional issue utACK

core/src/main/java/bisq/core/dao/DaoFacade.java
ripcurlx: I've already changed everything to Optional. Please see my force pushed state.

ManfredKarrer (approved):
Ah saw you pushed that already....

PR #2410, opened 2019-02-12
Add utm parameters to bisq.network links

ripcurlx:
To be able to learn if we are communicating the DAO within the client properly.
Also it adds the client language as campaign parameter to learn which translations are most important for the desktop client right now.


ManfredKarrer (approved):
utACK

PR #2412, opened 2019-02-12
Add null check

ManfredKarrer:
@ripcurlx Can you check if the syncMinAmountWithAmount= false is correct in case the minAmount is null? Not sure if it should be true or false.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/offer/MutableOfferViewModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/MutableOfferViewModel.java
@@ -732,16 +732,17 @@ public abstract class MutableOfferViewModel<M extends MutableOfferDataModel> ext
             InputValidator.ValidationResult result = isBtcInputValid(minAmount.get());
             minAmountValidationResult.set(result);
             if (result.isValid) {
-                syncMinAmountWithAmount = dataModel.getMinAmount().getValue().equals(dataModel.getAmount().getValue());
+                Coin minAmount = dataModel.getMinAmount().get();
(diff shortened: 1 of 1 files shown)


desktop/src/main/java/bisq/desktop/main/offer/MutableOfferViewModel.java
```diff
@@ -732,16 +732,17 @@ public void onFocusOutMinAmountTextField(boolean oldValue, boolean newValue) {
             InputValidator.ValidationResult result = isBtcInputValid(minAmount.get());
             minAmountValidationResult.set(result);
             if (result.isValid) {
-                syncMinAmountWithAmount = dataModel.getMinAmount().getValue().equals(dataModel.getAmount().getValue());
+                Coin minAmount = dataModel.getMinAmount().get();
```
ripcurlx: Not sure if it is a good idea to name it the same as the class field. Makes it a little bit hard to read with this.minAmount/minAmount.

ripcurlx (approved):
utACK - besides the optional changes suggested.
Having false for the case that minAmount is null is fine as there is a check for null that syncs the minAmount.
https://github.com/bisq-network/bisq/blob/2cbcf77a9e1fd9c14cfd7ec064d8d18ae66a75ba/desktop/src/main/java/bisq/desktop/main/offer/MutableOfferViewModel.java#L1037-L1042

desktop/src/main/java/bisq/desktop/main/offer/MutableOfferViewModel.java
ManfredKarrer: Yes, good point. Fixed it.

Issue #2286, opened 2019-01-18
Add dark mode


ripcurlx:
Todos:
- [ ] Clean up css colors (use derive and reduce color codes to a minimum)
- [ ] Move color codes into separate css file
- [ ] Enable switching of different color css files within preferences
- [ ] Create css file with dark color scheme

aejontargaryen:
YES !!!!

aejontargaryen:
also in other news... **Jan 2021**. Headlines read: 

### " Bisq Becomes Less Platform...More Trading Protocol "

a user by the name of 'ripcurlx' reportedly saying Bisq  

>  **" _...has now been integrated into thousands of crypto project's official wallets..._"**

with user 'cbeams' saying  

> **" it just works..."**

(a user 'manfredkarrer' was not available for comment but was last seen touring with the band U2 )

ManfredKarrer:
Haha.... but not U2 please ;-)

peterzen:
@ripcurlx  I'd like to take this on if nobody has started it yet.  Maybe it would be good to start with  refactoring the CSS into smaller chunks using SASS (the colors would reside in a separate configuration, defined as variables), then introduce classes that `derive()` from the base colors which the components can then refer to.   Creating the new color schemes would then be a case of adding new versions of the base color config.

Does that sound like a good plan?

ripcurlx:
@peterzen Please request @ManfredKarrer to add you as contributor and I'm able to assign you to this task.
Regarding SASS, I'm not sure if we can use the traditional tooling here as well. Javafx css looks very similar from the ruleset as regular css, but there are some differences. I did find a plugin that might work with our Gradle setup and JavaFX (https://github.com/robfletcher/gradle-compass), but you wouldn't need it if you just want to assign variables.
```
.root {
    -my-variable: #444;
}
``` 
You can split up the css file in smaller chunks to make it easier to work with. For that, we either have to load them manually or we use additional tooling. I leave the decision to you as long as it works with our current setup and integrates nicely within IntelliJ without doing any manual work during development.
https://github.com/bisq-network/bisq/blob/bf948101267ac459764cb1a012fb1aeefcfc7e4d/desktop/src/main/java/bisq/desktop/app/BisqApp.java#L221-L224
Still in the end I guess we'll need to load at least two files (colors + styles) or have two different compiled files (light style, dark style) that can be switched in the preferences.

Do you have experience using SASS within JavaFX?

peterzen:
[...]

PR #2414, opened 2019-02-12
Feature dark mode

peterzen:
Dark mode implementation (#2286 )

(For testing, use `Ctrl+D` to quickly cycle through the light/dark themes)


ripcurlx:
I'm not able to give proper feedback before the next release, but I quickly tried to build your PR (because I'm just to curious 😉). ATM it looks like something is not configured properly out of the box. Is there some manual setup involved for building the application with the new sass structured css?
bildschirmfoto 2019-02-13 um 10 47 19

ripcurlx:
The css resources are not build automatically.
```
Feb. 13, 2019 10:46:37 VORM. com.sun.javafx.css.StyleManager loadStylesheetUnPrivileged
WARNUNG: Resource "/bisq/desktop/theme-light.css" not found.
Feb. 13, 2019 10:46:37 VORM. com.sun.javafx.css.StyleManager loadStylesheetUnPrivileged
WARNUNG: Resource "/bisq/desktop/bisq.css" not found.
Feb. 13, 2019 10:46:37 VORM. com.sun.javafx.css.StyleManager loadStylesheetUnPrivileged
WARNUNG: Resource "/bisq/desktop/images.css" not found.
Feb. 13, 2019 10:46:37 VORM. com.sun.javafx.css.StyleManager loadStylesheetUnPrivileged
WARNUNG: Resource "/bisq/desktop/CandleStickChart.css" not found.
```

peterzen:
Looks like the compass build doesn't work correctly - I've added in the generated CSS files which is a good idea anyhow.  The SASS build should probably not be on the build critical path anyway.

peterzen:
@ripcurlx did you get a chance to fire this up?

ripcurlx:
> Looks like the compass build doesn't work correctly - I've added in the generated CSS files which is a good idea anyhow. The SASS build should probably not be on the build critical path anyway.

I'm not sure if it is a good idea to put precompiled files in the project, as this might cause confusion if someone changes the css code and later the sass files are processed and overwrite the changes. I think it should work that if someone runs `gradle build` also the required css files should be generated. The same goes with Intellij. If I change code in the SASS file and run the app I expect the changes to be reflected in the app without doing manual processing of the SASS files.

Is there a reason why it is not working like that?

ripcurlx:
> @ripcurlx did you get a chance to fire this up?

Works for me now with the pre-compiled css files.
bildschirmfoto 2019-02-14 um 16 22 45

As mentioned before I'll give a more in-depth review as soon as the new release is out. Should be mid/end of next weeks. 
I like the direction the dark mode is going, but I guess we'll need to get @pedromvpg involved for the final color codes 😄

peterzen:
Thanks for the feedback.  

Indeed, it would be good to coordinate with the color code update effort.  As part of this CSS refactoring I consolidated the colors into one place, it should be more straightforward to change them:

https://github.com/peterzen/bisq/blob/feature-dark-mode/desktop/src/main/java/bisq/desktop/styles/base-colors.scss

@pedromvpg please let me know if I can help.

In the meantime I'll fix the remaining couple issues like the tooltip background so that all of the palette can be controlled from the main configuration SASS file.

peterzen:
> > Looks like the compass build doesn't work correctly - I've added in the generated CSS files which is a good idea anyhow. The SASS build should probably not be on the build critical path anyway.
> 
> I'm not sure if it is a good idea to put precompiled files in the project, as this might cause confusion if someone changes the css code and later the sass files are processed and overwrite the changes. I think it should work that if someone runs `gradle build` also the required css files should be generated. The same goes with Intellij. If I change code in the SASS file and run the app I expect the changes to be reflected in the app without doing manual processing of the SASS files.
> 

Agreed, that would be the way most devs expect it to work.  (Though it can be argued that the SASS files are not likely to change very often, and the Compass build step does add a few seconds to the build.  Maybe a separate build step and a "this file is generated from..." warning at the top of the CSS files is a decent compromise?)

> Is there a reason why it is not working like that?

I couldn't replicate the issue in a fresh environment, am getting the `com.github.anbuck.compass` gradle plugin to load correctly.  Could you look out for this as part of your review next week?

ripcurlx:
@peterzen Just wanted to let you know, that I'm looking into your PR right now. Feedback should be available soon.

build.gradle
```diff
@@ -281,6 +287,7 @@ configure(project(':desktop')) {
         compile project(':p2p')
         compile project(':core')
         compile project(':common')
+
```
ripcurlx: No need for the extra new line.

common/src/main/proto/pb.proto
```diff
@@ -1292,6 +1292,7 @@ message PreferencesPayload {
     string rpc_user = 47;
     string rpc_pw = 48;
     string take_offer_selected_payment_account_id = 49;
+    bool use_dark_theme = 50;
```
ripcurlx: Not sure if we should use a boolean or if we want to have it more future proof if more themes are created over time?

core/src/main/java/bisq/core/locale/GlobalSettings.java
```diff
@@ -25,6 +25,7 @@
 
 public class GlobalSettings {
     private static boolean useAnimations = true;
+    private static boolean useDarkTheme = false;
```
ripcurlx: Is there a reason why we need it in the GlobalSettings object?

peterzen:
Thanks, please let me know if anything is unclear or wrong :)

desktop/src/main/resources/styles/theme-light.scss
```diff
@@ -0,0 +1,108 @@
+
+@import "_variables.scss";
```
ripcurlx: Is it necessary to import it here as well, as it is already imported in _base-colors?

ripcurlx (commented):
I left a couple of first remarks. In general I like the new structure a lot as it makes it easier to maintain, but I still have issues getting it to run on my local setup. Compass compile is running and I have the main css and themes css in my bisq/desktop/styles dir, but it seems to have issues loading them during startup:
```
Feb. 28, 2019 5:10:50 NACHM. com.sun.javafx.css.StyleManager loadStylesheetUnPrivileged
WARNUNG: Resource "/bisq/desktop/styles/theme-light.css" not found.
Feb. 28, 2019 5:10:51 NACHM. com.sun.javafx.css.StyleManager loadStylesheetUnPrivileged
WARNUNG: Resource "/bisq/desktop/styles/bisq.css" not found.
```
No idea why it uses the german locale just for this block log outpout. I'll see if I find the issue regarding the css loading. For you it works out-of-the-box?

common/src/main/proto/pb.proto
peterzen: It's indeed more future proof to use a string so that this file and the settings object won't require changing later.

core/src/main/java/bisq/core/locale/GlobalSettings.java
peterzen: No, this was an oversight on my part -- thanks for catching it.

peterzen:
@ripcurlx The changes you requested are now done, thanks for the review.

Instead of the boolean type light/dark switcher there's now a drop-down menu in the Settings that, as you point out, is more flexible when it comes to additional themes.  Please let me know what you think of this solution.

peterzen:
> I left a couple of first remarks. In general I like the new structure a lot as it makes it easier to maintain, but I still have issues getting it to run on my local setup. Compass compile is running and I have the main css and themes css in my bisq/desktop/styles dir, but it seems to have issues loading them during startup:
> 
> ```
> Feb. 28, 2019 5:10:50 NACHM. com.sun.javafx.css.StyleManager loadStylesheetUnPrivileged
> WARNUNG: Resource "/bisq/desktop/styles/theme-light.css" not found.
> Feb. 28, 2019 5:10:51 NACHM. com.sun.javafx.css.StyleManager loadStylesheetUnPrivileged
> WARNUNG: Resource "/bisq/desktop/styles/bisq.css" not found.
> ```
> No idea why it uses the german locale just for this block log outpout. I'll see if I find the issue regarding the css loading. For you it works out-of-the-box?

Those messages seem to indicate that the CSS files were not generated in the `compassCompile` step. Are the files there though, after the build?

I'm getting a clean build when running on a fresh system (Ubuntu 18.04) with `openjdk-11-jdk` installed, and the resulting binary starts up correctly with no CSS related error messages.  The theme switching works as well.

```
> Task :desktop:compassCompile
unsupported Java version "10", defaulting to 1.7
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.jruby.util.SecurityHelper (file:/home/peter/.gradle/caches/modules-2/files-2.1/org.jruby/jruby-complete/9.1.14.0/953b0ecf50980a4112fd0a2cd1dec1225b93c529/jruby-complete-9.1.14.0.jar) to field java.lang.reflect.Field.modifiers
WARNING: Please consider reporting this to the maintainers of org.jruby.util.SecurityHelper
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
    write build/resources/main/bisq/desktop/styles/bisq.css
    write build/resources/main/bisq/desktop/styles/theme-dark.css
    write build/resources/main/bisq/desktop/styles/theme-light.css
```

Are you running the build in an existing local copy?  Perhaps Maven is caching its dependencies and doesn't pick up the Compass dependency properly - unfortunately I'm not well versed enough with Gradle/Maven but it sounds like the problem might be there somewhere.  Could you perhaps try running the build on a clean working copy or maybe a fresh system inside Docker and see if it works that way?

Another idea to try would be to move away `$HOME/.gradle` to clean the gradle cache - that should force it to start with a clean sheet.

peterzen:
Rebased onto `master`

ripcurlx:
@peterzen Sorry again for the delayed response. We just have too many releases recently 😉. I'll try and response in more detail this afternoon.

ripcurlx:
The files are compiled for me as well and are in the defined build directory
```
WARNING: All illegal access operations will be denied in a future release
    write build/resources/main/bisq/desktop/styles/bisq.css
    write build/resources/main/bisq/desktop/styles/theme-dark.css
    write build/resources/main/bisq/desktop/styles/theme-light.css
```
But still when running the application they are not found and loaded. Maybe there is a classpath issue or something like that. I'll see if I'm able to figure it out, why it is not working for me.

peterzen:
What system/JDK version are you running?  I'll spin up a similar environment and see if I can replicate the problem.

ripcurlx:
> What system/JDK version are you running? I'll spin up a similar environment and see if I can replicate the problem.

I'm running macOS 10.13.6 with OpenJDK Runtime Environment 18.3 (build 10.0.2+13)

ripcurlx:
Atm moment I've an issue to startup a local seednode as some program arguments were changed recently on master. So I wasn't able to continue with testing right now.

ripcurlx:
@peterzen Just to be sure: Are you running the client from command line or are you running it from within IntelliJ?

build.gradle
```diff
@@ -314,6 +320,12 @@ configure(project(':desktop')) {
         testAnnotationProcessor "org.projectlombok:lombok:$lombokVersion"
     }
 
+    // https://github.com/anbuck/gradle-compass
+    compass {
+        sassDir = file("src/main/resources/styles")
+        cssDir = file("build/resources/main/bisq/desktop/styles")
```
ripcurlx: changing that path to `cssDir = file("src/main/java/bisq/desktop/styles")` solves my problems locally. Now I can build/run it from commandline and from within IntelliJ.

ripcurlx:
@peterzen One step further 👍 

Could you please have a look at the differences between your version
bildschirmfoto 2019-03-08 um 17 50 37
and the current one
bildschirmfoto 2019-03-08 um 17 54 46
so that they match.

I'm on vacation next week, so I can't give further feedback until then.
Thanks so far for your work!

ripcurlx:
Another thing we need to solve is, that there is no need to manually run the compile compass command after a change in the scss within IntelliJ. So developers who are working on the UI don't need to do anything manual that needs to be communicated.

peterzen:
> @peterzen Just to be sure: Are you running the client from command line or are you running it from within IntelliJ?

I'm running it from the CLI most of the time, but sometimes also from within the IntelliJ debugger.

peterzen:
> Another thing we need to solve is, that there is no need to manually run the compile compass command after a change in the scss within IntelliJ. So developers who are working on the UI don't need to do anything manual that needs to be communicated.

`./gradlew -t compassCompile` runs the compile task in continuous mode.  That, along with the CSS reload in [Scenic View](https://github.com/JonathanGiles/scenic-view) the app reloads the changes automatically.

https://github.com/peterzen/bisq/blob/aedbbe6c24cfe1b09acf6d09acfb0b6992303898/desktop/src/main/resources/styles/README.md

build.gradle
peterzen: I've tested this in Linux environments, it works correctly both with the `src/` being around and without it so the CSS gets properly compiled into the jar file.  I've added this fix to the PR.

peterzen:
> @peterzen One step further 
> 
> Could you please have a look at the differences between your version
> bildschirmfoto 2019-03-08 um 17 50 37
> and the current one
> bildschirmfoto 2019-03-08 um 17 54 46
> so that they match.
> 
> I'm on vacation next week, so I can't give further feedback until then.
> Thanks so far for your work!

This came out by accident - the white panel that's between the gradient background (which is actually invisible ATM as the white panel is opaque) is transparent.  I thought this looked rather cool so left it as transparent but it can of course be colored.  I'll fix the other items (buttons, nav etc)

I'll wait for all of your comments to come in as it's less work to do them all at once.

Are there any other GUI related items currently in the PR pipeline that are worked on currently and may get merged in the coming weeks?  If so, it would be good to coordinate them with this PR - possibly even work on them on top of this refactored CSS because in case there are larger changes in the current `bisq.css` it's quite a lot of work to manually rebase this one.

ripcurlx:
> Are there any other GUI related items currently in the PR pipeline that are worked on currently and may get merged in the coming weeks? If so, it would be good to coordinate them with this PR - possibly even work on them on top of this refactored CSS because in case there are larger changes in the current `bisq.css` it's quite a lot of work to manually rebase this one.

All redesign tasks are on hold until the DAO launch. So now would be a good to get this finished and into master before any further design work will happen. I think it also makes sense to merge this PR first without the dark mode, so there is more time to make it perfect, but still have this structural changes in asap.

Unfortunately I won't be able to look at every screen for differences to the current state. Maybe you could do this and I'll have another look at it when you're done. Would this work for you?

peterzen:
> > Are there any other GUI related items currently in the PR pipeline that are worked on currently and may get merged in the coming weeks? If so, it would be good to coordinate them with this PR - possibly even work on them on top of this refactored CSS because in case there are larger changes in the current `bisq.css` it's quite a lot of work to manually rebase this one.
> 
> All redesign tasks are on hold until the DAO launch. So now would be a good to get this finished and into master before any further design work will happen. I think it also makes sense to merge this PR first without the dark mode, so there is more time to make it perfect, but still have this structural changes in asap.
> 
> Unfortunately I won't be able to look at every screen for differences to the current state. Maybe you could do this and I'll have another look at it when you're done. Would this work for you?

I've been looking at the screens for a little too long to spot small differences so input from a fresh set of eyeballs would be great ;)  I'll have a go and go through them again, now after not having worked on them intensively for a couple weeks I'm noticing some differences which I'll fix for sure.

ripcurlx:
Yes, you get colorblind after some time 😉. Maybe you could use some image diff tool to help you out like: https://www.diffchecker.com/image-diff.

peterzen:
Rebased onto `master`

peterzen:
@ripcurlx What is the plan for integrating this PR?  There are some new conflicts now due to the DAO changes - I wonder what would be an appropriate time for doing a final rebase to add any other UI changes that might be in the pipeline.

peterzen:
^ Rebased onto `master`

wiz:
This is really cool, but I was kinda wondering what's the purpose of adding the com.github.anbuck.compass plugin? Does it use some kind of CSS inheritance to build CSS files? Since Bisq holds hot wallet private keys, it's probably best to try to minimize the amount of external code we use, even for development use only.

peterzen:
> This is really cool, but I was kinda wondering what's the purpose of adding the com.github.anbuck.compass plugin? Does it use some kind of CSS inheritance to build CSS files?

The plugin is a bridge to the SASS/Compass interpreter, it basically runs Compass on the SCSS sources.  

It was introduced in order to break up the monolithic CSS code into smaller, modular chunks and make the color scheme definitions configurable (so that more themes can be added with relative low amount of effort).

peterzen:
^Rebased on master

PR #1810, opened 2018-10-24
[WIP] Update to JDK11

devinbileck:
For fun, I decided to try JDK11 again and whaddya know, it actually worked this time :)

We can now build using either Oracle JDK or OpenJDK.

Resolves https://github.com/bisq-network/bisq/issues/1787

Issue #1787, opened 2018-10-19
Java 10 is EOL, move to Java 11


devinbileck:
As noticed by the CI builds failing lately, Oracle broke or removed their JDK10 downloads. JDK10 is End of Life anyways (https://www.oracle.com/technetwork/java/javase/eol-135779.html), so need to switch to JDK11 which will be their LTS version anyways.

ghost:
but does Bisq actually compile properly with JDK11 ?

devinbileck:
Not yet, but working on it :)

cbeams:
See https://github.com/bisq-network/bisq/pull/1792, in which I quickly (and unsuccessfully) attempted to switch Travis CI builds to OpenJDK10.

ManfredKarrer:
PR for OpenJDK 10 with javafx 11 is merged. 
I leave the issue open as we still want to migrate to java 11 after gradle support is available.

ripcurlx:
Un-stale this issue as we still want to keep this on our roadmap.

devinbileck:
Just an update on this. The primary item blocking us is there is no javapackager in JDK11. However, progress is being made on its replacement https://jdk.java.net/jpackage/ which is based on JDK 13. The next LTS release (JDK 17?) is still a few years away and no idea if this will be ported to the current LTS, JDK11. So unless we find an alternative solution, we may be stuck on JDK10 for a while.

leshik:
@devinbileck Could we use [this port](https://mail.openjdk.java.net/pipermail/openjfx-dev/2018-September/022500.html) while waiting for the full fledged solution?

devinbileck:
Thanks for the link. Potentially, yes. But I would be hesitant without further review and would especially be concerned using an unofficial/unsupported port. With the time required to ensure our code base is updated for Java 11 (and potentially BitcoinJ), and since we just recently went through all the hassle updating Java from 8 to 10, unless there is an urgent need I would think at least for now our time would be better suited focused on other high priority areas.

devinbileck:
Still relevant.

battleofwizards:
Minor update: a few days ago the first build of JEP 343 `jpackage` was published: https://jdk.java.net/jpackage/ This is early access build.

devinbileck:
Relevant

cd2357:
> The primary item blocking us is there is no javapackager in JDK11. (...) So unless we find an alternative solution, we may be stuck on JDK10 for a while.

[...]

PR #2420, opened 2019-02-13
update to OpenJDK 11

pfergi42:
the previous link to OpenJDK 10 was no longer valid


devinbileck (changes requested):
NACK
We are unable to move to JDK 11 yet. See open issue https://github.com/bisq-network/bisq/issues/1787 and further discussion here https://github.com/bisq-network/bisq/pull/1810.
OpenJDK 10 can be downloaded from https://jdk.java.net/archive/

PR #2421, opened 2019-02-14
Adapt DAO news view for Mainnet

### LINKED ISSUE #2349: Adapt teaser design for "DAO on Testnet" in the Mainnet client
In our upcoming release we are going to enable the DAO section by default to increase visibility and get more testers for Testnet. Atm the design is as follows.

@pedromvpg It would be great if you could come up with a more catchy visualization to make our users eager to try it out on Testnet ;-)

ripcurlx:
Fixes #2349.
@pedromvpg I tried to match your design as close as possible. I had to use a news tab to prevent a bigger rewrite of the DAO view navigation concept and add a external link indicator to the links which we use throughout the app. Still I changed it to a material design link icon which looks better IMHO.

bildschirmfoto 2019-02-14 um 15 02 53


ManfredKarrer (approved):
utACK

pedromvpg:
Cool

Just a couple of comments (ignore if not possible):
- Remove subnav (News btn)
- Increase padding overall, currently too close to the edge of the window
- Make the background white

PR #2424, opened 2019-02-15
Fix incorrect progress listener for spv file resync

ManfredKarrer:
- Rename model.getBtcSyncProgress() to model.getCombinedSyncProgress()
to reflect correct property
- Use bisqSetup.getBtcSyncProgress() instead of
getCombinedSyncProgress() in showFirstPopupIfResyncSPVRequested as we
need to listen to btc progress only not DAO progress.

@ripcurlx This is a bug fix for spv file resync.


devinbileck (approved):
utACK

ripcurlx (approved):
utACK

Issue #1806, opened 2018-10-23
Add support for Revolut payment links to avoid disputes


cbeams:
A common source of disputes with Revolut trades has emerged. The scenario is like this:

1. The seller publishes an email address in the Bisq trade details that is not actually registered with Revolut.

2. The buyer tries to send payment to that email address, but Revolut says it's not valid, i.e. not associated with a user account.

3. The buyer opens a dispute and explains that they can't complete payment.

4. I discuss the situation with the seller, and they explain that the buyer should send them a Revolut "payment link" via the interface at https://www.revolut.com/ie/payments.

5. The buyer does so, shares the payment link with me via arbitration chat, and I relay the payment link to the seller.

6. The seller accepts the payment via the payment link, and I then close out the issue.

The problem here is twofold:

1. the buyer has no idea whether the email address in the trade details is a registered Revolut account, or one that needs a payment link sent to it. They also may not know that payment links even exist, so when the get the 'invalid account' message from Revolut, they just go straight to a dispute.

2. Assuming they do know about payment links and create one, the seller does not get notified via email about the payment link. The buyer must manually email the payment link to the seller. This is not great for opsec on either side.

I see two options to solve this problem:

1. Require that email addresses published in trade details are actually registered with Revolut, not just arbitrary addresses that can receive a payment link.

2. Enhance the UI to allow the seller to specify how they expect the payment (direct via Revolut, or via a Revolut payment link), and add something to the UI (a popup, I guess) that explains to the buyer what they need to do in case of a payment link (give them the payment link creation URL as specified above, and explain that they must manually email the payment link to the seller's provided email address).

I would estimate I've dealt with a dozen or so disputes of this nature since we started supporting Revolut, and it requires an inordinate amount of back and forth to get these disputes resolved, so I think it's worth to do something to solve it.

My gut feeling is that we should go with something like option (2) above. While it's more complex, I don't think that simply making a rule as described in (1) will work very well. People won't read it, they'll still use unregistered email addresses, and we'll still end up with disputes. The only advantage there is that we'd be able to say it's a clear protocol violation. Right now we can't even do that.

/cc @alfsbs

sqrrm:
I agree that (2) sounds better.

With the information on how to handle this within the client the users should in general be able to handle things without bringing it to dispute, which is the aim here. I don't know enough about revolut to say, but if there is no added security risk with using payment links then that seems like a distinct advantage for bisq to be able to support them.

flix1:
[...]

PR #2426, opened 2019-02-15
Remove email for Revolut accounts

ManfredKarrer:
Fixes https://github.com/bisq-network/bisq/issues/1806

@ripcurlx I know it is after code freeze. It is not a critical PR, so if OK for you please merge. Otherwise we keep it for next release.


ripcurlx (approved):
utACK

PR #2430, opened 2019-02-17
List Veil (VEIL)

furszy:
* Official project URL: https://veil-project.com
* Official block explorer URL: https://explorer.veil-project.com

### DIFF
--- a/assets/src/main/java/bisq/asset/coins/Veil.java
+++ b/assets/src/main/java/bisq/asset/coins/Veil.java
@@ -1,0 +1,57 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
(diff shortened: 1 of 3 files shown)


assets/src/main/java/bisq/asset/coins/Veil.java
```diff
@@ -0,0 +1,63 @@
...
+        public VeilAddressValidator() {
+            super(new VeilParams());
+        }
+
+        @Override
+        public AddressValidationResult validate(String address) {
+            if (address.startsWith("V")) {
+                try {
+                    // Base58
+                    Address.fromBase58(getNetworkParameters(), address);
```
blabno: Do not expose network parameters publicly. Just call `super.validate(address)`.

blabno (changes requested):
NACK per comments

assets/src/main/java/bisq/asset/coins/Veil.java
furszy: Hi ok, no problem. 

Them are not exposed publicly, just made them protected to be able to get them in the future for bech32 and stealth addresses validation in case of not have them fully supported on the platform it selves.

furszy:
done.

blabno (approved):
ACK

Issue #2279, opened 2019-01-17
Wayland: 'GDK_IS_X11_DISPLAY (display)' failed


fredfortier:
I'm using Fedora 29 with Gnome and Wayland and a fractional screen resolution. I built the project from sources. I can run headless modules but get this error when starting the `bisq-desktop`. This seems to indicate that the app expects an X11 server but I'm not sure. If so, is there a way to build for Wayland.
```
(java:26713): Gdk-CRITICAL **: 11:35:39.602: gdk_x11_display_set_window_scale: assertion 'GDK_IS_X11_DISPLAY (display)' failed
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f9dc46ac42b, pid=26713, tid=26764
#
# JRE version: OpenJDK Runtime Environment (10.0.2+13) (build 10.0.2+13)
# Java VM: OpenJDK 64-Bit Server VM (10.0.2+13, mixed mode, tiered, compressed oops, g1 gc, linux-amd64)
# Problematic frame:
# C  [libX11.so.6+0x3142b]
#
# Core dump will be written. Default location: Core dumps may be processed with "/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h %e" (or dumping to /home/fredfortier/Downloads/bisq-0.9.3/core.26713)
#
# An error report file with more information is saved as:
# /home/fredfortier/Downloads/bisq-0.9.3/hs_err_pid26713.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
Aborted (core dumped)
```

ghost:
getting this exact same error.

freimair:
yes, I have stumbled over the issue as well. see https://github.com/bisq-network/roles/issues/78#issuecomment-452638986 for reference.

not sure what we can do about that. sounds to me as if the jdk is at fault.

3cola:
On manjaro gnome (Arch linux + gnome-wayland), I get the same error with the AUR package bisq (0.9.3-1) and the app crashes.
But with the AUR package bisq-bin (0.9.3-2), by executing the app from the terminal by typing bisq, I see that some errors are still there, but the app displays on the screen and is usable.

freimair:
[...]

PR #2431, opened 2019-02-17
Download rpm package on RedHat-based Linux

devinbileck:
Currently, the download update task will download the deb package for
any Linux distribution. Not only is this incorrect, but now that we are
able to provide an rpm package (see #2200), the download update task
needs to be able to differentiate Linux distributions and provide
the appropriate package.

The download update task will now differentiate between Debian and
RedHat based distributions (the two distributions for which we have an
install package) and download the appropriate package.

In addition, the isSupportedOS method was changed to exclusively check
for Debian and RedHat based distributions, as opposed to just Linux in
general. This means that any other distribution will encounter the
following error, which seems appropriate:

> Unable to determine the correct installer. Please download and verify
manually at https://bisq.network/downloads

### DIFF
--- a/common/src/main/java/bisq/common/util/Utilities.java
+++ b/common/src/main/java/bisq/common/util/Utilities.java
@@ -171,6 +171,14 @@ public class Utilities {
         return getOSName().contains("linux");
     }
 
+    public static boolean isDebianLinux() {
+        return isLinux() && new File("/etc/debian_version").isFile();
+    }
+
+    public static boolean isRedHatLinux() {
+        return isLinux() && new File("/etc/redhat-release").isFile();
+    }
+
     private static String getOSName() {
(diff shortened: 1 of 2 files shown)


devinbileck:
This is still a work in progress as I need to test my method of determining the appropriate distributions. If anyone is able to test on some distributions, that would be appreciated!

devinbileck:
I have tested this and it properly downloads the deb on Ubuntu 18.04. I was going to test on Fedora 29, but encountered #2279. I did verify the `/etc/redhat-release` file exists so this **should** work :)

ManfredKarrer:
Is this ready for merge despite #2279?

devinbileck:
Yes it is ready for merge.

ripcurlx (approved):
utACK

PR #2432, opened 2019-02-18
Monitor release: Settled

### LINKED ISSUE #62: Bisq Network Monitor Revisited
> _This is a Bisq Network proposal. Please familiarize yourself with the [submission and review process](https://docs.bisq.network/proposals.html)._

*Abstract: Tor and P2P network issues do and will affect the performance and acceptance of Bisq - a monitoring system greatly assists in finding their cause. Practically, the [current monitoring system](http://seedmonitor.0-2-1.net/) still leaves us with a lot of tedious guesswork. I propose a fresh monitoring solution which is properly designed for the task at hand (unlike fast quick and in a hurry as the current one had to be). The solution features monitoring node(s) in the P2P network gathering metrics while an external service takes care of history and presentation. The presentation can be suitable for developers and users alike, provide detailled insight into Bisq's network layer, lets us grasp the value of Bisq to the world, and prepare Bisq for the future.*

# Introduction

Bisq is getting bigger and bigger and thus, issues in Bisq's network layer appear more frequently. In late 2017, for example, the network simply did not perform (https://github.com/bisq-network/bisq/issues/1172). In an attempt to understand why, @ManfredKarrer created a [monitoring do-that](http://seedmonitor.0-2-1.net/) fast quick and in a hurry. However, while clearly showing the situation, the new monitor did not help very much in understanding the cause, let alone foster strategies to prevent such a situation in the future. Finally, the network magically recovered. Since then, Bisq has grown even bigger.

# Challenge

An ideal monitoring solution has to serve multiple purposes. First of all, it should support developers in finding fixes to pending issues (https://github.com/bisq-network/bisq/issues/1241, https://github.com/bisq-network/bisq/issues/1299, ). Second, it should support developers in analyzing and understanding the network. A better understanding of the network lets us anticipate upcoming issues and maybe stop them in their tracks before they become an actual issue. Third, numerical performance values let us evaluate the effectiveness of network tweaks more clearly and make informed decisions whether to keep the tweaks or not. Furthermore, numerical performance values can be fed to some sort of attack detection mechanisms which trigger countermeasures on demand. Leaving the realm of development, a historical display of numerical performance values allows users to get an idea why their offer is taking so long to be published and maybe pick a time where the network is less busy (https://github.com/bisq-network/bisq/issues/1575). And last but not least, the collective of metrics lets people get an idea of the value Bisq brings to the world.

The [current monitoring solution](http://seedmonitor.0-2-1.net/), unfortunately, is very limited (because it has been created fast quick and in a hurry in order to get hold of actual pending issues). First of all, there is no historical data: a dev cannot correlate historical data with other sources ([Tor metrics](https://metrics.torproject.org/) for example) in order to either conclude whether or not it is Bisq's fault when the network suffers from performance loss. Second, the current monitoring solution does not provde sole Tor performance values nor does it provide network load values. The only value available is a roundtrip time metric which might indicate performance loss, but does not say if it is caused by poor Tor performance, by a high network load or by congestion caused by a way-to-high network load. Whereas the latter should have been visible before congestion actually kicked in. Third and last, the data presentation is not suitable for people other than developers. The statistics site is static and has to be manually refreshed to get up-to-date data, there is no historical data, and the metrics displayed are too cryptic to be understood by the average (Bisq-affine) Joe. All in all, the current monitoring solution leaves us guessing what Bisq's network layer looks like inside, if Tor is blocking our request due to their DoS protection or if our optimiziations do really optimize things. Hence, the current monitoring solution does not come near the ideal solution sketched above.

# Proposal

It is time to create a proper monitoring solution (https://github.com/bisq-network/bisq/issues/1361). From a technical point of view, the shiny new monitoring solution of course has to conform to the usual [bullsh*t bingo](https://www.wikihow.com/Play-Bullshit-Bingo): it has to be clean, modular, extensible, easy to deploy, low maintenance effort, use existing solutions where possible, fast initial time to market, etc.

Having that in mind, I propose
1) starting fresh and design the monitoring solution from the ground up. A clean start does not trick us into reusing approaches just because they are already there, without thinking about their usefullness and effectiveness. Futhermore, by discarding old code we do not pull deprecated and/or dead code into a shiny new project, especially since the existing monitoring solution has been created in a hurry.
2) making the new monitoring solution easy to deploy and operate. I had my share of application servers, tomcats and IT departments - thus, I suggest instead of wasting time fighting these we invest a little more development time in order to create a simple executable which everyone, who is able to run the Bisq client, is able to run.
3) making the monitor highly modular and configurable so an operator can easily pick a (sub)set of metrics he wants to run. Furthermore, a simple Java-properties-based configuration should be used to control how these metrics behave.
4) to focus on extensibility. We might have a good idea of which metrics we need right now. However, there certainly are metrics we do not think of right now. Such future metrics have to be addable without rewriting the whole monitor thing. Furthermore, a new-to-Bisq developer should find her way around the monitoring tool quite easily. I am not saying that the actual metric themselves have to be super clean, just that a developer can add a set of new metrics without having to spend a lot of time solving riddles.

Following these proposals, we IMHO should be able to create a monitoring solution which properly allow for a deeper look ainsidet Bisq's network layer while not being outdated by tomorrow.

Please find a big picture of the proposed monitoring solution in the illustration below. There are two main components to the monitoring solution. First, *Monitoring Node*s are inserted into Bisq's P2P network. These monitor-bigpicture nodes only gather data, they do not keep a historical record of the data. Second, *Monitoring Service*s ~scrape~ accept data from the *Monitoring Node*(s), keep a historical record and visualize the data. ~Offline discussions yielded [Prometheus](https://prometheus.io/) to be the monitoring service to be used.~ Please note that the underlying Tor network is part of every connection in the illustration above.

EDIT: Prometheus turned out to be not suitable for our purpose, at least if we do not want to overload the network. Prometheus is meant to monitor system resources multiple times a minute, if we do that with the Bisq network or even only Tor, we might run into network overload and DoS countermeasures, respectively. However, there is another great open source tool out there: [Graphite](https://graphiteapp.org/). It does the same thing as Prometheus, except it does NOT scrape the data but waits for data to arrive (and features timestamps). However, as the chances of running into the limits of such a tool are quite high, I rearranged the proposal to allow each *Monitor Node* to publish its findings to its own set of *Monitoring Service*s. For example, for system information, use Prometheus as a *Monitoring Service*, for benchmarking the Tor and p2p networks use Graphite, and for event tracking use -InsertYourFavoriteMonitoringService-.

## Monitor Node(s)

Please find an architectual overview of a *Monitor Node* in the illustration below. A *Monitor Node* uses a *Scheduler* as its central component. the *Scheduler* executes *Metric*s and supplies them with their share of *Configuration*. The minimal *Configuration* for each *Metric* contains whether the *Metric* is enabled and if
monitor-architecture
yes, at which intervals the *Metric* is to be run. The collected data is published to a suitable *Monitoring Service* ~offered as a [Prometheus job exporter](https://prometheus.io/docs/introduction/overview/) (maybe we need to add a Pushgateway)~ via a Tor ~Hidden Service~. The *Monitor Node* is to be run as a simple executable from the command line (and thus can be easily turned into a system service). Furthermore, on Linux systems, the *Monitor Node* can be instructed to reload its configuration and react to changes (enable/disable metrics, change intervals) by a `kill -USR1` signal without the need for restarting the executable (since we expect to run multiple Tor binaries and restarting the whole thing would take an awful lot of time while at the same time loosing running average data).

## Monitoring Service

The *Monitoring Service* collects the data provided by the *Monitor Node*(s), keeps a historical record and by some means also provides a GUI. For starters, this service is to be provided by the open-source monitoring solution [Graphite](https://graphiteapp.org/) ~[Prometheus](https://prometheus.io/)~. It is well-maintained and active, takes a few minutes to set up and handles recording and displaying data quite nicely.

# Implementation Details

Please note the priority list and/or time line below on how I propose to get the proposed monitoring system up and running. Each release is meant to be set productive. The *Babysteps* release is meant to complement the existing [monitoring do-that](http://seedmonitor.0-2-1.net/) as it primarily adds Tor metrics. As some of the work is already done, I believe that we can provide *Babysteps* in January already. The *Showing Off* release is then ment to supersede Manfreds monitor as by then, the new monitor includes all metrics provided by Manfreds monitor. The *Settled* release then focuses on making the value of Bisq to the world somehow visible.

- [x] create basic infrastructure for a monitoring node
- [x] create Graphite instance
- [x] benchmark Tor startup
- [x] benchmark Hidden Service startup
- [x] benchmark Tor roundtrip time to torproject.org's hidden service
- [x] **release: Babysteps**
- [x] p2p RTT (using ping/pong)
- [x] p2p network load (messages per timeslot)
- [x] p2p network load histogram (messagetype per timeslot)
- [x] estimate the number of open offers (snapshot)
- [x] **release: Showing off**
~- [ ] *Monitoring Service* whitelisting~
- [x] exchange rates (integrate work of @HarryMacfinned)
- [x] fee estimation (integrate work of @HarryMacfinned)
- [x] p2p network load (include *Refresh*/*Remove* messages)
- [x] p2p network load histogram (include *Refresh*/*Remove* messages)
- [x] message number diffs between Seed Nodes per message type
- [x] open offers per market (teaser for what is possible)
- [x] notifications on alarms per email/slack/...
- [x] fix >6h empty data
- [x] create Readme
  - [x] running a *Monitor Node*
  - [x] running a *Monitor Service*
- [x] **release: Settled**
- [ ] **release: yourReleaseNameHere**

Future:
- [ ] BSQ price monitor
- [ ] benchmark Hidden Service startup on system Tor
- [ ] estimate the p2p network size (peer count)
- [ ] estimate the number of successful trades per timeslot
- [ ] maybe replace https://github.com/bisq-network/bisq-markets?
- [ ] realize parts of https://github.com/bisq-network/analytics/issues/3?
- [ ] market metrics
- [ ] [style](https://www.neteye-blog.com/2017/12/how-to-customize-your-grafana-theme/) Grafana
- [ ] ?

# Aftermath

The proposed monitoring solution should pave the way of Bisq's future quite a bit. With the solution, we are able to understand Bisqs network layer better and thus, take better care of it. 

Please feel free to suggest further Metrics and how they fit in the list above. Please also feel free to raise any questions and concerns! It is a rather big project and more minds usually perform better in creating a more complete picture.

freimair:
closes https://github.com/bisq-network/proposals/issues/62

### DIFF
--- a/monitor/README.md
+++ b/monitor/README.md
@@ -122,11 +122,66 @@ more information can be found [here](https://graphite.readthedocs.io/en/latest/i
 
 ### Configuration
 
-There is no further configuration necessary. However, you might change your iptables/firewalls to not let anyone access your Graphite instance from the outside.
+For configuration, you must adapt the whisper database schema to suit your needs. First, stop your docker container by running
+
+```
+docker stop graphite
+```
+
+Find your config files within the `Source` directory stated in
+
+```
+docker inspect graphite | grep -C 2 graphite/conf\",
+```
+
+Edit `storage-schemas.conf` so that the frequency of your incoming data (configured in the monitor configs `interval`) is matched. For example, insert 
+```
+[bisq]
+pattern = ^bisq.*
+retentions = 10s:1h,5m:31d,30m:2y,1h:5y
+```
+before the `[default...` blocks of the file. This basically says, that every incoming set of data reflects 5 minutes of the time series. Furthermore, every 30 minutes, the data is compressed and thus, takes less memory as it is kept for 2 years.
+
+Further, edit `storage-aggregation.conf` to configure how your data is compressed. For example, insert
+```
+[bisq]
+pattern=^bisq.*
+xFilesFactor = 0
+aggregationMethod = average
+```
+before the `[default...` blocks of the file. With this configuration, whenever data is aggregated, the `average` data is made available given that at least `0%` of the data points (i.e. floor(30 / 5 * 40%) = 2 data points) exist. Otherwise, the aggregated data is dropped. Since we start the first hour with a frequency of 10s but only supply data every 4 to 6 minutes, our aggregated values would get dropped.
+
+*Please note, that I have not been able to get the whole thing to work without the 10s:1h part yet*
+
+Finally, update the database. For doing that, go to the storage directory of graphite, the `Source` directory stated in
+```
+docker inspect graphite | grep -C 2 graphite/conf\",
+```
+Once there, you have two options:
+- delete the whisper directory
+```
+rm -r whisper
+```
+- update the database by doing
+```
+find ./ -type f -name '*.wsp' -exec whisper-resize.py --nobackup {} 10s:1h 5m:31d 30m:2y 1h:5y \;
+```
+
+and finally, restart your graphite container:
+```
+docker start graphite
+```
+
+
+Other than that, there is no further configuration necessary. However, you might change your iptables/firewalls to not let anyone access your Graphite instance from the outside.
 
 ### Backup your data
 
-*TBD*
+The metric data is kept in the `Source` directory stated in
+```
+docker inspect graphite | grep -C 2 graphite/conf\",
+```
+ready to be backed up regularly.
 
 ## Setting up Grafana
 
--- a/monitor/README.md
+++ b/monitor/README.md
@@ -2,12 +2,17 @@
 
 The Bisq monitor node collects a set of metrics which are of interest to developers and users alike. These metrics are then made available through reporters.
 
-The *Babysteps* release features these metrics:
+The *Settled* release features these metrics:
 - Tor Startup Time: The time it takes to start Tor starting at a clean system, unpacking the shipped Tor binaries, firing up Tor until Tor is connected to the Tor network and ready to use.
 - Tor Roundtrip Time: Given a bootstrapped Tor, the roundtrip time of connecting to a hidden service is measured.
 - Tor Hidden Service Startup Time: Given a bootstrapped Tor, the time it takes to create and announce a freshly created hidden service.
+- P2P Round Trip Time: A metric hitchhiking the Ping/Pong messages of the Keep-Alive-Mechanism to determine the Round Trip Time when issuing a Ping to a seed node.
+- P2P Seed Node Message Snapshot: Get absolute number and constellation of messages a fresh Bisq client will get on startup. Also reports diffs between seed nodes on a per-message-type basis.
+- P2P Network Load: listens to the P2P network and its broadcast messages. Reports every X seconds.
+- P2P Market Statistics: a demonstration metric which extracts market information from broadcast messages. This demo implementation reports the number of open offers per market .
 
-The *Babysteps* release features these reporters:
+
+The *Settled* release features these reporters:
 - A reporter that simply writes the findings to `System.err`
 - A reporter that reports the findings to a Graphite/Carbon instance using the [plaintext protocol](https://graphite.readthedocs.io/en/latest/feeding-carbon.html#the-plaintext-protocol)
 
@@ -22,6 +27,11 @@ The *Bisq Network Monitor Node* is to be configured via a Java properties file. 
 A sample configuration file looks like follows:
 
 ```
+## System configuration
+
+# true overwrites the reporters picked by the developers (for debugging for example) (defaults to false)
+System.useConsoleReporter=true
+
 ## Each Metric is configured via a set of properties.
 ##
 ## The minimal set of properties required to run a Metric is:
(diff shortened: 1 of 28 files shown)


ManfredKarrer:
Will review after relase (code freeze atm)

ManfredKarrer (changes requested):
> if(Capabilities.getSupportedCapabilities().isEmpty())
                Arrays.stream(Capabilities.Capability.values()).forEach(capability -> Capabilities.addCapability(capability.ordinal()));

We should not add all Capabilities automatically. For instance as you dont run it as full DAO node you will not handle BSQ Block requests.

> P2PSeedNodeSnapshot.run.hosts=3f3cu2yw7u457ztq.onion:8000, 723ljisnynbtdohi.onion:8000, fl3mmribyxgrv63c.onion:8000
PriceNodeStats.run.hosts=http://5bmpx76qllutpcyp.onion, http://xc3nh4juf2hshy7e.onion, http://44mgyoe2b6oqiytt.onion, http://62nvujg5iou3vu3i.onion, http://ceaanhbvluug4we6.onion

Maybe its better to use the nodes from the code, so if it gets updated there is not need to update config files? Or maybe to use default code based nodes and if config is set it will override the default nodes?

ManfredKarrer (approved):
utACK

PR #2435, opened 2019-02-18
Improve payout layout for arbitration

ripcurlx:
Instead of
bildschirmfoto 2019-02-18 um 18 33 22
buyer and seller amount have a floating top label if set
bildschirmfoto 2019-02-18 um 18 48 56


ManfredKarrer (approved):
utACK

Issue #2429, opened 2019-02-16
Implement URL redirects for GitHub downloads, compensation requests, and proposals


devinbileck:
This topic was mentioned in https://github.com/bisq-network/proposals/issues/72#issuecomment-463001006

Implement URL redirects for downloads, compensation requests, and proposals; all of which currently go to GitHub.
e.g.
https://bisq.network/download/... --> https://github.com/bisq-network/bisq/releases/download/...

Basically all hardcoded URLs should have a redirect under our control.

Having a redirect for downloads will allow us to feel confident hard-coding it indefinitely without worry if we switch version control providers or host downloads elsewhere.

Additionally, if we dont have redirects for compensation requests and proposals, GitHub will be forever recorded in compensation requests and could potentially become dead links.

ptmhd:
I can help with that.  
no, seriously. I can think, search/type and commin

ripcurlx:
@devinbileck Do you want to lead this issue? If yes, I'm happy to assign it to you.

devinbileck:
@ripcurlx yes, feel free to assign it to me. I will look into it.

devinbileck:
I have implemented the redirects. However, am considering what approach (if any) to take with regards to entering proposals within the UI to ensure people are using the correct URL. Should they only enter the proposal id, select from a dropdown, ...

devinbileck:
The URL redirects are implemented and I have entered #2473 to ensure proposals are entered using correct URL.

PR #2444, opened 2019-02-20
Use URL redirects to GitHub content

devinbileck:
URL redirects for downloads, source code, roles; all of which currently
go to GitHub.

See bisq-network/bisq#2429


devinbileck:
To be merged after https://github.com/bisq-network/bisq-website/pull/122

ManfredKarrer (approved):
utACK

PR #2445, opened 2019-02-21
Update bonded role unlock time

devinbileck:
Use the defined unlock time on mainnet, an arbitrarily low value on
regtest for dev testing, and a relatively short time on testnet for
testing purposes.


ManfredKarrer (approved):
utACK

PR #2447, opened 2019-02-21
Dao serverside regtest

ManfredKarrer:
As testnet is un-usable ;-( I have setup a regtest network for DAO testing.

1 master node which Bitcoind only. A script creates all 9-11 minutes a new block
4 seed nodes connecting only to that the node.
1 explorer node (https://explorer.bisq.network/testnet/) connecting only to the master node.

I added a new basecurrency enum DAO_TESTNET which is technically a regtest but uses the IP of the master node for the bitcoin connection.

It is important that nobody creates blocks and connects to the master node to avoid disruption. By controlling block creation with a script on only one node we can easier test reorg scenarios.
We are aware that this regtest set up is less realistic as mainnet, but the alternative to create a real testnet with mining nodes would have been much more effort and involve compiling of bitcoind.

I am still testing all so please don't merge yet but welcome to review and help testing.
The user can select the DAO TESTNET in settings same as with TESTNET.
I will send by email to the main testers a data directory with full wallets (BTC and BSQ) ;-).

Testnet will not work anymore as no seed nodes are running anymore.


ManfredKarrer:
Tested more now. So I think it is ready for merge after review.

devinbileck (approved):
ACK
Code looks good and I have tested it.

HarryMacfinned:
@devinbileck 
I guess building from source is as usual,
or are there any changes ?

ripcurlx:
ACK - even it is already merged 😉

oscarguindzberg:
utACK

PR #2450, opened 2019-02-21
Display BSQ marketcap using preferred currency

devinbileck:
In the BSQ dashboard view, the market capitalisation is shown in USD.
Instead, use the user's preferred currency.


ripcurlx:
ACK

PR #2453, opened 2019-02-21
Fix issues with rpm build

ripcurlx:


ManfredKarrer (approved):
utACK

desktop/package/linux/package.sh
```diff
@@ -117,7 +119,7 @@ fi
 echo Generating rpm package
 $JAVA_HOME/bin/javapackager \
     -deploy \
-    -BappVersion=$version_base \
```
devinbileck, line 120: @ripcurlx I seem to recall that I used `$version_base` which stripped the "-SNAPSHOT" portion because it causes trouble when generating the package. Maybe I forgot to commit where it sets `$version_base`...

ripcurlx: Feel free to add this anytime again 😄 I just changed what I needed to be able to create the builds.

PR #2454, opened 2019-02-21
List XDR (XDR)

vlad-khramov:
* Official project URL: http://mile.global/
* Official block explorer URL: https://explorer.mile.global/

XDR - the second asset of Mile project (Mile already added)


blabno (approved):
ACK

PR #2455, opened 2019-02-21
Dao deactivate localhost for dao testnet

ManfredKarrer:


ripcurlx (approved):
ACK - works for me now!

PR #2459, opened 2019-02-23
List Qwertycoin (QWC)

ghost:
Official project URL: https://qwertycoin.org
Official block explorer URL: https://explorer.qwertycoin.org

Included is a pop up message to let QWC senders know what is required of them in the event a trade goes to arbitration.

### DIFF
--- a/assets/src/main/java/bisq/asset/coins/Qwertycoin.java
+++ b/assets/src/main/java/bisq/asset/coins/Qwertycoin.java
@@ -1,0 +1,30 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.asset.coins;
+
+import bisq.asset.AltCoinAccountDisclaimer;
+import bisq.asset.Coin;
+import bisq.asset.CryptonoteAddressValidator;
+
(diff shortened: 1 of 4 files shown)


qwertycoin-org:
## Arbitration Information

This is a transaction over 5 QWC to this recipient:  
>QWC1K6XEhCC1WsZzT9RRVpc1MLXXdHVKt2BUGSrsmkkXAvqh52sVnNc1pYmoF2TEXsAvZnyPaZu8MW3S8EWHNfAh7X2xa63P7Y

**The Transaction Hash is:**  
> [9580508F8C9FFA7786CCD5BC33DDD34041F65F475F53155024B11B411AB8680A](https://explorer.qwertycoin.org/?hash=9580508F8C9FFA7786CCD5BC33DDD34041F65F475F53155024B11B411AB8680A#blockchain_transaction)

In Case of arbitration you can check this Transaction [on our Blockchain explorer](https://explorer.qwertycoin.org/?#check_payment). You need the additional Transaction Key, which can be seen in the senders Wallet (or by using getTxKey get_tx_key).

**This Transaction ID is:**  
> E58EB6F89C3EFCABFDE350E88657F6CB00E5474BE1F039EACF48A95184CC9D0E

Proof the Payment [here](https://explorer.qwertycoin.org/?#check_payment) and use the above Transaction Hash, the Transaction ID and the Recipient Address. And you will see a message like this:

> Specified address received **5.00000000 QWC** in transaction **9580508f8c9ffa7786ccd5bc33ddd34041f65f475f53155024b11b411ab8680a**

This test can be used to check if the receiver has received the coins from the sender.

blabno (commented):
Code looks good, no sign of ico. @ManfredKarrer do you think arbitration can be done with this asset?

qwertycoin-org:
> 
> 
> Code looks good, no sign of ico. @ManfredKarrer do you think arbitration can be done with this asset?

If that is not enough, we will revise the function. But I think for a Privacy Coin is this the best way. This can be proven that a recipient has received a sum from the sender. Only sender and receiver know about this transaction key.

ManfredKarrer:
Seems similar like Monero tx proof, so yes I think that is ok.

blabno (approved):
ACK

ripcurlx:
Damn - I thought the build for this PR already went through 😭. It currently breaks now master as the implementation is wrong.

qwertycoin-org:
@ripcurlx is fixed here: https://github.com/bisq-network/bisq/pull/2562 in a fresh PR.
There was a renamed Validator Class i didn't see that. I'm sorry for that.

Issue #2452, opened 2019-02-21
Two external screens on Ubuntu 18.xx result in blank window - JavaFX IllegalArgumentException: Window must not be zero


tomekit:
Tried to run Bisq on my Ubuntu 18.10, but have failed miserably.

Downloaded `deb` package from here: https://bisq.network/downloads/
Installed on my machine, then ran using the Search tool. Blank window has appeared, so I've ran the Bisq from command line, the blank window appears and here is the output:
```
/opt/Bisq 
15:42 $ ./Bisq
Feb-21 15:43:47.371 [main] INFO  bisq.core.setup.CoreSetup: 


Log files under: /home/user/.local/share/Bisq/bisq 
Feb-21 15:43:47.383 [main] INFO  bisq.common.util.Utilities: System info: os.name=Linux; os.version=4.18.0-15-generic; os.arch=amd64; sun.arch.data.model=64; JRE=10.0.2+13 (Oracle Corporation); JVM=10.0.2+13 (Java HotSpot(TM) 64-Bit Server VM) 
Feb-21 15:43:47.539 [main] INFO  bisq.common.app.Version: Version{VERSION=0.9.3, P2P_NETWORK_VERSION=1, LOCAL_DB_VERSION=1, TRADE_PROTOCOL_VERSION=1, BASE_CURRENCY_NETWORK=0, getP2PNetworkId()=10} 
Feb-21 15:43:47.540 [main] INFO  bisq.core.setup.CoreSetup: Path to Bisq jar file: /opt/Bisq/app/desktop-0.9.3-all.jar 
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.inject.internal.cglib.core.$ReflectUtils$1 (file:/opt/Bisq/app/desktop-0.9.3-all.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of com.google.inject.internal.cglib.core.$ReflectUtils$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
[...]

PR #2460, opened 2019-02-24
Handle IllegalArgumentException in multi-screen environment

devinbileck:
When the application window is being created, it checks what the
maximum bounds for a window is in order to set the window size.
However, multi-screen environments may encounter an
IllegalArgumentException (Window must not be zero).

Just ignore the exception and continue, which means the window will
use the minimum window size since we are unable to determine if we can
use a larger size.

Fixes https://github.com/bisq-network/bisq/issues/2452


ManfredKarrer (approved):
utACK

PR #2461, opened 2019-02-24
Security framework for the API

blabno:


api/src/main/java/bisq/api/http/service/auth/TokenRegistry.java
```diff
@@ -0,0 +1,63 @@
...
+public class TokenRegistry {
+
+    public static final long TTL = 30 * 60 * 1000;
+
+    private final TimeProvider timeProvider;
+    private final RandomStringGenerator randomStringGenerator;
+    private Map<String, Long> tokens = new HashMap<>();
+
+    public TokenRegistry() {
+        this(() -> UUID.randomUUID().toString(), System::currentTimeMillis);
```
ManfredKarrer, line 16: I think you have to use SecureRandom for cryptographically secure randomness.

api/src/main/java/bisq/api/http/service/auth/TokenRegistry.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    public TokenRegistry(RandomStringGenerator randomStringGenerator, TimeProvider timeProvider) {
+        this.timeProvider = timeProvider;
```
ManfredKarrer, line 20: I think you have to use SecureRandom for cryptographically secure randomness.

api/src/test/java/bisq/api/http/service/auth/ApiPasswordManagerTest.java
```diff
@@ -0,0 +1,391 @@
...
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+
+
+import com.github.javafaker.Faker;
```
ManfredKarrer, line 28: I dont find that class in the source code at https://github.com/Javafaker. 
Do we really need that? It is one of those examples where I don't want to get new dependencies (unknown developer, small functionality)

api/src/test/java/bisq/api/http/service/auth/ApiPasswordManagerTest.java
```diff
@@ -0,0 +1,391 @@
...
+
+    private String dataDir;
+    private BisqEnvironment bisqEnvironmentMock;
+    private TokenRegistry tokenRegistryMock;
+    private ApiPasswordManager apiPasswordManager;
+
+    private static String getRandomPasswordDifferentThan(String otherPassword) {
+        String newPassword;
+        do {
+            newPassword = Faker.instance().internet().password();
```
ManfredKarrer, line 43: What is that supposed to do? internet() is a weird method name...
As said above I prefer to not add that dependency.

api/src/test/java/bisq/api/http/service/auth/ApiPasswordManagerTest.java
ManfredKarrer: Ah sorry, just saw its a test class. So then I dont care so much about that dependency anymore, but still would like to know better what the class provides.

api/src/test/java/bisq/api/http/service/auth/ApiPasswordManagerTest.java
ManfredKarrer: Ah sorry, just saw its a test class. So then I dont care so much about that dependency anymore, but still would like to know better what the class provides.

ManfredKarrer (changes requested):

api/src/test/java/bisq/api/http/service/auth/ApiPasswordManagerTest.java
blabno: Faker is random data generator. On the top level it has methods with categories.

api/src/main/java/bisq/api/http/service/auth/TokenRegistry.java
blabno: `UUID.randomUUID` uses `SecureRandom` by default.

api/src/main/java/bisq/api/http/service/auth/TokenRegistry.java
blabno: Time provider is meant to provide actual time. It is used as abstraction for unit tests.

blabno:
I've substituted `Random` with `SecureRandom`. I'm using the one created with default constructor.
I've tried `SecureRandom.getInstanceStrong()` but it takes about a minute to generate single long. So it's absolutely unacceptable performance.

api/src/main/java/bisq/api/http/service/auth/TokenRegistry.java
ManfredKarrer: Ah ok.

ManfredKarrer (approved):
utACK

ManfredKarrer:
I will merge master after the PR. Please update your branch afterwards to be in sync.

PR #2462, opened 2019-02-24
Fix incorrect available balance for voting

ManfredKarrer:
- Subtract vote fee from available wallet balance
- Add log for errors
- Update view in error case


devinbileck (approved):
utACK

Issue #1996, opened 2018-11-29
0.9.0: Windows system tray icon is low resolution


devinbileck:
The new Bisq icon shown in the Windows system tray is low resolution and looks bad.
![image](https://user-images.githubusercontent.com/603793/49204273-2cdea980-f360-11e8-8dd8-763cebc5292d.png)

ripcurlx:
I'll add a high-res icon version and do the same retina check as we do for macOS. Hopefully that solves this problem.

devinbileck:
@ripcurlx Now I am encountering the following
```
Nov-30 00:21:17.300 [JavaFX Application Thread] ERROR bisq.desktop.app.BisqApp: Error during app init java.lang.IllegalArgumentException: input == null!
	at java.desktop/javax.imageio.ImageIO.read(ImageIO.java:1400)
	at bisq.desktop.app.SystemTray.init(SystemTray.java:107)
	at bisq.desktop.app.SystemTray.<init>(SystemTray.java:74)
	at bisq.desktop.app.SystemTray.create(SystemTray.java:68)
	at bisq.desktop.app.BisqApp.setupStage(BisqApp.java:231)
	at bisq.desktop.app.BisqApp.startApplication(BisqApp.java:137)
	at bisq.desktop.app.BisqAppMain.startApplication(BisqAppMain.java:126)
	at bisq.core.app.BisqExecutable.onApplicationLaunched(BisqExecutable.java:211)
	at bisq.desktop.app.BisqAppMain.onApplicationLaunched(BisqAppMain.java:94)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$10(PlatformImpl.java:428)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$11(PlatformImpl.java:427)
	at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:96)
	at com.sun.glass.ui.win.WinApplication._runLoop(Native Method)
	at com.sun.glass.ui.win.WinApplication.lambda$runLoop$3(WinApplication.java:174)
	at java.base/java.lang.Thread.run(Thread.java:844)
```
`task_bar_icon_windows@2x.png` needs to be renamed to `system_tray_icon_windows@2x.png`

[...]

PR #2467, opened 2019-02-26
Fix Windows system tray icon resolution

devinbileck:
For high resolution, using auto sizing provides better results.

Fixes https://github.com/bisq-network/bisq/issues/1996


ManfredKarrer (approved):
utACK

PR #2469, opened 2019-02-26
Refactor capabilities and other P2P-related stuff

freimair:
Got rid of the Capability ordinals being used all over the place - a first and major step towards a cleaner, more readable Bisq code base where Bugs have less places to hide. (Done in my intention to clean up the p2p part)

Doing that I even managed to remove some 60 loc: some stats using `github.com/AlDanial/cloc v 1.80`

EDIT: after a quick chat with @ManfredKarrer I did some more refactoring...

_|files|blank|comment|code
:-------|-------:|-------:|-------:|-------:
 same|0|0|1087|3842
 modified|32|0|7|208
 added|1|18|37|114
 removed|0|40|5|304

# Note
this is a fairly big change where functionality should remain the same. However, this needs thorough testing.
My suggestion is that every reviewer of this PR should have a testing session with me where we do the basic things such as a simple trade so major bugs are discovered before merging.
In-depth testing is done anyhow before the next release...

### DIFF
--- a/common/src/main/java/bisq/common/app/Capabilities.java
+++ b/common/src/main/java/bisq/common/app/Capabilities.java
@@ -17,55 +17,87 @@
 
 package bisq.common.app;
 
-import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 
-import lombok.Getter;
-import lombok.Setter;
-
+/**
+ * hold a set of capabilities and offers appropriate comparison methods.
+ *
+ * @author Florian Reimair
+ */
 public class Capabilities {
-    // We can define here special features the client is supporting.
-    // Useful for updates to new versions where a new data type would break backwards compatibility or to
-    // limit a node to certain behaviour and roles like the seed nodes.
-    // We don't use the Enum in any serialized data, as changes in the enum would break backwards compatibility. We use the ordinal integer instead.
-    // Sequence in the enum must not be changed (append only).
-    public enum Capability {
-        TRADE_STATISTICS,
-        TRADE_STATISTICS_2,
-        ACCOUNT_AGE_WITNESS,
-        SEED_NODE,
-        DAO_FULL_NODE,
-        PROPOSAL,
-        BLIND_VOTE,
-        ACK_MSG,
-        BSQ_BLOCK
+
+    /**
+     * The global set of capabilities, i.e. the capabilities if the local app.
+     */
+    public static final Capabilities app = new Capabilities();
+
+    protected final Set<Capability> capabilities = new HashSet<>();
+
+    public Capabilities(Capability... capabilities) {
+        this(Arrays.asList(capabilities));
     }
 
-    // Application need to set supported capabilities at startup
-    @Getter
-    @Setter
-    private static List<Integer> supportedCapabilities = new ArrayList<>();
-
-    public static void addCapability(int capability) {
-        supportedCapabilities.add(capability);
+    public Capabilities(Capabilities capabilities) {
+        this(capabilities.capabilities);
     }
 
-    public static boolean isCapabilitySupported(final List<Integer> requiredItems, final List<Integer> supportedItems) {
-        if (requiredItems != null && !requiredItems.isEmpty()) {
-            if (supportedItems != null && !supportedItems.isEmpty()) {
-                List<Integer> matches = new ArrayList<>();
-                for (int requiredItem : requiredItems) {
-                    matches.addAll(supportedItems.stream()
-                            .filter(supportedItem -> requiredItem == supportedItem)
-                            .collect(Collectors.toList()));
-                }
-                return matches.size() == requiredItems.size();
-            } else {
-                return false;
-            }
-        } else {
-            return true;
-        }
+    public Capabilities(Collection<Capability> capabilities) {
+        this.capabilities.addAll(capabilities);
+    }
+
+    public void resetCapabilities(Capability... capabilities) {
+        resetCapabilities(Arrays.asList(capabilities));
+    }
+
+    public void resetCapabilities(Capabilities capabilities) {
+        resetCapabilities(capabilities.capabilities);
+    }
+
+    public void resetCapabilities(Collection<Capability> capabilities) {
+        this.capabilities.clear();
+        this.capabilities.addAll(capabilities);
+    }
+
+    public boolean isCapabilitySupported(final Set<Capability> requiredItems) {
+        return capabilities.containsAll(requiredItems);
+    }
+
+    public boolean isCapabilitySupported(final Capabilities capabilities) {
+        return isCapabilitySupported(capabilities.capabilities);
+    }
+
+    public boolean hasCapabilities() {
+        return !capabilities.isEmpty();
+    }
+
+
+    /**
+     * helper for protobuffer stuff
+     *
+     * @param capabilities
+     * @return int list of Capability ordinals
+     */
+    public static List<Integer> toIntList(Capabilities capabilities) {
+        return capabilities.capabilities.stream().map(capability -> capability.ordinal()).sorted().collect(Collectors.toList());
+    }
+
+    /**
+     * helper for protobuffer stuff
+     *
+     * @param capabilities a list of Capability ordinals
+     * @return a {@link Capabilities} object
+     */
+    public static Capabilities fromIntList(List<Integer> capabilities) {
+        return new Capabilities(capabilities.stream().map(integer -> Capability.values()[integer]).collect(Collectors.toSet()));
+    }
+
+    @Override
+    public String toString() {
+        return Arrays.toString(Capabilities.toIntList(this).toArray());
     }
 }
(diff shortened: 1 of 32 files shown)


p2p/src/main/java/bisq/network/p2p/peers/peerexchange/Peer.java
```diff
@@ -39,47 +38,43 @@
 @Getter
 @EqualsAndHashCode(exclude = {"date"}) // failedConnectionAttempts is transient and therefore excluded anyway
 @Slf4j
-public final class Peer implements NetworkPayload, PersistablePayload, SupportedCapabilitiesListener {
+public final class Peer extends Capabilities implements NetworkPayload, PersistablePayload, SupportedCapabilitiesListener {
```
ManfredKarrer, line 41: Missing call to super in @EqualsAndHashCode

ManfredKarrer: Extension feels not really right to me here. A peer "has capabilities" and not "is capabilities".

ManfredKarrer: Same with SharedModel in Connection class..

seednode/src/main/java/bisq/seednode/SeedNodeMain.java
```diff
@@ -60,7 +63,10 @@ protected void doExecute(OptionSet options) {
 
     @Override
     protected void addCapabilities() {
-        Capabilities.addCapability(Capabilities.Capability.SEED_NODE.ordinal());
+        // TODO got even worse after refactoring
+        List<Integer> current = Capabilities.toIntList(Capabilities.app);
+        current.add(Capability.SEED_NODE.ordinal());
+        Capabilities.app.resetCapabilities(Capabilities.fromIntList(current));
```
ManfredKarrer, line 69: Why not add a method for appending a capability? Check to not add duplicates can be done with capability object itself rather then ordinal, no (missing here but was at another class where it was added after checking if not already there)?

p2p/src/main/java/bisq/network/p2p/network/Connection.java
```diff
@@ -280,31 +280,16 @@ public void sendMessage(NetworkEnvelope networkEnvelope) {
...
             final ProtectedStoragePayload protectedStoragePayload = (((AddDataMessage) msg).getProtectedStorageEntry()).getProtectedStoragePayload();
             return protectedStoragePayload instanceof CapabilityRequiringPayload && sharedModel.isCapabilitySupported(((CapabilityRequiringPayload) protectedStoragePayload).getRequiredCapabilities());
         } else if (msg instanceof AddPersistableNetworkPayloadMessage) {
             final PersistableNetworkPayload persistableNetworkPayload = ((AddPersistableNetworkPayloadMessage) msg).getPersistableNetworkPayload();
             return persistableNetworkPayload instanceof CapabilityRequiringPayload && sharedModel.isCapabilitySupported(((CapabilityRequiringPayload) persistableNetworkPayload).getRequiredCapabilities());
+        } else if(msg instanceof CapabilityRequiringPayload) {
+            return sharedModel.isCapabilitySupported(((CapabilityRequiringPayload) msg).getRequiredCapabilities());
         } else {
-            return msg instanceof CapabilityRequiringPayload && sharedModel.isCapabilitySupported(((CapabilityRequiringPayload) msg).getRequiredCapabilities());
+            return true;
```
ManfredKarrer: The noCapabilityRequiredOrCapabilityIsSupported is very hard to understand (was worse before). Maybe we should use that refactoring moment for improving the method so it becomes more easier.
I would need more time to review the corrtectness (compared to the old version), but would prefer to invest that time in a improved method version.

common/src/main/java/bisq/common/app/Capabilities.java
```diff
@@ -17,55 +17,87 @@
...
-        DAO_FULL_NODE,
-        PROPOSAL,
-        BLIND_VOTE,
-        ACK_MSG,
-        BSQ_BLOCK
+
+    /**
+     * The global set of capabilities, i.e. the capabilities if the local app.
+     */
+    public static final Capabilities app = new Capabilities();
```
ManfredKarrer, line 37: I would prefer a more descriptive name here. When I read Capabilities.app it was not very clear what it means.
Also does not need to be static?
Should the class extend HashSet instead of using the field capabilities - as that is basically what it holds?

common/src/main/java/bisq/common/app/Capabilities.java
```diff
@@ -17,55 +17,87 @@
...
+    public void resetCapabilities(Capability... capabilities) {
+        resetCapabilities(Arrays.asList(capabilities));
     }
 
-    // Application need to set supported capabilities at startup
-    @Getter
-    @Setter
-    private static List<Integer> supportedCapabilities = new ArrayList<>();
+    public void resetCapabilities(Capabilities capabilities) {
+        resetCapabilities(capabilities.capabilities);
```
ManfredKarrer, line 58: capabilities.capabilities reads strange. If we would use a HashSet as super class i think it becomes more natural.

common/src/main/java/bisq/common/app/Capabilities.java
```diff
@@ -17,55 +17,87 @@
...
+    }
+
+    public void resetCapabilities(Collection<Capability> capabilities) {
```
ManfredKarrer, line 61: Would replaceCapabilities be better than resetCapabilities? With reset I would expect more the clear funtion rather then the additional add.

common/src/main/java/bisq/common/app/Capabilities.java
ManfredKarrer: Maybe we should make it a Guice instance scoped as singleton? 
And then use it in peer and shared model as parameter passed to the constructor (I know there is missing guice and passing is painful)... So maybe keeping the Capabilities.app static is less painful?   
But then maybe its better to move it to another class which reflects that it holds the global capability state of the app?

ManfredKarrer (approved):
ACK

In Peer the EqualsAndHashCode should include a super call. I will push that afterwards. 
IntelliJ reports such issues with static code analysis (colored bg), otherwise I would not have spotted it as well ;-)
`@EqualsAndHashCode(exclude = {"date"}, callSuper = true)`

I tested just lightly. As that comes with many "risky" changes we have to observe well if there are any issues reported which might be connected to the topics of that PR.

p2p/src/main/java/bisq/network/p2p/peers/peerexchange/Peer.java
freimair: aggreed. but `peer.isCapabilitySupported(...)` and `connection.isCapabilitySupported(...)` feels quite right. Having them implement some sort of `HasCapabilities` interface would result in `peer.getCapabilities().isCapabilitySupported(...)` which imho is confusing.

seednode/src/main/java/bisq/seednode/SeedNodeMain.java
freimair: first, there is a TODO and second, we need that kind of code exactly once. and once more for the monitor.

common/src/main/java/bisq/common/app/Capabilities.java
freimair: yes I thought about that but having `Set` as superclass adds all the methods as well and well, makes the whole `Capabilities` class superfluous which then results in handling capabilities like a set. I do not know if we want that.

common/src/main/java/bisq/common/app/Capabilities.java
freimair: > I would prefer a more descriptive name here. When I read Capabilities.app it was not very clear what it means.

make a suggestion

> Also does not need to be static?

yes, as it holds the capabilities of the running Bisq itself and is used in multiple places all over the code.

> Should the class extend HashSet instead of using the field capabilities - as that is basically what it holds?

see below

common/src/main/java/bisq/common/app/Capabilities.java
freimair: hm interesting. A `setCapabilities` seems to me clear as it replaces all capabilities, a `resetCaps` indicates more strongly, that the capabilities are going to be replaced. I would understand a  `clearCapabilities` as a clear without add.

Again, in a collection of event logs, a reset clears the log.

Hm. `setCapabilities` or simply `set`?

p2p/src/main/java/bisq/network/p2p/network/Connection.java
freimair: Agreed. I thought about changing stuff around but it quickly gets dirty.

Most beautiful would be that we do not need the method at all but instead can use a `connection.isCapabilitySupported(msg)` or `isCapabilitySupported(msg)` directly. However, the class hierarchy has grown overly complicated over time and such a solution is not possible as things stand now (would result in a diamond pattern if Java would allow for that).

Even looks worse if depicted like that:
messages

I even tried to use java8/9 defaults in interfaces and create some sort of mixins. Nasty stuff. Definitely needs cleanup! Small steps...

PR #2470, opened 2019-02-26
API over TOR

blabno:


ManfredKarrer (commented):
Have you got feedback from @freimair about that approach?

ManfredKarrer:
@blabno @freimair Please have a discussion what is the best approach to support a hiddenservice for the API. Goal is to not add any security, performance or stability risks for the P2P network. Also we don't want to get code dependencies from API code to normal app as we will likely deliver 2 binaries: one with the API and one without API (and without any code dependencies).

freimair (approved):
just had a call with @blabno. Everything seems fine. No obvious risks get introduced. The P2P network remains untouched.

utAck

ManfredKarrer (approved):
utACK

ManfredKarrer:
Had a discussion with @freimair about that.
Summary:
This approach tells Tor to start a new HS and proxy the traffic to the API port, so not even a socket gets created from the Java side. Seems to have no risk and very low performance costs.

PR #2475, opened 2019-02-27
Fixed memory leak in message throttle mechanism

### LINKED ISSUE #599: Bounty for resolving a memory leak issue
It seems that there is a memory leak since the trade statistics release in the GUI app.
I could never see a leak before in the Bitsquare GUI app (on OSX), but the seed node app (running under Linux) has a leak since a while. It might be related to open files (network) which is treated differently on OSX and Unix.
I want to give out a bounty for anyone who is able to locate and fix the leak (get in touch to negotiate amount).
To test if the fix was sufficient, I will test both the seed node app and the GUI app to leave it running for min. 1 week to see if memory consumption does not grow.

freimair:
mostly showed during seed node operations (as the seed nodes tend to run for longer than a client), however, every Bisq instance has been affected.

Made a test with MSG_THROTTLE_PER_10_SEC = 50:

_ | diff byte[] in 15min |
|:------|------:|
before | +700.000 |
after: | +100 |

fixes #599 - finally (and hopefully)

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/network/Connection.java
+++ b/p2p/src/main/java/bisq/network/p2p/network/Connection.java
@@ -378,9 +378,10 @@ public class Connection implements MessageListener {
                             .collect(Collectors.toList()).toString());
                 }
             }
-            // we limit to max 1000 (MSG_THROTTLE_PER_10SEC) entries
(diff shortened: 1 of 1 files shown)


ManfredKarrer (approved):
utACK

PR #2476, opened 2019-02-28
Seed node repo refactoring

freimair:
shame on me...


ManfredKarrer:
@ripcurlx Lets delay that after release so we keep test efforts low.

ManfredKarrer (changes requested):
If I use a localhost seed node it tries to connect to one of the onion addresses.

`We send a PreliminaryGetDataRequest to peer rxdkppp3vicnbgqt.onion:8002. `

Seems the handling for localhost is missing something.

ManfredKarrer:
Are the duplicated resources in the test/resource directory needed? Might cause consfusion...

ManfredKarrer:
With a dev DAO_TESTNET seed it worked. Have not tested yet with the live seeds....

freimair:
> If I use a localhost seed node it tries to connect to one of the onion addresses.

how do you specify that you want to use a localhost seed node?
- `--useLocalhostForP2P`?
if so, that no longer works. We talked about that and not having peers like `localhost:4001` in the code base. If you want to use localhost seednodes you have to add them manually like `--seedNodes=<host:port[,...]>`

I would like to get rid of these dev-related command line parameters as far as possible before a shiny and user-friendly Bisq 1.0 is released...

> Are the duplicated resources in the test/resource directory needed?

for testing. tests do not have access to the main resources.

core/src/main/resources/btc_regtest.seednodes
```diff
@@ -1,10 +1,10 @@
 # For development you need to change that to your local onion addresses
-# 1. Run a seed node with prog args: --bitcoinNetwork=regtest --nodePort=8002 --myAddress=rxdkppp3vicnbgqt:8002 --appName=bisq_seed_node_rxdkppp3vicnbgqt.onion_8002
+# 1. Run a seed node with prog args: --bitcoinNetwork=regtest --nodePort=8002 --appName=bisq_seed_node_localhost_8002
```
ManfredKarrer: should be 2002 -> I edited it already...

ManfredKarrer (approved):
NACK Cannot get localhost dev setup running.

ManfredKarrer:
Dman, my #d9f9baf commit done on GH has committed to Bisq master instead to your PR. So I will merge now and try to find the issue why localhost is not working....

freimair:
ahm ok, but are you sure? I see your commit in my branch...

ManfredKarrer:
Don't know exactly what went wrong....

PR #2482, opened 2019-03-01
Allow spending of unconfirmed BSQ change outputs

ManfredKarrer:
When creating a BSQ transaction (actually at commit time as we can create a tx and then
cancel it in the confirmation popup) we store the change output (only that not the other
possible BSQ output) in a persisted list. The BsqCoinSelector will take that list to
allow spending those coins. We use the txType to find the index of the cahnge output.
We only have one change output in the transactions created in Bisq. Multiple change
outputs would be valid but our goal is only increased usability in the Bisq app and it is
not related to validation rules.

We update out list at each new block confirmation.

With that approach we avoid too much dependencies to the BitcoinJ side.

- Add UnconfirmedBsqChangeOutputListService and persisted UnconfirmedBsqChangeOutputList
for storing unconfirmed outputs
- Add lookup for unconfirmed BSQ change outputs at BsqCoinSelector and allow spending if
found
- Pass TxType for walletsManager.publishAndCommitBsqTx calls
- Add TxType to bsqWalletService.commitTx
- Refactor getPreparedSendTx methods for BSQ and BTC sending to one common method with a
coinselector parameter.
- Add getChangeAddress method to BsqWalletService to make change outputs more explicit
- Add unconfirmedChangeBalance to onUpdateBalances handlers
- Rename availableBalance to availableConfirmedBalance in onUpdateBalances
- Unify onUpdateBalances parameter names


core/src/main/java/bisq/core/btc/wallet/BsqCoinSelector.java
```diff
@@ -32,18 +34,34 @@
...
+            return true;
+
+        // It might be that it is an unconfirmed change output which we allow to be used for spending without requiring a confirmation.
+        // We check if we have the output in the dao state, if so we have a confirmed but unspendable output (e.g. confiscated).
+        if (daoStateService.getTxOutput(new TxOutputKey(parentTransaction.getHashAsString(), output.getIndex())).isPresent())
+            return false;
+
+        // Only if its not existing yet in the dao state (unconfirmed) we use our unconfirmedBsqChangeOutputList to
+        // check if it is an own change output.
+        return unconfirmedBsqChangeOutputListService.hasTransactionOutput(output);
```
sqrrm, line 66: Don't we need to check that it's also unspent or is the unconfirmed list only for unspent txs?

core/src/main/java/bisq/core/dao/state/unconfirmed/UnconfirmedBsqChangeOutputListService.java
```diff
@@ -0,0 +1,194 @@
...
+    // API
+    ///////////////////////////////////////////////////////////////////////////////////////////
+
+    /**
+     * Once a tx gets committed to out BSQ wallet we store the change output for allowing it to be spent in follow-up
+     * transactions.
+     */
+    public void onCommitTx(Transaction tx, TxType txType, Wallet wallet) {
+        // We remove all potential connected outputs from our inputs as they would have been spent.
+        removeConnectedOutputsOfInputsOfTx(tx);
```
sqrrm, line 77: Ok, I see the outputs are removed when spent, previous comment is then void since all transactions in the unconfirmed list are unspent.

core/src/main/java/bisq/core/dao/state/unconfirmed/UnconfirmedBsqChangeOutputListService.java
```diff
@@ -0,0 +1,194 @@
...
+
+        TransactionOutput change = outputs.get(changeOutputIndex);
+        if (!change.isMine(wallet))
+            return;
+
+        UnconfirmedTxOutput txOutput = UnconfirmedTxOutput.fromTransactionOutput(change);
+        if (unconfirmedBsqChangeOutputList.containsTxOutput(txOutput))
+            return;
+
+        log.error("We add txOutput {}", txOutput);
```
sqrrm: Why log an error?

core/src/main/java/bisq/core/dao/state/unconfirmed/UnconfirmedBsqChangeOutputListService.java
```diff
@@ -0,0 +1,194 @@
...
+        unconfirmedBsqChangeOutputList.add(txOutput);
+        persist();
+    }
+
+    //TODO not sure how we should handle it but seems safest to just clear all
```
sqrrm, line 148: Agree, rather be safe and delay trading for one new confirmed block.

sqrrm (commented):
I think this is a better way to handle it than through the parser. I tried using the parser and that's just too risky and hard to control in the end, this way is much more controlled and only very specifically designated change outputs can be spent without confirmation and parsing which is safer.

The problem would be that this is more fragile if we start changing anything in the BSQ parsing as this is now a separate kind of parsing that needs to be handled with great care.

core/src/main/java/bisq/core/btc/wallet/BsqCoinSelector.java
ManfredKarrer: As we remove entries for all connected outputs of any inputs we should not get that problem (e.g. user spend utxo from previously unconfirmed tx). If we would miss one it would throw an error on the BitcoinJ level if we try to use an input which is already spent. But I will have a look to add an additional check.

core/src/main/java/bisq/core/dao/state/unconfirmed/UnconfirmedBsqChangeOutputListService.java
ManfredKarrer: Yes, if data hanlding stays in sync it should be ok. Still as we manage parallel data now there might be theoretically an issue (e.g. if persistence fails after remove). I will try to test that scenario to verify that we get an exception from BitcoinJ.

core/src/main/java/bisq/core/dao/state/unconfirmed/UnconfirmedBsqChangeOutputListService.java
ManfredKarrer: Was dev debugging... was remove in follow up commit

ManfredKarrer:
@sqrrm I fixed and improved balance display as well as hanlding of confiscated bonds. Can you review the open commits?

ripcurlx (changes requested):
NACK - Besides the validation problems mentioned below for going over your available balance it looks fine.
bildschirmfoto 2019-03-04 um 11 13 10
@ManfredKarrer Is this error message correct? I don't think that is my available balance to send.
bildschirmfoto 2019-03-04 um 11 15 15
When I try to send an amount that I shouldn't be able to send I get this wrong mining fee warning popup.

ManfredKarrer:
@ripcurlx Thanks for the finding. Fixed with latest commit. Could you ACK now?

PR #2483, opened 2019-03-01
Add MacOS support in install_java.sh script

ben-kaufman:
Modified the `scripts/install_java.sh` script to support MacOS.
Added a step of setting the JDK in IntelliJ IDEA to the `import-idea` doc.


devinbileck (approved):
utACK

PR #2487, opened 2019-03-02
Update bond lockup prompt

devinbileck:
- Changed display string from "lockup time" to "unlock time" otherwise it may be confused with how long it takes to lock up the bond.
- Included estimated time duration (e.g. days) for the unlock time.

For reference, this is what the lockup prompt looked like before:
image

This is what it looks like now:
image

And the unlock prompt now:
image


ManfredKarrer (approved):
utACK

PR #2488, opened 2019-03-02
Delay broadcasting of taker fee tx

ManfredKarrer:
To avoid the issue with lost taker fees if the take-offer attempt fails,
we delay the publishing of the taker fee tx just before the deposit tx
gets published.
The taker fee tx must not be committed to the wallet to avoid that the
wallet would require a resync in case the tx is not published.
If the tx is not committed the addresses used there are not considered
as used and that would cause issues with the address entry context
assignment for the deposit tx. To avoid those issues we need to force
the creation of new addresses used for the multisig and payout address
entries. To be sure that happens before any of the following tasks make
use of those address entries we do that already in the createTakerFee
task.
For BSQ fee tx it behaves similar but needs separate testing.


ManfredKarrer:
@ripcurlx This would require good testing. We can discuss if it should go into the upcoming release. As it would fix an important issue with lsot taker fees I think it would be worth to try to get it in.

ripcurlx (approved):
ACK - Tested it by canceling send publish deposit tx. No fees where lost and if I run the same trade without canceling everything works as expected.

ripcurlx:
@ManfredKarrer I rebased your PR and solved the conflict. Besides the two missing merge commits to master, also the OfferPayloud TTL change commit is gone as it was changed by you to 9 minutes already in another PR.

ripcurlx:
I leave it to you to merge, if you want to have another look at my reworked commit history.

PR #2490, opened 2019-03-03
Add altcoin payment method for live trading

ManfredKarrer:
- Add LiveAsset account, payment method, AccountPayload
- Extract super classes for normal CryptoCurrenyAccount and payload and
LiveAssetAccount and payload
- Add isAsset method
- Add button for creating a live asset account

As it is a bit tricky to use the AccountForm for both methods and add a
checkbox there so defined which payment method to use I added a button
to add an AccountForm with the LiveAssets passed. This is just
temporary to be able to test a bit more and see if there are any
critical issues. We should unify that form but that might require a bit
of refactoring of the CryptoCurrencyForm.


core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -2363,6 +2363,11 @@ payment.country=Country
 payment.extras=Extra requirements
 payment.email.mobile=Email or mobile no.
 payment.altcoin.address=Altcoin address
+payment.altcoin.tradeInstantCheckbox=Trade instant
```
ripcurlx: Maybe it should be a little more descriptive like `Trade instant (within 1 hour) with this Altcoin` ⚡️?

ripcurlx (approved):
ACK - Besides my suggested text change.

core/src/main/resources/i18n/displayStrings.properties
ManfredKarrer: Ok, changed it.

PR #2492, opened 2019-03-03
Fix system tray visibility for dark mode Mac

ben-kaufman:
The system tray icon was not visible on MacOS when using dark mode. I added a white version of the graphics and added a check to see if the Mac is on dark mode or not. Working well on my machine.


ManfredKarrer (commented):
utACK

PR #2498, opened 2019-03-05
Use percentage based value for security deposits

ManfredKarrer:
Use percentage based value of trade amount for buyer and seller
security deposit.
To ensure that the BTC value is not getting too low, we apply a min.
value for both.


ripcurlx:
@ManfredKarrer I've added an informational icon tooltip and a new badge, as it is a big change that might lead to security deposit mistakes.

ripcurlx (approved):
ACK

ripcurlx:
@ManfredKarrer I've added the news badge for the instant trade to this PR as I'm reusing the new NewBadge component.

Issue #1943, opened 2018-11-19
Can launch Bisq only after system restart on Windows 


freimair:
This issue might address #1023, #1161, possibly #791, #1299 and more recent reports even from @ManfredKarrer himself. The issue is meant as a progress tracker.

### Issue
Bisq gets started. Bisq runs. Bisq gets closed and exists gracefully (?). Nothing wrong yet.
Now, Bisq gets started again, and it refuses to start with:
```
ERROR i.b.n.p.n.TorNetworkNode: Tor node creation failed: java.lang.RuntimeException: Could not delete file C:\Users\Name\AppData\Roaming\Bisq\btc_mainnet\tor\lock 
```

### Fixes
restart your machine

### Affected platforms
Windows

### Findings so far
The Tor process creates a file named `lock` in the Tor working directory. It has size 0 (as expected for a lock file), gets created as soon as the tor-binary is executed, and is in turn owned by the Tor process. However, at least on my Arch linux machine, the file does not get deleted after Tor shutdown and is only overwritten on a fresh start of Tor.

Ah yes, and this behaviour does apply to bisq, the netlayer tests and the TorBrowser-bundle as well. It seems like the issue might be located in Tor itself.

EDIT: I just found the piece of code in netlayer, which is trying to delete the file and fails doing it. It is preceeded by
```
        // It can take a little bit for the Tor OP to detect the connection is dead and kill itself
        Thread.sleep(1000)
```
Timing issue?

### Help needed
I cannot reproduce the issue right now. However, if anyone stumbles upon the issue, please check if a Tor process is still running on your machine. If so, try force-killing the process and start Bisq again.

ManfredKarrer:
I saw it when I installed Bisq on windows VM with AV software installed. The AV took ages at installation time to verify Bisq and either I killed it then (dont remember exactly) or the AV terminated Bisq. Then after trying to start it did not work anymore bc of the lock file. After a OS restart all was fine again. I think on Windows the new process did not had the permission to delete the old lock created by an prev. process.

freimair:
hm. Netlayer tries to delete the file itself when the first startup attempt is not successfull. So I do not believe that the new process has no permission to delete an old lock - in the end, it is always Windows and its permission system.

[...]

PR #2499, opened 2019-03-05
WIP Updated netlayer, jtorctl and tor-binary

freimair:
# Changes
- the latest tor binary
- a fresh jtorctl release (which should at least give some more information in #2398
- a fresh netlayer release (which might bring us forward in #1943)

# Before merging
We need a tester for each platform. I did my linux part and it works nicely.

Missing tests: 
- [x] Windows
- [x] OSX

The testing effort needed is simply building bisq and starting it up. If that works (or at least no other than it did before), the branch can be merged.


ManfredKarrer:
@freimair If a PR is not ready for merge/review please mark them with prefix WIP and use the draft feature (new in GH - when you create the PR you can select it as draft)

freimair:
@ManfredKarrer the PR is ready for merging and review. That is exaclty the point. I cannot test on OSX neither can I test on Windows. The reviewing process is meant to make it unlikely that a merge does break master. So please review and make it unlikely that the merge is going to break master by testing on OSX for example.
Since we still do not have integration testing on different platforms, we have to do it that way...

devinbileck:
I gave this a test on my primary Windows machine and it ran without issue. However, on my Windows VM I still encounter the auth cookie issue. But this time the behavior is slightly different. It is raising an IOException wrapped in a TorCtlException. So instead of performing onSetupFailed, it is performing restartTor.
image

freimair:
ah yes, and although the [Tor spec](https://github.com/torproject/torspec/blob/master/control-spec.txt#L1700) says that with this new version the default HS addresses are of the new format, a short chat with a tor dev resulted in https://trac.torproject.org/projects/tor/ticket/29669. so no worries. we are still with the old address format

ManfredKarrer:
@freimair Did a quick test and all looks ok on OSX

ManfredKarrer (approved):
ACK

PR #2500, opened 2019-03-05
Remove Log.traceCall methods

ManfredKarrer:
Florian reported that they consume quite a bit of performance.As
they have not been used for development testing since long we should
better remove those.


ripcurlx (approved):
utACK

PR #2501, opened 2019-03-06
Seednode optimizations

freimair:


p2p/src/main/java/bisq/network/p2p/P2PService.java
```diff
@@ -311,6 +311,7 @@ private void onNetworkReady() {
                 "seedNodeOfPreliminaryDataRequest must be present");
 
         requestDataManager.requestUpdateData();
+        UserThread.runPeriodically(() -> requestDataManager.requestUpdateData(), 1, TimeUnit.HOURS);
```
ManfredKarrer: Maybe we should make that configurable?

ManfredKarrer (approved):
<strike>ACK</strike>  EDIT: Did not see it was a draft...

ManfredKarrer:
Ah sorry just saw it is in draft.... So the ACk is for current state ;-)

p2p/src/main/java/bisq/network/p2p/P2PService.java
freimair: agreed. However, would you (for starters) really offer the user the chance to change that?

ManfredKarrer: not really for users, more for seed node operators. but yes not really needed, can be added later as well once needed...

freimair:
had the jhgcy2won7xnslrb.onion seed node running with this change since 2019-03-28 09:54 UTC. No issues showed up. yet.

The issue has been that, at times, data reached the seed nodes later than an active client. So whenever the client requested data from the seed node, it happened that the seed node delivered data which the client already removed due to a remove-message. As of now, we keep a (limited) history of messages (i.e. their hashes) we already removed and check if we already removed a message we get.

ManfredKarrer (commented):
I think we cannot do that. If the user removes and offer and then enables it again it would not be added until the 500 items get cleared out. 
Have you found out why the seed node gets delayed some messages? Might be overall too heavy load of seed or that some ddos protection causes that?

freimair:
Well, I thought that removing and offer and republishing an offer results in a fresh hash...

ad delayed messages: not only seed nodes are affected. the issue arises whenever one node (which gets asked for UpdateData) has not yet received a remove command but the the asking node has. A few ms are enough to cause the issue. Has nothing to do with heavy load or any DoS protection - just plain P2P network latency.

If there is no fresh hash for a fresh message, then we have these options:
- revert to making seed nodes crash and restart again
- reduce the 500 to say 5 minutes?
- only apply the filter to `UpdateDataResponse`s

ManfredKarrer (commented):
Requires another review. Previous ACK was a mistake.

freimair:
I just saw that this PR is still open so I

- made it work with the current master
- removed the bits with the repeated UpdateDataRequests, as this seems not to work so well

ripcurlx (approved):
utACK

seednode/src/main/java/bisq/seednode/SeedNodeMain.java
```diff
@@ -53,7 +53,6 @@ public static void main(String[] args) throws Exception {
     protected void doExecute(OptionSet options) {
         super.doExecute(options);
 
-        checkMemory(bisqEnvironment, this);
```
mrosseel, line 56: maybe I missed this in a discussion somewhere, but wat is the motivation to remove this check?

ManfredKarrer: Please discuss with @freimair If it is removed it might require that nodes run as systemd service to guarantee a restart in case of a out of memory. I would suggest that just a few nodes use that first to not risk that all fail if there is an issue.

freimair: The thing is, the checkMemory routine attempts to restart the seed node and does so by doing highly specific stuff. I had my seed node run with systemd and everytime the checkMemory did its thing, I had to manually restore and restart the node. Even without systemd, the restart did not work.
Restarting a system service after it fails due to external stuff (memory in this case) is not to be implemented in the service itself. That only complicates things and hardly works on a wider range of servers.

mrosseel: thx for the reply, I don't disagree with the change but seednode operators will have to be notified and as Manfred says maybe a staggered deploy would be best.

Issue #2477, opened 2019-02-28
Add export to csv feature for dao vote result screen


ManfredKarrer:
It would be nice to have a export to csv feature of all the voting cycles with their results. It might be a bit tricky hwo to get it into a flat format as it is a tree structure (cycle has proposals, proposal has received votes,...).

ManfredKarrer:
I would offer a 1000 BSQ bounty for any developer who can implemnet that.
A secondary goal is to have a determinisic data which we could easily compare between users to see if the vote result is 100% the same. So the file should not contain data specific to my own vote (as in the UI).

ben-kaufman:
@ManfredKarrer I'll try to implement that this week.

ben-kaufman:
I'm still thinking about how should this feature be implemented.
The problem is, as @ManfredKarrer noted, that the data is in a tree structure. This makes it difficult to output to a CSV file. I currently have a few ideas for possible implementations for this feature.

1. Export a separate CSV file for each data structure. This means that we will have a CSV file of all the cycles, a different CSV file of all the proposals etc. This is probably the simplest solution as most of the code for needed for that already exists. The problem is that the data is separated into multiple files.

2. Export all data into a single CSV. There are multiple ways we could export all the data into a CSV (e.g. Like with solution 1 but in a single file, Have all cycles proposals under cycle row and all votes under each row of a proposal). This is very messy in my opinion but can be done.

3. Export all data into as a JSON file, as suggested by @blabno on the Slack chat. JSON would be much easier here as it supports complex data structure as we have here. I think this would be the best solution but I'm not sure if it would be fine not to use the CSV format as specified in the issue.

@ManfredKarrer do you (or anyone else) have any thoughts on how this feature should be implemented?

ManfredKarrer:
@ben-kaufman 
I agree that the JSON option sounds like the best one. People can choose then toold for converting it further if needed. A main goal is to get a easy to share data structure for checking data consistency between peers. I will work on other more low level tools for that as well (hash chains) but to have a simple text format based output would be good to have.

PR #2504, opened 2019-03-06
Added button to export voting history into JSON

ben-kaufman:
Fix: #2477 

Added a button to export the voing history under DAO -> Governance -> Vote Results:
screen shot 2019-03-06 at 18 13 36

Attached here is the JSON file which was generated for me so it's possible to compare when you generate locally (zipped because GitHub doesn't support `.json`). 

Note: I'm not confident with the way I added the export button to the UI.

[voteResultsHistory.json.zip](https://github.com/bisq-network/bisq/files/2937159/voteResultsHistory.json.zip)


desktop/src/main/java/bisq/desktop/main/dao/governance/result/VoteResultView.fxml
```diff
@@ -29,4 +31,7 @@
     <columnConstraints>
         <ColumnConstraints percentWidth="100"/>
     </columnConstraints>
+    <HBox GridPane.rowIndex="8" GridPane.columnSpan="3" alignment="BOTTOM_RIGHT">
```
ManfredKarrer: We do not use FXML anymore beside of the containers (one day we want to get rid completely of it) as it has bad performance and is too limited for more dynamic content. There are still relicts but new code we want to have in Java only. See FormBuilder as out Util for UI components.

desktop/src/main/java/bisq/desktop/main/dao/governance/result/VoteResultView.java
```diff
@@ -105,6 +112,7 @@
     private final Preferences preferences;
     private final BsqFormatter bsqFormatter;
     private final Navigation navigation;
+    public AutoTooltipButton exportButton;
```
ManfredKarrer: Would be private once removed from FXML

desktop/src/main/java/bisq/desktop/main/dao/governance/result/VoteResultView.java
```diff
@@ -157,9 +165,11 @@ public VoteResultView(DaoFacade daoFacade,
     @Override
     public void initialize() {
         gridRow = phasesView.addGroup(root, gridRow);
+
         selectedVoteResultListItemListener = (observable, oldValue, newValue) -> onResultsListItemSelected(newValue);
 
         createCyclesTable();
+        exportButton.updateText(Res.get("shared.exportJSON"));
```
ManfredKarrer: Here would be the code or creating the Button. 
E.g. `button = FormBuilder.addButtonAfterGroup( gridPane,  rowIndex,  title);`

desktop/src/main/java/bisq/desktop/main/dao/governance/result/VoteResultView.java
```diff
@@ -172,6 +182,57 @@ protected void activate() {
         cyclesTableView.getSelectionModel().selectedItemProperty().addListener(selectedVoteResultListItemListener);
 
         fillCycleList();
+        exportButton.setOnAction(event -> {
```
ManfredKarrer, line 198: I would prefer to delegate that to a method to keep the activate() method short and easy to understand.
E.g. `exportButton.setOnAction(event -> doExportToJson()`

desktop/src/main/java/bisq/desktop/util/GUIUtil.java
```diff
@@ -268,6 +275,21 @@ public static void importAccounts(User user, String fileName, Preferences prefer
         }
     }
 
+    public static void exportJSON(String fileName, JsonElement data, Stage stage) {
```
ManfredKarrer, line 275: Is there a pretty print feature in the json framework? Would be nice to have, but if not no worry.

desktop/src/main/java/bisq/desktop/main/dao/governance/result/VoteResultView.java
ManfredKarrer: Could you add all fields (if missing, havve not checked all) from the domain objects which are currently not displayed in the UI.
We want to add popups for that anyway and for the json I think its good to have the complete domain data.

The CompensationProposal is missing `requestedBsq` and `bsqAddress`. As said the UI does not show all that yet but would be good to have that in the json.

ManfredKarrer: For Proposal there are a few subclasses with diff. fields...

ManfredKarrer (changes requested):
Thanks for the fast implementation! Looks basically very good, just a few Nits...

ben-kaufman:
@ManfredKarrer Fixed all the issues above. Attaching here new screenshot and JSON.
screen shot 2019-03-06 at 23 40 37

[voteResultsHistory.json.zip](https://github.com/bisq-network/bisq/files/2938409/voteResultsHistory.json.zip)

ManfredKarrer (approved):
utACK

ManfredKarrer:
Great thanks for the fast fixes! We can need devs like you! ;-)

ManfredKarrer:
@ben-kaufman 
Thanks for adding the translations as well but unfortunately those will be overwritten at next release from Transifex files. We only should edit the default property file in english and the translations have to be done in Transifex. I will leave the commits, just wanted to let you know that the will be lost if no one will move them to Transifex.

PR #2505, opened 2019-03-07
Add removed deprecated payment methods again as it would break trade history

ManfredKarrer:
We got some reports that the trade history was gone and it turned out
the removal of the deprecated payment methods caused an exception when
reading the persisted closed trades. So we have to keep that code to not
break old db files.


devinbileck (approved):
utACK
Looks good to me. However, this is my first exposure to this area of the code...

ripcurlx (approved):
ACK

Issue #2396, opened 2019-02-08
Error  Launching Bisq  on OSX


alemiranda19:
Hello,

I had download Bisq 5 times and restart it about 20 times. Not able  to make it  work.  I do not know what  the problem is. Any help, please? I always get out at 2/4 Thor Node  Create. After that it  just disappear without any other massage.

<img width="559" alt="captura de pantalla 2019-02-02 a las 11 30 57" src="https://user-images.githubusercontent.com/47457297/52497067-02fdfa80-2bcd-11e9-9a69-d924c07283b6.png">

Thanks!

ghost:
Hello @alemiranda19 
There are regularly problems with AV and Tor.
If you have an AV running, you should either tell it that Bisq is safe, either try turning the AV off.

I guess that you use the latest v0.9.3 Bisq version ?

alemiranda19:
Thank you, I’m using a <strike>IOS</strike> OSX computer (macpro) sorry for my ignorance, what is an AV? Yes I had downloaded the latest version.

Thank you

ghost:
apologies, AV = AntiVirus.
Sometimes it could also be a firewall issue.

alemiranda19:
ha,ha

Thanks!  

No AV, Malwarebytes OFF, VPN OFF, still the same problem.... Do not really know what the problem is...

thank you for your help.

Alex

alemiranda19:
Thank you,

I´m using an Apple Computer. MacOS Mojave.
I had tried everything, turn off AV, turn off the VPN just in case.  But the Thor node is not able to be created.
Weird thing is that the first time I download it, it work, then I was not able to log in again so it forced me to erase it and download it again. Since then, not able to connect…..

Thank you,


De: Harry MacFinned <notifications@github.com>
Responder a: bisq-network/bisq <reply@reply.github.com>
Fecha: sábado, 9 de febrero de 2019, 9:29
Para: bisq-network/bisq <bisq@noreply.github.com>
CC: Alejandro Miranda <1938uh@gmail.com>, Mention <mention@noreply.github.com>
Asunto: Re: [bisq-network/bisq] Error Launching Bisq on IOS (#2396)


Hello @alemiranda19<https://github.com/alemiranda19>
What OS do you use ? windows ?
There are regularly problems with AV and Tor under Windows.
If you have an AV running, you should either tell it that Bisq is safe, either try turning the AV off.

ripcurlx:
[...]

PR #2508, opened 2019-03-07
Netlayer 0.6.4

freimair:
should make error messages more clear (#2396)


ripcurlx (approved):
ACK - connecting over Tor is still working for me

PR #2511, opened 2019-03-07
Fix error and improve wording for security deposit

ripcurlx:
**For BTC seller**
bildschirmfoto 2019-03-07 um 18 14 35
**For BTC buyer**
bildschirmfoto 2019-03-07 um 18 14 22


ManfredKarrer (approved):
utACK

PR #2512, opened 2019-03-07
Show the correct BSQ or BTC trading fee if it is maker or taker

### LINKED ISSUE #2502: Trading fee incorrect displayed after successful trade
lf the maker of the trade uses as fee currency BSQ and the taker uses a different trade fee currency BTC than the fee is not displayed correctly. Is no production issue, but needs to be fixed for the Mainnet release.

bildschirmfoto 2019-03-06 um 10 39 10
bildschirmfoto 2019-03-06 um 10 35 00

ripcurlx:
Fixes #2502.


ManfredKarrer (approved):
utACK

PR #2513, opened 2019-03-07
Move BSQ dashboard into separate Facts & Figures section

ripcurlx:
~@ManfredKarrer Still in progress as I'll try to add the charts we were talking about as well.
But just to have a fallback scenario PR.

Current state of the "Facts & Figures" section. It sill needs some @pedromvpg love, but for this release it should be good enough I guess.
Bildschirmfoto 2019-03-27 um 15 21 21
Bildschirmfoto 2019-03-27 um 15 09 06
Bildschirmfoto 2019-03-27 um 15 09 20


ManfredKarrer:
Maybe tab should move right to governance and be called more generic. Ecosystem maybe?

Would be nice to have 2 buttons left:
- Network
- Economy

ManfredKarrer:
@ben-kaufman If you like you could join forces here as well ;-)
@ripcurlx is on holiday next week...

ben-kaufman:
@ManfredKarrer I'll take a look on that :)

ripcurlx:
> Maybe tab should move right to governance and be called more generic. Ecosystem maybe?
> 
> Would be nice to have 2 buttons left:
> 
> * Network
> * Economy

The reason why I would have it on the left is that it would open as the default tab when someone clicks on the DAO section. It should be in the end a dashboard with nice graphs and a high-level visualization of the DAO. Otherwise it would be the BSQ Send screen.

To find the right term for this section:
- What information would be in the network sub-section?

Maybe we could also label it `Overview` or `Fact and Figures`?

I was thinking about having in the final version two sub-sections:
- `Dashboard`: High-level graphs and nicely presented KPIs
- `Facts` or `Details`: Where we show something like we have right now

In general in the DAO > Facts and Figures > Dashboard should give the user an overview on the amount of BSQ available over time (BSQ colored), how much is burned and the BSQ/BTC conversion rate.
It should be visually very nice layouted (hint to @pedromvpg), so people go into the DAO section because they love to checkout the current state of the DAO.

ManfredKarrer:
Agree to all  you proposed!

ripcurlx:
@ManfredKarrer It is not perfect, but I think it should be enough for this release. What do you think?

ManfredKarrer:
Looks goot to me!

ManfredKarrer (approved):
utACK

Issue #2516, opened 2019-03-07
Wallet shows 0 balance after restore seed words


ManfredKarrer:
The AddressEntryList does not get deleted/updated and that is the reason why the balance in the UI does not show the balance of the wallet after a restore from seed words. With cmd+e (opens the emergency wallet tool) one can see and withdraw the full wallet balance. Also the Tx list is showing all transactions.

The AddressEntryList is important for the management of offer and trades utxos. I assume it can be deleted in a restore from seed case.

PR #2517, opened 2019-03-08
Fix missing balance at restore from seed words

ManfredKarrer:


ManfredKarrer:
Fixes https://github.com/bisq-network/bisq/issues/2516

devinbileck (approved):
ACK
Verified that the balance was correct and addresses were restored.

core/src/main/java/bisq/core/btc/model/AddressEntryList.java
```diff
@@ -110,16 +111,38 @@ public void onWalletReady(Wallet wallet) {
...
-
             persist();
         }
+
+        // We add a confidence listener to get notified about potential new transactions and
+        // add an address entry list in case it does not exist yet. This is mainly needed for restore from seed words
+        // but can help as well in case the addressEntry list would miss an address where the wallet was received
+        // funds (e.g. if the user sends funds to an address which has not been provided in the main UI - like from the
+        // wallet details window).
+        wallet.addTransactionConfidenceEventListener((w, tx) -> {
```
oscarguindzberg, line 131: TransactionConfidenceEventListener does the job but it seems to be an overkill. TransactionConfidenceEventListener is invoked for each wallet tx every time its "confidence" is updated (eg, when included in a block and every time a new block is mined on top of that block for the next 100 blocks).

Probably you want to use WalletCoinsReceivedEventListener and WalletCoinsSentEventListener (and  ScriptsChangeEventListener if you use the watched script feature of Wallet) instead to be notified of new addresses.

ManfredKarrer: I tried the WalletCoinsReceivedEventListener and WalletCoinsSentEventListener (both combined) but the balance was not correct that way. I have not investigated further why, but if you could find out would be good. I am aware that the TransactionConfidenceEventListener is a bit heavy.
If you test it you need a wallet which is a bit older and has a few addresses where it has send/received. During the spv chain sync you can check if the UI balance is the same as the one from the wallet shown at the popup opened from cmd+e.

PR #2523, opened 2019-03-11
Fix index out of bound issues

ManfredKarrer:
At Capabilities I got an exception after going back from a new dev
branch to master and getting p2p network data from nodes still running
on the dev branch as that newly added capability was outside of the
existing scope.

I found another potential issue with DaoPhase.Phase and added a check
there as well.


freimair (approved):
ACK

PR #2525, opened 2019-03-11
[WIP] Payment account support for the API

blabno:
The only non-experimental endpoint is to get list of payment accounts so that user can retrieve account's id to use during trade.
The rest of endpoints (create, remove payment account) are marked as experimental features and are required for integration tests.


ManfredKarrer:
@blabno I fear that PR is too far away from master. Sorry for the delay due all the circumstances, but probably better to make a new PR once devs have resources for a profound review and test.

Issue #2503, opened 2019-03-06
Invalid rate for btc/xdr


vlad-khramov:
![image](https://user-images.githubusercontent.com/1215671/53893765-e7590900-403f-11e9-8677-ab5590e0806f.png)

xdr/btc is 2768. It is wrong. Must be btc/xdr = 2768  (1 xdr ~= 1.39)

For example see https://www.coingecko.com/en/coins/bitcoin/xdr

--
Orders with XDR couldn't be created since difference with this rate > 30%

ghost:
@quantum13 
yes there seems to be a problem with this price.
(I can't even find XDR on poloniex.com)

vlad-khramov:
0.9.5

ripcurlx:
If I start in locally on Regtest I actually see `N/A` which is correct as Poloniex doesn't deliver a price ticker for this altcoin. On Production I get from the pricenodes the wrong value as mentioned above.

@ManfredKarrer Do you have an idea how the price node gets this ticker if Poloniex doesn't return anything?

ghost:
Something is provided, since 
eg http://5bmpx76qllutpcyp.onion/getAllMarketPrices
returns a XDR field.


180 |  
-- | --
currencyCode | "XCP"
price | 0.00042221
timestampSec | 1551969634894
provider | "POLO"
181 |  
-- | --
currencyCode | "XDR"
price | 2772.27
timestampSec | 1551969455149
provider | "BTCA_G"

but it's not from Poloniex.

(I guess BTCA_G means BitcoinAverage ?)

I notice that the field #180 is XCP, provided from Poloniex, and which seems delisted. (neither in the display altcoins list, neither in the add altcoin list).
Maybe there is some slippage between the 2 ?

ripcurlx:
The problem is that the PR for XDR didn't comply with https://docs.bisq.network/exchange/howto/list-asset.html#the-asset-ticker-symbol-must-not-conflict-with-national-currency-tickers.

There is a fiat currency with the same currency code XDR (https://en.wikipedia.org/wiki/Special_drawing_rights). 

So the PR shouldn't have been merged in the first place.

ManfredKarrer:
xdr has to be removed as it conflicts with our requirements.

vlad-khramov:
@ManfredKarrer I will create PR for renaming XDR to something else, sorry

ManfredKarrer:
https://bitcoinaverage.com/en/bitcoin-price/btc-to-xdr

PR #2529, opened 2019-03-11
rename XDR ticker (conflict with ISO 4217)

vlad-khramov:
See #2503

### DIFF
--- a/assets/src/main/java/bisq/asset/coins/XDR.java
+++ b/assets/src/main/java/bisq/asset/coins/XDR.java
@@ -22,7 +22,7 @@ import bisq.asset.Coin;
 public class XDR extends Coin {
 
     public XDR() {
-        super("XDR", "XDR", new Mile.MileAddressValidator());
+        super("XDR", "XDR0", new Mile.MileAddressValidator());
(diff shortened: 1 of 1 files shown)


blabno (commented):
Shouldn't tickers be uppercase?

vlad-khramov:
I have not found this restriction in https://docs.bisq.network/exchange/howto/list-asset.html
and there is at least one coin with not uppercase ticker
https://github.com/bisq-network/bisq/blob/master/assets/src/main/java/bisq/asset/coins/QMCoin.java

Should I make it uppercased?

ManfredKarrer:
I think we don't have a strict requirement for that but I would say lets keep it uppercase. Uppercase is the standard.

vlad-khramov:
Can I use digit in the beginning of ticker ("1XDR")?
Creators of blockchain want bXDR or 1XDR (reason - XDR is highlighted) since simple XDR is prohibited

blabno:
Man, ticker is a ticker. There is a name where you can play a bit.

blabno (approved):
ACK

PR #2532, opened 2019-03-11
Add hash of dao state

ManfredKarrer:
@sqrrm 
Here is the current state:
Screen Shot 2019-03-11 at 18 41 13

It is still missing the p2p network data monitoring and tests with reorgs.


core/src/main/java/bisq/core/dao/state/model/blockchain/BaseTxOutput.java
```diff
@@ -45,9 +45,10 @@
     protected final long value;
     protected final String txId;
 
-    // Only set if dumpBlockchainData is true
+    // Before v0.9.6 it was only set if dumpBlockchainData was set to true but we changed that with 0.9.6
+    // so that is is always set. We still need to support it because of backward compatibility.
```
sqrrm, line 49: Perhaps we should do a cleanup before starting testing on mainnet. That would break the regtestnet we have running now though.

ManfredKarrer:
Tested with DAO TESTNET and looks good so far:

Screen Shot 2019-03-12 at 00 33 51

core/src/main/java/bisq/core/dao/state/DaoStateService.java
```diff
@@ -162,6 +173,11 @@ public int getChainHeight() {
         return daoState.getCycles();
     }
 
+    public void addCycle(Cycle firstCycle) {
```
sqrrm: From what I understand this is to add any cycle, not just the first one, would be better to call the argument cycle rather than firstCycle

core/src/main/java/bisq/core/dao/state/monitoring/DaoStateMonitoringService.java
```diff
@@ -0,0 +1,263 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
+
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+/**
+ * Monitors the DaoState with using a hash fo the complete daoState and make it accessible to the network for
```
sqrrm:
```suggestion
 * Monitors the DaoState with using a hash fo the complete daoState and make it accessible to the network
```

core/src/main/java/bisq/core/dao/state/monitoring/DaoStateMonitoringService.java
```diff
@@ -0,0 +1,263 @@
...
+ * so we can detect quickly if any consensus issue arise. The data does not contain any private user
+ * data so sharing it on demand has no privacy concerns.
```
sqrrm: Could this be an attack vector, allowing too many requests as a DOS?

core/src/main/java/bisq/core/dao/state/monitoring/DaoStateMonitoringService.java
```diff
@@ -0,0 +1,263 @@
...
+
+
+    ///////////////////////////////////////////////////////////////////////////////////////////
+    // DaoStateListener
+    ///////////////////////////////////////////////////////////////////////////////////////////
+
+    @Override
+    public void onParseBlockChainComplete() {
+        parseBlockChainComplete = true;
+        daoStateNetworkService.addListeners();
```
sqrrm: Won't this add a listener every time a snapshot is reapplied. So after reorgs there would be duplicate listeners?

core/src/main/java/bisq/core/dao/state/monitoring/DaoStateMonitoringService.java
```diff
@@ -0,0 +1,263 @@
...
+            if (height == genesisTxInfo.getGenesisBlockHeight()) {
+                prevHash = new byte[0];
+            } else {
+                log.warn("DaoStateBlockchain is empty but we received the block which was not the genesis block. " +
+                        "We stop execution here.");
+                return;
+            }
+        } else {
+            // TODO check if in reorg cases it might be a valid case
+            checkArgument(height > daoStateBlockchain.getLast().getBlockHeight(),
```
sqrrm: Shouldn't this check be `height == daoStateBlockchain.getLast().getBlockHeight() + 1`

core/src/main/java/bisq/core/dao/state/monitoring/DaoStateMonitoringService.java
```diff
@@ -0,0 +1,263 @@
...
+        daoStateBlockchain.add(daoStateBlock);
+        listeners.forEach(Listener::onDaoStateBlockchainChanged);
+
+        log.info("Add daoStateBlock at processNewBlock:\n{}", daoStateBlock);
+
+        // We only broadcast after parsing of blockchain is complete
+        if (parseBlockChainComplete) {
+            // We delay broadcast to give peers enough time to have received the block.
+            // Otherwise they would ignore our data if received block is in future to their local blockchain.
+            int delayInSec = 1 + new Random().nextInt(5);
```
sqrrm: Is this a reasonable assumption, that most peers get the block within 1s of each other?

core/src/main/java/bisq/core/dao/state/monitoring/DaoStateMonitoringService.java
```diff
@@ -0,0 +1,296 @@
...
+        daoStateNetworkService.requestHashes(genesisTxInfo.getGenesisBlockHeight(), peersAddress);
+    }
+
+    public void applySnapshot(LinkedList<DaoStateHash> persistedDaoStateHashChain) {
+        // We could got a reset from a reorg, so we clear all and start over from the genesis block.
+        daoStateHashChain.clear();
+        daoStateBlockchain.clear();
+        daoStateNetworkService.reset();
+
+        log.info("Apply snapshot: Last daoStateHash={}", persistedDaoStateHashChain.getLast());
```
sqrrm: This throws an exception for an empty persistedDaoStateHashChain.

core/src/main/java/bisq/core/dao/state/DaoStateService.java
ManfredKarrer: Correct. Will change it in upcoming commits.

core/src/main/java/bisq/core/dao/state/monitoring/DaoStateMonitoringService.java
ManfredKarrer: On the P2P network level it would be disconnected it it gets heavy, but basically yes. We should consider to limit it somehow.

core/src/main/java/bisq/core/dao/state/monitoring/DaoStateMonitoringService.java
ManfredKarrer: Good spot. Will add a flag to protect against multiple calls.

core/src/main/java/bisq/core/dao/state/monitoring/DaoStateMonitoringService.java
ManfredKarrer: Yes correct. Will change and also error msg.

core/src/main/java/bisq/core/dao/state/monitoring/DaoStateMonitoringService.java
ManfredKarrer: Hm.... hard to  say. Prob. its too short. For testing its better ;-). Will will add a TODO to increase it later. 5-10 sec? It is not super critical fi you receive a future block you just drop it.

ManfredKarrer:
@sqrrm I think it is now complete. Could you cross check again? I tested re-orgs as well and seems that all works so far. Tested also backward compatibility with master and all ok.

sqrrm (commented):
My comment on throwing an exception only happened midway through these commits. I still think it can happen but couldn't recreate it.

Everything looks good while running on the current tip. I think it's basically good to go.

core/src/main/java/bisq/core/dao/state/monitoring/DaoStateMonitoringService.java
ManfredKarrer: Added a check

PR #2538, opened 2019-03-14
Fix exception handling during Tor setup

devinbileck:
A recent change in the netlayer is now seeing exceptions wrapped in
a TorCtlException if an error occurs while setting up Tor.

So when an issue such as "Auth cookie not created" (#2398) occurs,
which was previously raised as an IOException, it was restarting Tor
rather than showing the error message to the user.


devinbileck:
@freimair can you confirm exceptions wrapped in a TorCtlException is the expected behaviour?

ManfredKarrer:
@freimair I leave the review to you.

freimair (approved):
ACK

Slowly cleaning up the netlayer stuff. Maybe I introduce some custom exceptions instead of wrapping all and everything in this dubious TorCtlException...

PR #2541, opened 2019-03-15
Added mining fee to un/lock bond popup.

### LINKED ISSUE #2296: Show applicable mining fee in bond lockup/unlock confirmation dialog
The confirmation message dialog that appears when locking up or unlocking a bond does not mention that there is an applicable mining fee. Mentioning the mining fee would be beneficial so the user is aware there is a BTC cost involved.
image

For reference, a proposal transaction indicates the mining fee.
image

ben-kaufman:
Fix #2296 

Added mining fee costs to the lock and unlock bond popups.

Screen Shot 2019-03-15 at 20 08 11

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/dao/bonding/BondingViewUtils.java
+++ b/desktop/src/main/java/bisq/desktop/main/dao/bonding/BondingViewUtils.java
@@ -24,6 +24,8 @@ import bisq.desktop.main.funds.deposit.DepositView;
 import bisq.desktop.main.overlays.popups.Popup;
 import bisq.desktop.util.GUIUtil;
 
+import bisq.core.btc.exceptions.TransactionVerificationException;
+import bisq.core.btc.exceptions.WalletException;
 import bisq.core.btc.setup.WalletsSetup;
 import bisq.core.dao.DaoFacade;
 import bisq.core.dao.governance.bond.lockup.LockupReason;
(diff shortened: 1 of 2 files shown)


desktop/src/main/java/bisq/desktop/main/dao/bonding/BondingViewUtils.java
```diff
@@ -147,13 +162,18 @@ public void unLock(String lockupTxId, Consumer<String> resultHandler) {
 
             try {
                 if (!DevEnv.isDevMode()) {
+                    Tuple2<Coin, Integer> miningFeeAndTxSize = daoFacade.getMiningFeeAndTxSize(unlockAmount);
+                    Coin miningFee = miningFeeAndTxSize.first;
```
ManfredKarrer, line 166: Unfortunately daoFacade.getMiningFeeAndTxSize is not correctly named. It is only valid for the blind vote tx (I will rename the method in another commit). We need to use the correct tx which will be created at the unlock tx (without triggering any side effects as the user can cancel).

desktop/src/main/java/bisq/desktop/main/dao/bonding/BondingViewUtils.java
```diff
@@ -101,18 +105,29 @@ private void lockupBond(byte[] hash, Coin lockupAmount, int lockupTime, LockupRe
...
-                                lockupTime,
-                                duration
-                        ))
-                        .actionButtonText(Res.get("shared.yes"))
-                        .onAction(() -> publishLockupTx(hash, lockupAmount, lockupTime, lockupReason, resultHandler))
-                        .closeButtonText(Res.get("shared.cancel"))
-                        .show();
+                try {
+                    Tuple2<Coin, Integer> miningFeeAndTxSize = daoFacade.getMiningFeeAndTxSize(lockupAmount);
+                    Coin miningFee = miningFeeAndTxSize.first;
```
ManfredKarrer, line 110: Unfortunately daoFacade.getMiningFeeAndTxSize is not correctly named. It is only valid for the blind vote tx (I will rename the method in another commit). We need to use the correct tx which will be created at the lockup (without triggering any side effects as the user can cancel).

ManfredKarrer:
It might be not so trivial to do that as in the other cases, that was why I left it out. Would need to look myself more closely to give more hints.

ManfredKarrer:
Got replaced by https://github.com/bisq-network/bisq/pull/2577

PR #2546, opened 2019-03-16
Add dao betatest network

ManfredKarrer:
This adds a mainnet genesis tx for testing as a new network type. It is only intended to be used by testers who run from source code and get in direct contact with me (as it is now real BTC I prefer to keep the test group small and focussed).

### DIFF
--- a/core/src/main/java/bisq/core/dao/state/GenesisTxInfo.java
+++ b/core/src/main/java/bisq/core/dao/state/GenesisTxInfo.java
@@ -54,6 +54,10 @@ public class GenesisTxInfo {
     private static final String DAO_TESTNET_GENESIS_TX_ID = "cb316a186b9e88d1b8e1ce8dc79cc6a2080cc7bbc6df94f2be325d8253417af1";
     private static final int DAO_TESTNET_GENESIS_BLOCK_HEIGHT = 104; // 2019-02-19
 
+    //TODO create new one
+    private static final String DAO_BETANET_GENESIS_TX_ID = "81855816eca165f17f0668898faa8724a105196e90ffc4993f4cac980176674e";
+    private static final int DAO_BETANET_GENESIS_BLOCK_HEIGHT = 524717; // 2018-05-27
+
     private static final String REGTEST_GENESIS_TX_ID = "30af0050040befd8af25068cc697e418e09c2d8ebd8d411d2240591b9ec203cf";
     private static final int REGTEST_GENESIS_BLOCK_HEIGHT = 111;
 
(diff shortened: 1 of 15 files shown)


ManfredKarrer:
A BSQ blockexplorer is not supported yet but we are working on it.

ManfredKarrer:
I have started syncing btc core mainnet for the seed node. Will take a bit until that is ready. So if you want to run it before you need to provide your own seed node with --seedNodes=... can be localhost seed. 

Example local seed:
--seedNodes=localhost:3004 --baseCurrencyNetwork=BTC_DAO_BETANET --useLocalhostForP2P=true --nodePort=3004 --appName=bisq-BTC_DAO_BETANET_LOCAL_seed_3004 --daoActivated=true --fullDaoNode=true --rpcUser=xxx --rpcPassword=xxx--rpcPort=8332 --rpcBlockNotificationPort=5100

Example local Bisq app:
--seedNodes=localhost:3004 --baseCurrencyNetwork=BTC_DAO_BETANET --useLocalhostForP2P=true --nodePort=3111 --appName=bisq-BTC_DAO_BETANET_LOCAL_Ali_dao_lite --daoActivated=true

core/src/main/java/bisq/core/dao/state/GenesisTxInfo.java
```diff
@@ -43,19 +43,25 @@
...
     private static final int TESTNET_GENESIS_BLOCK_HEIGHT = 1446300; // 2018-12-02
+    private static final Coin TESTNET_GENESIS_TOTAL_SUPPLY = Coin.parseCoin("2.5"); // 2.5M BSQ / 2.50000000 BTC
 
     private static final String DAO_TESTNET_GENESIS_TX_ID = "cb316a186b9e88d1b8e1ce8dc79cc6a2080cc7bbc6df94f2be325d8253417af1";
     private static final int DAO_TESTNET_GENESIS_BLOCK_HEIGHT = 104; // 2019-02-19
+    private static final Coin DAO_TESTNET_GENESIS_TOTAL_SUPPLY = Coin.parseCoin("2.5"); // 2.5M BSQ / 2.50000000 BTC
+
+    private static final String DAO_BETANET_GENESIS_TX_ID = "0bd66d8ff26476b55dfaf2a5db0c659a5d8635566488244df25606db63a08bd9";
+    private static final int DAO_BETANET_GENESIS_BLOCK_HEIGHT = 567405; // 2019-03-16
+    private static final Coin DAO_BETANET_GENESIS_TOTAL_SUPPLY = Coin.parseCoin("0.49998644"); // 1000 BSQ / 0.49998644 BTC
```
devinbileck, line 60: Seems like an odd amount, but should be equal to 499,986.44 BSQ (not 1000)?

devinbileck (approved):
utACK

core/src/main/java/bisq/core/dao/state/GenesisTxInfo.java
ManfredKarrer: Ah yes, I did not care much with fee and change when i created it, but it is not really important as I made amount now flexible. The real will also be not an even number if we add add up all contributions with the 2.5M from the original genesis.
Will fix the comment in another commit

sqrrm (approved):
utACK

I'm syncing my bitcoin node for the explorer, should be ready tomorrow from the look of it.

PR #2552, opened 2019-03-18
Refactor capabilities

freimair:
Another step towards a cleaner code base.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/peers/PeerManager.java
+++ b/p2p/src/main/java/bisq/network/p2p/peers/PeerManager.java
@@ -176,7 +176,7 @@ public class PeerManager implements ConnectionListener, PersistedDataHost {
         PeerList persistedPeerList = storage.initAndGetPersistedWithFileName("PeerList", 1000);
         if (persistedPeerList != null) {
             long peersWithNoCapabilitiesSet = persistedPeerList.getList().stream()
-                    .filter(e -> e.getCapabilities().hasCapabilities())
+                    .filter(e -> e.getCapabilities().isEmpty())
                     .mapToInt(e -> 1)
                     .count();
             if (peersWithNoCapabilitiesSet > 100) {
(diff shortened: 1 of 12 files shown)


p2p/src/main/java/bisq/network/p2p/peers/PeerManager.java
```diff
@@ -176,7 +176,7 @@ public void readPersisted() {
         PeerList persistedPeerList = storage.initAndGetPersistedWithFileName("PeerList", 1000);
         if (persistedPeerList != null) {
             long peersWithNoCapabilitiesSet = persistedPeerList.getList().stream()
-                    .filter(e -> e.getCapabilities().hasCapabilities())
+                    .filter(e -> e.getCapabilities().isEmpty())
```
ManfredKarrer, line 179: I think here is the ! missing

p2p/src/main/java/bisq/network/p2p/peers/PeerManager.java
```diff
@@ -664,11 +664,11 @@ public void shutDownConnection(NodeAddress peersNodeAddress, CloseConnectionReas
                     // If we have a new connection the supportedCapabilities is empty.
                     // We lookup if we have already stored the supportedCapabilities at the persisted or reported peers
                     // and if so we use that.
-                    if (!supportedCapabilities.hasCapabilities()) {
+                    if (supportedCapabilities.isEmpty()) {
                         Set<Peer> allPeers = new HashSet<>(getPersistedPeers());
                         allPeers.addAll(getReportedPeers());
                         Optional<Peer> ourPeer = allPeers.stream().filter(peer -> peer.getNodeAddress().equals(connection.getPeersNodeAddressOptional().get()))
-                                .filter(peer -> !peer.getCapabilities().hasCapabilities())
+                                .filter(peer -> !peer.getCapabilities().isEmpty())
```
ManfredKarrer, line 671: Here is the boolean logic inverse to the older version.

ManfredKarrer (changes requested):

p2p/src/main/java/bisq/network/p2p/peers/PeerManager.java
freimair: asdf

freimair: This is a bug which has been introduced with the previous refactoring.

The goal is a list of peers which have no capabilities assigned. This list is counted and then reported either as `log.warn` (in case more than hardcoded 100 persisted peers have no capabilities assigned) or a `log.info`. Currently, however, the code counts peers which have capabilities (`hasCapabilities`) assigned. So no breaking bug that shows itself easily.

This change restores the functionality as intended originally by leaving only peers with empty capability sets in the list.

p2p/src/main/java/bisq/network/p2p/peers/PeerManager.java
freimair: This, again, is a bug which has been introduced with the previous refactoring.

The goal is to find "cached" capabilities for a given peer. By filtering for no capabilities (`!peer...hasCapabilities`), we only get a list of peers which have an empty set of capabilities - which does not help in achieving the goal. Again, this bug does not break anything as active peers (with time) report their capabilities and hence, the bug does not show itself easily.

This change, again, restores the intended functionality by filtering for peers with a non-empty (`!peer...isEmpty`) set of capabilities.

p2p/src/main/java/bisq/network/p2p/peers/PeerManager.java
ManfredKarrer: Ah ok. I did not look closer to the intended behaviour...

ManfredKarrer (approved):

ManfredKarrer:
utACK

Issue #2555, opened 2019-03-18
Check if adding a Param causes problems


ManfredKarrer:

ManfredKarrer:
It turned out that changing any field in the Param can break consensus.

Name of the params must not change as that is used for serialisation in Protobuffer. The data fields are not part of the PB serialisation so changes for those would not change the hash for the dao state hash chain.
Though changing the values might break consensus as the validations might return a different result (e.g. a param change proposal was accepted with older min/max values but then after a change it is not valid anymore and might break the consequences of that change. So in fact we MUST not change anything here, only way is to add new entries and don't use the deprecated enum in future releases anymore.

Issue #2554, opened 2019-03-18
Check if adding a bonded role causes problems


ManfredKarrer:

ManfredKarrer:
Changing values here is less critical as atm we dont use that in any consensus context. But it will become more critical on future once we verify mediators if they have set up the right bond.

I suggest we use 2 factors which gets multiplied. In the BondedRoleType there is factor A and in Param there is a new param for the multiplyer B. 

E.g. if Twitter role has factor 20 and initial mulitplyer is 1000 the bond is 20k BSQ. If BSQ prices changes to 5 USD we want to adjust the mulitplyer so that the USD value of the bond is the same. So we change it by voting to 200, and the BSQ amount for the bond will be 4000 BSQ (20k USD).

With that we only need to get the relation between bonds right. The expected volatility of BSQ will be handled by voting on the muliplyier in Params.

Issue #2545, opened 2019-03-16
Invalidate proposals with invalid data


ManfredKarrer:
As @devinbileck has tested out in the current proposals he managed to post a proposal with a too high amount (100 000 BSQ is limit). I assume he did that by editing the code as the UI does not permit that. But of course we need to verify that on the domain layer as well and invalidate such proposals so that they are not displayed and would not become valid if accepted in voting. I assume all those checks for the proposal values are atm not dont on the domain layer and also incomplete.

<img width="987" alt="Screen Shot 2019-03-16 at 15 33 19" src="https://user-images.githubusercontent.com/1449498/54481294-ed4da780-4800-11e9-8c29-f7efc18c4f27.png">

ManfredKarrer:
@ben-kaufman If you have time for that would be relatively high prio.

devinbileck:
@ManfredKarrer that was a valid proposal for changing the max BSQ amount for compensation requests. And within the acceptable max increase limit (x2).
I think you confused it with a compensation request?

ManfredKarrer:
Ah yes, I confused it... Then all is OK. 
But can be still that validation is missing on the domain side....

ghost:
Concerning this point of users trading with Bisq home compiled source code (and possibly modified with bad intentions).
Is there atm any signature/fingerprint/checksum (or other technical mean) produced by the appli and showed to the peer nodes in order to prove that the node uses the official and non modified source code ?

ManfredKarrer:
@HarryMacfinned Any modification need to be detected by the consensus code. So anything the user creates is only valid if your node considers it valid. Atm probl. some such checks are missing but of course we cannot rely on validation done at the UI level...

Issue #2539, opened 2019-03-15
Finalize ChangeParamValidator


ManfredKarrer:
The ChangeParamValidator requires more work to get it final and find reasonable values for min/max changes for all params.

PR #2556, opened 2019-03-18
Improve data structure, validation and performance

ManfredKarrer:
Fixes https://github.com/bisq-network/bisq/issues/2554
Fixes https://github.com/bisq-network/bisq/issues/2555
Fixes https://github.com/bisq-network/bisq/issues/2545
Fixes https://github.com/bisq-network/bisq/issues/2539
Fixes https://github.com/bisq-network/bisq/issues/2415

Adding new params and bonded roles is possible after adding the UNDEFINED enums and avoid throwing exceptions if enum is not found.


ManfredKarrer:
Sorry this PR becomes much bigger as expected. ;-( But as there are some breaking changes, I still prefer to have it in one and with one test session.

core/src/main/java/bisq/core/dao/monitoring/ProposalStateMonitoringService.java
```diff
@@ -134,15 +134,18 @@ public void onDaoStateChanged(Block block) {
         int blockHeight = block.getHeight();
         int genesisBlockHeight = genesisTxInfo.getGenesisBlockHeight();
 
+        boolean hashChainUpdated = false;
         if (proposalStateBlockChain.isEmpty() && blockHeight > genesisBlockHeight) {
             // Takes about 150 ms for dao testnet data
             long ts = System.currentTimeMillis();
             for (int i = genesisBlockHeight; i < blockHeight; i++) {
-                maybeUpdateHashChain(i);
+                hashChainUpdated = hashChainUpdated || maybeUpdateHashChain(i);
```
devinbileck: Does this not have the unintended consequence of only updating the hashchain for the first cycle? Since once maybeUpdateHashChain returns true (first block of blind vote phase), this statement will short-circuit and not call maybeUpdateHashChain for any other blocks?

core/src/main/java/bisq/core/dao/monitoring/ProposalStateMonitoringService.java
```diff
@@ -134,15 +134,18 @@ public void onDaoStateChanged(Block block) {
...
+            }
+            if (hashChainUpdated) {
+                log.info("updateHashChain for {} items took {} ms",
```
devinbileck: Perhaps could be clarified as blocks rather than items?

core/src/main/java/bisq/core/dao/monitoring/ProposalStateMonitoringService.java
devinbileck: Never mind. Just saw your subsequent commit :)

core/src/main/java/bisq/core/dao/governance/proposal/param/ChangeParamValidator.java
```diff
@@ -214,24 +222,64 @@ private void validateParamValue(Param param, String inputValue, int chainHeight)
...
                 break;
             case PHASE_RESULT:
                 if (isMainnet)
-                    checkArgument(newValueAsBlock >= 1, "The break must have at least 1 block");
+                    checkArgument(inputValueAsBlock >= 1, "The break must have at least 1 block");
                 break;
+        }
+    }
+
+    private void validateAddressValue(String currentParamValue, String inputValue) {
```
devinbileck: Should add `throws AddressFormatException` to this method signature

core/src/main/java/bisq/core/dao/governance/proposal/param/ChangeParamValidator.java
```diff
@@ -214,24 +222,64 @@ private void validateParamValue(Param param, String inputValue, int chainHeight)
...
+     */
+    @VisibleForTesting
+    static Result getChangeValidationResult(double currentValue, double newValue, double min, double max) {
+        checkArgument(min >= 0, "Min must be >= 0");
+        checkArgument(max >= 0, "Max must be >= 0");
+        if (currentValue == newValue)
+            return Result.SAME;
+
+        if (max == 0)
+            return Result.OK;
```
devinbileck: Are you assuming that if max is zero, then so is min? If max is zero but min is non-zero (not sure if there is a use case for that) and the change is <1, then it wont validate min.

core/src/main/java/bisq/core/dao/monitoring/ProposalStateMonitoringService.java
ManfredKarrer: Fixed in new commit.

core/src/main/java/bisq/core/dao/governance/proposal/param/ChangeParamValidator.java
ManfredKarrer: AddressFormatException is a RuntimeException. But agree whould be more clean to add it. Will consider to wrap it into ValidationException.

ManfredKarrer: Used ParamValidationException in last commit.

core/src/main/java/bisq/core/dao/governance/proposal/param/ChangeParamValidator.java
ManfredKarrer: Yes that is not correct. Was a bit too lazy... Basically there is no use case for setting one to 0 and the other not. But should handle it correclty... The whole validation code could need more love...

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1540,7 +1544,9 @@ dao.bond.bondedRoleType.BTC_NODE_OPERATOR=BTC node operator
 # suppress inspection "UnusedProperty"
 dao.bond.bondedRoleType.MARKETS_OPERATOR=Markets operator
 # suppress inspection "UnusedProperty"
-dao.bond.bondedRoleType.BSQ_EXPLORER_OPERATOR=BSQ explorer operator
+dao.bond.bondedRoleType.BSQ_EXPLORER_OPERATOR=Markets operator
```
devinbileck: I assume this string should not have been changed?

ManfredKarrer: Ups, thanks.

ManfredKarrer:
@sqrrm @devinbileck @ripcurlx
I know that is a monster PR and hard to review.
Still would love to get it soon merged so we have some time for dev testing before next release.
I will continue testing the next hours, but tested already quite a lot. If nobody can review in the next hour I will consider to merge despite an ACK. If any NACK arrives I will not merge of course....

devinbileck (approved):
utACK
Overall looks fine to me. However, I am not well experienced in this area of the code and can just make general observations.
Only concern I had was regarding translations for the validation messages. I assume that is still todo and will be completed separately?

ManfredKarrer:
Yes that is on my todo list. I try to focus on the critical issue first and keep that for later.

core/src/main/java/bisq/core/dao/DaoFacade.java
```diff
@@ -680,4 +685,27 @@ public double getRequiredThreshold(Proposal proposal) {
     public Coin getRequiredQuorum(Proposal proposal) {
         return proposalService.getRequiredQuorum(proposal);
     }
+
+    public long getRequiredBond(Optional<RoleProposal> roleProposal) {
```
sqrrm, line 673: I feel these functions should reside in DaoStateService for consistency. This works, most other actual work is done where it's more relevant rather than in DaoFacade itself.

sqrrm (commented):
utACK

core/src/main/java/bisq/core/dao/DaoFacade.java
ManfredKarrer: Hm, not sure. Proposal is another domain as DaoState so I prefer to keep it contained in ProposalService. Sure there is much overlap but DaoStateService is already huge so in doubt I prefer to keep it in ProposalService. With DaoFacade I am not sure if we should keep it mid term. Its quite boilerplate.

sqrrm: Just leave it for now. But would be nice to find a better place for it eventually.

PR #2567, opened 2019-03-21
Fix BTC status incorrectly hidden on splash screen

devinbileck:
Issue: When launching a new Bisq instance (i.e. new data directory) and
remaining on the user agreement screen for >90 seconds while reading it,
once you accept the agreement the BTC status was not being shown
on the splash screen.
image

Cause: showTorNetworkSettingsTimer gets triggered after 90 seconds
and since Tor is not started until after accepting the user agreement,
it was incorrectly assuming that Tor is not working and as a result
hiding the BTC status.

Fix: Don't hide the BTC status in showTorNetworkSettingsTimer.
image

If there is an issue with Tor, splashP2PNetworkErrorMsgListener handles
hiding the BTC status.
image


ManfredKarrer (commented):
utACK

PR #2569, opened 2019-03-22
Suppress popup at shutdown when open offers are disabled

### LINKED ISSUE #2115: Still receive warning about open offers at shutdown when trades is edited/set as INACTIVE.
as per subject.

devinbileck:
Issue: When shutting down the application with disabled open offers, the
confirmation popup was still being shown informing the user there are
open offers.

Fix: On shutdown, check all open offers and only show the popup if any
of them are enabled.

Fixes #2115


ManfredKarrer (approved):
utACK

PR #2576, opened 2019-03-23
Fix get all bonds not returning not active bonds

ManfredKarrer:
Completes https://github.com/bisq-network/bisq/pull/2526


devinbileck (approved):
utACK
Aside from the conflict...

ripcurlx (approved):
ACK - and resolved conflict with master.

PR #2577, opened 2019-03-23
Added mining fee to un/lock bond popup

ManfredKarrer:
Completes https://github.com/bisq-network/bisq/pull/2541
Fixes https://github.com/bisq-network/bisq/issues/2296


ben-kaufman (approved):
utACK

PR #2584, opened 2019-03-25
Add ISSUANCE_LIMIT param and check for max issuance per cycle

ManfredKarrer:
We limit the possible max. issuance (including reimbursements) to the
valued defined in the ISSUANCE_LIMIT param.
In case we exceed that limit the whole cycle becomes invalid.

The main reason for that feature is limitation of max.
damage in case that an attacker manages to create unjustified issuance.

For valid cases we consider such a case as social consensus problem as
the majority of voters should be aware of that limit before they
accept such a high issuance. If known in advance that an extraordinary
high issuance is expected the DAO parameter can be changed.


sqrrm (approved):
utACK

PR #2587, opened 2019-03-26
Add irregular txType, add check for total balance, prevent proposal withhold attack

ManfredKarrer:


core/src/main/java/bisq/core/dao/governance/votereveal/VoteRevealService.java
```diff
@@ -179,8 +179,11 @@ private void maybeRevealVotes(int chainHeight) {
                 .filter(myVote -> myVote.getRevealTxId() == null) // we have not already revealed
                 .forEach(myVote -> {
                     boolean isInVoteRevealPhase = periodService.getPhaseForHeight(chainHeight) == DaoPhase.Phase.VOTE_REVEAL;
+                    // If we would create the tx in the last block it would be confirmed in the best case in th next
+                    // block which would be already the break and would invalidate the vote reveal.
+                    boolean isNotLastBlockInPhase = chainHeight != periodService.getLastBlockOfPhase(chainHeight, DaoPhase.Phase.VOTE_REVEAL);
```
sqrrm: Would be easier to understand the code if the name was isLastBlockInPhase and the check later was for !isLastBlockInPhase.

sqrrm:
```suggestion
                    boolean isLastBlockInPhase = chainHeight == periodService.getLastBlockOfPhase(chainHeight, DaoPhase.Phase.VOTE_REVEAL);
```

core/src/main/java/bisq/core/dao/governance/votereveal/VoteRevealService.java
```diff
@@ -179,8 +179,11 @@ private void maybeRevealVotes(int chainHeight) {
...
                     boolean isBlindVoteTxInCorrectPhaseAndCycle = periodService.isTxInPhaseAndCycle(myVote.getTxId(), DaoPhase.Phase.BLIND_VOTE, chainHeight);
-                    if (isInVoteRevealPhase && isBlindVoteTxInCorrectPhaseAndCycle) {
+                    if (isInVoteRevealPhase && isNotLastBlockInPhase && isBlindVoteTxInCorrectPhaseAndCycle) {
```
sqrrm:
```suggestion
                    if (isInVoteRevealPhase && !isLastBlockInPhase && isBlindVoteTxInCorrectPhaseAndCycle) {
```

core/src/main/java/bisq/core/dao/node/parser/TxParser.java
```diff
@@ -249,17 +256,17 @@ private void processIssuance(int blockHeight, TempTx tempTx, long bsqFee) {
...
             // This could be a valid compensation request that failed to be included in a block during the
             // correct phase due to no fault of the user. Better not burn the change as long as the BSQ inputs
             // cover the value of the outputs.
-            // TODO don't burn in such cases
-            tempTx.setTxType(TxType.INVALID);
+            // We tolerate such an incorrect tx and do not burn the BSQ
+            tempTx.setTxType(TxType.IRREGULAR);
 
-            // TODO don't burn in such cases
+            // Make sure the optionalIssuanceCandidate is set the BTC
```
sqrrm:
```suggestion
            // Make sure the optionalIssuanceCandidate is set to BTC
```

core/src/main/java/bisq/core/dao/node/parser/TxParser.java
```diff
@@ -393,19 +399,19 @@ static TxType evaluateTxTypeFromOpReturnType(TempTx tempTx, OpReturnType opRetur
                 boolean hasCorrectNumOutputs = tempTx.getTempTxOutputs().size() >= 3;
                 if (!hasCorrectNumOutputs) {
                     log.warn("Compensation/reimbursement request tx need to have at least 3 outputs");
-                    // This is not an issuance request but it should still not burn the BSQ change
-                    // TODO do we want to burn in such a case?
-                    return TxType.INVALID;
+                    // This is not a valid issuance tx
+                    // We tolerate such an incorrect tx and do not burn the BSQ
+                    return TxType.IRREGULAR;
```
sqrrm: In this case I would also be ok to burn the BSQ. This is a malformed tx which should not be possible to create with an unmodified client. This is one of the cases where being strict might be worth it as this is about issuance.

core/src/main/java/bisq/core/dao/node/parser/TxParser.java
```diff
@@ -393,19 +399,19 @@ static TxType evaluateTxTypeFromOpReturnType(TempTx tempTx, OpReturnType opRetur
...
                 boolean hasIssuanceOutput = issuanceTxOutput.getTxOutputType() == TxOutputType.ISSUANCE_CANDIDATE_OUTPUT;
                 if (!hasIssuanceOutput) {
                     log.warn("Compensation/reimbursement request txOutput type of output at index 1 need to be ISSUANCE_CANDIDATE_OUTPUT. " +
                             "TxOutputType={}", issuanceTxOutput.getTxOutputType());
-                    // This is not an issuance request but it should still not burn the BSQ change
-                    // TODO do we want to burn in such a case?
-                    return TxType.INVALID;
+                    // This is not a valid issuance tx
+                    // We tolerate such an incorrect tx and do not burn the BSQ
+                    return TxType.IRREGULAR;
```
sqrrm: Same here about being strict. Possibly worth being stricter unless you can see a way that normal usage of the client would result in this kind of malformed issuance tx.

core/src/main/java/bisq/core/dao/node/parser/TxParser.java
```diff
@@ -423,8 +429,8 @@ static TxType evaluateTxTypeFromOpReturnType(TempTx tempTx, OpReturnType opRetur
                 return TxType.PROOF_OF_BURN;
             default:
                 log.warn("We got a BSQ tx with an unknown OP_RETURN. tx={}, opReturnType={}", tempTx, opReturnType);
-                // TODO do we want to burn in such a case?
-                return TxType.INVALID;
+                // We tolerate such an incorrect tx and do not burn the BSQ
+                return TxType.IRREGULAR;
```
sqrrm, line 430: This case I'm not sure about. Perhaps it's good to be this lenient as this is no special kind of tx and would just result in a normal BSQ tx sending BSQ from one output to the next.

desktop/src/main/java/bisq/desktop/main/dao/wallet/dashboard/BsqDashboardView.java
```diff
@@ -199,18 +200,25 @@ private void updateWithBsqBlockChainData() {
         Coin totalUnlockingAmount = Coin.valueOf(daoFacade.getTotalAmountOfUnLockingTxOutputs());
         Coin totalUnlockedAmount = Coin.valueOf(daoFacade.getTotalAmountOfUnLockedTxOutputs());
         Coin totalConfiscatedAmount = Coin.valueOf(daoFacade.getTotalAmountOfConfiscatedTxOutputs());
-        availableAmount = issuedAmountFromGenesis
+        Coin totalUtxoAmount = Coin.valueOf(daoFacade.getTotalAmountOfUnspentTxOutputs());
+
+        daoFacade.getTotalAmountOfConfiscatedTxOutputs();
```
sqrrm:
```suggestion
```

sqrrm (approved):
utACK

I have some suggested changes that you can use if you feel like.

There is also some questions on how lenient we should be with irregular txs for failed issuance, just added my thoughts.

core/src/main/java/bisq/core/dao/governance/votereveal/VoteRevealService.java
ManfredKarrer: Ok, changed it

core/src/main/java/bisq/core/dao/node/parser/TxParser.java
ManfredKarrer: Agree.

core/src/main/java/bisq/core/dao/node/parser/TxParser.java
ManfredKarrer: Agree. But changing it later will be likely not possible (see mess it has created with currnet change).

core/src/main/java/bisq/core/dao/node/parser/TxParser.java
ManfredKarrer: Ok.

core/src/main/java/bisq/core/dao/governance/param/Param.java
```diff
@@ -142,55 +142,55 @@
                     "4" :       // regtest
                     BisqEnvironment.getBaseCurrencyNetwork().isDaoBetaNet() ?
                             "144" :       // daoBetaNet; 1 day
-                            "380",      // testnet or dao testnet (server side regtest); 2.6 days
+                            "134",      // testnet or dao regtest (server side regtest); 2.6 days
```
sqrrm:
```suggestion
                            "134",      // testnet or dao regtest (server side regtest); 0.93 days
```

core/src/main/java/bisq/core/dao/governance/param/Param.java
```diff
@@ -142,55 +142,55 @@
...
+                            "10",       // testnet or dao regtest (server side regtest)
             ParamType.BLOCK, 2, 2),
     PHASE_BLIND_VOTE(BisqEnvironment.getBaseCurrencyNetwork().isMainnet() ?
             "451" :     // mainnet; 3 days
             BisqEnvironment.getBaseCurrencyNetwork().isRegtest() ?
                     "2" :       // regtest
                     BisqEnvironment.getBaseCurrencyNetwork().isDaoBetaNet() ?
                             "144" :       // daoBetaNet; 1 day
-                            "300",      // testnet or dao testnet (server side regtest); 2 days
+                            "134",      // testnet or dao regtest (server side regtest); 2 days
```
sqrrm:
```suggestion
                            "134",      // testnet or dao regtest (server side regtest); 0.93 days
```

core/src/main/java/bisq/core/dao/governance/param/Param.java
```diff
@@ -142,55 +142,55 @@
...
+                            "10",       // testnet or dao regtest (server side regtest)
             ParamType.BLOCK, 2, 2),
     PHASE_VOTE_REVEAL(BisqEnvironment.getBaseCurrencyNetwork().isMainnet() ?
             "451" :     // mainnet; 3 days
             BisqEnvironment.getBaseCurrencyNetwork().isRegtest() ?
                     "2" :       // regtest
                     BisqEnvironment.getBaseCurrencyNetwork().isDaoBetaNet() ?
                             "144" :       // daoBetaNet; 1 day
-                            "300",      // testnet or dao testnet (server side regtest); 2 days
+                            "132",      // testnet or dao regtest (server side regtest); 2 days
```
sqrrm:
```suggestion
                            "132",      // testnet or dao regtest (server side regtest); 0.93 days
```

sqrrm (approved):
utACK

ManfredKarrer:
Sorry again for pumping more in that PR. But as it will change data, hashchain, genesis tx... I prefer to have that all in one merge.

core/src/main/java/bisq/core/dao/governance/voteresult/VoteResultService.java
```diff
@@ -358,29 +358,46 @@ private DecryptedBallotsWithMerits getDecryptedBallotsWithMerits(
...
-                        // received from the network?
-                        return new Ballot(ballot.getProposal(), vote);
+
+                        if (ballot.getVote() != null) {
+                            // If we had set a vote it was an own active vote
+                            if (!entry.getValue().isPresent()) {
+                                log.warn("We found a local vote but don't have that vote in the data from the " +
+                                        "blind vote. ballot={}", ballot);
+                            }
+                            if (ballot.getVote() != entry.getValue().get()) {
```
sqrrm: This needs to be an `else if`, `entry.getValue()` is checked in the previous `if`but code execution continues.

core/src/main/java/bisq/core/dao/governance/voteresult/VoteResultService.java
```diff
@@ -358,29 +358,46 @@ private DecryptedBallotsWithMerits getDecryptedBallotsWithMerits(
...
+                                log.warn("We found a local vote but the vote from the " +
+                                                "blind vote does not match. ballot={}, vote from blindVote data={}",
+                                        ballot, entry.getValue().get());
+                            }
+                        }
+
+                        // We only return accpeted or rejected votes
```
sqrrm:
```suggestion
                        // We only return accepted or rejected votes
```

core/src/main/java/bisq/core/dao/governance/voteresult/VoteResultService.java
```diff
@@ -396,6 +413,17 @@ private BallotList createBallotList(VoteWithProposalTxIdList voteWithProposalTxI
         if (!missingBallots.isEmpty())
             throw new VoteResultException.MissingBallotException(ballots, missingBallots);
 
+        // If we received a proposal after we had already voted we consider it as an proposla withhold attack and
```
sqrrm:
```suggestion
        // If we received a proposal after we had already voted we consider it as an proposal withhold attack and
```

core/src/main/java/bisq/core/dao/governance/voteresult/VoteResultService.java
ManfredKarrer: Yes will fix it.

sqrrm:
utACK

Took me a while to get the reject vote on proposals already seen locally but not included in a vote. It's clear that it does that, the comments are correct but it was still hard to follow. Most likely some detailed documentation of how the blindvote/payload/votereveal/voteresult process works is needed to be able to follow why all this happens and is needed.

core/src/main/java/bisq/core/dao/governance/proposal/ProposalService.java
```diff
@@ -219,12 +222,14 @@ private void publishToAppendOnlyDataStore() {
...
-                    else
-                        log.warn("publishToAppendOnlyDataStore failed for proposal " + proposalPayload.getProposal());
+                    UserThread.runAfterRandomDelay(() -> {
+                        boolean success = p2PService.addPersistableNetworkPayload(proposalPayload, true);
+                        if (success)
+                            log.info("We published a ProposalPayload to the P2P network as append-only data. proposalTxId={}",
+                                    proposalPayload.getProposal().getTxId());
+                        else
+                            log.warn("publishToAppendOnlyDataStore failed for proposal " + proposalPayload.getProposal());
+                    }, 100, 5000, TimeUnit.MILLISECONDS);
```
sqrrm:
```suggestion
                    }, 1, 1800, TimeUnit.SECONDS);
```

sqrrm (approved):
utACK

Perhaps a longer delay might be better, this is still quite a burst, why not spread it over half an hour? BREAK1 is 149 blocks long. Either way is fine though.

core/src/main/java/bisq/core/dao/governance/proposal/ProposalService.java
ManfredKarrer: We don't want to risk that the user shuts down the app in the meantime. The long we delay the higher the risk.

sqrrm: There are a lot of clients and they will all be sending the same data more or less so it's not dependent on one single instance. If you think that's not as safe I'm ok with the 5s interval as well.

ManfredKarrer: Yes, maybe we need a more sophisticated solution here... or just increase the delay?

core/src/main/java/bisq/core/dao/state/DaoStateService.java
```diff
@@ -811,6 +816,27 @@ public long getTotalAmountOfConfiscatedTxOutputs() {
...
     }
 
+    public long getBurnedBsqOfAllInvalidTxs() {
+        return getTxStream()
+                .filter(e -> e.getTxType() == TxType.INVALID)
+                .mapToLong(this::getBurnedBsqOfInvalidTx)
+                .sum();
+    }
+
+    public long getBurnedBsqOfInvalidTx(Tx tx) {
```
sqrrm: This is not specific to burned txs. A better name might be `getBsqInputValue(Tx tx)`

core/src/main/java/bisq/core/dao/monitoring/DaoStateMonitoringService.java
```diff
@@ -142,13 +148,13 @@ public void onDaoStateChanged(Block block) {
...
-        long totalConfiscatedAmount = daoStateService.getTotalAmountOfConfiscatedTxOutputs();
+        long totalInvalidAmount = daoStateService.getBurnedBsqOfAllInvalidTxs();
         // confiscated funds are still in the utxo set
         long sumUtxo = daoStateService.getUnspentTxOutputMap().values().stream().mapToLong(BaseTxOutput::getValue).sum();
-        long sumBsq = genesisTotalSupply + compensationIssuance + reimbursementIssuance - totalBurntFee;
+        long sumBsq = genesisTotalSupply + compensationIssuance + reimbursementIssuance - totalBurntFee - totalInvalidAmount;
 
         if (sumBsq != sumUtxo) {
-            throw new RuntimeException("There is a mismatch between the UTXO set and the DAO state. Please contact the Bisq devlopers.");
+            utxoMismatches.add(new UtxoMismatch(block.getHeight(), sumUtxo, sumBsq));
```
sqrrm, line 156: If we get a mismatch it seems that will stay mismatched in the future as well and we'll get a new `UtxoMismatch` added at every block after the first one. Maybe this is good but could be a bit annoying if there is a popup for every new block.

sqrrm (approved):
utACK

Looks ok but feel free to rename the function I commented on.

core/src/main/java/bisq/core/dao/state/DaoStateService.java
ManfredKarrer: Hm... I consider it burned BSQ as well in contrast to burnedFee.

core/src/main/java/bisq/core/dao/monitoring/DaoStateMonitoringService.java
ManfredKarrer: Yes. Need to improve that. Popups are handled in the UI to not repeat but still we don't want to add it at each block again. But anyway should not happen at all, is just to double check if we missed something. So I consider it more as a developer tool...

core/src/main/java/bisq/core/dao/state/DaoStateService.java
sqrrm: Yes, my point was that the function can calculate the input value from any tx, not only invalid ones.

ManfredKarrer: Ah ok. I am not sure if it is valid for normal txs as well as all special cases handled in the input parser are ignored here. It only checks the txoutput types. I think for the use case of invalid txs that should be good enough. A maybe more safe approach would be to add another field to tx like burnedFee for the burned BSQ from an invalid tx. Only at parse time we have all checks done and we could not do that later as well. Hm... not sure if its worth to add that field but might be more safe. In fact we need to add a comment to that method that this is not intended to be used otherwise as it is not a perfect check....

sqrrm: Yeah, good point. If we can add the field and set in the parser it would be more correct and safer I think.

ManfredKarrer: Implemented as discussed. We use the burntFee field in Tx but renamed it.

core/src/main/java/bisq/core/dao/governance/proposal/ProposalService.java
ManfredKarrer: Implemented as discussed. We set reBroadcast to false to avoid flooding. Other broadcast handling is less critical as user only publishes its own proposals/votes. But still some case where I added a longer random delay to avoid network stress.

core/src/main/java/bisq/core/dao/monitoring/DaoStateMonitoringService.java
ManfredKarrer: I leave that for now but will keep it on my todo list

core/src/main/java/bisq/core/dao/governance/merit/MeritConsensus.java
```diff
@@ -54,28 +54,27 @@ public static MeritList decryptMeritList(byte[] encryptedMeritList, SecretKey se
...
+                .filter(merit -> merit.getIssuance().getChainHeight() <= blindVoteTxHeight)
                 .mapToLong(merit -> {
                     try {
                         Issuance issuance = merit.getIssuance();
                         checkArgument(issuance.getIssuanceType() == IssuanceType.COMPENSATION,
                                 "issuance must be of type COMPENSATION");
                         return getWeightedMeritAmount(issuance.getAmount(),
                                 issuance.getChainHeight(),
-                                txChainHeight,
+                                blindVoteTxHeight,
```
sqrrm, line 77: This seems like a good change, more consistent calculations. It also counts the merit at the time of voting which is what the user would expect.

sqrrm (approved):
utACK

I think changing the name to burntBsq made it a lot easier to read and understand. Looks like a good solution.

common/src/main/java/bisq/common/util/ExtraDataMapValidator.java
```diff
@@ -0,0 +1,75 @@
...
+    // We don't expect many entries there.
+    public final static int MAX_SIZE = 10;
+    public final static int MAX_KEY_LENGTH = 100;
+    public final static int MAX_VALUE_LENGTH = 100000; // 100 kb
+
+    public static Map<String, String> getValidatedExtraDataMap(@Nullable Map<String, String> extraDataMap) {
+        return getValidatedExtraDataMap(extraDataMap, MAX_SIZE, MAX_KEY_LENGTH, MAX_VALUE_LENGTH);
+    }
+
+    public static Map<String, String> getValidatedExtraDataMap(@Nullable Map<String, String> extraDataMap, int maxSize, int maxKeyLength, int maxValueLength) {
```
sqrrm: Please try to break these long lines, it's harder to read when they're too long.

sqrrm (approved):
utACK

common/src/main/java/bisq/common/util/ExtraDataMapValidator.java
ManfredKarrer: Done

ManfredKarrer:
Last commits have been rather trivial, but you are welcome to review them post merge as well.

ManfredKarrer:
Build error is prob caused due cache issue. Local build works and witness has not changed. Pushed anotehr change to trigger re-build.

sqrrm:
Latest changes also look good.

PR #2590, opened 2019-03-26
Show close button when creating offer without required account

devinbileck:
Issue: When attempting to create an offer without a trading account for
the selected currency, a prompt is shown asking the user whether to still
create an offer or to create a new account. But this prompt is missing a
cancel/close button in case you don't want to do either action. And
pressing Esc on the keyboard proceeds to create the account.
image

Cause: The close button on the prompt was renamed and being used to
create a new account.

Fix: Use the close button as a close button and add a secondary action
button to create a new account. The popup width was increased to
accommodate the long button text without being truncated.
image


ManfredKarrer (approved):
utACK

PR #2591, opened 2019-03-27
Disable "add peers from addr message" custom bitcoinj feature when a …

oscarguindzberg:
…list of peers is provided.


ManfredKarrer (approved):
utACK

PR #2592, opened 2019-03-27
Prevent BSQ offers below dust limit

### LINKED ISSUE #2343: Min. offer amount for BSQ must be 5.46 BSQ
When creating an offer we must ensure that the min. BSQ amount is larger than the dust limit of 546 satoshi (5.46 BSQ).

ripcurlx:
Fixes #2343.

Not 100% perfect as I show the error message below the volume input as we don't display a minVolume field right now. But as this is an edge case I think it is good enough.


ManfredKarrer (approved):
utACK

PR #2595, opened 2019-03-28
DAO governance and bond improvements

### LINKED ISSUE #2593: Small UI improvements
- In the votes tab we could add the blind vote tx id and move the vote column to the right.
Screen Shot 2019-03-27 at 12 46 32

- In the vote results once a detail popup at a proposal has been opened and closed the user cannot open it up again as it only opens on the select event. Maybe to add a columns with a details button might make it also more clear that there is a popup available?

- Maybe we should add the date/time to the popup as well.

- At confiscation requests we link the tx id to the confiscated bond. Atm the screen switches in the background but the popup is not closed.  A link symbol would be good and the text for boned reputation should mention that the displayed data is the reputation hash.

- If its bonded role confiscation we show role + name. Tx ID would be good as well to be sure its unique. Now there are 2 times a colon. Would be good to avoid that.
If we use tx id we could use that in boned reputation instead of hash as well. A shortened ID might be enough.
Screen Shot 2019-03-27 at 12 51 03

ripcurlx:
Addresses and fixes #2593.

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -198,7 +198,7 @@ shared.actions=Actions
 shared.buyerUpperCase=Buyer
 shared.sellerUpperCase=Seller
 shared.new=NEW
-shared.new=NEW
+shared.blindVoteTxId=Blind vote transaction ID
 
 ####################################################################
 # UI views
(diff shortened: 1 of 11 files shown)


ManfredKarrer (commented):
utACK

PR #2598, opened 2019-03-29
Show fiat rounding info popup also when switching between payment accounts

### LINKED ISSUE #2597: Fiat rounding info popup not shown when switching between payment accounts
**Expected behavior:**
Fiat rounding info popup is always shown when a fiat currency is selected.

**Actual behavior:**
Fiat rounding info popup is only shown when create offer view is activated and a fiat currency is selected.

ripcurlx:
Fixes #2597.


ManfredKarrer (approved):
utACK

Issue #2604, opened 2019-03-30
Protection against dust attack utxos


ManfredKarrer:
There is some discussion regaring some dust attacks where chainanalysis companies send out a lot of dust outputs to addresses with the intention to learn about other addresses of those address owners (coin merge reveal ownership).
See: https://twitter.com/manfred_karrer/status/1111435913744117760
https://twitter.com/manfred_karrer/status/1111707029327015938

One solution to that attack is to isolate those dust outputs. But they still would pollute utxo set and mempool. Better would be to spend them in an aggregated tx in a way that no coin merge happens.

As for paying the miner fee other utxos must be added it would only work in a tx which uses the dust utxos of other users. In Bisq we could send out those signed inputs (signed with sighash ANYONECANPAY) to a service which aggregate the utxos into a tx to some donation address once there are sufficient inputs to be efficient with the miner fee.

The dontation should be a project which is an active counterforce against those surveillance capitalistic companies. The Tor project would be a natural fit.....

The aggegator should run as Tor onion service with a simple HTTP API so that anyone beside Bisq users can use it. The aggregated funds should be small enough to not create incentives that the aggregator steals the funds. There should be some random delays to avoid analysis attempts to map all those users to the same wallet software / platform.

It is important that this service would be used by several wallets. Wasabi and Samurai Wallet would be natural fits, but other like Greenaddress or Electruum should be invited as well.
Only if there are sufficiently broad distribution of that feature the spies cannot assume that all those who gave their dust utxos are Bisq users.

Here is a rough idea how it could be implemented:
- Add radio buttons to preferences where the user can select of he wants to isolate the dust outputs or if he prefers to send it to a aggregator service. Third option would be to use it as normal outputs but we should display a warning popup about the privacy risks if the user choose that option.
- If user has activated the 'dust to aggregator service option' he signs the dust utxo with the ANYONECANPAY sighash and sends the signed input to a dedicated onion service (operated by a Bisq developer). The receiver address need to be defined at that point. The aggregator service will provide that.
- The aggregator service checks the current miner fee and the aggregated input values when receiving new dust utxos and if the balance hits a defined efficienty threshold it creates and broadcasts the tx to the donation address.

Anyone up for imlementing it?

ManfredKarrer:
[...]

PR #2621, opened 2019-04-03
Ignore dust utxo

ManfredKarrer:
Implements https://github.com/bisq-network/bisq/issues/2604

I think we should keep that for after the DAO launch. Not really high risk but prob. not that important to add any risk for the DAO release....


oscarguindzberg:
utACK

PR #2622, opened 2019-04-03
Fix max available stake for voting showing negative

devinbileck:
When attempting to vote on proposals with insufficient BSQ,
the max available stake for voting was showing a negative value.

image


ManfredKarrer (approved):
utACK

PR #2630, opened 2019-04-03
Fix incorrect required BSQ amount in warning text

devinbileck:
When submitting a proposal with insufficient BSQ balance, the warning
message showed an incorrect required BSQ amount. The amount needed to be
divided by 100 (1 BSQ = 100 satoshis).

image


ManfredKarrer (approved):
utACK

PR #2643, opened 2019-04-04
Update insufficient BTC warning for issuance proposals

devinbileck:
When attempting to submit a compensation request or reimbursement
request with insufficient BTC, the warning message only indicated that
BTC is required for mining fees. It did not indicate that BTC is also
required for the BSQ issuance.
image

So another display string was added and displayed when submitting an
issuance proposal with insufficient BTC funds available.
image


ManfredKarrer (commented):
utACK

PR #2652, opened 2019-04-05
Fix incorrect transaction size shown for bonding

devinbileck:
When locking up or unlocking a bond, the displayed transaction size was
in bytes, but the prompt was indicating it was Kb. So adjust the value
so it is shown as Kb.

Before:
image

After:
image


ripcurlx (approved):
utACK - Good catch 👍! I quickly checked the other places where we display Kb and it looks like it is correctly calculated everywhere else.

PR #2655, opened 2019-04-05
Improve tx display for BTC withdrawal from BSQ wallet

### LINKED ISSUE #2637: Transaction information when withdrawing non-BSQ funds from BSQ wallet not handled properly
After unsuccessful issuance proposals (e.g. compensation requests) you have non-colored BTC available in your BSQ Wallet. When sending them to your local Bisq BTC wallet following transactions are displayed.

*In BSQ wallet: Information shouldn't be not recognized*
Bildschirmfoto 2019-04-04 um 12 59 16

*In BTC wallet: Address shouldn't be empty*
Bildschirmfoto 2019-04-04 um 12 58 30

ripcurlx:
Fixes #2637.

Bildschirmfoto 2019-04-05 um 16 30 40
Bildschirmfoto 2019-04-05 um 16 30 20

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
+++ b/desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
@@ -297,7 +297,7 @@ public class BsqSendView extends ActivatableView<GridPane, Void> implements BsqB
                     Transaction signedTx = bsqWalletService.signTx(txWithBtcFee);
                     Coin miningFee = signedTx.getFee();
 
-                    if (miningFee.getValue() > receiverAmount.getValue())
+                    if (miningFee.getValue() >= receiverAmount.getValue())
                         GUIUtil.showWantToBurnBTCPopup(miningFee, receiverAmount, btcFormatter);
                     else {
                         int txSize = signedTx.bitcoinSerialize().length;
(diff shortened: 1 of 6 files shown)


ManfredKarrer:
Maybe the N/A could be changed here to 0?

ripcurlx:
> Maybe the N/A could be changed here to 0?

Yes, would make sense. I'll update that.

ManfredKarrer (approved):
utACK

desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
```diff
@@ -296,18 +296,25 @@ private void addSendBtcGroup() {
...
-                            miningFee,
-                            txSize, receiversBtcAddressInputTextField.getText(),
-                            btcFormatter,
-                            btcFormatter,
-                            () -> {
-                                receiversBtcAddressInputTextField.setText("");
-                                btcAmountInputTextField.setText("");
-                            });
+
+                    if (miningFee.getValue() > receiverAmount.getValue())
```
ManfredKarrer: I think it should be >=

ManfredKarrer: I think a check that the output is no below dust limit is in the wallet service, but maybe better to cross check?

ManfredKarrer (approved):
See comment for small improvemet and double check dust limit, beside that utACK

desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
ripcurlx: Yes, we probably should already prevent it, if the amount will be ```0``` in the end. Maybe we should also prevent it already before, but there it is hard where to draw a line.

ripcurlx: > I think a check that the output is no below dust limit is in the wallet service, but maybe better to cross check?

There is a dust check in place already.

PR #2657, opened 2019-04-05
Reduce tx broadcast timeout

ManfredKarrer:
Currently there is a bug in BitcoinJ causing the timeout at all BSQ
transactions.
It is because BitcoinJ does not handle confidence object correctly in
case as tx got altered after the
Wallet.complete() method is called which is the case for all BSQ txs.
We will work on a fix for that but that
will take more time. In the meantime we reduce the timeout to 5
seconds to avoid that the trade protocol runs
into a timeout when using BSQ for trade fee.

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/TxBroadcaster.java
+++ b/core/src/main/java/bisq/core/btc/wallet/TxBroadcaster.java
@@ -72,7 +72,12 @@ public class TxBroadcaster {
         void onFailure(TxBroadcastException exception);
     }
 
-    private static final int DEFAULT_BROADCAST_TIMEOUT = 30;
+    // Currently there is a bug in BitcoinJ causing the timeout at all BSQ transactions.
+    // It is because BitcoinJ does not handle confidence object correctly in case as tx got altered after the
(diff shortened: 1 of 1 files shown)


ManfredKarrer:
This follows after a discussion with @oscarguindzberg about his finding of the tx broadcast bug.
It will require more time to get the proper bugfix for the underlying problem (tx confidence handling in btcj). The issues only happens at BSQ txs as there the complete methods is called before the final tx is created. We discussed different intermediate fixes but decided to not add any risk and just reduce the timeout so that it does not cause issues on the trade protocol as well to not force a too long delay at proposal or blind vote publishings.
This happen only if you are connected to more then 1 btc peer, not on local regtest or dao_regtest....

The tx broadcast itself has no bug, it is just that we don't hear back rom peers because the confidence object has changed and the success handler depends on that. So changing to a rather short time of 5 seconds is justified. The trade protocol has 1 minute timeout and to lose with the trade fee tx publishing too much time would risk that the trade protocol gets into a timeout error. This would be only the case when BSQ is used as trade fee.

sqrrm (approved):
utACK

Issue #2660, opened 2019-04-06
Writing json file to disc on Windows fails


BrettCawley:
On a windows machine, launch bisq using the "--dumpStatistics=true" argument.

Observe many temp files in "Bisq\btc_mainnet\db" are continually created.

Issue is due to a failure when renaming the temp file to the correct file name.

ManfredKarrer:
Can you post the log file?

BrettCawley:
Log attached.
My assumption is it's not possible to rename the file when it's still in use
[bisq.log](https://github.com/bisq-network/bisq/files/3050804/bisq.log)

ManfredKarrer:
I think its a windows only issue. There are some temp file handling issues. As that feature is used for exploreres only and they run on linux I did not care much, but yes should be fixed...

PR #2661, opened 2019-04-06
Fix issue writing json file to disk

BrettCawley:


BrettCawley:
Fixes #2660

common/src/main/java/bisq/common/storage/JsonFileManager.java
```diff
@@ -78,6 +78,7 @@ public void writeToDisc(String json, String fileName) {
 
                 printWriter = new PrintWriter(tempFile);
                 printWriter.println(json);
+                printWriter.close();
```
ManfredKarrer, line 81: It is in the finally statement:
```
if (printWriter != null)
                    printWriter.close();
```

ManfredKarrer (commented):
NACK

ManfredKarrer:
I checked at another class and added some fix there which seems to have been needed on Windows. 
At the end its the same as you suggested but it comes with a comment why as the close is in the finally statment anyway. 
See https://github.com/bisq-network/bisq/pull/2669

Issue #2659, opened 2019-04-06
Test asset fee tx


ManfredKarrer:
The asset listing fee tx does not use a BTC input in case the BSQ fee is >= miner fee.

A valid tx can look like (tested):
Input 0: 222216.00 BSQ (22221600 Sat)
Output 0: 122217.00 BSQ (12221700 Sat)
Output 1: 0,09961716 BTC (9961716 Sat)
Miner fee: 0,00038184 (38184 Sat)
 
We should test if in case the input is exact the amount of the required miner fee we cannot create the tx as a tx without outputs is not valid. We need to have at least 1 output with 546 sat. This output can be either BSQ or BTC.

sqrrm:
Should be closed, see #2664

PR #2664, opened 2019-04-06
Handle #2659, no output for asset fee tx

sqrrm:
Avoid creating tx with no outputs apart from opreturn.
Use completePreparedBsqTx instead of completePreparedProposalTx for
asset fee txs as it's more appropriate and also handles the no
output issue.

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
@@ -150,7 +150,7 @@ public class BtcWalletService extends WalletService {
 
     public Transaction completePreparedBurnBsqTx(Transaction preparedBurnFeeTx, byte[] opReturnData)
             throws WalletException, InsufficientMoneyException, TransactionVerificationException {
-        return completePreparedProposalTx(preparedBurnFeeTx, opReturnData, null, null);
+        return completePreparedBsqTx(preparedBurnFeeTx, false, opReturnData);
     }
 
     ///////////////////////////////////////////////////////////////////////////////////////////


core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
```diff
@@ -150,7 +150,7 @@ String getWalletAsString(boolean includePrivKeys) {
 
     public Transaction completePreparedBurnBsqTx(Transaction preparedBurnFeeTx, byte[] opReturnData)
             throws WalletException, InsufficientMoneyException, TransactionVerificationException {
-        return completePreparedProposalTx(preparedBurnFeeTx, opReturnData, null, null);
+        return completePreparedBsqTx(preparedBurnFeeTx, false, opReturnData);
```
ManfredKarrer, line 153: That method is also called from BaseProposalFactory for proposals. 
I need more time to check if completePreparedBurnBsqTx could be used for proposals.

ManfredKarrer: Not sure if that fix is worth the effort and risk atm, it is a very unlikely event that the input matches exactly the miner fee. And I assume it would fail anyway on the BitcoinJ side at least for the case with no outputs. Maybe we postpone that for a later time as it is not relevant for consensus anyway?

sqrrm: Indeed, it's almost impossible to make the burn fee match mining fee and completely spend the remaining BSQ in an output. I had to write some code just to make that happen.

Since there is an output (the opReturn) I think the tx might even be standard. I thought that wasn't the case at first but reading up I can't find anything saying otherwise. Best not merge this PR then.

sqrrm: About your comments, the difference between `completePreparedProposalTx` and `completePreparedBsqTx` is the reimbursement option so they should work the same for non reimbursement proposals except that `completePreparedBsqTx` also handles the no output case. I think we have a bit of a problem with too much copy pasted code here.

ManfredKarrer: Yes I agree, the code need to be cleaned up and refactored, I am just too scary to do that now. Needs much care and testing. I prefer to focus on issues which are consensus relevant as there we cannot change later anymore (or it becomes much harder).

ManfredKarrer: Maybe lets close it for now.

sqrrm:
Decided to not go this way

Issue #2662, opened 2019-04-06
Hash for Role is not using cyrptographic hash


ManfredKarrer:
The hash which is used at Role is useing hashCode instead of a cryptographic hash.
hashCode is not guaranteed to be determinisitic across JVM implementations and versions so we must not use that in any consensus relevant use case. The hash for Role seems to have not much usage atm but will likely get more at least for arbitrators, mediators and maybe seed node operators. So better to fix that before the mainnet launch...

ManfredKarrer:
The consequences with the change with the current data and network is that a because of the hash change a bonded role is not found anymore and interpreted as bonded reputation.
If a bonded role was locked up it cannot be revoked in the UI. So when we add this change there can be smaller confusions but it should not break consensus.

PR #2665, opened 2019-04-07
Use cryptographic hash from protobuffer serialized data instead of hashCode

ManfredKarrer:
Fixes https://github.com/bisq-network/bisq/issues/2662

Let's wait with merge until we are sure when we want to deploy it.


sqrrm:
I think it's worth getting this right. The change itself is simple and those that want to can upgrade testnet to get the monitoring right. Slightly annoying vs keeping this into live launch.

utACK

ripcurlx (approved):
utACK

PR #2667, opened 2019-04-07
Protect against late blind vote publishing

ManfredKarrer:
We want to wait with merge until we know how we will deploy it.

### DIFF
--- a/core/src/main/java/bisq/core/dao/governance/blindvote/BlindVoteListService.java
+++ b/core/src/main/java/bisq/core/dao/governance/blindvote/BlindVoteListService.java
@@ -102,7 +102,7 @@ public class BlindVoteListService implements AppendOnlyDataStoreListener, DaoSta
     @Override
     public void onNewBlockHeight(int blockHeight) {
         // We only add blindVotes to blindVoteStorageService if we are not in the vote reveal phase.
-        blindVoteStorageService.setInCorrectPhase(isInCorrectPhase(blockHeight));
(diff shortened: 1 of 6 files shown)


core/src/main/java/bisq/core/dao/governance/blindvote/BlindVoteListService.java
```diff
@@ -142,18 +157,30 @@ private void onAppendOnlyDataAdded(PersistableNetworkPayload persistableNetworkP
...
                     }
-                    blindVotePayloads.add(blindVotePayload);
                 } else {
                     log.warn("We received an invalid blindVotePayload. blindVoteTxId={}", txId);
                 }
             }
         }
     }
+
+    private boolean isInCorrectPhase(int blockHeight) {
```
sqrrm: It would be better to just call this `notInRevealPhase` or `isAcceptingProposals` to better describe what it does.

sqrrm (approved):
utACK

This might be a better way than using the date, but I don't have a really good model in my head so I'm not saying this with conviction.

core/src/main/java/bisq/core/dao/governance/blindvote/BlindVoteListService.java
ManfredKarrer: Ok, will rename (had it that way before ;-))

ripcurlx (approved):
utACK

PR #2674, opened 2019-04-08
Handle dao state conflicts better

ManfredKarrer:
- Check if conflicting node is seed node or not. Show info in list entry
and only show resync button and red color in text if conflict is with
seed node.
- At resync move all p2p network data to a dedicated directory so they
will be reloaded from the seeds.

### DIFF
--- a/common/src/main/java/bisq/common/storage/FileManager.java
+++ b/common/src/main/java/bisq/common/storage/FileManager.java
@@ -162,14 +162,19 @@ public class FileManager<T extends PersistableEnvelope> {
         }
     }
 
-    public synchronized void removeAndBackupFile(String fileName) throws IOException {
-        File corruptedBackupDir = new File(Paths.get(dir.getAbsolutePath(), "backup_of_corrupted_data").toString());
+    public static void removeAndBackupFile(File dbDir, File storageFile, String fileName, String backupFolderName)
+            throws IOException {
+        File corruptedBackupDir = new File(Paths.get(dbDir.getAbsolutePath(), backupFolderName).toString());
         if (!corruptedBackupDir.exists())
             if (!corruptedBackupDir.mkdir())
                 log.warn("make dir failed");
 
-        File corruptedFile = new File(Paths.get(dir.getAbsolutePath(), "backup_of_corrupted_data", fileName).toString());
+        File corruptedFile = new File(Paths.get(dbDir.getAbsolutePath(), backupFolderName, fileName).toString());
         FileUtil.renameFile(storageFile, corruptedFile);
+    }
+
+    public synchronized void removeAndBackupFile(String fileName) throws IOException {
+        removeAndBackupFile(storageFile, dir, fileName, "backup_of_corrupted_data");
     }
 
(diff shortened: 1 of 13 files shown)


ManfredKarrer:
@sqrrm @ripcurlx 
I am not sure if we shoud merge that into 1.0.0 or keep it for later. It is not high risk but also not trivial to test. I tested quite a bit though... It should not introduce much risk as well.
Lets discuss if we should add it or not. Would at least reduce confusion if some peers are out of sync as we have seen in dao-regtest last days... Also the missing file deletion of proposals, blind votes and ballots would not have lead to a recovery in current release if the cause for the issue is that one of those db files is out of sync.

common/src/main/java/bisq/common/storage/FileManager.java
```diff
@@ -162,16 +162,21 @@ void shutDown() {
...
             if (!corruptedBackupDir.mkdir())
                 log.warn("make dir failed");
 
-        File corruptedFile = new File(Paths.get(dir.getAbsolutePath(), "backup_of_corrupted_data", fileName).toString());
+        File corruptedFile = new File(Paths.get(dbDir.getAbsolutePath(), backupFolderName, fileName).toString());
         FileUtil.renameFile(storageFile, corruptedFile);
     }
 
+    public synchronized void removeAndBackupFile(String fileName) throws IOException {
+        removeAndBackupFile(storageFile, dir, fileName, "backup_of_corrupted_data");
```
sqrrm: The argument order is `dbDir`, `storageFile`..., this looks inverted.

desktop/src/main/java/bisq/desktop/main/dao/monitor/StateMonitorView.java
```diff
@@ -102,11 +115,15 @@
     public StateMonitorView(DaoStateService daoStateService,
                             DaoFacade daoFacade,
                             CycleService cycleService,
-                            PeriodService periodService) {
+                            PeriodService periodService,
+                            SeedNodeRepository seedNodeRepository,
+                            File storageDir) {
```
sqrrm: Should this be `@Named(Storage.STORAGE_DIR) File storageDir`?

sqrrm (commented):
I think it's worth adding this before 1.0 as it would avoid a lot of extra support issues from conflicted non seed nodes. It's also likely that users will be able to recover through the UI with this resync from seed feature.

No ACK until my comments are taken care of though.

desktop/src/main/java/bisq/desktop/main/dao/monitor/StateMonitorView.java
ManfredKarrer: No its the base class, I will remove the Inject annotation and make it protected.

common/src/main/java/bisq/common/storage/FileManager.java
ManfredKarrer: Good finding! Yes that is wrong. Thanks!

sqrrm:
utACK

This version I think is worth merging before 1.0, it's not that complex and much time will be saved by not having users worry about client not in synch.

PR #2675, opened 2019-04-09
List ParsiCoin (PARS)

m4hdi1995:
Official project URL: https://parsicoin.net
Official block explorer URL: http://explorer.parsicoin.net


blabno (approved):
ACK

Issue #2189, opened 2019-01-02
In the MARKET/SPREAD tab, some listed cryptos don't have a description field


ghost:
v0.9.1/Debian
In the MARKET/SPREAD tab, some listed cryptos don't have a description field
eg N/A (BCH) , N/A (STEEM) , N/A (YTN)
Probably delisted cryptos.

If it's not too complicated (or dangerous), feel free to assign this issue to me.

ripcurlx:
@HarryMacfinned The classes for this assets were removed from the repository, which causes the ```N/A``` in the Spread view. As this will be a non issue as soon as all these single loitering offers are gone, I'll close this issue.

ghost:
@ripcurlx , thanks for answering,
However I wonder how those concerned offers will go by themselves ? Do you expect the makers to delete them ?
I have seen many cases with old offers lasting for months.
There are 5 such offers now in v0.9.2.

ripcurlx:
@HarryMacfinned Feel free to add removed assets over time if you see new "N/A" appear. https://github.com/bisq-network/bisq/blob/9927f3a3c76d67ab309ab679dd180dcbb7b15a0f/core/src/main/java/bisq/core/locale/CurrencyUtil.java#L148-L155 I didn't want to add and check for all assets that have been removed since the beginning, so I just added the ones where a open offer was available at the time of implementation.

ghost:
Thanks for the precisions @ripcurlx 
I'll have a look at this java piece of code.

PR #2680, opened 2019-04-09
Update CurrencyUtil.java

ghost:
add BURST+STEEM to the list of removed assets.
Should solve https://github.com/bisq-network/bisq/issues/2189
I built the code and it seems to work.


ripcurlx (approved):
utACK

PR #2687, opened 2019-04-10
Fix ticker to Starwels (USDH)

ghost:
Official project URL:
https://0net.io/139H5wy8H3tgXwJ33mXpwmm4oCfjKUo7dY/
Official block explorer URL: http://91.240.86.126:3001/


blabno (approved):
ACK

blabno:
@ManfredKarrer the change is only about symbol change. I'm not sure how it could affect existing offers.

ripcurlx:
> @ManfredKarrer the change is only about symbol change. I'm not sure how it could affect existing offers.

I don't see any open offers nor trades for this coin so I don't think there should be an issue with this change.

PR #2699, opened 2019-04-11
Improve space management in button bar of overlays

### LINKED ISSUE #2681: Adjust margin on "Don't show again" alert for translated strings
Unfortunately due to the nature of translations, strings can vary in length from language to language. I have my adjusted some translations for harmony with the UI where possible, but some places just can't be helped.
Alerting about the security deposit function
alert
Alerting that I've accepted an offer
offeraccpted

ripcurlx:
Fixes #2681.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/overlays/windows/DaoLaunchWindow.java
+++ b/desktop/src/main/java/bisq/desktop/main/overlays/windows/DaoLaunchWindow.java
@@ -127,8 +127,7 @@ public class DaoLaunchWindow extends Overlay<DaoLaunchWindow> {
     protected void addButtons() {
         super.addButtons();
 
-        closeButton.setPrefWidth(162);
-        actionButton.setPrefWidth(162);
+        closeButton.prefWidthProperty().bind(actionButton.widthProperty());
     }
 
     ///////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


ManfredKarrer (approved):
utACK

desktop/src/main/java/bisq/desktop/main/overlays/windows/DaoLaunchWindow.java
```diff
@@ -127,8 +127,7 @@ protected void onShow() {
     protected void addButtons() {
         super.addButtons();
 
-        closeButton.setPrefWidth(162);
-        actionButton.setPrefWidth(162);
+        closeButton.prefWidthProperty().bind(actionButton.widthProperty());
```
ManfredKarrer, line 130: Are you sure the setters are never called in any other popup? It would cause an error if binding is set and we call setWidth somewhere else....

Ah its not in the super class, so should be isolated then to that window.

utACK

ripcurlx: Yes, it was just to have the same button width based on Pedro's design for this popup.

ripcurlx: But still to support bigger button sizes for different languages

PR #2700, opened 2019-04-11
Add missing check to prevent re-selection of proposal

ripcurlx:
To be able to re-select a proposal that was clicked before, without selecting a different proposal.


ManfredKarrer (approved):
utACK

PR #2701, opened 2019-04-11
Manage startup popups

ManfredKarrer:


ManfredKarrer:
If localhost BTC node is running we would have seen that popup behind the DAO launch popup. This PR fixes that and introduce a generic solution to handle multiple popups by priority.

ripcurlx (approved):
utACK

PR #2703, opened 2019-04-12
Add slide out animation

ripcurlx:
...and delay initial slide in animation to increase the chance that it is seen (dependent on available processing power on startup)


ManfredKarrer (approved):
utACK

PR #2705, opened 2019-04-12
Add specific URL validation for proposal links

### LINKED ISSUE #2473: Ensure proposals are entered using correct URL
URL redirects for proposals, compensation requests, and roles have been implemented by #2429. We now need to consider what approach (if any) to take with regards to entering proposal links within the UI to ensure people are using the correct URL.

If users don't use our redirected URLs, GitHub will be forever recorded in compensation requests and could potentially become dead links if we later relocate them elsewhere. This would be an issue if someone wants to view past proposals and attempts to follow the link to the proposal.

Likely the simplest approach is to only allow the user to enter the proposal id.

devinbileck:
Since all proposals should be created in GitHub, and we have URL
redirects in place in case we switch platforms in the future, enforce
proposal links to utilize the specific URL redirect.

Resolves #2473


devinbileck:
I assume we don't need to be lenient for the link, and it must be submitted using that specific URL?
image

ManfredKarrer (commented):
Are the redirects in place? https://bisq.network/dao-proposals gets a 404.
Once redirects are set utACK.

devinbileck:
The redirect should work now that https://github.com/bisq-network/bisq-website/pull/164 was merged.

PR #2710, opened 2019-04-14
Add Docker configuration

BusyDigits:
Running Bisq under Docker is appealing because it (probably?) isolates the host from any malicious behavior that may have somehow gotten into Bisq. This pull adds Docker configuration and instructions. It's only been tested using a Linux system running X11.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,14 @@
+FROM openjdk:10-jdk
+
+# This docker image should be used with
+# `docker run ... -v /path/to/host/dir:/root/.local/share/Bisq` in order to
+# persist Bisq configuration to the host.
+
+RUN apt-get update \
+  && apt-get install -y x11-apps \
+  && rm -rf /var/lib/apt/lists/*
+
+RUN mkdir /src && git clone https://github.com/bisq-network/bisq /src/bisq
+WORKDIR /src/bisq
+RUN ./gradlew build
+VOLUME /home/root/.local/share/Bisq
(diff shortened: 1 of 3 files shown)


ripcurlx:
Maybe @devinbileck and/or @blabno could you have a look at this?

devinbileck:
I leave it to @blabno to review as I am not very familiar with docker.

blabno:
There is already a docker related stuff here: https://github.com/bisq-network/docker

docker/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+# persist Bisq configuration to the host.
+
+RUN apt-get update \
+  && apt-get install -y x11-apps \
+  && rm -rf /var/lib/apt/lists/*
+
+RUN mkdir /src && git clone https://github.com/bisq-network/bisq /src/bisq
+WORKDIR /src/bisq
+RUN ./gradlew build
+VOLUME /home/root/.local/share/Bisq
```
blabno, line 14: add `CMD ./bisq-desktop`
so that user could invoke `run` without additional command param

docker/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+Running Bisq using Docker has currently only been tested using Linux.
+
+## Build the image
+
+```sh
+docker build -t bisq .
+```
+
+## Running Bisq
```
blabno, line 11: It would be nice to have docker-compose configured with several containers like 
- prod, which would have default volume bound to `$HOME/.local/share/Bisq`
- dev, without data dir volume bound

That way user would only need to call `docker-compose up prod` or `docker-compose up dev`

blabno:
Also, I think that the docker directory should be moved to `desktop` directory

PR #2722, opened 2019-04-16
Add missing pricenode project generated shell script

devinbileck:
When generating shell scripts in the root project directory, the
pricenode project was missing.


ripcurlx (approved):
ACK

PR #2723, opened 2019-04-16
Fix wrong calculation of percentage based security deposit

### LINKED ISSUE #2713: can not edit some open offers in v 1.0.0
I can not edit any offer made from previous versions so i deleted all offers and recreated new ones.  2 of them i can edit (for amounts of 1 and 2 BTC) but can not edit 2 others (0.1, 0.25 BTC) as the confirm button is disabled for them.  Maybe it only likes integer trades?

ripcurlx:
Only affects editing of an existing offer and was introduced in v0.9.5.
It wasn't detected until now as we increased the minimum security deposit with v1.0.0.

Fixes #2713.


ManfredKarrer (approved):
utACK

PR #2725, opened 2019-04-16
Only show rounding popup once during activation

### LINKED ISSUE #2708: Fiat rounding privacy popup pops up twice when creating a Fiat offer
This also happens when editing a fixed price offer - in that case it pops up 5 times.

ripcurlx:
...and not show it in edit offer screen. Fixes #2708.


ManfredKarrer (approved):
utACK

PR #2728, opened 2019-04-16
Doichain: set p2shHeader.

oscarguindzberg:
See https://github.com/Doichain/core/blob/master/src/chainparams.cpp#L181

Not setting the p2shHeader makes the p2shHeader to be 0, that has problems on address validation on bitcoinj 0.15


ManfredKarrer (commented):
utACK

PR #2730, opened 2019-04-17
Request blocks in case we have not received it

ManfredKarrer:
Once we get a new block from BitcoinJ we start a timer and check in 10
seconds if we have received the block from the P2P network. If not we
request it from the seed nodes.

### DIFF
--- a/core/src/main/java/bisq/core/dao/node/lite/LiteNode.java
+++ b/core/src/main/java/bisq/core/dao/node/lite/LiteNode.java
@@ -103,11 +103,12 @@ public class LiteNode extends BsqNode {
             // We expect to receive the new BSQ block from the network shortly after BitcoinJ has been aware of it.
             // If we don't receive it we request it manually from seed nodes
             checkForBlockReceivedTimer = UserThread.runAfter(() -> {
-                if (daoStateService.getChainHeight() < height) {
+                int chainHeight = daoStateService.getChainHeight();
+                if (chainHeight < height) {
                     log.warn("We did not receive a block from the network {} seconds after we saw the new block in BicoinJ. " +
                                     "We request from our seed nodes missing blocks from block height {}.",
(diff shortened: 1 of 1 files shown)


core/src/main/java/bisq/core/dao/node/lite/LiteNode.java
```diff
@@ -78,6 +86,31 @@ public void start() {
...
+            }
+
+            // We expect to receive the new BSQ block from the network shortly after BitcoinJ has been aware of it.
+            // If we don't receive it we request it manually from seed nodes
+            checkForBlockReceivedTimer = UserThread.runAfter(() -> {
+                if (daoStateService.getChainHeight() < height) {
+                    log.warn("We did not receive a block from the network {} seconds after we saw the new block in BicoinJ. " +
+                                    "We request from our seed nodes missing blocks from block height {}.",
+                            CHECK_FOR_BLOCK_RECEIVED_DELAY_SEC, daoStateService.getChainHeight());
+                    liteNodeNetworkService.requestBlocks(daoStateService.getChainHeight());
```
sqrrm: This means we're asking for block `daoStateService.getChainHeight()` and up, but we already have the block at `daoStateService.getChainHeight()`, wouldn't it be better to ask for `daoStateService.getChainHeight() + 1`?

It's also only asking seednodes, any full node should be fine to ask and then we can use the block hash to verify we have the right blocks, and if not we block that ban that full node and ask the next. That's not really relevant to this PR though.

sqrrm (changes requested):
I think it would be better to ask for the first block we don't have rather than the last block we have.

core/src/main/java/bisq/core/dao/node/lite/LiteNode.java
ManfredKarrer: I think you are right, thanks! Will look into it again after I finish current new PR...

ManfredKarrer: Fixed

PR #2732, opened 2019-04-17
Add nodeaddress and capabilities at getblocks request

ManfredKarrer:
When we request at startup the blocks from a seed which has not yet
received out capabilities or node address we will not get a response.
We do not want to depend on the state in the getData domain in the
p2p network by using only a seed which has already responded but we
prefer to make sure the seednode will get all the data by our request
to be able to respond.


sqrrm (approved):
utACK

PR #2736, opened 2019-04-17
Fix issues with DAO full mode in preferences

ManfredKarrer:
There was missing the block notify port as well as the options have
not been applied correctly which led to the situation that when you
run the same application in full mode and later in lite mode that
data got persisted.

Added also to the footer a note if full mode is running.

Screen Shot 2019-04-17 at 18 26 50


ripcurlx (approved):
ACK

ripcurlx:
Only conflicts need to be resolved

PR #2737, opened 2019-04-18
Add UTF8 for getBytes calls

### LINKED ISSUE #2729: Update prompt not shown on Windows
There are multiple reports of Windows 10 users not receiving the prompt to update to v1.0. Whereas Linux and macOS users are receiving the prompt.

ManfredKarrer:
Fixes https://github.com/bisq-network/bisq/issues/2729

### DIFF
--- a/common/src/main/java/bisq/common/crypto/PGP.java
+++ b/common/src/main/java/bisq/common/crypto/PGP.java
@@ -17,6 +17,8 @@
 
 package bisq.common.crypto;
 
+import com.google.common.base.Charsets;
+
 import org.bouncycastle.bcpg.BCPGKey;
 import org.bouncycastle.bcpg.RSAPublicBCPGKey;
 import org.bouncycastle.openpgp.PGPException;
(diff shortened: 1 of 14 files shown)


ManfredKarrer:
@devinbileck Would be great if you could test that on Windows with whole smoke test (trade) as I cahnged all ocurrances of getByte(). I tested on regtest and all worked.

ripcurlx (approved):
ACK - Still works for me on macOS and tried it with the changed update message on Windows 10, which works as well.

devinbileck:
ACK. Works for me as well on regtest using Windows 10.

devinbileck (commented):
ACK

ManfredKarrer:
Has anyone tested notifiactions? Not sure if the change there could break something.

devinbileck:
Yes, I tested alert notifications as well as private messages (via arbitration). And also used some random UTF-8 characters in the messages. Didn't encounter any issues.

PR #2740, opened 2019-04-18
Improve handling of editing of offers with invalid security deposit

ripcurlx:
Shows warning popup when offer with invalid security deposit is edited.
@m52go Could you please review the text? Thanks!
Bildschirmfoto 2019-04-18 um 13 19 24

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -493,6 +493,7 @@ editOffer.confirmEdit=Confirm: Edit offer
 editOffer.publishOffer=Publishing your offer.
 editOffer.failed=Editing of offer failed:\n{0}
 editOffer.success=Your offer has been successfully edited.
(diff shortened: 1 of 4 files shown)


m52go:
Here is a general message: 

> The buyer's security deposit is not within the constraints defined by the Bisq DAO and can no longer be edited.

If this message will always be shown to the person making the edit, I would suggest simply saying:

> Your security deposit is not within the constraints defined by the Bisq DAO and can no longer be edited.

ripcurlx:
> Here is a general message:
> 
> > The buyer's security deposit is not within the constraints defined by the Bisq DAO and can no longer be edited.
> 
> If this message will always be shown to the person making the edit, I would suggest simply saying:
> 
> > Your security deposit is not within the constraints defined by the Bisq DAO and can no longer be edited.

It will always be the maker, but she can be the buyer or seller. So I'll go with the general message -thx!

ManfredKarrer (approved):
utACK

PR #2742, opened 2019-04-18
Use different security deposit for fiat-crypto and crypto-crypto trades

ripcurlx:
Sets back security deposit for crypto-crypto trades to values before the DAO and only keeps the new ones for fiat-crypto trades.

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/Restrictions.java
+++ b/core/src/main/java/bisq/core/btc/wallet/Restrictions.java
@@ -18,8 +18,12 @@
 package bisq.core.btc.wallet;
 
 import bisq.core.app.BisqEnvironment;
+import bisq.core.payment.PaymentAccount;
+import bisq.core.payment.PaymentAccountUtil;
 
 import org.bitcoinj.core.Coin;
+
+import javax.annotation.Nullable;
 
 public class Restrictions {
     private static Coin MIN_TRADE_AMOUNT;
@@ -50,16 +54,25 @@ public class Restrictions {
         return MIN_TRADE_AMOUNT;
     }
 
-    public static double getDefaultBuyerSecurityDepositAsPercent() {
-        return 0.1; // 10% of trade amount.
+    public static double getDefaultBuyerSecurityDepositAsPercent(@Nullable PaymentAccount paymentAccount) {
+        if (PaymentAccountUtil.isCryptoCurrencyAccount(paymentAccount))
+            return 0.02; // 2% of trade amount.
+        else
+            return 0.1; // 10% of trade amount.
     }
 
-    public static double getMinBuyerSecurityDepositAsPercent() {
-        return 0.05; // 5% of trade amount.
+    public static double getMinBuyerSecurityDepositAsPercent(@Nullable PaymentAccount paymentAccount) {
+        if (PaymentAccountUtil.isCryptoCurrencyAccount(paymentAccount))
+            return 0.005; // 0.5% of trade amount.
+        else
+            return 0.05; // 5% of trade amount.
     }
 
-    public static double getMaxBuyerSecurityDepositAsPercent() {
-        return 0.5; // 50% of trade amount. For a 1 BTC trade it is about 800 USD @ 4000 USD/BTC
+    public static double getMaxBuyerSecurityDepositAsPercent(@Nullable PaymentAccount paymentAccount) {
+        if (PaymentAccountUtil.isCryptoCurrencyAccount(paymentAccount))
+            return 0.2; // 20% of trade amount. For a 1 BTC trade it is about 800 USD @ 4000 USD/BTC
+        else
+            return 0.5; // 50% of trade amount. For a 1 BTC trade it is about 2000 USD @ 4000 USD/BTC
     }
 
     // We use MIN_BUYER_SECURITY_DEPOSIT as well as lower bound in case of small trade amounts.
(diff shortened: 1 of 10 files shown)


ripcurlx:
I'll change the condition to differentiate from currency code to payment account so we are more flexible in the future if we ever decide to have different deposit limits for different payment accounts.

core/src/main/java/bisq/core/btc/wallet/Restrictions.java
```diff
@@ -18,9 +18,12 @@
 package bisq.core.btc.wallet;
 
 import bisq.core.app.BisqEnvironment;
+import bisq.core.payment.PaymentAccount;
 
 import org.bitcoinj.core.Coin;
 
+import static bisq.core.payment.PaymentAccountUtil.isCryptoCurrencyAccount;
```
sqrrm: Is it appropriate to do a static import here? It probably doesn't matter in this class, I feel it makes code less readable when it's not that frequently used. More a matter of our policy than anything.

core/src/main/java/bisq/core/btc/wallet/Restrictions.java
```diff
@@ -50,16 +53,25 @@ public static Coin getMinTradeAmount() {
         return MIN_TRADE_AMOUNT;
     }
 
-    public static double getDefaultBuyerSecurityDepositAsPercent() {
-        return 0.1; // 10% of trade amount.
+    public static double getDefaultBuyerSecurityDepositAsPercent(PaymentAccount paymentAccount) {
```
sqrrm: It's nice to know when nullable is allowed

desktop/src/main/java/bisq/desktop/main/offer/MutableOfferDataModel.java
```diff
@@ -282,6 +282,8 @@ public boolean initWithData(OfferPayload.Direction direction, TradeCurrency trad
         setTradeCurrencyFromPaymentAccount(paymentAccount);
         tradeCurrencyCode.set(this.tradeCurrency.getCode());
 
+        setTradeCurrencyFromPaymentAccount(paymentAccount);
```
sqrrm: I don't understand why this needs to be done twice, it's set three lines up.

sqrrm (commented):
I think this is mostly correct as far as I know, I'm not that familiar with the trade protocol though so I can't tell if there is some case where the deposit should be updated.

Also, see inline comments.

core/src/main/java/bisq/core/btc/wallet/Restrictions.java
ripcurlx: I'll add it again. I was too much in IntelliJ automation mode 😉

core/src/main/java/bisq/core/btc/wallet/Restrictions.java
ripcurlx: I'll add the Nullable annotation where appropriate.

desktop/src/main/java/bisq/desktop/main/offer/MutableOfferDataModel.java
ripcurlx: Good catch was added by mistake.

ripcurlx:
> I think this is mostly correct as far as I know, I'm not that familiar with the trade protocol though so I can't tell if there is some case where the deposit should be updated.
> 
> Also, see inline comments.

Thanks for the review. I force pushed all changes.

core/src/main/java/bisq/core/user/PreferencesPayload.java
```diff
@@ -47,6 +47,8 @@
 
 import javax.annotation.Nullable;
 
+import static bisq.core.btc.wallet.Restrictions.getDefaultBuyerSecurityDepositAsPercent;
```
sqrrm, line 50: More static imports, in case you meant to revert them all, otherwise it's ok, just feels a bit excessive for these kinds of use cases.

desktop/src/main/java/bisq/desktop/util/validation/SecurityDepositValidator.java
```diff
@@ -17,21 +17,28 @@
...
 
-import bisq.core.btc.wallet.Restrictions;
 import bisq.core.locale.Res;
+import bisq.core.payment.PaymentAccount;
 import bisq.core.util.BSFormatter;
 
 import javax.inject.Inject;
 
+import static bisq.core.btc.wallet.Restrictions.getMaxBuyerSecurityDepositAsPercent;
+import static bisq.core.btc.wallet.Restrictions.getMinBuyerSecurityDepositAsPercent;
```
sqrrm: More static imports

sqrrm (approved):
utACK

As a reviewer I prefer when the previous commits are kept and just add a new one rather than a force push. Easier to follow which changes were made after comments were made that way.

ripcurlx:
> utACK
> 
> As a reviewer I prefer when the previous commits are kept and just add a new one rather than a force push. Easier to follow which changes were made after comments were made that way.

True - I wanted to reduce the amount of commits, but in a PR that gets reviewed it makes it harder to follow.

core/src/main/java/bisq/core/user/PreferencesPayload.java
ripcurlx: Here I left it by intent to reduce the line length for L125 & L127

desktop/src/main/java/bisq/desktop/util/validation/SecurityDepositValidator.java
ripcurlx: Here it doesn't help to make the code more readable. Commit below.

sqrrm (approved):
utACK

ManfredKarrer (approved):
utACK

PR #2752, opened 2019-04-20
Add filter to trade history

ManfredKarrer:


ripcurlx (approved):
ACK - just added a prompt text for filter textfield to give the user an idea what she can filter for

ManfredKarrer:
Local build works, so failure above seems to be Travis issue

PR #2753, opened 2019-04-20
Fix account age display of peer

ManfredKarrer:
We only showed makers account age before also at trade screens.
This PR fixes that so that the trade peers account age is displayed at
the peer info icon.


ripcurlx (approved):
utACK

PR #2763, opened 2019-04-21
Use same precision for all x-axis labels

sqrrm:
Make sure all labels use the same precision for the x-axis, looks really strange when some labels are precision 8 and some 3.


ripcurlx (approved):
utACK

PR #2767, opened 2019-04-22
Fix pricenode gradle config

devinbileck:
When launching the bisq-pricenode start script in the root dir,
the following error was encountered:

> Error: Could not find or load main class bisq.price.Main
> Caused by: java.lang.ClassNotFoundException: bisq.price.Main


build.gradle
```diff
@@ -354,7 +354,7 @@ configure(project(':monitor')) {
 
 
 configure(project(':pricenode')) {
-    apply plugin: "org.springframework.boot"
+    mainClassName = 'bisq.price.Main'
```
ManfredKarrer, line 357: I am not familiar with Spring. Is org.springframework.boot not needed amymore?

devinbileck: To be honest, I am not familiar with it either. But the only way I found to eliminate the error with unable to load main class when running the start script was to not apply the plugin. However, there is now an issue with trying to run the jar. I will continue to investigate.

PR #2768, opened 2019-04-23
Sign account usage

ManfredKarrer:
Implements https://github.com/bisq-network/proposals/issues/78


core/src/main/java/bisq/core/account/sign/SignedWitnessService.java
```diff
@@ -0,0 +1,237 @@
...
+                addToMap((SignedWitness) e);
+        });
+    }
+
+
+    ///////////////////////////////////////////////////////////////////////////////////////////
+    // API
+    ///////////////////////////////////////////////////////////////////////////////////////////
+
+    // Arbitrators sign with EC key
```
sqrrm, line 105: Could you add a short explanation why arbitrators are using EC keys?

core/src/main/java/bisq/core/account/sign/SignedWitnessService.java
```diff
@@ -0,0 +1,237 @@
...
+
+    // We go one level up by using the signer Key to lookup for SignedWitness objects which contain the signerKey as
+    // witnessOwnerPubKey
+    public Set<SignedWitness> getSignedWitnessSetBySignerPubKey(byte[] signerPubKey) {
+        return signedWitnessMap.values().stream()
+                .filter(e -> Arrays.equals(e.getWitnessOwnerPubKey(), signerPubKey))
+                .collect(Collectors.toSet());
+    }
+
+    public boolean isValidAccountAgeWitness(AccountAgeWitness accountAgeWitness) {
```
sqrrm, line 196: There need to be a break in case of circular signing. If someone creates two witnesses that sign each other I think this recursion would go into an infinite loop.

sqrrm (commented):
Looks like a reasonable approach. Quite simple and not tying in to other parts.

core/src/main/java/bisq/core/account/sign/SignedWitnessService.java
ManfredKarrer: Hm... not sure if that can happen. Can you describe the case? If Alice signs Bob and Bob signs Carol and Carol signs Bob it is not an issue as the witness objects are the relevant objects which build the chain, not the signers. But in the lookup for the signers witness we need to probably use the date to use the oldest.

It also does not support yet handling of mulitple signers and when you are allowed to sign...

sqrrm: Alice signs Bob's witness and Bob signs Alice's witness.

First level, trying to validaty Bob's witness, we find that it was signed by Alice. We then try to validate Alice's witness and during the second recursion we find that it's signed by Bob and will then try to validate it. Back to where we started in an infinite recursion.

Some leeway has to be given to the time stamps so an attack could be to create this pair and set the same date for both signedWitnesses.

I might well be missing something, and I don't think that's a risk during normal operation, just good to be safe and not add any attack vectors.

ManfredKarrer: I think we need to pass through the list and remove items once processed.

ManfredKarrer:
Replaced by https://github.com/bisq-network/bisq/pull/2787

PR #2769, opened 2019-04-23
Reflect pricenode operator change in ProvidersRepository

devinbileck:
I have taken over one of the pricenodes from Manfred.

See bisq-network/roles#14


ManfredKarrer (approved):
utACK

PR #2770, opened 2019-04-23
Update mobile notification relay node DEV_URL IP address

devinbileck:
I have taken over the relay node from Manfred. The onion address
remains the same, but the IP address used in the DEV_URL has
changed.

See bisq-network/roles#82


ManfredKarrer (approved):
utACK

PR #2765, opened 2019-04-22
[WIP] Delay payout if fiat buyer has new account

ManfredKarrer:
This PR has merged https://github.com/bisq-network/bisq/pull/2753 as some of those features are used.

Implements proposal https://github.com/bisq-network/proposals/issues/77

Here are the relevant screens:

Create account:
Screen Shot 2019-04-22 at 00 10 28

Popup after account creation:
Screen Shot 2019-04-22 at 00 11 10

Popup at create buy offer:
Screen Shot 2019-04-22 at 00 11 38

Popup at take buy offer:
Screen Shot 2019-04-22 at 00 12 35

Popup at confirm fiat receipt:
Screen Shot 2019-04-22 at 00 14 07

Status while waiting until buyers account matures:
Screen Shot 2019-04-22 at 00 14 44

After buyers account has matured:
Screen Shot 2019-04-22 at 00 16 36

PR #2787, opened 2019-04-28
[WIP] Add protection tools

ManfredKarrer:
Merges https://github.com/bisq-network/bisq/pull/2765 and https://github.com/bisq-network/bisq/pull/2768


core/src/main/java/bisq/core/account/score/ScoreInfo.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Optional;
+
+import lombok.Value;
+
+@Value
+public class ScoreInfo {
+    private AccountScoreCategory accountScoreCategory;
+    private long accountAge;
+    private Optional<Long> signedTradeAge;
+    private Coin minDepositAsCoin; //TODO remove?
```
ripcurlx, line 31: I think this will be gone as we are not able to force this from the maker's side - right?

core/src/main/java/bisq/core/account/sign/SignedWitness.java
```diff
@@ -0,0 +1,156 @@
...
+        DateTolerantPayload, CapabilityRequiringPayload {
+    private static final long TOLERANCE = TimeUnit.DAYS.toMillis(1);
+
+    private final boolean signedByArbitrator;
+    private final byte[] witnessHash;
+    private final byte[] signature;
+    private final byte[] signerPubKey;
+    private final byte[] witnessOwnerPubKey;
+    private final long date;
+    //TODO should we add trade amount?
```
ripcurlx: Is the idea to add the trade amount to add this as a weight later on as well? Of course a 1000 USD trade might not be undetected compared to a 10 USD trade.

core/src/main/java/bisq/core/account/sign/SignedWitness.java
```diff
@@ -0,0 +1,156 @@
...
+
+
+    ///////////////////////////////////////////////////////////////////////////////////////////
+    // API
+    ///////////////////////////////////////////////////////////////////////////////////////////
+
+    @Override
+    public boolean isDateInTolerance() {
+        // We don't allow older or newer then 1 day.
+        // Preventing forward dating is also important to protect against a sophisticated attack
```
ripcurlx, line 128: The date is not part of the hash. Is this by intent?

core/src/main/java/bisq/core/account/sign/SignedWitnessService.java
```diff
@@ -0,0 +1,257 @@
...
+
+    //TODO pass list and remove items once processed to avoid endless loop in case of multiple sigs
+    public boolean isValidAccountAgeWitness(AccountAgeWitness accountAgeWitness) {
+        Set<SignedWitness> arbitratorsSignedWitnessSet = getArbitratorsSignedWitnessSet(accountAgeWitness);
+        if (!arbitratorsSignedWitnessSet.isEmpty()) {
+            // Our peer was signed by arbitrator. We only check it at least one is valid and don't need to go further.
+            return arbitratorsSignedWitnessSet.stream().anyMatch(this::verify);
+        } else {
+            Set<SignedWitness> trustedPeerSignedWitnessSet = getTrustedPeerSignedWitnessSet(accountAgeWitness);
+            // We have some SignedWitness singed by any trusted peer and need to see if it is valid and has a
```
ripcurlx: typo `singed` should be `signed`

core/src/main/java/bisq/core/account/sign/SignedWitness.java
ManfredKarrer: Will think about it

ManfredKarrer: Just added a comment:
  // Date is not added to hash to avoid that the same account with same sigs can be stored multiple 
        // times if date would differ

core/src/main/java/bisq/core/account/sign/SignedWitness.java
ManfredKarrer: Not sure yet if needed. Simplicity might be better here.... but leave the comment open until 100% clear.

core/src/main/java/bisq/core/account/score/ScoreInfo.java
ManfredKarrer: Probably

ManfredKarrer:
I am considering that implementation approach to be on hold. See https://github.com/bisq-network/proposals/issues/77#issuecomment-489200973 for my concerns with the approach of using a delayed payout.

ManfredKarrer:
That approach has been dropped in favor of other ideas (see https://github.com/bisq-network/proposals/issues/78, https://github.com/bisq-network/proposals/issues/83)

Issue #2754, opened 2019-04-20
Not sure, software asked to report.


inpharmaticist:
[Uploading bisq error.txt…]()

ghost:
The link leads to no logfile.

inpharmaticist:
[bisq error.txt](https://github.com/bisq-network/bisq/files/3100191/bisq.error.txt)

ManfredKarrer:
@oscarguindzberg Could you have a look to the log file, there are many connection failed logs. We got another similar report recently. Any idea what can cause that?

oscarguindzberg:
I had a look at the logs.
I see all the "[BlockingClient network thread for xxxxx.onion:8333] ERROR o.b.n.BlockingClient: Error trying to open/read from connection: xxxxx.onion:8333".
I see these error lines on my Bisq node too.
I tried to figure out what could be wrong, but could not find anything.
So,
- tor network instability could be the reason, #2773 was fixed by trying again after an OS restart.
- "Use bitcoinj with connection handling audit fixes" PR (https://github.com/bisq-network/bisq/pull/2618) was not merged. Although there is no direct fix for this problem, I would like to see whether we keep seeing this problem when that PR is merged.

ghost:
BlockingClient
is that not the banned clients because of the running scam ?

oscarguindzberg:
@HarryMacfinned BlockingClient is a bitcoinj class unrelated to the scam.

oscarguindzberg:
@inpharmaticist were you able to get your funds back?

inpharmaticist:
I don't think I lost any funds on the transaction. Everything worked after I saw the error to submit the bug (I'm not a programmer and new to github so I'm out of my element here).  I just submitted because I got the alert. I believe I am still whole though.

oscarguindzberg:
I will post what I could find out:

This is the relevant part of the logs 
```
Apr-19 21:22:11.782 [JavaFX Application Thread] INFO  b.c.a.DisputeManager: Received DisputeResultMessage with tradeId PVByYn-15ccdcc8-436f-42a7-a638-ea093047ba61-097 and uid bb4145dc-e013-4553-bbb0-55be3613dbb0 
Apr-19 21:22:11.788 [JavaFX Application Thread] ERROR b.c.b.w.BtcWalletService: multiSigAddressEntry not found for trade ID=PVByYn-15ccdcc8-436f-42a7-a638-ea093047ba61-097.
[...]

PR #2789, opened 2019-04-29
Improve logging

oscarguindzberg:
Print full stack trace on the log file 
This could be useful to investigate this issue: https://github.com/bisq-network/bisq/issues/2754


ManfredKarrer (approved):
utACK

PR #2801, opened 2019-05-02
Add check for account age to apply restrictions

ManfredKarrer:
This PR provides some security improvements to enable us to unblock the banned SEPA counties and Interac due to the chargeback scams. 

We limit accounts created after 15.3. to 0.01 BTC trades in case they are the fiat buyer of payment methods considered risky regarding chargebacks (basically all bank transfer methods).

This is a intermediary solution as the planned proposals will require more time to get implemented and deployed.

To ensure that users with old version cannot be caught by the scammer once we unblock the banned counties and Interac we will send out a message which enforces all users to update, otherwise their trades will fail. We will do that shortly after the release (about 2 days).

### DIFF
--- a/core/src/main/java/bisq/core/payment/AccountAgeWitnessService.java
+++ b/core/src/main/java/bisq/core/payment/AccountAgeWitnessService.java
@@ -63,6 +63,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 public class AccountAgeWitnessService {
     private static final Date RELEASE = Utilities.getUTCDate(2017, GregorianCalendar.NOVEMBER, 11);
     public static final Date FULL_ACTIVATION = Utilities.getUTCDate(2018, GregorianCalendar.FEBRUARY, 15);
+    public static final long SAFE_ACCOUNT_AGE_DATE = Utilities.getUTCDate(2019, GregorianCalendar.MARCH, 15).getTime();
 
     public enum AccountAge {
         LESS_ONE_MONTH,
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -345,22 +345,22 @@ offerbook.warning.noTradingAccountForCurrency.msg=You don't have a trading accou
 offerbook.warning.noMatchingAccount.headline=No matching trading account.
 offerbook.warning.noMatchingAccount.msg=To take this offer, you will need to set up a payment account using this payment method.\n\nWould you like to do this now?
 
-offerbook.warning.riskyBuyOfferWithImmatureAccountAge=This offer cannot be taken because of security restrictions based on those criteria:\n\
-  - The maker''s payment account was created after March 15th 2019\n\
-  - The min. trade amount is above 0.01 BTC\n\
-  - The payment method for that offer is considered risky for bank chargebacks\n\n{0}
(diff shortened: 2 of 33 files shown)


core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -344,6 +344,16 @@ offerbook.warning.noTradingAccountForCurrency.headline=No trading account for se
 offerbook.warning.noTradingAccountForCurrency.msg=You don't have a trading account for the selected currency.\nDo you want to create an offer with one of your existing trading accounts?
 offerbook.warning.noMatchingAccount.headline=No matching trading account.
 offerbook.warning.noMatchingAccount.msg=You don't have a trading account with the payment method required for that offer.\nYou need to setup a trading account with that payment method if you want to take this offer.\nDo you want to do this now?
+offerbook.warning.makerHasNoMatureAccountForBuyOffer=You cannot create an offer because you do not have a payment account which was created before March 15th 2019. \
```
m52go: Not show-stopping but I made some minor changes:

```
offerbook.warning.noMatchingAccount.msg=To take this offer, you will need to set up a payment account using this payment method.\n\nWould you like to do this now?
offerbook.warning.makerHasNoMatureAccountForBuyOffer=You cannot create this offer because you do not have a payment account which was created before March 15th 2019. \
 The selected payment method is considered risky for bank chargebacks. We needed to deploy this restriction as a short-term measure for enhanced security.\n\n\
 The next software release will provide more robust chargeback protection tools, and this restriction for new accounts will be removed.
offerbook.warning.riskyBuyOfferWithImmatureAccountAge=This offer cannot be taken because the maker's payment account \
 was created after March 15th 2019, and the payment method is considered risky for bank chargebacks. We needed to deploy this restriction as a \
 short-term measure for enhanced security.\n\n\
 The next software release will provide more robust protection tools so that offers with this risk profile can be traded again.
offerbook.warning.sellOfferAndAnyTakerPaymentAccountForOfferMature=This offer cannot be taken because your payment account \
 was created after March 15th 2019 and the payment method is considered risky for bank chargebacks. We needed to deploy this restriction as a \
 short-term measure for enhanced security.\n\n\
 The next software release will provide more robust protection tools so that offers with this risk profile can be traded again.
```

core/src/main/java/bisq/core/payment/AccountAgeWitnessService.java
```diff
@@ -63,6 +63,7 @@
 public class AccountAgeWitnessService {
     private static final Date RELEASE = Utilities.getUTCDate(2017, GregorianCalendar.NOVEMBER, 11);
     public static final Date FULL_ACTIVATION = Utilities.getUTCDate(2018, GregorianCalendar.FEBRUARY, 15);
+    public static final long SAFE_ACCOUNT_AGE_DATE = Utilities.getUTCDate(2019, GregorianCalendar.MARCH, 15).getTime();
```
ripcurlx, line 66: I think we should re-use the static from AccountAgeRestrictions

ripcurlx (approved):
ACK - besides a small UI issue with popups everything worked as expected. @devinbileck and I experienced a problem on Regtest that an old account wasn't able to take a sell offer by a new account that exceeded the demo amount of 0.01 BTC. The reason for this error was that the account age witness data wasn't propagated in the local network at the point the offer was taken. Shouldn't be a problem on Mainnet.

PR #2811, opened 2019-05-05
List burnt BlackCoin (BLK-BURNT)

KrzysiekJ:
- Official project URL: https://blackcoin.org/
- Block explorer URL: https://chainz.cryptoid.info/blk/

For context, see https://bisq.community/t/best-way-to-add-an-ibo/7477 .

### DIFF
--- a/assets/src/main/java/bisq/asset/coins/BurntBlackCoin.java
+++ b/assets/src/main/java/bisq/asset/coins/BurntBlackCoin.java
@@ -1,0 +1,33 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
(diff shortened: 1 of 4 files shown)


assets/src/main/java/bisq/asset/coins/BurntBlackCoin.java
```diff
@@ -0,0 +1,33 @@
...
+import bisq.asset.Coin;
+import bisq.asset.RegexAddressValidator;
+
+@AltCoinAccountDisclaimer("account.altcoin.popup.blk-burnt.msg")
+public class BurntBlackCoin extends Coin {
+    public static final short PAYLOAD_LIMIT = 15000;
+
+    public BurntBlackCoin() {
+        super("Burnt BlackCoin",
+              "BLK-BURNT",
```
KrzysiekJ, line 30: Note that burnt blackcoins are just blackcoins with a peculiar script, hence the ticker should be actually `BLK`. However this would leave no room for adding BlackCoin with normal addresses. On the other hand, `BLK-BURNT` may be problematic in regard to price node (as external exchanges will use the `BLK` ticker).

blabno (approved):
ACK

PR #2813, opened 2019-05-06
Allow seller to take any offers

ManfredKarrer:


sqrrm (approved):
utACK

ripcurlx (approved):
ACK - was able to re-produce it with Zelle in current master and it was fixed in this PR.

ripcurlx:
☝️ Just improved layout of Zelle buyer form on the way to prevent scrolling

sqrrm (approved):
ACK

Looks good

ripcurlx:
Just found a little detail that needed to be changed. Translations and references of March 15th needs to be changed to March 1st. And so the release build starts again 😉

PR #2821, opened 2019-05-11
Fix getBlocks issue

ManfredKarrer:
We disconnected seed nodes of we exceeded our connectionlimit which caused that getBlockResponses where not received. This PR fixes that.


devinbileck (approved):
ACK
Without this fix, I tested rebuilding the DAO state on mainnet and encountered the issue where I lost connection to all seednodes and was stuck synchronizing. I then tested again with this fix and it did not disconnect from the seeds and it successfully resynced the DAO state.

PR #2844, opened 2019-05-17
Fix confirm lockup transaction prompt showing incorrect estimated time

devinbileck:
When attempting to lock up a bond, the confirmation prompt was showing an incorrect estimated time:
image

This PR fixes the estimated time:
image


sqrrm (approved):
utACK

Issue #2840, opened 2019-05-17
Tails interop issue - useless use of IPv6


agb19:
**Summary**: on Tails, Bisq is difficult to use due to a JVM misconfiguration causing Bisq to use an IPv6 socket to connect to IPv4 address 127.0.0.1, which is unsupported. 

**Background**
The Tails live OS uses a Tor control protocol sanitizer called `onion-grater`, which acts as a proxy between the user applications and the Tor control port and restricts the Tor control commands to a limited subset defined on a per-application basis. The `onion-grater` daemon listens on localhost port 9051 and passes the commands to the real Tor control port, which on Tails is set to 9052. An application requiring access to the Tor control port should connect to port 9051 and expect to be allowed to operate within the limits defined in a per-application file located in `/etc/onion-grater.d/` (usually supplied as part of the application package). This is realized by matching the TCP connection to PID to executable via the `psutil` package. 
See for details:
https://github.com/Whonix/onion-grater/blob/f01a889cb97a6afb86044407e307fc819aae7f7e/usr/lib/onion-grater#L162

**Details**: 
When Bisq opens the connection to the Tor control port, it first queries the loopback address - (actually this happens in the netlayer library):
https://github.com/JesusMcCloud/netlayer/blob/3467ae96d1abd834246e6c3629b16e16f6c38cdc/tor.external/src/main/kotlin/org/berndpruenster/netlayer/tor/ExternalTor.kt#L32

Depending on the JVM settings and the local IP configuration, this may return an IPv6 address (::1), an IPv4 address (127.0.0.1) or an IPv4-in-IPv6-format (::ffff:127.0.0.1). In the case of Bisq on Tails, it's the third case. 

Bisq then proceeds to open the Tor control connection using IPv6 formatting, to the IPv4 address. This causes the connection to be represented in the kernel using IPv6 format. However, from onion-grater's perspective, the connection is still IPv4. 
For this reason, the connection-to-PID matching fails in onion-grater and Bisq is denied access. 

**Steps to reproduce**
Bisq version: 1.1.1 installed from .deb
OS: Tails 3.13.2 (Debian 9)
- boot up Tails with administration password enabled
- download and install the Bisq DEB:
  `sudo dpkg -i Bisq-64bit-1.1.1.deb`
- start up Bisq using onion-grater control port:
  `/opt/Bisq/Bisq --torControlPort=9051`

**Expected result**
[...]

PR #2862, opened 2019-06-03
Fix "useless use of IPv6" in Tor socket

agb19:
Implements the "possible fix" described in issue #2840 by appending the directive `java.net.preferIPv4Stack=true` to the JVM options. 

Tested successfully on Debian 9 and Tails 3.14, using both native Tor and onion-grater.

### DIFF
--- a/desktop/package/linux/package.sh
+++ b/desktop/package/linux/package.sh
@@ -108,6 +108,7 @@ $JAVA_HOME/bin/javapackager \
     -srcfiles desktop-$version-all.jar \
     -appclass bisq.desktop.app.BisqAppMain \
     -BjvmOptions=-Xss1280k \
+    -BjvmOptions=-Djava.net.preferIPv4Stack=true \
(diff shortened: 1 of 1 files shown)


sqrrm (commented):
I'm not familiar with the previous discussion and what the problems are of using IPv6 here. The PR looks correct but I'd like to understand the background.

agb19:
@sqrrm - issue #2840 should provide some background. In a nutshell, due to an unfortunate configuration default in JDK 10, Bisq sometimes tries to speak IPv6 to the IPv4 loopback address, causing an interop issue on Tails. 
This change tells Java to speak IPv4 to IPv4 addresses and IPv6 to IPv6, which is a sane choice.

sqrrm:
@agb19 My bad, didn't see that link to the issue actually explaining it.

sqrrm (approved):
utACK

PR #2870, opened 2019-06-05
Monitor knows about the DAO

freimair:
- brought the Monitor up to speed again
- added metric to see if the DAO state between all seed nodes is in sync (https://monitor.bisq.network/d/iVpBgXXmk/dao-sync-status?refresh=5m&orgId=1)


ripcurlx (approved):
utACK

PR #2871, opened 2019-06-05
Add trader chat

sqrrm:
Implementation of https://github.com/bisq-network/proposals/issues/90


christophsturm:
This PR needs some unit tests. @sqrrm do you think there is a way to cherry pick most of it to master without enabling it?

sqrrm:
Yes, I tried to make the first commits only refactoring, then adding the chat functionality to traders, and then adding the UI.

PR #2872, opened 2019-06-06
[WIP] Add account signing

ripcurlx:


ripcurlx:
@sqrrm I moved the code of the existing domain layer into this PR. I'm reviewing the existing codebase right now. Would be great if you also could give it a look, so we don't overlook any issues in the first place.

core/src/main/java/bisq/core/account/sign/SignedWitness.java
```diff
@@ -0,0 +1,170 @@
+/*
+ * This file is part of Bisq.
+ *
+ * bisq is free software: you can redistribute it and/or modify it
```
christophsturm: Bisq

sqrrm: There is an "insert a suggestion" feature where you can make easy one liner code changes through github.

christophsturm: yeah i know but i think in this case it should be a search and replace operation.

christophsturm:
I think we should merge this pr without the UI changes, to get all the new code into master. its mostly new code and I don't see how it can break things.

christophsturm:
most of this is now merged to master. if you rebase this pr with master there should only be the ui parts left

ripcurlx:
> most of this is now merged to master. if you rebase this pr with master there should only be the ui parts left

I'll rebase to see what will be left in this PR.

PR #2884, opened 2019-06-14
Add checkbox to hide extreme offers in chart

sqrrm:


mrosseel:
since no issue is linked I'll comment here, is this the final solution? Other solutions would be a log scale, or an algorithm to cut off offers which are too far away. Available to brain-storm that algorithm ;)

sqrrm:
@mrosseel  Not much thought went into this. I just got annoyed with the BSQ market chart being unusable. This solution hides the extreme offers from the chart (more than a factor 5 from the best offers). I don't really want to put much more work into this right now, which the log scale solution probably would be.

ghost:
Thanks @sqrrm ,
btw, this extreme offers seems really here to intentionally damage.
Sometimes it briefly disappears ... and some hours later they are here again.

ripcurlx (changes requested):
I personally would just remove the extremes without adding new controls as I don't think there is a use case where someone wants to have this kind of display. If we find time in the future to adapt to a logarithmic scale or similar we can just adapt the code again. Besides that ACK.

sqrrm:
Removed the checkbox. There is no indication that some offers might not be displayed in the chart now, but I suppose that's ok, it's about usability, not correctness.

ripcurlx:
> Removed the checkbox. There is no indication that some offers might not be displayed in the chart now, but I suppose that's ok, it's about usability, not correctness.

Yes, I understand your concerns, but we had a similar solution in place before (only display the best offers at all), so I think it is OK if we continue to iterate on this over time.

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -53,6 +53,7 @@ shared.oneOffer=offer
 shared.multipleOffers=offers
 shared.Offer=Offer
 shared.openOffers=open offers
+shared.hideExtremes=Hide extremes
```
ripcurlx: Please remove the translation as well

desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java
```diff
@@ -146,14 +147,20 @@ public OfferBookChartView(OfferBookChartViewModel model, Navigation navigation,
     public void initialize() {
         createListener();
 
+        // Header with currency combobox and checkbox to show extreme values
+        GridPane topGrid = new GridPane();
```
ripcurlx: Please also remove the additional GridPane again, so we don't use more containers as necessary.

ripcurlx (changes requested):
Please have a look at my comments.

sqrrm:
Removed all the extras except for the column option to a checkbox in FormBuilder.

ripcurlx (approved):

PR #2888, opened 2019-06-15
Late vote reveal handling

sqrrm:
Vote reveal has to be published to complete the vote process even after the reveal phase. To indicate votes I added an indication that votes were not counted in result
image

Also added text to show failed votes in proposaldisplay
image

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -1788,7 +1788,8 @@ dao.proposal.table.icon.tooltip.changeVote=Current vote: ''{0}''. Change vote to
 dao.proposal.display.myVote.accepted=Accepted
 dao.proposal.display.myVote.rejected=Rejected
 dao.proposal.display.myVote.ignored=Ignored
-dao.proposal.myVote.summary=Voted: {0}; Vote weight: {1} (earned: {2} + stake: {3});
+dao.proposal.display.myVote.unCounted=Vote was not included in result
+dao.proposal.myVote.summary=Voted: {0}; Vote weight: {1} (earned: {2} + stake: {3}) {4}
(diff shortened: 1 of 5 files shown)


devinbileck (changes requested):
I am encountering a display issue with this new indicator. When initially clicking on a prior voting cycle for which I successfully published the vote reveal, it is showing this new icon indicating it wasn't counted:
image

And clicking on it to bring up the details shows my vote was not included:
image

But viewing the individual votes shows each vote:
image

But when clicking on the current cycle and back to the prior cycle resolves it:
image

sqrrm:
@devinbileck Good find. It was the vote inclusion check that was wrong. Now fixed.

devinbileck:
That fixed it.

However, I am now noticing if I had ignored a proposal and my vote is not included, the ignored proposal is not shown in the results for the prior cycle, but the ones that I attempted to vote upon do show up.

As seen in this screenshot, Bob (bottom) made all the proposals and remained online and published his vote reveal tx, while Alice (top) went offline after the blind vote. Alice ignored the BTC maker fee proposal and it is not shown for her.
image

I haven't yet determined if this is related to this change or not.

devinbileck:
My above mentioned issue is not related to this change. I tested on master and had the same behaviour. In fact after restarting Bob's client, the ignored proposals no longer showed up. Not sure if related but I seem to have a conflict with my DAO state as it is not in sync with my seed. Not sure why at this point.

sqrrm:
It's pretty strange to lose a proposal, once received the client should save it. If you can reproduce that issue it would be good to take a look at. Might cause some issue during voting if it splits the votes.

devinbileck:
Agreed. I will see if I can reproduce it.

devinbileck (approved):
ACK

PR #2890, opened 2019-06-16
Display changed param value at proposal time, fix #2692

### LINKED ISSUE #2692: Results of param change proposal for a previous cycle shows current value of parameter rather than previous value
For example, the BTC maker fee was 0.001 and a proposal was accepted to change it to 0.0015. When viewing the results of that proposal (when currently in a subsequent voting cycle), it shows the current value along with the proposed value, which are both the same. It would be more useful to show the previous value of the parameter, compared to the proposed value.
image

sqrrm:
image


ripcurlx (approved):
utACK

PR #2904, opened 2019-06-19
[WIP] Increase tolerated trade limit from 0.01 BTC to 0.02 BTC

### LINKED ISSUE #95: Increase buy limit for new accounts from 0.01 to 0.02 BTC
> _This is a Bisq Network proposal. Please familiarize yourself with the [submission and review process](https://docs.bisq.network/proposals.html)._

Please vote this proposal on the DAO.

A restriction on buying more than 0.01 BTC for users with fiat payment accounts created after March 1st 2019 was put into effect to increase security after 1 scammer issue. Such small amounts are uneconomic to trade as mining fees can represent 10%+ of traded amount.

### **We propose increasing this limit to 0.02 BTC**. 
This is still low enough to make stolen bank account fraud uneconomic (as each trade increases chance of detection for the scammer) while allowing new users to buy a meaningful amount of BTC with fiat on Bisq.

--Clarification: we propose enacting this change _regardless_ of what happens with protection tools discussed in #79 #93 #83. The limit should be increased even if these new tools have not yet been implemented, as proposed in #91.

ripcurlx:
Fixes https://github.com/bisq-network/proposals/issues/95.

### DIFF
--- a/core/src/main/java/bisq/core/offer/OfferRestrictions.java
+++ b/core/src/main/java/bisq/core/offer/OfferRestrictions.java
@@ -23,7 +23,7 @@ import bisq.core.trade.Trade;
 import org.bitcoinj.core.Coin;
 
 public class OfferRestrictions {
-    public static Coin TOLERATED_SMALL_TRADE_AMOUNT = Coin.parseCoin("0.01");
+    public static Coin TOLERATED_SMALL_TRADE_AMOUNT = Coin.parseCoin("0.02");
 
     public static boolean isOfferRisky(Offer offer) {
         return offer != null &&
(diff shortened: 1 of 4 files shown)


devinbileck (approved):
utACK

sqrrm:
What is the effect on compatibility of old and new clients when we hard code a change like this? Will old clients just see new offers that they consider risky and not be able to trade with them? Will new clients try to take offers of > 0.01 from old clients and get rejected? Should we try to handle at least future cases with some kind of versioning so that it doesn't cause too much extra support and arbitration work?

ripcurlx:
Closing branch for now because of the backwards compatibility risk and the upcoming account signing feature.

PR #2915, opened 2019-06-23
Wait for initialized wallet before processing dispute msgs, fix #2914

### LINKED ISSUE #2914: Nullpointer Exception while startup
Just starting (before entering passwd) I get the stack see attachement
[NullPointerWhileStart.txt](https://github.com/bisq-network/bisq/files/3316631/NullPointerWhileStart.txt)
After entering passwd app hangs in startscreen showing number of peers

sqrrm:


core/src/main/java/bisq/core/arbitration/DisputeManager.java
```diff
@@ -266,6 +268,13 @@ private boolean isReadyForTxBroadcast() {
     }
 
     private void applyMessages() {
+        // Some messages can't be handled until all the services are properly initialized.
+        // In particular it's not possible to complete the signing of a dispute payout
+        // by an encrypted wallet until after it's been decrypted.
+        if (!servicesInitialized) {
+            return;
```
devinbileck: Just wondering if it would make more sense to put this check in tryApplyMessages? Since it appears to be the only caller of applyMessages and has an additional check.

devinbileck (approved):
utACK

core/src/main/java/bisq/core/arbitration/DisputeManager.java
sqrrm: I moved to the readiness check, looks much more reasonable, thanks.

sqrrm: Please ack again if it looks good.

devinbileck (approved):
utACK

PR #2920, opened 2019-06-24
Resurrect testnet

devinbileck:
In order to perform more in depth testing, as opposed to simply using regtest, particularly for potential bitcoinj 0.15 integration, I have resurrected testnet. Since the old testnet seednode is no longer maintained by Manfred, I am currently running a seednode and have also created a BSQ genesis transaction. In the process of learning how to create the BSQ genesis transaction and struggling to do it using the Bitcoin core wallet, I created a shell script that can be used to quickly and easily generate one on either testnet or regtest.

Note, as we encountered previously, testnet continues to be volatile and frequently encounter very quick blocks. This means it cannot be used for reliable timing of DAO phases/cycles. Even so, it is still a good test environment.


sqrrm (approved):
utACK

Great script to have. Would've been good to have while developing the DAO.

Issue #2910, opened 2019-06-21
Major errors happening when user tries to accept trade.


spottedmarley:
This error has happened several times since version 1.0 but this is the first time I have had the opportunity to get a screen capture of the actual error dialog and the log file to go with. I have traded alot using bisq and this never used to happen before 1.0.

What is common to each time I've seen this issue is that I have several open "money order" trades that are in the "Payment started" faze and waiting for me to receive and confirm payment. Then someone tries to accept one of my open trade offers and I get flooded with errors. The initial error states that someone attempted to accept my trade offer and an error occured.. this dialog only has a "close" button.. when I click "close" then I get the dialog that you see in the screen shot: Please see attached screen shot of the error dialog and log file.

You can even see in the screen shot that Bisq is telling me that "Your offer has been accepted ..." that is what sets this error off. The offer is not actually accepted though.

Here is the error that is copied to my clipboard: PeerAddress must not be null (sendEncryptedMailboxMessage)

When this error occures it tends to also corrupt my entire portfolio so that ALL of my trades are no longer listed. All my open trades and open offers are gone.

I have incremental backups of my data folder so I can recover from this but why does this keep happening? Like I said, it never happened until v1.0 and the DAO was turned on in the client.

![bisq-error-2](https://user-images.githubusercontent.com/2573814/59896012-db0d7780-93b4-11e9-96fa-446174229c12.png)
[bisq.log](https://github.com/bisq-network/bisq/files/3312911/bisq.log)

You may also notice from the screen shot that, yes, I am currently watching the first season of Miami Vice.

spottedmarley:
I'm going back to v0.9.8 until the platform is a little more stable. Works like a charm :)

a123b:
@spottedmarley The error message popped up ~3 minutes before the "Your offer has been accepted" message, right? It would be interesting to know whether the portfolio gets wiped after the error message or only after the trade notification.

Quick breakdown of the log file:
- We are a maker selling BTC
- Someone tries to take the offer
- We disagree about the market price with a difference slightly above the max. allowed 1%
- Sending the `AckMessage` for the taker's `PayDepositRequest` with the error message about the price disagreement fails because the `PeerAddress` is `null`
- ~3 minutes later, the taker tries taking the offer again, this time, the price difference is within tolerance and everything works as expected (deposit tx gets published)

[...]

PR #2926, opened 2019-06-27
[WIP] Don't ask for trade volume if non positive price, fix #2925

### LINKED ISSUE #2925: Locked up funds from a failed trade
See also issue #2924 reported just minutes ago.

This issue came with to pop-ups.

First pop-up is a warning about locked-up funds from a failed trade
Warning-locked-up-funds

Second pop-up is an error with java.lang.IllegalArgumentException. This one asks me to file an issue.
Error-illegal-argument

Finally I'm sending you the corresponding log file
[bisq.log](https://github.com/bisq-network/bisq/files/3334369/bisq.log)

Thanks

sqrrm:
I have been able to test this, not sure how to provoke the 0 price trade. Seems reasonable to return null amount for non positive prices though.

### DIFF
--- a/core/src/main/java/bisq/core/trade/Trade.java
+++ b/core/src/main/java/bisq/core/trade/Trade.java
@@ -716,7 +716,8 @@ public abstract class Trade implements Tradable, Model {
 
     @Nullable
     public Volume getTradeVolume() {
-        if (getTradeAmount() != null && getTradePrice() != null) {
+        if (getTradeAmount() != null && getTradePrice() != null &&
+                getTradePrice().getMonetary().getValue() > 0) {
             Volume volumeByAmount = getTradePrice().getVolumeByAmount(getTradeAmount());
             if (offer != null) {
                 if (offer.getPaymentMethod().getId().equals(PaymentMethod.HAL_CASH_ID))


a123b:
Wouldn't it be much better to fix this by ensuring that the `tradePrice` property can never be non-positive? Negative or zero values don't make any sense whatsoever and might cause even bigger problems somewhere else in the code...

devinbileck:
I agree it would be ideal to find and resolve the underlying issue. I am wondering if it is similar to #2915 in that something wasn't completely initialized when the taker took the offer, hence a null peer address and invalid trade price. Perhaps we need an additional check in https://github.com/bisq-network/bisq/blob/master/desktop/src/main/java/bisq/desktop/util/GUIUtil.java#L737

devinbileck:
I think it may be due to the fact that if the price is not in tolerance, it does not set the trade price.
https://github.com/bisq-network/bisq/blob/master/core/src/main/java/bisq/core/trade/protocol/tasks/maker/MakerProcessPayDepositRequest.java#L85

Perhaps should setTradePrice prior to checkTradePriceTolerance.

ManfredKarrer:
@devinbileck 
No setTradePrice have to be after the check as we don't want to write a possible incorrect price into the trade object. If the price provider never delivered a price the check will fail but I don't remember to have seen that ever and I am not aware of any issues with provider nodes, they run usually very stable.

ManfredKarrer:
To add the extra check in getTradeVolume would return a null and it can be that this will cause other issuer later, so I am not sure if that fix would make the problem better. I agree we should find out what happened exactly. If the user can provide the full logs that might help. If the price was never delivered that should be visible.

a123b:
I _think_ I know what's happening:
- `TradeManager` adds trade in `handlePayDepositRequest()`
- Price disagreement leads to NullPointerException (see [my comment on #2910](https://github.com/bisq-network/bisq/issues/2910#issuecomment-505533956))
- Thus, `cleanupTradableOnFault()` never gets called in [TradeProtocol.java#L173](https://github.com/bisq-network/bisq/blob/v1.1.2/core/src/main/java/bisq/core/trade/protocol/TradeProtocol.java#L173) and the trade falsely stays in the TradeManager/Open Trades list
- On the next `handlePayDepositRequest()`, the trade gets added _again_, so we have it in the TradeManager _twice_: One correct version and one version without a price

Thoughts?
If this is correct, do you think my fix of the underlying NPE in #2928 is enough or should we ensure that whenever a trade is added to the TradeManager's `tradableList`, already existing entries with the same ID are removed first to prevent possible similar issues like this in the future?

ManfredKarrer:
@a123b Great finding! Would be best if you can reproduce that on a local regtest setup to verify the assuption but looks likely that this is a correct analysis. Once reproduceable it's easier to see what path is the best to fix it. The price disagreement should already clean up the trade state, so I think that might be the best place to fix it. 

To strategy to fail in case of uncertainty is intentional. Better to end up in a dispute where a human arbitrator can resolve the issue than that the protocol continues and make a wrong payout.

sqrrm:
I have managed to reproduce it using the following steps in regtest:
1. Alice create offer
1. Bob, running modified source doubling the price when taking offers, tries to take the offer. Alice sees the same issue as #2910 
1. Bob restarts running from unmodified master and takes the same offer. Alice sees the same issue as #2925

sqrrm:
@a123b After reproducing and following what's going on you seem to be spot on. The fix in https://github.com/bisq-network/bisq/pull/2928 solves all of the issues as you describe in https://github.com/bisq-network/bisq/issues/2910#issuecomment-505533956 and https://github.com/bisq-network/bisq/pull/2926#issuecomment-506949237

This PR solves the issue of displaying the already erroneously added 0 price trade item which could be nice for the users who already has such trades in their history. We could alternatively look at purging the saved trades for those users.

sqrrm:
I opened another PR for the purging of failed trades due to price disagreement at https://github.com/bisq-network/bisq/pull/2930

I think that one makes the most sense to merge since it will fix the issues users with broken trades.

@a123b I wasn't able to suggest you as a reviewer, I suppose you're not in the bisq org. Consider joining, your contributions are great.

sqrrm:
When I think about it, perhaps it makes more sense to just make a separate build for those with broken histories from that branch to avoid affecting everyone else. Once run and the purged trade list is saved they can return to using the main release.

PR #2928, opened 2019-06-28
Use tempTradingPeerNodeAddress if the trade doesn't have a trading peer set yet

### LINKED ISSUE #2910: Major errors happening when user tries to accept trade.

### LINKED ISSUE #2924: Error while someone took my offer
I missed to take screenshots of the error message, but it was along the lines that the price was off the allowed range. However I can provide the log file.
[bisq.log](https://github.com/bisq-network/bisq/files/3334337/bisq.log)
Later I got another warning and error regarding locked funds, which I'm going to report separately

a123b:
This should fix the errors on the maker side when there's a disagreement
about market price.

Fixes #2910
Fixes #2924

### DIFF
--- a/core/src/main/java/bisq/core/trade/protocol/TradeProtocol.java
+++ b/core/src/main/java/bisq/core/trade/protocol/TradeProtocol.java
@@ -198,7 +198,9 @@ public abstract class TradeProtocol {
                 tradeId,
                 result,
                 errorMessage);
-        final NodeAddress peersNodeAddress = trade.getTradingPeerNodeAddress();
+        // If there was an error during offer verification, the tradingPeerNodeAddress of the trade might not be set yet.
+        // We can find the peer's node address in the processModel's tempTradingPeerNodeAddress in that case.
+        final NodeAddress peersNodeAddress = trade.getTradingPeerNodeAddress() != null ? trade.getTradingPeerNodeAddress() : processModel.getTempTradingPeerNodeAddress();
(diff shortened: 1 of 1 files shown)


devinbileck (approved):
utACK

sqrrm (approved):
ACK

PR #2929, opened 2019-06-29
Fix offer editing for low volume offers

a123b:
A user on the forum had problems editing an offer:
https://bisq.community/t/alt-buy-offer-not-editable/7930

For very low volume offers, it can happen that the minimum buyer security
deposit by coin value is bigger than the maximum security deposit by
percentage of trade amount.
This caused the security deposit validity check in the offer editor to fail,
making it impossible to edit these offers.
This PR fixes the problem by setting the default percentage value to
the model instead of the calculated one in this specific case.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/portfolio/editoffer/EditOfferDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/portfolio/editoffer/EditOfferDataModel.java
@@ -23,6 +23,7 @@ import bisq.desktop.main.offer.MutableOfferDataModel;
 import bisq.core.btc.TxFeeEstimationService;
 import bisq.core.btc.wallet.BsqWalletService;
 import bisq.core.btc.wallet.BtcWalletService;
+import bisq.core.btc.wallet.Restrictions;
 import bisq.core.filter.FilterManager;
 import bisq.core.locale.CurrencyUtil;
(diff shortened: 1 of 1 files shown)


sqrrm (commented):
NACK

While testing this seems to allow editing an offer but while editing the amount to buy is now forced to 1 BTC and grayed out. After editing and saving an offer I'm not able to edit it again.

a123b:
Strange... Can't reproduce it with my test offer.
If you send the details of the problematic offer I might take another look at it tomorrow.

sqrrm:
Open offer as:
image
edit usd offer:
image
edit alt coin offer:
image

Attaching log but I don't think it's very interesting
[issue_2929.log](https://github.com/bisq-network/bisq/files/3350244/issue_2929.log)

Looks like the amount defaults to 1 BTC after editing. It might not be related to your fix though.

a123b:
Oh, I think that's just the [dev mode setting test values](https://github.com/bisq-network/bisq/blob/v1.1.2/desktop/src/main/java/bisq/desktop/main/offer/MutableOfferViewModel.java#L225)

sqrrm:
Indeed, I don't think I tried editing an offer while testing before.

sqrrm:
ACK

Sorry, that was my bad with dev mode.

PR #2930, opened 2019-06-29
Purge saved trades that failed market price check

sqrrm:
This is a way to clean saved failed take offers due to price disagreement. I don't know if there is any other signature apart from the errormessage to uniquely identify such saved failed trades so the matching for purges is done on the error message string.

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqSetup.java
+++ b/core/src/main/java/bisq/core/app/BisqSetup.java
@@ -161,7 +161,7 @@ public class BisqSetup {
     private Consumer<String> cryptoSetupFailedHandler, chainFileLockedExceptionHandler,
             spvFileCorruptedHandler, lockedUpFundsHandler, daoErrorMessageHandler, daoWarnMessageHandler,
             filterWarningHandler, displaySecurityRecommendationHandler, displayLocalhostHandler,
-            wrongOSArchitectureHandler;
+            wrongOSArchitectureHandler, cleanupDone;
     @Setter
     @Nullable
     private Consumer<Boolean> displayTorNetworkSettingsHandler;
@@ -283,6 +283,9 @@ public class BisqSetup {
     }
 
     public void start() {
+        if (cleanupDone != null)
+            cleanupDone.accept("Cleanup done, shutting down");
+
         maybeReSyncSPVChain();
         maybeShowTac();
     }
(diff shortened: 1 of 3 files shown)


sqrrm:
This branch is for those that have a broken trade history due to the issue with failed acks on price disagreement. This branch will fix the data dir so they can keep the histories.

sqrrm:
I created a new branch since this should not go into master. That way those that need to clean up their failed offers can do so by building and running from this branch as a one time clean up.

devinbileck:
I worry that the barrier of building from source will keep a lot of less technical users from being able to utilize this. Perhaps instead we could implement a separate utility that the user just executes and it repairs their trade history. And if necessary we can also add additional cleanup tasks to this utility in order to resolve other issues that arise. What do you think?

sqrrm:
@devinbileck I agree that building from source is probably too hard for a lot of users. We could build this branch for them as well, like the main release.

Having a separate tool would be perhaps more correct but I don't really see many benefits. This fix is very specific to repair the data dir for a bug that has now been fixed so it's a one time use. The separate tool would have to be maintained just as much as bisq itself if it were to be reused and we would have to build it for different platforms as well. To add any task that users might want to perform regularly in a separate tool doesn't make sense to me. Why wouldn't we just implement that in core to never have the issue in the first place?

ghost:
If there is no risk, indeed putting it in the core seems the simplest solution.
And if really this fix will be of limited use, it could be removed in future release when no more needed.

sqrrm:
@HarryMacfinned This is not a suggestion to include it in master for future releases. The suggestion is to keep a separate branch to avoid including it in master and avoid that everyone has to run it. It is basically a full bisq version cleanup tool rather than a purpose built cleanup tool as suggested by @devinbileck

devinbileck:
Fair enough. My idea was to have a utility that could purge any trade matching a given criteria. Which could be used to resolve this specific issue and any that arise in the future. I agree it is a little too ambitious at this point requiring implementation and maintenance time. I suppose providing a build of this branch for those that require it should be sufficient at this point.

a123b:
~~The faulty trades also need to be purged from the history tab in case the trade already got closed by an arbitrator, see https://github.com/bisq-network/bisq/issues/2925#issuecomment-507079829~~ Just realized they already do

Don't know if it's not too much work, but how about showing a popup like "Repairs in data directory completed. Please switch to the standard Bisq version now." that closes the app if acknowledged to prevent users from continuing to use the special build?

sqrrm:
@devinbileck That's a good idea, having a separate tool to clean the database selectively. Could be useful for development and testing too. Something to keep in mind going forward that when in doubt, it might be worth building rather than doing some hack to test stuff.

@a123b Perhaps I should add a popup and force shut down after the startup process, including the cleanup, is completed. That way it's not possible to use this version for trading but only as a cleanup tool.

ripcurlx (approved):
utACK

PR #2939, opened 2019-07-06
Bundle of envelopes

freimair:
Seeing an oszillating pattern of broadcast message numbers in [monitoring](https://monitor.bisq.network/d/wTDl6T_iz/p2p-stats?refresh=30s&orgId=1), @ManfredKarrer and I suspected that the pattern might be caused by the DoS protection mechanisms in place.

This PR introduces a new message that containing a number of existing messages. The hopes are:
- less connection setup overhead on the network level leading to less network load and hopefully better performance
- not to spread the bursts of e.g. RefreshOffer messages over a long time

In case this change has a bad effect on network stability (although we do not believe it is going to) we can simply remove the newly created `Capability` in a future release and thus, quickly get rid of the feature.

I tested the code extensively by running
- 4 nodes locally
- they knew each other as seed nodes
- one of the nodes was connected to the live bisq network
- had this setup running for days for each iteration

### DIFF
--- a/common/src/main/proto/pb.proto
+++ b/common/src/main/proto/pb.proto
@@ -66,12 +66,18 @@ message NetworkEnvelope {
         NewBlindVoteStateHashMessage new_blind_vote_state_hash_message = 40;
         GetBlindVoteStateHashesRequest get_blind_vote_state_hashes_request = 41;
         GetBlindVoteStateHashesResponse get_blind_vote_state_hashes_response = 42;
+
+        EnvelopeOfEnvelopes envelope_of_envelopes = 43;
     }
 }
 
 ///////////////////////////////////////////////////////////////////////////////////////////
 // Implementations of NetworkEnvelope
 ///////////////////////////////////////////////////////////////////////////////////////////
+
+message EnvelopeOfEnvelopes {
+    repeated NetworkEnvelope envelopes = 1;
+}
 
 // get data
 
--- a/common/src/main/proto/pb.proto
+++ b/common/src/main/proto/pb.proto
@@ -67,7 +67,7 @@ message NetworkEnvelope {
         GetBlindVoteStateHashesRequest get_blind_vote_state_hashes_request = 41;
         GetBlindVoteStateHashesResponse get_blind_vote_state_hashes_response = 42;
 
-        EnvelopeOfEnvelopes envelope_of_envelopes = 43;
+        BundleOfEnvelopes bundle_of_envelopes = 43;
     }
 }
 
@@ -75,7 +75,7 @@ message NetworkEnvelope {
 // Implementations of NetworkEnvelope
 ///////////////////////////////////////////////////////////////////////////////////////////
 
-message EnvelopeOfEnvelopes {
+message BundleOfEnvelopes {
     repeated NetworkEnvelope envelopes = 1;
 }
 
--- a/p2p/src/main/java/bisq/network/p2p/network/Connection.java
+++ b/p2p/src/main/java/bisq/network/p2p/network/Connection.java
@@ -72,8 +72,10 @@ import java.io.StreamCorruptedException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.Queue;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.CopyOnWriteArraySet;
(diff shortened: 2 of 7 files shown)


common/src/main/proto/pb.proto
```diff
@@ -66,13 +66,19 @@ message NetworkEnvelope {
         NewBlindVoteStateHashMessage new_blind_vote_state_hash_message = 40;
         GetBlindVoteStateHashesRequest get_blind_vote_state_hashes_request = 41;
         GetBlindVoteStateHashesResponse get_blind_vote_state_hashes_response = 42;
+
```
sqrrm, line 69:
```suggestion
```

p2p/src/main/java/bisq/network/p2p/network/Connection.java
```diff
@@ -280,6 +277,48 @@ public void sendMessage(NetworkEnvelope networkEnvelope) {
...
+
+                        // check if BundleOfEnvelopes is supported
+                        if(getCapabilities().containsAll(new Capabilities(Capability.ENVELOPE_OF_ENVELOPES))) {
+                            synchronized (lock) {
+                                // check if current envelope fits size
+                                // - no? create new envelope
+                                if(queueOfBundles.isEmpty() || queueOfBundles.element().toProtoNetworkEnvelope().getSerializedSize() + networkEnvelope.toProtoNetworkEnvelope().getSerializedSize() > MAX_PERMITTED_MESSAGE_SIZE * 0.9) {
+                                    // - no? create a bucket
+                                    queueOfBundles.add(new BundleOfEnvelopes());
+                                    System.err.println("added fresh container");
```
sqrrm: Why print?  Debug cruft?

p2p/src/main/java/bisq/network/p2p/network/Connection.java
```diff
@@ -280,6 +277,48 @@ public void sendMessage(NetworkEnvelope networkEnvelope) {
...
+
+                                    // - and schedule it for sending
+                                    lastSendTimeStamp += sendMsgThrottleSleep;
```
sqrrm, line 299: This will distort the lastSendTimeStamp by setting it to the future. Won't this add cumulatively more time to lastSendTimeStamp if there are multiple messages that need to be sent in a short times pan?

p2p/src/main/java/bisq/network/p2p/network/Connection.java
```diff
@@ -393,7 +432,13 @@ private boolean violatesThrottleLimit(NetworkEnvelope networkEnvelope) {
     @Override
     public void onMessage(NetworkEnvelope networkEnvelope, Connection connection) {
         checkArgument(connection.equals(this));
-        UserThread.execute(() -> messageListeners.forEach(e -> e.onMessage(networkEnvelope, connection)));
+
+        if(networkEnvelope instanceof BundleOfEnvelopes)
```
sqrrm: Style is off (missing space). Best to use some tool that handles that automatically.

sqrrm (commented):
Seems like it's a bit risky, but not a bad idea. Need a bit more time to think on it.

p2p/src/main/java/bisq/network/p2p/network/Connection.java
freimair: yes it will. that is exactly the point. if we handle it otherwise, we would be left with no DoS protection at all. Besides, these messages are not lost but send as a bundle. I tested it in the real Bisq network with 4 nodes with testing conditions to only send every 2 seconds. Yes, my queue once went up to 27 bundles to be sent but went down again as soon as traffic slowed.

p2p/src/main/java/bisq/network/p2p/network/Connection.java
freimair: well :). PR updated.

p2p/src/main/java/bisq/network/p2p/network/Connection.java
freimair: ahm. That is what Intellij does (is that the tool you are referring to?). Besides, I do not spot the issue

common/src/main/proto/pb.proto
freimair: NACK. The change has nothing to do with any DAOStateHashes.

ripcurlx:
Travis already ran through, but it is still showing pending. I'll merge it anyway.
@sqrrm If you find any issues with this implementation I'll revert this changes before the next release.

PR #2945, opened 2019-07-09
Update command to generate blocks

huey735:
Specify that from Bitcoin Core's v.0.18 the 'generate' command will be
deprecated and to use 'generatetoaddress' instead.


ripcurlx (commented):
utACK

PR #2947, opened 2019-07-09
Update CurrencyUtil.java

ghost:
PR similar to previous https://github.com/bisq-network/bisq/pull/2680
add DAC + RDD to the list of removed assets.
Should solve #2189 for those assets
I built the code and it seems to work.


ripcurlx (approved):
utACK

PR #2951, opened 2019-07-11
Don't dump blockchain data during initial parsing

sqrrm:
Avoid dumping blockchain after every parsed block during initial parsing, it's really slow.


devinbileck (approved):
utACK

PR #2952, opened 2019-07-12
Envelope of envelopes #2

freimair:
- Renamed the Capability to match the actual Feature name
- only send a 'BundleOfEnvelopes' if necessary


ripcurlx (approved):
utACK

PR #2963, opened 2019-07-14
Initial block sync fixes

sqrrm:
The fix to continue requesting blocks is essential if we are to go ahead with the limit of 6000 blocks max per request.


ripcurlx (approved):
utACK

PR #2941, opened 2019-07-07
List Know Your Developer (KYD)

Bob-Thomas:
- Official project URL: https://kydcoin.io
- Official block explorer URL: https://explorer.kydcoin.io

PR #2964, opened 2019-07-15
Update KYD to KYDC listing to be conform ISO_4217

Bob-Thomas:
Update for the KYD listing #2941


ripcurlx:
I banned KYD from trading right now by using the developer filter ([conflict with existing currency symbol](https://docs.bisq.network/exchange/howto/list-asset.html#the-asset-ticker-symbol-must-not-conflict-with-national-currency-tickers)) to prevent any issues with existing offers when the currency code changes. As soon as the next release will be published KYDC will be available for trading again.

ripcurlx (approved):
ACK

PR #1657, opened 2018-09-07
List Neos (NEOS)

BoubonicNugz:
- Re-Submit
- Previous PR: https://github.com/bisq-network/bisq-assets/pull/117

Project Information:
- Official project URL: https://www.neos.sh
- Official block explorer URL: https://chainz.cryptoid.info/neos/
- Official Github Source:  https://github.com/neoscoin/neos-core/

PR #2977, opened 2019-07-18
Rebrand Neos to Donu

ksynb:
Project Information:

* Official project URL: https://www.donu.io
* Official block explorer URL: https://chainz.cryptoid.info/donu/
* Official Github Source: https://github.com/donito-inc/donu-core

### DIFF
--- a/assets/src/main/java/bisq/asset/coins/Donu.java
+++ b/assets/src/main/java/bisq/asset/coins/Donu.java
@@ -22,17 +22,17 @@ import bisq.asset.Base58BitcoinAddressValidator;
 import bisq.asset.Coin;
 import bisq.asset.NetworkParametersAdapter;
 
-public class Neos extends Coin {
+public class Donu extends Coin {
 
-    public Neos() {
-        super("Neos", "NEOS", new NeosAddressValidator());
+    public Donu() {
+        super("Donu", "DONU", new DonuAddressValidator());
     }
 
 
-    public static class NeosAddressValidator extends Base58BitcoinAddressValidator {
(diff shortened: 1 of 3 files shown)


blabno:
NEOS has been submitted in this PR https://github.com/bisq-network/bisq/pull/1657 by @BoubonicNugz. 

@BoubonicNugz could you confirm this is legit re-branding?

blabno:
The code looks ok, although I'm not sure how it will influence existing offers.
Wouldn't it be safer to delist NEOS and add Donu as a new asset?

buzztiaan:
the address formats are the same and all users know it moved, but i can see your motivation yeah
there's no open offers at the moment

BoubonicNugz:
ACK'd

This is a legit re-branding.  @ksynb is the lead developer of the project.

blabno:
@ManfredKarrer @ripcurlx Do you think it's safe to accept this re-brand with regards to potential existing offers?

sqrrm:
Regarding rebranding, I think we might want to discuss a bit and have some kind of template on how to handle that. Might be best to just add it as a new asset for any rebranding.

ksynb:
> Regarding rebranding, I think we might want to discuss a bit and have some kind of template on how to handle that. Might be best to just add it as a new asset for any rebranding.

That would be reasonable. I know folks in the community would appreciate it being back up and available for trade. With the current PR containing required details, is it possible for inclusion in the next update or would you prefer a new PR is created with the intent specified as initial inclusion?

blabno (approved):
Since we've got a lot of other topics queued for next few dev calls and taking into consideration that the developer who initially create PR to list this asset with previous name I'm giving this ACK.

PR #2983, opened 2019-07-21
make rpc host configurable

mrosseel:
What: 
* make the rpc host configurable
* rename blockNotifyPort to rpcBlockPort
* add information to log text on RpcService startup

Why: 
when using docker (or different servers), the bitcoin full node and seednode run on different ip's.


ripcurlx (commented):
utACK

PR #2999, opened 2019-07-24
rename Clock to ClockWatcher

christophsturm:
reasons for the change: 
* this class is not a clock but it watches the clock, detects standby and runs periodic tasks.
* there is already a jdk method called clock

First i thought it should be called PeriodicTaskManager, now i find
ClockWatcher more fitting.


christophsturm:
@sqrmm @ripcurlx  can we get this one merged, then i can start injecting clocks to fix unit tests on jdk 11 and 12.

sqrrm (approved):
utACK

PR #3000, opened 2019-07-24
fix a generics warning and some idea autofixes

christophsturm:


christophsturm:
I just noticed that the diff is much bigger than it should be, probably because i have "rearrange code" checked in my idea git dialog. is rearrange code not something that should be use?

ManfredKarrer:
> is rearrange code not something that should be use?

We never used it. I tried it out sometimes but I was not happy with its decisions....

Organize Imports is mandatory to be used and defined in the projects IDEA settings.

christophsturm:
ok i will keep rearrange code off from now on.

sqrrm:
@christophsturm I should probably have merged the PRs in order, now there is a conflict. Looks reasonable.

christophsturm:
I'll just redo this PR without rearrange code.

PR #3003, opened 2019-07-25
Avoid validating proposals during initial sync

sqrrm:


ManfredKarrer (approved):
ACK. I tested a full node with resync from genesis and it came to the same daostate hash as the seed nodes. Without that PR it did not.

PR #3005, opened 2019-07-25
split the pb.java file into multiple files

christophsturm:
the pb.java file was making idea crazy.

since we have many protobuf classes that have the same name as our own classes I changed the protobuf package to "protobuf" to have a short way to reference it via its whole package in case of a naming collision.


sqrrm (approved):
ACK

PR #3030, opened 2019-07-30
Cherry pick non ui commits from #2872

christophsturm:
to get the code into master.  see #2872

### DIFF
--- a/core/src/main/java/bisq/core/setup/CoreNetworkCapabilities.java
+++ b/core/src/main/java/bisq/core/setup/CoreNetworkCapabilities.java
@@ -29,7 +29,7 @@ import lombok.extern.slf4j.Slf4j;
 public class CoreNetworkCapabilities {
     public static void setSupportedCapabilities(BisqEnvironment bisqEnvironment) {
         Capabilities.app.addAll(Capability.TRADE_STATISTICS, Capability.TRADE_STATISTICS_2, Capability.ACCOUNT_AGE_WITNESS, Capability.ACK_MSG);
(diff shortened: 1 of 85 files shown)


christophsturm:
@sqrrm like we discussed on slack

core/src/main/java/bisq/core/setup/CoreNetworkCapabilities.java
```diff
@@ -29,7 +29,7 @@
 public class CoreNetworkCapabilities {
     public static void setSupportedCapabilities(BisqEnvironment bisqEnvironment) {
         Capabilities.app.addAll(Capability.TRADE_STATISTICS, Capability.TRADE_STATISTICS_2, Capability.ACCOUNT_AGE_WITNESS, Capability.ACK_MSG);
-        Capabilities.app.addAll(Capability.BUNDLE_OF_ENVELOPES);
+        Capabilities.app.addAll(Capability.BUNDLE_OF_ENVELOPES,Capability.SIGNED_ACCOUNT_AGE_WITNESS);
```
christophsturm, line 32: not sure if that line is safe. maybe not add the capability for now?

sqrrm: No issues, this is the way to extend the protocol. We're saying we handle `SIGNED_ACCOUNT_AGE_WITNESS` but no one is going to ask for it until we add the code actually using it.

sqrrm (approved):
utACK

PR #3034, opened 2019-07-30
improve the signedwitnessservicetest

christophsturm:
Tests should contain no conditional logic (no `if`s) and be self contained (not call private methods)
that way the test is easier to read without context switches.


sqrrm:
The `protobuf` commit message doesn't seem relevant

sqrrm:
Looks reasonable, but there seems to be many extra empty lines

sqrrm (approved):
utACK

PR #3037, opened 2019-07-30
inject clock for usage in isDateInTolerance

christophsturm:


mrosseel:
Clock doesn't seem to be used yet, are those commits coming?

christophsturm:
i have a different PR that adds unit tests for the `SignedWitness` class. once this is merged i will use the clock for more unit testing. I just wanted to do the clock injection in a separate pr to make prs small and easy to review and not block a bigger pr if someone has an objection.

sqrrm:
For these kinds of PRs it would be good to have an explanation of why they're needed. It's not obvious why this one is needed without the follow up conversation.

sqrrm (approved):
utACK

christophsturm:
> For these kinds of PRs it would be good to have an explanation of why they're needed. It's not obvious why this one is needed without the follow up conversation.

ok.

Issue #3036, opened 2019-07-30
add unit tests for guice setup


christophsturm:
there should be unit tests covering the guice wiring for all the different app setups. This should be not hard to do

PR #3041, opened 2019-08-01
Move Clock binding to P2PModule

sqrrm:
To make seednode happy


christophsturm:
sorry for the breakage. i did write an issue for guice testing because i was worried about different setups https://github.com/bisq-network/bisq/issues/3036 but then i thought that everything probably depends on core

Issue #2873, opened 2019-06-06
Support of Torv3 (Required for BTCPay Server)


NicolasDorier:
I recently added Bisq to the supported software to connect to your BTCPay hosted node.

![image](https://user-images.githubusercontent.com/3020646/59045298-c01e0c00-88ba-11e9-90b1-90c3642bf2d7.png)

I tried it today, and sadly it does not work. 

BTCPay use Tor version 3 for exposing the full node. 

Tor v2 is not adapted to connect to your own node because you want your own BTC node to whitelist any connection using the Tor address to guarantee having a connection slot available.

But if your Tor is version 2, then your address become discoverable and anybody could connect to it and DDoS you.

```
Jun-07 00:23:18.233 [JavaFX Application Thread] INFO  b.c.b.s.WalletsSetup: Socks5Proxy for bitcoinj: socks5Proxy=sampleapi.example.com:61780	Version 5 
Jun-07 00:23:18.256 [JavaFX Application Thread] INFO  o.b.c.Context: Creating bitcoinj 0.14.7.bisq.1-SNAPSHOT context. 
Jun-07 00:23:18.263 [JavaFX Application Thread] ERROR b.c.s.CommonSetup: Uncaught Exception from thread JavaFX Application Thread 
Jun-07 00:23:18.263 [JavaFX Application Thread] ERROR b.c.s.CommonSetup: throwableMessage= Invalid hostname: ######################.onion 
Jun-07 00:23:18.264 [JavaFX Application Thread] ERROR b.c.s.CommonSetup: throwableClass= class java.lang.IllegalArgumentException 
Jun-07 00:23:18.273 [JavaFX Application Thread] ERROR b.c.s.CommonSetup: Stack trace:
java.lang.IllegalArgumentException: Invalid hostname: ######################.onion
	at org.bitcoinj.net.OnionCatConverter.onionHostToIPV6Bytes(OnionCatConverter.java:28)
	at org.bitcoinj.net.OnionCatConverter.onionHostToInetAddress(OnionCatConverter.java:46)
	at bisq.core.btc.nodes.BtcNodeConverter$Facade.onionHostToInetAddress(BtcNodeConverter.java:130)
	at bisq.core.btc.nodes.BtcNodeConverter.convertOnionHost(BtcNodeConverter.java:61)
	at bisq.core.btc.nodes.BtcNodesRepository.lambda$getOnionHosts$1(BtcNodesRepository.java:74)
```

[...]

PR #3044, opened 2019-08-01
Backup, restore and create a new onion address via GUI

### LINKED ISSUE #909: Privacy enhancement request.
Hello,
Currently, if you trade fiat-btc and later btc-crypto trade with the same person/onion.address, the btc/crypto trade is not anonymous because of the prior trade, the other person knows who you are.

Work around is you can choose to not to transact with this person or launch another bisq instance, but this is inconvenient, due to the extra hassle & the low volume.

Not sure if this is on the radar to separate somehow to enhance privacy.

I hope i'm not imagining things, i think this was a case, i noticed this some time back, but forgot about it till now.

### LINKED ISSUE #1055: Export/import feature for onion address
To keep one's local reputation if changing the data directory an export/import feature for the onion address would be nice to have. It simply requires hostname and private_key files from the hidden service directory.

### LINKED ISSUE #1056: Feature for renewing the onion address
For gaining more privacy a feature for renewing the onion address would be nice to have.
It would simply delete (backup first) the filed in the hidden service directory. It would require some checks that the user has no open offer, trades or disputes and that he understands that he loses his local reputation by changing the onion address.

freimair:
This PR adds features that allow the user to 
- backup her current onion address,
- restore a previously onion address backup
- and even create a completely new onion address

all from within the GUI.

**How do you make sure that there are no more open offers or even ongoing trades before you import/renew the address?**
I don't. I have reworked the P2P backend so that a single client can have multiple hidden services up and running at any time. If the user renews/imports an onion address, a new hidden service is started. Hidden services are only removed once no more open offers or ongoing trades are there (checked for after every completed trade).

**All during runtime?**
No. A restart is required for the import and renew features.

Contains DE translations as well.

Here are some impressions:
Screenshot from 2019-08-01 13-28-50
Screenshot from 2019-08-01 13-29-57
Screenshot from 2019-07-30 12-09-40

**Expected behaviour (for testers)**
- after renew/import onion address and an application restart, there should be one more hidden service alive (except when you import an address that you already had)
- the hidden services will only be removed when no more trades and offers use an address. The decision which hidden service to remove is made when a trade completes (successful or not). If an address is not used anymore it is marked for removal and will not show up after an application restart.

closes #1055, closes #1056, subsequently closes #909 and does take care of a lot of preparation work for #2873.


core/src/main/java/bisq/core/trade/TradeManager.java
```diff
@@ -543,6 +544,14 @@ public void removePreparedTrade(Trade trade) {
 
     private void removeTrade(Trade trade) {
         tradableList.remove(trade);
+
+        // collect all node addresses (hidden services) we still need for active offers/trades
```
christophsturm: isnt that exactly the same code as above?

freimair: in OpenOfferManager you mean? yes, somewhat.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -97,6 +98,8 @@
     private final P2PService p2PService;
     private final BtcWalletService btcWalletService;
     private final TradeWalletService tradeWalletService;
+    @Inject
+    private TradeManager tradeManager;
```
ManfredKarrer: Why not use constructor injection? It would be final then and it would be more in sync with current code style.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -492,14 +495,29 @@ private void onRemoved(@NotNull OpenOffer openOffer, ResultHandler resultHandler
...
+        openOffers.remove(offer);
+
+        // extract my node address
+        NodeAddress myNodeAddress = offer.getOffer().getOfferPayload().getOwnerNodeAddress();
+
+        // check
+        if(openOffers.stream().noneMatch(openOffer -> openOffer.getOffer().getOfferPayload().getOwnerNodeAddress().equals(myNodeAddress))) {
+            // check ongoing trades
+            if(tradeManager.getTradableList().stream().noneMatch(trade -> trade.getContract().getBuyerNodeAddress().equals(myNodeAddress) || trade.getContract().getSellerNodeAddress().equals(myNodeAddress)))
+                        p2PService.reportUnusedNodeAddress(myNodeAddress);
```
ManfredKarrer: I think here is missing the check for your own role. You are adding both trade peers here.

trade.getContract() can be null, so the code should be safe for that case.

```
result.addAll(tradeManager.getTradableList().stream()
                .map(Trade::getContract)
                .filter(Objects::nonNull)
                .map(contract -> {
                    if (contract.isBuyerMakerAndSellerTaker()) {
                        return contract.getBuyerNodeAddress();
                    } else {
                        return contract.getSellerNodeAddress();
                    }
                })
                .collect(Collectors.toSet()));
```
will do it.
We should rename tradeManager.getTradableList() to tradeManager.getTradesList(). Tradeable can be also an OpenOffer...

core/src/main/java/bisq/core/trade/TradeManager.java
```diff
@@ -543,6 +543,22 @@ public void removePreparedTrade(Trade trade) {
...
+            myNodeAddress = trade.getContract().getSellerNodeAddress();
+        else
+            myNodeAddress = trade.getContract().getBuyerNodeAddress();
+
+        // check ongoing trades
+        if (tradableList.stream().noneMatch(currentTrade -> currentTrade.getContract().getSellerNodeAddress().equals(myNodeAddress) || currentTrade.getContract().getBuyerNodeAddress().equals(myNodeAddress))) {
+            // check open offers
+            if (openOfferManager.getObservableList().stream().noneMatch(openOffer -> openOffer.getOffer().getOfferPayload().getOwnerNodeAddress().equals(myNodeAddress)))
+                p2PService.reportUnusedNodeAddress(myNodeAddress);
+        }
```
ManfredKarrer: Same code as in OpenOfferManager. Better use one method for both.

desktop/src/main/java/bisq/desktop/main/settings/network/NetworkSettingsView.java
```diff
@@ -302,6 +307,8 @@ public void deactivate() {
 
         useTorForBtcJCheckBox.setOnAction(null);
 
+        renewIdButton.setOnAction(null);
```
ManfredKarrer, line 382: Missing the setOnAction(null); on the other 2 buttons.

p2p/src/main/java/bisq/network/p2p/network/TorMode.java
```diff
@@ -49,7 +49,7 @@
      *                         sits in. Note that, due to the inner workings of the
      *                         <code>Netlayer</code> dependency, it does not
      *                         necessarily equal
```
ManfredKarrer: @param hiddenServiceDir is not used anymore. Can you remove it?

p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
```diff
@@ -315,35 +315,37 @@ private void createTor(TorMode torMode) {
...
-                log.info("It will take some time for the HS to be reachable (~40 seconds). You will be notified about this");
-                return nodeAddress;
+                }.start();
+            } catch (final Exception e) {
+                log.error(e.toString());
+                e.printStackTrace();
+            }
+            return null;
+        });
+        hiddenServiceSocket.addReadyListener(hiddenServiceSocket1 -> {
```
ManfredKarrer, line 508: Why add an new listender and not use the above one?

desktop/src/main/java/bisq/desktop/main/settings/network/NetworkSettingsView.java
```diff
@@ -293,6 +294,27 @@ public void activate() {
                 p2PService.exportHiddenService(file);
         });
 
+        importIdButton.setOnAction(event -> {
+            new Popup<>().information(Res.get("settings.net.importAddress"))
+                    .actionButtonText(Res.get("shared.applyAndShutDown"))
```
ManfredKarrer: Button text is confusing. Should be something like: "Select backup file"

ManfredKarrer:
If I make a backup of first HS and later restore that I have 3 HS where first and third are the same but have diff. dir names (fist is '0'). Leads to a timeout then... I think very first HS should also use timestamp as directory name. And seems handling of the original in case of restore is not correct.

ManfredKarrer:
On OSX there is a  .DS_Store file and caused an exception: 
Aug-14 20:39:07.081 [JavaFX Application Thread] ERROR b.n.p2p.network.TorNetworkNode: Hidden service creation failed: java.io.IOException: Not a directory

ManfredKarrer (changes requested):
NACK

freimair:
> If I make a backup of first HS and later restore that I have 3 HS where first and third are the same but have diff. dir names (fist is '0'). Leads to a timeout then...

is fixed in https://github.com/bisq-network/bisq/pull/3044/commits/b4818d8cf552999f081fcdd282441822c17e99f2

> On OSX there is a .DS_Store file and caused an exception:

that is a bug. onto it...

ripcurlx:
@freimair The build of your last commit failed on travis.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
freimair: TradeManager and OpenOfferManager each have to know each other when instantiated. Even guice cannot resolve that. Having guice inject the TradeManager a tad later resolves the circular dependency and things work again. I am not happy with that, but that is the best we can do.

freimair:
@ripcurlx yes I know and I have not figured out why suddenly, the desktop app tests start to fail. I am on to adding tests and get to the ground of that.

Update: Travis tests the merge (that is what surprised me). Recently, the test failing was introduced in master -> that is why it only recently showed up. The code works fine but the test fails as soon as something concerned about the TradeManager is at stake. Looking into resolving that circular dependency...

core/src/main/java/bisq/core/offer/OpenOfferManager.java
christophsturm: can we break up the circularity maybe by introducing a new component?

ManfredKarrer: Oh ok. Please don't do that. We are just hiding a circular dependency that way and we should rather avoid it s it shows a wrong design.
I think in that case its anyway a code duplication issue and that should move out to probably a 3rd class. Need to look closer for more details...

core/src/main/java/bisq/core/offer/OpenOfferManager.java
freimair: ad missing role check: well, it is not that easy to determine what the clients role is at that point. Your code suggestion only works for trades you created. It fails if you took an offer from someone else. If one is doing the check thoroughly, it is much more complex, also because necessary info may not be provided (null) in certain cases. The approach now is to add any and all onion addresses that are anywhere related to the cause and prevent them from being deleted. Plus, even if a HS is wrongly kept once, it may be released in the future without any further calls and checks and complexity in the code.

ad null-check: fixed.

ad rename: I haven't touched the names as there is no Javadoc that explains what is what.

ManfredKarrer: Ah, yes you are right, OpenOfferManager handles it differently than tradeManager. But for OpenOffers my suggestion is correct as you are the maker. But for trades we need to use the pubKey to detect which role is ours. A util method in Contract might be good as that is used anyway at other pleces as well (arbitration as far I remember).

p2p/src/main/java/bisq/network/p2p/network/TorNetworkNode.java
freimair: because of readability. separation of concerns.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
freimair: well, that it not so easy. It is definitely not solvable with best-practice Java. Here is some food-for-thought: 
Screenshot from 2019-08-20 15-06-27
(black is status, green works out of the box, red requires dependency, the grey thought experiment introduces 2 circular dependencies). All in all, some devilish trickery seems due.

Or: What if we **only check** for EOL hidden services **on trade complete**? Bisq is about trading and not about cancelling offers. The HS might live longer without being needed but does that hurt? Eventually a trade completes and thus, the HS is cleaned up.

freimair: I changed the code so that hidden services are only cleaned up when a trade somehow completes.

ManfredKarrer: With a 3rd component as listener for both open offer and trades it works:
Here is a patch:

```
Index: core/src/main/java/bisq/core/RemoveTradeOrOpenOfferListener.java
IDEA additional info:
<+>UTF-8
--- core/src/main/java/bisq/core/RemoveTradeOrOpenOfferListener.java	(date 1566337782000)
+++ core/src/main/java/bisq/core/RemoveTradeOrOpenOfferListener.java	(date 1566337782000)
@@ -0,0 +1,88 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * License for more details.
+ *
+ */
+
+package bisq.core;
+
+import bisq.core.offer.OpenOfferManager;
+import bisq.core.trade.Trade;
+import bisq.core.trade.TradeManager;
+
+import bisq.network.p2p.NodeAddress;
+import bisq.network.p2p.P2PService;
+
+import bisq.common.crypto.PubKeyRing;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+
+@Slf4j
+@Singleton
+public class RemoveTradeOrOpenOfferListener implements TradeManager.RemoveTradeListener, OpenOfferManager.RemoveOpenOfferListener {
+
+    private final TradeManager tradeManager;
+    private final OpenOfferManager openOfferManager;
+    private final P2PService p2PService;
+    private final PubKeyRing pubKeyRing;
+
+    @Inject
+    public RemoveTradeOrOpenOfferListener(TradeManager tradeManager,
+                                          OpenOfferManager openOfferManager,
+                                          P2PService p2PService,
+                                          PubKeyRing pubKeyRing) {
+        this.tradeManager = tradeManager;
+        this.openOfferManager = openOfferManager;
+        this.p2PService = p2PService;
+        this.pubKeyRing = pubKeyRing;
+
+        tradeManager.addRemoveTradeListener(this);
+        openOfferManager.addRemoveOpenOfferListener(this);
+    }
+
+    @Override
+    public void onRemoveTrade() {
+        reportRequiredHiddenServices();
+    }
+
+
+    @Override
+    public void onRemoveOpenOffer() {
+        reportRequiredHiddenServices();
+
+    }
+
+    private void reportRequiredHiddenServices() {
+        // collect all node addresses (hidden services) we still need for active offers/trades
+        Set<NodeAddress> result = new HashSet<>();
+        result.addAll(openOfferManager.getOpenOffers().stream()
+                .map(openOffer -> openOffer.getOffer().getOfferPayload().getOwnerNodeAddress())
+                .collect(Collectors.toSet()));
+        result.addAll(tradeManager.getTradableList().stream()
+                .map(Trade::getContract)
+                .filter(Objects::nonNull)
+                .map(contract -> contract.getMyNodeAddress(pubKeyRing))
+                .collect(Collectors.toSet()));
+
+        p2PService.reportRequiredHiddenServices(result);
+    }
+}
Index: core/src/main/java/bisq/core/app/BisqSetup.java
IDEA additional info:
<+>UTF-8
--- core/src/main/java/bisq/core/app/BisqSetup.java	(date 1566334741000)
+++ core/src/main/java/bisq/core/app/BisqSetup.java	(date 1566337886000)
@@ -17,6 +17,7 @@
 
 package bisq.core.app;
 
+import bisq.core.RemoveTradeOrOpenOfferListener;
 import bisq.core.account.witness.AccountAgeWitnessService;
 import bisq.core.alert.Alert;
 import bisq.core.alert.AlertManager;
@@ -154,6 +155,8 @@
     private final AssetService assetService;
     private final TorSetup torSetup;
     private final TradeLimits tradeLimits;
+    // Need to keep a reference for RemoveTradeOrOpenOfferListener
+    private final RemoveTradeOrOpenOfferListener removeTradeOrOpenOfferListener;
     private final BSFormatter formatter;
     @Setter
     @Nullable
@@ -232,6 +235,7 @@
                      AssetService assetService,
                      TorSetup torSetup,
                      TradeLimits tradeLimits,
+                     RemoveTradeOrOpenOfferListener removeTradeOrOpenOfferListener,
                      BSFormatter formatter) {
 
 
@@ -272,6 +276,7 @@
         this.assetService = assetService;
         this.torSetup = torSetup;
         this.tradeLimits = tradeLimits;
+        this.removeTradeOrOpenOfferListener = removeTradeOrOpenOfferListener;
         this.formatter = formatter;
     }
 
Index: core/src/main/java/bisq/core/offer/OpenOfferManager.java
IDEA additional info:
<+>UTF-8
--- core/src/main/java/bisq/core/offer/OpenOfferManager.java	(date 1566334741000)
+++ core/src/main/java/bisq/core/offer/OpenOfferManager.java	(date 1566337963000)
@@ -78,6 +78,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import lombok.Getter;
+
 import org.jetbrains.annotations.NotNull;
 
 import javax.annotation.Nullable;
@@ -87,6 +89,10 @@
 public class OpenOfferManager implements PeerManager.Listener, DecryptedDirectMessageListener, PersistedDataHost {
     private static final Logger log = LoggerFactory.getLogger(OpenOfferManager.class);
 
+    public interface RemoveOpenOfferListener {
+        void onRemoveOpenOffer();
+    }
+
     private static final long RETRY_REPUBLISH_DELAY_SEC = 10;
     private static final long REPUBLISH_AGAIN_AT_STARTUP_DELAY_SEC = 30;
     private static final long REPUBLISH_INTERVAL_MS = TimeUnit.MINUTES.toMillis(40);
@@ -108,7 +114,9 @@
     private final Map<String, OpenOffer> offersToBeEdited = new HashMap<>();
     private boolean stopped;
     private Timer periodicRepublishOffersTimer, periodicRefreshOffersTimer, retryRepublishOffersTimer;
+    @Getter
     private TradableList<OpenOffer> openOffers;
+    private List<RemoveOpenOfferListener> removeOpenOfferListeners = new ArrayList<>();
 
 
@@ -479,6 +487,7 @@
         }
     }
 
+    //todo
     private void onRemoved(@NotNull OpenOffer openOffer, ResultHandler resultHandler, Offer offer) {
         offer.setState(Offer.State.REMOVED);
         openOffer.setState(OpenOffer.State.CANCELED);
@@ -487,6 +496,8 @@
         log.info("onRemoved offerId={}", offer.getId());
         btcWalletService.resetAddressEntriesForOpenOffer(offer.getId());
         resultHandler.handleResult();
+
+        removeOpenOfferListeners.forEach(RemoveOpenOfferListener::onRemoveOpenOffer);
     }
 
     // Close openOffer after deposit published
@@ -497,6 +508,8 @@
             offerBookService.removeOffer(openOffer.getOffer().getOfferPayload(),
                     () -> log.trace("Successful removed offer"),
                     log::error);
+
+            removeOpenOfferListeners.forEach(RemoveOpenOfferListener::onRemoveOpenOffer);
         });
     }
 
@@ -792,6 +805,10 @@
         startPeriodicRepublishOffersTimer();
     }
 
+    public void addRemoveOpenOfferListener(RemoveOpenOfferListener removeOpenOfferListener) {
+        removeOpenOfferListeners.add(removeOpenOfferListener);
+    }
+
 
     // Private
Index: core/src/main/java/bisq/core/trade/Contract.java
IDEA additional info:
<+>UTF-8
--- core/src/main/java/bisq/core/trade/Contract.java	(date 1566334741000)
+++ core/src/main/java/bisq/core/trade/Contract.java	(date 1566337963000)
@@ -193,6 +193,13 @@
         return isBuyerMakerAndSellerTaker ? makerPubKeyRing : takerPubKeyRing;
     }
 
+    public NodeAddress getMyNodeAddress(PubKeyRing myPubKeyRing) {
+        if (myPubKeyRing.equals(getBuyerPubKeyRing()))
+            return buyerNodeAddress;
+        else
+            return sellerNodeAddress;
+    }
+
     public PubKeyRing getSellerPubKeyRing() {
         return isBuyerMakerAndSellerTaker ? takerPubKeyRing : makerPubKeyRing;
     }
Index: core/src/main/java/bisq/core/trade/TradeManager.java
IDEA additional info:
<+>UTF-8
--- core/src/main/java/bisq/core/trade/TradeManager.java	(date 1566334741000)
+++ core/src/main/java/bisq/core/trade/TradeManager.java	(date 1566337963000)
@@ -104,6 +104,11 @@
 public class TradeManager implements PersistedDataHost {
     private static final Logger log = LoggerFactory.getLogger(TradeManager.class);
 
+
+    public interface RemoveTradeListener {
+        void onRemoveTrade();
+    }
+
     private final User user;
     private final KeyRing keyRing;
     private final BtcWalletService btcWalletService;
@@ -130,6 +135,7 @@
     private ErrorMessageHandler takeOfferRequestErrorMessageHandler;
     @Getter
     private final LongProperty numPendingTrades = new SimpleLongProperty();
+    private List<RemoveTradeListener> removeTradeListeners = new ArrayList<>();
 
 
@@ -543,6 +549,7 @@
         cleanUpAddressEntries();
     }
 
+    //todo
     private void removeTrade(Trade trade) {
         tradableList.remove(trade);
 
@@ -553,6 +560,8 @@
         result.addAll(getTradableList().stream().map(Trade::getContract).filter(Objects::nonNull).map(contract -> contract.getSellerNodeAddress()).collect(Collectors.toSet()));
 
         p2PService.reportRequiredHiddenServices(result);
+
+        removeTradeListeners.forEach(RemoveTradeListener::onRemoveTrade);
     }
 
 
@@ -660,4 +669,8 @@
             }
         });
     }
+
+    public void addRemoveTradeListener(RemoveTradeListener removeTradeListener) {
+        removeTradeListeners.add(removeTradeListener);
+    }
 }

```

freimair:
do you want me to change something?

ripcurlx:
> do you want me to change something?

No, I'll review the PR asap.

ManfredKarrer:
@freimair Can you add the code part for correctly filtering the users node address I sent you in the patch? It is same code as used in arbitration and it maked the list containing only the correct addresses even if wrong addressed would not have any effect in your use case.

ripcurlx:
> @freimair Can you add the code part for correctly filtering the users node address I sent you in the patch? It is same code as used in arbitration and it maked the list containing only the correct addresses even if wrong addressed would not have any effect in your use case.

@ManfredKarrer Is there something still open after the last commit from @freimair?

ManfredKarrer:
I don't have time for reviewing again and testing. Also I think @freimair did not implemented the suggestions I made yet. Maybe better to keep it for later as it carries high risk.

battleofwizards:
Sorry for being very late to the party :( This is very loosely related to the PR but maybe will be helpful for the future.

IMO all secret keys should be **derived from the Bitcoin seed**. This includes the private key behind the onion address (or many of them if we want to support that).

This would entirely mitigate the need for multiple separate backups we now expect from the user. It would also make Bisq inline with what Bitcoin users expect from wallets software.

To large extent the standards for this are ready as BIP-39 (+family) and SLIP-s.

freimair:
is there still something I should change?

chimp1984:
The filed inside the zip are executable, original files not.
Files in zip:
```
-rwxr-xr-x   1 user  staff   22 Sep  8 22:40 hostname
-rwxr-xr-x   1 user  staff  874 Sep  8 22:40 private_key
```

Original:
```
-rw-r--r--  1 user  staff   22 Sep  4 10:26 hostname
-rw-r--r--  1 user  staff  874 Sep  4 10:26 private_key
```

chimp1984:
Tested again backup and restore with app with no offers or trades open.

7jkfjawas3o3vu7k was original address. Backed it up, the renewed it. trkdvguyptjajozb was the new one. The I restored 7jkfjawas3o3vu7k. 
Expected behaviour: I have only 7jkfjawas3o3vu7k as HS.
Result: I have both HS....
 
################################################################
Tor hidden service published after 31505 ms. Socked=HiddenServiceSocket[addr=7jkfjawas3o3vu7k.onion,port=9999]
.....

################################################################
Tor hidden service published after 33176 ms. Socked=HiddenServiceSocket[addr=trkdvguyptjajozb.onion,port=9999]

chimp1984:
NACK

freimair:
ad backup and restore: that is expected. unused hidden services are cleaned up once a trade is complete and only after a restart of the application.

as for the executable flags: I do not see this behavior on my linux machine... onto it.

EDIT: zip by itself only stores files, no attributes. There are additional headers for storing such things but whether they are used and read properly strongly depends on the creator/reader software. I have seen different zip file viewers show different file mods on the same zip file. Hence, there is not a whole lot one can do against that.

Besides, it really does not matter here, as the files in the zip-file are text only.

freimair:
is there anything else I should change?

freimair:
is there anything else I should change?

devinbileck:
Is it not possible to also cleanup unused hidden services on shutdown or startup? To prevent the unnecessary HS as mentioned by chimp?

After generating a new address and restarting, it listed its new address as an inbound peer and the old address as an outbound peer (gitfxt...). Listing the old address is likely fine, but it shouldn't list the new address.
image
Perhaps the GUI should explicitly indicate all onion addresses in use?

devinbileck (changes requested):
NACK
While testing on regtest using Tor (i.e. --useLocalhostForP2P=false):
- Given Alice has created an offer and Bob has taken the offer but not initiated payment, and Alice has generated a new address and restarted her client, when Bob starts the payment, Alice does not receive a notification in her client of the payment initiated.

ripcurlx:
> NACK
> While testing on regtest using Tor (i.e. --useLocalhostForP2P=false):
> 
> * Given Alice has created an offer and Bob has taken the offer but not initiated payment, and Alice has generated a new address and restarted her client, when Bob starts the payment, Alice does not receive a notification in her client of the payment initiated.

@freimair Could you please have a look at this case?

ripcurlx (changes requested):
NACK - Please see https://github.com/bisq-network/bisq/pull/3044#pullrequestreview-302929972 and resolve conflict with master

freimair:
I will be working on this one again after we cleaned up the PR backlog...

freimair:
I reworked the whole PR
- did some squashing for a cleaner history
- added functionality to also clean unused hidden services on shutdown
- fixed the connecting-to-self issue
- fixed the issue where messages sent to an "old" hidden service got discarded

@ripcurlx, @devinbileck ready for review again! thanks.

devinbileck:
I verified the connecting-to-self issue and the issue where messages sent to an "old" hidden service got discarded have been resolved.

Regarding cleaning unused hidden services on shutdown, I noticed when generating a new address and restarting (without open trades/offers), both addresses are still in use (as seen in the log). And possibly as a result, it listed a seednode twice in the connected peers list (possibly from each address connecting to it?):
image

On the next restart, the old address is no longer active. So it does take an additional shutdown to clean the unused address. I am not sure if it failed to clean it up initially or if it needs an additional check on startup for cleanup?

Additionally, I still feel the GUI should explicitly indicate all onion addresses in use so it is clear to the user. Though I am not sure the best way to present it since each could potentially have different peer connections. Possibly a future improvement idea?

freimair:
>  I noticed when generating a new address and restarting (without open trades/offers), both addresses are still in use

~hm I have not encountered this case. Will look into it.~
That is expected of course since at the time of shutdown, the hidden service in question is still the main hidden service.

> it listed a seednode twice in the connected peers list 

I have not encountered this case. Will look into it.

> I am not sure if it failed to clean it up initially or if it needs an additional check on startup for cleanup?

if cleaning unused hidden services on startup would be easy, I would have done it right away. The issue is the async way of services getting started - by the time Bisq can assess which HS to keep, all HSs are already up and running (at least most of the time). So no luck here.

> Additionally, I still feel the GUI should explicitly indicate all onion addresses in use...

we could just add a field "legacy onion addresses" with a list of onion addresses still in use, but I am not sure if that can easily be understood by the average Joe. As for the connections, the legacy HS should only receive connections from peers which know the onion address so I see no need for adding an additional connection listing. All in all, I would suggest to get the GUI as is into production and see if and how people actually use it.

devinbileck:
> That is expected of course since at the time of shutdown, the hidden service in question is still the main hidden service.

Makes sense.

> All in all, I would suggest to get the GUI as is into production and see if and how people actually use it.

Sounds good to me.

devinbileck:
I will revisit this and perform additional testing after the upcoming release.

ripcurlx (dismissed):

ripcurlx:
> I will revisit this and perform additional testing after the upcoming release.

@devinbileck Would be great if you could give this a shot for v1.2.5. Thanks!

devinbileck:
To help test this, I have compiled some test scenarios. They may need to be refined, but should be a good starting point. Let me know if there are any additional scenarios that should be included.


1. **Scenario**: Backing up the current onion address saves hostname and private_key files
**When** selecting the option to back up the current onion address
**And** a location/filename is chosen for the *.bisq file
**Then** the *.bisq (zip) file will be saved containing the hostname and private_key files for the active onion address


2. **Scenario**: Informational prompt is shown prior to creating a new onion address
**When** selecting the option to create a new onion address
**Then** an informational prompt is shown indicating that current reputation will be lost and a restart is required


3. **Scenario**: Creating a new onion address activates a new hidden service after restarting the application
**When** selecting the option to create a new onion address
**And** restarting the application
**Then** a new hidden service using the new onion address will be active along with the old hidden service using the old onion address – can be confirmed by observing the tor/hiddenservice directory and the log output for “Tor hidden service published” messages
_// Both addresses are still in use after restarting the application since unused address are only removed at shutdown and in this case the hidden service in question is still used as the main hidden service during shutdown_
**And** The connected peers list will not show duplicates


4. **Scenario**: Informational prompt is shown prior to restoring an onion address
**When** selecting the option to restore an onion address
**Then** an informational prompt is shown indicating that current reputation will be lost but will be restored from the backup, and a restart is required


5. **Scenario**: Restoring a previously saved onion address activates a new hidden service after restarting the application
**When** selecting the option to restore a previous onion address and choosing a previous backup file
**And** restarting the application
**Then** a new hidden service using the restored onion address will be active along with the old hidden service using the old onion address – can be confirmed by observing the tor/hiddenservice directory and the log output for “Tor hidden service published” messages
_// Both addresses are still in use after restarting the application since unused address are only removed at shutdown and in this case the hidden service in question is still used as the main hidden service during shutdown_
**And** The connected peers list will not show duplicates


6. **Scenario**: Unused hidden services are removed after restarting the application
**Given** there are no open offers or trades
**When** selecting the option to create a new onion address
**And** restarting the application two times
_// The application must be restarted two times because the onion address (hidden service) in question cannot be used as the main hidden service during shutdown_
**Then** the old onion address (hidden service) will no longer be active and will be removed from the tor/hiddenservice directory


7. **Scenario**: Old hidden services (onion addresses) that have open offers or trades are not removed after restarting the application
**Given** there are open offers and/or trades
**When** selecting the option to create a new onion address
**And** restarting the application two times
**Then** the old onion address (hidden service) will continue to be active and remain in the tor/hiddenservice directory, along with the new one
**And** the open offers and/or trades remain active


8. **Scenario**: Hidden services for trades that have completed are removed after restarting the application
**Given** there are open offers and/or trades
**When** a new onion address has been created
**And** the application has been restarted
**And** the open offers/trades have been completed
**And** the application has been restarted
**Then** the old onion address (hidden service) will no longer be active and will be removed from the tor/hiddenservice directory


9. **Scenario**: Restoring an onion address that is already active does not create a new hidden service
**When** selecting the option to restore a previous onion address and choosing a previous backup file containing the same onion address as is already active
**And** restarting the application
**Then** the current onion address (hidden service) will remain active and no new hidden service will be created


10. **Scenario**: Creating a new onion address comes with no reputation
**Given** the active onion address has a reputation (i.e. completed several trades with Bob)
**And** selecting the option to create a new onion address
**And** restarting the application
**When** taking an offer published by Bob
**Then** it indicates you have not traded with that peer


11. **Scenario**: Restoring an onion address restores its reputation
**Given** the active onion address has a reputation (i.e. completed several trades with Bob)
**And** there are no open offers or trades
**And** the active onion address has been backed up
**And** a new onion address has been created
**And** the application has been restarted
**When** the previous onion address is restored
**And** the application is restarted
**And** an offer published by Bob has been taken
**Then** it indicates you have traded with that peer


12. **Scenario**: An open trade can complete successfully when one peer generates a new address in the middle of the trade
**Given** Alice has created an offer and Bob has taken the offer but not initiated payment
**And** Alice has generated a new address and restarted her client
**When** Bob initiates payment
**Then** Alice will receive a notification in her client of the payment initiated
**And** the trade proceeds/completes as normal


13. **Scenario**: Old hidden services (onion addresses) that have failed trades are removed after restarting the application
_Is this correct? Or is there a valid situation where a failed trade can be reopened (i.e. by the mediator/refund agent to republish a tx)?_


14. **Scenario**: Multiple hidden services can run simultaneously without negative effect
**Given** there are >10 open offers and/or trades each using different onion addresses
**When** starting the application
**Then** the application starts in a reasonable amount of time
**And** without consuming an excessive amount of system resources


15. **Scenario**: An error message is shown when attempting to restore an invalid backup file
**When** selecting the option to restore a previous onion address and choosing an invalid backup file
**Then** an error message will be shown for each of the following scenarios
  - not a zip file
  - missing either hostname or private_key files
  - empty hostname or private_key file
  - invalid hostname
  - incorrect private_key

devinbileck:
Additionally, implementing #3780 would have an impact on this. Something to keep in mind, but perhaps there should be a different behavior. For example, perhaps should not remove hidden services that have closed trades <30 days? Or allow the user choose when they want to remove old hidden services?

devinbileck:
Regarding scenario 15 above, the restore and restart will proceed when using an invalid backup file without any error indication to the user, and it will create a new onion address. This seems like a concern in case the backup had become corrupted - the user is unaware and may not realize the address was not actually restored. Can it validate the backup file prior to restore/restart?

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1073,6 +1073,15 @@ setting.preferences.dao.fullNodeInfo.cancel=No, I stick with lite node mode
 settings.net.btcHeader=Bitcoin network
 settings.net.p2pHeader=Bisq network
 settings.net.onionAddressLabel=My onion address
+settings.net.renewAddressButton=Create new address
+settings.net.renewAddress=Please be aware that you will loose your reputation when you create and use a new onion address. Furthermore, you need to restart the application to set your new onion address active.\n\nProceed?
```
devinbileck, line 1077:
```suggestion
settings.net.renewAddress=Please be aware that you will lose your reputation when you create and use a new onion address. Furthermore, you need to restart the application to set your new onion address active.\n\nProceed?
```

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1073,6 +1073,15 @@ setting.preferences.dao.fullNodeInfo.cancel=No, I stick with lite node mode
...
+settings.net.exportAddressButton=Backup address
+settings.net.exportAddressFileDialog=Select the backup file
+settings.net.exportAddressFileEnding=Bisq Onion Address Backup (*.bisq)
+settings.net.importAddressButton=Restore address
+settings.net.importAddressFileDialog=Select the backup file
+settings.net.importAddress=Please be aware that you will loose your reputation when you restore a previous onion address (but restore your reputation connected to the restored onion address). Furthermore, you need to restart the application to set your new onion address active.
```
devinbileck, line 1083:
```suggestion
settings.net.importAddress=Please be aware that you will lose your reputation when you restore a previous onion address (but restore your reputation connected to the restored onion address). Furthermore, you need to restart the application to set your new onion address active.
```

core/src/main/resources/i18n/displayStrings.properties
devinbileck, line 1077: Perhaps we should emphasize the significance of this action. Although it wont remove the old address immediately if there are open offers/trades, they will not be able to back up the old address after creating a new address. So perhaps highly recommend that they backup prior to creating a new address if there are open trades/offers or perhaps even if there are closed trades (just in case).

core/src/main/resources/i18n/displayStrings.properties
devinbileck, line 1083: As mentioned with renew address, perhaps we should emphasize the significance of this action as well.

devinbileck (changes requested):
NACK.
Based on my prior comment requiring validation during restore.

ripcurlx (changes requested):
NACK - Could you please address https://github.com/bisq-network/bisq/pull/3044#pullrequestreview-335577097? Thanks!

ripcurlx:
@freimair I activated yesterday codacy's PR Quality Review for the Bisq repository. It found for your PR a couple of issues which you can view in the Details link next to the PR check. Please review and push fixes for this issues to trigger a re-review by codacy. Thanks!

freimair:
this PR is postponed. it might be reactivated later. see https://github.com/bisq-network/projects/issues/23

PR #3045, opened 2019-08-01
Monitor fixes

freimair:
Finally got the monitor to state that is does not kill some of its metrics after a few hours.

Furthermore, I refactored the P2P Roundtriptime metric so it does not always report all seeds as offline if one or more of them takes too long to respond (the issue showed up just recently, seems that our seeds are under heavier load than they have been).

### DIFF
--- a/monitor/src/main/java/bisq/monitor/metric/P2PSeedNodeSnapshot.java
+++ b/monitor/src/main/java/bisq/monitor/metric/P2PSeedNodeSnapshot.java
@@ -192,6 +192,8 @@ public class P2PSeedNodeSnapshot extends P2PSeedNodeSnapshotBase {
                                 String.valueOf(((Counter) count).value() - referenceValues.get(messageType).value()));
                     } catch (MalformedURLException ignore) {
                         log.error("we should never got here");
+                    } catch (NullPointerException e) {
+                        // we got no reference-Value if we got here. strange.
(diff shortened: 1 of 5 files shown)


christophsturm:
would it be easy to add some unit tests to this pr?

freimair:
as the monitor is our least critical infrastructure I rather spend the time fixing stuff in Bisq itself.

christophsturm:
At some point in time there should be a rule that any change requires at least simple unit tests. And that time could as well be now.

freimair:
agreed. however, as we do not have the resources to do all and any testing in an instant we need to spend resources in thoughtful way

monitor/src/main/java/bisq/monitor/metric/P2PSeedNodeSnapshot.java
```diff
@@ -192,6 +192,8 @@ void report() {
                                 String.valueOf(((Counter) count).value() - referenceValues.get(messageType).value()));
                     } catch (MalformedURLException ignore) {
                         log.error("we should never got here");
+                    } catch (NullPointerException e) {
+                        // we got no reference-Value if we got here. strange.
```
christophsturm: maybe log here too?

sqrrm (approved):
utACK

PR #3051, opened 2019-08-03
Add checksum-dependency-plugin to verify plugin and dependency checksums

vlsi:


sqrrm:
While it might be a good idea to verify plugin cheksums, I worry about adding more dependencies. I don't feel I can do a review of this dependency at this point.

vlsi:
>I don't feel I can do a review of this dependency at this point.

Oh, that is interesting.

How about adding the dependency for Travis only?

christophsturm:
I'm not sure but I think that newer gradle versions have their own checksum support.

christophsturm:
does this replace gradle-witness? the pr does not at all touch the gradle witness config.

vlsi:
> I'm not sure but I think that newer gradle versions have their own checksum support.

It is not there yet. However would likely appear sooner or later.

>does this replace gradle-witness? the pr does not at all touch the gradle witness config.

It does. However, if I replaced `gradle-witness` with `checksum-dependency-plugin` in a single PR, then it would be hard to tell if I change "the expected checksums".

However as I just add `checksum-dependency-plugin` that makes it very simple to reason about: all the dependencies still match previous assumptions.
If `checksum-dependency-plugin` is merged, then `gradle-witness` could be removed.

christophsturm:
How does the checksumming of the plugin itself work? If someone replaces the plugin with a hostile version that just does not check anything, how will that be detected?

vlsi:
> If someone replaces the plugin with a hostile version that just does not check anything, how will that be detected?

Checksumming is done by an added code in settings.gradle.kts:  https://github.com/bisq-network/bisq/pull/3051/files#diff-88b7c47e47b8ee65263b6784b86fa0a7R30-R39
The plugin is not used to checksum itself.

The sequence is as follows:
1) Dependency is declared
2) Dependency is resolved (jar file is downloaded) by `buildscript.configurations.classpath.resolve()`
3) It is verified for the checksum
4) Then the plugin is applied

I probably need to add an explicit test to `checksum-dependency-plugin` codebase (to ensure that holds for further Gradle versions), however I did try to add a static initializer block to the plugin to see if it could execute before checksum verification, and the outcome is static initializer is NOT executed before checksumming (at least in Gradle 5.5.1)

vlsi:
Of course there's always an option to keep the jar in Git. The jar file is ~40KiB.

However I personally like that the jar could be downloaded as a dependency because storing binary files under source control is not that good.

ManfredKarrer:
@vlsi Please have patient with that PR. We are very short on devs and reviewers atm and such a security critical PR need to get a extended review, including your repo with the added dependencies. Bisq is very conservative adding dependencies...

vlsi:
@ManfredKarrer , I see.

Just in case, I've implemented PGP-based verification.

ripcurlx:
@vlsi Could you please update your PR to use signed commits only.
@ManfredKarrer @sqrrm @christophsturm Are you still up for reviewing this PR?

vlsi:
I've signed the commit, and updated the plugin to a newer version.

ripcurlx:
> I've signed the commit, and updated the plugin to a newer version.

The current build seems to be broken (see Travis)

vlsi:
@ripcurlx , I've rebased the change, Travis test passes. Please review.

vlsi:
Hi there. I'm not pushing, but you might be interested that Gradle 6.2 is going to implement in-core verification of both checksums and signatures which might might cover more cases than checksum-dependency-plugin.
For instance, the plugin does not verify `pom.xml` files if they are not used as explicit dependencies, while Gradle verifies `pom.xml` as well. I don't know the other cases, but they might be there as well.

Here's the relevant documentation: https://github.com/gradle/gradle/pull/11755
Please feel free to review/comment.

ripcurlx:
> Hi there. I'm not pushing, but you might be interested that Gradle 6.2 is going to implement in-core verification of both checksums and signatures which might might cover more cases than checksum-dependency-plugin.
> For instance, the plugin does not verify `pom.xml` files if they are not used as explicit dependencies, while Gradle verifies `pom.xml` as well. I don't know the other cases, but they might be there as well.
> 
> Here's the relevant documentation: [gradle/gradle#11755](https://github.com/gradle/gradle/pull/11755)
> Please feel free to review/comment.

Thanks for the hint!

PR #3057, opened 2019-08-05
WIP show strange guice behavior

christophsturm:
Can anybody explain this? why does bisq.seednode.GuiceSetupTest work even when the OfferBook registration is commented out?

@ManfredKarrer @ripcurlx @sqrrm @freimair

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -176,7 +176,7 @@ configure(project(':common')) {
         compile "ch.qos.logback:logback-classic:$logbackVersion"
         compile 'com.google.code.findbugs:jsr305:3.0.2'
         compile 'com.google.guava:guava:20.0'
-        compile('com.google.inject:guice:4.1.0') {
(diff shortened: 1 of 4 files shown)


ManfredKarrer:
If you don't define a binding Guice creates a new instance each time when a dependey is injected. So thats why it is important to set bindings, specially if only 1 instance must exist. But best practice to define all bindings.

christophsturm:
so in the context of bisq with its different guice modules per app this means that a component that is on the classpath but not in the module of that app will still be available but possibly with the wrong lifecycle (not singleton)

ManfredKarrer:
yes.

christophsturm:
so it would be better to use the `@Singleton` annotation on our own classes instead of binding them as `Singleton.class` in the module. then its always correct. binding is mostly for classes that you don't have control over.

PR #3058, opened 2019-08-06
use singleton annotation instead of binding as singleton

christophsturm:
this is the correct way to do it. the way it is currently done means
that if a component is not declared in a module, guice still finds it
but does not use the singleton lifecycle

binding as singleton in the module is meant to be used for classes that
we don't have the sourcecode for (i.e. jdk classes)

also update guice because the current version has some incompatibilites with our asm lib and outputs internal errors when the wiring fails. 

this pr only converts the components that were declared in the desktop module. will convert the others later.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -176,7 +176,7 @@ configure(project(':common')) {
         compile "ch.qos.logback:logback-classic:$logbackVersion"
         compile 'com.google.code.findbugs:jsr305:3.0.2'
         compile 'com.google.guava:guava:20.0'
-        compile('com.google.inject:guice:4.1.0') {
+        compile('com.google.inject:guice:4.2.2') {
             exclude(module: 'guava')
         }
         compile("com.github.bisq-network.bitcoinj:bitcoinj-core:$bitcoinjVersion") {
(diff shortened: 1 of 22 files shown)


ManfredKarrer:
Should it be in the responsibility of the class to know in which context it will be used or the application (via guice module)?

christophsturm:
definately the class, the class knows if its a singleton. also the problem is that even when the class is not defined in the module but still on the classpath it will be found by guice, but have the wrong lifecycle. doing it by annotation ensures that the lifecycle is always correct.

mrosseel:
Is there a way to detect these 'orphan' classes which are automatically
wired by guice? Breaking the build would be better because even annotations
can be forgotten.

christophsturm:
It's not really orphan classes, it's just how guice works. Binding a class in the module is for classes that you don't have the sourcecode for. If you add an `Inject` annotation and the class is a singleton you need to add a `Singleton` annotation too.

blabno (approved):
The code looks good. 
I see pros and cons in both approaches. Just as forgetting to configure class as singleton in a module you can forget about annotating the class. Defining life cycle inside module allows for class reuse in different context, but that's very theoretical benefit.

sqrrm (approved):
ACK

PR #3061, opened 2019-08-07
also fix singleton config for core module

christophsturm:
follow up to #3058 that converts the core module too. 
and unit test it all

### DIFF
--- a/core/src/main/java/bisq/core/CoreModule.java
+++ b/core/src/main/java/bisq/core/CoreModule.java
@@ -25,15 +25,6 @@ import bisq.core.btc.BitcoinModule;
 import bisq.core.dao.DaoModule;
 import bisq.core.filter.FilterModule;
 import bisq.core.network.p2p.seed.DefaultSeedNodeRepository;
-import bisq.core.notifications.MobileMessageEncryption;
-import bisq.core.notifications.MobileModel;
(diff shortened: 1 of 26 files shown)


blabno (approved):
The code looks good.
I see pros and cons in both approaches. Just as forgetting to configure class as singleton in a module you can forget about annotating the class. Defining life cycle inside module allows for class reuse in different context, but that's very theoretical benefit.

core/src/main/java/bisq/core/CoreModule.java
```diff
@@ -118,15 +84,6 @@ protected void configure() {
         String referralId = environment.getProperty(AppOptionKeys.REFERRAL_ID, String.class, "");
         bind(String.class).annotatedWith(Names.named(AppOptionKeys.REFERRAL_ID)).toInstance(referralId);
 
-        bind(MobileNotificationService.class).in(Singleton.class);
-        bind(MobileMessageEncryption.class).in(Singleton.class);
```
sqrrm, line 122: MobileMessageEncryption doesn't have singleton annotation and not tested in GuiceSetupTest.java

sqrrm (changes requested):
Please fix the missing singleton

sqrrm (approved):
utACK

PR #3080, opened 2019-08-13
use pubkeyring instead of keyring where possible. 

christophsturm:
upgrade mockito and start some tests.


christophsturm:
not ready

PR #3081, opened 2019-08-13
fix build on jdk11 and jdk12. remove powermock and jmockit

christophsturm:
on jdk 12 the app still throws a NPE at startup (but afterwards it seems to work),  but runs fine on jdk11.

build and test suite works on jdk10,11,12, inside IDEA or with `./gradlew clean build` 

jdk 10 is probably still needed for binaries. but contributors can now use any jdk 10+ that they like

had to disable a very small amount  of tests (i think 3) that were not so easy to convert.

### DIFF
--- a/core/src/main/java/bisq/core/offer/OfferPayload.java
+++ b/core/src/main/java/bisq/core/offer/OfferPayload.java
@@ -51,7 +51,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 @EqualsAndHashCode
 @Getter
 @Slf4j
-public final class OfferPayload implements ProtectedStoragePayload, ExpirablePayload, RequiresOwnerIsOnlinePayload {
+public class OfferPayload implements ProtectedStoragePayload, ExpirablePayload, RequiresOwnerIsOnlinePayload {
 
     ///////////////////////////////////////////////////////////////////////////////////////////
     // Enum
--- a/core/src/main/java/bisq/core/offer/OfferPayload.java
+++ b/core/src/main/java/bisq/core/offer/OfferPayload.java
@@ -51,7 +51,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 @EqualsAndHashCode
 @Getter
 @Slf4j
-public class OfferPayload implements ProtectedStoragePayload, ExpirablePayload, RequiresOwnerIsOnlinePayload {
+public final class OfferPayload implements ProtectedStoragePayload, ExpirablePayload, RequiresOwnerIsOnlinePayload {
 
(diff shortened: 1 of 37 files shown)


desktop/src/test/java/bisq/desktop/main/market/trades/TradesChartsViewModelTest.java
```diff
@@ -121,10 +107,11 @@
             null,
             1
     );
-
     @Before
     public void setup() throws IOException {
-
+        tsm = mock(TradeStatisticsManager.class);
```
blabno: Please avoid acronyms.

blabno (changes requested):
Great job. I have only one small remark.

desktop/src/test/java/bisq/desktop/main/market/trades/TradesChartsViewModelTest.java
christophsturm: yeah that's what it was called before, will change.

blabno (approved):
utACK

core/src/main/java/bisq/core/offer/OfferPayload.java
```diff
@@ -51,7 +51,7 @@
 @EqualsAndHashCode
 @Getter
 @Slf4j
-public final class OfferPayload implements ProtectedStoragePayload, ExpirablePayload, RequiresOwnerIsOnlinePayload {
+public class OfferPayload implements ProtectedStoragePayload, ExpirablePayload, RequiresOwnerIsOnlinePayload {
```
ManfredKarrer: Should stay final for security reasons

desktop/src/main/java/bisq/desktop/main/offer/MakerFeeMaker.java
```diff
@@ -0,0 +1,13 @@
+package bisq.desktop.main.offer;
+
+import bisq.core.btc.wallet.BsqWalletService;
+import bisq.core.offer.OfferUtil;
+import bisq.core.user.Preferences;
+
+import org.bitcoinj.core.Coin;
+
+public class MakerFeeMaker {
```
ManfredKarrer, line 9: MakerFeeMaker sounds a bit weird. Maybe MakerFeeProvider?

core/src/main/java/bisq/core/offer/OfferPayload.java
christophsturm: it is final (not in this commit but later)

ManfredKarrer: Ah saw it now... thanks.

ManfredKarrer (approved):
utACK

Issue #123, opened 2014-10-03
First time startup: Display info 


ManfredKarrer:
We should display a info to first time user that he can play around without registration, just when he wants to trade he need to register.

PR #3086, opened 2019-08-14
inject publickeyring

christophsturm:
based on #3081

### DIFF
--- a/common/src/main/java/bisq/common/crypto/PubKeyRing.java
+++ b/common/src/main/java/bisq/common/crypto/PubKeyRing.java
@@ -42,7 +42,7 @@ import javax.annotation.Nullable;
 @Slf4j
 @EqualsAndHashCode
 @Getter
-public final class PubKeyRing implements NetworkPayload, UsedForTradeContractJson {
+public class PubKeyRing implements NetworkPayload, UsedForTradeContractJson {
     private final byte[] signaturePubKeyBytes;
     private final byte[] encryptionPubKeyBytes;
     @Nullable
--- a/common/src/main/java/bisq/common/crypto/PubKeyRing.java
+++ b/common/src/main/java/bisq/common/crypto/PubKeyRing.java
@@ -42,7 +42,7 @@ import javax.annotation.Nullable;
 @Slf4j
 @EqualsAndHashCode
 @Getter
-public class PubKeyRing implements NetworkPayload, UsedForTradeContractJson {
+public final class PubKeyRing implements NetworkPayload, UsedForTradeContractJson {
     private final byte[] signaturePubKeyBytes;
     private final byte[] encryptionPubKeyBytes;
     @Nullable
--- a/core/src/main/java/bisq/core/CoreModule.java
+++ b/core/src/main/java/bisq/core/CoreModule.java
@@ -40,6 +40,8 @@ import bisq.network.p2p.seed.SeedNodeRepository;
 import bisq.common.CommonOptionKeys;
 import bisq.common.app.AppModule;
 import bisq.common.crypto.KeyStorage;
+import bisq.common.crypto.PubKeyRing;
+import bisq.common.crypto.PubKeyRingProvider;
 import bisq.common.proto.network.NetworkProtoResolver;
 import bisq.common.proto.persistable.PersistenceProtoResolver;
 import bisq.common.storage.Storage;
@@ -96,6 +98,7 @@ public class CoreModule extends AppModule {
         install(alertModule());
         install(filterModule());
         install(corePresentationModule());
+        bind(PubKeyRing.class).toProvider(PubKeyRingProvider.class);
     }
 
     private TradeModule tradeModule() {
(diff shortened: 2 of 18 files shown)


core/src/test/java/bisq/core/dao/governance/proposal/MyProposalListServiceTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class MyProposalListServiceTest {
+    @Test
+    public void canInstantiate() {
+        P2PService p2PService = mock(P2PService.class);
+        when(p2PService.getNumConnectedPeers()).thenReturn(new SimpleIntegerProperty(0));
+        Storage storage = mock(Storage.class);
```
mrosseel, line 24: why is storage mocked separately? It's just passed as an argument like the rest.

desktop/src/main/java/bisq/desktop/main/offer/MakerFeeMaker.java
```diff
@@ -0,0 +1,13 @@
+package bisq.desktop.main.offer;
+
+import bisq.core.btc.wallet.BsqWalletService;
+import bisq.core.offer.OfferUtil;
+import bisq.core.user.Preferences;
+
+import org.bitcoinj.core.Coin;
+
+public class MakerFeeMaker {
```
mrosseel: not sure about the name, too many makers making things :)
Don't have any good alternatives though... MakerFeeFactory, MakerFeeCalculation, MakerFeeCreator, ... (if you don't find anything better don't worry)

This is just a wrapper for one of the util classes, so will you need to do this for every method in every Util class? Are you trying to avoid calls to static methods for unit testing purposes?

mrosseel (commented):
utACK

Don't know how the 'mergers' think about this but wouldn't it be cleaner if you rebase this on your previous PR (this PR doesn't only contain the inject publickeyring changes) ? You could add a 'blocked by #123' comment so it doesn't get merged before the parent. Would make for cleaner diffs

core/src/test/java/bisq/core/dao/governance/proposal/MyProposalListServiceTest.java
christophsturm: yeah that test is just a stub that got me thinking about the pubkey.

desktop/src/main/java/bisq/desktop/main/offer/MakerFeeMaker.java
christophsturm: I'm trying to get rid of the static util classes, for testing purpose but also for architectural purposes. this is part of the parent pr, see the description of this pr

christophsturm:
> utACK
> 
> Don't know how the 'mergers' think about this but wouldn't it be cleaner if you rebase this on your previous PR (this PR doesn't only contain the inject publickeyring changes) ? You could add a 'blocked by #123' comment so it doesn't get merged before the parent. Would make for cleaner diffs

i can only rebase this pr once the other one is in master. and then i will rebase it and it will be much smaller.

mrosseel (commented):
Is this not confusing, there are already the guice providers?

core/src/main/java/bisq/core/CoreModule.java
```diff
@@ -96,6 +98,7 @@ protected void configure() {
         install(alertModule());
         install(filterModule());
         install(corePresentationModule());
+        bind(PubKeyRing.class).toProvider(PubKeyRingProvider.class);
```
ManfredKarrer, line 101: Is this guaranteed to be in singleton context?
Can you move it to the other bindings above, so th einstall calls are not mixed with bind calls?

ManfredKarrer: Ah from the test below its seems its singleton.

common/src/main/java/bisq/common/crypto/PubKeyRing.java
```diff
@@ -42,7 +42,7 @@
 @Slf4j
 @EqualsAndHashCode
 @Getter
-public final class PubKeyRing implements NetworkPayload, UsedForTradeContractJson {
+public class PubKeyRing implements NetworkPayload, UsedForTradeContractJson {
```
ManfredKarrer: Might be used in p2p messages so must be final. I know its prob. not used now but further refactoring might use it... all NetworkPayload classes should be final.

ManfredKarrer (changes requested):
NACK

See comment for missing final.

I will review again after rebase with other PR is done.

ManfredKarrer:
> Is this not confusing, there are already the guice providers?

We use the term provider in a wider range. IMO its not bad, better then factory IMO as thats more overloaded.

core/src/main/java/bisq/core/CoreModule.java
christophsturm: yes i made sure that it is.

desktop/src/main/java/bisq/desktop/main/offer/MutableOfferDataModel.java
```diff
@@ -106,7 +106,7 @@
     private final TxFeeEstimationService txFeeEstimationService;
     private final ReferralIdService referralIdService;
     private final BSFormatter btcFormatter;
-    private MakerFeeMaker makerFeeMaker;
+    private MakerFeeProvider makerFeeProvider;
```
ManfredKarrer, line 109: Can be final

ManfredKarrer (approved):
ACK

PR #3087, opened 2019-08-14
Remove legacy hack from build.gradle installDist task

battleofwizards:
The rationale for the "delete fileTree" hack no longer exists because the conflicting files are not created anymore.


ManfredKarrer (approved):
utACK

freimair (approved):
Ack

PR #3090, opened 2019-08-14
Render proper column headers and values for CSV exports; fix #3075

### LINKED ISSUE #3075: Portfolio CSV export Headers
The first line of a CSV export is
`,,,,,,,,,,,,`

and it is quite hard to figure out what is what. Would be nice to have a header line on what column is what.

battleofwizards:
When exporting to CSV, both portfolio/trades and funds/transactions were lacking the column headers.

Additionally, portfolio/trades colum values were outdated and not matching the intended export.

Both issues are fixed.


ManfredKarrer (approved):
utACK

freimair (approved):
Ack

PR #3091, opened 2019-08-14
jdk10 is no longer needed to build and run the test suite

christophsturm:


ManfredKarrer (commented):
Is JDK 11 easy accesible from Oracle? If not maybe keep the links to the install scripts or to openJDK? Maybe some note that Bisq is only tested with JDK 10 and shipped with JDK 10 might be good as well?

freimair (approved):
Ack

PR #3093, opened 2019-08-15
Takeover jhgcy2won7xnslrb.onion from @devinbileck

wiz:
As of 20190815 @ 0615Z, I have taken over the jhgcy2won7xnslrb.onion seed node from @devinbileck


devinbileck (approved):
ACK

PR #3092, opened 2019-08-14
Define more version variables, fix style problems

ghubstan:
Define fontawesomefx version variables.

Fix testCompile org.mockito:mockito-core syntax.

PR #3100, opened 2019-08-16
Define more version variables, fix style problems

ghubstan:
Define fontawesomefx version variables.

Fix testCompile org.mockito:mockito-core syntax.

Replaces PR #3092.


ManfredKarrer (commented):
utACK

Issue #3042, opened 2019-08-01
Normalize Phone Numbers


kdeldycke:
Just happened to have an issue on setting up a new Revolut account on Bisq: the phone number I registered is taken as-is and will be displayed as such to the receiving end of the transaction.

I entered my Revolut account's phone number in the local (French) format, i.e. `0612345678` and the other party, not located in the same country as I am, did not find my account. It had to be formatted in the international form: `+ 33 6 12 34 56 78`.

I failed my first transaction on Revolut for that reason. But was able to solve the dispute thanks to a great arbitrator.

To fix this issue for good and reduce the number of disputes, I propose to have all phone number fields in Bisq normalised into their international form.

Good news, there's a great library made by Google to normalise all of them at: https://github.com/google/libphonenumber . Used it myself, it's solid, well maintained and clean! :)

ghubstan:
I'm working on a validator / formatter and tests for this issue.
Please let me know what objections there are to this approach:

- Adding a new 343 kB Google [libphonenumber](https://search.maven.org/artifact/com.googlecode.libphonenumber/libphonenumber/8.10.16/jar) dependency.

- Converting all input phone numbers to [E.164](https://en.wikipedia.org/wiki/E.164) format.
    @kdeldycke's example input `0612345678` would be re-formatted as `+33612345678`.

ripcurlx:
Closing as complete

PR #3102, opened 2019-08-16
Normalize phone numbers

ghubstan:
Reluctantly added new dependency: 343 kb Google libphonenumber.
See https://github.com/google/libphonenumber

Created new PhoneNumberValidator + Test.  Validator hides
ISO country code for validating national numbers.

If input passes validation, it is transformed into E.164 format
and cached in a class field, accessible via getter method.
See https://en.wikipedia.org/wiki/E.164

Five new i18n error message properties (validation.phone.*) were
added to the default and Portuguese display string property files.
A native Portuguese speaker should review/change the PT strings.

This is partial fix for Issue #3042 -- validator will be integrated
into the GUI after PR(s) approved / merged.


battleofwizards:
Thanks, looks promising! Also, very nice you provided the tests.

Short on time / quick and dirty comments:

- have you reviewed dependencies of libphonenumber, if any?
- have you checked the library is self-contained and **never calls home**?
- can we migrate away from static fields to instance fields? if not, is it verified libphonenumber is fully thread safe?
- validate() is a huge method, would be nice to name its fragments by extracting private methods
- the tests beg for helpers encapsulating the intent, like `assertInvalid` or whatever is idiomatic for this test framework (`refuteValid`, `assertNotValid`?)
- you cannot directly edit translations except English. See https://github.com/bisq-network/bisq/blob/master/docs/translation-process.md for details

ManfredKarrer:
I am not sure if the use case here is worth the risk of adding a new dependency (even if its a google library). We basically want to go the opposite way to reduce dependencies as far as possible.

ghubstan:
Closing PR.  Will redo w/out another dependency.

PR #3103, opened 2019-08-17
Auto remove whitespace in the IBAN input field

battleofwizards:
When adding a payment account Bisq requires IBAN to be whitespace free.

It is annoying for the user to manually clean pasted IBAN of spaces. In online banking IBAN-s are usually displayed **with** spaces for better readability.

This commit auto-removes whitespace when user is leaving the input field, right before validation, effectively allowing to paste real world IBAN-s.


ripcurlx (approved):
ACK

PR #3107, opened 2019-08-19
inject CorruptedDataFilesHandler instead of making it static

christophsturm:
just a small rabbit-hole that i went down while looking at the code. 

storage can probably be injected directly in more places but i didn't want to change any logic in this pr.

### DIFF
--- a/core/src/main/java/bisq/core/arbitration/DisputeManager.java
+++ b/core/src/main/java/bisq/core/arbitration/DisputeManager.java
@@ -56,8 +56,6 @@ import bisq.common.handlers.FaultHandler;
 import bisq.common.handlers.ResultHandler;
 import bisq.common.proto.network.NetworkEnvelope;
 import bisq.common.proto.persistable.PersistedDataHost;
-import bisq.common.proto.persistable.PersistenceProtoResolver;
-import bisq.common.storage.CorruptedDatabaseFilesHandler;
 import bisq.common.storage.Storage;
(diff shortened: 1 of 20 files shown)


core/src/main/java/bisq/core/arbitration/DisputeManager.java
```diff
@@ -137,7 +138,7 @@ public DisputeManager(P2PService p2PService,
                           OpenOfferManager openOfferManager,
                           KeyRing keyRing,
                           PersistenceProtoResolver persistenceProtoResolver,
-                          @Named(Storage.STORAGE_DIR) File storageDir) {
+                          @Named(Storage.STORAGE_DIR) File storageDir, CorruptedDatabaseFilesHandler corruptedDatabaseFilesHandler) {
```
ManfredKarrer: Can you add a line break. So params are better separated/readable?

ManfredKarrer (approved):
utACK

core/src/main/java/bisq/core/arbitration/DisputeManager.java
christophsturm: sure, i wonder if there is a way to make idea do it automatically.

ripcurlx:
Just a small thing regarding commit messages. Please start them with uppercase https://chris.beams.io/posts/git-commit/

ripcurlx (approved):
utACK - But please see my comment for future commits.

PR #3108, opened 2019-08-19
make sure that signed witness is immutable.

christophsturm:
btw I do hate byte arrays

### DIFF
--- a/core/src/test/java/bisq/core/account/sign/SignedWitnessTest.java
+++ b/core/src/test/java/bisq/core/account/sign/SignedWitnessTest.java
@@ -10,22 +10,51 @@ import com.google.common.base.Charsets;
 
 import java.time.Instant;
 
+import org.junit.Before;
 import org.junit.Test;
 
+import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 
 public class SignedWitnessTest {
+
+    private ECKey arbitrator1Key;
+    private byte[] witnessOwner1PubKey;
+    private byte[] witnessHash;
+    private byte[] witnessHashSignature;
+
+    @Before
+    public void setUp() {
+        arbitrator1Key = new ECKey();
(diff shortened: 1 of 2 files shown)


ManfredKarrer:
We have a wrapper class as well (P2PDataStorage.ByteArray in p2p module) but its only used where needed (e.g. when the bytarray is a key in a hashmap).

ManfredKarrer (commented):
utACK

Good find!

core/src/test/java/bisq/core/account/sign/SignedWitnessTest.java
```diff
@@ -10,22 +10,51 @@
...
+    public void testProtoRoundTrip() {
         SignedWitness signedWitness = new SignedWitness(true, witnessHash, witnessHashSignature, arbitrator1Key.getPubKey(), witnessOwner1PubKey, Instant.now().getEpochSecond(), 100);
         assertEquals(signedWitness, SignedWitness.fromProto(signedWitness.toProtoMessage().getSignedWitness()));
+    }
 
+    @Test
+    public void isImmutable() {
+        byte[] signerPubkey = arbitrator1Key.getPubKey();
+        SignedWitness signedWitness = new SignedWitness(true, witnessHash, witnessHashSignature, signerPubkey, witnessOwner1PubKey, Instant.now().getEpochSecond(), 100);
+        byte[] witnessHash = signedWitness.getWitnessHash().clone();
```
ManfredKarrer: Maybe better to use a different name? Now the this.witnessHash and witnessHash are different, which makes the test a bit harder to read IMO.

ManfredKarrer (approved):
utACK

see Nit comment...

core/src/test/java/bisq/core/account/sign/SignedWitnessTest.java
christophsturm: done

sqrrm (approved):
utACKl

PR #3110, opened 2019-08-20
List Ndau (XND)

coriolinus:
- Official project URL: https://ndau.io/
- Official block explorer URL: https://explorer.service.ndau.tech

~~Note: as ndau address validation is fairly complicated, this PR adds
a dependency to a library which handles this. This library can be
inspected at https://github.com/oneiro-ndev/ndauj. It is licensed
under the LGPL 3.0.~~

This code has been removed.

### DIFF
--- a/assets/src/main/java/bisq/asset/coins/Ndau.java
+++ b/assets/src/main/java/bisq/asset/coins/Ndau.java
@@ -1,0 +1,37 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * Copyright © 2019 Oneiro NA, Inc.
+ */
+
+package bisq.asset.coins;
+
+import bisq.asset.Coin;
+import bisq.asset.RegexAddressValidator;
+
+
+public class Ndau extends Coin {
+
+    public Ndau() {
+        // note: ndau addresses contain an internal checksum which was deemed too complicated to include here.
+        // this regex performs superficial validation, but there is a large space of addresses marked valid
+        // by this regex which are not in fact valid ndau addresses. For actual ndau address validation,
+        // use the Address class in github.com/oneiro-ndev/ndauj (java) or github.com/oneiro-ndev/ndaumath/pkg/address (go).
+        super("ndau", "XND", new RegexAddressValidator("nd[anexbm][abcdefghijkmnpqrstuvwxyz23456789]{45}"));
+    }
+}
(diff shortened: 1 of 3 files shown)


battleofwizards:
NACK

I guess we would rather skip the altcoin altogether (it's not even present on CMC), or have no full address validation rather than adding a new dependency.

build.gradle
```diff
@@ -75,6 +75,7 @@ configure(subprojects) {
 
     dependencies {
         testCompile "junit:junit:$junitVersion"
+        implementation 'com.github.oneiro-ndev:ndauj:v1.0.3'
```
ManfredKarrer: Big NACK!

ManfredKarrer (changes requested):
NACK for adding dependency!

wiz:
Why would you expect Bisq to trust your 13 day old jar dependency? 🤔

coriolinus:
The jar is about 500 lines of code and easily auditable. It's 13 days old because our primary development language is go, not Java; that JAR was developed specifically to support this project. 

If you prefer, it would be possible for us to inline the necessary code here. It just means adding a substantial amount more code than other altcoins have.

coriolinus:
For bringing the code into this repo, what is the preferred style: to keep all necessary code within the `Ndau.java` file, or to create multi-file packages as required?

battleofwizards:
@coriolinus inlining this dependency won't help.

coriolinus:
> I guess we would rather... have no full address validation rather than adding a new dependency.

If I were to use a regex validator, it would necessarily accept addresses which the actual validator would reject: you can't check internal hash codes with a regex. That said, I do not know what the actual effects would be if the bisq validator accepted addresses which are in fact invalid. If the consequences are acceptable, it would be easy to implement. 

Doing so will necessitate changing the test suite, many of the invalid cases of which are single-character replacements which invalidate the internal hash.

@battleofwizards would that be more palatable?

coriolinus:
I have updated this PR to remove the external dependency. I would appreciate a re-review of the code as it now stands.

mrosseel:
utACK
dependency is gone and replaced by simple regex address validation. As @coriolinus mentioned some invalid addresses can slip by this regex, but I'm not aware that this is a requirement to include an altcoin.

ripcurlx:
@coriolinus please resolve the conflict with Asset (because of merged new assets)

coriolinus:
@ripcurlx Done.

ripcurlx:
@blabno Any comments regarding our other asset requirements? Codewise it looks fine to me now.

assets/src/main/java/bisq/asset/coins/Ndau.java
```diff
@@ -0,0 +1,41 @@
...
+import bisq.asset.RegexAddressValidator;
+
+
+public class Ndau extends Coin {
+
+    // note: ndau addresses contain an internal checksum which was deemed too complicated to include here.
+    // this regex performs superficial validation, but there is a large space of addresses marked valid
+    // by this regex which are not in fact valid ndau addresses. For actual ndau address validation,
+    // use the Address class in github.com/oneiro-ndev/ndauj (java) or github.com/oneiro-ndev/ndaumath/pkg/address (go).
+    private static AddressValidator validator = new RegexAddressValidator("nd[anexbm][abcdefghijkmnpqrstuvwxyz23456789]{45}");
```
blabno: There is no need to make this an object property. Move validator instantiation and declaration into constructor.

assets/src/main/java/bisq/asset/coins/Ndau.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public Ndau() {
+        super("ndau", "XND", Ndau.validator, Network.MAINNET);
```
blabno: Mainnet is the default value, so no need for that parameter.

blabno: Name of asset should be uppercased I guess.

blabno (changes requested):
NACK per comments.

Squash commits into one, keeping proper commit message.

assets/src/main/java/bisq/asset/coins/Ndau.java
coriolinus: Stet: style guide rules for this asset mandate lowercase usage.

ripcurlx:
@blabno Could you please re-visit this PR if it is up to be merged now? Thanks!

blabno (approved):
ACK

PR #3111, opened 2019-08-20
bump gradle to 5.6

christophsturm:
and use source distribution to get better idea support.


christophsturm:
closing this for now because it does not work with ancient IDEA versions.

PR #3112, opened 2019-08-20
Various optimizations

freimair:
Various optimizations to the Bisq code
- reduce info to debug log on spamming infos
- do not do parsing until everything is completed and thus, save on resources and time

### DIFF
--- a/core/src/main/java/bisq/core/dao/governance/proposal/param/ChangeParamValidator.java
+++ b/core/src/main/java/bisq/core/dao/governance/proposal/param/ChangeParamValidator.java
@@ -63,9 +63,13 @@ public class ChangeParamValidator extends ProposalValidator implements Consensus
     public void validateDataFields(Proposal proposal) throws ProposalValidationException {
         try {
             super.validateDataFields(proposal);
-            ChangeParamProposal changeParamProposal = (ChangeParamProposal) proposal;
-            validateParamValue(changeParamProposal.getParam(), changeParamProposal.getParamValue(), getBlockHeight(proposal));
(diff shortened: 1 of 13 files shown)


core/src/main/java/bisq/core/dao/governance/proposal/ProposalValidator.java
```diff
@@ -55,7 +55,8 @@ public boolean areDataFieldsValid(Proposal proposal) {
             validateDataFields(proposal);
             return true;
         } catch (ProposalValidationException e) {
-            log.warn("proposal data fields are invalid. proposal={}, error={}", proposal, e.toString());
+            // TODO only log once dao state is complete
+            //log.warn("proposal data fields are invalid. proposal={}, error={}", proposal, e.toString());
```
ManfredKarrer: That should be still used. The fix in the subclass avoids the wrong logs

ManfredKarrer (changes requested):
NACK

See comment

core/src/main/java/bisq/core/dao/governance/proposal/ProposalValidator.java
freimair: force-pushed without this very change

christophsturm: where is the fix that avoids that?

ManfredKarrer: Would have been better in a new commit so make it more transparent.

ManfredKarrer (approved):
utACK

sqrrm (approved):
ACK

PR #3113, opened 2019-08-20
also run tests on jdk12

christophsturm:
Right now the build works on 10, 11 and 12, and the desktop app runs on 10 and 11, and with some minor problems on jdk12.
It makes no sense to ever officially support JDK11 because it is as unsupported as 10.
So I think we should run CI on jdk 10 to make sure we can build official binaries, and jdk 12 to make sure anyone with a supported JDK can contribute.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,10 @@ jdk:
     - openjdk12
 
 before_install:
-  grep -v '^#' assets/src/main/resources/META-INF/services/bisq.asset.Asset | sort --check --dictionary-order --ignore-case
+    grep -v '^#' assets/src/main/resources/META-INF/services/bisq.asset.Asset | sort --check --dictionary-order --ignore-case
 notifications:
-  slack:
-    on_success: change
-    on_failure: always
-    rooms:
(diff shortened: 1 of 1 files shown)


sqrrm (commented):
Why change all the indentation?

christophsturm:
i made the change in one commit and then i reformatted the file with IDEA.

sqrrm:
My idea doesn't format this file at all, shouldn't .editorconfig handle this? Although it would use 4 space indents from what I understand so that this auto format of yours should be correct.

christophsturm:
maybe your older idea version does not format yaml files?

christophsturm:
anyway when a reformat changes the whole file i do it in a separate commit to make the changes easier to see

battleofwizards:
Concept ACK. Thanks!

ripcurlx (commented):
I just ran a manual autoformat of this file with the latest Intellij (2019.2.1) and it added an extra space after the `-` in `- secure`. Maybe add this there as well so if someone runs manual autoformat it won't change this file again.

Issue #3026, opened 2019-07-29
convert source code to kotlin


christophsturm:
I think we should convert all source files that we change to kotlin *and* write new source files in kotlin.

This may seem a bit far fetched but:
* I have done this on several projects already and it was always a very smooth transition
* kotlin code is easy to read for java programmers.
* idea can automatically convert java files to kotlin, without changing the semantics or api.
* kotlin can be uses as just a "better java" while we transition, and we can refactor to more idiomatic kotlin if we desire
* kotlin makes things like lombok obsolete, all kotlin objects automatically have getters and setters. 
* kotlin is from jetbrains so 1. it must be great and 2. it has great support in IDEA.
* its basically java without boilerplate and with easier ways to write funtional code. 
* kotlin makes a lot of discussions from our style guide obsolete.

for example in https://github.com/bisq-network/style/issues/12 instead of 
```
if (a) return X
else if (b) return Y
else return Z
```
you can just write 
```return if (a) X else if (b) Y else Z```

or for https://github.com/bisq-network/style/issues/11 vals are automatically constant which makes the discussion about final or not for local variables obsolete

If people like this idea i would just change the gradle file to allow kotlin sources too, and then we can transition as fast or as slow as we want. My suggestion would be that anytime we do a significant change to a class we first convert it to kotlin via the convert to kotlin function. Also gui classes and classes that use lombok are primary candidates to be converted.

I think reading and writing kotlin is just more fun.

freimair:
The only way I see this happening is by having integration test in place. https://github.com/bisq-network/proposals/issues/74

christophsturm:
i think converting the sourcode to kotlin on a file per file basis should not make any logic changes or require refactorings. Still we could start using kotlin for new code and for unit and integration tests.

christophsturm:
also we could start using kotlin for unit tests. there are really nice kotlin test libs
for example https://strikt.io which works with any test runner but has really great assertions. 
I also like minutest https://github.com/dmcg/minutest but that could probably be too much for some people

freimair:
> converting the sourcode to kotlin on a file per file basis should not make any logic changes

yes, but the code looks different and a reviewer has to make a decision. I see that only happening when test coverage comes up to speed and therefore can take the heat off the reviewer.

> Still we could start using kotlin for new code and for unit and integration tests.

agreed

battleofwizards:
I am super late to the project and **feel free to ignore my suggestions**.

[...]

PR #3115, opened 2019-08-20
Apply kotlin plugin and convert one unused class to kotlin

christophsturm:
just to check that everything works and to show how much simpler
kotlin code looks. see #3026 

cherry picked the kotlin part to a separate pr as requested by @freimair


build.gradle
```diff
@@ -173,6 +175,7 @@ configure(project(':common')) {
     }
 
     dependencies {
+        implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.3.41"
```
```suggestion
        implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion"
```

would you consider introducing a version variable for kotlin?

freimair (changes requested):

build.gradle
christophsturm: done

freimair (approved):
Ack

Issue #3116, opened 2019-08-20
Governance -> Browse Open Proposals -> NullPointerException


niyid:
I experienced an uncaught exception. Stack trace is below:

```
Aug-20 18:01:35.875 [JavaFX Application Thread] ERROR bisq.common.setup.CommonSetup: Uncaught Exception from thread JavaFX Application Thread 
Aug-20 18:01:35.875 [JavaFX Application Thread] ERROR bisq.common.setup.CommonSetup: throwableMessage= null 
Aug-20 18:01:35.876 [JavaFX Application Thread] ERROR bisq.common.setup.CommonSetup: throwableClass= class java.lang.NullPointerException 
Aug-20 18:01:35.916 [JavaFX Application Thread] ERROR bisq.common.setup.CommonSetup: Stack trace:
java.lang.NullPointerException
	at bisq.desktop.main.dao.governance.proposals.ProposalsView.lambda$activate$5(ProposalsView.java:212)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$9(PlatformImpl.java:418)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$10(PlatformImpl.java:417)
	at javafx.graphics/com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:96)
	at javafx.graphics/com.sun.glass.ui.gtk.GtkApplication._runLoop(Native Method)
	at javafx.graphics/com.sun.glass.ui.gtk.GtkApplication.lambda$runLoop$11(GtkApplication.java:277)
	at java.base/java.lang.Thread.run(Thread.java:844)
 
java.lang.NullPointerException
	at bisq.desktop.main.dao.governance.proposals.ProposalsView.lambda$activate$5(ProposalsView.java:212)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$9(PlatformImpl.java:418)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$10(PlatformImpl.java:417)
	at javafx.graphics/com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:96)
[...]

PR #3118, opened 2019-08-21
Prevent NullPointerException in ProposalsView

### LINKED ISSUE #2883: "java.lang.NullPointerException" when opening up Bisq
java.lang.NullPointerException

[bisq.log](https://github.com/bisq-network/bisq/files/3273138/bisq.log)

It had been a day or two since I closed Bisq and this occurred on my first attempt at reopening the app.

It seemed to be syncing the blockchain and taking a little longer than usual.

devinbileck:
When activating the ProposalsView, the table height is updated on the
user thread. Sometimes the scene is unavailable and causes a
NullPointerException. This should at least prevent the exception.

Fixes #2883 and #3116


ripcurlx (approved):
utACK

PR #3119, opened 2019-08-21
Add wiz's 2nd Bitcoin node

wiz:


ripcurlx:
@wiz Should this new node be publicly available already? I had issues connecting to it. Your other node works without any problems for me.

ripcurlx (approved):
ACK - I did have issues in the beginning using bitnodes.earn.com for testing connections. Testing it with the Bisq client seems to be fine.

wiz:
That site actually shows your node as "unreachable" as well, so I think it's an issue on their end. The node is ready for production use and is reachable on both clearnet and onion. Thanks for testing and merging.

PR #3121, opened 2019-08-21
Arbitrator to sign

sqrrm:


core/src/main/java/bisq/core/arbitration/BuyerDataItem.java
```diff
@@ -0,0 +1,60 @@
...
+                         Coin tradeAmount,
+                         PublicKey sellerPubKey) {
+        this.paymentAccountPayload = paymentAccountPayload;
+        this.accountAgeWitness = accountAgeWitness;
+        this.tradeAmount = tradeAmount;
+        this.sellerPubKey = sellerPubKey;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
christophsturm: I'm not sure, but can't lombok generate the equals and hash code methods?

btw, in kotlin this class would be just one line

core/src/main/java/bisq/core/account/sign/SignedWitnessService.java
```diff
@@ -295,4 +313,45 @@ private boolean verifyDate(SignedWitness signedWitness, long childSignedWitnessD
...
+                .filter(this::isBuyerWinner)
+                .map(this::getBuyerData)
+                .filter(Objects::nonNull)
+                .filter(buyerDataItem -> buyerDataItem.getAccountAgeWitness().getDate() < safeDate)
+                .distinct()
+                .collect(Collectors.toList());
+    }
+
+    private boolean hasChargebackRisk(Dispute dispute) {
+        return PaymentMethod.hasChargebackRisk(dispute.getContract().getPaymentMethodId());
```
christophsturm: we should stop calling static util methods. Maybe we can move some methods from PaymentMethod to a class that we can inject, or as temp solution just create a class that we can inject that forwards to the static method. (maybe a `ChargeBackRiskDetector`)

christophsturm:
see inline comments. also CI failed which is probably because the branch is not up to date with master.

desktop/src/test/java/bisq/desktop/GuiceSetupTest.java
```diff
@@ -131,6 +132,7 @@ public void testGuiceSetup() {
         assertSingleton(TradeEvents.class);
         assertSingleton(PriceAlert.class);
         assertSingleton(MarketAlerts.class);
+        assertSingleton(ChargeBackRisk.class);
```
christophsturm, line 135: ChargebackRisk is stateless so it does not need to be a singleton. of course it does not hurt to be a singleton, but I think guice best practice for stateless services is to keep them as default scope.

christophsturm:
ack

PR #3122, opened 2019-08-21
fix random test failure.

christophsturm:


battleofwizards (approved):

battleofwizards:
uACK

The fix seems to be safe. The `DefaultSeedNodeRepository` is ~~only ever instantiated once~~ in the production code so making the cache instance-level should not break anything.

Good catch, thanks for improving Bisq's tests!

ripcurlx:
I'm not 100% sure if there wouldn't be a problem for the monitor as the class is configured in the Core Module
https://github.com/bisq-network/bisq/blob/c9e6cd84fd94d4e8458183e1a2e42b7da67a478f/core/src/main/java/bisq/core/CoreModule.java#L69
and manually instantiated in the P2PNetworkLoad as well https://github.com/bisq-network/bisq/blob/c9e6cd84fd94d4e8458183e1a2e42b7da67a478f/monitor/src/main/java/bisq/monitor/metric/P2PNetworkLoad.java#L147

Maybe @freimair could have an extra look on this?

battleofwizards:
Indeed that seems like a problem. The `DefaultSeedNodeRepository` is a `@Singleton` but that is only enforced by Guice and not when instantiating manually. So there could be more instances flying around.

It feels like combining automatic and manual instantiation of a supposedly `@Singleton` class is a problem on its own.

christophsturm:
I think its not ideal that the monitor does not use guice, but in its manual wiring its still only creating one instance. also thats not really related to this fix.

ripcurlx:
> I think its not ideal that the monitor does not use Guice, but in its manual wiring its still only creating one instance. also thats not really related to this fix.

I looked at the code again It is true that it shouldn't be a problem as the Guice autowiring is kind of overwritten in all used classes in that use-case by manual instantiation. So no matter if the Guice autowiring wins over the manual instantiation or the other way round it shouldn't cause any issues.

so utACK

freimair (commented):
just had a look and even if there are 2 caches, it does not hurt anything because the monitor is only run once or twice.

Ack.

Issue #2712, opened 2019-04-15
Allow text input in all currency drop downs


huey735:
Right now, only when creating a new altcoin account one is able to type the name of the currency to select it from the long drop down options. This option should be expanded to other drop down currency menus in the software.

ripcurlx:
Yes, we could have a look where else it makes sense to add this functionality.

ManfredKarrer:
Yes that would be definitely good to have at all currency drop downs. We have the code in tha altcoin account setup.

shamoons:
I'll take a look and start working on this

huey735:
@shamoons @battleofwizards @ripcurlx 
This function has been added to the all the Market dropdown menus except for the Settings > Preferences.

battleofwizards:
@huey735 yes, that is correct.

The first PR only introduced autocomplete on major tabs to limit the volume of changes.

I intend to apply autocomplete more widely in the application over the coming weeks.

PR #3124, opened 2019-08-21
Fix bugs in SearchComboBox (improve autocomplete)

battleofwizards:
Bisq's `SearchComboBox` implements a simple autocomplete. It is used on the "add new altcoin account" form.

This fixes the following bugs in the `SearchComboBox`:

* no more dropdown flickering (fixed hide+show hack)
* space key no longer resets the text input
* unconfirmed changes no longer affect the form
* empty list no longer renders empty dropdown

This PR is the first step towards using autocomplete more widely in the project. Once merged I will attempt to address https://github.com/bisq-network/bisq/issues/2712 (but can't promise anything).


ripcurlx (approved):
ACK - Just tested it and its looking good 👍

PR #3126, opened 2019-08-22
Refactor checkMaxConnections

devinbileck:
Fix connection limit checks so as to prevent the following warning:

> WARN  b.n.p2p.peers.PeerManager: No candidates found to remove (That
case should not be possible as we use in the last case all
connections).


mrosseel:
this might be a good opportunity to add a unit test for PeerManager (possibly limited to the scope of connection handling)

ripcurlx (dismissed):
utACK

devinbileck:
> this might be a good opportunity to add a unit test for PeerManager (possibly limited to the scope of connection handling)

It seems the PeerManagerTest is ignored and in need of an overhaul.

ripcurlx:
Could someone of the exisiting contributors try out this branch before merging, as it would cause big problems if some use-case handling is wrong. I just read through the code and it looks fine, but it would be good to give this a local test as well.

ManfredKarrer:
I think that code part is really a good candidate for a test. The logic is a bit involved and bigger changes here are tough to test manually. I would postpone otherwise merge until someone had dedicated enough time to be sure all is correct.

devinbileck:
Agreed. I will try to look into implementing some tests. Just not familiar with this area so will need some time to ramp up.

ripcurlx:
> Agreed. I will try to look into implementing some tests. Just not familiar with this area so will need some time to ramp up.

@devinbileck Are you still on this PR?

devinbileck:
I started looking into it but didn't make much progress. I haven't focused on it very much lately. If someone else is able to contribute something, feel free. Otherwise I will try to get to it.

ripcurlx:
@devinbileck Could you please resolve the conflict and add some tests?
@freimair Could you review the code?

devinbileck:
I have added some tests. I will be committing them shortly.

devinbileck:
@ripcurlx @freimair ready for review.

ripcurlx:
@freimair As I haven't touched this part of the codebase so far, could you please do the review? Thanks!

build.gradle
```diff
@@ -205,6 +205,7 @@ configure(project(':common')) {
 configure(project(':p2p')) {
     dependencies {
         compile project(':common')
+        testCompile project(':core')
```
```suggestion

```

p2p/src/test/java/bisq/network/p2p/MockNode.java
```diff
@@ -0,0 +1,86 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.network.p2p;
+
+import bisq.core.network.p2p.seed.DefaultSeedNodeRepository;
```
freimair:
```suggestion
import bisq.network.p2p.seed.SeedNodeRepository;
```

p2p/src/test/java/bisq/network/p2p/MockNode.java
```diff
@@ -0,0 +1,86 @@
...
+    @Getter
+    public Set<Connection> connections;
+    @Getter
+    public int maxConnections;
+
+    public MockNode(int maxConnections) {
+        this.maxConnections = maxConnections;
+        networkNode = mock(NetworkNode.class);
+        Storage<PeerList> storage = new Storage<>(mock(File.class), mock(PersistenceProtoResolver.class), mock(CorruptedDatabaseFilesHandler.class));
+        peerManager = new PeerManager(networkNode, mock(DefaultSeedNodeRepository.class), new ClockWatcher(), maxConnections, storage);
```
freimair:
```suggestion
        peerManager = new PeerManager(networkNode, mock(SeedNodeRepository.class), new ClockWatcher(), maxConnections, storage);
```

freimair (changes requested):
IMO we could get away with mocking the `SeedNodeRepository` superclass, thus eliminating the dependency to `core`. Can you verify that this is correct?

devinbileck:
Yes, good point. Updated!

freimair (approved):

PR #3127, opened 2019-08-22
Fix incorrect DAO full node check when loading PreferencesView

devinbileck:
When loading the preferences view, it would always log a warning
indicating that DAO full node is selected without providing required
parameters even though it wasn't selected.


ManfredKarrer (approved):
utACK

ripcurlx (approved):
utACK

PR #3131, opened 2019-08-23
fix logging of peer address

christophsturm:
fix a small problem that i noticed when looking at the log file:
don't log peer as "Optional [....onion]"


ripcurlx (approved):
utACK

PR #3133, opened 2019-08-23
Increase top navigation button text size for Japanese locale

wiz:
Before / After screenshots
bisq1
bisq2


battleofwizards:
Minor suggestion: consider renaming the `.nav-button-large` to `.nav-button-for-japanese-locale`. 

The class name as it is now suggests some broader use or importance and so is misleading.

Of course, if other languages turn out to need this larger nav, the name may be further changed to be a bit more generic, like `.nav-button-for-asian-locales`.

Thanks!

ManfredKarrer (approved):
utACK

wiz:
@battleofwizards good point, I've changed it to `-japanese`, leaving it up to other CJK languages to tweak the font-size on their own

PR #3134, opened 2019-08-23
Validate phone numbers

ghubstan:
Created new PhoneNumberValidator + Test.  Validator hides
ISO country code for validating inputs.

After successful validation, inputs are transformed into E.164
format and cached in a field, accessible via a getter method.
However, area and region codes are not checked for correctness.
End users are responsible for correctness of area/region codes.

Four i18n properties (validation.phone.*) were added to
displayStrings.properties

This is a partial fix for Issue bisq-network#3042.  The validator
will be integrated into the GUI if PR is approved & patch is
merged.

Question for Bisq devs:
Another new pkg protected class, CountryCallingCodes, contains
an immutable map of ISO country codes -> country calling codes;
it is the only non-validator class in its package.  The map
declaration is too large to live in the PhoneNumberValidator class,
even after tweaking idea.properties in an attempt to prevent
the IDE from freezing.  Is it OK to leave CountryCallingCodes
where it is, or is there a more appropriate home for it?


desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * Phone number in E.164 format.
+     */
+    @Nullable
+    @Getter
+    private String normalizedPhoneNumber;
+
+    // Hide no-arg constructor
+    private PhoneNumberValidator() {
+        this.isoCountryCode = null;
```
battleofwizards, line 46: All these attributes are guaranteed null. Assigning null is superflous.

desktop/src/main/java/bisq/desktop/util/validation/CountryCallingCodes.java
```diff
@@ -0,0 +1,278 @@
...
+    /**
+     * Immutable mapping of ISO 3166 alpha-2 country code to national dialing number.
+     * <p>
+     * In some regions, such as American Samoa ('AS'), there is only one area code, and
+     * it is included in the mapping as part of the calling code.
+     *
+     * @see {@link <a href="https://en.wikipedia.org/wiki/E.164">https://en.wikipedia.org/wiki/E.164</a>}
+     * @see {@link <a href="https://en.wikipedia.org/wiki/List_of_country_calling_codes">https://en.wikipedia.org/wiki/List_of_country_calling_codes</a>}
+     */
+    private final Map<String, String> callingCodeMap = Map.ofEntries(
```
battleofwizards: This should be static unless there is specific reason against.

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
```diff
@@ -0,0 +1,143 @@
...
+        this.countryCallingCodes = null;
+        this.callingCode = null;
+        this.normalizedCallingCode = null;
+    }
+
+    public PhoneNumberValidator(String isoCountryCode) {
+        this.isoCountryCode = isoCountryCode;
+        this.countryCallingCodes = new CountryCallingCodes();
```
battleofwizards: Why do we keep an instance here? The CountryCallingCodes is essentially a global immutable resource.

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+
+    @Override
+    public ValidationResult validate(String input) {
+        normalizedPhoneNumber = null;
+        ValidationResult result = super.validate(input);
+        if (!result.isValid) {
+            return result;
+        }
```
battleofwizards, line 63: I might be wrong here but it feels like the dominating convention in the project is to skip `{}` for single statement blocks.

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
```diff
@@ -0,0 +1,143 @@
...
+        final String trimmedInput = input.trim();
```
battleofwizards: A matter of personal taste but I suggest instead of making local variables `final` make your methods very tiny. Then there is no need for the `final` keyword boilerplate.

desktop/src/test/java/bisq/desktop/util/validation/PhoneNumberValidatorTest.java
```diff
@@ -0,0 +1,291 @@
...
+    @Before
+    public void setup() {
+        Res.setup();
+    }
+
+    @Test
+    public void testNoInput() {
+        validator = new PhoneNumberValidator("AT");
+        validationResult = validator.validate("");
+        assertFalse("'' should not be a valid number in AT", validationResult.isValid);
```
battleofwizards, line 27: It really helps with test readability to define custom assertions. In this case `assertInvalid`.

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
```diff
@@ -0,0 +1,143 @@
...
+        boolean isCountryDialingCodeExplicit = trimmedInput.startsWith("+");
+        // remove non-digit chars:  plus sign, dashes, parens, hashes, slashes and white spaces
+        String pureNumber = input.replaceAll("[\\+\\s+()\\-\\−\\#\\\\]", "");
```
battleofwizards: Instead of removing specific chars remove everything that is not a digit: `^\\d` (regexp not tested).

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
```diff
@@ -0,0 +1,143 @@
...
+        boolean hasValidCallingCodePrefix = pureNumber.startsWith(normalizedCallingCode);
+        boolean isCountryCallingCodeImplicit = !isCountryDialingCodeExplicit && hasValidCallingCodePrefix;
+
+        result = validateIsNumeric(input, pureNumber)
+                .and(validateHasSufficientDigits(input, pureNumber))
+                .and(validateIsTooLong(input, pureNumber))
```
battleofwizards: Maybe `validateIsNotTooLong`? Because other validations are positive.

battleofwizards (commented):
Big plus for the test cases and nice meaningful names!

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
ghost: There are some style reco at https://github.com/bisq-network/style

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
ghubstan: OK.  I thought all final instance variables had to be initialized in a constructor, and in this class, the no-arg constructor had to be hidden (private) to prevent another class from using it.  In the former case, maybe it was true in an earlier JDK?  In the latter, I'm still not sure it was a good idea to remove the private no-arg constructor, but I followed your advice anyway.

desktop/src/main/java/bisq/desktop/util/validation/CountryCallingCodes.java
ghubstan: OK

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
ghubstan: OK, did does seem to be the convention.

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
ghubstan: OK.  Since working with Scala,  making Java vars that should never change value immutable became a habit.

desktop/src/test/java/bisq/desktop/util/validation/PhoneNumberValidatorTest.java
ghubstan: I didn't create a custom assertion; I think test names and framework's existing assertions are sufficient.

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
ghubstan: OK

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
ghubstan: OK

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
ghubstan: Letters need to be included in the pureNumber string for the isPositiveNumber test.  I had a US number with letters, and I'm sure they still exist in other countries.  A number with letters may be valid, but the user will be prompted to convert them to  numbers for successful validation.

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
battleofwizards: Sorry! I didn't mean to imply removing private constructor. I do agree it is important here.

So please either revert back to assigning nulls **or** simply get rid off the problematic final. In the context of clean code (small methods and classes) Java finals are overrated. They only protect the reference anyway and not the object's content.

Whatever you pick it will be OK with me.

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
battleofwizards: It does make perfect sense in Scala b/c vals are actually immutable. In Java final objects are mutable. It is only the reference that cannot be reassigned. Also `final` is longer than `val` so more visual clutter. Anyway, congrats for working with Scala!

desktop/src/test/java/bisq/desktop/util/validation/PhoneNumberValidatorTest.java
battleofwizards: Most importantly it is very nice you provided comprehensive unit tests. I am perfectly okay with your original version as it is already superior to what we have in the project (tests missing).

But please consider: tests like any code should be about intent and essence, not technical details. Good tests rarely use raw assertions directly. Framework assertions are just the building blocks for your meaningful ones.

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
battleofwizards: Sure but that does not change the essence. Remove everything that is not digit-or-letter. That is better approach than haunting for specific chars. Try `"[^A-Za-z0-9]"` and adjust regarding expected capitalization.

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
ghubstan: But these are Java primitives.  Once set, final Strings and primitives are immutable.  (Class level final primitives must be set in in constructors, as I was doing in the hidden no-arg constructor.)   There are also tricks to create an immutable map (see CountryCallingCodes).

There are other scenarios where final primitives can be reassigned, such as ...

final int[] x = int[0];
function or anonymous inline class -->
  x[0] = 1;
  x[0] = 100;  
  etc...

battleofwizards: I was referring to objects in general to explain why local finals are overkill. Also, it wouldn't make sense to use local finals selectively depending on the class (im)mutability.

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
ghubstan: Yeah, "[^A-Za-z0-9]" is a much better regex for this purpose.  I'll make that change soon, but maybe not tonight.

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
ghubstan: Reverted back to using braces in if() {single line} blocks.

desktop/src/main/java/bisq/desktop/bisq.css
```diff
@@ -804,6 +804,10 @@ bg color of non edit textFields: fafafa
     -fx-fill: -bs-rd-nav-deselected;
 }
 
+.nav-button-japanese .text {
```
ripcurlx, line 807: Why is this change part of this PR? I think something got wrong during rebasing of your branch.

desktop/src/main/java/bisq/desktop/main/MainView.java
```diff
@@ -809,6 +809,10 @@ private void setupBadge(JFXBadge buttonWithBadge, StringProperty badgeNumber, Bo
 
             this.setToggleGroup(navButtons);
             this.getStyleClass().add("nav-button");
+            // Japanese fonts are dense, increase top nav button text size
```
ripcurlx, line 812: Why is this change part of this PR? I think something got wrong during rebasing of your branch.

ripcurlx (changes requested):
NACK - based on my comments. Please update your PR with the required changes and I'll give it a spin 👍

desktop/src/main/java/bisq/desktop/bisq.css
ghubstan: It happened while unnecessarily syncing that branch in with master.  A clean PR #3149 replaces this.

desktop/src/main/java/bisq/desktop/main/MainView.java
ghubstan: It happened while unnecessarily syncing this branch with master.    The clean, replacement PR #3149 should be ready to test after CI.

PR #3136, opened 2019-08-24
Changed "Portfolio" to "Orders"

RajaMBZ:
The word "Portfolio" makes absolutely no sense at the menu when it's only supposed to show the open/closed orders. The portfolio usually refers to the distribution of your OWN funds, that you're not trading. It makes no sense at this point and it should be renamed as "Orders".


ManfredKarrer (changes requested):
NACK
Trades are not orders and we intentionally do not use the domain language from centralized exchanged where orders get executed by a matching engine. Bisq does not work like that. Also what you edited is not the app code but a demo class as the name suggests.

RajaMBZ:
Hmmm... it'll remain quite confusing for anyone who jumps from a centralized exchange to Bisq, but anyways, you do have a good point there.

PR #3142, opened 2019-08-27
Suppress RPC service log entries when running as DAO lite node

devinbileck:
When running as DAO lite node, the following would get logged:

> INFO  b.c.dao.node.full.RpcService: The value of rpchost is
> INFO b.c.dao.node.full.RpcService: Starting RPCService with btcd-cli4j
version 0.5.8.4 on 127.0.0.1:8332 with user null

So instead of logging it when instantiating the RPCService, log it when
setting up the service.


ripcurlx (approved):
utACK

PR #3148, opened 2019-08-27
Improve TempProposal processing (fixes #3143)

wiz:

Issue #3143, opened 2019-08-27
Node becomes unresponsive causing all peers to be dropped


wiz:
Here is one example where node was completely unresponsive for 30+ seconds while processing these events

```
Aug-27 08:01:28.521 [JavaFX Application Thread] INFO  b.c.d.g.p.ProposalService: We received a remove request for a TempProposalPayload and have removed the proposal from our list. proposal creation date=Sun Jun 09 02:06:48 JST 2019, proposalTxId=adfe3550b36d8d644fb600ec43b9eeb1ebbc03946631a2d8c93811812e61c3f8, inPhase=true, txInPastCycle=true, unconfirmedOrNonBsqTx=false 
Aug-27 08:01:29.165 [JavaFX Application Thread] INFO  b.c.d.g.p.ProposalService: We received a remove request for a TempProposalPayload and have removed the proposal from our list. proposal creation date=Wed Jun 05 04:17:02 JST 2019, proposalTxId=dc561b2cf7ebc1636c986db8d710e14e01e0072bdf1a0b621094d3c2ce4a4276, inPhase=true, txInPastCycle=true, unconfirmedOrNonBsqTx=false 
Aug-27 08:01:29.782 [JavaFX Application Thread] INFO  b.c.d.g.p.ProposalService: We received a remove request for a TempProposalPayload and have removed the proposal from our list. proposal creation date=Tue May 07 06:46:12 JST 2019, proposalTxId=583d4bc4e9bd7d1b0b6bfa47ce542dd33deea9febd7e26ac31de0ad18b39e9fb, inPhase=true, txInPastCycle=true, unconfirmedOrNonBsqTx=false 
Aug-27 08:01:30.386 [JavaFX Application Thread] INFO  b.c.d.g.p.ProposalService: We received a remove request for a TempProposalPayload and have removed the proposal from our list. proposal creation date=Wed May 08 06:19:09 JST 2019, proposalTxId=1fbc48fc1052287dcd33da555184a3e095bd089244b2375cf1a4a3afe67c2367, inPhase=true, txInPastCycle=true, unconfirmedOrNonBsqTx=false 
[...]

PR #3150, opened 2019-08-27
Update data storage on expired data removal

wiz:
Also related to earlier #3148 and #3143 - datastore was not updated on remove calls for expired data


ManfredKarrer (approved):
ACK

PR #3151, opened 2019-08-27
Do not send BSQ blocks to full nodes

chimp1984:


ripcurlx:
Seems to make sense to me. @sqrrm @ManfredKarrer could either of you ACK this PR if you also think it is ok?

ManfredKarrer (approved):
ACK

PR #3153, opened 2019-08-27
Add backup operator for seeds 3f3cu2yw7u457ztq and fl3mmribyxgrv63c

devinbileck:


ManfredKarrer (commented):
utACK

ripcurlx (changes requested):
NACK - I think it is better to use my GitHub handle in that case: @ripcurlx

devinbileck:
Fair enough - updated. I wasn't sure if the intention is to tie it to a GitHub account or Slack account in order to notify in case of issue.

ripcurlx (approved):
utACK

PR #3155, opened 2019-08-28
Merge P2P bugfixes into seednode branch to help node stability issues

wiz:


ManfredKarrer (approved):
ACK

core/src/main/java/bisq/core/dao/governance/proposal/ProposalListPresentation.java
```diff
@@ -75,13 +87,30 @@ public ProposalListPresentation(ProposalService proposalService,
...
+    @Override
+    public void addListeners() {
+    }
+
+    @Override
+    public void start() {
+        // We must set the listeners initially and not on onParseBlockChainComplete as activeOrMyUnconfirmedProposals
+        // is used in voteResults which can be called earlier during sync.
+        // To avoid unneeded upDateLists calls we delay one render frame so that once the proposalService is complete we
+        // register out listeners.
```
freimair, line 110:
```suggestion
        // register our listeners.
```

core/src/main/java/bisq/core/dao/governance/proposal/ProposalListPresentation.java
```diff
@@ -75,13 +87,30 @@ public ProposalListPresentation(ProposalService proposalService,
...
+
+    @Override
+    public void addListeners() {
+    }
+
+    @Override
+    public void start() {
+        // We must set the listeners initially and not on onParseBlockChainComplete as activeOrMyUnconfirmedProposals
+        // is used in voteResults which can be called earlier during sync.
+        // To avoid unneeded upDateLists calls we delay one render frame so that once the proposalService is complete we
```
freimair, line 109:
```suggestion
        // To avoid unneeded updateLists(); calls, we delay one render frame so that once the proposalService is complete, we
```

common/src/main/java/bisq/common/app/Capability.java
```diff
@@ -31,7 +31,7 @@
     PROPOSAL,
     BLIND_VOTE,
     ACK_MSG,
-    BSQ_BLOCK,
+    RECEIVE_BSQ_BLOCK,
```
freimair, line 34: I am not sure we should change stuff like that. What I am sure of is that we should add Javadoc to describe what a capability is for and what it does, only so can we make sure devs can understand what these 2 or 3 words mean.

freimair (approved):
utAck

I am not sure if we should keep adding code and complexity to the project like this. Sure, there is an issue there that needs fixing, hence, utack. However, it seems to me that there is a simpler solution just around the corner.

wiz:
@freimair all of these commits were already merged into master in PRs #3148, #3150, and #3151 - this PR is only for the `seednode_temporary` branch, so if you want to make changes please make a PR that branches off master because this branch will probably go away soon.

sqrrm (approved):
utACK

PR #3157, opened 2019-08-28
Group burned BSQ by day

ripcurlx:
As only the issuance is taking place roughly once a month, I've changed the grouping of burned BSQ to daily numbers.
Bildschirmfoto 2019-08-28 um 11 15 59


ManfredKarrer (commented):
utACK

PR #3164, opened 2019-08-29
Set active currencies as high chargeback risk

sqrrm:
Specify the currencies that are considered high risk for chargeback based on volume. As new markets grow we will probably have to add them as well, but it might be that some currencies don't have as much chargeback risk.

By allowing new markets to trade without any restrictions they get a chance to grow without too much pain of bootstrapping signed account witnesses.

### DIFF
--- a/core/src/main/java/bisq/core/account/sign/SignedWitnessService.java
+++ b/core/src/main/java/bisq/core/account/sign/SignedWitnessService.java
@@ -26,6 +26,7 @@ import bisq.core.arbitration.DisputeManager;
 import bisq.core.arbitration.DisputeResult;
 import bisq.core.payment.ChargeBackRisk;
 import bisq.core.payment.payload.PaymentAccountPayload;
+import bisq.core.payment.payload.PaymentMethod;
 
 import bisq.network.p2p.P2PService;
 import bisq.network.p2p.storage.P2PDataStorage;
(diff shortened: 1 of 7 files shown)


ManfredKarrer (approved):
utACK

huey735:
If the AUD market is in that list so should the BRL as it has been more active recently, and If I'm not wrong already surpassed AUD in trade numbers and volume.

What is the list of payment methods that remain under this limit list?

ripcurlx:
> If the AUD market is in that list so should the BRL as it has been more active recently, and If I'm not wrong already surpassed AUD in trade numbers and volume.
> 
> What is the list of payment methods that remain under this limit list?

True - the AUD/BTC market is not as big as the BRL right now.

AUD/BTC
Bildschirmfoto 2019-08-30 um 11 07 38

BRL/BTC
Bildschirmfoto 2019-08-30 um 11 07 47

@ManfredKarrer Do you remember if we had scam issues in the AUD market already?
If not, I agree with @huey735 that we might not need to add AUD as well or add BRL if we have the feeling that it might be too risky to exclude it from our current limit.

@huey735 As most limited payment methods only support one currency (e.g. EUR, USD) the ones that have this multiple currency options and are limited right now are:  National bank transfer, same bank transfer, specific bank transfer, Revolut and Uphold.

ManfredKarrer:
I would prefer to be cautious than sorry. So maybe add BRL as well as it has significantly enough volume for a scammer to be targeted.

ripcurlx:
> I would prefer to be cautious than sorry. So maybe add BRL as well as it has significantly enough volume for a scammer to be targeted.

@sqrrm Could you add BRL to the list of countries as well?

ripcurlx (changes requested):
NACK - Please add BRL as well as suggested by @huey735

ripcurlx (approved):

PR #3165, opened 2019-08-29
Traders chat

chimp1984:
Based on PR  #2871 from @sqrrm.

I share the PR so that others can add optimisations for design and wording.

Open questions:
- Do we need a flag in the settings to allow the user to not receive chat messages? Would require that the peer gets that information as well which is a bit of an effort. Maybe we can keep that for a later version if needed at all.
- Do we need a popup or "news" badge to make sure users will find the new feature? I guess it is obvious enough, specially if you get a message.
- Do we want a badge icon showing new messages if the selected screen is not pending trades?
- Do we want to send mobile notifications?

I think all the above is not needed, at least not for a first release. But some UI beautification (@ripcurlx) and better wording (@m52go) would be nice...

Screen Shot 2019-08-30 at 01 13 20


battleofwizards:
> I think all the above is not needed, at least not for a first release.

I agree. None of the mentioned improvements sound critical for the first release. Better to release the thinnest vertically complete features possible.

ripcurlx:
@chimp1984 Looking great already! I'll give it a spin and provide feedback soon.

ripcurlx:
I fixed the badge positioning and the tooltip font size. I'd also suggest following changes to the icon: Using a material design version of the chat icon and color it in our action color (green), which is more consistent throughout the app.
Bildschirmfoto 2019-08-30 um 12 37 25

ripcurlx:
I just did more in-depth testing and everything works as expected. Also having arbitration case in parallel to the trader chat for a specific message doesn't cause any issues. Great work @sqrrm and @chimp1984 👍 
@m52go Could you review the text content if you want to have some changes there?

If that is done I'm happy to merge this PR to master.

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -608,6 +608,18 @@ portfolio.pending.step2_seller.waitPayment.msg=The deposit transaction has at le
 portfolio.pending.step2_seller.warn=The BTC buyer still has not done the {0} payment.\nYou need to wait until they have started the payment.\nIf the trade has not been completed on {1} the arbitrator will investigate.
 portfolio.pending.step2_seller.openForDispute=The BTC buyer has not started his payment!\nThe max. allowed period for the trade has elapsed.\nYou can wait longer and give the trading peer more time or contact the arbitrator for opening a dispute.
 
+tradeChat.chatWindowTitle=Chat window for trade with ID ''{0}''
+tradeChat.openChat=Open chat window
+tradeChat.rules=You can communicate with your trade peer to resolve potential problems with this trade. It is not mandatory to reply in \
```
```
tradeChat.rules=Here you can communicate with your trading peer to resolve potential problems with this trade.\n\n\
It is not mandatory to reply to any messages.\n\nIf a trader violates any of the rules below, open a dispute with \
'Ctrl/Cmd + o' and report it to the arbitrator.\n\n\
```

Main updates here: 

- put "it is not mandatory to reply to any messages" on its own line so it is crystal clear
- Add 'Ctrl/Cmd' for non-Mac users

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -608,6 +608,18 @@ portfolio.pending.step2_seller.waitPayment.msg=The deposit transaction has at le
...
+  the chat. If a trader violates the below rules, open a dispute with 'cmd + o' and report it to the arbitrator.\n\n\
+  Chat rules:\n\
+  \t● Do not send links to unknown URLs (risk of malware). Links to a block explorer are permitted.\n\
```
m52go: Should we also expressly disallow link shorteners?

Separately: even bad links can be disguised to look like known URLs with weird TLDs. I think all links should be disallowed, but maybe I'm paranoid.

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -608,6 +608,18 @@ portfolio.pending.step2_seller.waitPayment.msg=The deposit transaction has at le
...
+  \t● Do not try to motivate peer to trade outside of Bisq (no security).\n\
```
```suggestion
  \t● Do not encourage trading outside of Bisq (no security).\n\
```

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -608,6 +608,18 @@ portfolio.pending.step2_seller.waitPayment.msg=The deposit transaction has at le
...
+  \t● Keep conversation friendly and respectful.\n\
+  \t● Do not engage in any form of social engineering scam attempts.\n\
+  \t● Respect if the peer is not responding and prefers to not communicate via the chat.\n\
```
```suggestion
  \t● If a peer is not responding and prefers to not communicate via chat, respect their decision.\n\
```

m52go (commented):
Text looks good to me overall...just a potential issue with links (see comment).

core/src/main/resources/i18n/displayStrings.properties
m52go: I would also add an additional rule, as it comes up quite a bit, but is a huge problem for privacy:

```
\t● Do not attempt to contact the trading peer outside of this chat.\n\
```

wiz: Do not copy and paste your seed words, private keys, passwords, or other sensitive information into this chat!

core/src/main/resources/i18n/displayStrings.properties
chimp1984: Yes, good point with link shortener. I think it should be added. Maybe also recommend to use Tor browser as that is more safe with deactivated js etc...

chimp1984: Right. Actually beside links to the tx at a block-explorer no links would be needed. And those can be looked up by the other used with the tx id as well. So right maybe disallow any URLs?

m52go: > So right maybe disallow any URLs?

Yes, I would say so, unless anyone has a solid reason to keep them.

```suggestion
  \t● URLs are not allowed. If you would like to send a link to a block explorer, for example, just send the transaction ID, and the other peer can look it up on their own.\n\
```

core/src/main/resources/i18n/displayStrings.properties
chimp1984: > Do not attempt to contact the trading peer outside of this chat

Not so sure if that should be a strict rule as in cases of network instability users do communicate via email if the email is known. I would say lets leave that to common sense of users...

chimp1984: Updated version:

```
tradeChat.rules=You can communicate with your trade peer to resolve potential problems with this trade.\n\
  It is not mandatory to reply in the chat.\n\
  If a trader violates the below rules, open a dispute with 'Cmd/Ctrl + o' and report it to the arbitrator.\n\n\
  Chat rules:\n\
  \t● Do not send any links (risk of malware). You can send the transaction ID and the name of a block explorer.\n\
  \t● Do not send your seed words, private keys, passwords or other sensitive information!\n\
  \t● Do not encourage trading outside of Bisq (no security).\n\
  \t● Do not engage in any form of social engineering scam attempts.\n\
  \t● If a peer is not responding and prefers to not communicate via chat, respect their decision.\n\
  \t● Keep conversation scope limited to the trade. This chat is not a messenger replacement or trollbox.\n\
  \t● Keep conversation friendly and respectful.
```

ripcurlx (approved):
ACK

desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/PendingTradesView.java
```diff
@@ -316,6 +316,19 @@ private void openChat(Trade trade) {
         if (chatPopupStage != null)
             chatPopupStage.close();
 
+        if (trade.getCommunicationMessages().isEmpty()) {
```
sqrrm, line 328: This will make receivers of messages from peers not see the code of conduct message

PR #3167, opened 2019-08-30
Do not specify indirect module dependencies

battleofwizards:
One is supposed to only specify direct dependencies in `build.gradle`. This commit removes two superfluous deps statements for the `desktop` module. Other modules already follow the convention.

This does **not** remove any dependencies - it only cleans up how they are specified.

The build does work fine after the change and `desktop` app does run as expected.


ManfredKarrer (approved):
utACK

PR #3169, opened 2019-08-30
[WIP] Add mediator support

chimp1984:
This is heavily work in progress but I thought it might be useful for other developers to share the development progress.

For the new trade protocol we need a mediator which will be very similar to arbitration just that mediators do not have a 3rd key and cannot do a payout but only give the traders a recommendation for the payout. If both traders agree to that result they can complete the trade by both signing the payout tx. Only if they do not find consensus they need to go to the next step which is currently the arbitration. 
So before traders can open a dispute to arbitrators they need to go through the mediation first and only if both do not agree to the mediators decision arbitration will be used.

This should reduce arbitration cases a lot and is a necessary feature for the new trade protocol but can be deployed as addition with the current protocol.

To make sure the feature is available for both traders (both have updated) we can use an activation date and set a mandatory version number after a few days after release, so we enforce traders to update and enable that feature. Otherwise it would be complicated if one peer has updated but the other not...


core/src/main/java/bisq/core/dispute/DisputeResolverService.java
```diff
@@ -0,0 +1,127 @@
...
+            if (!map.containsKey(disputeResolverNodeAddress))
+                map.put(disputeResolverNodeAddress, disputeResolver);
+            else
+                log.warn("disputeResolverAddress already exist in disputeResolver map. Seems an disputeResolver object is already registered with the same address.");
+        }
+        return map;
+    }
+
+    @NotNull
+    protected abstract Set<T> getCollect(List<String> bannedDisputeResolvers);
```
sqrrm: `getCollect()` is not a great name, `getSet()` might be better

desktop/src/main/java/bisq/desktop/main/account/register/DisputeResolverRegistrationView.java
```diff
@@ -0,0 +1,270 @@
...
+        revokeButton.setOnAction(e -> onRevoke());
+
+        final TitledGroupBg titledGroupBg = addTitledGroupBg(gridPane, ++gridRow, 2,
+                Res.get("shared.information"), Layout.GROUP_DISTANCE);
+
+        titledGroupBg.getStyleClass().add("last");
+
+        Label infoLabel = addMultilineLabel(gridPane, gridRow);
+        GridPane.setMargin(infoLabel, new Insets(Layout.TWICE_FIRST_ROW_AND_GROUP_DISTANCE, 0, 0, 0));
+        infoLabel.setText(Res.get("account.arbitratorRegistration.info.msg"));
```
sqrrm: This info text is not generic enough for both mediator and arbitrator

core/src/main/java/bisq/core/dispute/DisputeResolverManager.java
```diff
@@ -100,32 +99,16 @@ public DisputeResolverManager(KeyRing keyRing,
...
-                ));
+        publicKeys = useDevPrivilegeKeys ? Collections.singletonList(DevEnv.DEV_PRIVILEGE_PUB_KEY) : getPubLeyList();
     }
 
 
     ///////////////////////////////////////////////////////////////////////////////////////////
     // Abstract methods
     ///////////////////////////////////////////////////////////////////////////////////////////
 
+    protected abstract List<String> getPubLeyList();
```
sqrrm: Should be getPub**K**eyList

sqrrm (commented):
Some minor comments

core/src/main/java/bisq/core/dispute/DisputeResolverService.java
ManfredKarrer: Upps, was not intended. Will change it.

chimp1984: Upps, was not intended. Will change it.

desktop/src/main/java/bisq/desktop/main/account/register/DisputeResolverRegistrationView.java
chimp1984: Ah good point. Wanted to ask @ripcurlx how it is for translators if we change key of string. I fear they will see he old key removed and need to add the new one again. That's why I left it currently, but I will add a TODO.

core/src/main/java/bisq/core/dispute/DisputeResolverManager.java
chimp1984: Ups, will fix it.

chimp1984:
Rebased with current master.

chimp1984:
I will close that branch as I run into merge issues I cannot resolve ;-(. Did some master merge and rebase locally without keeping in sync with remote and now any merge attempt fails. Was heavy refactorings with many moved and removed classes, so some complexity there screwed it up...

PR #3170, opened 2019-08-30
Fix bug with not showing system msg at peer who receives msg

chimp1984:
- We only added the system msg when one opens the chat but not when
one received a message.


sqrrm (approved):
ACK

PR #3174, opened 2019-09-01
Remove Byteball in order to remove commons-codec.

blabno:
This asset is not traded anymore.


blabno:
Well, it seems that p2p module depends on `org.apache.httpcomponents:httpclient` which in turn depends on `commons-codec`.

chimp1984 (approved):
utACK

chimp1984:
> Well, it seems that p2p module depends on org.apache.httpcomponents:httpclient which in turn depends on commons-codec.

Ah damn. But remove it anyway, better to get one explicit dependency removed...

battleofwizards:
Once we migrate to Java 11+ there is hope with the new standard HttpClient:
http://cr.openjdk.java.net/~chegar/httpclient/02/javadoc/api/java.net.http/java/net/http/package-summary.html
Maybe one day it could replace the Apache HttpClient.

PR #3177, opened 2019-09-01
Fix performance issue in BsqWalletService

### LINKED ISSUE #3175: Startup problems
In case of an application which have not been opened for a while it happens often that the application freezes. After some investiagtion with an old app I could reproduce it and <strike>I found where the issue is likely caused (TradeStatistic and AccountAgeWitness objects - there are a lot if the user have not been online for a while)</strike>. I still have not found the exact cause but hopefully will soon...

chimp1984:
The updateBsqWalletTransactions method got called at each block for all
transactions. During block download that wasted a lot of cpu and
led to stuck UI thread and lost connections.
The updateBsqBalance is not cheap (a few ms) and called for 100s of txs
at each block was very problematic.
Furthermore the listeners on the walletTransactions observableList got
triggered which made the situation worse.

We changed the observableList to a ArrayList and use a listener which
gets called after the list is updated.
We also make sure the onTransactionConfidenceChanged listener is not
calling updateBsqWalletTransactions if bsq parsing is not complete and
if the depth of the tx is > 1.
In the updateBsqWalletTransactions method we use a flag and a delay
to ensure that the updateBsqBalance is not called more then once
in 100 ms.

We changed also the getter to return a cloned version of the list to
avoid potential concurrent modification exceptions at clients.

Closes #3175

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
@@ -35,6 +35,8 @@ import bisq.core.dao.state.unconfirmed.UnconfirmedBsqChangeOutputListService;
 import bisq.core.provider.fee.FeeService;
 import bisq.core.user.Preferences;
 
+import bisq.common.UserThread;
+
 import org.bitcoinj.core.Address;
 import org.bitcoinj.core.AddressFormatException;
 import org.bitcoinj.core.BlockChain;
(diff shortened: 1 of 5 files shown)


sqrrm:
Changes look reasonable but while running I notice that the `awaiting blocks... Verified nnn blocks out of mmm` is now counting out both the verified number and the total expected.

Screenshot_2019-09-02_00-07-13
Screenshot_2019-09-02_00-07-30
Screenshot_2019-09-02_00-07-42
Screenshot_2019-09-02_00-07-59

chimp1984:
@sqrrm That is expected while block download as the chain tip gets updated with new blocks. The right one is from BitcoinJ, the left one the parsed blocks.

sqrrm:
@chimp1984 The right one should be the known block height from bitcoinj but it keeps changing, that's strange.

chimp1984:
> but it keeps changing, that's strange.

If that is during block download it is expected, otherwise its a bug. I will test again.

chimp1984:
Just tested again but did the block download completed before parsing started. But taht way all was as expected.

ripcurlx (approved):
utACK

sqrrm:
Made a video. The total used to be known at start, now it's changing, and the progress is not showing correctly.


blocksync

chimp1984:
@sqrrm 
Yes that shows that you received the BSQ blocks before the blocks had been synced (you should see the % of download not completed on the left bottom side for BTC). The code should handle that case so that you repeat requesting BSQ blocks. But maybe there are bugs in that part if your BSQ blocks don't get updated again.

PR #3178, opened 2019-09-01
Improve logging

chimp1984:
This should help for debugging to enable to use debug level without getting the output filled up completely. Also adds some duration measurements to help to spot performance bottlenecks.
We should add those duration logs at each method which is suspect to be expensive.


core/src/main/java/bisq/core/dao/monitoring/DaoStateMonitoringService.java
```diff
@@ -316,6 +325,13 @@ private void updateHashChain(Block block) {
             int delayInSec = 5 + new Random().nextInt(10);
             UserThread.runAfter(() -> daoStateNetworkService.broadcastMyStateHash(myDaoStateHash), delayInSec);
         }
+        long duration = System.currentTimeMillis() - ts;
+        // We don't want to spam the outpu. We log accumulated time after parsing is completed.
```
sqrrm:
```suggestion
        // We don't want to spam the output. We log accumulated time after parsing is completed.
```

sqrrm (approved):
utACK

PR #3179, opened 2019-09-02
Update default fiat and crypto currencies

devinbileck:
Within the settings, the default list of displayed national currencies and altcoins is outdated, in terms of top traded, and most are unnecessary to be displayed for most users and a nuisance to have to remove (which I am sure is the first thing most users do).

I have updated it so that the preferred currency and only displayed national currency is chosen based on the users locale/county code, and the only altcoin displayed by default is BSQ. If the user wants anything else it is up to them to add it.

Before:
image

After:
image

### DIFF
--- a/core/src/main/java/bisq/core/locale/CurrencyUtil.java
+++ b/core/src/main/java/bisq/core/locale/CurrencyUtil.java
@@ -86,8 +86,7 @@ public class CurrencyUtil {
         list.add(new FiatCurrency("GBP"));
         list.add(new FiatCurrency("CAD"));
         list.add(new FiatCurrency("AUD"));
-        list.add(new FiatCurrency("RUB"));
-        list.add(new FiatCurrency("INR"));
+        list.add(new FiatCurrency("BRL"));
 
         list.sort(TradeCurrency::compareTo);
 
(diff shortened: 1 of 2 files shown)


ripcurlx:
@devinbileck You've to fix the PreferencesTest. If you have time to fix it right now, I'll wait for the codefreeze until you've updated this PR.

ripcurlx (changes requested):
NACK - Please see my comment

ripcurlx:
@devin I just looked at the code part. I think this change needs to have some discussion as we are using it right now to push a couple of currencies. @ManfredKarrer @sqrrm @m52go might want to join this discussion as well.

ripcurlx:
> to push a couple of currencies

On the fiat side I agree that it might not make sense to have all other currencies there. On the Altcoin side we have to discuss which ones we want to keep to make it easier for new users to find their coin immediately. To select your Fiat and/or Altcoin market you want to trade in should be part of the future onboarding process. In that case we can remove all of them afterwards.

devinbileck:
Ah, good points. I will wait to receive feedback before proceeding.

chimp1984:
I don't think it is good to reduce the fiat currencies to the users default currency, as some users trade on several currencies and might be interested to look up other markets. With one currency it might give the impression Bisq is supporting only that currency for newbies.

chimp1984:
Also on altcoins I think we should keep a list of "recommended" coins. It is not only about trade volume but about coins which we think makes sense (e.g. Bitcoin Rhodium had high volume but is irrelevant long term, stablecoins will become important as alternative to fiat so we should promote those to make them more visible, as well as privacy coins like Grin - once added again).
Also XMR is 99% volume so that is for all reasons a must have.

chimp1984 (commented):
Concept NACK

battleofwizards:
I confirm the multi fiat use case. On top of that I never traded against my locale-implied currency.

PR #3181, opened 2019-09-02
Remove cbeams pricenode from ProvidersRepository

cbeams:
The BitcoinAverage developer plan API keys in use for this node have
recently expired. Given that we already have four other pricenodes
up and running, there's no compelling reason to keep this fifth one in
place and to pay for hosting and API keys for it every month.

bisq-network/roles#14 will be updated accordingly.


ripcurlx (approved):
utACK

PR #3182, opened 2019-09-02
fix NPE in shutdown hook

christophsturm:
I saw some NPEs when running the test suite. this probably only happens when the app is quit very early. Still its a bug


chimp1984 (approved):
utACK

PR #3188, opened 2019-09-02
Fix CSS styling bug caused by using Text() instead of Label()

wiz:
Before:
Screen Shot 2019-09-02 at 22 58 53
After:
Screen Shot 2019-09-03 at 5 59 30

Also fixes it here:
Screen Shot 2019-09-03 at 5 57 21


ripcurlx (approved):
utACK - Actually there is also a way to style Text components. You have to use `-fx-fill` instead of `-fx-text-fill`. Very counterintuitive I know 🤦‍♂ . I used the Text component as it is more lightweight than the label. But as it might not make a big difference and it makes the css more complicated to maintain I'm fine with this change.

PR #3192, opened 2019-09-03
Reduce size of chart axis font to fit Japanese string

wiz:
Before:
Screen Shot 2019-09-03 at 21 22 39

After:
Screen Shot 2019-09-03 at 21 45 18


ripcurlx (approved):
utACK

PR #3194, opened 2019-09-03
Fix offer volume jumps

### LINKED ISSUE #3050: Field "Amount in USD to spend" is not working properly.
Scenario 1:

1. Create a new offer to buy bitcoin for dollars
2. Put $600 in the "Amount in USD to spend" field
3. Set the "Below % from market price" field to 2%

Result: The "Amount in USD to spend" has been changed to $588.
Expected: The "Amount in USD to spend" should remain $600. 

Scenario 2:

1. Create a new offer to buy bitcoin for dollars
2. Put $600 in the "Amount in USD to spend" field
3. Set the "Fixed price in USD for 1 BTC" field to $100

Result: The "Amount in USD to spend" has been changed to $50.
Expected: The "Amount in USD to spend" should remain $600.

Scenario 3:

1. Create a new offer to buy bitcoin for dollars
2. Put 0.5 BTC in the "Amount of BTC to buy" field
3. Set the "Fixed price in USD for 1 BTC" field to $1000.  The "Amount in USD to spend" field will be updated to read $500.
4. Put $600 in the "Amount in USD to spend" field

Result: The "Amount in USD to spend" field will revert to reading $500.
Expected: The "Amount in USD to spend" field will revert to reading $600 - the number you input. 

Note that this behavior doesn't happen for all inputs. For example, if you delete a 0 in $500, it will update to $50.00. If you put in $50 it will revert. If you put in $50.00 it *won't* revert. The logic on this field is VERY broken. Please don't just try fixing this bug with a patchover. The code obviously needs refactoring so the logic is cleaner.

Scenario 4:

1. Create a new offer to buy bitcoin for dollars
2. Put $600 in the "Amount in USD to spend" field
3. Put 0.05 BTC in the "Amount of BTC to buy" field

Result: The "Fixed price in USD for 1 BTC" or "Below % from market price" remain empty.
Expected: The "Fixed price in USD for 1 BTC" or "Below % from market price" should be updated so the equation is fulfilled (just like the  "Amount in USD to spend" field).

These fields just don't work intuitively or even correctly. Updating one part of the equation should update other parts of the equation, but it should anchor on the values the USER input, not on an arbitrary field.

ben-kaufman:
Fix #3050 , scenarios 1 and 4.

I'm not sure the code is good enough, there's probably a cleaner solution I missed, but from my local testing it works well.


chimp1984 (approved):
utACK for commit 
f393625 

For other commit I don't have time to look more closely atm.

Another case where missing brackets caused a bug. I woul drecommend to make it part of our syle guide to always use brackets - even in one liners.

desktop/src/main/java/bisq/desktop/main/offer/MutableOfferViewModel.java
```diff
@@ -1208,4 +1216,35 @@ private void stopTimeoutTimer() {
     private BSFormatter getFormatterForMakerFee() {
         return dataModel.isCurrencyForMakerFeeBtc() ? btcFormatter : bsqFormatter;
     }
+
```
ripcurlx, line 1221: The code works as expected atm, but I think this part of the could should be moved into the MutableOfferViewDataModel.

ripcurlx (changes requested):
NACK - Please see my comment.

desktop/src/main/java/bisq/desktop/main/offer/MutableOfferViewModel.java
ben-kaufman: I'm not sure what part of the code should I move there. It seems like I can move just a few lines of the actual math calculation, while I'll need to keep the rest. Is that what you wanted me to move?

ripcurlx: I've to look at the code in more detail. I'll give it a quick try if I'm able to move it without triggering undesired updates.

desktop/src/main/java/bisq/desktop/main/offer/MutableOfferViewModel.java
```diff
@@ -1207,4 +1216,35 @@ private void stopTimeoutTimer() {
...
+    private void updateMarketPriceAToManual() {
```
ripcurlx: `update market price a to manual` - Should the `a` mean automatic?

desktop/src/main/java/bisq/desktop/main/offer/MutableOfferViewModel.java
ripcurlx: On second look, yes it is not really feasible to move everything into the data model. Please move only the calculation part as you suggested to the data model. There is already a block of calculate... methods there. Besides that and the naming of the update method ACK

ben-kaufman:
@ripcurlx I'm not completely sure this is what you meant but hopefully it should be good enough now :)

ripcurlx:
@ben-kaufman Sorry for the delayed re-review, but I'm focusing on the new release right now. Will come back to you soon.

ripcurlx:
@ben-kaufman Just a quick ping, that I'm going through your PR right now.

ripcurlx (approved):

ripcurlx:
ACK

fresheneesz:
Is this in v1.1.6? Neither scenario 1 nor scenario 4 are fixed in v1.1.6.

ben-kaufman:
@fresheneesz no, this was merged after the release of v1.1.6.

PR #3195, opened 2019-09-03
Remove bcprov from direct dependencies - a major step towards eventually removing Bouncy Castle

battleofwizards:
This removes `org.bouncycastle:bcprov` from **direct** dependencies.

We still depend on this jar **indirectly** via `bouncycastle:bcpg`, which is used for PGP signature verification of Bisq Desktop updates.

This gets us closer to prospect of removing Bouncy Castle dependency entirely in the future.

Rationale for this direction:

* We are really only using Bouncy Castle for PGP sig verification
* Java's builtin cryptography got vastly better since 2000; no more key length restrictions
* We should prefer boring and proven cryptography anyway
* Australian projects should be considered compromised and Bouncy Castle is managed by Australian non-profit organization
* Bouncy Castle is a heavy dependency totaling 3.6MB

In the process:

* Ergo coin got removed
* BC Base64 got replaced with Java's builtin version (which is much faster BTW)
* BC Hex got replaced with Guava's version

Note to reviewers: individual commits are easier to review than full diff. They also provide more details.


common/src/main/java/bisq/common/util/Hex.java
```diff
@@ -0,0 +1,31 @@
...
+public class Hex {
+
+    public static byte[] decode(String hex) {
+        return BaseEncoding.base16().lowerCase().decode(hex.toLowerCase());
+    }
+
+    public static String encode(byte[] bytes) {
+        return BaseEncoding.base16().lowerCase().encode(bytes);
+    }
+}
```
chimp1984, line 31: We have in the bisq.common.util.Utilities calls methods for encode and decode.

chimp1984 (changes requested):
Can you please use the existing methods in Utilities?

common/src/main/java/bisq/common/util/Hex.java
chimp1984: Actually it would be more clean to use the class as you added here. But maybe in a second PR would be good to refactor the usage of the Utilities methods to have only one method used for hex. There might be also use of the HEX class from BitcoinJ (same impl).

chimp1984 (approved):
utACK

common/src/main/java/bisq/common/util/Hex.java
battleofwizards: Yes, I am aware of Utilities and BitcoinJ versions of Hex being used in the code. It did cross my mind to standardize this but I decided it is too broad for this PR.

In this PR I focused on minimizing the diff and ensuring backwards compatibility. We need exactly the same behavior from replacements as original Bounce Castle versions. Hence the "new" Hex.

jasondavies:
> Ergo coin got removed

We've only just added support for Ergo!  I didn't realise you were about to remove Bouncy Castle.  It's only required for validating the base58 checksum, which in Ergo uses blake2b-256.  Any suggestions for the preferred approach here?

* Use an external blake2 library e.g. <https://github.com/alphazero/Blake2b>
* Use our own implementation of blake2b-256 in-tree.
* Skip checksum validation for Ergo addresses.

battleofwizards:
Hi @jasondavies! The Ergo was removed only b/c it had a bad luck of depending on the Bouncy Castle. If you want to re-add it please make it w/o relying on external dependencies and w/o introducing new cryptographic code to Bisq. One option is to only partially validate addresses with a regexp. Once Ergo explodes in popularity (if ever) we may then consider more proper validation. 

Thanks for your understanding and sorry for the turmoil.

jasondavies:
Thanks for the quick reply!  I've simply removed the checksum validation (which uses blake2b-256) and the existing base58 decode should be better than a regexp.  See latest PR.

PR #3200, opened 2019-09-04
Add scripts for testing jars

ripcurlx:
To make it easy to create fat testing jars with signature for desktop and seednode.


sqrrm (approved):
utACK

PR #3202, opened 2019-09-04
periodical UpdateDataReq for Seednodes

freimair:
This PR introduces a recurring syncronization with other seed nodes if Bisq is run in seed node-mode. In detail, every seed node issues an UpdateDataRequest to the connected seed nodes once every hour, receives data such as it would during restart. By having the same effect as a restart but without having to wait 24h and being offline for some time, this PR deprecates https://github.com/bisq-network/bisq/tree/seednode_temporaryfix.

I did a lot of debugging and analyzing in order to get this thing up and running. First and foremost, I found that by counting the unique messages of all seed nodes together results in a higher number of messages than any seed node has on its own. In other words, no seed node has every message. Here is a graph with distinct snapshots: 
Screenshot from 2019-09-04 18-51-24

Secondly, I monitored the proposed fix over periods of time to see how the number evolve. I only monitored Offer- and MailboxStoragePayloads, as the other messages are pretty much in sync (according to https://monitor.bisq.network/d/wTDl6T_iz/p2p-stats). The graphs below denote the number of messages on their y-axis, the time of measurement on the x-axis, the "got-" values reflect the number of messages received from a seed node during a periodical UpdateDataRequest, the "have-" values denote the number of messages we already have.
Screenshot from 2019-08-30 10-05-09
Screenshot from 2019-08-30 10-05-22
Screenshot from 2019-08-30 10-05-29
Screenshot from 2019-09-05 08-03-23
Screenshot from 2019-09-06 16-36-54
Screenshot from 2019-09-09 18-41-25

All in all, it seems that the recurring querying of other seed nodes might fill in the missing messages just fine. I even believe that we might get rid of at least some arbitration cases, where MailboxMessages simply did not arrive at the receiver. However, I suggest a slow adoption as we had our difficulties with these attempts in the past.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/P2PService.java
+++ b/p2p/src/main/java/bisq/network/p2p/P2PService.java
@@ -311,6 +311,7 @@ public class P2PService implements SetupListener, MessageListener, ConnectionLis
                 "seedNodeOfPreliminaryDataRequest must be present");
 
         requestDataManager.requestUpdateData();
+        UserThread.runPeriodically(() -> requestDataManager.requestUpdateData(), 1, TimeUnit.HOURS);
 
         // If we start up first time we don't have any peers so we need to request from seed node.
         // As well it can be that the persisted peer list is outdated with dead peers.
--- a/p2p/src/main/java/bisq/network/p2p/P2PService.java
+++ b/p2p/src/main/java/bisq/network/p2p/P2PService.java
@@ -47,6 +47,8 @@ import bisq.network.p2p.storage.payload.ProtectedStorageEntry;
 import bisq.network.p2p.storage.payload.ProtectedStoragePayload;
 
 import bisq.common.UserThread;
+import bisq.common.app.Capabilities;
+import bisq.common.app.Capability;
 import bisq.common.crypto.CryptoException;
 import bisq.common.crypto.KeyRing;
 import bisq.common.crypto.PubKeyRing;
@@ -311,7 +313,8 @@ public class P2PService implements SetupListener, MessageListener, ConnectionLis
                 "seedNodeOfPreliminaryDataRequest must be present");
 
         requestDataManager.requestUpdateData();
-        UserThread.runPeriodically(() -> requestDataManager.requestUpdateData(), 1, TimeUnit.HOURS);
+        if (Capabilities.app.containsAll(Capability.SEED_NODE))
+            UserThread.runPeriodically(() -> requestDataManager.requestUpdateData(), 1, TimeUnit.HOURS);
 
         // If we start up first time we don't have any peers so we need to request from seed node.
         // As well it can be that the persisted peer list is outdated with dead peers.
--- a/p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
+++ b/p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
@@ -35,6 +35,8 @@ import bisq.network.p2p.storage.payload.ProtectedStoragePayload;
 
 import bisq.common.Timer;
 import bisq.common.UserThread;
+import bisq.common.app.Capabilities;
+import bisq.common.app.Capability;
 import bisq.common.proto.network.NetworkEnvelope;
 import bisq.common.proto.network.NetworkPayload;
 
@@ -60,6 +62,12 @@ class RequestDataHandler implements MessageListener {
     private static final long TIMEOUT = 90;
     private NodeAddress peersNodeAddress;
 
+    /**
+     * when we are run as a seed node, we spawn a RequestDataHandler every hour. However, we do not want to receive
+     * {@link PersistableNetworkPayload}s (for now, as there are hardly any cases where such data goes out of sync). This
+     * flag indicates whether we already received our first set of {@link PersistableNetworkPayload}s.
+     */
+    private static boolean firstRequest = true;
 
     ///////////////////////////////////////////////////////////////////////////////////////////
     // Listener
@@ -219,7 +227,11 @@ class RequestDataHandler implements MessageListener {
                         });
                         log.info("Processing {} protectedStorageEntries took {} ms.", counter.get(), System.currentTimeMillis() - ts);
 
-                        if (persistableNetworkPayloadSet != null) {
+                        // engage the firstRequest logic only if we are a seed node. Normal clients get here twice at most.
+                        if (!Capabilities.app.containsAll(Capability.SEED_NODE))
+                            firstRequest = true;
+
+                        if (persistableNetworkPayloadSet != null && firstRequest) {
                             ts = System.currentTimeMillis();
                             persistableNetworkPayloadSet.forEach(e -> {
                                 if (e instanceof LazyProcessedPayload) {
@@ -241,6 +253,7 @@ class RequestDataHandler implements MessageListener {
 
                         cleanup();
                         listener.onComplete();
+                        firstRequest = false;
                     } else {
                         log.warn("Nonce not matching. That can happen rarely if we get a response after a canceled " +
                                         "handshake (timeout causes connection close but peer might have sent a msg before " +
--- a/p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
+++ b/p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
@@ -194,49 +194,10 @@ class RequestDataHandler implements MessageListener {
             if (connection.getPeersNodeAddressOptional().isPresent() && connection.getPeersNodeAddressOptional().get().equals(peersNodeAddress)) {
                 if (!stopped) {
                     GetDataResponse getDataResponse = (GetDataResponse) networkEnvelope;
-                    Map<String, Set<NetworkPayload>> payloadByClassName = new HashMap<>();
                     final Set<ProtectedStorageEntry> dataSet = getDataResponse.getDataSet();
-                    dataSet.forEach(e -> {
-                        final ProtectedStoragePayload protectedStoragePayload = e.getProtectedStoragePayload();
-                        if (protectedStoragePayload == null) {
-                            log.warn("StoragePayload was null: {}", networkEnvelope.toString());
-                            return;
-                        }
-
-                        // For logging different data types
-                        String className = protectedStoragePayload.getClass().getSimpleName();
-                        if (!payloadByClassName.containsKey(className))
-                            payloadByClassName.put(className, new HashSet<>());
-
-                        payloadByClassName.get(className).add(protectedStoragePayload);
-                    });
-
-
                     Set<PersistableNetworkPayload> persistableNetworkPayloadSet = getDataResponse.getPersistableNetworkPayloadSet();
-                    if (persistableNetworkPayloadSet != null) {
-                        persistableNetworkPayloadSet.forEach(persistableNetworkPayload -> {
-                            // For logging different data types
-                            String className = persistableNetworkPayload.getClass().getSimpleName();
-                            if (!payloadByClassName.containsKey(className))
-                                payloadByClassName.put(className, new HashSet<>());
 
-                            payloadByClassName.get(className).add(persistableNetworkPayload);
-                        });
-                    }
-
-                    // Log different data types
-                    StringBuilder sb = new StringBuilder();
-                    sb.append("\n#################################################################\n");
-                    sb.append("Connected to node: " + peersNodeAddress.getFullAddress() + "\n");
-                    final int items = dataSet.size() +
-                            (persistableNetworkPayloadSet != null ? persistableNetworkPayloadSet.size() : 0);
-                    sb.append("Received ").append(items).append(" instances\n");
-                    payloadByClassName.forEach((key, value) -> sb.append(key)
-                            .append(": ")
-                            .append(value.size())
-                            .append("\n"));
-                    sb.append("#################################################################");
-                    log.info(sb.toString());
+                    if (log.isDebugEnabled()) logContents(networkEnvelope, dataSet, persistableNetworkPayloadSet);
 
                     if (getDataResponse.getRequestNonce() == nonce) {
                         stopTimeoutTimer();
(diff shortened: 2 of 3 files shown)


p2p/src/main/java/bisq/network/p2p/P2PService.java
```diff
@@ -337,11 +340,13 @@ public void onPreliminaryDataReceived() {
...
-        p2pServiceListeners.stream().forEach(P2PServiceListener::onUpdatedDataReceived);
-        if(!isBootstrapped)
-            p2PDataStorage.onBootstrapComplete();
+        if (!isBootstrapped() || Capabilities.app.containsAll(Capability.SEED_NODE)) {
+            isBootstrapped = true;
+            maybeProcessAllMailboxEntries();
+            p2pServiceListeners.stream().forEach(P2PServiceListener::onUpdatedDataReceived);
+            if (!isBootstrapped)
+                p2PDataStorage.onBootstrapComplete();
+        }
```
chimp1984: In your new code it is not executed is isBootStapped=true and not a seed node. Without having looked to a wider context that seems wrong or at lease a behaviour change.

chimp1984: Ah ignore previous comment... See in the first commit the original code...

chimp1984: Maybe worth to squash all commits, makes review easier

chimp1984 (commented):
Would need a bit more time for testing PR... On first look it looks ok.

p2p/src/main/java/bisq/network/p2p/P2PService.java
freimair: Ahm I am not a fan of squashing commits, because:
- if a reviewer wants to see individual commits (because there is far less a change than all of them), he can just click them
- if a reviewer wants to see all changes, he can review in the [Files]-Tab...

as for your comment, these `if`s are strange. Onto it.

freimair: fixed by doing
```
        if (!isBootstrapped() || Capabilities.app.containsAll(Capability.SEED_NODE)) {
            maybeProcessAllMailboxEntries();
            p2pServiceListeners.stream().forEach(P2PServiceListener::onUpdatedDataReceived);
        }

        if (!isBootstrapped()) {
            isBootstrapped = true;
            p2PDataStorage.onBootstrapComplete();
        }
```

p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
```diff
@@ -304,6 +278,56 @@ public void stop() {
...
+        sb.append("Connected to node: " + peersNodeAddress.getFullAddress() + "\n");
+        final int items = dataSet.size() +
+                (persistableNetworkPayloadSet != null ? persistableNetworkPayloadSet.size() : 0);
+        sb.append("Received ").append(items).append(" instances\n");
+        payloadByClassName.entrySet().stream().forEach(e -> sb.append(e.getKey())
+                .append(": ")
+                .append(e.getValue().size())
+                .append("\n"));
+        sb.append("#################################################################");
+        log.info(sb.toString());
```
christophsturm: info is probably enabled most of the time. still this whole method should be in a `if (log.isInfoEnabled)` block

p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
```diff
@@ -190,53 +196,21 @@ public void onFailure(@NotNull Throwable throwable) {
...
-                    final int items = dataSet.size() +
-                            (persistableNetworkPayloadSet != null ? persistableNetworkPayloadSet.size() : 0);
-                    sb.append("Received ").append(items).append(" instances\n");
-                    payloadByClassName.forEach((key, value) -> sb.append(key)
-                            .append(": ")
-                            .append(value.size())
-                            .append("\n"));
-                    sb.append("#################################################################");
-                    log.info(sb.toString());
+                    if (log.isDebugEnabled()) logContents(networkEnvelope, dataSet, persistableNetworkPayloadSet);
```
christophsturm, line 208: this checks for debug, and then the method logs as info

christophsturm:
I really think the log level check and the log statement should bei in the same method. maybe the logContent method should just build the stringbuffer and the logging should be outside.

freimair:
disagree. the method is about `logContents` and if you do not want to log it, you should not call it.

christophsturm:
IMO It’s the methods concern to build the string only when needed.

p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
```diff
@@ -208,7 +208,7 @@ else if (networkEnvelope instanceof RemoveDataMessage)
                     final Set<ProtectedStorageEntry> dataSet = getDataResponse.getDataSet();
                     Set<PersistableNetworkPayload> persistableNetworkPayloadSet = getDataResponse.getPersistableNetworkPayloadSet();
 
-                    if (log.isDebugEnabled()) logContents(networkEnvelope, dataSet, persistableNetworkPayloadSet);
+                    if (log.isInfoEnabled()) logContents(networkEnvelope, dataSet, persistableNetworkPayloadSet);
```
chimp1984: logContents prints very important information which we also want to log in debug mode.

p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
```diff
@@ -190,6 +193,11 @@ public void onFailure(@NotNull Throwable throwable) {
 
     @Override
     public void onMessage(NetworkEnvelope networkEnvelope, Connection connection) {
+        if (networkEnvelope instanceof RemoveMailboxDataMessage)
+            recentlyRemovedMessages.add(new P2PDataStorage.ByteArray(P2PDataStorage.get32ByteHash(((RemoveMailboxDataMessage) networkEnvelope).getProtectedMailboxStorageEntry().getProtectedStoragePayload())));
+        else if (networkEnvelope instanceof RemoveDataMessage)
+            recentlyRemovedMessages.add(new P2PDataStorage.ByteArray(P2PDataStorage.get32ByteHash(((RemoveDataMessage) networkEnvelope).getProtectedStorageEntry().getProtectedStoragePayload())));
+
```
chimp1984: What is the intended usage of recentlyRemovedMessages? It is only filled but not queried beside the cleanup loop. Handling for remove msg was added recently to P2PDataStorage. It seems it still covers only MailboxStoragePayload but that would probably require further implementaions of the AddOncePayload interface... Would need moe time to have a clear opinion about that....

chimp1984: Also I would prefer a more verbose style as otherwise its very hard to follow:

``` 
if (networkEnvelope instanceof RemoveMailboxDataMessage) {
            RemoveMailboxDataMessage removeMailboxDataMessage = (RemoveMailboxDataMessage) networkEnvelope;
            ProtectedStoragePayload protectedStoragePayload = removeMailboxDataMessage.getProtectedMailboxStorageEntry().getProtectedStoragePayload();
            P2PDataStorage.ByteArray byteArray = new P2PDataStorage.ByteArray(P2PDataStorage.get32ByteHash(protectedStoragePayload));
            recentlyRemovedMessages.add(byteArray);
        } else if (networkEnvelope instanceof RemoveDataMessage) {
            RemoveDataMessage removeDataMessage = (RemoveDataMessage) networkEnvelope;
            ProtectedStoragePayload protectedStoragePayload = removeDataMessage.getProtectedStorageEntry().getProtectedStoragePayload();
            P2PDataStorage.ByteArray byteArray = new P2PDataStorage.ByteArray(P2PDataStorage.get32ByteHash(protectedStoragePayload));
            recentlyRemovedMessages.add(byteArray);
        }
```

p2p/src/main/java/bisq/network/p2p/P2PService.java
```diff
@@ -336,12 +337,11 @@ public void onPreliminaryDataReceived() {
...
     public void onUpdatedDataReceived() {
-        if (!isBootstrapped) {
-            isBootstrapped = true;
-            maybeProcessAllMailboxEntries();
-            p2pServiceListeners.stream().forEach(P2PServiceListener::onUpdatedDataReceived);
+        isBootstrapped = true;
+        maybeProcessAllMailboxEntries();
+        p2pServiceListeners.stream().forEach(P2PServiceListener::onUpdatedDataReceived);
+        if(!isBootstrapped)
             p2PDataStorage.onBootstrapComplete();
```
chimp1984: The new version would behave different, even it might be not drastic and visible changes, I think that this is dangerous.

To make it more clear I post the old and the new method:
New version:
```
    public void onUpdatedDataReceived() {
        if (!isBootstrapped() || Capabilities.app.containsAll(Capability.SEED_NODE)) {
            maybeProcessAllMailboxEntries();
            p2pServiceListeners.stream().forEach(P2PServiceListener::onUpdatedDataReceived);
        }

        if (!isBootstrapped()) {
            isBootstrapped = true;
            p2PDataStorage.onBootstrapComplete();
        }
    }
```
Old version:
```
    public void onUpdatedDataReceived1() {
        if (!isBootstrapped) {
            isBootstrapped = true;
            maybeProcessAllMailboxEntries();
            p2pServiceListeners.stream().forEach(P2PServiceListener::onUpdatedDataReceived);
            p2PDataStorage.onBootstrapComplete();
        }
    }
```

  isBootstrapped = true; is set after maybeProcessAllMailboxEntries(); is called, but there we check for isBootstrapped so that will behave differently. the flag need to be set initially before any other calls.


`p2pServiceListeners.stream().forEach(P2PServiceListener::onUpdatedDataReceived);`

I am not sure if a repeated onUpdatedDataReceived notification might cause issues. It is expected to be only called once. I know its for seed node only and likely will not cause issued in that use case, but I think it would be better to add a new handler (can be `default` in the interface so it will not require clients to implement new handler method).

And a minor issue:
`isBootstrapped()` should be replaced by `isBootstrapped`.

chimp1984 (changes requested):
NACK, see comments

p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
freimair: Firstly, I do not agree that the information is very important. It is just numbers. To be exact, it is number of how many messages of each type arrive after a Preliminary- or UpdateDataRequest. Users only get poorer performance and a bigger log file out of that.

Secondly, if you enable "DEBUG log verbosity" (I assume that is what you meant), INFO is enabled as well. The logging behavior does not change, only the block of undocumented logging code is moved out of the way so one can see the functional code faster.

p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
freimair: oh my. merge gone wrong. onto it.

p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
chimp1984: When I look into logs thets one of the most important info. There I see if seeds deliver same numbers, if I got the arbitrators and filters etc. 

> Secondly, if you enable "DEBUG log verbosity" (I assume that is what you meant), INFO is enabled as well. 

Ah ok. did not know that. But yes makes sense  ;-)

> The logging behavior does not change, only the block of undocumented logging code is moved out of the way so one can see the functional code faster.

Yes moving it out in a method is fine.

p2p/src/main/java/bisq/network/p2p/P2PService.java
freimair: I am running a long-time test right now with the "old version". It seems that everything works just fine without the change. God I love when minor changes somewhere in core change the behavior elsewhere without any trace :).

freimair: here is another long-term test run result which unfortunately was interrupted by a power outage. This time, with the old version. Looks fine as well.
Screenshot from 2019-09-12 14-21-33

Just started another long-term test with data on how much messages have been rejected because of the recentlyRemoved filter.

p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
freimair: fixed

freimair:
here is the last long-term test and it looks good to me. ready to be merged again.
Screenshot from 2019-09-16 13-27-29

EDIT: ah yes, and 1249 messages have not been re-added due to the "removed"-cache over a period of 4 days (that is 10+ messages per hour).

chimp1984:
> God I love when minor changes somewhere in core change the behavior elsewhere without any trace :).

Not sure what you refer to but I think it might be the AddOncePayload interface we added MailboxStoragePayload and which avoids now in P2PDataStorage that MailboxMessages can be added again after they have been removed.

freimair:
I am not sure either. Back then, I tried doing the resync and the "new version" has been a critical part of the process. If not for the "new version", updateDataResponses did not get processed at all.

However, now, a few weeks/months later, it works with the "old version". Lucky, I guess!

ripcurlx:
> I am not sure either. Back then, I tried doing the resync and the "new version" has been a critical part of the process. If not for the "new version", updateDataResponses did not get processed at all.
> 
> However, now, a few weeks/months later, it works with the "old version". Lucky, I guess!

@freimair @chimp1984 Does this mean this PR is now up to be merged or still [WIP]?

chimp1984:
@ripcurlx Not ready yet, need more time for review

chimp1984:
NACK

If an offer gets added, then removed then again added it would be listed in the `recentlyRemovedMessages` map but it should not. This would lead to the situation that this offer is not provided.

Also it does not handle PersistableNetworkPayload items like Alert, Filter, TempProposal...

Risks for big inconsistencies might be low in case of offers due its short TTL and republishing but with other objects like Filter TTL is very long and there is no frequent republishing. Some of those data are very critical (like Filter).

freimair:
another iteration. looks good still.
Screenshot from 2019-09-24 10-14-44

there have been only 9 messages not being added because they already have been removed over the course of 4 days. However, 38 mailbox messages have been synced to the seednode under test only by the recurring updates to other seed nodes.

chimp1984 (approved):
utACK

wiz:
jhgcy2won7xnslrb.onion is now running this for testing 4b66b0450 Only store PeristablePayloads once

wiz:
In case this was not clear from my Slack comments in the seednode channel, the testing failed, the node crashed and got stuck due to an apparent memory leak, and I had to revert to the previous branch.

wiz:
@freimair please check https://github.com/bisq-network/bisq/issues/3468#issue-512540243

chimp1984:
@wiz @freimair 
That issue is resolved
https://github.com/bisq-network/bisq/pull/3476

PR #3203, opened 2019-09-05
Add autogenerated entry by Intellij (2019.2)

ripcurlx:


bodymindarts:
Do we care about line endings? Cause this file is missing a CR at the bottom. May as well add it.

ripcurlx:
> Do we care about line endings? Cause this file is missing a CR at the bottom. May as well add it.

Atm we don't have it in other config files as well. No strong opinion about that. But to be consistent with the other files right now I'd push it as it is.

battleofwizards:
@bodymindarts this file is auto generated by IntelliJ IDEA and for that reason it is not subject to the best practice you mentioned.

chimp1984:
Ups... from slack: " to fix that annoying PREFERRED_PROJECT_CODE_STYLE open code style and change to project. it its already set, change ot other, apply and the change back to project. same with code stype/java. that fixed it for me. seems IDEA has from old versions the setting and does not rewrite it if not newly applied"

PR #3206, opened 2019-09-05
Update to Tor v0.4.1.5

freimair:
first time I tried that tor version, it took an awful long time to extract the new binaries. but only once. maybe someone wants to check that.


christophsturm:
ack. also build was quick even the first time

freimair:
I would like to remind the reviewers and maintainers, that Tor is the infrastructure most critical to Bisq. Every Tor update is there for a reason. This should be a priority. Always.

PR #3207, opened 2019-09-05
Remove PGP keys from the KeyRing

battleofwizards:
The PGP support was envisioned in the KeyRing but never fully implemented. This removes the (essentially) dead code.

This code impeded my understanding on how we use signatures. My initial impression was that we use PGP for p2p message signing, which is not the case.

The protobuf message compatibility is implemented with `reserved`.


ripcurlx:
Looks good to me, but I'd like to have an ACK by @freimair, @sqrrm or @ManfredKarrer before merging.

bodymindarts:
utAck removing dead code is great. I was just doing a deep dive into this code yesterday and also thought this should simply be removed!

freimair (approved):
ACK

sqrrm (approved):
utACK

PR #3215, opened 2019-09-06
Remove old capability requirements

chimp1984:
We leave the DAO related capabilities as it might be useful in future for flexibility if there is a use case for a non-DAO node.


sqrrm (approved):
utACK

PR #3216, opened 2019-09-06
More logging optimisations. 

christophsturm:
Those don't need to get into 1.1.6 because they are not hotspots. based on #3214


freimair (approved):
Ack

PR #3217, opened 2019-09-06
Add "squash commits" to contributor's workflow

erciccione:
At the moment the git history of all repos is not properly handled. This is a first step to introduce a better workflow where contributors squash their own commits; but maintainers could squash as well if the contributor doesn't/cannot.

See bisq-network/style#14


freimair (approved):
definitely

ripcurlx (approved):
ACK

ripcurlx:
As I think this is in interest of all, I'll merge this PR.

PR #3218, opened 2019-09-06
Only build string when its actually output

christophsturm:
this is also a performance hotspot and should go into 1.1.6 cc @ripcurlx


ripcurlx (approved):
utACK

ripcurlx:
ah damn - I missed that travis hadn't finished yet. @christophsturm can you please fix the build error!

ripcurlx:
My bad. As there are commits on the release branch, that didn't make it into master yet, I forgot to test the build after cherry-picking the latest performance optimization PR.

PR #3226, opened 2019-09-08
Add bsq price trend indicator

chimp1984:


chimp1984:
The median price is not based on amount weight, so the average is considered more exact IMO.
This PR replaces the median 90 day price wioth the average 30 days and adds an indicator if price is going up or down.

Screen Shot 2019-09-08 at 12 02 18
Screen Shot 2019-09-08 at 12 03 48

sqrrm (approved):
utACK

The median is now gone, perhaps it would've been nice to keep it, but I have no real strong opinion there.

chimp1984:
> perhaps it would've been nice to keep it, but I have no real strong opinion there.

I was alos wondering to make 2 lines, but I think its not worth it. Too much info weakens the important info.

wiz:
It's backwards, should be BTC/BSQ

Screen Shot 2019-09-09 at 10 38 18

ripcurlx (approved):
ACK

chimp1984:
> It's backwards, should be BTC/BSQ

No altcoins are displayed with Altcoin/BTC - not logical but the standard...

PR #3228, opened 2019-09-09
Add new BondedRoleType for Analytics Operator

wiz:


ripcurlx (approved):
utACK

chimp1984:
NACK!

@ripcurlx Please revert that commit.

It was not sufficiently tested if adding a field can break anything. It is low risk but still a risk and not worth it. It would at least require quite a bit of effort for DAO devs to test very properly.

Please consider DAO code as not changeable by dafault and never merge changed in the dao package without explicit ACK from dao devs.

PR #3232, opened 2019-09-09
Remove analytics operator bond requirement

wiz:
As discussed in:
https://github.com/bisq-network/proposals/issues/115#issuecomment-529490565

### DIFF
--- a/core/src/main/java/bisq/core/dao/state/model/governance/BondedRoleType.java
+++ b/core/src/main/java/bisq/core/dao/state/model/governance/BondedRoleType.java
@@ -53,7 +53,7 @@ public enum BondedRoleType {
 
     // operators
     WEBSITE_OPERATOR(50, 110, "https://bisq.network/roles/12", true),
-    ANALYTICS_OPERATOR(1, 110, "https://bisq.network/roles/88", true),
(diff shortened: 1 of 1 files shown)


chimp1984 (commented):
NACK

Changing those values was never tested well in the DAO and its pretty complex to test.

wiz:
@chimp1984 since the analytics operator role hasn't been released in any version of Bisq yet, it should be safe 😅

ripcurlx:
> NACK
> 
> Changing those values was never tested well in the DAO and its pretty complex to test.

utACK - As long as it wasn't used nor released I think it should be quite safe to change.

ripcurlx (approved):
utACK

chimp1984:
Please revert! It is risky to change that even if it was not used. Breaking DAO consensus can have severe effects and that change does not justify the risk.

sqrrm:
I have to agree with @chimp1984 on this one. It seems like a very small benefit compared to the risk of breaking the DAO consensus.

wiz:
@m52go yeah I spoke with the Chimp and he said that while the changes we made are probably fine, even if there's a 0.1% risk of breaking the DAO, we shouldn't do it without huge testing, which is a valid point, so we're going to revert the changes and remove the bond requirement for now.

PR #3236, opened 2019-09-10
Integrate PhoneNumberValidator into HalCashForm

ghubstan:
Added country selector to HalCashForm, requiring the following changes.

1.  A new CountryUtil method getAllHalCashCountries, returns [ES, PL].
2.  HalCashValidator now extends PhoneNumberValidator.
3.  HalCashAccountPayload now extends CountryBasedPaymentAccountPayload.
4.  HalCashAccount now extends CountryBasedPaymentAccount.
5.  Added two new fields to pb.proto.
    a.  HalCashAccountPayload hal_cash_account_payload field to
        CountryBasedPaymentAccountPayload message.
    b.  accepted_country_codes field to HalCashAccountPayload message.
    Note:  pb.proto changes may break backward compatibility.
6.  Reformatted all modified source files except pb.proto.

This patch is part of solution to Issue #3042.

### DIFF
--- a/common/src/main/proto/pb.proto
+++ b/common/src/main/proto/pb.proto
@@ -836,6 +836,7 @@ message CountryBasedPaymentAccountPayload {
         WesternUnionAccountPayload western_union_account_payload = 5;
         SepaInstantAccountPayload sepa_instant_account_payload = 6;
         F2FAccountPayload f2f_account_payload = 7;
+        HalCashAccountPayload hal_cash_account_payload = 8;
     }
 }
 
@@ -888,6 +889,7 @@ message MoneyGramAccountPayload {
 
 message HalCashAccountPayload {
     string mobile_nr = 1;
+    repeated string accepted_country_codes = 2;
 }
 
 message WesternUnionAccountPayload {
--- a/core/src/main/java/bisq/core/locale/CountryUtil.java
+++ b/core/src/main/java/bisq/core/locale/CountryUtil.java
@@ -34,6 +34,14 @@ import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class CountryUtil {
+    public static List<Country> getAllHalCashCountries() {
+        List<Country> list = new ArrayList<>();
+        String[] codes = {"ES", "PL"};
+        populateCountryListByCodes(list, codes);
+        list.sort((a, b) -> a.name.compareTo(b.name));
+        return list;
+    }
+
     public static List<Country> getAllSepaEuroCountries() {
         List<Country> list = new ArrayList<>();
         String[] codes = {"AT", "BE", "CY", "DE", "EE", "FI", "FR", "GR", "IE",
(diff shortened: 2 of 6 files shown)


ripcurlx (changes requested):
NACK - I think it is not a good idea to add the account creation validation and push it as far to invalidate all existing HalCash accounts for offers created with phone number validated HalCash accounts. Doing it like that it would also force people with correct phone numbers to create a new account and loose their account age. I think it is better for this PR only to add the phone number validation in the account creation form and leave everything else as is.

ripcurlx:
As it is implemented right now, it is as if we are introducing two new payment methods. HalCash (ES) and HalCash (PL). Also I'm not sure if it is maybe also possible to transfer money with HalCash between ES and PL.

ghubstan:
It is clear you can send international remittances to a halcash networked ATM in other countries.
https://www.halcash.com/es/usos   (See "REMESAS")

It is not clear halcash users can transfer fiat from one bank account to another, so I think it should be assumed to not be possible, though I don't see how transfers to ATMs in other countries can happen if the funds aren't deposited in a recipient's bank account.

ghubstan:
@ripcurlx, Thanks for the feedback.   I'll revert protobuf changes to limit validation to the create payment method use case.

Phone validation depends on country.  How would you suggest I add a country selector without implying the payment method can be used to transfer fiat between Spain and Poland?

ripcurlx:
If it is not possible to send it across borders than the old implementation was false in that regard and what you came up with is the correct implementation. I’ll think about the best case migration scenario.

ghubstan:
According to [this EN site](https://affgambler.com/banking/halcash-casinos), recipients  _need a cell phone_ (not email), and access to any HalCash ATM in Spain or Poland.   Money can be sent across borders, is not deposited in the recipient's bank account, and  HalCash ATM withdrawal limits apply.

> To send money users should have a bank account with one of participating institutions. In this case they will need to provide phone number of the receiver, PIN code and specify how much money they want to send. Once all required information is sent, receiver gets an SMS telling a reference code. It takes few moments for money transfer to be processed, and withdrawal is possible just seconds later.

> After getting a message about money transfer, users should find closest HalCash ATM. There will be “Withdrawal without using a card” option on a display. Select it and then push HalCash button to find simple instructions you should follow. If you are trying to withdraw money, you’ll need to enter your phone number, transfer size, reference code and PIN code. The latter should be provided by a sender. If the information is accurate, you’ll receive cash.

According to this [HalCash FAQ](https://www.halcash.com/en/quienes-somos), there are plans to add US ATM and POS systems to the network.

But FYI, [this page](https://www.paymentssource.com/news/spains-halcash-tests-the-marketing-potential-of-atms-in-the-us) was published in 2016, which makes me doubt the US roll out is going to happen.

ripcurlx:
> It is clear you can send international remittances to a halcash networked ATM in other countries.
> https://www.halcash.com/es/usos (See "REMESAS")

But isn't that our use case? So someone wants to buy BTC with Halcash and sends the code to the phonenumber of the recipient. The recipient goes to a Halcash ATM receiveds the money and releases the BTC. To make that possible we shouldn't split up Halcash on a country basis, but just validate a general phone number on creation or leave it as it is right now.

ghubstan:
Leaving it as is and closing PR.

PR #3237, opened 2019-09-10
Hide account age verification salt for altcoin accounts

### LINKED ISSUE #2537: Remove salt for account age verification field when adding altcoin account
When adding an altcoin account, the salt for account age verification field can be removed since it is not relevant for altcoins (see related #1743).
image

devinbileck:
The salt for account age verification is not relevant for altcoins,
so hide the field.

Fixes https://github.com/bisq-network/bisq/issues/2537


ripcurlx (approved):
utACK

sqrrm (approved):
ACK

PR #3239, opened 2019-09-11
Make stateless BSFormatter functions static and remove dead code

bodymindarts:
Making these functions static is a preparation for extraction.

Reducing the public interface of the BSFormatter instances reduces coupling and enables moving these functions to places where it makes more sense for them to live.

### DIFF
--- a/core/src/main/java/bisq/core/app/WalletAppSetup.java
+++ b/core/src/main/java/bisq/core/app/WalletAppSetup.java
@@ -120,7 +120,7 @@ public class WalletAppSetup {
                             result = Res.get("mainView.footer.btcInfo",
                                     peers,
                                     Res.get("mainView.footer.btcInfo.synchronizingWith"),
-                                    getBtcNetworkAsString() + ": " + formatter.formatToPercentWithSymbol(percentage));
+                                    getBtcNetworkAsString() + ": " + BSFormatter.formatToPercentWithSymbol(percentage));
                         } else {
                             result = Res.get("mainView.footer.btcInfo",
                                     peers,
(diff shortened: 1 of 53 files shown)


christophsturm:
I think this pr is too small :)

making the method static is only useful if you really move the method. so to see if it makes sense the move needs to be in the same pr. 
While its important to keep prs small, they need to be self contained and make sense on their own.

bodymindarts:
> making the method static is only useful if you really move the method. so to see if it makes sense the move needs to be in the same pr.

I disagree. Reducing the public api of the BSFormatter _instance_ is useful in itself as it makes it possible to see in the calling code if there is a dependency on hidden state or the function is stateless.

Also removing dead code (there are a couple functions deleted that were never called) is also useful in itself, without a further commit required.

Either way I will open another PR with the first move included.

.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -1,5 +1,6 @@
 <component name="ProjectCodeStyleConfiguration">
   <state>
     <option name="USE_PER_PROJECT_SETTINGS" value="true" />
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default" />
```
freimair:
```suggestion

```

that gets me as well every time. Had it solved temporarily by doing https://github.com/bisq-network/bisq/pull/3205, yet, it just reappeared again.

freimair (changes requested):
thanks for splitting that up. yet, I found another LOC I need you to change.

Other than that, I agree with both of you.
- christoph: the full advantages will only show once alle BSFormatter-Stuff is gone from the constructors. agreed.
- bodymindarts: reducing the API is useful as well as removing dead code. agreed.

nothing is perfect.

please refer the followup PR to this one and we are good.

.idea/codeStyles/codeStyleConfig.xml
bodymindarts: oh whoops... sorry. I thought I had fixed this but it came back :-(

freimair (approved):
Ack.

PR #3241, opened 2019-09-11
Revert DAO changes to add bonded role for Analytics Operator

wiz:
We decided to revert this, per discussion in https://github.com/bisq-network/proposals/issues/115

### DIFF
--- a/core/src/main/java/bisq/core/dao/state/model/governance/BondedRoleType.java
+++ b/core/src/main/java/bisq/core/dao/state/model/governance/BondedRoleType.java
@@ -53,7 +53,7 @@ public enum BondedRoleType {
 
     // operators
     WEBSITE_OPERATOR(50, 110, "https://bisq.network/roles/12", true),
-    ANALYTICS_OPERATOR(2, 110, "https://bisq.network/roles/88", true),
+    ANALYTICS_OPERATOR(1, 110, "https://bisq.network/roles/88", true),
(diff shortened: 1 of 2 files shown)


ripcurlx (approved):
utACK - As mentioned by DAO devs it might work, but it is too much risk to do so without in-depth testing.

chimp1984:
@wiz @ripcurlx Thanks!

I was thinking more about the possible consquences and I am now pretty sure that it would have had serious issues. I try to explain (not 100% sure without going through the code and testing but 99% sure).

If a bonded role request would have been done, not updated users would have been out of sync (diff. proposal hash, their vote would be invalid). If not more then 80% of voters (80% by stake) would have either updated or not updated the whole voting cycle would have been invalid. We require that a super majority (> 80%) of voters have the same p2p network data (by hash) on which voting is based on.

So you see tiny trivial seeming changes can have severe consequences in the DAO.

If we need to add a field we need to announce it early enough so DAO stakeholders have time to update. We can enforce it by Filter.disableDaoBelowVersion so not updated users cannot partizipate in governance activities in the DAO. So it it is well planned we are able to do small changes but also those would require very careful thoughts and testing and contain some risks.

While Betanet testing we had such a case (caused by other reasons) where we did not reach 80% and voting round failed. This was also due unexpected reasons and it took a while to find out why it happened (so far to risks even if you are sure all will go fine).

**We have to consider the DAO as set in stone and changes in concensus need to justify the risk and should be considered exceptional.**

Issue #3240, opened 2019-09-11
Remove / refactor BSFormatter (tracking issue)


bodymindarts:
While making myself familiar with the code and thinking about improvements I kept seeing complexities connected to the [BSFormatter](https://github.com/bisq-network/bisq/blob/master/core/src/main/java/bisq/core/util/BSFormatter.java) class.

The code base is quite tangled up and that class is one of the places that contributes because:
- it is bucket class with many many unrelated functions
- it has a _huge_ public interface even though most functions don't depend on instance state
- it depends on globals
- it is depended on by many classes
- it is part of a type hierarchy

This led me to refactor it out of existence. The final state of the code is [here](https://github.com/bisq-network/bisq/pull/3230) but it is too much to review and merge in 1 step.
To get the code changes in I can break it down into 5-6 trivial commits / PRs, but don't want to do this work unless other ppl agree that removing this class makes sense.
Please give 👍 or 👎 as feedback on wether I should go ahead. Perhaps I am the only one who has seen these issues, or they aren't hurting anyone and the code can stay as is. In that case I don't want to make the effort.

bodymindarts:
If you do think this should be taken care of please review the first (trivial) PR https://github.com/bisq-network/bisq/pull/3239.

bodymindarts:
Reposting the design principles that led me to this refactoring:

- Minimize dependencies on all levels. ie. small public interfaces for classes and libraries.
- Prefer static functions for stateless helpers that don't contain domain logic
- Prefer composition over inheritance for code reuse
- Prefer interfaces over inheritance for polymorphism
- Don't depend on global state in regular classes, better to inject it via constructor

I assume these are not contentious and am omitting justification. Happy to discuss these points if there is dissent.

bodymindarts:
Here is step 2. Extract 2 classes https://github.com/bisq-network/bisq/pull/3244

bodymindarts:
Completed

PR #3244, opened 2019-09-11
Extract helper classes from BSFormatter

bodymindarts:
Next step in removing BSFormatter. Here I extract 2 classes.

DisplayUtils which lives in the desktop project and has all methods that are only used in desktop. This is to reduce the number of dependencies going from desktop -> core.

ParsingUtils which has all the parsing related methods (ie is not formatting).

Link to tracking issue https://github.com/bisq-network/bisq/issues/3240


freimair (approved):
Ack

PR #3246, opened 2019-09-11
Fix unstyled chat trade message states and icons

wiz:
Screen Shot 2019-09-12 at 2 11 19
Screen Shot 2019-09-12 at 2 10 46


ripcurlx (approved):
utACK

PR #3250, opened 2019-09-12
Remove CapabilityRequiringPayload from NewBlockBroadcastMessage

chimp1984:
We remove the CapabilityRequiringPayload interface to avoid risks
that new BSQ blocks are not well distributed in case the capability is
not exchanged at the time when the message is sent. We need to improve
the capability handling so that we can be sure that we know the actual
capability of the peer.


ripcurlx (approved):
utACK

PR #3251, opened 2019-09-13
Fix invalid comparators

stejbac:
Many of the table column comparators (as well as _PaymentMethod.compareTo_) are invalid, since they immediately return 0 upon encountering a null/missing field. This violates the _Comparator_/_Comparable_ contract, which expects a total order on the objects and, in particular, an equivalence relation given by the result of "_compare(x, y) == 0_".

This PR attemps to fix all the invalid _Comparator_ / _Comparable_ instances (which are almost all table column comparators), by ordering objects with a null/missing (sub)field ahead of the populated ones. In most cases this will probably have no effect, as nulls won't be encountered for that column in practice. However, the current flawed comparators are at least breaking row sorting in the Portfolio/History table, as seen in the following two screenshots:

Closed Trades with Broken Comparators 1
Closed Trades with Broken Comparators 2

The ascending/descending sorting of the "Amount in BTC" field does not make sense, due to a broken comparator passed to the underlying FX library insertion sort. After the fix, the rows appear as follows:

Closed Trades with Fixed Comparators 1
Closed Trades with Fixed Comparators 2

The missing cells are deemed smaller than all the others in the column, which is consistent with _TableColumnBase.DEFAULT_COMPARATOR_ from the openjfx library.

### DIFF
--- a/core/src/main/java/bisq/core/payment/payload/PaymentMethod.java
+++ b/core/src/main/java/bisq/core/payment/payload/PaymentMethod.java
@@ -42,7 +42,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 @EqualsAndHashCode(exclude = {"maxTradePeriod", "maxTradeLimit"})
 @ToString
 @Slf4j
-public final class PaymentMethod implements PersistablePayload, Comparable {
+public final class PaymentMethod implements PersistablePayload, Comparable<PaymentMethod> {
 
     ///////////////////////////////////////////////////////////////////////////////////////////
     // Static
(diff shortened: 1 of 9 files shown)


ripcurlx:
@stejbac Could you please resolve the conflict because of the recently merged release branch.

stejbac:
I've just rebased the two commits to resolve the conflict, which appears to be simply due to a file rename: _TraderDisputeView_ -> _DisputeView_.

freimair (approved):
Ack

PR #3258, opened 2019-09-13
Fix various bugs related to Japan Bank Transfer payment method

wiz:
* Add missing call to addFormForBuyer(), improve BuyerStep2View UI
* Fix Account Creation UI elements width to fill entire pane
* Fix Account Creation UI elements padding to avoid overlap
* Tweak trade display popup text and various strings
* Remove @Deprecated tag


ripcurlx:
Here is an unnecessary line break that creates this single "." in the next line.
Bildschirmfoto 2019-09-14 um 19 31 46
Bildschirmfoto 2019-09-14 um 19 46 08

To prevent scrolling the fields should be distributed more efficiently between the two columns.
Bildschirmfoto 2019-09-14 um 19 37 47
Bildschirmfoto 2019-09-14 um 19 46 36
It is still not perfect. Maybe for the next release we could combine two fields into one, so we don't have a scrollbar at all.

To save time I'll fix this issues and will push it directly to this branch.

ripcurlx (approved):
ACK - after my changes

PR #3259, opened 2019-09-14
Remove CapabilityRequiringPayload from GetBlocksRequest

chimp1984:
We remove CapabilityRequiringPayload as it would cause problems if the
lite node connects to a new seed node and they have not exchanged
capabilities already. We need to improve capability handling the we
can re-enable it again. As this message is sent any only to seed nodes
it does not has any effect. Even if a lite node receives it it will
be simply ignored.


ripcurlx (approved):
utACK

PR #3262, opened 2019-09-15
Update IP address for @mrosseel's 2 hard-coded BTC nodes

wiz:
My monitoring system discovered that the hard-coded IP addresses for @mrosseel 2 BTC nodes were unreachable. After looking into it, seems he changed the IP addresses for his nodes but didn't update the hard-coded ones:
```
btc.vante.me has address 94.23.21.80
btc2.vante.me has address 94.23.205.110
```


chimp1984 (approved):
utACK

chimp1984:
@ripcurlx Coulf we get that intot the release as well?

mrosseel:
correct, thx for the PR. 
Should only have impact if DNS resolving fails.

PR #3263, opened 2019-09-15
Temporarily disable onion host for @KanoczTomas's BTC node

wiz:
My monitoring system is reporting that the Tor hidden service for this BTC node becomes unreachable every few minutes - I'm guessing it's a firewall configuration issue on their side, but since I have no way to contact them I'm submitting this PR to temporarily remove it for now until the issue can be resolved. Hopefully if someone can contact @KanoczTomas to fix it we won't need to merge this PR :)

Screen Shot 2019-09-15 at 9 55 51

### DIFF
--- a/core/src/main/java/bisq/core/btc/nodes/BtcNodes.java
+++ b/core/src/main/java/bisq/core/btc/nodes/BtcNodes.java
@@ -69,7 +69,7 @@ public class BtcNodes {
                         new BtcNode("btc2.sqrrm.net", "i3a5xtzfm4xwtybd.onion", "81.171.22.143", BtcNode.DEFAULT_PORT, "@sqrrm"),
 
                         // KanoczTomas
-                        new BtcNode("btc.ispol.sk", "mbm6ffx6j5ygi2ck.onion", "193.58.196.212", BtcNode.DEFAULT_PORT, "@KanoczTomas"),
+                        new BtcNode("btc.ispol.sk", null, "193.58.196.212", BtcNode.DEFAULT_PORT, "@KanoczTomas"),
 
                         // Devin Bileck
                         new BtcNode("btc1.dnsalias.net", "lva54pnbq2nsmjyr.onion", "165.227.34.198", BtcNode.DEFAULT_PORT, "@devinbileck"),


chimp1984:
As far I remember from monitoring with http://seedmonitor.0-2-1.net @KanoczTomas's BTC node was mostly stable. So would prefer to wait for feedback from @KanoczTomas befor removing the node.

wiz:
To be clear, yes, the node itself is stable, my monitoring system is only reporting issues for this node's *.onion hidden service. As you can see below, the onion hidden service does not return a valid response most of the time, and its graph is mostly empty (black).

Clearnet:
clearnet

Onion:
onion

mrosseel:
Did you try restarting tor? I've seen it get stuck on a bad route.
If it's consistently unreachable and the operator does not respond then disabling seems the right move.

ripcurlx:
As mentioned in https://github.com/bisq-network/roles/issues/67#issuecomment-532622661 I filtered out the onion address of @KanoczTomas node right now. Based on @wiz monitoring feedback it seems as if the node recovered its connectivity because of the reduced load. I won't merge this PR right now until @KanoczTomas comes back addressing the Tor issue.

KanoczTomas:
Hi! It is certainly not a firewall issue, as there was no change on the network. It might be a congestion issue, I will try stopping a service that I was testing a few days ago, hope that is the culprit. @wiz do  you have a date when it started, or is the current state still the same. There was a power fluctuation issue as well in the building and hardware behaved weirdly because of that. I did power cycle the boxes and it seemed fine on my side.

wiz:
Hey @KanoczTomas glad to hear you are okay, we were worried. Basically as soon as I setup the new btcnode monitoring system, it showed your node and a couple others were not reliably responding to requests on the onion URI - we gave it a few days but since we didn't hear back we added a temporary filter. I can give you access to the monitoring system's web interface if you want to see the data, but currently it only replied to about 15 of the last 60 queries:

Screen Shot 2019-09-20 at 5 41 58

ripcurlx:
Hi @KanoczTomas! As now also your public node 193.58.196.212:8333 is throwing more and more errors (see the #alerts channel) I have to filter it out as well. I'll merge this PR for now, but I'm happy to add your nodes again as soon as you have time again to monitor them on a regular basis.

PR #3266, opened 2019-09-16
Fix DAO sync issue requiring a manual kill-and-restart

### LINKED ISSUE #3094: Improve BSQ block requests
Currently we request BSQ blocks from only one seed node. If it fails there is a rather long timeout until we try the next. In case a Bisq node does not receive a new block (after initial parsing is completed) there is currenttly no mechanism in place to request missing blocks and the user has to restart. 
We should look into strategies how to get the initial block request more reliable and resiliant as well as to repeat requests in case we are not on the chaintip with our BSQ blocks.

Another open issue is that in case of a resync from genesis is triggered, the node only receives max. 6000 blocks as that is the limit a seednode provides (ddos protection). At the moment there is no mechanisma for repeating the call so that the missing blocks are requested again and to be able to get to the chain tip. A restart is atm required to get the rest of the blocks. We should add a request from another seednode for the missing blocks once the parsing of the first 6000 blocks is done.

The parsing must not be interrupted, so we need some smart handling of parallel received data and handling of yhe data deltas.

freimair:
This PR prevents the Bisq application to never get its DAO status updated without a force-kill and restart of the application over and over again.

This fix is not final, however, this core issue is not easily resolved. The DAO sync logic tries one seed node after another until the DAO status is received. The very first attempt targets the seed node that already delivered the initial data. However, when the first attempt fails for whatever reason, other seed nodes are tried. The issue is that even if a connection can be summoned to such a seed node, the local Bisq client does not know its capabilities (these only get updated once a node responds with a suitable message), the DAO sync message, however, requires a capability. Bottom line is, the local Bisq app does not send the DAO sync message to seed nodes other than the initial one.

This PR fixes the issue by having the DAO sync message not requiring any capabilities. As this change only affects clients and not other nodes, this does not change the attack surface of the Bisq network.

fixes #3094

### DIFF
--- a/core/src/main/java/bisq/core/dao/node/messages/GetBlocksRequest.java
+++ b/core/src/main/java/bisq/core/dao/node/messages/GetBlocksRequest.java
@@ -39,8 +39,7 @@ import javax.annotation.Nullable;
 @EqualsAndHashCode(callSuper = true)
 @Getter
 @Slf4j
-public final class GetBlocksRequest extends NetworkEnvelope implements DirectMessage, SendersNodeAddressMessage,
-        CapabilityRequiringPayload, SupportedCapabilitiesMessage {
+public final class GetBlocksRequest extends NetworkEnvelope implements DirectMessage, SendersNodeAddressMessage, SupportedCapabilitiesMessage {
     private final int fromBlockHeight;
     private final int nonce;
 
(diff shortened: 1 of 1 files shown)


chimp1984 (commented):
The same was done in PR https://github.com/bisq-network/bisq/pull/3259 which is merged into 1.1.6 and will get into master soon.

PR #3278, opened 2019-09-17
Change dao.proposal.display.name to 'Exact GitHub username'

### LINKED ISSUE #3274: Make more clear that DAO users use their Github handle as name
When making a proposal we have currently "Name/Nickname" in the name field. We should use a more strict term to make clear that the name should match exactly the Github user name. This helps to avoid confusions and possible "squatter" attacks.
I would suggest: "Exact Github username"

erciccione:
Fixes #3274


chimp1984 (approved):
utACK

ripcurlx (approved):
utACK

PR #3279, opened 2019-09-17
Remove reactfx dependency

battleofwizards:
The reactfx was a 0.33MB dependency / 238 Java source files.

We only used it for FxTimer helper and to measure FPS in development mode.

The small FxTimer class got copied over into the project, so there is no change in semantics.

The FPS measurement got replaced w/ com.sun.javafx.perf.PerformanceTracker, an internal but long stable part of JavaFx. This is probably not a problem, because it only concerns a rarely used, developer feature.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -64,7 +64,6 @@ configure(subprojects) {
         protobufVersion = '3.9.1'
         pushyVersion = '0.13.2'
         qrgenVersion = '1.3'
-        reactfxVersion = '2.0-M3'
         sarxosVersion = '0.3.12'
         slf4jVersion = '1.7.22'
         sparkVersion = '2.5.2'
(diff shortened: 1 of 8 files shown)


ripcurlx:
@battleofwizards Doing so I think we need to add the copyright notice into the copied file as mentioned in https://opensource.org/licenses/BSD-2-Clause.

chimp1984:
@battleofwizards The FPS tool could be removed as well.

chimp1984 (approved):
utACK

battleofwizards:
LICENSE added. The old FPS tool removed. Thanks @ripcurlx and @chimp1984!

ripcurlx (approved):
utACK

PR #3283, opened 2019-09-17
use gradle dist with sources for better IDE experience.

christophsturm:
since my pr that upgrades to latest gradle is still not merged here is a more conservative pr.


battleofwizards:
ACK (tested).

Issue #3276, opened 2019-09-16
If audio is muted on OSX standby gets activated 


chimp1984:
The hack to play a sound to avoid network losses due throttled resources by standby mode does not work if you have muted audio. 
We should find a more reliable way to fix that problem.

christophsturm:
what about `pmset noidle`? 
http://osxdaily.com/2011/03/21/temporarily-prevent-a-mac-from-sleeping/

```
~% pmset noidle
Preventing idle sleep (^C to exit)...
```

christophsturm:
even better: `/usr/bin/caffeinate ./bisq-desktop`. if this works we can just add caffeinate to the osx start script.

christophsturm:
it seems `AvoidStandbyModeService` runs on all systems. is it only needed on osx? then we could remove it and use caffeinate instead. Or otherwise we just keep it in and run caffeinate in addition via the mac os startscript.

chimp1984:
It is not tested for other systems but it can be assumed that Windows might have similar behaviour.

devinbileck:
I haven't done any testing with muted audio on Windows, but can if necessary.
Additionally, perhaps this is the cause of #3223?

wiz:
Nah I think Linux power management is very primitive and doesn't care if you are listening to audio

christophsturm:
maybe we can do something similar to #3284 on linux too? https://wiki.archlinux.org/index.php/Power_management#Console

freimair:
looks like this would be a good time to rethink the avoid-standby stuff. I will provide info for arch linux.

PR #3284, opened 2019-09-17
Use /usr/bin/caffeinate to disable power management on macOS

christophsturm:
a possible fix for #3276 

I verified that caffeinate is correctly called and that it keeps running until bisq exits. the other standby mode service is still running, not sure if we want to disable it on macOS


sqrrm (approved):
utACK

I would be hesitant to disable the other mode (playing a sound) until it's well tested that this fix actually works as intended. The issues resulting from sleep mode are quite severe.

chimp1984:
I tested a bit with removed sound and still received all dao hashes (over night). I have not tested taking offers yet but testing with dao hashes is not sufficient. Taking offers failed if the makers app was throttled by standby mode, but exact circumstances are unclear. To test it reliable someone need to make it reproduceable without the sound and then use the caffeinate and see it it avoids the issue. Its cumbersome to test but we should not risk to remove the sound hack without being sure...

PR #3287, opened 2019-09-18
Remove controlsfx dependency

battleofwizards:
The `controlsfx` is a huge JavaFX components library (0.83MB as JAR) offering a number of advanced components.

Turns out, we only use a single component, the `PopOver`. Fortunately, the `PopOver` is self-contained and easy to pull out from `controlsfx`.

This PR embeds the `PopOver` in the project and removes `controlsfx` dependency. The embedded version is exactly the same as previously used in the project, so no change in semantics should occur.

Note for reviewers: see README.md for exact versions and links to files that were copied over. No new code is introduced. The easiest way to verify is to diff the embedded vs external (3 files total).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,6 @@ configure(subprojects) {
         bitcoinjVersion = 'a88d36d'
         btcdCli4jVersion = '975ff5d4'
         codecVersion = '1.9'
-        controlsfxVersion = '8.0.6_20'
         easybindVersion = '1.0.3'
         easyVersion = '4.0.1'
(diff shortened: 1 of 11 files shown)


battleofwizards:
"Some checks haven't completed yet" - they did complete green. Apparently GitHub lost notification from Travis. Click "Details" below to view directly.

chimp1984 (approved):
Great thanks! Master of dependency reduction!

utACK

I compared all changed classes and beside packages and formatting all is same.

PR #3125, opened 2019-08-21
List Ergo (ERG)

jasondavies:
- Official project URL: https://ergoplatform.org/
- Official block explorer URL: https://explorer.ergoplatform.com/

Example link to transaction: https://explorer.ergoplatform.com/en/transactions/fc9d78a3ca55d7950a928d55ef1676cc9ba516e4ef107d955de2fe541390ecc3

PR #3289, opened 2019-09-18
Add Ergo (ERG) without Bouncy Castle dependency.

jasondavies:
See #3195.


jasondavies:
Note: this is exactly the same as the code in #3125, which is currently in v1.1.6 and v1.1.7, just with the checksum validation removed as it required Bouncy Castle.

blabno (approved):
ACK

PR #3296, opened 2019-09-19
chop down long chained calls and force brackets on if

christophsturm:


chimp1984 (approved):
utACK

Just quickly tried it out with streams and looks ok.

mrosseel:
Not sure what the current status is of .editorconfig ? 
Being ide-agnostic seems like a win for me, maybe the idea config is more convenient but then maybe also keep .editorconfig in sync as much as possible?

chimp1984:
NACK

Actually it does not behave as i would like:

If its not a long line it makes it a one-liner. I prefer linebreaks at all non-trivial stream expressions for better redibility.
```
 public Optional<TxOutput> getLockupTxOutput(String txId) {
        return getTx(txId).flatMap(tx -> tx.getTxOutputs().stream().filter(this::isLockupOutput).findFirst());
    }
```

```
   public Optional<TxOutput> getLockupTxOutput(String txId) {
        return getTx(txId).flatMap(tx -> tx.getTxOutputs().stream()
                .filter(this::isLockupOutput)
                .findFirst());
    }
```

ripcurlx:
> NACK
> 
> Actually it does not behave as i would like:
> 
> If its not a long line it makes it a one-liner. I prefer linebreaks at all non-trivial stream expressions for better redibility.
> 
> ```
>  public Optional<TxOutput> getLockupTxOutput(String txId) {
>         return getTx(txId).flatMap(tx -> tx.getTxOutputs().stream().filter(this::isLockupOutput).findFirst());
>     }
> ```
> 
> ```
>    public Optional<TxOutput> getLockupTxOutput(String txId) {
>         return getTx(txId).flatMap(tx -> tx.getTxOutputs().stream()
>                 .filter(this::isLockupOutput)
>                 .findFirst());
>     }
> ```

Unfortunately merged it already because of the utACk before. @christophsturm could you address the comments in another PR?

chimp1984:
@ripcurlx @christophsturm @freimair 
Lets try to get to a final version soon and then make a reformat of the whole code base to avoid future format diffs. After that, there should be only changes if super important as a full reformat should not become a standard procedure.  Leaving a bit of freedom to devs about their preferred style is ok IMO as well. Usually there is a main dev for a class so it can carry a bit of his signature...

PR #3302, opened 2019-09-20
Monitor fixes

freimair:
The monitor has been offline for a couple of hours because:
- the Capability stuff changed
- data responses finally got too big

This PR fixes all of that and adds a few tweaks as well:
- Refer Seednode diffs to the global unique number of messages instead of a random seed node
- add a new metric that lets us get an idea on how the bisq software version landscape looks like
- make the DAO monitoring more deterministic


chimp1984 (approved):
utACK

Just had a very rough look over it....

PR #3309, opened 2019-09-22
Revert recent code format changes

chimp1984:
The recent changes breaks current formatting radically and does not
lead to better results as it was before.
Before applying new changes, please test it with bigger classes to see
if the result is as intended.


ripcurlx (approved):
utACK

PR #3310, opened 2019-09-22
Fix missing handling of mediator mailbox msg

chimp1984:
We have a bug with the handling of missing mailbox messaged from mediation result and the peers accept message (carrying the signature). This PR fixes that.

For testing shut down the other trade peer at each step after mediation gets closed.
1. Mediation is opened 
2. Bob shutd down
3. Alice accept result
4. Alice shuts down
5. Bob starts and accepts result
6. Alice starts and sees the trade got closed


ripcurlx (approved):
utACK

ripcurlx:
ACK - found another small translation bug, that will be added to the hotfix as well.

PR #3311, opened 2019-09-22
New trade protocol

chimp1984:
Implements https://github.com/bisq-network/proposals/issues/52

This is work in progress but as the protocol is already implemented I thought it might be useful to share progress on it via making a draft PR. 

It has already lots of code change and more will come. It is not expected that other devs review every commit here but to review the full diff once it is completed (similar to mediation support PR).

The new trade protocol will have a new TRADE_PROTOCOL_VERSION to separate it from not updated users, but it will not require a hard fork. Users need to complete their pending trades and distputes before updating. Updates while having pending trades and distputes would cause failures of the trade.


chimp1984:
Replaced by https://github.com/bisq-network/bisq/pull/3333

Issue #3231, opened 2019-09-09
Table view columns should use a numeric-sort algorithm


wiz:
<img width="103" alt="Screen Shot 2019-09-09 at 22 34 48" src="https://user-images.githubusercontent.com/232186/64535664-a3c6bb00-d352-11e9-912d-40b0251e4bcb.png">

niyid:
@freimair 
Who is assigning? I suppose this issue is to be resolved in all table views and not only Bonds view?

niyid:
Hi @sqrrm 

Thanks for this. If you do not mind, please also take a look at this:

https://github.com/bisq-network/bisq/pull/3932

PR #3316, opened 2019-09-24
Issue #3231 Fix

niyid:
Fixed sorting in views with TableView instances with no implemented sort.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/dao/bonding/bonds/BondsView.java
+++ b/desktop/src/main/java/bisq/desktop/main/dao/bonding/bonds/BondsView.java
@@ -185,9 +185,9 @@ public class BondsView extends ActivatableView<GridPane, Void> {
                 };
             }
         });
+        tableView.getColumns().add(column);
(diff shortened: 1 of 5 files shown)


wiz:
Thanks for fixing this issue, haven't tested yet but looks good so far. Could you please fix the following?
- Swap the title and description of this PR, so that the title is "Fixes sorting in views with TableView instances with no implemented sort" and the description is "fixes #3231"
- Squash the 3 commits into a single commit
- The PR branch should not be named "master" but instead something like "fix-table-sort" to identify what it does

wiz (changes requested):
Could you please fix the following:
- Swap the title and description of this PR, so that the title is "Fixes sorting in views with TableView instances with no implemented sort" and the description is "fixes #3231"
- Squash the 3 commits into a single commit
- The PR branch should not be named "master" but instead something like "fix-table-sort" to identify what it does

niyid:
OK. I will get to it.

freimair:
closed as duplicate of https://github.com/bisq-network/bisq/pull/3319

PR #3320, opened 2019-09-24
Show only active roles in "Bonded Roles" table

### LINKED ISSUE #3271: Rejected role is shows in bonded roles
A rejected bonded role request (lio cycle 5) is displeyed in the bonded roles table. By clicking the role it throws an exception.
Expected: Rejected requests are filtered out.

```
java.lang.IllegalArgumentException: bondedRoleType must be present
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:122)
	at bisq.core.dao.DaoFacade.getRequiredBond(DaoFacade.java:721)
	at bisq.desktop.main.dao.bonding.roles.RoleDetailsWindow.addContent(RoleDetailsWindow.java:88)
	at com.sun.javafx.tk.quantum.GlassViewEventHandler$MouseEventNotification.run(GlassViewEventHandler.java:295)
	at bisq.desktop.main.dao.bonding.roles.RoleDetailsWindow.show(RoleDetailsWindow.java:66)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at bisq.desktop.main.dao.bonding.roles.RolesView$3$1.lambda$updateItem$0(RolesView.java:214)
```

ben-kaufman:
Fix #3271.
Filter the bonded roles list to contain only active bonds in the RolesView table.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/dao/bonding/roles/RolesView.java
+++ b/desktop/src/main/java/bisq/desktop/main/dao/bonding/roles/RolesView.java
@@ -27,6 +27,7 @@ import bisq.desktop.util.FormBuilder;
 import bisq.desktop.util.GUIUtil;
 
 import bisq.core.dao.DaoFacade;
+import bisq.core.dao.governance.bond.Bond;
 import bisq.core.dao.governance.bond.BondState;
 import bisq.core.dao.governance.bond.role.BondedRole;
 import bisq.core.dao.state.model.governance.BondedRoleType;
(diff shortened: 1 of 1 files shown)


chimp1984 (approved):
utACK

wiz:
I don’t think this is the correct fix, because the table needs to show inactive bonds that have been approved but not yet posted.

sqrrm (changes requested):
NACK

As @wiz says, this will filter out all non committed bonds. What needs to be done is to filter out failed proposals to take on a bonded role but still show those role proposals that have been accepted by the DAO but not yet locked up their funds in a bond.

chimp1984:
>I don’t think this is the correct fix, because the table needs to show inactive bonds that have been approved but not yet posted.

Yes you are right! Sorry for the too quick utACK!  Good to have more eyes on that!   

NACK

ripcurlx:
@ben-kaufman Are you up for making the required changes? See comments above ^^

wiz:
@ripcurlx I looked into this at the time the issue was created, but I didn't try to fix it because I thought it would be too dangerous to make the necessary changes to DAO code to filter the data as needed.. which I guess is why the bug exists in the first place. I suggest to close as WONTFIX unless you or chimp wants to fix it.

chimp1984:
Replaced by https://github.com/bisq-network/bisq/pull/3676

This PR can be closed.

PR #3321, opened 2019-09-24
inject DaoKillSwitch

christophsturm:
just a small refactoring to reduce global state, while i investigate transaction broadcasting.

### DIFF
--- a/core/src/main/java/bisq/core/dao/DaoKillSwitch.java
+++ b/core/src/main/java/bisq/core/dao/DaoKillSwitch.java
@@ -70,8 +70,8 @@ public class DaoKillSwitch implements DaoSetupService {
         daoDisabled = requireUpdateToNewVersion || filter.isDisableDao();
     }
 
-    public static void assertDaoIsNotDisabled() {
-        if (INSTANCE.isDaoDisabled()) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/bisq/core/dao/DaoKillSwitch.java
```diff
@@ -70,8 +70,8 @@ private void applyFilter(@Nullable Filter filter) {
         daoDisabled = requireUpdateToNewVersion || filter.isDisableDao();
     }
 
-    public static void assertDaoIsNotDisabled() {
-        if (INSTANCE.isDaoDisabled()) {
```
mrosseel, line 74: is the INSTANCE still needed after this refactoring?

christophsturm: no.

mrosseel:
did the simplify commit accidentally get added to this PR? Seems out of scope. Otherwise utACK

christophsturm:
i added it because its a local change that i made and that seemed simple enough. i can put it into a different pr where it fits better

christophsturm:
removed the unrelated commit

sqrrm (approved):
utACK

Issue #3223, opened 2019-09-07
V1.1.6 "Avoid Standby Mode" does not work on Ubuntu 18


ghubstan:
[How to Reproduce]

1.  Create a VM:  Ubuntu 18.04.3 LTS
	Kernel 5.0.0-27-generic 28~18.04.1-Ubuntu
	GNOME Shell 3.28.4 
	Apt update/upgrade should give you the same or more recent kernel & gnome versions.
	
2.  Open VM's Settings->Power->Automatic Suspend->Plugged In
	Set ON with Delay of 15 minutes.
	
3.  Install Bisq (w/ v1.1.6 test jar) and start an Arbitration client.
    Enable Settings -> "Avoid Stanby Mode".
    
4.  Don't touch anything for 15 minutes; watch Ubuntu 18 go into standby mode.

[Expected Results]

The Ubuntu 18 VM should not go into standby mode.

[Actual Results]

The Ubuntu 18 VM goes into standby mode.

wiz:
Ah yeah, it's not going to work on Linux. IMO the correct fix for Linux would be to simply hide the feature. Users on Linux can disable the OS sleep mode on their own.

niyid:
Hello. Is anybody working on this yet?

ghubstan:
@niyid , I am working on this now.

ghubstan:
@chimp1984, the row adjustment fix has been committed.

I have not confirmed there is no need for the Unix sound hack.  Can @wiz or @niyid confirm there is no need for the Unix sound hack?

PR #3322, opened 2019-09-24
Hide avoid standby mode feature on *nux OS

ghubstan:
Displays standby mode button on only on Windows and OSX,
and hides it on Linux and Unix distributions.

Formatted source file, resulting in rearranged class
level field declarations.

Fix for Issue #3223

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/settings/preferences/PreferencesView.java
+++ b/desktop/src/main/java/bisq/desktop/main/settings/preferences/PreferencesView.java
@@ -53,6 +53,7 @@ import bisq.core.util.validation.IntegerValidator;
 import bisq.common.UserThread;
 import bisq.common.app.DevEnv;
 import bisq.common.util.Tuple3;
+import bisq.common.util.Utilities;
 
 import org.bitcoinj.core.Coin;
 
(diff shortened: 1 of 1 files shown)


chimp1984 (changes requested):
Missing adjustment of num rows.
`TitledGroupBg titledGroupBg = addTitledGroupBg(root, gridRow, 8, Res.get("setting.preferences.general"));` -> 8 need to be 7 in case its not displayed

Is it confirmed that on Unix there is no need for the hack with playing the sound to avoid standby?

ghubstan:
@chimp1984, the row adjustment fix has been committed.

I have not confirmed there is no need for the Unix sound hack. Can @wiz or @niyid confirm there is no need for the Unix sound hack?

desktop/src/main/java/bisq/desktop/main/settings/preferences/PreferencesView.java
```diff
@@ -97,14 +98,22 @@
...
+    private final Preferences preferences;
+    private final FeeService feeService;
+    //private final ReferralIdService referralIdService;
+    private final AssetService assetService;
+    private final FilterManager filterManager;
+    //private ComboBox<BaseCurrencyNetwork> selectBaseCurrencyNetworkComboBox;
+    private final DaoFacade daoFacade;
+    private final BSFormatter formatter;
+    private final boolean daoOptionsSet;
+    private final boolean displayStandbyModeFeature;
```
freimair, line 110: please do not commit non-changes. makes it hard to review and introduces unnecessary big change sets.

freimair (changes requested):

desktop/src/main/java/bisq/desktop/main/settings/preferences/PreferencesView.java
ghubstan: I don't understand.   Are the non-changes you mentioned references to the comments (lines 103, 106)?  Those were there from previous commits (before this PR).  I did not add these comments.  Should I delete them?

Or, are you referring to the Intellij code reformatting which rearranged the order of field definitions?  I have been reformatting files I touch, using the IDE code style versioned in git.  Should I not do that?

freimair: > Or, are you referring to the Intellij code reformatting which rearranged the order of field definitions? I have been reformatting files I touch, using the IDE code style versioned in git. Should I not do that?

that is what I meant. please only commit changes that have something to do with the topic of your PR. Thanks!

PR #3327, opened 2019-09-25
Check for dust after subtracting fee, not before.

christophsturm:
I think this is a bug.


sqrrm (approved):
utACK

@chimp1984 could you have a look as well to make sure this is right.

freimair (approved):
utAck

PR #3334, opened 2019-09-27
Signed witness trading

sqrrm:


sqrrm:
There is a failing test but I can't make it fail locally, not sure what the issue is.

ripcurlx (approved):
ACK - it is in an in-development state, but we decided to merge it with the new trade protocol to increase developer testing for both features

PR #3339, opened 2019-09-27
New trade protocol rc 1.2.0

chimp1984:


chimp1984:
Ah wrong target branch ;-(

PR #3341, opened 2019-09-27
Remove log.isDebugEnabled() check

chimp1984:
If log level is INFO it would not call logContents which is important
logging for seed nodes.


ripcurlx (approved):
utACK

PR #3344, opened 2019-09-28
Translate name of Japan Bank Transfer for non-Japanese users

wiz:
For foreigners in Japan, let's use translation instead of romanization


ripcurlx (approved):
utACK

PR #3348, opened 2019-10-01
Changes needed for minimal api

blabno:


mrosseel:
utACK, seems like it's mainly about adding some extra hooks during setup.

blabno:
@ripcurlx @sqrrm If you could merge this I would pull bisq/master into incubator-api/master and rebase https://github.com/bisq-network/incubator-bisq-api/pull/1 on top of it.

core/src/main/java/bisq/core/app/BisqHeadlessApp.java
```diff
@@ -43,7 +43,7 @@
     @Setter
     private GracefulShutDownHandler gracefulShutDownHandler;
     private boolean shutDownRequested;
-    private BisqSetup bisqSetup;
+    protected BisqSetup bisqSetup;
```
ripcurlx, line 46: Why do you need this to be protected instead of private?

desktop/src/main/java/bisq/desktop/app/BisqAppMain.java
```diff
@@ -61,7 +61,7 @@ public static void main(String[] args) throws Exception {
 
     @Override
     public void onSetupComplete() {
-        log.debug("onSetupComplete");
+        log.info("onSetupComplete");
```
ripcurlx: Is this really something we want to see and log already in the info debug level?

blabno:
So that it can be used by subclasses.

-------- Original Message --------

desktop/src/main/java/bisq/desktop/app/BisqAppMain.java
blabno: Good catch. I increased log level to better see this event inside docker. I'll revert it.

core/src/main/java/bisq/core/app/BisqHeadlessApp.java
blabno: So that it can be used by subclasses.

ripcurlx (approved):
utACK

PR #3356, opened 2019-10-03
Fix proposal table hide when switching to another screen and back

### LINKED ISSUE #3269: Vote result update issue
If one selects a cycle in vote result the proposals are listed below the cycle table. After switching to another screen and getting back the same syvle stays selected but the proposal table is not displayed. After switching another cycle and getting back the proposal table is visible again. 
Expected behaviour. Last selected cycle triggers also display of proposals.

stejbac:
Do not remove the proposals table when VoteResultView is deactivated, as the cycle selection which triggered its display is still active upon switching to another screen and back again. This fixes #3269.

Also remove a spurious _GUIUtil.removeChildrenFromGridPaneRows_ call when a proposal result is selected, as it is no longer needed since switching from an inline votes table to a popup.


freimair (approved):
Ack

PR #3358, opened 2019-10-03
Add an extra byte to TX size when calculating Bitcoin TX fees

wiz:
After investigating, it seems Bisq will sometimes create Bitcoin transactions with less than 1 sat/byte fee for TX, caused by:

1) User has set 1 sat/byte fee in settings
2) TX size gets rounded down when converting from bytes to vbytes (25% chance)

For example:
2080/4 = 520 (good)
2081/4 = 520.25 (bad, gets rounded down to 520)
2082/4 = 520.5 (good, gets rounded up to 521)
2083/4 = 520.75 (good, gets rounded up to 521)

In the case TX is 2081 bytes, 520 sats of fee gets used, for what should be considered a 521 vbyte TX. This results in a net fee of 0.998 sats/vbyte, which won't get relayed through Bitcoin network with the latest Bitcoin Core default settings.

This PR fixes the above issue by adding a workaround to add 1 byte to the TX size when calculating the fee, so it always "rounds up"


wiz:
This PR was replaced by #3387

PR #3364, opened 2019-10-03
Refactor FilterWindow

lusarz:
In this pull request I introduced `List<String> readAsList(InputTextField field)` function. It allowed me to remove a few repetitions of code. 

For example instead of:

    if (!nodesInputTextField.getText().isEmpty()) {
        nodes = new ArrayList<>(Arrays.asList(StringUtils.deleteWhitespace(nodesInputTextField.getText()).split(",")));
    }

I do:
    
    List<String> nodes = readAsList(offerIdsInputTextField)

Finally I was able to eliminate temporary variables like `nodes`, `bannedCurrencies` etc, and invoke `readAsList` directly.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/overlays/windows/FilterWindow.java
+++ b/desktop/src/main/java/bisq/desktop/main/overlays/windows/FilterWindow.java
@@ -41,8 +41,8 @@ import javafx.scene.layout.HBox;
 import javafx.geometry.HPos;
 import javafx.geometry.Insets;
 
-import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


desktop/src/main/java/bisq/desktop/main/overlays/windows/FilterWindow.java
```diff
@@ -301,4 +248,12 @@ private void addContent() {
         gridPane.getChildren().add(hBox);
         GridPane.setMargin(hBox, new Insets(10, 0, 0, 0));
     }
+
+    private List<String> readAsList(InputTextField field) {
+        if (field.getText().isEmpty()) {
+            return new ArrayList<>();
```
blabno: Use `Collections.emptyList()` instead of `new ArrayList<>()`.

blabno (changes requested):

desktop/src/main/java/bisq/desktop/main/overlays/windows/FilterWindow.java
lusarz: `Collections.emptyList()` returns an immutable list, so operations like `add` will throw `UnsupportedOperationException`. Are we sure that these lists are not modified anywhere ?

blabno: I'm not 100% sure how the result is being used. You can run Bisq with this change and open filter editor and see if anything blows up.

lusarz: Ok, I changed into `Collections.emptyList()` and did like you said - nothing wrong happened.

blabno (dismissed):
ACK

lusarz:
@ripcurlx merge ?

ripcurlx:
> @ripcurlx merge ?

As this is a very sensitive view I want to test it myself before merging. But atm I'm focusing on the v1.2.0 release branch right now.

Just something general: Refactorings are great, but it would be better to do them in the future while implementing feaures/fixing bugs with direct user benefit. As reviewing PRs is a bottleneck we need to be efficient and try to reduce review times as much as possible.

lusarz:
@ripcurlx I'm trying to do small refactorings which eliminates code repetition, according to the `Leave your code better than you found it.` rule. They are rather easy to review - I'm explaining everything in pull request description. 

Doing such a small refactorings help me to familiarize myself with the code.

freimair:
please resolve the conflicts so we can merge it

lusarz:
@freimair Conflicts resolved

freimair (dismissed):
utAck. I did fire up the thing and created a filter. Everything works as I would expect it. @ripcurlx?

ripcurlx:
@lusarz  This PR contains unsigned commits: https://help.github.com/en/github/authenticating-to-github/signing-commits

ripcurlx:
> utAck. I did fire up the thing and created a filter. Everything works as I would expect it. @ripcurlx?

@freimair Why utACK and not ACK in that case?

ripcurlx (changes requested):
NACK - please sign your commits. Besides that ACK. Tested it by filtering two currencies on Regtest.

lusarz:
@ripcurlx Commit is singed now

ripcurlx (approved):

PR #3374, opened 2019-10-08
Account Signing: Add information popups for signing state

ripcurlx:
... and other UI improvements

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqSetup.java
+++ b/core/src/main/java/bisq/core/app/BisqSetup.java
@@ -17,7 +17,9 @@
 
 package bisq.core.app;
 
+import bisq.core.account.sign.SignedWitness;
 import bisq.core.account.sign.SignedWitnessService;
+import bisq.core.account.witness.AccountAgeWitness;
 import bisq.core.account.witness.AccountAgeWitnessService;
 import bisq.core.alert.Alert;
 import bisq.core.alert.AlertManager;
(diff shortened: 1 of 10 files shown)


core/src/main/java/bisq/core/app/BisqSetup.java
```diff
@@ -729,9 +731,65 @@ private void maybeShowSecurityRecommendation() {
...
+        maybeTriggerDisplayHandler(key, displayHandler, signingStateFound);
+    }
+
+    private boolean isSignedWitnessOfMineWithState(PersistableNetworkPayload payload,
+                                                   AccountAgeWitnessService.SignState state) {
+        if (payload instanceof SignedWitness && user.getPaymentAccounts() != null) {
+            // We know at this point that it is already added to the signed witness list
+            // Check if new signed witness is for one of my own accounts
+
+            return user.getPaymentAccounts().stream()
```
sqrrm, line 777: I find this quite odd. First checking if the payload is a SignedWitness and then check if the user has a SignedWitness with that state. It's not actually checking if the payload witness has the desired SignState.

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -2541,6 +2539,13 @@ popup.accountSigning.signAccounts.ECKey.error=Bad arbitrator ECKey
 
 popup.accountSigning.success.headline=Congratulations
 popup.accountSigning.success.description=All {0} payment accounts were successfully signed!
+popup.accountSigning.generalInformation=You'll find the signing state of all your accounts in the account section.\n\n\
+  For further information please visit docs.bisq.network/account-signing.html.
+popup.accountSigning.signedByArbitrator=One of your payment accounts has been verified and signed by an arbitrator. Trading with this account will automatically sign your trading peer''s account after a successful trade.\n\n{0}
+popup.accountSigning.signedByPeer=One of your payment accounts has been verified and signed by a trading peer. You''re initial limit will be lifted within 30 days from now. \
```
sqrrm:
```suggestion
popup.accountSigning.signedByPeer=One of your payment accounts has been verified and signed by a trading peer. Your initial limit will be lifted within 30 days from now. \
```

sqrrm (approved):
ACK

PR #3377, opened 2019-10-08
Add combo block explorer for Blockstream.info + Mempool.space

wiz:
This PR adds a new option in the Preferences for a combination blockstream.info block explorer and mempool.space transaction visualizer.

https://mempool.space is a mempool visualizer that lets you see where your unconfirmed TX is and notifies you when it gets mined into a block like this:

Screen Shot 2019-10-09 at 1 03 27
Screen Shot 2019-10-09 at 1 03 00


ripcurlx (approved):
utACK

PR #3378, opened 2019-10-08
Disable @KanoczTomas btcnode on both clearnet and onion

wiz:
In PR #3263 we temporarily disabled @KanoczTomas btcnode's onion host, as it was having issues and it seems that @KanoczTomas is unavailable to maintain his btcnode. Unfortunately now it is generating monitoring system alerts on both clearnet and onion. This PR disables both clearnet and onion hosts, until @KanoczTomas returns and fixes his node. Hope he is okay.

Screen Shot 2019-10-09 at 1 07 50


ripcurlx (approved):
utACK - We'll re-add @KanoczTomas nodes as soon as they are stable again.

PR #3386, opened 2019-10-10
Fix mistakes in English source

ripcurlx:
Thanks to @MwithM for pointing them out.

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -789,9 +789,8 @@ portfolio.pending.disputeOpenedMyUser=You opened already a dispute.\n{0}
 portfolio.pending.disputeOpenedByPeer=Your trading peer opened a dispute\n{0}
 portfolio.pending.supportTicketOpenedByPeer=Your trading peer opened a support ticket.\n{0}
 portfolio.pending.noReceiverAddressDefined=No receiver address defined
-portfolio.pending.removeFailedTrade=If the mediator or arbitrator could not close that trade you can move it yourself \
-  to the failed trades screen.\n\
-  Do you want to remove that failed trade from the open trades screen?
+portfolio.pending.removeFailedTrade=Is this a failed trade? If so, would you like to manually \
+  close it, so that it no longer shows as an open trade?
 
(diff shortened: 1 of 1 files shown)


core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -2441,7 +2441,7 @@ popup.warning.lockedUpFunds=You have locked up funds from a failed trade.\n\
   Locked up balance: {0} \n\
   Deposit tx address: {1}\n\
   Trade ID: {2}.\n\n\
-  Please open a support ticket by selecting the trade in the pending trades screen and clicking \"alt + o\" or \"option + o\"."
+  Please open a support ticket by selecting the trade in the open trades screen and clicking \"alt + o\" or \"option + o\"."
```
```suggestion
  Please open a support ticket by selecting the trade in the open trades screen and pressing \"alt + o\" or \"option + o\"."
```

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -791,7 +791,7 @@ portfolio.pending.supportTicketOpenedByPeer=Your trading peer opened a support t
 portfolio.pending.noReceiverAddressDefined=No receiver address defined
 portfolio.pending.removeFailedTrade=If the mediator or arbitrator could not close that trade you can move it yourself \
```
m52go: There's no "failed trades" panel right now, as far as I know...just a "history" panel that shows successful trades and failed trades. 

Is that still the case?

Regardless, this text sounds overly complicated to me. Suggestion:

```
portfolio.pending.removeFailedTrade=Is this a failed trade? If so, would you like to manually \
close it, so that it no longer shows as an open trade?
```

ripcurlx: > There's no "failed trades" panel right now, as far as I know...just a "history" panel that shows successful trades and failed trades.

There are failed trades still in the code at least 😉 It is a separate tab next to the history tab.
Bildschirmfoto 2019-10-10 um 13 40 45
You can trigger this popup in the pending trades screen with `ALT+Y`
I'll adapt the text accordingly.
Bildschirmfoto 2019-10-10 um 13 43 49
Bildschirmfoto 2019-10-10 um 13 44 05

PR #3387, opened 2019-10-10
Increase minimum TX fee to 2 sats/vByte to fix #3106

### LINKED ISSUE #3106: Improve withdraw fee calculation
I tried to withdraw some BSQ from Bisq with a 1 sat/byte fee but the software miscalculated it and it ended outputting a value bellow 1 sat/byte. This caused the transaction to not be broadcast. I had to resync the SPV file and increase the custom withdraw transaction fee to 2 sat/byte.

There are two possible solutions:
* a) set the withdraw transaction fee to an obligatory minimum of 2 sat/byte
* b) fix the calculation of the transaction fee

I'd also like to say that this is a first occurrence, I've withdrawn many times at 1 sat/byte without coming across this issue.

wiz:
Trigger k3tan and also Don't allow users to set less than 2 sats/vByte TX fee in settings to prevent invalid TX with too low of a fee from being created

### DIFF
--- a/core/src/main/java/bisq/core/btc/BaseCurrencyNetwork.java
+++ b/core/src/main/java/bisq/core/btc/BaseCurrencyNetwork.java
@@ -73,6 +73,6 @@ public enum BaseCurrencyNetwork {
     }
 
     public long getDefaultMinFeePerByte() {
-        return 1;
+        return 2;
     }
 }


ripcurlx (approved):
utACK

ripcurlx:
@sqrrm Could you give this a quick look as well, if I overlooked any side effects of this change?

sqrrm (approved):
utACK

PR #3389, opened 2019-10-10
Verify the assumption that a HashMap orders keys.

christophsturm:
to better understand the inconsitent hash problem that we currently have. @bodymindarts

### DIFF
--- a/common/src/test/java/bisq/HashMapOrderTest.java
+++ b/common/src/test/java/bisq/HashMapOrderTest.java
@@ -1,6 +1,6 @@
 package bisq;
 
-import java.util.Map;
+import java.util.TreeMap;
 
 import org.junit.Test;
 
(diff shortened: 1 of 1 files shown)


bodymindarts:
Nack. Can we test each case against a hard coded `byte[]`? Then we will see if a JVM update or something like that changes what we expect.

christophsturm:
well that test just verifies what we talked about, and you can always add another test that uses a byte array.

bodymindarts:
Afaict internally `LinkedHashMap` is [being used](https://github.com/protocolbuffers/protobuf/blob/f9d8138376765d229a32635c9209061e4e4aed8c/java/core/src/main/java/com/google/protobuf/MapField.java#L171) which actually preserves order and would thus make this test fail. In other words the behaviour is different depending on if we create it then serialize it (all key-orders lead to same bytes), or de-serialize it from the wire (due to `LinkedHashMap` internally it becomes order sensitive).

Its really funny that the whole of Bisq only works due to an implementation detail of the Java pb library (meaning hashes are reproducible on different nodes). Even in the reference implementation of protobuf (golang) a non-order-preserving map type is being used.

bodymindarts:
I don't really see how this test is useful. What would be actionable if it breaks?
In order for bisq to remain stable it is not really important that insert order is deterministic. Its much more important that order remains stable during the deserialize -> reserialize process. If _that_ breaks down bisq is completely broken.

christophsturm:
this test does not really test any of our own code. its just an investigation and documentation on how protobuf behaves with maps.

ripcurlx:
@christophsturm Is this intended for discussion purpose only?

christophsturm:
it was the result of a discussion but it could also be useful as documentation,.

PR #3392, opened 2019-10-11
Account Signing: Fix verified usage

sqrrm:


desktop/src/main/java/bisq/desktop/main/account/content/PaymentAccountsView.java
```diff
@@ -130,6 +130,8 @@ public void updateItem(final PaymentAccount item, boolean empty) {
                                 switch (signState) {
                                     case PEER_SIGNER:
                                     case ARBITRATOR:
+                                    case PEER_LIMIT_LIFTED:
+                                    case PEER_INITIAL:
```
ripcurlx: As discussed we only show the approval icon for signer accounts. Every other state should indicate to the user that she should trade with caution.

ripcurlx (changes requested):
NACK - please see my comment. Everything else utACK

desktop/src/main/java/bisq/desktop/main/account/content/PaymentAccountsView.java
sqrrm: Reverted

ripcurlx (approved):

PR #3410, opened 2019-10-15
NTP: Fix a couple of UI issues in the New Trade Protocol

ripcurlx:


core/src/main/java/bisq/core/presentation/SupportTicketsPresentation.java
```diff
@@ -42,6 +43,11 @@
     @Getter
     private final BooleanProperty showOpenMediationTicketsNotification = new SimpleBooleanProperty();
 
+    @Getter
+    private final StringProperty numOpenRefundTickets = new SimpleStringProperty();
+    @Getter
+    private final BooleanProperty showOpenRefundTicketsNotification = new SimpleBooleanProperty();
```
lusarz, line 49: Are `numOpenRefundTickets` and `showOpenRefundTicketsNotification` variables used somewhere outside of `SupportTicketsPresentation` ?

sqrrm (approved):
utACK

core/src/main/java/bisq/core/presentation/SupportTicketsPresentation.java
sqrrm: Not that I can see, just like the other properties.

ripcurlx: > Are `numOpenRefundTickets` and `showOpenRefundTicketsNotification` variables used somewhere outside of `SupportTicketsPresentation` ?

But yes, the getters for all properties are not necessary anymore (there was a use case in the past). I'll change it in one of my next PRs

lusarz: So why we need these properties ? Seems like we only set values, but never use them. Or I miss something ?

ripcurlx: > So why we need these properties ? Seems like we only set values, but never use them. Or I miss something ?

I changed it already locally where all except of ~one~ two properties (numOpenSupportTickets, showOpenSupportTicketsNotification) are removed

chimp1984:
@ripcurlx @sqrrm  It seems support for RefundManager is missing. ArbitrationManager is for the current arbitrator, but RefundManager fro the NTP version of arbitration.

PR #3413, opened 2019-10-15
Apply rule to not allow BSQ outputs after BTC output for regular txs

chimp1984:
With block 602500 (about 2 weeks after v1.2.0 release) we enforce a new rule which represents a
hard fork. Not updated nodes would see an out of sync dao state hash if a relevant transaction would
happen again.
Further (highly unlikely) consequences could be:
If the BSQ output would be sent to a BSQ address the old client would accept that even it is
invalid according to the new rules. But sending such a output would require a manually crafted tx
(not possible in the UI). Worst case a not updated user would buy invalid BSQ but that is not possible as we
enforce update to 1.2.0 for trading a few days after release as that release introduced the new trade protocol
and protection tool. Only of both both traders would have deactivated filter messages they could trade.

Problem description:
We did not apply the check to not allow BSQ outputs after we had detected a BTC output.
The supported BSQ transactions did not support such cases anyway but we missed an edge case:
A trade fee tx in case when the BTC input matches exactly the BTC output
(or BTC change was <= the miner fee) and the BSQ fee was > the miner fee. Then we
create a change output after the BTC output (using an address from the BTC wallet) and as
available BSQ was >= as spent BSQ it was considered a valid BSQ output.
There have been observed 5 such transactions where 4 got spent later to a BTC address and by that burned
the pending BSQ (spending amount was higher than sending amount). One was still unspent.
The BSQ was sitting in the BTC wallet so not even visible as BSQ to the user.
If the user would have crafted a custom BSQ tx he could have avoided that the full trade fee was burned.

Not an universal rule:
We cannot enforce the rule that no BSQ output is permitted to all possible transactions because there can be cases
where we need to permit this case.
For instance in case we confiscate a lockupTx we have usually 2 BSQ outputs: The first one is the bond which
should be confiscated and the second one is the BSQ change output.
At confiscating we set the first to TxOutputType.BTC_OUTPUT but we do not want to confiscate
the second BSQ change output as well. So we do not apply the rule that no BSQ is allowed once a BTC output is
found. Theoretically other transactions could be confiscated as well and all BSQ tx which allow > 1 BSQ outputs
would have the same issue as well if the first output gets confiscated.
We also don't enforce the rule for irregular or invalid txs which are usually set and detected at the end of
the tx parsing which is done in the TxParser. Blind vote and LockupTx with invalid OpReturn would be such cases
where we don't want to invalidate the change output (See comments in TxParser).

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
@@ -573,16 +573,21 @@ public class BsqWalletService extends WalletService implements DaoStateListener 
 
 
     ///////////////////////////////////////////////////////////////////////////////////////////
-    // Burn fee tx
+    // Burn fee txs
     ///////////////////////////////////////////////////////////////////////////////////////////
 
     // We create a tx with Bsq inputs for the fee and optional BSQ change output.
     // As the fee amount will be missing in the output those BSQ fees are burned.
-    public Transaction getPreparedProposalTx(Coin fee, boolean requireChangeOutput) throws InsufficientBsqException {
-        return getPreparedBurnFeeTx(fee, requireChangeOutput);
+    public Transaction getPreparedProposalTx(Coin fee) throws InsufficientBsqException {
+        return getPreparedBurnBsqFeeTx(fee);
     }
 
-    public Transaction getPreparedBurnFeeTx(Coin fee) throws InsufficientBsqException {
+    //todo
+    public Transaction getPreparedIssuanceTx(Coin fee) throws InsufficientBsqException {
+        return getPreparedBurnBsqFeeTx(fee);
+    }
+
+    public Transaction getPreparedTradeFeeTx(Coin fee) throws InsufficientBsqException {
         return getPreparedBurnFeeTx(fee, false);
     }
 
@@ -633,7 +638,7 @@ public class BsqWalletService extends WalletService implements DaoStateListener 
                 log.warn("We increased required input as change output was zero or dust: New change value={}", change);
                 String info = "Available BSQ balance=" + coinSelection.valueGathered.value / 100 + " BSQ. " +
                         "Intended fee to burn=" + fee.value / 100 + " BSQ. " +
-                        "Please reduce the fee to burn to " + (coinSelection.valueGathered.value - minDustThreshold.value) / 100 + " BSQ.";
+                        "Please increase your balance to at least " + (coinSelection.valueGathered.value + minDustThreshold.value) / 100 + " BSQ.";
                 checkArgument(coinSelection.valueGathered.compareTo(fee) > 0,
                         "This transaction require a change output of at least " + minDustThreshold.value / 100 + " BSQ (dust limit). " +
                                 info);
@@ -655,6 +660,7 @@ public class BsqWalletService extends WalletService implements DaoStateListener 
         }
     }
 
+    //todo
     private Transaction getPreparedBurnFeeTx(Coin fee, boolean requireChangeOutput) throws InsufficientBsqException {
         daoKillSwitch.assertDaoIsNotDisabled();
         final Transaction tx = new Transaction(params);
--- a/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
@@ -586,6 +586,68 @@ public class BsqWalletService extends WalletService implements DaoStateListener 
         return getPreparedBurnFeeTx(fee, false);
     }
 
+    public Transaction getPreparedBurnFeeTxForAssetListing(Coin fee) throws InsufficientBsqException {
+        daoKillSwitch.assertDaoIsNotDisabled();
+
+        // We need to require one BSQ change output as we could otherwise not be able to distinguish between 2
+        // structurally same transactions where only the BSQ fee is different and the asset listing fee is
+        // a user input when creating the asset listing, so it is not know to the parser.
+        // Instead we derive the asset listing fee from the parser.
+
+        // Case 1: 10 BSQ asset listing fee
+        // In: 15 BSQ
+        // Out: BSQ change 5 BSQ -> valid BSQ
+        // Out: OpReturn
+        // Miner fee: 1000 sat  (10 BSQ burned)
+
+
+        // Case 2: 15 BSQ asset listing fee
+        // In: 15 BSQ
+        // Out: burned BSQ change 5 BSQ -> BTC (5 BSQ burned)
+        // Out: OpReturn
+        // Miner fee: 1000 sat  (10 BSQ burned)
+
+        Transaction tx = new Transaction(params);
+        // We look for inputs covering out BSQ fee we want to pay.
+        CoinSelection coinSelection = bsqCoinSelector.select(fee, wallet.calculateAllSpendCandidates());
+        try {
+            Coin change = bsqCoinSelector.getChange(fee, coinSelection);
+            if (change.isZero() || Restrictions.isDust(change)) {
+                // If change is zero or below dust we increase required input amount to enforce a BSQ change output.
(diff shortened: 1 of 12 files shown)


core/src/main/java/bisq/core/dao/node/parser/TxOutputParser.java
```diff
@@ -118,19 +170,39 @@ void processTxOutput(TempTxOutput tempTxOutput) {
...
+                // After the hard fork activation we fix a bug with a transaction which would have interpreted the
+                // vote fee output as BSQ if the vote fee was >= miner fee.
+                // Such a tx was never created but as we don't know if it will happen before activation date we cannot
+                // enforce the bug fix which represents a rule change before the activation date.
+                handleIssuanceCandidateOutput(tempTxOutput);
+            } else if (isHardForkActivated(tempTxOutput) && isBlindVoteFeeOutput(tempTxOutput)) {
+                // After the hard fork activation we fix a bug with a transaction which would have interpreted the
+                // issuance output as BSQ if the availableInputValue was >= issuance amount.
+                // Such a tx was never created but as we don't know if it will happen before activation date we cannot
+                // enforce the bug fix which represents a rule change before the activation date.
```
sqrrm: Comments are switched, vote fee comment for issuance and vice versa

core/src/main/java/bisq/core/dao/node/parser/TxOutputParser.java
```diff
@@ -171,17 +244,118 @@ private void handleUnlockBondTx(TempTxOutput txOutput) {
...
+            return false;
+        }
+
+        // If it is the vote stake output we return false.
+        if (tempTxOutput.getIndex() == 0) {
+            return false;
+        }
+
+        // There must be a vote fee left
+        if (availableInputValue <= 0) {
```
sqrrm: I think there is always a vote fee left here if any fee was paid at all

core/src/main/java/bisq/core/dao/node/parser/TxOutputParser.java
```diff
@@ -171,17 +244,118 @@ private void handleUnlockBondTx(TempTxOutput txOutput) {
...
+        // outputs are parsed and check the phase. The TxParser will do that....
+        optionalIssuanceCandidate = Optional.of(tempTxOutput);
+    }
+
+    private void handleBlindVoteFeeOutput(TempTxOutput tempTxOutput) {
+        tempTxOutput.setTxOutputType(TxOutputType.BTC_OUTPUT);
+        prohibitMoreBsqOutputs = true;
+    }
+
+    private boolean isBlindVoteFeeOutput(TempTxOutput tempTxOutput) {
```
sqrrm: I don't understand the naming of this method

core/src/main/java/bisq/core/dao/node/parser/TxOutputParser.java
```diff
@@ -171,17 +244,118 @@ private void handleUnlockBondTx(TempTxOutput txOutput) {
...
+        // be detected at blindVoteFee check.
+        // We always have the BSQ change before the burned BSQ output if both are present.
+        checkArgument(optionalOpReturnIndex.isPresent());
+        if (tempTxOutput.getIndex() != optionalOpReturnIndex.get() - 1) {
+            return false;
+        }
+
+        // Without checking the fee we would not be able to distinguish between 2 structurally same transactions, one
+        // where the output is burned BSQ and one where it is a BSQ change output.
+        long blindVoteFee = daoStateService.getParamValueAsCoin(Param.BLIND_VOTE_FEE, tempTxOutput.getBlockHeight()).value;
```
sqrrm: I feel it would be better to disallow any other BSQ outputs than the vote stake output for blind vote txs. I think the code would be easier to read and I don't see any other use case.

core/src/main/java/bisq/core/dao/node/parser/TxOutputParser.java
chimp1984: Will fix

core/src/main/java/bisq/core/dao/node/parser/TxOutputParser.java
chimp1984: Is isBlindVoteBurnedFeeOutput more clear? It it the output of the burned BSQ

core/src/main/java/bisq/core/dao/node/parser/TxOutputParser.java
chimp1984: There can be an optional BSQ change output.

Inputs [1-n]: BSQ inputs for BSQ fee + stake
Inputs [0-n]: BTC inputs for miner fee
Output [1]: Mandatory BSQ output of stake
Output [0-1] Optional BSQ change output
Outputs [0-1]: BTC change output OR optional burned BSQ as BTC output
Outputs [0-1]: Output [1]: OP_RETURN with OpReturnData and amount 0
Mining fee: BTC mining fee + burned BSQ fee

chimp1984: Without knowing the fee you cannot distinguish between a BSQ change and burned BSQ

core/src/main/java/bisq/core/dao/node/parser/TxOutputParser.java
chimp1984: It can be a BTC change output. We don't cover that in this code branch but later.

ManfredKarrer (approved):
ACK

ripcurlx (approved):
utACK

core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
```diff
@@ -573,60 +582,128 @@ private Transaction getPreparedSendTx(String receiverAddress, Coin receiverAmoun
...
+    private Transaction getPreparedTxWithMandatoryBsqChangeOutput(Coin fee) throws InsufficientBsqException {
         daoKillSwitch.assertDaoIsNotDisabled();
-        final Transaction tx = new Transaction(params);
-        addInputsAndChangeOutputForTx(tx, fee, bsqCoinSelector, requireChangeOutput);
-        // printTx("getPreparedFeeTx", tx);
-        return tx;
+
+        Transaction tx = new Transaction(params);
+        // We look for inputs covering out BSQ fee we want to pay.
+        CoinSelection coinSelection = bsqCoinSelector.select(fee, wallet.calculateAllSpendCandidates());
```
sqrrm, line 643: It would be better to include the dust here in the initial coin selection since it is a requirement that it's included. Then there is no need to do a double take when this initial selection wasn't enough.

core/src/main/java/bisq/core/dao/node/parser/TxOutputParser.java
```diff
@@ -171,17 +237,108 @@ private void handleUnlockBondTx(TempTxOutput txOutput) {
...
+                        // We always have the BSQ change before the burned BSQ output if both are present.
+                        checkArgument(optionalOpReturnIndex.isPresent());
+                        if (index != optionalOpReturnIndex.get() - 1) {
+                            return false;
+                        }
+
+                        // Without checking the fee we would not be able to distinguish between 2 structurally same transactions, one
+                        // where the output is burned BSQ and one where it is a BSQ change output.
+                        long blindVoteFee = daoStateService.getParamValueAsCoin(Param.BLIND_VOTE_FEE, tempTxOutput.getBlockHeight()).value;
+                        return availableInputValue == blindVoteFee;
```
sqrrm, line 293: What happens with a vote that's using the wrong fee. It would set this output to BSQ, but would it fail to register as a valid vote?

sqrrm: It's set to IRREGULAR in the TxParser so that's fine.

core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
chimp1984: Not clear to me what you mean exactly?

PR #3415, opened 2019-10-16
Account Signing: Don't sign filtered accounts

sqrrm:
I don't know how to add banned accounts so please test that if you know how to.


ripcurlx (approved):
utACK

PR #3420, opened 2019-10-17
NTP: merge with master

ripcurlx:


ripcurlx:
^^ This Travis JDK 12 problem every now and then slowly gets annoying.

Issue #2457, opened 2019-02-22
Show more details about connected peers within Settings > Network Info


devinbileck:
It would be nice to see more details about connected peers, for both Bitcoin and P2P networks, within Settings > Network Info.

For example, Bitcoin network peers only lists their addresses. It would be beneficial to list their version and possibly block height. These details are visible within the log file:
`Feb-22 00:36:13.388 [BlockingClient network thread for lva54pnbq2nsmjyr.onion:8333] INFO  o.b.core.Peer: [lva54pnbq2nsmjyr.onion]:8333: Got version=70015, subVer='/Satoshi:0.17.1/', services=0x1037, time=2019-02-22 00:36:09, blocks=564134`

Similarly for the P2P network, it would be nice to show the application version that each peer is running.

Additionally, it would be nice to be able to determine a peers version prior to taking an offer from them. For example, if you wanted to only trade with peers running the latest version to avoid potential issues. However, that doesn't prevent a peer with an older version taking an offer from you, unless we added that capability.

Ultimately, with this information, we could potentially allow a user to decide to avoid/exclude/block certain peers that they do not want to connect/interact with (e.g. if the peer is not up to date).

ManfredKarrer:
Feel free to add the info t0 the UI. Version of seeds is probably not propagated atm but as seeds usually run the lastest version it might be less relevant.

For the other suggestion regarding info of version of peer, I am not so sure as it adds complexity for users. They have to understand some context to make here choices and I think we should removed instead of adding complexity / work load for user. Beside that I think a big part of ur users are updating quickly, so I don't see much need to pushing that harder. Delivering useful features in each release might be the best way to softly convince them to update.

devinbileck:
Note, using [BitcoinJ 0.15](https://github.com/bisq-network/bisq/pull/2772) the BTC nodes now show the additional details:
![image](https://user-images.githubusercontent.com/603793/57988674-f1d34e00-7a45-11e9-9385-ac27c32d9210.png)

freimair:
seems like that issue became deprecated

PR #3423, opened 2019-10-17
Show connected Bitcoin network peer info

devinbileck:
Before:
image

After:
image

This partially addresses #2457, perhaps we can add additional P2P network peer details in the future.


ripcurlx (approved):
utACK

PR #3425, opened 2019-10-18
Not show payment account details for blocked offers

ripcurlx:
Atm blocking offers that abuse the general information section of the payment method is useless,
as it does exactly what the offer maker wants. Show the information and prevent the offer from being taken.


sqrrm (approved):
utACK

PR #3434, opened 2019-10-19
Refactor BisqEnvironment - change type of ignoreLocalBtcNode to boolean

lusarz:


freimair:
can we have a 
```
  --ignoreLocalBtcNode
        If set, a Bitcoin core node running locally will be ignored
```

instead of 

```
  --ignoreLocalBtcNode=<value>
        If set to true a Bitcoin core node running locally will be ignored
```
while you are at it?

freimair (changes requested):
can we have a 
```
  --ignoreLocalBtcNode
        If set, a Bitcoin core node running locally will be ignored
```

instead of 

```
  --ignoreLocalBtcNode=<value>
        If set to true a Bitcoin core node running locally will be ignored
```
while you are at it?

lusarz:
Conflicts resolved.

I think we can do `--ignoreLocalBtcNode` instead of `--ignoreLocalBtcNode=<value>`, but we should consider situation when someone pass for example `--ignoreLocalBtcNode=false`. I propose to create dedicated issue for that and discuss it a bit there.

freimair (approved):
agreed.

Ack.

PR #3435, opened 2019-10-21
Update account signing strings for v1.2 release

m52go:
I'm making a couple of in-line comments -- please see them.

EDIT

Cannot comment on unchanged code...I'll just add comments here.

---

```
offerbook.timeSinceSigning.info.peerLimitLifted=signed by a peer and limits were lifted
offerbook.timeSinceSigning.info.signer=signed by peer and can sign peer accounts
```

Same thing now?

---

Changed `popup.warning.tradeLimitDueAccountAgeRestriction` to fit new rules (since March 1 2019 date is no longer the determining factor for signing).

---

Changed single quotes to double quotes in `portfolio.pending.mediationResult.popup.info`...seems like mistake since variable includes {n} insertions.

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -343,8 +343,8 @@ offerbook.timeSinceSigning.info.signer=signed by peer and can sign peer accounts
 offerbook.timeSinceSigning.daysSinceSigning={0} days
 offerbook.timeSinceSigning.daysSinceSigning.long={0} since signing
 
-offerbook.timeSinceSigning.help=By trading with a payment account that was verified by an arbitrator or a peer, your account gets signed as well.\n\
-   {0} days later the initial limit of {1} gets lifted and your account can sign other peers as well.
+offerbook.timeSinceSigning.help=When you successfully complete a trade with a peer who has a signed payment account, your payment account is signed.\n\
+   {0} days later, the initial limit of {1} is lifted and your account can sign other peers'' payment accounts.
 offerbook.timeSinceSigning.notSigned=Not signed yet
 offerbook.timeSinceSigning.notSigned.noNeed=Unsigned
 shared.notSigned=This account hasn't been signed yet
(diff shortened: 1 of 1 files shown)


ripcurlx:
> Same thing now?

The `offerbook.timeSinceSigning.info.peerLimitLifted` won't be displayed right now, but I'd like to keep it there if we decide to change this again later. Still I think we should extend the `offerbook.timeSinceSigning.info.signer` value with the limits lifted state. What do you think?

ripcurlx:
e.g. `offerbook.timeSinceSigning.info.signer=signed by peer and can sign peer accounts (limits lifted)`?

ripcurlx:
Or we leave this info out, as it might not be of any help for the other trading peer?

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -376,15 +376,17 @@ offerbook.warning.noMatchingAccount.msg=To take this offer, you will need to set
...
- By trading with a peer that has already been verified your account will be signed after a successful trade as well and the limits will be lifted after a certain time frame based on the verification method.\n\n\
-  For more information on account signing, please visit our documentation section on docs.bisq.network/account-signing.
+offerbook.warning.newVersionAnnouncement=With this version of the software, trading peers can verify and sign each others' payment accounts to create a network of trusted payment accounts.\n\n\
+ After successfully trading with a peer with a verified payment account, your payment account will be signed and trading limits will be lifted after a certain time interval (length of this interval is based on the verification method).\n\n\
+  For more information on account signing, please see the documentation at https://docs.bisq.network/payment-methods#account-signing.
 
 popup.warning.tradeLimitDueAccountAgeRestriction.seller=The allowed trade amount is limited to {0} because of security restrictions based on the following criteria:\n\
- - The buyers account was created after March 1st 2019\n\
+ - The buyer''s account is not at least 60 days old
```
ripcurlx: You are missing a `\n\` at the end of the line

ripcurlx:
> Changed `popup.warning.tradeLimitDueAccountAgeRestriction` to fit new rules (since March 1 2019 date is no longer the determining factor for signing).

I think this change is not correct. The criteria should be something like this:
- The buyer''s account has not been signed by an arbitrator or a peer
- The time since signing of the buyer''s account is not at least 30 days 
- The payment method for this offer is considered risky for bank chargebacks

m52go:
> Or we leave this info out, as it might not be of any help for the other trading peer?

I added the change -- I think it's better to be as clear as possible.

> I think this change is not correct. The criteria should be something like this:

Makes sense, changed. I was under the impression that this string describes why payment accounts were not signed by an arbitrator at first, not why they aren't signed in general.

ripcurlx:
> Makes sense, changed. I was under the impression that this string describes why payment accounts were not signed by an arbitrator at first, not why they aren't signed in general.

It is triggered whenever the seller's or buyer's account has the 0.01 BTC limit. In the past that was for all accounts after 1st of March 2019. Good that you found that this translation needed to be changed as well 👍

ripcurlx (approved):
utACK

PR #3445, opened 2019-10-23
Use correct formatter and add missing value for placeholder

### LINKED ISSUE #3444: [1.2.0] Informational prompt explaining trade amount limit displays the limit in BSQ rather than BTC
When attempting to enter a value greater than the trade limit for the amount of BTC to buy/sell, the informational prompt displayed the limit in BSQ rather than BTC.
image

ripcurlx:
Fixes #3444.


devinbileck (approved):
utACK

PR #3448, opened 2019-10-23
Arbitrator republish signedWitnesses on startup

sqrrm:
This republishes all signed witnesses 60 seconds after boostrap complete for arbitrators

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqSetup.java
+++ b/core/src/main/java/bisq/core/app/BisqSetup.java
@@ -692,6 +692,9 @@ public class BisqSetup {
 
         accountAgeWitnessService.onAllServicesInitialized();
         signedWitnessService.onAllServicesInitialized();
+        if (user.getRegisteredArbitrator() != null) {
+            signedWitnessService.setRepublishOnStartup(true);
+        }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/bisq/core/app/BisqSetup.java
```diff
@@ -692,6 +692,9 @@ private void initDomainServices() {
 
         accountAgeWitnessService.onAllServicesInitialized();
         signedWitnessService.onAllServicesInitialized();
+        if (user.getRegisteredArbitrator() != null) {
```
chimp1984: Why not do it inside of signedWitnessService? User can be injected. Better to not spread out responsibilities...

chimp1984 (changes requested):

core/src/main/java/bisq/core/app/BisqSetup.java
sqrrm: Yes, perhaps that's neater. Fixed

chimp1984 (commented):
utACK

ripcurlx (approved):
utACK

Issue #3463, opened 2019-10-24
[1.2.0] Time locked payout is paid without initiating support (arbitration)


devinbileck:
### Description

When taking an offer and not initiating/confirming payment or opening a support ticket, when I generate blocks until exceeding the timelock period nothing happens (as expected). But once either client restarts (Alice or Bob), it pays out to the donation address immediately upon restart. As far as I am aware, the payout should only occur if any trader opens an arbitration ticket (refund agent).

#### Version

release/v1.2.0 branch (commit cd4cd39206c22b29db77a6b082a00127605416c4)

### Steps to reproduce

Preconditions:
- Start with a fresh regtest environment (using the dao-setup.zip file)
- Adjust the delay in MakerSetsLockTime to only 5 blocks to make testing easier
- Change RECIPIENT_BTC_ADDRESS to an address of your local regtest Bitcoin Core node

Steps:
1. Alice creates a buy offer and Bob takes the offer (do not initiate payment).
2. Generate blocks until exceeding the timelock delay period.
3. Restart Bob's client.

### Expected behaviour

The time locked payout should NOT be paid.

### Actual behaviour

The time locked payout is paid to the RECIPIENT_BTC_ADDRESS as soon as the client restarts.

#### Device or machine

Windows 10

#### Additional info

There are no errors in the log, and I don't see any indication that publishDelayedPayoutTx was called. However, I do see the transaction broadcast:
```
Oct-24 14:56:08.150 [ STARTING] INFO  org.bitcoinj.wallet.Wallet: New broadcaster so uploading waiting tx 93e1c52388f235136d213a5f4204c1508e48a9553d392e6b49ff792ce074d6fa 
Oct-24 14:56:08.155 [ STARTING] INFO  o.b.core.TransactionBroadcast: Waiting for 1 peers required for broadcast, we have 0 ... 
```
```
Oct-24 14:56:10.349 [JavaFX Application Thread] INFO  o.b.core.TransactionBroadcast: broadcastTransaction: We have 1 peers, adding 93e1c52388f235136d213a5f4204c1508e48a9553d392e6b49ff792ce074d6fa to the memory pool 
Oct-24 14:56:10.353 [JavaFX Application Thread] INFO  o.b.core.TransactionBroadcast: Sending to 1 peers, will wait for 0, sending to: [127.0.0.1]:18444 
```

devinbileck:
[...]

PR #3464, opened 2019-10-25
Do not commit delayedPayoutTx to avoid publishing at restart

chimp1984:
Fixes https://github.com/bisq-network/bisq/issues/3463

BitcoinJ publishes automatically committed transactions.
We committed it to the wallet to be able to access it later after a
restart. We stored the txId in Trade and used that to request the tx
from the wallet (as it was committed). Now we store the
bitcoin serialized bytes of the tx and do not commit the tx before
broadcasting it (if a trader opens refund agent ticket).


ripcurlx (approved):
utACK

devinbileck:
I tested this and the original issue has been resolved.
However, I ran into a couple other issues that I am not sure if related, but I will enter separately.

Issue #3426, opened 2019-10-18
Create new intro pop-up for v1.2


m52go:
To replace the current DAO pop-up that shows when new users open Bisq.

I think it's critical that users see **both** announcements—account signing and new trade protocol. If we put them on different panels, it's likely they'll miss at least one of them.

This is also why I've removed the "Dismiss" button from the DAO pop-up and replaced it with a small X in the top-right. Would be great if the pop-up persists until the X is clicked, so that the user can click on both "Learn More" buttons before the box disappears.

The "Learn More" links will lead to documentation, which will hopefully include video descriptions of the new mechanisms too.

@pedromvpg could use your feedback here! cc @ripcurlx.

_Proposed pop-up_

![v12-intro-mockup](https://user-images.githubusercontent.com/735155/67089382-718c5280-f175-11e9-9c0d-46a0d4598fe2.png)

_Old DAO pop-up (for reference)_

![Screenshot from 2019-10-18 06-32-47](https://user-images.githubusercontent.com/735155/67089344-59b4ce80-f175-11e9-8e29-fea57b0d3de0.png)

ripcurlx:
@m52go I'll do my best to layout as good as possible.
@pedromvpg If you have any suggestions on the exact layout feel free to provide one here within EOD.

m52go:
@ripcurlx sounds good. For links:

* Account Signing button can go to https://docs.bisq.network/payment-methods#account-signing.
* New Trade Protocol button can go to https://docs.bisq.network/trading-rules.

ripcurlx:
How about this: 
![Bildschirmfoto 2019-10-21 um 15 39 12](https://user-images.githubusercontent.com/170962/67210221-fea30600-f418-11e9-8a60-b5e241af6883.png)

I kept the cancel icon in green as we don't have action buttons right now that are black.

pedromvpg:
Sorry, only seeing this now.
I think it looks good but I can do a couple of adjustments tonight.

m52go:
Cool. BTW images may be improved too...the ones I used are workable but meant to be placeholders.

pedromvpg:
I think the images are ok.
Here's my proposal with slightly different typographic approach. 


![Bisq Critical Updates](https://user-images.githubusercontent.com/807505/67251698-e2fe2680-f43d-11e9-8158-a26331e9c901.png)

pedromvpg:
[...]

PR #3465, opened 2019-10-25
[1.2.0] Improve new feature popup

ripcurlx:
Based on feedback in https://github.com/bisq-network/bisq/issues/3426#issuecomment-546138227

Bildschirmfoto 2019-10-25 um 10 09 04


desktop/src/main/java/bisq/desktop/main/MainViewModel.java
```diff
@@ -621,14 +621,14 @@ public BooleanProperty getShowAccountUpdatesNotification() {
 
     private void maybeAddNewTradeProtocolLaunchWindowToQueue() {
         String newTradeProtocolWithAccountSigningLaunchPopupKey = "newTradeProtocolWithAccountSigningLaunchPopup";
-        if (DontShowAgainLookup.showAgain(newTradeProtocolWithAccountSigningLaunchPopupKey)) {
+//        if (DontShowAgainLookup.showAgain(newTradeProtocolWithAccountSigningLaunchPopupKey)) {
```
sqrrm: I don't think this should be commented out

sqrrm (changes requested):

PR #3471, opened 2019-10-25
Set amount for delayed payout tx to 0

chimp1984:
We have shown the spent funds from the deposit tx to the bisq donation
address before. But that was incorrect from the wallet perspective and
would have lead to incorrect summary of all transaction amounts. We set
it now to 0 as we are not spending funds nor receiving any in our wallet.


sqrrm (approved):
utACK

Issue #3470, opened 2019-10-25
Clean up trade statistics from duplicate entries


chimp1984:
Currently we have some duplicate entries (see https://github.com/bisq-network/bisq/issues/3468) in the trade statistic map because the hash has changed with changes in the extraMap field and we added new enties in recent versions. As there was a republish routine for past trade statistics we added duplicates with such updates.
To fix that we should add some cleanup code which deletes those duplicates. We can check for trade ID and the most significant fields like price and amount. 
The republish routine was removed for v.1.2. so there should not be added new duplicates.
For future updates when we add entries to that map it would only be an issue if maker and taker have a different version. 
If the cleanup code does not cause performance issues we could leave it so it would handle such potential future scenarios as well. Alternatively we could add a Json Ignore annotation to the map as well.

mrosseel:
is there a way to easily unit test this so new discrepancies can be detected early?

chimp1984:
The intention of the extra data hashmap is to be flexible to add data without breaking backward compatibility. So if anyone adds a new entry there it will break the hash. Best is probably to exclude the extra map from the hash by json exclude.
A unit test is trivial here but will not help you discover if any new feature adds a new entry.

chimp1984:
@bodymindarts Can you post your code how you observed those duplicates?
I just tried to find duplicates in the TradeStatistics2Store but could not see any using the offerID, amount and price as key.

chimp1984:
I am working on a fix....

PR #3476, opened 2019-10-26
Clean up trade statistics from duplicate entries

chimp1984:
Fixes https://github.com/bisq-network/bisq/issues/3470

At software updates we added new entries to the extraMap which caused
duplicate entries (if one if the traders was on the new and the other on
the old version or at republishing). We set it now json exclude so avoid
that in future and clean up the map.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
+++ b/p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
@@ -58,6 +58,8 @@ import org.jetbrains.annotations.Nullable;
 @Slf4j
 class RequestDataHandler implements MessageListener {
     private static final long TIMEOUT = 90;
+    private static boolean initialRequestApplied = false;
+
     private NodeAddress peersNodeAddress;
     /*
      */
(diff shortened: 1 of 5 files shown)


sqrrm (approved):
utACK

p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
```diff
@@ -240,7 +242,12 @@ public void onMessage(NetworkEnvelope networkEnvelope, Connection connection) {
                                     // Processing 82645 items took now 61 ms compared to earlier version where it took ages (> 2min).
                                     // Usually we only get about a few hundred or max. a few 1000 items. 82645 is all
                                     // trade stats stats and all account age witness data.
-                                    dataStorage.addPersistableNetworkPayloadFromInitialRequest(e);
+
+                                    // We only apply it once from first response
+                                    if (!initialRequestApplied) {
```
julianknutsen, line 247: Why do we only handle the first requests for the PersistableNetworkPayload objects, but all of them for the ProtectedStoragePayload objects? Do we expect more inconsistencies between the ProtectedStorageEntrys so we use all of them? Or, were the PersistableNetworkPayload handlers not dealing with duplicates well?

I haven't been able to follow the entire TradeStatstics issue, but I am just trying to understand more about what is expected/unexpected in this startup path.

chimp1984: The hash issue was caused by the extraMap where at certain updates data got added and combined with the republishing of the objects created dupicates entries as the hashes have been different due the added entries in that map (was used for tracking usage of arbitrators and mediators for a optimized selection of them). 

The optimisation for handling the PersistableNetworkPayload objects was to avoid performance spikes when processing a large set of PersistableNetworkPayload objects. For ProtectedStorageEntrys we do not expect such high amount of objects.  PersistableNetworkPayload are mainly Tradestatistic objects and AccountAgeWitnes objects and if a user have not been online for longer time those can be several 1000 objects which caused longer processing time if the normal method was used where unnecessary checks (as we got those obejcts at startup and not at runtime later) have caused most of the cpu costs.

julianknutsen: Ok, I think I understand. So we can have up to two loops through the ProtectedStoragePayload path and only one through the PersistableNetworkPayload path. The code handles duplicates just fine, but processing 2x the PersistableNetworkPayload objects takes a lot of time and we don't expect processing them from the second seed to provide meaningful data. Thanks for the rundown.

PR #3477, opened 2019-10-26
Create modal pop-up with single button taking user to Support screen if there are active support tickets

### LINKED ISSUE #3401: Notification of pending arbitration issue
The bubble notification with the number of pending arbitration issues does provide the information   required but does not convey any urgency and is a bit too unobtrusive.

I will suggest a pop-up alert/dialog once the app launches; and on clicking the **`OK`** button (which will be the only available button - no **`Cancel`** button) will navigate to the arbitration list screen.

Screenshot from 2019-10-13 23-11-57

niyid:
Resolves issue #3401

closes #3401


freimair (approved):
utAck

ripcurlx:
@niyid Please resolve conflict with displayStrings.properties.

niyid:
Will do. Apologies.

ripcurlx:
@niyid Please sign all your commits, before I start with the review. Thanks!

desktop/src/main/java/bisq/desktop/main/MainView.java
```diff
@@ -413,6 +414,7 @@ protected Tooltip computeValue() {
 
         // Delay a bit to give time for rendering the splash screen
         UserThread.execute(() -> onUiReadyHandler.run());
+        UserThread.execute(() -> checkOpenSupportTickets());
```
ripcurlx, line 417: This should be done more like the new trade protocol window like https://github.com/bisq-network/bisq/blob/95257b10f9741e7d3548c3630dca0b2b2168b366/desktop/src/main/java/bisq/desktop/main/MainViewModel.java#L271 in the MainViewModel. So it is shown after the Splash screen based on priority.

desktop/src/main/java/bisq/desktop/main/MainView.java
```diff
@@ -429,6 +431,18 @@ public void onCheckpointFail() {
...
+    
+    private void checkOpenSupportTickets() {
+		// Create modal pop-up with single button taking user to Support screen if there
+		// are active support tickets
+		if (model.getNumOpenSupportTickets() != null && Integer.parseInt(model.getNumOpenSupportTickets().get()) > 0) {
+			new Popup<>().attention(Res.get("portfolio.pending.support.popup.attention"))
+					.closeButtonTextWithGoTo("navigation.support")
+					.onClose(
+							() -> navigation.navigateTo(MainView.class, SupportView.class, ArbitrationClientView.class))
+					.show();
```
ripcurlx, line 443: I think we should add a don't show again option for power users that might get annoyed having it popup every time it opens as long they have a support case open.

ripcurlx (changes requested):
NACK - please see my comments

desktop/src/main/java/bisq/desktop/main/MainView.java
niyid: Thanks for the suggestions. I will update ASAP.

desktop/src/main/java/bisq/desktop/main/MainView.java
niyid: I will look into that. Thanks.

niyid: I assume by power user, you refer to an Arbitrator/DisputeAgent?

niyid: This PR has moved to:

https://github.com/bisq-network/bisq/compare/master...bisq-3401:master

https://github.com/bisq-network/bisq/pull/3477/commits/aae6fc7b6fd07152d1b475593dd9f14b96e6d037

PR #3478, opened 2019-10-26
Update RECIPIENT_BTC_ADDRESS for regtest

devinbileck:
Use an address that is owned by the regtest wallet in the dao-setup.zip
file. This allows for easily verifying BTC trading fees are sent to
this address correctly. Also, it helps verify spending of the time lock
payout.


sqrrm (approved):
utACK

Issue #3482, opened 2019-10-27
[1.2.0] Clicking on avatar icon of peers offer indicates payment account age is not known


devinbileck:
### Description

When clicking on the avatar icon of a peers offer, it indicates payment account age is not known.

#### Version

release/v1.2.0 (commit 8a000c100517fe5c2059742774b3a9d4ce1b22e3)

### Steps to reproduce

- Alice and Bob have created fiat accounts (e.g. National Bank transfer)
- Alice has published a buy offer
- Bob clicks on Alice's avatar icon for the offer

### Expected behaviour

The payment account age should indicate the age of the account (e.g. 1 day).

### Actual behaviour

The payment account age indicates it is not known.

### Screenshots

![image](https://user-images.githubusercontent.com/603793/67631436-bbe79080-f853-11e9-963d-fb379530213a.png)

![image](https://user-images.githubusercontent.com/603793/67631534-12a19a00-f855-11e9-88b7-9f24c332602b.png)

#### Device or machine

Windows 10

sqrrm:
Please check #3490 if any more details need fixing

devinbileck:
Looks good. Still a little room for improvement though:
1. hide account signing status for account types that don't use signing (#3484)
2. show account signing status above time since signing - seems logical to show the status first, and then the age
3. if account is not signed, hide time since signing - seems redundant and unnecessary to show "Not signed yet" and "Age not known"

devinbileck:
Verified. Closing.

PR #3490, opened 2019-10-27
Display appropriate account age info header

sqrrm:
Depending on charge back risk type, accounts should show
accountAgeWitness age or time since signing

Screenshots of popups with this fix for #3482 
Accounts:
image

First account popup:
image

Second account popup:
image

Third account popup:
image

Fourth account popup:
image


ripcurlx (approved):
utACK

Issue #3487, opened 2019-10-27
[1.2.0] Results of DAO voting cycle are not visible under governance > vote results until restarting the app


devinbileck:
### Description

Results of the DAO voting cycle are not visible under governance > vote results until restarting the app. It should be visible as soon as in the vote result phase.

#### Version

release/v.1.2.0 (commit 8a000c100517fe5c2059742774b3a9d4ce1b22e3)

### Steps to reproduce

Complete a DAO cycle including submitting a compensation request, voting, vote reveal, and vote results. Then view the vote results for the cycle under governance > vote results.

### Expected behaviour

The vote results for the cycle should be shown.

### Actual behaviour

No proposals or vote results are shown, even after progressing to the start of the following cycle. However, once you restart the app, the results are shown.

### Screenshots

Before restarting app:
![image](https://user-images.githubusercontent.com/603793/67631921-f2280e80-f859-11e9-9dfd-925681060d61.png)

After restarting app:
![image](https://user-images.githubusercontent.com/603793/67631962-6c589300-f85a-11e9-9918-0831e8a6906d.png)

#### Device or machine

Windows 10

chimp1984:
Thanks for reporting!
Fixed with https://github.com/bisq-network/bisq/pull/3492

devinbileck:
Verified. Closing.

PR #3492, opened 2019-10-28
[1.2.0] Check for result phase at activate method

chimp1984:
Fixes https://github.com/bisq-network/bisq/issues/3487

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/dao/governance/result/VoteResultView.java
+++ b/desktop/src/main/java/bisq/desktop/main/dao/governance/result/VoteResultView.java
@@ -202,6 +202,7 @@ public class VoteResultView extends ActivatableView<GridPane, Void> implements D
         cyclesTableView.getSelectionModel().selectedItemProperty().addListener(selectedVoteResultListItemListener);
 
         if (daoStateService.isParseBlockChainComplete()) {
+            checkForResultPhase(daoStateService.getChainHeight());
             fillCycleList();
         }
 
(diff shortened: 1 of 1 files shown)


desktop/src/main/java/bisq/desktop/main/dao/governance/result/VoteResultView.java
```diff
@@ -238,7 +239,11 @@ protected void deactivate() {
 
     @Override
     public void onParseBlockCompleteAfterBatchProcessing(Block block) {
-        int chainHeight = daoStateService.getChainHeight();
+        checkForResultPhase(daoStateService.getChainHeight());
+        fillCycleList();
+    }
+
+    private void checkForResultPhase(int chainHeight) {
         if (periodService.getFirstBlockOfPhase(chainHeight, DaoPhase.Phase.RESULT) == chainHeight) {
```
devinbileck, line 247: Shouldn't this be `<= chainHeight`?

devinbileck (changes requested):
NACK
The vote results are only populated if viewing the results on the first block of the results phase. If you try to view the results after that, they are still empty.

desktop/src/main/java/bisq/desktop/main/dao/governance/result/VoteResultView.java
sqrrm: `periodService.getFirstBlockOfPhase` gets the height of the first block of the current phase so this check will return true for all `chainHeight` in the `RESULT` phase.

sqrrm (approved):
NACK

ripcurlx:
@sqrrm I think you approved by mistake -> NACK

sqrrm:
I first approved without testing. Then I read a bit more, checked out and tested and it fails, working on a fix. Don't know how to undo the approve.

sqrrm:
I made a new PR at #3500 that fixes this

ripcurlx:
Superseded by #3500

PR #3494, opened 2019-10-28
[1.2.0] Update data stores and adding SignedWitnessStore

ripcurlx:


sqrrm:
I have run through this quite a few times now and I think this is fine. I get anything from 1055 to 1057 signed witnesses after bootstrap is done. After restart I always get 1057 so it's just a matter of the propagation of the two missing signed witnesses.

sqrrm (approved):
ACK

PR #3499, opened 2019-10-28
[1.2.0] Unsigned should be N/A

### LINKED ISSUE #3486: [1.2.0] Offers for account types that do not utilize account signing show confusing "unsigned" status
### Description

Offers for account types that do not utilize account signing are shown as "unsigned" in the offer list. I don't think this is the best term to use as it may be confused with "not signed yet". Perhaps it should show something like "N/A"?

#### Version

release/v1.2.0 (commit 8a000c100517fe5c2059742774b3a9d4ce1b22e3)

### Screenshots

image

ripcurlx:
Fixes #3486.


devinbileck (approved):
utACK

PR #3504, opened 2019-10-29
[1.2.0] Improve dispute section

ripcurlx:
Improve next steps for mediation and add it for refund agent cases as well.
When entering the support section automatically select the first case.


m52go:
utACK

m52go (approved):

PR #3506, opened 2019-10-29
chore: add btc node

kilrau:


wiz:
The btcnode operator role is typically done by regular or bonded Bisq contributors - is your first PR to Bisq?

ripcurlx:
> The btcnode operator role is typically done by regular or bonded Bisq contributors - is your first PR to Bisq?

As far as I know it's his first.

@kilrau Because of the potential privacy risk (bloom filter) if a BTC node operator harvests data from Bisq users we only allow this role for long time contributors. So any contribution on the codebase is highly appreciated first - Thanks!

kilrau:
Not taking this personally, since what you are saying makes sense as of bisq status quo.

Thinking a bit more about this, I want to object to the concept of `btc node operator` or a `federation of Bitcon Core Nodes` in general. According to [the docs](https://docs.bisq.network/btcnode.html), this role was introduced to
> (1) avoid relying on bloom filters to protect user privacy when connecting to arbitrary public nodes
> (2) ensure Bisq connects to nodes with consistent Bitcoin Core implementations.
> (3) in the case of a hard fork, it would be impossible for Bisq to tell which fork a connected node was running

IMHO all 3 reasons have failed/serious problems:

(1) A lot of unstructured thoughts. I believe it's worse for privacy to give my tx info to a handful of bisq operators copmared to spreading it across arbitrary public nodes. I believe it's worse to trust a handful of bisq nodes to give me correct chain info, which somewhat even know the context of what I'm doing than arbitrary public nodes (in fact that's the reason why my PR was closed - I am not trustworthy enough). Ultimately I believe it's bad to create a highly centralized `federation` (that's even the word that is used in the docs) unless there is a real good reason to do so. It creates a honeypot to be ddosed or regulated for any adversary, especially a regulator. Also - are we sure that these hard-coded node operators where real world identities are *very* easily attached - never become liable for failed/fraudulent btc transactions? Additionally: now that it's ready to be used with enough btcd's on the network to support it, shouldn't bisq consider switching to [neutrino](https://github.com/lightninglabs/neutrino/)?

(2) This clearly failed so far, quite some nodes are on satoshi 16 or lower, missing security patches etc.

(3) I agree this is a huge issue, but it's a problem the entire bitcoin network and all SPV wallets are facing. I as a bisq user do object a bisq node operator deciding for a particular fork on my behalf. Even though not happening frequently it's not an if, but a when there will be another contentious hard fork in the future. Bisq could do the following IMHO:
* warn user in the UI that the fork is happening (explain in plain english they why, how, what)
* give user a choice: `a) I want to deal with it on my own` vs. `b) I want some trusted bisq operators to decide for me`.
If the user decides for a), disable trading from the scheduled blockheight of the fork and only re-enable by the user actively agreeing to a disclaimer.
Only if the user decided for b) and only then the bisq client should connect to the hardcoded btc node operator's nodes and only for the limited time of uncertainty while the hard fork battle is happening. The SPV wallet should connect to all nodes again once the hardfork is clearly resolved via a bisq client software update.
One could argue now that having the proposed button above is "bad UX", but I believe bitcoin thrives from users participating, having opinions and making decisions. Bisq should not take away this fundamental right.

Taking all of the above into account I think there should be a path towards `bisq without permanent btc node federation`. Request for comments ;)

kilrau:
Also (can't find the issue rn), I remember a discussion that bisq users will be able to use use own wallets and nodes in future, which renders the btc node operator features 1,2,3 useless anyways.

chimp1984:
Chain analysis companies run a good chunk of full nodes to spy in SPV wallets. Even if you dont connect to one at a time, it is enough if you connect once at all (after any restart) that you lose your wallet privacy (they will learn that all your wallet addresses are one entity). Federated nodes operated by Bisq devs are the less problematic solution here. But of course not the perfect one but as there is no feasible alternative atm for SPV / light wallet we have to stick around for longer with that. We researched a while back all alternatives and it turned out alternatives like Neutrino have not been solid enough (at least 6 months back) to be used. To use Electum servers seems to be the most realistic approach but that is the same model with federated servers. 
With the v2 trade protocol this issue would become less of an issue as Bisq would not require a BTC wallet anymore and people can use whatever they want (full node, hardware wallet,...).

kilrau:
> We researched a while back all alternatives and it turned out alternatives like Neutrino have not been solid enough (at least 6 months back) to be used.

Neutrino is still not rock solid, but it's usable. It's an official mainnet backend option for one of the major [LN implementations](https://github.com/lightningnetwork/lnd/pull/3358) by now and gets filters from `btcd` full nodes. Full bitcoind support to serve the filters in one of the upcoming two releases (https://github.com/bitcoin/bitcoin/pull/16442). I believe it's worth considering to start implementation and testing now since this is not done over night.

> To use Electum servers seems to be the most realistic approach but that is the same model with federated servers.

Agree, forgot to list electrum and definitely a good option, a much better option compared to bisq trying to maintain it's own federation.

> With the v2 trade protocol this issue would become less of an issue as Bisq would not require a BTC wallet anymore and people can use whatever they want (full node, hardware wallet,...).

Right, that's what I meant above in https://github.com/bisq-network/bisq/pull/3506#issuecomment-549331826. Still many users might not have or want to use a separate wallet and there should be a decent default choice.

In summary I believe it's time to consider switching to either Neutrino or Electrum.

PR #3522, opened 2019-10-30
Release/v1.2.0

ripcurlx:


ripcurlx:
hmm - I thought I can get rid of this error by merging master upfront. I think I'll have to add another commit to solve this issue

ripcurlx:
Deleting as v1.2.0 will still be used atm

PR #3523, opened 2019-10-30
Add capability to tradestats

chimp1984:


ripcurlx:
@chimp1984 I think this PR should be against v1.2.0 not master

chimp1984:
Was mistake

PR #3524, opened 2019-10-30
Add CapabilityRequiringPayload to TradeStatistics2

chimp1984:
With v1.2.0 we changed the way how the hash is created.
To not create too heavy load for seed nodes from
requests from old nodes we use the SIGNED_ACCOUNT_AGE_WITNESS
capability to send trade statistics only to new nodes.
As trade statistics are only used for informational purpose it will
not have any critical issue for the old nodes beside that they don't see the latest trades.


ripcurlx (approved):
utACK

PR #3529, opened 2019-10-31
Fix tradestat hash issue

chimp1984:
We used the hash from the protobuf file of the trade statistics and that can be a hash which was created with an older version where we used a different hash method (extra map included). This causes that the excludedKeys at GetDataRequest are not matching the seed nodes keys so our deltat handling fails which results in too much data transmitted (7 MB instead of 3MB). To fix that we force the tradestat to update its hash by passing a null to the constructor which causes a call of the new hash method. Further we need to add a new capability so that seed nodes only deliver trade stat objects in case the node has updated.


ripcurlx (approved):
utACK

PR #3539, opened 2019-11-01
Fix BTC valuation popups

### LINKED ISSUE #3515: Available balance pop-up currency wrong
### Description

At the top right corner of the window, we have the drop down menu for BTC rates, the available balance, the reserved funds and locked funds (all these last 3 in BTC units).
When the mouse is over any of these last three, the software calculates the value in the respective rate and displays a pop-up with the value converted to whatever currency was selected in the drop down menu.
Even though the conversion is working alright, the symbol that is always shown is "USD".
For example, check below what happens when BTC/COP is selected (Colombian Peso):

image

I wish I had 702k USD!!! :-)

#### Version

v1.2.0

### Steps to reproduce

Change the rate at the drop down and observe.

### Expected behaviour

In the example above, the pop-up should show 702.324,79 COP.

### Actual behaviour

Instead of showing "COP", it showed "USD" (it shows "USD" always)

devinbileck:
When hovering over the BTC balances, the popup was using the market
price based on the locale combined with the preferred currency as its
units. This lead to the situation where it could show a mismatch
between value and units.

Instead, use the market price based on the preferred currency.

Fixes #3515

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/MainView.java
+++ b/desktop/src/main/java/bisq/desktop/main/MainView.java
@@ -254,11 +254,12 @@ public class MainView extends InitializableView<StackPane, MainViewModel>
             protected Tooltip computeValue() {
                 String tooltipText = Res.get("mainView.balance.available");
                 try {
+                    String preferredTradeCurrency = model.getPreferences().getPreferredTradeCurrency().getCode();
                     double availableBalance = Double.parseDouble(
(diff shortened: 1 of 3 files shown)


freimair (approved):
Ack.

Although one has to restart the application in order to get a new preferred currency to work for the tooltip. Since one would not change the preferred currency on a regular basis I recon that is ok.

PR #3551, opened 2019-11-04
Add missing network capability for trade statistics hash update

ripcurlx:
I think we missed to add the new statistics hash capability to the core network capabilities.
At least this worked for me in my local nodes when I recognized some missing trade statistics.

### DIFF
--- a/core/src/main/java/bisq/core/setup/CoreNetworkCapabilities.java
+++ b/core/src/main/java/bisq/core/setup/CoreNetworkCapabilities.java
@@ -39,7 +39,8 @@ public class CoreNetworkCapabilities {
                 Capability.BUNDLE_OF_ENVELOPES,
                 Capability.MEDIATION,
                 Capability.SIGNED_ACCOUNT_AGE_WITNESS,
-                Capability.REFUND_AGENT
+                Capability.REFUND_AGENT,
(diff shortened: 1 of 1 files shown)


ripcurlx:
@chimp1984 @sqrrm Could you review if my finding is correct?

sqrrm:
I think you're right. I don't have the time right now to test it though.

ripcurlx:
> I think you're right. I don't have the time right now to test it though.

It did solve the problem on two of my nodes. Unfortunately during my release smoke test I got two runtime trade stats for 1st of Nov and I thought everything is right now and we just didn't have any trades the day before because everyone was updating and too afraid to trade during this transitional period.

ripcurlx:
Tested it now on three different live nodes and it seems to work now. Just waiting for an ACK from someone else. Also we need to decide if we want to shoot out another release soon (latest on Thursday).

freimair (approved):
Ack.

Assuming that without this fix, a fresh client would not get all trade stats on startup, I just fired up a fresh 1.2.2 instance and compared the trade stats with my alway-on instance. All trades are there. 26 trades on 2019-11-01.

chimp1984 (approved):
utACK

PR #3552, opened 2019-11-04
Only show approval icon for signer accounts

ripcurlx:
The icon state for this account
Bildschirmfoto 2019-11-04 um 15 10 57
should be the alert not the approval icon
Bildschirmfoto 2019-11-04 um 15 38 43


sqrrm (approved):
utACK

PR #3556, opened 2019-11-04
(2/8) [BUGFIX] Don't try and remove() if addMailboxData() fails 

julianknutsen:
Fix a bug where remove() was called instead of removeMailboxData() and
guard against future callers doing the same thing.

Additionally, fix a potential bug where the sequence number for the
remove was not incremented leading to unexpected add after remove behavior
documented in P2PDataStorageTest::addProtectectedStorageEntry_afterRemoveSameSeqNr().

For mailbox messages, this wouldn't actually be a problem due to the
AddOncePayload behavior, but all removes now increase the sequence number
making refactoring and analysis easier.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/P2PService.java
+++ b/p2p/src/main/java/bisq/network/p2p/P2PService.java
@@ -700,12 +700,12 @@ public class P2PService implements SetupListener, MessageListener, ConnectionLis
                     };
                     boolean result = p2PDataStorage.addProtectedStorageEntry(protectedMailboxStorageEntry, networkNode.getNodeAddress(), listener, true);
                     if (!result) {
-                        //TODO remove and add again with a delay to ensure the data will be broadcasted
-                        // The p2PDataStorage.remove makes probably sense but need to be analysed more.
-                        // Don't change that if it is not 100% clear.
                         sendMailboxMessageListener.onFault("Data already exists in our local database");
-                        boolean removeResult = p2PDataStorage.remove(protectedMailboxStorageEntry, networkNode.getNodeAddress(), true);
-                        log.debug("remove result=" + removeResult);
+
+                        // This should only fail if there are concurrent calls to addProtectedStorageEntry with the
+                        // same ProtectedMailboxStorageEntry. This is an unexpected use case so if it happens we
+                        // want to see it, but it is not worth throwing an exception.
+                        log.error("Unexpected state: adding mailbox message that already exists.");
                     }
                 } catch (CryptoException e) {
                     log.error("Signing at getDataWithSignedSeqNr failed. That should never happen.");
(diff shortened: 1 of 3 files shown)


p2p/src/main/java/bisq/network/p2p/P2PService.java
```diff
@@ -704,8 +704,14 @@ public void onBroadcastFailed(String errorMessage) {
...
-                        log.debug("remove result=" + removeResult);
+
+                        // Generate a new entry so sequence number is larger than previous add()
+                        ProtectedMailboxStorageEntry removeProtectedMailboxStorageEntry = p2PDataStorage.getMailboxDataWithSignedSeqNr(
+                                expirableMailboxStoragePayload,
+                                keyRing.getSignatureKeyPair(),
+                                receiversPublicKey);
+                        boolean removeResult = p2PDataStorage.removeMailboxData(removeProtectedMailboxStorageEntry, networkNode.getNodeAddress(), true);
+                        log.error("Unexpected state: adding mailbox message that already exists. removeMailboxData result=" + removeResult);
                     }
```
chimp1984, line 709: If the result from addProtectedStorageEntry is false we have not added data to the map or the sequence nr map. From my understanding I think the remove is pointless as there is nothing to remove anyway in case of a failed add. But maybe I am missing something. Do you see any case where this would make sense? Beside that your code looks correct to me. So if you agree that the whole result=false path is pointless, I would suggest to remove it instead.

julianknutsen: After more testing, it is even simpler than that. Even if the add() did update some state and still return false (which isn't the case), only the receiver of a mailbox entry can remove it so the sender shouldn't try.

This brings up one of my TODO list items for this module which is the fact that concurrent calls through the client API (getProtectedStorageEntry) can result in interleaved sequence numbers that cause unexpected failures. Moving to a contract where client API calls never fail due to stale sequence numbers will make the system easier to reason about. This could be done by synchronizing the client API paths to not hand out duplicate seq numbers.

I've added multiple tests in the normal add() and mailbox add() paths to document the current behavior to ensure my future patches don't change it.

julianknutsen:
Ok. I've rebased this branch with master now to see how it looks. In my opinion, this is much easier to review than the wall of merges before and it will have a clean history once it is goes through the merge-commit path. The downside is the prior files are gone that the code review comments referenced, but the comments are still visible so the reviewers can verify the changes were actually implemented.

Let me know what you think and if this is a good path moving forward.

freimair (dismissed):
utAck, however, the title is not correct anymore isn't it?

julianknutsen:
Absolutely right. I've updated the commit text and PR title to reflect the actual changes after PR review and comments. Thanks!

freimair (approved):
utAck

ripcurlx (approved):
utACK

PR #3557, opened 2019-11-04
(3/8) [REFACTOR] P2PDataStore::add/remove/refresh

julianknutsen:
Improve clarity and add comments for future readers to help them reason about the behavior.

This also paves the way for deduplicating the remove() paths and helps identify the shared code paths
that can be merged in future refactoring work.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
+++ b/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
@@ -316,33 +316,41 @@ public class P2PDataStorage implements MessageListener, ConnectionListener, Pers
                                                 boolean reBroadcast,
                                                 boolean checkDate) {
         log.trace("addPersistableNetworkPayload payload={}", payload);
-        byte[] hash = payload.getHash();
-        if (payload.verifyHashSize()) {
-            ByteArray hashAsByteArray = new ByteArray(hash);
-            boolean containsKey = getAppendOnlyDataStoreMap().containsKey(hashAsByteArray);
-            if (!containsKey || reBroadcast) {
-                if (!(payload instanceof DateTolerantPayload) || !checkDate || ((DateTolerantPayload) payload).isDateInTolerance(clock)) {
-                    if (!containsKey) {
-                        appendOnlyDataStoreService.put(hashAsByteArray, payload);
-                        appendOnlyDataStoreListeners.forEach(e -> e.onAdded(payload));
-                    }
-                    if (allowBroadcast)
-                        broadcaster.broadcast(new AddPersistableNetworkPayloadMessage(payload), sender, null, isDataOwner);
 
-                    return true;
-                } else {
-                    log.warn("Publish date of payload is not matching our current time and outside of our tolerance.\n" +
-                            "Payload={}; now={}", payload.toString(), new Date());
-                    return false;
-                }
-            } else {
-                log.trace("We have that payload already in our map.");
-                return false;
-            }
-        } else {
+        // Payload hash size does not match expectation for that type of message.
+        if (!payload.verifyHashSize()) {
             log.warn("We got a hash exceeding our permitted size");
             return false;
         }
+
+        ByteArray hashAsByteArray = new ByteArray(payload.getHash());
+        boolean payloadHashAlreadyInStore = getAppendOnlyDataStoreMap().containsKey(hashAsByteArray);
+
+        // Store already knows about this payload. Ignore it unless the caller specifically requests a republish.
+        if (payloadHashAlreadyInStore && !reBroadcast) {
+            log.trace("We have that payload already in our map.");
+            return false;
+        }
+
+        // DateTolerantPayloads are only checked for tolerance from the onMessage handler (checkDate == true). If not in
+        // tolerance, ignore it.
+        if (checkDate && payload instanceof DateTolerantPayload && !((DateTolerantPayload) payload).isDateInTolerance((clock))) {
+            log.warn("Publish date of payload is not matching our current time and outside of our tolerance.\n" +
+                    "Payload={}; now={}", payload.toString(), new Date());
+            return false;
+        }
+
+        // Add the payload and publish the state update to the appendOnlyDataStoreListeners
+        if (!payloadHashAlreadyInStore) {
+            appendOnlyDataStoreService.put(hashAsByteArray, payload);
+            appendOnlyDataStoreListeners.forEach(e -> e.onAdded(payload));
+        }
+
+        // Broadcast the payload if requested by caller
+        if (allowBroadcast)
+            broadcaster.broadcast(new AddPersistableNetworkPayloadMessage(payload), sender, null, isDataOwner);
+
+        return true;
     }
 
     // When we receive initial data we skip several checks to improve performance. We requested only missing entries so we
--- a/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
+++ b/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
@@ -319,7 +319,7 @@ public class P2PDataStorage implements MessageListener, ConnectionListener, Pers
 
         // Payload hash size does not match expectation for that type of message.
         if (!payload.verifyHashSize()) {
-            log.warn("We got a hash exceeding our permitted size");
+            log.warn("addPersistableNetworkPayload failed due to unexpected hash size");
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -316,33 +316,41 @@ public boolean addPersistableNetworkPayload(PersistableNetworkPayload payload,
...
-                }
-            } else {
-                log.trace("We have that payload already in our map.");
-                return false;
-            }
-        } else {
+        // Payload hash size does not match expectation for that type of message.
+        if (!payload.verifyHashSize()) {
             log.warn("We got a hash exceeding our permitted size");
             return false;
```
chimp1984, line 323: I think we do not check for exceeding, just if hash size is as expected.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -390,50 +390,64 @@ public boolean addProtectedStorageEntry(ProtectedStorageEntry protectedStorageEn
...
+
+            return false;
+        }
+
+        // Verify the ProtectedStorageEntry is well formed and valid for the add operation
+        if (!checkPublicKeys(protectedStorageEntry, true) ||
+                !checkSignature(protectedStorageEntry)) {
+
+            log.trace("addProtectedStorageEntry failed due to invalid entry");
+            return false;
```
chimp1984, line 399: Maybe better to split that to 2 checks and more concrete log messages.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -390,50 +390,64 @@ public boolean addProtectedStorageEntry(ProtectedStorageEntry protectedStorageEn
...
+        }
 
         boolean containsKey = map.containsKey(hashOfPayload);
-        if (containsKey) {
-            result = result && checkIfStoredDataPubKeyMatchesNewDataPubKey(protectedStorageEntry.getOwnerPubKey(), hashOfPayload);
+
+        // In a hash collision between two well formed ProtectedStorageEntry, the first item wins and will not be overwritten
+        if (containsKey &&
```
chimp1984: > In a hash collision between two well formed ProtectedStorageEntry, the first item wins and will not be overwritten
       
I think that comment is misleading. The checkIfStoredDataPubKeyMatchesNewDataPubKey methods just checks if the pubKey is the same and if the hash is matching. Hash collusion is not an issue as we use cryptographic hashes as keys.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -390,50 +390,64 @@ public boolean addProtectedStorageEntry(ProtectedStorageEntry protectedStorageEn
...
+                !checkIfStoredDataPubKeyMatchesNewDataPubKey(protectedStorageEntry.getOwnerPubKey(), hashOfPayload)) {
+
+            log.trace("addProtectedStorageEntry failed due to hash collision");
```
chimp1984: See above comment...

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -461,39 +461,64 @@ public boolean refreshTTL(RefreshOfferMessage refreshTTLMessage,
...
             return false;
         }
+
+        int sequenceNumber = refreshTTLMessage.getSequenceNumber();
+
+        // If we have seen a more recent operation for this payload, we ignore the current one
+        // TODO: I think we can return false here. All callers use the get() public API which increments the sequence number
+        // leaving only the onMessage() handler which doesn't look at the return value. It makes more intuitive sense that operations that don't
+        // change state return false.
+        if (sequenceNumberMap.containsKey(hashOfPayload) && sequenceNumberMap.get(hashOfPayload).sequenceNr == sequenceNumber) {
```
chimp1984, line 463: In OfferBookService it is used and would cause a error message in that case (thought I think its an impossible case). Have not followed further if that error message would cause more troubles but I think it would require more checks of all client usage code to be sure we can change return value.  Conceptually I agree.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -390,50 +390,64 @@ public boolean addProtectedStorageEntry(ProtectedStorageEntry protectedStorageEn
...
-            if (!containsKey || hasSequenceNrIncreased) {
-                // At startup we don't have the item so we store it. At updates of the seq nr we store as well.
-                map.put(hashOfPayload, protectedStorageEntry);
-                hashMapChangedListeners.forEach(e -> e.onAdded(protectedStorageEntry));
-                // printData("after add");
-            } else {
-                log.trace("We got that version of the data already, so we don't store it.");
-            }
+        // If we have seen a more recent operation for this payload, we ignore the current one
+        // TODO: I think we can return false here. All callers use the get() public API which increments the sequence number
```
chimp1984: There is quite a bit of client code dealing with the return value. Would require more analysis if we can safely change it. Conceptually I agree.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
julianknutsen: The helper methods already have verbose logging and each case will print out a separate message. I updated this PR to only add new messages for the failures that didn't have them before. It is worth another pass to clean these up again, but I don't want to pollute this stack with those changes.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
julianknutsen: I agree. I'll add my thoughts about this on (4/4) to explain more about the current state and my opinion on where it should go to make it a bit easier to reason about. I agree that it is possible to client code may expect the current behavior and I will dig into each path individually to understand what happens on failure.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
julianknutsen: Do you have much context on when this condition is expected?

My understanding of the check is to verify that the NewEntry.ownerPubkKey == StoredEntry.ownerPubKey, but how could this fail unless there were identical SeqNr & Payload hashes from different owners? Maybe a misbehaving node sets the value wrong?

I may be missing something really straight forward, but I am happy to update the comment to be more appropriate for the check and not focus on hash collisions.

julianknutsen:
Commits start at: 5512f34

This got stale and polluted with merge commits. Updated against master and all previous comments have been addressed.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -380,50 +390,50 @@ public boolean addProtectedStorageEntry(ProtectedStorageEntry protectedStorageEn
...
-            } else {
-                log.trace("We got that version of the data already, so we don't store it.");
-            }
+        boolean hasSequenceNrIncreased = hasSequenceNrIncreased(protectedStorageEntry.getSequenceNumber(), hashOfPayload);
 
-            if (hasSequenceNrIncreased) {
-                sequenceNumberMap.put(hashOfPayload, new MapValue(protectedStorageEntry.getSequenceNumber(), System.currentTimeMillis()));
-                // We set the delay higher as we might receive a batch of items
-                sequenceNumberMapStorage.queueUpForSave(SequenceNumberMap.clone(sequenceNumberMap), 2000);
+        // If we have seen a more recent operation for this payload, we ignore the current one
```
freimair, line 409:
```suggestion
        // If we have seen a more recent version of this payload, we ignore the current one
```

?

freimair (approved):
utAck

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
julianknutsen: This comment has been updated.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
julianknutsen: This has been renamed to matchesRelevantPubKey in later patches. The idea is that for ProtectedStorageEntrys we want the Entry.ownerPubKey to match the Payload.ownerPubKey.

For ProtectedMailboxStorageEntrys we want the Entry.receiversPubKey to match the Payload.ownerPubKey.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
julianknutsen: We don't necessarily see a different version of the payload as the contents are identical, but we do see different operations on that payload that increment the sequence number.

The producers of change (add/remove/refresh) always increment the SeqNr so it acts like a timestamp for which operations take precedence over one another. For example, a Remove operation with SeqNr 2 is superseded by an add with SeqNr 3.

In that way of thinking, it made sense to me that the SeqNrs are more tied to the operations that occur and less about the payload version.

Hope that helps.

ripcurlx (approved):
utACK - 👍 makes it way easier to read than before.

PR #3558, opened 2019-11-04
(4/8) Update behavior of add/remove/refresh on duplicate sequence numbers

julianknutsen:
These patches changs some unexpected behavior when handling operations that have duplicate sequence numbers.

Now, if no internal state is changed the entry points will always return false.

It also fixes a subtle behavior of remove() where the code would allow duplicate add() operations if a
remove was processed with an equivalent sequence number.

Now, all operations require an increasing sequence number to update any internal state. I think this is a much easier way to think about the system, but I have no historical context for if/when these operations required equivalent sequence numbers between add() and remove().

I think it also closes a hole where a malicious node could rebroadcast an add() as a remove() and have it succeed. There are still other holes, but it moves it in the right direction.

This is the highest risk change in this stack.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
+++ b/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
@@ -316,33 +316,41 @@ public class P2PDataStorage implements MessageListener, ConnectionListener, Pers
                                                 boolean reBroadcast,
                                                 boolean checkDate) {
         log.trace("addPersistableNetworkPayload payload={}", payload);
-        byte[] hash = payload.getHash();
-        if (payload.verifyHashSize()) {
-            ByteArray hashAsByteArray = new ByteArray(hash);
-            boolean containsKey = getAppendOnlyDataStoreMap().containsKey(hashAsByteArray);
-            if (!containsKey || reBroadcast) {
-                if (!(payload instanceof DateTolerantPayload) || !checkDate || ((DateTolerantPayload) payload).isDateInTolerance(clock)) {
-                    if (!containsKey) {
-                        appendOnlyDataStoreService.put(hashAsByteArray, payload);
-                        appendOnlyDataStoreListeners.forEach(e -> e.onAdded(payload));
-                    }
-                    if (allowBroadcast)
-                        broadcaster.broadcast(new AddPersistableNetworkPayloadMessage(payload), sender, null, isDataOwner);
 
-                    return true;
-                } else {
-                    log.warn("Publish date of payload is not matching our current time and outside of our tolerance.\n" +
-                            "Payload={}; now={}", payload.toString(), new Date());
-                    return false;
-                }
-            } else {
-                log.trace("We have that payload already in our map.");
-                return false;
-            }
-        } else {
+        // Payload hash size does not match expectation for that type of message.
+        if (!payload.verifyHashSize()) {
             log.warn("We got a hash exceeding our permitted size");
             return false;
         }
+
+        ByteArray hashAsByteArray = new ByteArray(payload.getHash());
+        boolean payloadHashAlreadyInStore = getAppendOnlyDataStoreMap().containsKey(hashAsByteArray);
+
+        // Store already knows about this payload. Ignore it unless the caller specifically requests a republish.
+        if (payloadHashAlreadyInStore && !reBroadcast) {
+            log.trace("We have that payload already in our map.");
+            return false;
+        }
+
+        // DateTolerantPayloads are only checked for tolerance from the onMessage handler (checkDate == true). If not in
+        // tolerance, ignore it.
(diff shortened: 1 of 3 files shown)


chimp1984 (dismissed):
utACK

I think all changes are correct but I am not 100% sure if some sublte changes might have unwanted consequences. It will require very careful testing to ensure that we don't break anything. We should merge that into master once we don't have release-pressure to have enough time that devs are running from source and potential issues are discovered.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -391,41 +391,27 @@ public boolean addProtectedStorageEntry(ProtectedStorageEntry protectedStorageEn
...
+        // If we have already seen an Entry with the same hash, verify the new Entry has the same owner
+        if (containsKey && !checkIfStoredDataPubKeyMatchesNewDataPubKey(protectedStorageEntry.getOwnerPubKey(), hashOfPayload))
             return false;
-        }
 
         boolean hasSequenceNrIncreased = hasSequenceNrIncreased(protectedStorageEntry.getSequenceNumber(), hashOfPayload);
 
         // If we have seen a more recent operation for this payload, we ignore the current one
         // TODO: I think we can return false here. All callers use the get() public API which increments the sequence number
         // leaving only the onMessage() handler which doesn't look at the return value. It makes more intuitive sense that adds() that don't
```
chimp1984: This comment ("leaving only the onMessage() handler which doesn't look at the return value") is not true. In P2PService is a method which further gets used by client code ('p2PService.addProtectedStorageEntry') and some are operating on the return value. There are a few of those cases (I have not looked up now if it was in that particular case but I guess the statement that only the onMessage handler is calling the method is not correct.

chimp1984 (commented):
Regarding the concurrency tests:
Bisq uses single threading model as far as possible (exceptions, P2P connection class, BitcoinJ internal classes, disc persistance classes, some util stuff,...). The P2PDataStorage has no concurrent access as all network data are mapped to the UserThread.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
julianknutsen: Thanks again for looking through this.

I think my terminology may just be off here. When I refer to the "get() public API" I mean the P2PService callers that use the getProtectedStorageEntry/getRefreshTTLMessage/getMailboxDataWithSignedSeqNr APIs which grab an increasing sequence number before calling back into addProtectedStorageEntry/refreshTTL/remove/removeMailboxData. 

Maybe I should use the term "internal callers" or "P2PService users" or "non-onMessage callers" or something else to describe the code path that follows the pattern:

```
addProtectedStorageEntry(getProtectedStorageEntry());
```

What I am trying to communicate is that there are two paths through this code, the first is the onMessage handler which doesn't care about the return value so no behavior change will be noticed.

The second is the P2PService ("get() public API") which always uses an increasing sequence number so any failure paths that are due to sequence numbers won't trigger and the behavior will be the same as before.

I hope that helps communicate my intention on the deduplication of these checks.

julianknutsen:
Thanks for the information about the threading model. I'll dig into it tomorrow to get a better understanding and refresh this stack to remove the tests and comments around impossible interleavings.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
julianknutsen: All of these comments were addressed in the pull request they were added (3/4).

julianknutsen:
Commits start at: c1ad6b4

The request got stale and polluted with merge commits and has now been updated with master. Chimp's utAck outlined the risk of possible subtle changes to the client behavior since we may return false value to the client code now when it may previously have returned true.

After more review of the client caller code, I don't believe there is risk to client callers. All client code paths generate new Entrys that have sequence numbers greater than any previously seen sequence number (old + 1). This means that any new behavior in the P2PDataStore that requires increasing sequence numbers won't have any effect on those client callers since it isn't possible to generate a non-valid sequence number (with the old or new checks).

Instead, this new behavior changes what happens when messages are received from either peer nodes or seed nodes that have sequence numbers identical to those that have been seen before. It was possible to allow adds-after-removes in the old code which would have re-added data and signaled listeners.

Since this is a fundamental change to the P2P message handling logic. I encourage a merge when things are stable and there is no release pressure as suggested. I've done my best to add thorough tests, validate it in my regtest cluster, and run a production node with my changes and audit the logs to catch any glaring issues, but I've only been looking at this code for 3 weeks and don't have nearly the breadth of knowledge as the other developers who have touched this code recently.

Thanks for considering this request and I am happy to answer any other questions that come up.

freimair (approved):
utAck

ripcurlx (approved):
utACK

PR #3559, opened 2019-11-05
Limit data response

chimp1984:
To avoid that seed nodes get overloaded with requests for too many
PersistableNetworkPayload and ProtectedStorageEntry data we limit nr. of
entries to max 5000 each.


chimp1984:
Closed as superseeded by https://github.com/bisq-network/bisq/pull/3562 to master

PR #3562, opened 2019-11-05
Limit max. nr. of PersistableNetworkPayload and ProtectedStorageEntries

chimp1984:
To avoid that seed nodes get overloaded with requests for too many
PersistableNetworkPayload and ProtectedStorageEntry data we limit nr. of
entries to max 10000.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/peers/getdata/GetDataRequestHandler.java
+++ b/p2p/src/main/java/bisq/network/p2p/peers/getdata/GetDataRequestHandler.java
@@ -17,6 +17,7 @@
 
 package bisq.network.p2p.peers.getdata;
 
+import bisq.network.p2p.NodeAddress;
 import bisq.network.p2p.network.CloseConnectionReason;
 import bisq.network.p2p.network.Connection;
 import bisq.network.p2p.network.NetworkNode;
@@ -39,6 +40,7 @@ import com.google.common.util.concurrent.SettableFuture;
 
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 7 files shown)


ripcurlx (approved):
utACK

p2p/src/main/java/bisq/network/p2p/peers/getdata/GetDataRequestHandler.java
```diff
@@ -138,25 +140,40 @@ public void onFailure(@NotNull Throwable throwable) {
...
 
+        AtomicInteger maxSize = new AtomicInteger(MAX_ENTRIES);
         Set<P2PDataStorage.ByteArray> excludedKeysAsByteArray = P2PDataStorage.ByteArray.convertBytesSetToByteArraySet(getDataRequest.getExcludedKeys());
         Set<ProtectedStorageEntry> filteredSet = dataStorage.getMap().entrySet().stream()
                 .filter(e -> !excludedKeysAsByteArray.contains(e.getKey()))
+                .filter(e -> maxSize.decrementAndGet() >= 0)
                 .map(Map.Entry::getValue)
                 .collect(Collectors.toSet());
+        if (maxSize.get() <= 0) {
+            log.warn("The peer request caused too much ProtectedStorageEntry entries to get delivered. We limited the entries for the response to {} entries", MAX_ENTRIES);
```
julianknutsen: Is the UpdateDataRequest the only way to retrieve items > 10,000? What if the total set is larger than what the Initial & Update can handle? Is there a periodic on as well to sync over time?

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -504,6 +504,37 @@ public boolean remove(ProtectedStorageEntry protectedStorageEntry,
...
+
+    /**
+     * This method must be called only from client code not from network messages! We omit the ownership checks
+     * so we must apply it only if it comes from our trusted application code. It is used from client code which detects
+     * that the domain object violates specific domain rules.
+     * We could make it more generic by adding an Interface with a generic validation method.
+     *
+     * @param protectedStorageEntry     The entry to be removed
+     */
+    public void removeInvalidProtectedStorageEntry(ProtectedStorageEntry protectedStorageEntry) {
```
julianknutsen, line 516: This leads to data structure inconsistency in the event that two consumers listen for the same Payload type and one consumer calls into this function to remove the data. It also breaks encapsulation having the consumers know about the internals of the P2PDataStore. This isn't perfect now, anyway, but moving towards a model where P2PDataStore is private to P2PService may be a good goal.

It seems like a cleaner way to do the same thing would be to have the P2PDataStore validate each stored entry prior to becoming "active" or signaling listeners. The ProtectedStorageEntry and/or payload could implement an interface that would do this check and remove the data before it was every available for consumers.

Is that what you were thinking with the "generic validation method"? Doing this would dovetail well with my existing refactoring and I can probably try it out and see how it looks. I think it would give a lot more flexibility for the future if there are certain Entrys or Payloads that we need to purge but made it past prior validation. Here is just one example I found while testing:
https://github.com/bisq-network/bisq/blob/0ab156f2496d1dda2dd0543b9653bdd126280f76/p2p/src/test/java/bisq/network/p2p/storage/payload/ProtectedMailboxStorageEntryTest.java#L96-L109

PR #3563, opened 2019-11-05
Hide avoid standby mode feature on *nux OS

### LINKED ISSUE #3223: V1.1.6 "Avoid Standby Mode" does not work on Ubuntu 18

ghubstan:
Displays standby mode button on only on Windows and OSX,
and hides it on Linux and Unix distributions.

TitledGroupBg num rows reduced to 7 when standby mode button
is not displayed.

Fixes #3223

Replaces PR #3322 -- rejected because source file reformat
rearanged class level field declarations, making review
more difficult.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/settings/preferences/PreferencesView.java
+++ b/desktop/src/main/java/bisq/desktop/main/settings/preferences/PreferencesView.java
@@ -53,6 +53,7 @@ import bisq.core.util.validation.IntegerValidator;
 import bisq.common.UserThread;
 import bisq.common.app.DevEnv;
 import bisq.common.util.Tuple3;
+import bisq.common.util.Utilities;
 
 import org.bitcoinj.core.Coin;
 
@@ -144,7 +145,7 @@ public class PreferencesView extends ActivatableViewAndModel<GridPane, Preferenc
     private ChangeListener<Boolean> useCustomFeeCheckboxListener;
     private ChangeListener<Number> transactionFeeChangeListener;
     private final boolean daoOptionsSet;
-
+    private final boolean displayStandbyModeFeature;
 
     ///////////////////////////////////////////////////////////////////////////////////////////
     // Constructor, initialisation
@@ -173,6 +174,7 @@ public class PreferencesView extends ActivatableViewAndModel<GridPane, Preferenc
                 rpcUser != null && !rpcUser.isEmpty() &&
                 rpcPassword != null && !rpcPassword.isEmpty() &&
                 rpcBlockNotificationPort != null && !rpcBlockNotificationPort.isEmpty();
+        this.displayStandbyModeFeature = Utilities.isOSX() || Utilities.isWindows();
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


chimp1984:
We still enable it on Unix, no? See AvoidStandbyModeService

ghubstan:
> We still enable it on Unix, no? See AvoidStandbyModeService

I see nothing in the issues or AvoidStandbyModeService src explicitly saying there are any problems with standby mode on Unix.  I had no good reason to assume standby mode won't work on Unix too, but I did.    

This PR would disable the feature on all flavors of Unix.  I can make sure the feature is not hidden on Unix.  Do you want me to do that?

ripcurlx:
But don't we have a problem for all not Windows and macOS systems if that preference was already set to true? So the sound gets played and they are not able to deactivate it.

ghubstan:
Closing and reopening PR to force travis-ci check.

desktop/src/main/java/bisq/desktop/main/settings/preferences/PreferencesView.java
```diff
@@ -806,8 +811,10 @@ private void activateDisplayPreferences() {
 
         // We use opposite property (useStandbyMode) in preferences to have the default value (false) set as we want it,
         // so users who update gets set avoidStandbyMode=true (useStandbyMode=false)
-        avoidStandbyMode.setSelected(!preferences.isUseStandbyMode());
-        avoidStandbyMode.setOnAction(e -> preferences.setUseStandbyMode(!avoidStandbyMode.isSelected()));
+        if (displayStandbyModeFeature) {
+            avoidStandbyMode.setSelected(!preferences.isUseStandbyMode());
+            avoidStandbyMode.setOnAction(e -> preferences.setUseStandbyMode(!avoidStandbyMode.isSelected()));
+        }
```
freimair, line 819:
```suggestion
        } else {
            preferences.setUseStandbyMode(false)
        }
```

I agree with @ripcurlx, thus, the change suggestion. I believe that is a good place to put it. Instead of attaching a switch, we set it to false.

freimair (changes requested):
other than that, I share the opinion of @wiz: Linux users have to disable sleep-mode on their own. Showing the switch on Linux promises something that does not work.

chimp1984:
I think before changing behaviour we should be sure if the existing solution works on Linux and if it is needed as well if there is another better option to do it (e.g. via VM argument). 
To test this on OSX was pretty time consuming and difficult. As the result of the sleep mode behaviour can be that take offer attempts fail because the app is too slow to respond in time I think it is a rather serious issue which we only should change if we are sure to improve it. There have been only 1 or 2 people complaining about the sound (probably bad sound card produce artefacts) but before that fix we have many cases of failed take-offer attempts.

ripcurlx (approved):
ACK - it still runs as before on OSX (tested myself) and Windows and also removes this option on Linux. I agree that it is not good if we leave the user in the impression that this will prevent the application from going into standby mode, which seems to be not true as mentioned in https://github.com/bisq-network/bisq/issues/3223 .

freimair (approved):

PR #3566, opened 2019-11-06
Improve handling of failed trades and offers

chimp1984:
- Check if deposit tx is null
- Check if trade fee tx is rejected
- Listen to reject tx errors
- Cleanup addressEntryList
- Prevent opening disputes with if deposit tx is null
- Add null checks
- Improve logs
- Cleanups

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqSetup.java
+++ b/core/src/main/java/bisq/core/app/BisqSetup.java
@@ -55,6 +55,7 @@ import bisq.core.support.dispute.refund.RefundManager;
 import bisq.core.support.dispute.refund.refundagent.RefundAgentManager;
 import bisq.core.support.traderchat.TraderChatManager;
 import bisq.core.trade.TradeManager;
+import bisq.core.trade.TradeTxException;
 import bisq.core.trade.statistics.AssetTradeActivityCheck;
 import bisq.core.trade.statistics.TradeStatisticsManager;
 import bisq.core.user.Preferences;
@@ -107,6 +108,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
@@ -192,7 +194,8 @@ public class BisqSetup {
             spvFileCorruptedHandler, lockedUpFundsHandler, daoErrorMessageHandler, daoWarnMessageHandler,
             filterWarningHandler, displaySecurityRecommendationHandler, displayLocalhostHandler,
             wrongOSArchitectureHandler, displaySignedByArbitratorHandler,
-            displaySignedByPeerHandler, displayPeerLimitLiftedHandler, displayPeerSignerHandler;
+            displaySignedByPeerHandler, displayPeerLimitLiftedHandler, displayPeerSignerHandler,
+            rejectedTxErrorMessageHandler;
     @Setter
     @Nullable
     private Consumer<Boolean> displayTorNetworkSettingsHandler;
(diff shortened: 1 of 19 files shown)


core/src/main/java/bisq/core/app/BisqSetup.java
```diff
@@ -601,28 +604,56 @@ private void initWallet() {
...
+                    .filter(e -> setOfAllTradeIds.contains(e.getOfferId()) &&
+                            e.getContext() == AddressEntry.Context.MULTI_SIG)
+                    .forEach(e -> {
+                        Coin balance = e.getCoinLockedInMultiSig();
+                        if (balance.isPositive()) {
+                            String message = Res.get("popup.warning.lockedUpFunds",
+                                    formatter.formatCoinWithCode(balance), e.getAddressString(), e.getOfferId());
+                            log.warn(message);
+                            if (lockedUpFundsHandler != null) {
+                                lockedUpFundsHandler.accept(message);
```
sqrrm, line 630: Is there something preventing automatic opening of disputes here, rather than asking the trader to do it manually?

core/src/main/java/bisq/core/app/BisqSetup.java
```diff
@@ -651,13 +682,60 @@ private void initDomainServices() {
...
+                                }
+
+                                if (details != null) {
+                                    // We delay to avoid concurrent modification exceptions
+                                    String finalDetails = details;
+                                    UserThread.runAfter(() -> {
+                                        trade.setErrorMessage(newValue.getMessage());
+                                        rejectedTxErrorMessageHandler.accept(Res.get("popup.warning.trade.txRejected",
+                                                finalDetails, trade.getShortId(), txId));
+                                        tradeManager.addTradeToFailedTrades(trade);
```
sqrrm, line 731: It would be nice to avoid losing the offer fee tx and revert this to an active offer.

core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
```diff
@@ -221,6 +225,20 @@ protected void onSetupCompleted() {
...
                     blocksDownloadedFromPeer.set(peer);
                 });
+
+                // Need to be Threading.SAME_THREAD executor otherwise BitcoinJ will skip that listener
+                peerGroup.addPreMessageReceivedEventListener(Threading.SAME_THREAD, (peer, message) -> {
+                    if (message instanceof RejectMessage) {
+                        UserThread.execute(() -> {
+                            RejectMessage rejectMessage = (RejectMessage) message;
+                            String msg = rejectMessage.toString();
+                            log.error(msg);
```
sqrrm: I think this should be warning log level

core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
```diff
@@ -672,7 +672,13 @@ public void resetAddressEntriesForOpenOffer(String offerId) {
...
     public void resetAddressEntriesForPendingTrade(String offerId) {
         swapTradeEntryToAvailableEntry(offerId, AddressEntry.Context.MULTI_SIG);
-        // Don't swap TRADE_PAYOUT as it might be still open in the last trade step to be used for external transfer
+        // We swap also TRADE_PAYOUT to be sure all is cleaned  up. There might be cases where a user cannot send the funds
+        // to an external wallet directly in the last step of the trade, but the funds are in the Bisq wallet anyway and
+        // the dealing with the external wallet is pure UI thing. The user can move the funds to the wallet and then
+        // send out the funds to the external wallet. As this cleanup is a rare situation and most users do not use
+        // the feature to send out the funds we prefer that strategy (if we keep the address entry it might cause
+        // complications in some edge cases after a SPV resync).
+        swapTradeEntryToAvailableEntry(offerId, AddressEntry.Context.TRADE_PAYOUT);
```
sqrrm, line 681: Is there a risk of address reuse by re-labeling this address?

core/src/main/java/bisq/core/trade/TradeManager.java
```diff
@@ -681,27 +702,46 @@ public boolean isBuyer(Offer offer) {
...
+                    if (depositTx != null) {
+                        TransactionConfidence confidence = btcWalletService.getConfidenceForTxId(depositTx.getHashAsString());
+                        if (confidence != null && confidence.getConfidenceType() != TransactionConfidence.ConfidenceType.BUILDING) {
+                            tradeTxException.set(new TradeTxException(Res.get("error.closedTradeWithUnconfirmedDepositTx", trade.getShortId())));
+                        } else {
+                            log.warn("We found a closed trade with locked up funds. " +
+                                    "That should never happen. trade ID=" + trade.getId());
+                        }
+                    } else {
+                        tradeTxException.set(new TradeTxException(Res.get("error.closedTradeWithNoDepositTx", trade.getShortId())));
```
sqrrm, line 736: This means there are no funds locked in, should it still be part of the tradesIdSet representing locked in funds?

sqrrm (commented):
I have some comments, haven't tested it yet. I think it's basically good though.

core/src/main/java/bisq/core/app/BisqSetup.java
chimp1984: I am not sure how to reproduce that and if deposit tx is null then the user cannot open a dispute.

core/src/main/java/bisq/core/app/BisqSetup.java
chimp1984: I feat that is too complex to get it right. It should be exceptional case anyway. And users can request for reimbursement.

core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
chimp1984: Ok. I will change it.

core/src/main/java/bisq/core/app/BisqSetup.java
sqrrm: At this point just getting things more stable is good. Perhaps we could look at it if it's recurring event that people lose their fee this way.

core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
chimp1984: I think it was never used in that use case (failed trade), so its a fresh address.

core/src/main/java/bisq/core/trade/TradeManager.java
chimp1984: Hm... Not sure. But after the pending trade has been moved to failed trade I think the tradesIdSet is correct again. All those cases are very hard to reproduce/test...

sqrrm: Looks like it will work due to how the result of `getSetOfFailedOrClosedTradeIdsFromLockedInFunds` is used, but perhaps the function name is a bit misguiding since there is no locked in fund. Could return `null` there and filter out nulls before collect().

sqrrm (approved):
utACK

Issue #2996, opened 2019-07-24
Inject a clock instead of mocking currentTimeMS


christophsturm:
time is a dependency like any other and it needs to be injected where needed. 
see https://docs.oracle.com/javase/8/docs/api/java/time/Clock.html
https://github.com/robfletcher/test-clock

mrosseel:
would indeed be better for unit testing, any idea of the impact of this change?

christophsturm:
will probably be a big diff but it does not really change any logic, so it should be safe.
one additional issue is that we already have a class called "clock", that should probably be renamed to SecondsScheduler or something.
I can do both in separate prs. if people agree that its worthwile

sqrrm:
While it's nice to clean up the code I wonder if it's really worthwhile right now. I see issues with current trade protocol and performance issues as more urgent. People are having problem at startup with the app not starting or being very slow and there are a few bugs out there causing people to lose money trying to initiate trades or transactions failing for unclear reasons.

This kind of refactor is probably a lot easier to understand though, and perhaps a good start rather than diving straight into the most sensitive parts of the code. All depends on how much work and the risk to stability.

christophsturm:
I'm just writing issues as I notice problems, so others can chime in with their opinion.

sqrrm:
Yes, that's how I understood it. I'm adding my opinion on where effort is probably best spent at this point. I think it's good that you open these issues and I don't disagree with the issue itself.

PR #3568, opened 2019-11-07
(5/8) Test P2PDataStore expiration code and make testing time sensitive code easier.

julianknutsen:
This new set of patches starts at de72d39 (The rest are just merges of my previous pull requests).

The end goal of this stack is to test the expiration code that runs periodically in P2PDataStore. Since the tests require modification of time and the current objects made use of System.currentTimeMillis() (an unmockable static function) a detour of plumbing and dependency injection was required to get the objects in a state where they could be tested. #3037 looks to have started the pattern in other modules.

The end result is that all tests in P2PDataStore now have fine-grained control of the clock and can easily test the various expiration and purge functions. The periodic expiration code is now fully tested.

With the addition of these tests, the code coverage now stands at 81% line and 70% branch coverage.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
+++ b/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
@@ -316,33 +316,41 @@ public class P2PDataStorage implements MessageListener, ConnectionListener, Pers
                                                 boolean reBroadcast,
                                                 boolean checkDate) {
         log.trace("addPersistableNetworkPayload payload={}", payload);
-        byte[] hash = payload.getHash();
-        if (payload.verifyHashSize()) {
-            ByteArray hashAsByteArray = new ByteArray(hash);
-            boolean containsKey = getAppendOnlyDataStoreMap().containsKey(hashAsByteArray);
-            if (!containsKey || reBroadcast) {
-                if (!(payload instanceof DateTolerantPayload) || !checkDate || ((DateTolerantPayload) payload).isDateInTolerance(clock)) {
-                    if (!containsKey) {
-                        appendOnlyDataStoreService.put(hashAsByteArray, payload);
-                        appendOnlyDataStoreListeners.forEach(e -> e.onAdded(payload));
-                    }
-                    if (allowBroadcast)
-                        broadcaster.broadcast(new AddPersistableNetworkPayloadMessage(payload), sender, null, isDataOwner);
 
-                    return true;
-                } else {
-                    log.warn("Publish date of payload is not matching our current time and outside of our tolerance.\n" +
-                            "Payload={}; now={}", payload.toString(), new Date());
-                    return false;
-                }
-            } else {
-                log.trace("We have that payload already in our map.");
-                return false;
-            }
-        } else {
+        // Payload hash size does not match expectation for that type of message.
+        if (!payload.verifyHashSize()) {
             log.warn("We got a hash exceeding our permitted size");
             return false;
         }
+
+        ByteArray hashAsByteArray = new ByteArray(payload.getHash());
+        boolean payloadHashAlreadyInStore = getAppendOnlyDataStoreMap().containsKey(hashAsByteArray);
+
+        // Store already knows about this payload. Ignore it unless the caller specifically requests a republish.
+        if (payloadHashAlreadyInStore && !reBroadcast) {
+            log.trace("We have that payload already in our map.");
+            return false;
+        }
+
+        // DateTolerantPayloads are only checked for tolerance from the onMessage handler (checkDate == true). If not in
+        // tolerance, ignore it.
+        if (checkDate && payload instanceof DateTolerantPayload && !((DateTolerantPayload) payload).isDateInTolerance((clock))) {
+            log.warn("Publish date of payload is not matching our current time and outside of our tolerance.\n" +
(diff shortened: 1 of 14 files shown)


mrosseel:
have a look at: #2996 for a similar Clock related proposal.

I found this in P2PModule as an example of injecting a clock:
`bind(Clock.class).toInstance(Clock.systemDefaultZone());`
which is then used in the constructor of P2PDataStorage.

At first glance I did not see you use injection to instantiate Clocks, sorry if I missed it.

julianknutsen:
Thanks for taking a look. I use the same injection that you found. Work was done previously in #3037 to do the injection, it just wasn't actually used yet, so I pushed it across the finish line.

p2p/src/main/java/bisq/network/p2p/storage/payload/ProtectedStorageEntry.java
```diff
@@ -41,25 +43,29 @@
     private int sequenceNumber;
     private byte[] signature;
     private long creationTimeStamp;
+    transient private final Clock clock;
```
julianknutsen: There is an alternative here that doesn't involve storing the clock inside the data structure. The Public API methods that need a clock (ctor/refresh/backdate/isExpired) can take one as a parameter for the work they need individually. This will involve plumbing a Clock into OfferBookService or P2PService.

The protobuf deserialization path still needs access to a clock, so it will still need the resolver instance where it can be used in maybeAdjustCreationTimeStamp().

If someone has a strong opinion on this or maybe @christophsturm can comment on what he envisioned with the clock injections throughout the code. I'd rather stick to a common pattern to make this work consistent in the future.

chimp1984: I did not have time to look closer but adding Clock to ProtectedStorageEntry does not feel right to me. Using injecting for service classes is fine and adding clock there, but I think it should not be part of the value objects. 
Those methods on ProtectedStorageEntry using time could be moved to the service class and the ProtectedStorageEntry is passed as an object (just a rough idea, had not looked closer to the PR...). Those heavy value objects doing too much and having external dependencies (time) have been a bad idea anyway and maybe thats a good opportunity to refactor that.

julianknutsen: After sleeping on it and coming back to it today, I don't like the idea of storing the clock in the object. Not necessarily because it is value-heavy, but because it is more testable and cleaner to pass in a new instance if needed instead of relying on the one that was generated at instantiation time

I've updated the stack with the patch that removes it in favor of passing it into the functions that need it.

After reviewing and refactoring this code over the last few weeks I am actually of the opinion that this needs more OO not less to stay maintainable and testable. Having one large integration test for everything isn't a great path forward. I've got a stack almost ready that leverages the inheritance that already exists to make the code a lot easier test and deduplicates a lot of the redundant helper functions.

I'll send it out tomorrow and whoever is interested can take a look and see if the pros outweigh the cons.

Thanks again for your feedback.

julianknutsen:
I got some small out-of-band requests for style and naming in 5-9. I'll get those cleaned up tomorrow and 5-9 will be ready for final review.

Thanks again to everyone for taking the time to look through these. I know it wasn't planned dev or review work and bringing on new devs is more time consuming than normal, but it is appreciated.

p2p/src/main/java/bisq/network/p2p/storage/payload/ProtectedStorageEntry.java
julianknutsen: This has been fixed. ProtectedStorageEntry takes in a Clock in all usages instead of storing it.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -121,6 +124,8 @@
     private final Set<ProtectedDataStoreListener> protectedDataStoreListeners = new CopyOnWriteArraySet<>();
     private final Clock clock;
 
+    protected int maxSequenceNumberMapSizeBeforePurge;
```
freimair, line 127: can we have a 
```suggestion
    protected static final int MAX_SEQUENCE_NUMBER_MAP_SIZE_BEFORE_PURGE;
```
or at least 
```suggestion
    protected final int maxSequenceNumberMapSizeBeforePurge;
```
?

or as I am sure you thought about that, why did you decide against that in the first place?

freimair (changes requested):
utAck. However, see inline comment...

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
julianknutsen: https://github.com/julianknutsen/bisq/commit/850cd9180213f396180b447d6164e7e8c3900ac2

Guice is really new to me, but from what I found I think this solution will be better. The App will just inject the static value and the tests can pass it in through the normal constructor.

The constant didn't feel right to put in NetworkOptionKeys so let me your preference for it.

Let me know your comments and I'll create the PR to fix this up.

freimair: I agree with the changes you made in the commit you referenced.

ripcurlx (approved):
utACK

freimair (approved):

PR #3570, opened 2019-11-07
Update data stores and bitcoinj checkpoints

ripcurlx:
I compared the hashes with my lite node and everything seems to be the same.
Bildschirmfoto 2019-11-07 um 08 46 55
Bildschirmfoto 2019-11-07 um 08 47 08
Bildschirmfoto 2019-11-07 um 08 47 20


sqrrm (approved):
ACK

PR #3572, opened 2019-11-07
PaymentAccount object removal from ObservableSet issue

### LINKED ISSUE #3461: revolut/uphold account - unable to remove after creation
### Description

revolut account is not able to be removed immediately after creation using minus button
Though able to delete after restart.

Able to observe this behavior for uphold too
#### Version

1.2.0

### Steps to reproduce

1. create revolut account
2. remove that account using minus button in account list
3. click yes on confirm popup
4. It is not removed from listing

### Expected behaviour

Account should be removed

### Actual behaviour

Account is not removed

#### Device or machine

Ubuntu

beingindot:
Issue:
https://github.com/bisq-network/bisq/issues/3461

Addresses issue by adding equals/hashcode override.
Also added one unit test.

fixes #3461

### DIFF
--- a/core/src/main/java/bisq/core/payment/PaymentAccount.java
+++ b/core/src/main/java/bisq/core/payment/PaymentAccount.java
@@ -43,15 +43,17 @@ import javax.annotation.Nullable;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-@EqualsAndHashCode
+@EqualsAndHashCode(onlyExplicitlyIncluded = true)
 @ToString
 @Getter
 @Slf4j
 public abstract class PaymentAccount implements PersistablePayload {
     protected final PaymentMethod paymentMethod;
     @Setter
+    @EqualsAndHashCode.Include
     protected String id;
     @Setter
+    @EqualsAndHashCode.Include
     protected long creationDate;
     @Setter
     public PaymentAccountPayload paymentAccountPayload;
--- a/core/src/test/java/bisq/core/payment/PaymentAccountsTest.java
+++ b/core/src/test/java/bisq/core/payment/PaymentAccountsTest.java
@@ -21,12 +21,17 @@ import bisq.core.account.witness.AccountAgeWitness;
 import bisq.core.account.witness.AccountAgeWitnessService;
 import bisq.core.offer.Offer;
 import bisq.core.payment.payload.PaymentAccountPayload;
+import bisq.core.user.UserPayload;
+
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableSet;
 
 import java.util.Collections;
 
 import org.junit.Test;
 
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.mock;
@@ -70,5 +75,22 @@ public class PaymentAccountsTest {
         when(service.getMyWitness(payload)).thenReturn(witness);
 
         return account;
+    }
+
+    @Test
+    public void testRemoveAccount() {
+        PaymentAccount account = new RevolutAccount();
+        account.setAccountName("account name");
+
+        ObservableSet<PaymentAccount> paymentAccountsAsObservable;
+        paymentAccountsAsObservable = FXCollections.observableSet(new UserPayload().getPaymentAccounts());
+
+        //add element
+        paymentAccountsAsObservable.add(account);
+
+        //remove element with some state change
+        account.setAccountName("new account name for new hash code");
+
+        assertTrue(paymentAccountsAsObservable.remove(account));
     }
 }


freimair:
I just tried to reproduce the issue by only using your Test but without your fix. Turns out the test does not fail then either. Have you been able to reproduce the issue with the test?

(because I wanted to try if a ```@EqualsAndHashCode.Exclude protected String accountName;``` might do the trick as well.)

EDIT: it seems mockito does not memorize the account name, nor does it memorize anything else. Hence, the hash-method returns always the same (null being null being null).  So it seems that this kind of test is pretty useless. @christophsturm do you have an opinion on that?

core/src/test/java/bisq/core/payment/PaymentAccountsTest.java
```diff
@@ -71,4 +76,24 @@ private static PaymentAccount createAccountWithAge(AccountAgeWitnessService serv
...
         return account;
     }
+
+    @Test
+    public void testRemoveAccount() {
+        PaymentAccountPayload payload = mock(PaymentAccountPayload.class);
+
+        PaymentAccount account = mock(PaymentAccount.class);
+        account.setAccountName("account name");
+        when(account.getPaymentAccountPayload()).thenReturn(payload);
```
julianknutsen: I believe mockito overrides equals() and hashCode() internally and you can't even stub them [Mockito 2.x limitations](https://github.com/mockito/mockito/wiki/FAQ#mockito-2x-specific-limitations). My guess is it results in standard memory address comparisons which is why it works before and after this patch and supersedes lomboks's generated methods.

Even if you could, the PaymentAccount is the system under tests so mocking it doesn't make sense. You would just be testing your mock implementation. Here is an example using a real object that fails before your patch and succeeds after it.

```
    @Test
    public void testRemoveAccount() {
        PaymentAccount account = new RevolutAccount();
        account.setAccountName("account name");

        ObservableSet<PaymentAccount> paymentAccountsAsObservable;
        paymentAccountsAsObservable = FXCollections.observableSet(new UserPayload().getPaymentAccounts());

        //add element
        paymentAccountsAsObservable.add(account);

        //remove element with some state change
        account.setAccountName("new account name");

        assertTrue(paymentAccountsAsObservable.remove(account));
    }
```

If you are intent on not using a production class, you can just subclass PaymentAccount with a PaymentAccountStub for this test that just sets the two instance variables you care about and inherits the new equals() and hashCode(). Lots of options.

core/src/main/java/bisq/core/payment/PaymentAccount.java
```diff
@@ -148,6 +149,20 @@ public TradeCurrency getSingleTradeCurrency() {
             return null;
     }
 
+    @Override
```
julianknutsen: This seems to work even though the EqualsAndHashCode annotation is still on the class. But, if you are overriding them it may make sense to remove it so lombok doesn't generate the code or future readers assume the generate equals is valid for this class.

julianknutsen (changes requested):
Test needs some work so it will fail before your patch. I've given an example of one inline.

beingindot:
Thanks @freimair for the test issue. Have resolved it now.

core/src/test/java/bisq/core/payment/PaymentAccountsTest.java
beingindot: Thanks @julianknutsen for providing alternate object creation. 
Have included the changes now. pls review

core/src/main/java/bisq/core/payment/PaymentAccount.java
```diff
@@ -148,6 +148,20 @@ public TradeCurrency getSingleTradeCurrency() {
...
+        PaymentAccount that = (PaymentAccount) o;
+        return creationDate == that.creationDate &&
+                id.equals(that.id);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, creationDate);
+    }
+
```
freimair: I would rather see a simple
```
@EqualsAndHashCode.Exclude
```
added to the `accountName` field rather than getting rid of lombok at all.

freimair (changes requested):
great! Now testing is good for something. thanks!

Could you try if my `@EqualsAndHashCode.Exclude` suggestion works? and if yes, update the PR accordingly?

core/src/main/java/bisq/core/payment/PaymentAccount.java
beingindot: `accountName` field is for test only I have added. In reality, we couldn't delete the object immediately after creation.(Not sure which field causes this)
Hence i feel it should be applied to entire class. pls let me know if otherwise.

chimp1984:
This issues also happens at Uphold or MoneyGram. All payment methods using a list of tradeCurrencies. 
After more debugging I found out that if you exclude `selectedTradeCurrency` in `PaymentAccount` with `@EqualsAndHashCode.Exclude` the issue is resolved.
But it is very weird as at a breakpoint at the remove method both accounts the one in the set and the one to remove are the same object identity and have exact the same data.
Anyone an idea whats going on here?

chimp1984 (changes requested):
NACK
We should find the real reason for the issue and not reduce the check to id and date only. See comment above...

chimp1984:
After further investigation it seems that ObservableSet behaves weird. If you copy the ObservableSet entries into a HashSet you can remove it.

```
   public void removePaymentAccount(PaymentAccount paymentAccount) {
        HashSet<PaymentAccount> temp = new HashSet<>(paymentAccountsAsObservable);
        boolean changed = temp.remove(paymentAccount);
        paymentAccountsAsObservable.clear();
        paymentAccountsAsObservable.addAll(temp);
        if (changed)
            persist();
    }
```

beingindot:
After this change i could remove uphold/revolut without any issue.

What i believe is because of hashcode change, we are not able to remove. hashcode change can be because of some state change. so usually we restrict that to unique identifying fields (id, date in this case) to maintain hashcode contract. 

I couldn't create moneygram account even in prod. will open a bug. https://github.com/bisq-network/bisq/issues/3598

ripcurlx (changes requested):
NACK
> After this change i could remove uphold/revolut without any issue.
> 
> What i believe is because of hashcode change, we are not able to remove. hashcode change can be because of some state change. so usually we restrict that to unique identifying fields (id, date in this case) to maintain hashcode contract.
> 
> I couldn't create moneygram account even in prod. will open a bug. #3598

But as @chimp1984 mentioned above I'm not sure we should fix it like that with only using id and date.  The account name cannot be changed in the interface so this can't be the problem of this bug IMO. Probably this part of the code doesn't need changes, but more the `removePaymentAccount` method in the User class as @chimp1984 pointed out. We have to find out why ObservableSet is behaving not as expected in that case.

> ```
>   public void removePaymentAccount(PaymentAccount paymentAccount) {
>         HashSet<PaymentAccount> temp = new HashSet<>(paymentAccountsAsObservable);
>         boolean changed = temp.remove(paymentAccount);
>         paymentAccountsAsObservable.clear();
>         paymentAccountsAsObservable.addAll(temp);
>         if (changed)
>             persist();
>     }
> ```

beingindot:
I have done deep debugging of this issue.

- when you insert a payment account, object (with hashcode x) will be inserted in hashset with key as x.
- when remove method called, that time hashcode of the same object will be y (some state change). so when we look for that object in hashset it wont find that key y (because x only will be there as key).
so remove wont happen.

 `paymentAccountsAsObservable` is just a wrapper of HashSet entries of payment accounts.

what the above code suggests:
1. when we again construct hashset entries using the same collection, it will build the hashset again using current hashcode value( will be y) and remove will work now.
2. we are _re-constructing same hashset_ everytime for remove method.

Real Issue: we need to implement hashcode contract correctly. that's what i believe have done.
@ripcurlx @chimp1984 if that's not the case, i'll add the above code, pls let me know.

julianknutsen:
I tracked it down and it brings up one of the dangers of modifying elements while they are inside Hash containers. The hash bucket for an item is computed at insert time by calling item.hashCode().

But, if the item is later changed such that hashCode() returns a different value, the remove will look for the item in the bucket specified by the hashCode() return value at removal time. This explains why just moving them all to a map and deleting also worked. Also, why after restart it works.

The "fix" is to ensure the hashCode is stable before inserting it into the HashSet. The attached patch works, but I would recommend testing it more or auditing the other users that do something similar. For example, the code is almost duplicated in AltCoinAccountsDataModel.

revolut-add
revolut-remove

```
diff --git a/desktop/src/main/java/bisq/desktop/main/account/content/fiataccounts/FiatAccountsDataModel.java b/desktop/src/main/java/bisq/desktop/main/account/content/fiataccounts/FiatAccountsDataModel.java
index 568b3d8c2..ed281def8 100644
--- a/desktop/src/main/java/bisq/desktop/main/account/content/fiataccounts/FiatAccountsDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/account/content/fiataccounts/FiatAccountsDataModel.java
@@ -106,7 +106,6 @@ class FiatAccountsDataModel extends ActivatableDataModel {
 
     public void onSaveNewAccount(PaymentAccount paymentAccount) {
-        user.addPaymentAccount(paymentAccount);
         TradeCurrency singleTradeCurrency = paymentAccount.getSingleTradeCurrency();
         List<TradeCurrency> tradeCurrencies = paymentAccount.getTradeCurrencies();
         if (singleTradeCurrency != null) {
@@ -128,6 +127,8 @@ class FiatAccountsDataModel extends ActivatableDataModel {
             });
         }
 
+        user.addPaymentAccount(paymentAccount);
+
         accountAgeWitnessService.publishMyAccountAgeWitness(paymentAccount.getPaymentAccountPayload());
     }
```

chimp1984:
@julianknutsen 
> modifying elements while they are inside Hash containers

Could you see that the objects got modified? I have not seen that but if so that would explain it.
I suspect the `selectedTradeCurrency` as when it was excluded it worked.

julianknutsen:
Yes, the screenshots show the tradeCurrencies object was different inside the debugger at insert time and remove time. This is confirmed with the attached patch that "fixes" it by moving the insert until after the tradeCurrencies are set. Previously, it was inserted, then modified.

A quick run through the UI shows that none of the other fields are modifiable after creation time, but I don't have enough background to know other operations that modify that state and if it will have the same issue. For example, can the list of supported currencies change after creation time which would cause the hash to change again?

It seems like using the fields that are guaranteed to be unique and stable is the way to go, but I'll leave that to the devs with more knowledge of that area. Knowing why it behaved strangely should be good enough to move forward now.

chimp1984:
> can the list of supported currencies change after creation time which would cause the hash to change again?

No. That the tradeCurrencies are not set initially is a bug. Once the account is set it is immutable.

chimp1984:
Fixed with https://github.com/bisq-network/bisq/pull/3612
Thanks @julianknutsen for your hint!

beingindot:
> It seems like using the fields that are guaranteed to be unique and stable is the way to go, 

This is what i tried in this PR.

freimair:
no worries, although the bug does not appear anymore due to @chimp1984 s PR, we are currently discussing if your approach breaks anything (we like your idea). But please use lombok for the Equals and Hash stuff.

julianknutsen:
I think the issue was more with understanding why it failed initially and not directly tied to the fix. The initial PR suggested that the state changed, but until there was a good understanding of which specific state changed, was it changed intentionally, and what is the scope of a PaymentAccount mutating, it is hard to come to a consensus on the right set of fields to guarantee uniqueness.

@beingindot I really appreciate the work to identify and work on this bug. You've started a good conversation on how to approach this type of issue in the future.

beingindot:
Thanks @freimair @julianknutsen. Have pushed the review changes accordingly.

chimp1984 (changes requested):
NACK
I don't see the need for excluding the other fields. It add some unneeded risk and testing effort. The objects should be immutable so any fields has to be fix. If we go further on that area I would recommend to think how we can refactor to make all those accounts immutable. But I think there are many other higher prio stuff to do....

beingindot:
Is it ok to add `@EqualsAndHashCode.Exclude` to only `selectedTradeCurrency`? 
If not required, we can close this PR.

chimp1984:
> Is it ok to add @EqualsAndHashCode.Exclude to only selectedTradeCurrency?
If not required, we can close this PR.

I think we should not change the @EqualsAndHashCode as that was not the source of the problem and the reason for the problem is solved. Manipulating the @EqualsAndHashCode  should be only done in cases where we really need it IMO. So I suggest to close that PR.

PR #3578, opened 2019-11-08
Fix failing test when there are spaces in the project path

stejbac:
Use URI.getPath() rather than URL.getFile() to get the absolute path of a test resource, since the latter fails to perform URL decoding. This breaks the tests when the project root contain chars needing %-encoding.

This fixes errors of the following form in _(BisqInstallerTest)_ _verifySignature_:

>java.io.FileNotFoundException: /home/steven/Documents/Java%20Projects/bisq/desktop/build/resources/test/downloadUpdate/F379A1C6.asc (No such file or directory)


freimair (dismissed):
utAck

ripcurlx:
@stejbac You have to sign your commits: https://help.github.com/en/github/authenticating-to-github/signing-commits

stejbac:
I've now added a GPG key to Github and amended the commit to include a signature.

freimair (approved):
Ack

PR #3580, opened 2019-11-08
Re-introduce account age to chargeback risk payment methods

ripcurlx:
We are not showing the account age info for payment methods that
have a high chargeback risk. This PR adds them back again.

Bildschirmfoto 2019-11-08 um 11 35 07
Bildschirmfoto 2019-11-08 um 11 35 20
Bildschirmfoto 2019-11-08 um 11 36 08

### DIFF
--- a/common/src/main/java/bisq/common/util/Tuple5.java
+++ b/common/src/main/java/bisq/common/util/Tuple5.java
@@ -19,31 +19,34 @@ package bisq.common.util;
 
 import java.util.Objects;
 
-public class Tuple4<A, B, C, D> {
+public class Tuple5<A, B, C, D, E> {
     final public A first;
     final public B second;
     final public C third;
     final public D forth;
+    final public E fifth;
 
-    public Tuple4(A first, B second, C third, D forth) {
+    public Tuple5(A first, B second, C third, D forth, E fifth) {
         this.first = first;
         this.second = second;
         this.third = third;
         this.forth = forth;
+        this.fifth = fifth;
     }
 
     @SuppressWarnings("SimplifiableIfStatement")
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
-        if (!(o instanceof Tuple4)) return false;
+        if (!(o instanceof Tuple5)) return false;
 
-        Tuple4<?, ?, ?, ?> tuple4 = (Tuple4<?, ?, ?, ?>) o;
+        Tuple5<?, ?, ?, ?, ?> tuple5 = (Tuple5<?, ?, ?, ?, ?>) o;
 
-        if (!Objects.equals(first, tuple4.first)) return false;
-        if (!Objects.equals(second, tuple4.second)) return false;
-        if (!Objects.equals(third, tuple4.third)) return false;
-        return Objects.equals(forth, tuple4.forth);
+        if (!Objects.equals(first, tuple5.first)) return false;
+        if (!Objects.equals(second, tuple5.second)) return false;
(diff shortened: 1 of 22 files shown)


desktop/src/main/java/bisq/desktop/components/PeerInfoIcon.java
```diff
@@ -146,19 +143,19 @@ private PeerInfoIcon(NodeAddress nodeAddress,
...
-                Res.get("peerInfoIcon.tooltip.trade.notTraded", role, fullAddress, accountAge);
+                Res.get("peerInfoIcon.tooltip.trade.traded", role, fullAddress, numTrades, accountAgeTooltip) :
+                Res.get("peerInfoIcon.tooltip.trade.notTraded", role, fullAddress, accountAgeTooltip);
 
         // outer circle
         Color ringColor;
         if (isFiatCurrency) {
 
-            switch (accountAgeWitnessService.getPeersAccountAgeCategory(peersAccount.first)) {
+            switch (accountAgeWitnessService.getPeersAccountAgeCategory(accountAge)) {
```
sqrrm: I think this should switch on `accountAge` for non signed accounts and switch on `signAge` for signed accounts.

Possibly it should switch on `signAge` for all accounts of risky payment methods. Otherwise it will show that an older account that isn't signed is safe, even though it's a risky method.

common/src/main/java/bisq/common/util/Tuple5.java
```diff
@@ -0,0 +1,62 @@
...
+
+package bisq.common.util;
+
+import java.util.Objects;
+
+public class Tuple5<A, B, C, D, E> {
+    final public A first;
+    final public B second;
+    final public C third;
+    final public D forth;
```
sqrrm: fourth

common/src/main/java/bisq/common/util/Tuple5.java
```diff
@@ -0,0 +1,62 @@
...
+        if (!(o instanceof Tuple5)) return false;
+
+        Tuple5<?, ?, ?, ?, ?> tuple5 = (Tuple5<?, ?, ?, ?, ?>) o;
+
+        if (!Objects.equals(first, tuple5.first)) return false;
+        if (!Objects.equals(second, tuple5.second)) return false;
+        if (!Objects.equals(third, tuple5.third)) return false;
+        if (!Objects.equals(forth, tuple5.forth)) return false;
+        return Objects.equals(fifth, tuple5.fifth);
+
```
sqrrm: Extra blank line

sqrrm (changes requested):
Please have look at comments, especially on the usage of account age category

desktop/src/main/java/bisq/desktop/components/PeerInfoIcon.java
ripcurlx: Yes, I wasn't sure in that case if we should use always accountAge or selectively switch the colors. But true for people used to the color codes it might trick them into false security if always accountAge is used. Will adapt it based on chargeback risk.

common/src/main/java/bisq/common/util/Tuple5.java
ripcurlx: ah damn English 😉

ripcurlx: or let's say damn copy & paste. I'll fix it in the Tuple4 as well

sqrrm (approved):
utACK

PR #3582, opened 2019-11-09
(6/8) Remove mutating functions of ProtectedStorageEntry

julianknutsen:
New commit is 2c2a57e

This is good clean up anyway, but it will make more sense with the next 2 pull requests that start changing the ProtectedStorageEntry and ProtectedMailboxStorageEntry APIs.


freimair (approved):
utAck

ripcurlx (approved):
utACK

PR #3583, opened 2019-11-09
(7/8) Develop APIs for ProtectedStorageEntry to enable testing and code refactoring

julianknutsen:
New commits start 54b4b4d

I expect a bit of discussion on this one based on my previous feedback, so I wanted to outline my thoughts here so that the best path forward can be taken with regards to testing, maintainability, and clean programming practices.

This entire stack was based on testing the P2PDataStore and refactoring it so it is more maintainable and easier to add new features in the future.

The module was almost impossible to test when starting out. I decided to write very heavy-handed integration tests while looking at the implementation to verify I covered all the branches and edge cases. The end result #3554 allowed refactoring to be done while ensuring the various combinations of ProtectedStorageEntrys and Payloads still worked correctly.

But in order to do that, the tests needed to create REAL ProtectedStorageEntry & ProtectedMailboxStorageEntry objects from scratch and manipulate them into the various valid and invalid configurations to exercise the code. Real keys, real hashes, a lot of overhead for testing that should be pretty straight forward.

The issue is that too much of the code exists in private helper functions inside the P2PDataStore itself. The only testing that can be done is front door integration testing where the tests know way too much about the implementation.

My solution to the complexity that I present in the next 2 pull requests is to turn ProtectedStorageEntry into a real object with a real API that can be unit tested. Determining whether or not an Entry is valid for an add operation shouldn't depend on the P2PDataStore internal state, or have to be run through the Client API to verify correctness.

If you are at all interested, I urge you to look at the unit tests in the following pull request (ProtectedStorageEntryTest & ProtectedMailboxStorageEntryTest) and see how easy it is to verify the correctness of an add or remove operation versus the path the integration test had to take in the beginning.

In addition to adding the ability to unit test the Entrys, real unit tests can now exist for the P2PDataStore. Instead of constructing a real object, they now just build a mock Entry (something that wasn't possible with the old object relationship). The result is that the unit tests for P2PDatastore can test only the code they need to (what data structures to update, who to signal, etc) and not worry about the implementation specifics of Mailbox vs. Normal Entrys.

The end result is code that has unit tests, has integration tests, is easy to navigate, and has line and branch coverage over 80%. More important than the numbers is that a developer can go in and change the code and get feedback on whether or not it worked. Something that wasn't possible before.

In summary, I would request that anyone who is interested look through the testing and weigh in on the maintainability and testing strategy of this refactor. The reality is that this code wasn't tested, has plenty of bugs, and was hard to maintain and develop against. It is only one of many that likely have the same problem and starting to have a discussion on the practices that everyone wants in Bisq should happen sooner than later.

Pull Request (7/8)

This pull request will introduce the API, test it, and use the new benefits to combine the remove() and removeMailboxData() paths. 

Pull Request(8/8)

This pull request will make use of the new objects to create clean mocks that clean up the integration tests and remove a lot of the heavy-handed object creation.

If you got this far, thanks again for taking the time to look.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
+++ b/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
@@ -400,7 +400,7 @@ public class P2PDataStorage implements MessageListener, ConnectionListener, Pers
             return false;
 
         // Verify the ProtectedStorageEntry is well formed and valid for the add operation
-        if (!protectedStorageEntry.isValidForAddOperation() || !checkSignature(protectedStorageEntry))
+        if (!protectedStorageEntry.isValidForAddOperation())
             return false;
 
         // In a hash collision between two well formed ProtectedStorageEntry, the first item wins and will not be overwritten
--- a/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
+++ b/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
@@ -316,33 +316,41 @@ public class P2PDataStorage implements MessageListener, ConnectionListener, Pers
                                                 boolean reBroadcast,
                                                 boolean checkDate) {
         log.trace("addPersistableNetworkPayload payload={}", payload);
-        byte[] hash = payload.getHash();
-        if (payload.verifyHashSize()) {
-            ByteArray hashAsByteArray = new ByteArray(hash);
-            boolean containsKey = getAppendOnlyDataStoreMap().containsKey(hashAsByteArray);
-            if (!containsKey || reBroadcast) {
-                if (!(payload instanceof DateTolerantPayload) || !checkDate || ((DateTolerantPayload) payload).isDateInTolerance(clock)) {
-                    if (!containsKey) {
-                        appendOnlyDataStoreService.put(hashAsByteArray, payload);
-                        appendOnlyDataStoreListeners.forEach(e -> e.onAdded(payload));
-                    }
-                    if (allowBroadcast)
-                        broadcaster.broadcast(new AddPersistableNetworkPayloadMessage(payload), sender, null, isDataOwner);
 
-                    return true;
-                } else {
-                    log.warn("Publish date of payload is not matching our current time and outside of our tolerance.\n" +
-                            "Payload={}; now={}", payload.toString(), new Date());
-                    return false;
-                }
-            } else {
-                log.trace("We have that payload already in our map.");
-                return false;
-            }
-        } else {
+        // Payload hash size does not match expectation for that type of message.
+        if (!payload.verifyHashSize()) {
             log.warn("We got a hash exceeding our permitted size");
             return false;
         }
+
+        ByteArray hashAsByteArray = new ByteArray(payload.getHash());
+        boolean payloadHashAlreadyInStore = getAppendOnlyDataStoreMap().containsKey(hashAsByteArray);
+
+        // Store already knows about this payload. Ignore it unless the caller specifically requests a republish.
+        if (payloadHashAlreadyInStore && !reBroadcast) {
+            log.trace("We have that payload already in our map.");
+            return false;
+        }
+
+        // DateTolerantPayloads are only checked for tolerance from the onMessage handler (checkDate == true). If not in
+        // tolerance, ignore it.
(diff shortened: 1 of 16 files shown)


p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -747,35 +747,6 @@ private boolean checkSignature(PublicKey ownerPubKey, byte[] hashOfDataAndSeqNr,
...
-        return result;
-    }
-
-    private boolean checkIfStoredMailboxDataMatchesNewMailboxData(PublicKey receiversPubKey, ByteArray hashOfData) {
-        ProtectedStorageEntry storedData = map.get(hashOfData);
-        if (storedData instanceof ProtectedMailboxStorageEntry) {
-            ProtectedMailboxStorageEntry entry = (ProtectedMailboxStorageEntry) storedData;
-            // publicKey is not the same (stored: sender, new: receiver)
-            boolean result = entry.getReceiversPubKey().equals(receiversPubKey)
-                    && get32ByteHashAsByteArray(entry.getProtectedStoragePayload()).equals(hashOfData);
```
freimair, line 794: you did not migrate the hash calculation and comparison to the `ProtectedMailboxStorageEntry`. Can you elaborate?

freimair (commented):
you are correct, this will trigger some discussion. On the one hand, I really like how this clears up the code base, makes everything testable and more readable as well. On the other hand, you rely on an object to report on its own validity. From my side, I like the testing motivation so that is OK with me. We can always revisit this topic in the future.

please note the inline comment. Awaiting your response there.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
julianknutsen: Absolutely. This one was really confusing at first and I'm happy to help explain so you can verify it is correct.

By renaming some variables it makes it a bit easier to understand:

```
private boolean checkIfStoredMailboxDataMatchesNewMailboxData(PublicKey removeEntryReceiversPubKey, ByteArray removeEntryPayloadHash) {	
        ProtectedStorageEntry existingEntry = map.get(removeEntryPayloadHash);	
        if (storedData instanceof ProtectedMailboxStorageEntry) {	
            ProtectedMailboxStorageEntry existingMailboxEntry = (ProtectedMailboxStorageEntry) existingEntry;	
            // publicKey is not the same (stored: sender, new: receiver)	
            boolean result = existingMailboxEntry.getReceiversPubKey().equals(removeEntryReceiversPubKey)	
                    && get32ByteHashAsByteArray(existingMailboxEntry.getProtectedStoragePayload()).equals(removeEntryPayloadHash);
```

So, it was verifying that the new removeEntry's payload has the same hash as the existingEntry's payload. But, the only way that we can get to this branch in the code is if the payload hashes are equal. Line 762 grabs the existingEntry based on the hash of the removeEntry's payload so they already must be identical because the existingEntry was inserted into the map based on its payload.

As an aside, it made me feel more confident that this was just duplicate behavior because it didn't occur in the non-mailbox check AND conceptually, the only thing different between the Mailbox and non-Mailbox checks is which Entry field should stay the same  between the initial add and update/remove.

julianknutsen:
Another way I have thought about this refactor and why I think it is a good path moving forward is that all of these instance methods are equivalent to just testable static methods that operate on final variables. For example, you could just have a bunch of helper functions that look like this:

```
public static boolean isValidForAddOperation(ProtectedStorageEntry entry) {
  // check signature
  Sig.verify(entry.getPubKey(), hash(entry.getProtectedPayload()))

  // check key matches... etc...
}
```

BUT, by using them as instance methods you get the benefit of being able to mock them easily for testing which I think outweighs the anxiety around hiding these checks behind an interface. It may make sense in the future to guarantee that these types of functions only operate on `final` variables so they are equivalent to static methods with test benefits.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
freimair: ah yes, thanks for sharing. Missed that one, although it is fairly obvious now.

freimair (approved):
utAck

ripcurlx (approved):
utACK

PR #3585, opened 2019-11-09
Sort existing gradle-witness.gradle

blabno:
This should make comparison in future much easier.


freimair:
I would love to see that in action! Thanks for showing that up!

However, I suggest we implement https://github.com/signalapp/gradle-witness/issues/34 ourselves, as the original seems to be unmaintained for quite some time and we have our own fork anyhow.

As for the implementation part, I already did https://github.com/bisq-network/gradle-witness/pull/2.

For going forward, I suggest we get the implementation merged, and then add the resulting gradle-witness.jar in this PR and be done with it.

blabno:
@freimair I think your fix is better. But this PR also does sort the current contents of the file so it's a good start.

freimair:
@blabno, Chris has merged the changes in the gradle-witness repo. Would you build the new gradle-witness.jar and include the binary in this PR so we can get this merged?

ripcurlx (dismissed):
NACK - Could you please address https://github.com/bisq-network/bisq/pull/3585#issuecomment-554433803?

blabno:
@ripcurlx @freimair new included.

freimair (approved):
Ack. gradle-witness.jar checksum matches, library checksums did not change.

PR #3586, opened 2019-11-09
Move Payment account creation and removal from UI to core

blabno:


core/src/main/java/bisq/core/exceptions/NotFoundException.java
```diff
@@ -0,0 +1,24 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.core.exceptions;
+
+public class NotFoundException extends RuntimeException {
```
chimp1984, line 20: I think the class name should be more specific. What is not found?

core/src/main/java/bisq/core/payment/validation/AltCoinAddressValidator.java
```diff
@@ -30,14 +34,8 @@
...
 
-
-
-import bisq.asset.AddressValidationResult;
-import bisq.asset.Asset;
-import bisq.asset.AssetRegistry;
-
 @Slf4j
-public final class AltCoinAddressValidator extends InputValidator {
+public class AltCoinAddressValidator extends InputValidator {
```
chimp1984: If that was removed for testing consider to use the mockito-extensions like in: https://github.com/bisq-network/bisq/pull/3587/commits/6d983dec8d37a038a6e9f1cc024f941050632140#diff-1bf51dc85c38e8dd66e85a3d5286d1e0

core/src/main/java/bisq/core/user/Preferences.java
```diff
@@ -78,7 +78,7 @@
 
 @Slf4j
 @Singleton
-public final class Preferences implements PersistedDataHost, BridgeAddressProvider {
+public class Preferences implements PersistedDataHost, BridgeAddressProvider {
```
chimp1984: If 'final' was removed for testing consider to use the mockito-extensions like in: https://github.com/bisq-network/bisq/pull/3587/commits/6d983dec8d37a038a6e9f1cc024f941050632140#diff-1bf51dc85c38e8dd66e85a3d5286d1e0

core/src/main/java/bisq/core/payment/PaymentAccountManager.java
```diff
@@ -0,0 +1,128 @@
...
+
+    public void removePaymentAccount(String id) {
+        PaymentAccount paymentAccount = user.getPaymentAccount(id);
+        if (paymentAccount == null) {
+            throw new NotFoundException(format("Payment account %s not found", id));
+        }
+        boolean isPaymentAccountUsed = openOfferManager.getObservableList().stream()
+                .anyMatch(openOffer -> id.equals(openOffer.getOffer().getMakerPaymentAccountId()));
+        if (isPaymentAccountUsed) {
+            throw new PaymentAccountInUseException(format("Payment account %s is used for open offer", id));
```
chimp1984: I prefer to use exceptions only in exceptional cases not in valid use case (user tries to remove account but app does not allow it as its used in trades or offers). Maybe best to refactor the check out to 'canRemove()' method and only call the method if it is allowed to remove.

desktop/src/main/java/bisq/desktop/main/account/content/altcoinaccounts/AltCoinAccountsDataModel.java
```diff
@@ -103,40 +92,16 @@ protected void deactivate() {
...
-        isPaymentAccountUsed = isPaymentAccountUsed || tradeManager.getTradableList().stream()
-                .filter(t -> t.getOffer().getMakerPaymentAccountId().equals(paymentAccount.getId()) ||
-                        paymentAccount.getId().equals(t.getTakerPaymentAccountId()))
-                .findAny()
-                .isPresent();
-        if (!isPaymentAccountUsed)
-            user.removePaymentAccount(paymentAccount);
-        return isPaymentAccountUsed;
+        try {
+            paymentAccountManager.removePaymentAccount(paymentAccount.getId());
```
chimp1984: Why not pass the paymentAccount and avoid the extra lookup and null check in paymentAccountManager.removePaymentAccount?

desktop/src/main/java/bisq/desktop/main/account/content/altcoinaccounts/AltCoinAccountsDataModel.java
```diff
@@ -103,40 +92,16 @@ protected void deactivate() {
...
-                    preferences.addCryptoCurrency((CryptoCurrency) tradeCurrency);
-            });
-        }
-
-        if (!(paymentAccount instanceof AssetAccount))
-            accountAgeWitnessService.publishMyAccountAgeWitness(paymentAccount.getPaymentAccountPayload());
+        paymentAccountManager.addPaymentAccount(paymentAccount);
     }
 
     public boolean onDeleteAccount(PaymentAccount paymentAccount) {
```
chimp1984, line 97: I think we should change the API here to not return a boolean but to add a 'canRemove()' method at the caller and only call remove if that returns true.

core/src/main/java/bisq/core/exceptions/NotFoundException.java
blabno: This is general purpose exception that can be applied to any missing entity. The caller should know what is the main entity referenced by method.

core/src/main/java/bisq/core/payment/PaymentAccountManager.java
blabno: No! The method must not make any assumptions that it is being called in happy path scenario.
Exceptions are exactly for this type of situations where something else needs to be signaled then the happy path result.

desktop/src/main/java/bisq/desktop/main/account/content/altcoinaccounts/AltCoinAccountsDataModel.java
blabno: Because this method is going to be reused also by the API. I can create 2 methods, one working with PaymentAccount and the other with working with id and delegating to the first one after the lookup.
However I thought it's better to have less code, especially that the performance penalty is insignificant.

desktop/src/main/java/bisq/desktop/main/account/content/altcoinaccounts/AltCoinAccountsDataModel.java
blabno: I think it's better to propagate the exception and handle it on the view layer, but wanted to do as little changes as possible in order not to overwhelm maintainers.

core/src/main/java/bisq/core/payment/PaymentAccountManager.java
chimp1984: I disagree. I think the method removePaymentAccount should do what its name suggests and not throw an unchecked exception if the valid case happens that the user tries to remove an account but has an offer open using it. 
Bisq is using exceptions for exceptional cases not for flow control. I think we should stick with that patter and not introduce tons of exceptions for all kind of checks.

desktop/src/main/java/bisq/desktop/main/account/content/altcoinaccounts/AltCoinAccountsDataModel.java
chimp1984: I prefer to have then 2 methods if you need that or just do the lookup in your client code.

chimp1984: We do not use ids as params for lookup if the object is available anyway.

core/src/main/java/bisq/core/exceptions/ValidationException.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Copy of ValidationException from javax.validation:validation-api
+ */
+public class ValidationException extends RuntimeException {
+    public ValidationException(String message) {
+        super(message);
+    }
+
+    public ValidationException() {
```
lusarz, line 28: Only first one constructor with `String message` as param is used anywhere in application.

core/src/main/java/bisq/core/payment/PaymentAccountInUseException.java
```diff
@@ -0,0 +1,24 @@
...
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.core.payment;
```
lusarz: I wonder if it's appropriate place for this Exception class. What about `bisq.core.payment.exceptions` ? 

Although I don't see consistency in project codebase - some exceptions are [inside](https://github.com/blabno/exchange/blob/2ad86abe9fc3ea802c95d7c25d1195eacb587d60/core/src/main/java/bisq/core/dao/exceptions/DaoDisabledException.java) `exceptions` package, others [no](https://github.com/blabno/exchange/blob/d55114e019fda469266dab50f63916289b2e71cb/core/src/main/java/bisq/core/support/dispute/DisputeAlreadyOpenException.java).

core/src/main/java/bisq/core/payment/PaymentAccountManager.java
```diff
@@ -0,0 +1,128 @@
...
+        this.tradeManager = tradeManager;
+        this.user = user;
+    }
+
+    public PaymentAccount addPaymentAccount(PaymentAccount paymentAccount) {
+        if (paymentAccount instanceof CryptoCurrencyAccount) {
+            CryptoCurrencyAccount cryptoCurrencyAccount = (CryptoCurrencyAccount) paymentAccount;
+            TradeCurrency tradeCurrency = cryptoCurrencyAccount.getSingleTradeCurrency();
+            if (tradeCurrency == null) {
+                throw new ValidationException("CryptoCurrency account must have exactly one trade currency");
```
lusarz, line 70: Shouldn't `cryptoCurrencyAccount.getSingleTradeCurrency()` throws exception instead returning `null` ?

I'm just curious your opinion - such a change would require changes in many places, and it's out of scope of current pull request.

core/src/main/java/bisq/core/payment/PaymentAccountManager.java
lusarz: I see this exception is handled [here](https://github.com/blabno/exchange/blob/a98cf1b9e84ceca658692547a92b0ec45653460b/desktop/src/main/java/bisq/desktop/main/account/content/fiataccounts/FiatAccountsDataModel.java#L100-L107). Maybe `removePaymentAccount` method can return boolean directly ?

core/src/main/java/bisq/core/payment/PaymentAccountManager.java
```diff
@@ -0,0 +1,128 @@
...
+            }
+            altCoinAddressValidator.setCurrencyCode(tradeCurrency.getCode());
+            InputValidator.ValidationResult validationResult = altCoinAddressValidator.validate(cryptoCurrencyAccount.getAddress());
+            if (!validationResult.isValid) {
+                throw new ValidationException(validationResult.errorMessage);
+            }
+        }
+//        TODO we should validate payment account here as well
```
lusarz: Is it out of scope of current pull request ?

desktop/src/main/java/bisq/desktop/main/account/content/fiataccounts/FiatAccountsDataModel.java
```diff
@@ -50,30 +45,23 @@
...
-    private final AccountAgeWitnessService accountAgeWitnessService;
     final ObservableList<PaymentAccount> paymentAccounts = FXCollections.observableArrayList();
     private final SetChangeListener<PaymentAccount> setChangeListener;
     private final String accountsFileName = "FiatPaymentAccounts";
     private final PersistenceProtoResolver persistenceProtoResolver;
     private final CorruptedDatabaseFilesHandler corruptedDatabaseFilesHandler;
 
     @Inject
-    public FiatAccountsDataModel(User user,
+    public FiatAccountsDataModel(PaymentAccountManager paymentAccountManager, User user,
```
lusarz, line 57: Seems like `FiatAccountsDataModel` and `AltCoinAccountsDataModel` are very similar, especially after changes in this pull request. Shouldn't we introduce some common abstraction for them ?

lusarz (changes requested):

core/src/main/java/bisq/core/payment/PaymentAccountManager.java
chimp1984: Policy in Bisq is to use exceptions for exceptional cases. If a field can be null we use null checks or `Optional` but we do not throw an exception. Only if a fields is expected to be not null we use `checkNotNull` which throws an exception. I know there are different philosophies how to deal with exceptions but I think it is not a good idea to start to mix 2 different philosophies when the projects is using already one.

core/src/main/java/bisq/core/payment/PaymentAccountManager.java
chimp1984: > Maybe removePaymentAccount method can return boolean directly ?

That would be consistent with classical `remove` methods which return a boolean as well about the success of the intended remove. In that case the client (e.g. UI) need to know why it was not possible. For that I recommend using custom methods.

desktop/src/main/java/bisq/desktop/main/account/content/fiataccounts/FiatAccountsDataModel.java
chimp1984: Yes, might be a refactor candidate if fuctionality has duplications there.

ripcurlx (changes requested):
NACK - could you please address the comments by @lusarz and @chimp1984 and resolve conflicts with master. Thanks!

desktop/src/main/java/bisq/desktop/main/account/content/fiataccounts/FiatAccountsDataModel.java
blabno: Yes, but let's keep this PR focused and small.

core/src/main/java/bisq/core/payment/PaymentAccountManager.java
blabno: Not using exceptions for flow control? What should they be used for then?

desktop/src/main/java/bisq/desktop/main/account/content/altcoinaccounts/AltCoinAccountsDataModel.java
blabno: >We do not use ids as params for lookup if the object is available anyway.

That's cool for desktop app, but as soon as you get any kind of API you will have to reference stuff by id.

chimp1984: Then use 2 methods with diff. params. Forcing other clients to go down to id just to get back the object what is already available is not good.

core/src/main/java/bisq/core/payment/PaymentAccountManager.java
chimp1984: > Not using exceptions for flow control? What should they be used for then?

This is considered bad practice:
https://duckduckgo.com/?q=java+exceptions+for+flow+control&t=ffab&ia=web

chimp1984: > What should they be used for then?

For cases which happens exceptionally and sometimes lack of a clear way to handle it, so you delegate it to higher level classes how to deal with it. Writing to disk but get a IOError for instance. This use case is a valid circumstance that a user want to remove an account but does not know that there is a policy that he cannot remove it if used in offers or trades. This is not exceptional. I guess you want to provide information for the caller about the reason why the remove was not successful in the exception. You could also create a resultObject which contains diff. reuslt types and meta data if needed. Or just let the client call some validation methods like `isAccountUsedInOpenOffers` which is the usual way how Bisq handles such cases.

blabno: So instead of of throwing Exception we should return some Object that either contains happy path result, or some additional data in case of failure?
Exceptions are much cleaner.

desktop/src/main/java/bisq/desktop/main/account/content/altcoinaccounts/AltCoinAccountsDataModel.java
blabno: Ok. Though I will get another comment from @lusarz that method is not used.

core/src/main/java/bisq/core/payment/PaymentAccountManager.java
chimp1984: I give up.... 
NACK to the PR

chimp1984 (changes requested):
NACK, see comments. Bisq does not use exceptions for flow control which is considered by many developers as anti-pattern. Even if you tolerate that view it is bad to mix 2 different "philosophies" in a project.

core/src/main/java/bisq/core/payment/PaymentAccountManager.java
blabno: Yes, give up, cause you don't have better arguments. When there are many reasons why a method might fail then the typed exceptions are the way to inform other component about the reason.
Strings are good for humans, but it's not only the humans that interact with this business logic component.

lusarz: Why strings? You can create some kind of OperationResult class, like [ValidationResult](https://github.com/bisq-network/bisq/blob/049655b6857449aaca789c9cee5358daf0a54419/core/src/main/java/bisq/core/util/validation/InputValidator.java#L37-L64).

You'll be able to map it to appropriate response in API.

blabno: @lusarz do you honestly use OperationResult for every non-void method that you write?

blabno:
@ripcurlx comments applied although I don't agree with some of them at all.

core/src/main/java/bisq/core/payment/PaymentAccountManager.java
lusarz: No, and I don't have strong opinion what is better - Exceptions or ResultObject. I usually use Exceptions, but now I see a lot of opinions in net that recommends ResultObject. I think this approach might be interesting. Client code with a lot of `catch` blocks doesn't looks too nice. 

Just give it a try, maybe you'll find some pros :)

Also consistency argument is relevant here. Validation in bisq [(for example here)](https://github.com/bisq-network/bisq/blob/049655b6857449aaca789c9cee5358daf0a54419/core/src/main/java/bisq/core/util/validation/InputValidator.java#L26-L28) uses ResultObject instead of throwing typed exceptions.

blabno: Validation is completely different case. When you call `validate` method the main thing you want to know is the result of the validation.
Look at classic example of ArrayList: http://www.docjar.com/html/api/java/util/ArrayList.java.html#441

Does remove method return RemoveMethodResult? No it returns the removed object, which is the main intended thing. In order to tell user that object cannot be removed, because index is wrong it uses IndexOutOfBoundsException.

lusarz: Look at ArrayList [30 lines lower](http://www.docjar.com/html/api/java/util/ArrayList.java.html#472), it has `remove(Object o)` method. Does it throw `NotFoundException` ? No, it returns false.

chimp1984:
I don't have time for another review. I leave review to other devs and if they ACK my NACK can be ignored.

PR #3587, opened 2019-11-09
(9/9) [PR COMMENTS] Split up P2PDataStoreTest

julianknutsen:
New commits start eb2f8f3

Review of #3554 asked for some iteration on the number of different tests that all existed in one file. To that end, this PR splits up the tests into multiple files, cleans up the common helper functions, and adds Javadoc comments on all of the test fixtures to help future reviewers understand how the tests work.

I would prefer to just put this PR on the end of the rest since it is strictly limited to test code. That will save a lot of the hassle of having to merge the various iterations of tests throughout the refactor effort across the new files.

I think the end result is in a good place and there is a clean separation between the common test helpers and the various entry points.


chimp1984 (dismissed):
utACK
I did some smoke tests but as that is high risk area we should wait for an ACK from @freimair as well.

freimair (approved):
Ack

ripcurlx (approved):
utACK

ripcurlx:
I don't merge this PR as the commits will be part of https://github.com/bisq-network/bisq/pull/3609 anyways. So I'll close it as soon as #3609 gets merged. In the future I might do it differently as I do it right now. If we have bigger changes that rely on each other and are pushed as separate PRs, it might be better to only merge the final PR (squash and merge) and use the other PRs for initial feedback. So we don't end up with in-progress commits in our git history. Still the title of the final PR will need some change as it should sum up the whole change. Within the descriptions we should reference all other closed PRs for documentation purpose.
@freimair @julianknutsen @sqrrm What do you think?

julianknutsen:
From the author's side of things, each PR could have been independently merged and it would not have affected product stability. For cases like that AND if PRs are handled within a day or so then it allows iterative development and commits to master that can make smaller improvements that other devs will pull. You get some additional free "testing" although this should never be the sole reason to do it. The major benefit is any development on the same files or modules can be merged easier because you don't have 50 commits that may have merge conflicts.

If we expect longer cycles, I think just merging the last PR may make sense. This is similar to a feature branch merge, but you still have the benefit of going back to the second parent to see the individual commits in the event of bugs. With squash-merging, it all shows up as one huge changeset and it is hard to dig through.

After reading more about the different merge options available, I think I prefer merge-commits over squash-and-merge whenever possible. The ability to track a change back to the individual commits is useful for bisecting and understanding the original intention of a specific fix. If it is hard for devs to read `git log` I think `git log --first-parent` will give you what you want. It just removes the second parent from the merge commits so you just see the individual PRs.

I'm interested to hear everyone else's thoughts on the benefits of squash-and-merge so I can see if that would make my workflow any better. There may be something I am missing since I don't have a ton of experience with Git.

PR #3588, opened 2019-11-09
Updating instructions for trading XMR (Monero)

rating89us:
- clarifying that official Monero wallets (GUI or CLI) are NOT required to send XMR, since user can use alternative wallets that provide the transaction key, transaction ID and destination address (like desktop wallets Exodus and MyMonero, or mobile wallets like Cake Wallet) and do the verification outside the wallet (Explore Monero website or XMR checktx tool).
- listing Monero GUI as first option (for common users), and Monero CLI as second option (for advanced users)
- to get transaction key in Monero GUI, user must go to Transactions tab (P button is only used to generate payment proof)
- adding Explore Monero website (https://www.exploremonero.com/receipt) as alternative to proof/verify payments
- using _transaction key_ and _transaction ID_, instead of _tx private key_ and _transaction hash_, which are the terms used in the official Monero wallets
- removing payment ID instructions, since it is being deprecated at the end of November (v0.15)
- directing to subreddit Monero support (https://www.reddit.com/r/monerosupport/), which is actively maintained, instead of Monero forum (https://forum.getmonero.org).

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -1226,25 +1226,24 @@ If you are not sure about that process visit ArQmA discord channel (https://disc
 or the ArQmA forum (https://labs.arqma.com) to find more information.
 account.altcoin.popup.xmr.msg=Trading XMR on Bisq requires that you understand and fulfill \
 the following requirements:\n\n\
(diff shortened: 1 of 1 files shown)


ripcurlx:
@rating89us Could you please squash it into one commit?

PR #3596, opened 2019-11-11
Updating trading instructions for XMR (Monero)

rating89us:
Prove payments:
- clarifying that the official wallets (Monero GUI or CLI) are NOT required to send XMR, since user can use some alternative wallets (desktop: MyMonero, Exodus / mobile: Cake Wallet, MyMonero, Monerujo) that provide the information required to prove a payment (transaction key, transaction ID and destination address). This information is not provided by the remaining desktop and mobile wallets that currently support Monero, but since they could provide it in the future, I kept the warning about using other wallets different from the previously mentioned.
- listing Monero GUI as first option (for most users), and CLI as second option (for advanced users)
- using "transaction ID", which is used in official wallets, and keeping "transaction hash" as synonymous, which is used in XMR checktx tool
- using "transaction key", which is used in official wallets, and adding synonymous like "Secret key", which is used in MyMonero wallet
- adding "destination address" and keeping  "recipient's address" as synonymous
- renaming History tab to Transactions tab in Monero GUI
- adding "save recipient address" option must be enabled in Cake Wallet settings
- add warning to never delete or restore a Monero wallet before a Bisq trade is completed (since the transaction key and the destination address are stored in the Monero wallet software, and they cannot be recovered in the Monero blockchain)

Check payments:
- adding Monero GUI must be in Advanced mode
- adding Monero GUI verification must be done in Check Transaction section of Prove/check page
- adding parameters TXID TXKEY ADDRESS to the command check_tx_key in Monero CLI, as instructions in (https://www.getmonero.org/resources/user-guides/prove-payment.html)
- adding Explore Monero website (https://www.exploremonero.com/receipt) as alternative to verify payments
- removing payment ID instructions (it is being deprecated at the end of November in v0.15)

More info:
- directing to subreddit Monero support (https://www.reddit.com/r/monerosupport/), which is actively maintained, instead of Monero forum (https://forum.getmonero.org).

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -1227,25 +1227,33 @@ If you are not sure about that process visit ArQmA discord channel (https://disc
 or the ArQmA forum (https://labs.arqma.com) to find more information.
 account.altcoin.popup.xmr.msg=Trading XMR on Bisq requires that you understand and fulfill \
 the following requirements:\n\n\
-For sending XMR, you need to use either the official Monero GUI wallet or Monero CLI wallet with the \
-store-tx-info flag enabled (default in new versions). Please be sure you can access the tx key as \
-that would be required in case of a dispute.\n\
-monero-wallet-cli (use the command get_tx_key)\n\
-monero-wallet-gui (go to history tab and click on the (P) button for payment proof)\n\n\
-In addition to XMR checktx tool (https://xmr.llcoins.net/checktx.html) verification can also be accomplished in-wallet.\n\
-monero-wallet-cli : using command (check_tx_key).\n\
-monero-wallet-gui : on the Advanced > Prove/Check page.\n\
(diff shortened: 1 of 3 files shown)


ripcurlx:
@niyid @woodser Could you give an ACK if this Monero info update is correct?

niyid:
> @niyid @woodser Could you give an ACK if this Monero info update is correct?

OK. @woodser ?

Regards.

woodser:
Looks good to me.

ripcurlx:
@rating89us I had to extend your PR to make the content fit within our Popup. It exceeded the max character length we were supporting until now. I changed the window width for Monero and extended the max allowed character length.
**Before**
Bildschirmfoto 2019-11-14 um 15 07 05
**After**
Bildschirmfoto 2019-11-14 um 15 22 00

rating89us:
Looks good to me.

ripcurlx (approved):
ACK

Issue #3598, opened 2019-11-12
Not able to create moneygram account 


beingindot:
### Description

Not able to create moneygram account. save new account button is not enabled after filling all required fields.

#### Version
1.2.3

### Steps to reproduce
1. create new account
2. select payment method as moneygram
3. fill required fileds
4. save new account button not enabling

### Expected behaviour
We could save the moneygram account

### Actual behaviour
For moneygram, save new account not enabling

### Screenshots
![Screenshot from 2019-11-12 13-33-15](https://user-images.githubusercontent.com/55691179/68653135-099d1200-0551-11ea-95e5-cb0be0236c6d.png)

#### Device or machine
Ubuntu 18.04

beingindot:
working on it.

PR #3600, opened 2019-11-12
Added padding for flowpane to resolve moneygram save issue.

beingindot:
For https://github.com/bisq-network/bisq/issues/3598

Basically flowpane is hiding email label, hence we could not save. so 
@ripcurlx suggested this workaround fix for immeidate push.


ripcurlx (dismissed):
~ACK~ NACK - as the commit needs to be signed.

ripcurlx:
@beingindot ACK on the code change - NACK- you need to sign your commits

ripcurlx (changes requested):
NACK - as the commit needs to be signed

ripcurlx (approved):

PR #3608, opened 2019-11-14
(12/12) [BUGFIX] Remove from protectedDataStore in onDisconnect path

julianknutsen:
New commits start: b10a603

Fixes a bug in the P2PDataStore::onDisconnect path where a PersistablePayload would not be removed from the protectedDataStore.


chimp1984 (commented):
utACK

freimair (approved):
utAck

ripcurlx (approved):
utACK - I don't merge this PR as the commits will be part of #3609 anyways. So I'll close it as soon as #3609 gets merged.

PR #3611, opened 2019-11-15
Refactorings for api support

chimp1984:
For API access we need to move code for creating an offer to the core module.
https://github.com/chimp1984/bisq/tree/grpc is based on that code. This PR does not add anythign API related just prepares the required refactoring.


freimair (approved):
Ack - ie. passed a small smoke test. Changes look good though.

ripcurlx (approved):
ACK - Tested trading and offer editing on Regtest and everything still seems to work as is. Code looks fine as well.

PR #3612, opened 2019-11-15
Fix bug with mutating account after saving 

chimp1984:
Fixes https://github.com/bisq-network/bisq/pull/3572

We saved the account and afterwards called setSelectedTradeCurrency.
That caused a bug with remove as the hash at insert was different as at
remove.


freimair:
I recon that this also fixes #3461?

ripcurlx (approved):
utACK

chimp1984:
> I recon that this also fixes #3461?

Yes. I have not tested it but I am pretty sure.

Issue #3537, opened 2019-10-31
possiblePaymentAccounts.isEmpty()


bryanlivingston:
Got the error in the title when trying to take order to sell btc.
Log is attached.
[bisq.log](https://github.com/bisq-network/bisq/files/3795727/bisq.log)

Criatura83:
I got the same error when trying to take an order to buy btc.
Log attached.

[bisq.log](https://github.com/bisq-network/bisq/files/3796453/bisq.log)

fflo:
ditto

Manew-des:
I have the same problem, any words guys ?

mattmill30:
Is the issue related to the new "signed accounts" feature for bank accounts?

The issue doesn't occur when purchasing against my Faster Payments (GB) account, but that account doesn't require signing.

I'm wondering whether the situation is along the lines of:
TakeOfferDataModel.java (line 194): ObservableList<PaymentAccount> possiblePaymentAccounts = getPossiblePaymentAccounts();

v1.1.7 - Lists all offers for National Bank Account and the possiblePaymentAccounts variable is populated, but if Account isn't mature, will error when making a purchase with "account created after 1st March 2019".

v1.2.2 - Lists all offers for National Bank Account and the possiblePaymentAccounts variable is populated with _only signed_ accounts which may enumerate to 0. Then when attempting to make a purchase using the TAKE OFFER TO BUY BTC button, the TakeOfferDataModel.java (line 195) exception is thrown.

Just my two Satoshis.

[bisq - possiblePaymentAccounts.isEmpty.log](https://github.com/bisq-network/bisq/files/3814878/bisq.-.possiblePaymentAccounts.isEmpty.log)

Perlover:
Same error, Bisq 1.2.3, Linux (official deb package)
My accoung is signed but not aged

I clicked by offer (The Tab "Sell bitcoin" - i wanted to sell bitcoin, filter BTC/EUR Revolut)
The destination offer was in range sum (0.05-0.1 BTC). Destination account is signed too (i see now 19 days ago). My account has 0.02 BTC (but a button of order was actived and i clicked). I got `isEmpty()`

Here is logs:

```
нояб.-15 15:24:35.407 [JavaFX Application Thread] INFO  b.d.app.BisqApp: System load: Memory (MB): 819 / No. of threads: 142 
[...]

PR #3619, opened 2019-11-16
Fixes #3601 #3537 and Add colon after paymentMethod on tooltip

### LINKED ISSUE #3601: Bisq - Error: possiblePaymentAccounts.isEmpty()
-->

### Description

 When clicking on a "suitable" offer (buy BTC) error msg appeares....

'Error: possiblePaymentAccounts.isEmpty()'

It is reproduceable on two different pc's (fresh installation) when chosing the same offer. Picking another offer (sometimes) works.

### Screenshots

screenshots and log file attached

[bisq_log.txt](https://github.com/bisq-network/bisq/files/3838151/bisq_log.txt)
issue screenshot

#### Device or machine

x86 and x64

rafaelpac:
This is a silly fix of the tooltip. Every item on the tooltip have colons, some added with getWithCol and some hardcoded on the string. This one doesn't have.

### DIFF
--- a/core/src/main/java/bisq/core/payment/PaymentAccountUtil.java
+++ b/core/src/main/java/bisq/core/payment/PaymentAccountUtil.java
@@ -65,7 +65,7 @@ public class PaymentAccountUtil {
                                                 AccountAgeWitnessService accountAgeWitnessService) {
         boolean hasChargebackRisk = PaymentMethod.hasChargebackRisk(offer.getPaymentMethod(), offer.getCurrencyCode());
         boolean hasValidAccountAgeWitness = accountAgeWitnessService.getMyTradeLimit(paymentAccount,
-                offer.getCurrencyCode(), offer.getMirroredDirection()) >= offer.getAmount().value;
+                offer.getCurrencyCode(), offer.getMirroredDirection()) >= offer.getMinAmount().value;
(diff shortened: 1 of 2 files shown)


rafaelpac:
Sorry, I made a little confusion...
My first change was a silly change to add a colon missing in a tooltip.
This was my first ever pull request, I am a total noob with github.
Later, I think I found the bug that causes issues #3601 #3537 
The bug only happens when the offer has min-max value range.
The new `isAmountValidForOffer` in `PaymentAccountUtil.java` consider only the max amount, but the button to take the offer is ungrayed considering the min amount, which causes the bug.
Both fixes came together here, I don't really know how should I have done it to make them separate.

chimp1984:
@rafaelpac 
The second commit needs to be removed from that PR as it is a completely different issue.
You can close the PR and make a new one with the colon fix. The second one requires @ripcurlx or @sqrrm to verify.

chimp1984 (changes requested):
NACK
The 2nd commit needs to be in a separate PR

rafaelpac:
Yes, I'm sorry... I know there should be two separate pull requests, but I can't find out how to do it. I need separate branches in my fork, is it? I am afraid of clicking more and make it worse...

rafaelpac:
Ok, I will close this and open two separate ones. I think I know how to do it now.

PR #3621, opened 2019-11-17
(2/6) Remove onBatch interfaces in P2PDataStore

julianknutsen:
Patches start 78211a6

Before:
update-lists-perf

After:
update-lists-perf-after

A bug was introduced in #3148 where ProposalListPresentation would incorrectly call `updateLists()` when the proposal state didn't change. This pull request should help #3143 (the target of #3148) in a more maintainable way that removes the bug and UI performance issues by calling `updateLists()` less often. There is still an outstanding bug that `updateLists()` can take **_seconds_** to run on each invocation.

**Initial Bug #3143**
If the `P2PDataStore` removes multiple items during the expiration work, each item is signaled separately. The `ProposalService` is a listener for these removes and would eventually call `updateLists()` for each remove.

__Exacerbating Issue__
Old TempProposals were not correctly deleted from disk. They would be recreated at startup and the expire code would always remove them. This was fixed with #3150.

**Follow Up Bug #3148**
The fix for the initial bug was to create a new listener interface on `P2PDataStore` that would signal when the expire code starts and ends. The `ProposalListPresentation` would subscribe to this interface and remove its listener from the ProposalService so that all of the extra `onRemoved()` calls would not trigger `updateLists()`.

The code did remove the proposal listener when the expire code started, but it had a bug where it would **_ALWAYS_** call `updateLists()` when the expire code finished if **_ANY_** `TempProposal` object was received (not necessarily a new one).

In the pre-fix code, the `ProposalListPresentation` code only called `updateList()` when the `ProposalService` proposal list changed. The `ProposalService` had validation to ensure that it only signaled listeners if the `Proposal` was new. By going around the `ProposalService` and directly listening to the `P2PDataStore`, the new code was running every time an existing `Proposal` object was seen.

Duplicate `Proposal` objects (same payload different sequence number) can be added during startup, on `restart()`, or republished on startup by owners `MyProposalListService::rePublishMyProposalsOnceWellConnected()`. The effect is that the next call to the expiration code would cause `updateLists()` to run when it didn't need to.

**Fix**
The root of the original performance issue is straightforward. Don't update listeners of the `P2PDataStore` or the `ProposalService` on each remove when batching them is possible.

This pull request implements batching by:

- Modifying the `onAdded()` and `onRemoved()` listener functions to take a `Collection` of `ProtectedStorageEntry` objects instead of just one.
- Combining all `remove()` operations done in `P2PDataStore::removeExpiredEntries()` to a single `onRemoved()` call that includes all of the removed `ProtectedStorageEntry` objects.
- Using this new behavior in `ProposalService` to only update its listeners once for each `onRemoved()` call.
- Reverting the `onBatch` listener changes

The end result is that any number of removes from `P2PDataStore::removeExpiredEntries()` will result in 1 call to `ProposalListPresentation::updateLists()` which was the original intention of the fix.

It also cleans up one of the existing code smells for the P2PDataStorage layer that requires listeners to understand if and when `removeExpiredEntries()` is called and if they should do something special.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/P2PModule.java
+++ b/p2p/src/main/java/bisq/network/p2p/P2PModule.java
@@ -105,5 +105,6 @@ public class P2PModule extends AppModule {
         bindConstant().annotatedWith(named(NetworkOptionKeys.MSG_THROTTLE_PER_10_SEC)).to(environment.getRequiredProperty(NetworkOptionKeys.MSG_THROTTLE_PER_10_SEC));
         bindConstant().annotatedWith(named(NetworkOptionKeys.SEND_MSG_THROTTLE_TRIGGER)).to(environment.getRequiredProperty(NetworkOptionKeys.SEND_MSG_THROTTLE_TRIGGER));
         bindConstant().annotatedWith(named(NetworkOptionKeys.SEND_MSG_THROTTLE_SLEEP)).to(environment.getRequiredProperty(NetworkOptionKeys.SEND_MSG_THROTTLE_SLEEP));
+        bindConstant().annotatedWith(named("MAX_SEQUENCE_NUMBER_MAP_SIZE_BEFORE_PURGE")).to(1000);
     }
 }
--- a/core/src/main/java/bisq/core/alert/AlertManager.java
+++ b/core/src/main/java/bisq/core/alert/AlertManager.java
@@ -44,6 +44,8 @@ import java.security.SignatureException;
 
 import java.math.BigInteger;
 
+import java.util.Collection;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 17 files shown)


freimair (dismissed):
utAck

ripcurlx:
> utAck

Commit https://github.com/bisq-network/bisq/pull/3628/commits/833611f1df5e9cd886d82c49bd7d597736d062a3 is also part of this PR https://github.com/bisq-network/bisq/pull/3628#pullrequestreview-318859217

p2p/src/main/java/bisq/network/p2p/P2PModule.java
```diff
@@ -105,5 +105,6 @@ protected void configure() {
         bindConstant().annotatedWith(named(NetworkOptionKeys.MSG_THROTTLE_PER_10_SEC)).to(environment.getRequiredProperty(NetworkOptionKeys.MSG_THROTTLE_PER_10_SEC));
         bindConstant().annotatedWith(named(NetworkOptionKeys.SEND_MSG_THROTTLE_TRIGGER)).to(environment.getRequiredProperty(NetworkOptionKeys.SEND_MSG_THROTTLE_TRIGGER));
         bindConstant().annotatedWith(named(NetworkOptionKeys.SEND_MSG_THROTTLE_SLEEP)).to(environment.getRequiredProperty(NetworkOptionKeys.SEND_MSG_THROTTLE_SLEEP));
+        bindConstant().annotatedWith(named("MAX_SEQUENCE_NUMBER_MAP_SIZE_BEFORE_PURGE")).to(10000);
```
ripcurlx: Same concern as in https://github.com/bisq-network/bisq/pull/3628/commits/833611f1df5e9cd886d82c49bd7d597736d062a3#r347789865

ripcurlx (commented):
Besides https://github.com/bisq-network/bisq/pull/3621#pullrequestreview-318868875 everything looks good.

p2p/src/main/java/bisq/network/p2p/P2PModule.java
julianknutsen: Yes, this was intended to just be a refactor, not a policy change. An extra 0 was added accidentally during the move. Thanks for the keen eye.

ripcurlx (approved):
utACK

PR #3624, opened 2019-11-17
Add colon after "Payment method" on tooltip

rafaelpac:
This is a cosmetic fix of the tooltip over the payment method in the offer view tab. Every item on the tooltip have colons, some added with Res.getWithCol() and some hardcoded on the string. This one doesn't have.

image


freimair (approved):
utAck

ripcurlx (approved):
utACK

PR #3626, opened 2019-11-17
Avoid unchecked casts/assignments/calls

stejbac:
Improve type safety by avoiding most of the current unchecked operations, instead of simply suppressing the warnings (as is being done for almost all of them now). This mostly involves adding some missing arguments to raw types, using wildcards in a few extra places and making some minor changes to the logic (such as passing an explicit _Class\<T\>_ argument or replacing a generic vararg list with _List.of(..)_).

(In the case of _OfferBookChartView.updateChartData()_, I made a somewhat larger change to the method body. I also replaced some _ActivatableViewAndModel_ superclasses with _ActivatableView\<R, Void\>_ where this made sense.)

The test classes and some of the harder cases of potentially type-unsafe unchecked operations have been skipped for now, such as _bisq.common.storage.FileManager.read(File)_.


freimair (approved):
Ack. Thanks for cleaning that up!

ripcurlx (approved):
ACK

PR #3628, opened 2019-11-18
(1/6) [PR COMMENTS] Remove P2PDataStorage subclass in tests

julianknutsen:
Requested changes from #3568

Instead of using a subclass that overwrites a value, utilize Guice
to inject the real value of 10000 in the app and let the tests overwrite
it with their own.


p2p/src/main/java/bisq/network/p2p/P2PModule.java
```diff
@@ -105,5 +105,6 @@ protected void configure() {
         bindConstant().annotatedWith(named(NetworkOptionKeys.MSG_THROTTLE_PER_10_SEC)).to(environment.getRequiredProperty(NetworkOptionKeys.MSG_THROTTLE_PER_10_SEC));
         bindConstant().annotatedWith(named(NetworkOptionKeys.SEND_MSG_THROTTLE_TRIGGER)).to(environment.getRequiredProperty(NetworkOptionKeys.SEND_MSG_THROTTLE_TRIGGER));
         bindConstant().annotatedWith(named(NetworkOptionKeys.SEND_MSG_THROTTLE_SLEEP)).to(environment.getRequiredProperty(NetworkOptionKeys.SEND_MSG_THROTTLE_SLEEP));
+        bindConstant().annotatedWith(named("MAX_SEQUENCE_NUMBER_MAP_SIZE_BEFORE_PURGE")).to(10000);
```
freimair: hasn't that been only 1000 (one thousand) previously? I guess the memory impact is marginal anyways but can you elaborate your decision?

freimair (changes requested):

p2p/src/main/java/bisq/network/p2p/P2PModule.java
julianknutsen: Yes, this was intended to just be a refactor, not a policy change. An extra 0 was added accidentally during the move. Thanks for the keen eye.

ripcurlx:
@freimair Could you please review the changes and resolve them afterwards?

freimair (approved):

PR #3630, opened 2019-11-19
Bump javafx version to 11.0.2

### LINKED ISSUE #2279: Wayland: 'GDK_IS_X11_DISPLAY (display)' failed

devinbileck:
Aside from various fixes and improvements, it appears that this
should resolve #2279, as fixed in
https://bugs.openjdk.java.net/browse/JDK-8210411

Reference:
https://github.com/javafxports/openjdk-jfx/blob/jfx-11/doc-files/release-notes-11.0.1.md
https://github.com/javafxports/openjdk-jfx/blob/jfx-11/doc-files/release-notes-11.0.2.md

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -46,6 +46,7 @@ configure(subprojects) {
         httpclientVersion = '4.5.3'
         ioVersion = '2.4'
         jacksonVersion = '2.8.10'
+        javafxVersion = '11.0.2'
         jcsvVersion = '1.4.0'
         jetbrainsAnnotationsVersion = '13.0'
         jfoenixVersion = '9.0.6'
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
ACK - @devinbileck Do you remember why we removed the checksums for javafx from the witness file? I think it was because of the different os jars which we somehow weren't able to implement properly in the witness file.

devinbileck:
Yes, as far as I remember that is correct. It would be nice to find a solution for that.

PR #3631, opened 2019-11-19
Fix package.bat version parsing

devinbileck:
The package.bat script parses the first 3 tokens of the version string
to determine the file version. This is done to strip "-SNAPSHOT" from
the version string since the file version can only contain a numeric
version number.


ripcurlx (approved):
utACK

PR #3634, opened 2019-11-19
Coin formatter interface

bodymindarts:
Last step in my ongoing effort to improve the design around the `BSFormatter` class. Possible next steps are to improve the related `BsqFormatter` as well. But that would be another task.

Tracking issue: https://github.com/bisq-network/bisq/issues/3240


freimair (approved):
Ack

PR #3636, opened 2019-11-19
(3/6) [BUGFIX] #3629 (Duplicate payloads w/ 20-byte and 32-byte keys)

### LINKED ISSUE #3629: TempProposal objects transmit on every startup when already present on starting node
As part of investigating #3610, I found another bug in how `ProtectedStorageEntry` objects are persisted to disk. The result is that a node starting up does not properly format the `GetData` request to the seed node leading to retransmission of `TempProposal` payloads that it already knows about.

I am already working on this area as part of bisq-network/proposals#140 and it will likely show up as a failing test. Documenting this here in case people see behavior they don't understand.

## Tech Info
`ProtectedPayload` objects implementing the `PersistablePayload` interface are saved to disk by the `Sha256Ripemd160hash` but are loaded from disk expecting the `Sha256Hash`. This results in the keys in the hashmap being 20-bytes for the items loaded from disk and 32-bytes for the items sent from the seed node.

So, the starting node asks the seednode to exclude the 20-byte key and it sends back the 32-byte key which both point to the exact same payload.

This is relevant for bisq-network/proposals#140 because we need to understand the behavior of non-upgraded nodes and their ability to propagate old `TempProposals` to the network.

More to come with how to address it.

julianknutsen:
New commits start at 849155a

fixes #3629

Hotfix Needed: No

There was a bug when reconstructing the in-memory HashMap for `ProtectedStoragePayload` objects implementing the `PersistablePayload` object. This had two related bugs:

1) The in-memory map would end up with duplicate Payloads if a node rebuilt a ProtectedStorageEntry from the 20-byte key, then received the 32-byte key from the seednode during initialization or from an update on the P2P network.

2) An initializing node would resync known payloads from the seednode since the `Payload` existed with both the 20-byte and 32-byte key.

This PR cleans up the usages so that all outward users see the 32-byte key and all internal users can use the existing 20-byte key for persistence. This design is much safer than transitioning all persistent users to 32-byte keys and writing on-disk upgrade code.

All external listeners of Payloads don't look at the key. So, although they will receive 2 onAdded() calls instead of 1, they do the right thing by excluding the second payload since it is identical. I don't think there is any reason to publish a hotfix for this.

## Upgrade behavior
**New Seednode <--> Old Client**
The seednode will only have 1 32-byte key for each unique payload, but the client node may still have a 20-byte key that it will send as a known key. The seednode will ignore this known key, send back the 32-byte key `Payload` and the client node will update its map with the 32-byte key. This mimics existing behavior.

**Old Seednode <--> New Client**
The seednode will have 1 20-byte key and 1 32-byte key for the same unique payload. The client will only have the 32-byte key and will send it is a known key. The seednode will transmit the 20-byte `Payload` back to the client who will ignore it.

**New Seednode <--> New Client**
Both the seednode and the client will only have 1 32-byte key for each unique payload. The client will include its 32-byte key as a known key which will match the key in the seednode map. No duplicate `Payload` objects will be transmitted.

### DIFF
--- a/core/src/main/java/bisq/core/alert/AlertManager.java
+++ b/core/src/main/java/bisq/core/alert/AlertManager.java
@@ -44,6 +44,8 @@ import java.security.SignatureException;
 
 import java.math.BigInteger;
 
+import java.util.Collection;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -79,22 +81,26 @@ public class AlertManager {
         if (!ignoreDevMsg) {
             p2PService.addHashSetChangedListener(new HashMapChangedListener() {
                 @Override
-                public void onAdded(ProtectedStorageEntry data) {
-                    final ProtectedStoragePayload protectedStoragePayload = data.getProtectedStoragePayload();
-                    if (protectedStoragePayload instanceof Alert) {
-                        Alert alert = (Alert) protectedStoragePayload;
-                        if (verifySignature(alert))
-                            alertMessageProperty.set(alert);
-                    }
+                public void onAdded(Collection<ProtectedStorageEntry> protectedStorageEntries) {
+                    protectedStorageEntries.forEach(protectedStorageEntry -> {
+                        final ProtectedStoragePayload protectedStoragePayload = protectedStorageEntry.getProtectedStoragePayload();
+                        if (protectedStoragePayload instanceof Alert) {
+                            Alert alert = (Alert) protectedStoragePayload;
+                            if (verifySignature(alert))
+                                alertMessageProperty.set(alert);
+                        }
+                    });
                 }
 
                 @Override
-                public void onRemoved(ProtectedStorageEntry data) {
-                    final ProtectedStoragePayload protectedStoragePayload = data.getProtectedStoragePayload();
-                    if (protectedStoragePayload instanceof Alert) {
-                        if (verifySignature((Alert) protectedStoragePayload))
-                            alertMessageProperty.set(null);
-                    }
+                public void onRemoved(Collection<ProtectedStorageEntry> protectedStorageEntries) {
+                    protectedStorageEntries.forEach(protectedStorageEntry -> {
+                        final ProtectedStoragePayload protectedStoragePayload = protectedStorageEntry.getProtectedStoragePayload();
+                        if (protectedStoragePayload instanceof Alert) {
+                            if (verifySignature((Alert) protectedStoragePayload))
+                                alertMessageProperty.set(null);
+                        }
+                    });
                 }
             });
         }
--- a/core/src/main/java/bisq/core/dao/governance/proposal/ProposalListPresentation.java
+++ b/core/src/main/java/bisq/core/dao/governance/proposal/ProposalListPresentation.java
@@ -20,15 +20,10 @@ package bisq.core.dao.governance.proposal;
 import bisq.core.btc.wallet.BsqWalletService;
 import bisq.core.dao.DaoSetupService;
 import bisq.core.dao.governance.proposal.storage.appendonly.ProposalPayload;
-import bisq.core.dao.governance.proposal.storage.temp.TempProposalPayload;
 import bisq.core.dao.state.DaoStateListener;
 import bisq.core.dao.state.DaoStateService;
 import bisq.core.dao.state.model.blockchain.Block;
 import bisq.core.dao.state.model.governance.Proposal;
-
-import bisq.network.p2p.storage.HashMapChangedListener;
-import bisq.network.p2p.storage.P2PDataStorage;
-import bisq.network.p2p.storage.payload.ProtectedStorageEntry;
 
 import bisq.common.UserThread;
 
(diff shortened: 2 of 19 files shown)


freimair (changes requested):
I might be missing something here so here are some control questions to foster my understanding:

> ## Upgrade behavior
> **New Seednode <--> Old Client**
> The seednode will only have 1 32-byte key for each unique payload, but the client node may still have a 20-byte key that it will send as a known key. The seednode will ignore this known key, send back the 32-byte key `Payload` and the client node will update its map with the 32-byte key. This mimics existing behavior.

- So in the real world the old client will send roughly 2MB of 20-byte known keys to the new seed node.
- The new seed node ignores all of them and respond with _all_ data (close to our single-message size limit of 10MB)? (since it only received 20-byte keys and no 32-byte key)
- Does the old client memorize the 32-byte keys? after restart?
 
> **Old Seednode <--> New Client**
> The seednode will have 1 20-byte key and 1 32-byte key for the same unique payload. The client will only have the 32-byte key and will send it is a known key. The seednode will transmit the 20-byte `Payload` back to the client who will ignore it.

- why does the old seednode have 32-byte keys?
- the client sends only 32-byte keys for its known payloads, so the seed node will send _all_ 20-byte addressed payloads to the client (roughly 10MB of data) to every client.
- every time a client connects. Because new clients will not memorize the 20-byte keys?

julianknutsen:
> I might be missing something here so here are some control questions to foster my understanding:
> 
> > ## Upgrade behavior
> > **New Seednode <--> Old Client**
> > The seednode will only have 1 32-byte key for each unique payload, but the client node may still have a 20-byte key that it will send as a known key. The seednode will ignore this known key, send back the 32-byte key `Payload` and the client node will update its map with the 32-byte key. This mimics existing behavior.

**_This behavior is only for `ProtectedStoragePayload` objects implementing `PersistablePayload`. For now, this is just `TempProposalPayload`._**

> 
> * So in the real world the old client will send roughly 2MB of 20-byte known keys to the new seed node.

**_No, the client sends roughly 1kb of unnecessary 20-byte keys at startup for the `TempProposalPayload` objects._**

> * The new seed node ignores all of them and respond with _all_ data (close to our single-message size limit of 10MB)? (since it only received 20-byte keys and no 32-byte key)

**_No, the seednode sends back around 50kb of duplicate `TempProposal` data in `GetDataResponse`._**

> * Does the old client memorize the 32-byte keys? after restart?

**_No, old clients use the 20-byte key for persistence. So after restart all `TempProposalPayload` objects
are reconstructed with 20-byte keys. It only "works" because the seed nodes send 32-byte keys at
startup that the client nodes use to update their internal map._**

> 
> > **Old Seednode <--> New Client**
> > The seednode will have 1 20-byte key and 1 32-byte key for the same unique payload. The client will only have the 32-byte key and will send it is a known key. The seednode will transmit the 20-byte `Payload` back to the client who will ignore it.
> 
> * why does the old seednode have 32-byte keys?

**_During initial creation of `TempProposalPayload` and during `MyProposalListService::rePublishMyProposalsOnceWellConnected()` the owners send them to the network. If these are "new", the seednode will save it as a 20-byte key in the `ProtectedDataStoreService`, but also keep it in-memory with the 32-bye key like the other `ProtectedStoragePayloads`. This allows it to give the 32-byte keys to new nodes on startup._**

> * the client sends only 32-byte keys for its known payloads, so the seed node will send _all_ 20-byte addressed payloads to the client (roughly 10MB of data) to every client.
> * every time a client connects. Because new clients will not memorize the 20-byte keys?

**_It isn't that the new clients won't memorize it, it is that they only have in-memory representations using 32-byte keys now. So, until the seednode is updated to ONLY have 32-byte keys and NOT have 20-byte keys in the internal map, the seednodes will still send back the 20-byte key payloads. It isn't harmful because the duplicate paylaods sent back are ignored from the new client._**

julianknutsen:
After a deeper investigation of the persistence code, I found that TempProposals are stored to disk as a List<ProtectedStorageEntry> and the Map.Entry<ByteArray, ProtectedStorageEntry> is actually reconstructed in-memory. So, I've updated e212240 accordingly with a much cleaner patch that just changes the 20-byte key to a 32-byte key.

ripcurlx:
@freimair Can you please review the changes and ACK if ok?

ripcurlx (approved):
utACK

freimair (approved):
Ack.

Now is the time to do the change, as only `TempProposalPayload` and the deprecated `TradeStatistics` are affected by the change. But it will generate additional network load if we are not quick in updating the seednodes.

julianknutsen:
I updated the bug as well but wanted to include it here.

After more investigation, 455f7d2 also fixes an additional issue where we were retransmitting every non-persistable `ProtectedStorageEntry` in the `GetUpdatedDataRequest` path.  By using the in-memory map instead of the protectedDataStore, we now stop transmitting duplicates. Here was the breakdown of the duplicate transmissions before that patch. This likely also caused additional bandwidth when seednodes were trying to sync with each other since they would be resending everything.

```
#################################################################
Received 724 instances
RefundAgent: 1 (915 B)
Filter: 2 (5.7 KB)
TempProposalPayload: 56 (33.9 KB)
MailboxStoragePayload: 457 (2.0 MB)
Alert: 1 (826 B)
Mediator: 2 (1.8 KB)
OfferPayload: 205 (249.6 KB)
################################################################# 
```

PR #3638, opened 2019-11-20
(4/6) Implement proper remove-before-add behavior in P2PDataStorage

julianknutsen:
New commits are at 526aee5

According to the original [design doc](https://docs.google.com/document/d/1kdZf1ZaPuj8MmtueBhv0f8ztvnfd8hDJwNO1_TWH4MM/edit#), removes-before-adds were expected and should have the behavior that a remove with a larger sequence number will block a future add with a lower sequence number.

The idea is that you don't want out of order remove messages to get lost otherwise you may do more work then necessary by adding a stale `ProtectedStorageEntry` when you could have ignored it entirely. I'm not sure of any existing cases, but this may have been the cause of 'flapping' or 'undeleted' messages as nodes were starting.

I've left the existing behavior w.r.t. broadcasting. Even if this `remove()` is the first message we have seen for a payload, we don't rebroadcast it to our peers.  I'm not exactly sure what we want here, but rebroadcasting, in this case, could help network propagation and we may want to adopt the invariant that **_ANY_** state updates caused by a message are worth of rebroadcasting because the same state updates may be relevant to our peers.

### DIFF
--- a/core/src/main/java/bisq/core/alert/AlertManager.java
+++ b/core/src/main/java/bisq/core/alert/AlertManager.java
@@ -44,6 +44,8 @@ import java.security.SignatureException;
 
 import java.math.BigInteger;
 
+import java.util.Collection;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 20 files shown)


ripcurlx:
> I've left the existing behavior w.r.t. broadcasting. Even if this `remove()` is the first message we have seen for a payload, we don't rebroadcast it to our peers. I'm not exactly sure what we want here, but rebroadcasting, in this case, could help network propagation and we may want to adopt the invariant that **_ANY_** state updates caused by a message are worth of rebroadcasting because the same state updates may be relevant to our peers.

@chimp1984 What do you think about that?

freimair (approved):
utAck

ad rebroadcasting even if not applicable: I had that feeling for a long time now. We have stale messages in the network. Thought about resending remove messages e.g. 3 times to spread the news even if a couple of clients missed the first one. Pretty much the same for every remove message. Since the seq number is touched, there is no risk of overloading the network.

my opinion: rebroadcast even if we did not yet receive the respective add. revert if something bad happens.

julianknutsen:
I'll just make an additional PR that adds rebroadcasting in the remove-before-add case. It will involve a bit of validation work that will conflict with my patch stack so I won't be able to get it out until tomorrow. A different PR will also allow the approved patches to go in without forcing them back to stale.

ripcurlx (approved):
utACK

PR #3639, opened 2019-11-20
Create modal pop-up with single button taking user to Support screen if there are active support tickets

### LINKED ISSUE #3401: Notification of pending arbitration issue

niyid:
Resolves #3401

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/MainViewModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/MainViewModel.java
@@ -17,6 +17,7 @@
 
 package bisq.desktop.main;
 
+import bisq.desktop.Navigation;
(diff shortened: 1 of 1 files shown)


niyid:
Pending issue: restrict DontShowAgainLookup only to "power users".

freimair (changes requested):
formatting is off.

freimair (changes requested):
formatting is still off

PR #3640, opened 2019-11-20
(8/8) Persist changes to ProtectedStoragePayload objects implementing PersistablePayload

julianknutsen:
New commits start at 66f71e5

With the addition of `ProtectedStorageEntrys`, there are now persistable
maps that have different payloads and the same keys. In the
`ProtectedDataStoreService` case, the value is the `ProtectedStorageEntry`
which has a `createdTimeStamp`, `sequenceNumber`, and `signature` that can
all change, but still contain an identical payload.

Previously, the service was only updating the on-disk representation on
the first object and never again. So, when it was recreated from disk it
would not have any of the updated metadata. This was just copied from the
append-only implementation where the value was the Payload
which was immutable.

This hasn't caused any issues to this point, but it causes strange behavior
such as always receiving seqNr==1 items from seednodes on startup. It
is good practice to keep the in-memory objects and on-disk objects in
sync and removes an unexpected failure in future dev work that expects
the same behavior as the append-only on-disk objects.


p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -418,9 +418,8 @@ public boolean addProtectedStorageEntry(ProtectedStorageEntry protectedStorageEn
 
         // Persist ProtectedStorageEntrys carrying PersistablePayload payloads and signal listeners on changes
         if (protectedStoragePayload instanceof PersistablePayload) {
-            ProtectedStorageEntry previous = protectedDataStoreService.putIfAbsent(hashOfPayload, protectedStorageEntry);
-            if (previous == null)
-                protectedDataStoreListeners.forEach(e -> e.onAdded(protectedStorageEntry));
+            protectedDataStoreService.put(hashOfPayload, protectedStorageEntry);
```
freimair, line 429: This means that we will have a lot more save procedures. Looking at the code in `FileManager:75` ff. I feel not confident to do this change now as we significantly up the risk of having corrupted data stores - which is very bad.

Just by a quick look at the filemanager code, I doubt that this is thread-safe. There seems to be an obvious race condition in 79-83. Can we have that tested and fixed if needed before we apply this patch?

freimair (changes requested):

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
julianknutsen: I'm not sure I understand the full concern. These objects use the same path as all of the other persistable objects that are written on trades, account signings, etc. Are you suggesting we hold the release until FileManager.java:75 is looked into?

A ConcurrentHashMap is used in all the TempProposal code, but I think your comments are more related to the fact that all FileManager tasks aren't threadsafe regardless of the `Storage` data structures. Is that right?

freimair: > I'm not sure I understand the full concern.

The storage stuff is most probably broken and we had corrupted databases in the past (which are very hard to recover from especially because the max message size is nearly reached and seednodes/clients have a hard time sending/receiving the whole 10MB of data). If we increase the number of writes (by persisting stuff everytime something we already have gets updated) we up the chance of hitting that race condition. Nothing more nothing less.

> Are you suggesting we hold the release until FileManager.java:75 is looked into?

no, there is no need to hold the release. The broken code is shipped already, so there is nothing we can do about that. If you mean holding the PR by "hold the release" until this is fixed: yes.

> aren't threadsafe regardless of the Storage data structures. Is that right?

yes, correct. I believe the storage stuff needs a closer look and fast and with high priority.

chimp1984: I agree the FileManager would need some love. I am not aware of many file corruption issues and there is some routines to handle that: Move corrupted file to a backup folder and recreate it. If recreating is problematic depends on the file. Some files are then taken from resources and just require the delta update (trade stat, account age). Others like PendingTrades, Disputes etc. are more problematic as its user data. But there are rolling backups and so far I am not aware that it ever causes a serious problem. 
More problems happen with corrupted wallet files but that is done inside BitcoinJ (from where I borrowed most of the persistence code base).
So conclusion: Yes would be good to get the persistence code thread safe and improved. But I don't consider that we have lots of problems in that area atm (have not looked into the code changes and cannot comment on the increased risk). 
Beside that I think it is one of the major performance bottlenecks. Some objects get written too often to disk and specially if they are larger and users have non-SSD disks it comes with considerable costs.

freimair (approved):

Issue #3605, opened 2019-11-13
Bisq not opening on windows [With solution]


FKrauss:
### Description

Bisq simply doesn't open on windows. You click the icon, nothing happens.

This is an issue to help others solve the problem as we had users who found how to fix it and we probably should use this as the starting point for a bug fix

#### Version

all (known the start on 1.7)

### Steps to reproduce

Have Special chara

1. Install bisq
2. Have special characters in your username, and thus the bisq path on Windows, such as `á` (very common in Portuguese) or `ø` (Danish)
3. try to open Bisq
4. sit and stare and the empty screen. No error messages nothing happens

### Expected behaviour

Bisq opens and you are able to use it

### Actual behaviour

Nothing happens

#### Device or machine

Windows 7 and 10

## Solution
Change the installation directory to **NOT** include any special characters anywhere in its path.

ripcurlx:
@devinbileck Do you think there is anything we can configure in install anywhere or the Windows build configuration to fix this problem?

devinbileck:
I will look into reproducing this and investigating a solution.

devinbileck:
I am able to reproduce the issue, but have yet to find the cause. The exe exits shortly after launching it with a return code of 1 while attempting to read runtime\lib\modules. I have been unable to determine anymore than that. I am leaning towards this being an issue with javapackager. We can re-evaluate this issue once we move to a newer Java version with a newer javapackager tool.

For now, I have updated the installer in order to mitigate and workaround this issue in #3642.

ripcurlx:
Closing as complete.

PR #3642, opened 2019-11-20
Implement installation path check on Windows

devinbileck:
Prior to starting the installation, check if the install location
contains special characters and if so use a different default install
location as well as display a prompt to the user providing additional
details.

This is in response to #3605, for which a proper solution has yet to be
found.


ripcurlx (approved):
utACK - Thanks for fixing this! 👍

PR #3643, opened 2019-11-20
Fix typos and cleanup

rex4539:


ripcurlx (approved):
utACK - I guess now you have found all of them  😉

rex4539:
I actually found some more later and filed https://github.com/bisq-network/bisq/pull/3647 :)

julianknutsen:
@ripcurlx I didn't have any pain rebasing these fixes this time, but this is the type of change that can cause massive headaches for other developers with outstanding PRs and in-process development work. In the future, it might be nice to either have a pseudo code-freeze for style patches that touch so many files. That way it is a known rebase point for master to limit merge pain. I appreciate the intention of the patch, but future ones in a similar style could be a lot more painful.

PR #3644, opened 2019-11-20
Account number label in Swedish fix

### LINKED ISSUE #3641: Wrong bank account type in GUI
### Description

#### Version

1.2.3

### Steps to reproduce

Konto

### Expected behaviour

NOK

### Actual behaviour

SEK

### Screenshots

Se above.

#### Additional info

rafaelpac:
Fixes #3641 
According to this issue, the label "Bankgiro number", as it was hardcoded in `BankUtil.java`, should not be used in Sweden.
The label "Kontonummer" is the correct one to be used in Sweden and Norway.

Account number validation is implemented to Norwegian accounts in `AccountNrValidator.java`.
There is no validation implemented to Swedish accounts, so nothing to change there.
If account validation for Sweden is needed, a different method would be needed (see Wikipedia reference below):

image

### DIFF
--- a/core/src/main/java/bisq/core/locale/BankUtil.java
+++ b/core/src/main/java/bisq/core/locale/BankUtil.java
@@ -156,10 +156,8 @@ public class BankUtil {
             case "CA":
             case "HK":
                 return Res.get("payment.accountNr");
-            case "NO":
-                return "Kontonummer"; // do not translate as it is used in norwegian only
-            case "SE":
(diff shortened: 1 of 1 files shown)


rafaelpac:
Wrong use of OR operator ( || )

Bisq2People:
Swedish account validation checks that the account number consists of:

11 or 13 or 15 decimal digits.

PR #3645, opened 2019-11-20
Account number label in Swedish fix

### LINKED ISSUE #3641: Wrong bank account type in GUI

rafaelpac:
Fixes #3641
According to this issue, the label "Bankgiro number", as it was hardcoded in BankUtil.java, should not be used in Sweden.
The label "Kontonummer" is the correct one to be used in Sweden and Norway.

Account number validation is implemented to Norwegian accounts in AccountNrValidator.java.
There is no validation implemented to Swedish accounts, so nothing to change there.
If account validation for Sweden is needed, a different method would be needed (see Wikipedia reference below):

image

### DIFF
--- a/core/src/main/java/bisq/core/locale/BankUtil.java
+++ b/core/src/main/java/bisq/core/locale/BankUtil.java
@@ -156,10 +156,8 @@ public class BankUtil {
             case "CA":
             case "HK":
                 return Res.get("payment.accountNr");
-            case "NO":
-                return "Kontonummer"; // do not translate as it is used in norwegian only
-            case "SE":
-                return "Bankgiro number"; // do not translate as it is used in swedish only
+            case "NO" || "SE":
+                return "Kontonummer"; // do not translate as it is used in norwegian and swedish only
             case "MX":
                 return "CLABE"; // do not translate as it is used in spanish only
             case "CL":
--- a/core/src/main/java/bisq/core/locale/BankUtil.java
+++ b/core/src/main/java/bisq/core/locale/BankUtil.java
@@ -156,7 +156,9 @@ public class BankUtil {
             case "CA":
             case "HK":
                 return Res.get("payment.accountNr");
-            case "NO" || "SE":
+            case "NO":
+                return "Kontonummer"; // do not translate as it is used in norwegian and swedish only
+            case "SE":
                 return "Kontonummer"; // do not translate as it is used in norwegian and swedish only
             case "MX":
                 return "CLABE"; // do not translate as it is used in spanish only


ghost:
Swedish account validation checks that the account number consists of:

11 or 13 or 15 decimal digits.

ripcurlx:
@rafaelpac Could you please resolve the conflict with BankUtil and update your commit messages to follow https://chris.beams.io/posts/git-commit/#7-rules - we know that you updated BankUtil.java 😉 . Thanks!

ripcurlx (changes requested):
NACK - https://github.com/bisq-network/bisq/pull/3645#issuecomment-556984208

ripcurlx:
And please squash it into one commit in that case.

camahama:
Hoping this is the correct way and place to post this bug. When I try to register a new SEPA account (Swedish Skandiabanken), the SAVE NEW ACCOUNT button remains grey and unclickable. No error messages. The same account has worked previously. Running Bisq v. 1.2.3.

ghost:
@camahama
This is not the right place but it looks like a bug, because when you toggle SEK/EUR drop-down menu the grey out disappears.

rafaelpac:
I'm sorry... I can't find how to change the commit message. I think it is not doable in the web interface, right?

ripcurlx:
> I'm sorry... I can't find how to change the commit message. I think it is not doable in the web interface, right?

You have to change your commit history locally and force push your changes to your remote PR branch.
See https://www.internalpointers.com/post/squash-commits-into-one-git for reference how to do it.

core/src/main/java/bisq/core/locale/BankUtil.java
```diff
@@ -157,9 +157,9 @@ public static String getAccountNrLabel(String countryCode) {
             case "HK":
                 return Res.get("payment.accountNr");
             case "NO":
-                return "Kontonummer"; // do not translate as it is used in Norwegian only
+                return "Kontonummer"; // do not translate as it is used in norwegian and swedish only
```
ripcurlx, line 160: As it looks like the label is exactly the same I think we could merge those two cases.

ripcurlx:
> I'm sorry... I can't find how to change the commit message. I think it is not doable in the web interface, right?

@rafaelpac Besides my comment on the cases here is how you squash multiple commits into one locally: https://www.internalpointers.com/post/squash-commits-into-one-git

rafaelpac:
I'm sorry, I completely forgot about this PR.
This was a quick fix I did for somebody's issue about a label in Swedish, I was studying this file at that time, so it was easy for me to do it.
I don't use git locally, I only use the github web interface (only for very small fixes and to study the code, I am not a developer).
So I can't easily fix the mess I did here. It is easier for now to just close this and open a new clean one.
I will try that, I hope I don't mess it up again.

PR #3649, opened 2019-11-20
Use global price data instead of local

### LINKED ISSUE #3527: Use global price data instead of local

**Forum discussion**

ghost:
Fixes #3527

--Changed timestamp from Local to Global (shouldn't matter).
--Changed the component order of Local and Global to give global data precedence over local. 
   Matters according to the base class comment below.

From pricenode/src/main/java/bisq/price/spot/ExchangeRateProvider.java

Abstract base class for providers of bitcoin ExchangeRate data. Implementations are marked with the org.springframework.stereotype.Component annotation in order to be discovered via classpath scanning. Implementations are also marked with the org.springframework.core.annotation.Order annotation to determine their precedence over each other in the case of two or more services returning exchange rate data for the same currency pair. In such cases, results from the provider with the higher order value will taking precedence over the provider with a lower value, presuming that such providers are being iterated over in an ordered list.

### DIFF
--- a/pricenode/src/main/java/bisq/price/spot/ExchangeRateService.java
+++ b/pricenode/src/main/java/bisq/price/spot/ExchangeRateService.java
@@ -87,7 +87,7 @@ class ExchangeRateService {
             t.printStackTrace();
         }
 
-        if (provider instanceof BitcoinAverage.Local) {
+        if (provider instanceof BitcoinAverage.Global) {
             metadata.put("btcAverageTs", timestamp);
         }
 
--- a/pricenode/src/main/java/bisq/price/spot/providers/BitcoinAverage.java
+++ b/pricenode/src/main/java/bisq/price/spot/providers/BitcoinAverage.java
@@ -142,18 +142,17 @@ public abstract class BitcoinAverage extends ExchangeRateProvider {
 
     @Component
     @Order(1)
-    public static class Global extends BitcoinAverage {
-        public Global(Environment env) {
-            super("BTCA_G", "btcAverageG", 0.3, "global", env);
-        }
-    }
-
-
-    @Component
-    @Order(2)
     public static class Local extends BitcoinAverage {
         public Local(Environment env) {
             super("BTCA_L", "btcAverageL", 0.7, "local", env);
         }
     }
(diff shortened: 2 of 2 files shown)


chimp1984:
This changes would apply it for all markets which is likely not intended. Big markets are more accurate with local price.

ghost:
With global price one is assured that all market prices have the same average reference price calculated by BitcoinAverage. Thus all the prices are behaving the same because all prices are then calculated by currency conversion. That is how prices normally are viewed as being on a single global market. In opposite local prices are set separate thus the price appears way off and not properly aligned to the single global market, that is for low volume currency markets. One can go to BitcoinAverage and check that for USD the difference between local and global price is max 0.1%. But for minor currencies on low volume markets the price is way off, typically 10-30% for SEK. 
There are reports that this price can not be used for serious trading. I would myself not make a market order in SEK in Bisq because of this. It would almost be the same as Bisq wouldn't have an external price reference. Because of low volume we need good external price references.

Thank you for your concern, you have a point, but I have studied this phenomena in depth on BitcoinAverage and it should be clear after some evaluation that using global price is the optimal solution that doesn't make special exceptions for selected currencies making it messy and more difficult to implement.

Bisq2People:
https://github.com/bisq-network/bisq/pull/3649#issuecomment-556540535

chimp1984:
> As told above it differs about 0.1% for the biggest market (USD). That's is accurate enough.

Ask a market maker. I doubt he will agree.

I think we need to fix that in a more sophisticated way rather to move all to global mode. E.g. use a list of main markets where we want to use local rate and others use global rate.

Bisq2People:
You take BitcoinAverage's local prices as absolute truth then!?

Don't you think 0.1% can be seen in the fault tolerance of BitcoinAverage own weighted average price calculation?

Up to you, I am not changing the code...

ghost:
AMENDMENT

The 0.1% difference on USD is actually a standard deviation. **Mean error is 0.**

Also **0.1% is presumably in the fault margin** of BitcoinAverage weighted average.

Please pull code to fix all minor currency markets.

PR #3651, opened 2019-11-21
Un-generify Popup class

stejbac:
Remove the type parameter `T` from `Popup<T> extends Overlay<Popup>`, as it appears to have never been used or set anywhere in the code. (This mainly involves replacing a lot of `new Popup<>` occurrences, which was done entirely automatically by the IDE.)

I guess the original purpose of the type parameter was to allow different kinds of popup, all extending the same class, e.g. `Foo extends Popup<Foo> extends Overlay<Foo>`. However, due to the fluent interface of `Overlay`, this wouldn't be entirely type safe as long as `Popup` has a public constructor and standalone instances of it can be created. For example,

> `Foo foo = new Popup<Foo>().headline("foo");`

would typecheck fine but throw a `ClassCastException` at runtime, when it tries to assign the underived `Popup` instance to `foo`. (In more complex cases this could cause heap pollution.)

If different popup subclasses are later required, it would probably be better to define an abstract base class but keep all the concrete subclasses non-generic, e.g. `BasePopup<T extends BasePopup<T>> extends Overlay<T>`, similar to the `TabbedOverlay` class.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/app/BisqApp.java
+++ b/desktop/src/main/java/bisq/desktop/app/BisqApp.java
@@ -65,7 +65,6 @@ import javafx.stage.StageStyle;
 
 import javafx.scene.Parent;
 import javafx.scene.Scene;
-import javafx.scene.image.Image;
 import javafx.scene.input.KeyCode;
 import javafx.scene.input.KeyEvent;
 import javafx.scene.layout.StackPane;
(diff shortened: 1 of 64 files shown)


ripcurlx (approved):
utACK - Now I'm officially `<>` blind after reviewing this PR

PR #3655, opened 2019-11-21
Refactor OfferView constructor

lusarz:
`direction` is now passed directly to `OfferView` constructor, instead of doing `(this instanceof BuyOfferView)` checks in superclass.

Before that is was hard to check what is difference between `SellOfferView` and `BuyOfferView` classes.


ripcurlx (approved):
utACK

PR #3658, opened 2019-11-21
Add missing @Named annotations for CoinFormatter injection

bodymindarts:
I missed a few places where `@Named(BTC_FORMATTER_KEY)` annotation should be added in my recent PR https://github.com/bisq-network/bisq/pull/3634.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/funds/transactions/TransactionsView.java
+++ b/desktop/src/main/java/bisq/desktop/main/funds/transactions/TransactionsView.java
@@ -36,6 +36,7 @@ import bisq.core.offer.OpenOffer;
 import bisq.core.trade.Tradable;
 import bisq.core.trade.Trade;
 import bisq.core.user.Preferences;
+import bisq.core.util.FormattingUtils;
 import bisq.core.util.coin.CoinFormatter;
 
 import bisq.network.p2p.P2PService;
@@ -55,6 +56,7 @@ import org.bitcoinj.wallet.listeners.WalletEventListener;
 import com.googlecode.jcsv.writer.CSVEntryConverter;
 
 import javax.inject.Inject;
+import javax.inject.Named;
 
 import de.jensd.fx.fontawesome.AwesomeIcon;
 
@@ -127,7 +129,7 @@ public class TransactionsView extends ActivatableView<VBox, Void> {
     private TransactionsView(BtcWalletService btcWalletService,
                              P2PService p2PService,
                              WalletsSetup walletsSetup,
-                             CoinFormatter formatter,
+                             @Named(FormattingUtils.BTC_FORMATTER_KEY) CoinFormatter formatter,
                              Preferences preferences,
                              TradeDetailsWindow tradeDetailsWindow,
                              OfferDetailsWindow offerDetailsWindow,
--- a/desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferView.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferView.java
@@ -54,6 +54,7 @@ import bisq.core.offer.OfferPayload;
 import bisq.core.payment.PaymentAccount;
 import bisq.core.payment.payload.PaymentMethod;
 import bisq.core.user.DontShowAgainLookup;
+import bisq.core.util.FormattingUtils;
 import bisq.core.util.coin.BsqFormatter;
 import bisq.core.util.coin.CoinFormatter;
 
@@ -70,6 +71,7 @@ import net.glxn.qrgen.QRCode;
 import net.glxn.qrgen.image.ImageType;
 
 import javax.inject.Inject;
+import javax.inject.Named;
 
 import de.jensd.fx.glyphs.materialdesignicons.MaterialDesignIcon;
 
@@ -172,7 +174,7 @@ public class TakeOfferView extends ActivatableViewAndModel<AnchorPane, TakeOffer
     @Inject
     private TakeOfferView(TakeOfferViewModel model,
                           Navigation navigation,
-                          CoinFormatter formatter,
+                          @Named(FormattingUtils.BTC_FORMATTER_KEY) CoinFormatter formatter,
                           BsqFormatter bsqFormatter,
                           OfferDetailsWindow offerDetailsWindow,
                           Transitions transitions) {


desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeView.java
```diff
@@ -139,7 +142,7 @@
     public DisputeView(DisputeManager<? extends DisputeList<? extends DisputeList>> disputeManager,
                        KeyRing keyRing,
                        TradeManager tradeManager,
-                       CoinFormatter formatter,
+                       @Named(FormattingUtils.BTC_FORMATTER_KEY) CoinFormatter formatter,
```
lusarz: I think in this case it's unnecessary, because this constructor is not annotated with `@Inject`

lusarz (changes requested):

desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeView.java
bodymindarts: That is what I thought too, which is why I left these out originally. But it turns out `@inject` is not infact mandatory and the program crashes on these views without this annotation. Thanks to @stejbac for discovering this.

lusarz: Are you sure that app crash when you ommit `@Named` annotation on this particular file (`DisputeView`) ? It's an abstract class and it does not "participate" in dependency injection.

stejbac: On second thought, I might have been wrong about `@Inject` not being mandatory for Guice, since looking closer the modified `Dispute*` classes all have subclasses with `@Inject`-annotated constructors - probably only `TransactionsView` and `TakeOfferView` actually needed fixing. (I'm pretty sure it isn't mandatory by default for Spring, but perhaps Guice is stricter.)

Also, there are (probably leftover) `@FxmlView` annotations on the 3 `Dispute*` classes which mislead me to think that they were concrete components - perhaps those annotations can be removed?

ripcurlx: I think here it is not necessary as mentioned by @lusarz. At least it didn't crash for me here.

desktop/src/main/java/bisq/desktop/main/support/dispute/agent/DisputeAgentView.java
```diff
@@ -48,7 +49,7 @@
     public DisputeAgentView(DisputeManager<? extends DisputeList<? extends DisputeList>> disputeManager,
                             KeyRing keyRing,
                             TradeManager tradeManager,
-                            CoinFormatter formatter,
+                            @Named(FormattingUtils.BTC_FORMATTER_KEY) CoinFormatter formatter,
```
ripcurlx: I think here it is not necessary.

desktop/src/main/java/bisq/desktop/main/support/dispute/client/DisputeClientView.java
```diff
@@ -30,16 +30,19 @@
...
 
+import javax.inject.Named;
+
 @FxmlView
 public abstract class DisputeClientView extends DisputeView {
     public DisputeClientView(DisputeManager<? extends DisputeList<? extends DisputeList>> DisputeManager,
                              KeyRing keyRing,
                              TradeManager tradeManager,
-                             CoinFormatter formatter,
+                             @Named(FormattingUtils.BTC_FORMATTER_KEY) CoinFormatter formatter,
```
ripcurlx: I think here it is not necessary.

ripcurlx (changes requested):
NACK - please see my comments

desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeView.java
ripcurlx: > But it turns out `@inject` is not infact mandatory and the program crashes on these views without this annotation.

Did it crash for you while testing?

lusarz: I agree that probably only these two views needs to be annotated. 

Yeah, I suppose `@FxmlView` on abstract views are leftovers, and it would be nice to remove them.

ripcurlx:
One more question: Is there a reason why we use for these injects the Named annotation from javax.inject.Named and on other places com.google.inject.name.Named?

ripcurlx:
> One more question: Is there a reason why we use for these injects the Named annotation from javax.inject.Named and on other places com.google.inject.name.Named?

It seems both work the same as described in https://github.com/google/guice/wiki/JSR330

lusarz:
@ripcurlx I think there is no reason - I suppose everyone use autoimport feature in Intellij and no one cares which one is imported :)

ripcurlx:
@bodymindarts It looks like you still left the unused imports in and a small typo in the commit message `@nAmed` 😉

bodymindarts:
I have force pushed removing the annotation from the abstract classes.

I have consistently used
```
import javax.inject.Named;
```
To mirror the usage of 
```
import javax.inject.Inject;
```

These are library agnostic and should still be effective if a different DI framework is used. I didn't notice that another import was used elsewhere.

ripcurlx:
> I have force pushed removing the annotation from the abstract classes.
> 
> I have consistently used
> 
> ```
> import javax.inject.Named;
> ```
> 
> To mirror the usage of
> 
> ```
> import javax.inject.Inject;
> ```
> 
> These are library agnostic and should still be effective if a different DI framework is used. I didn't notice that another import was used elsewhere.

I guess we should get rid of the Guice bound annotations in favor of the generic one at some point.

bodymindarts:
Removed the unused imports

bodymindarts:
And commit message is no typo. If you look at the message locally its fine. It seems to be a display bug in github.

ripcurlx:
> It seems to be a display bug in github.

Yes it looks like GitHub tries to be clever finding a user with that name 🤦‍♂

ripcurlx (approved):
ACK

lusarz:
@bodymindarts Would you prepare pull request that removes unnecessary `@FxmlView` annotations from abstract views ? You should be able to find them using `@FxmlView[\s\n]*public abstract` regex.

bodymindarts:
@lusarz https://github.com/bisq-network/bisq/pull/3661

desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeView.java
chimp1984: Inject annotation is needed for the concrete classes but not for the base classes (should not be used there). It probably crashed due some nullpointers caused if the annotation is missing but with view classed you don't get proper error logs.

chimp1984: Inject annotation is needed for the concrete classes but not for the base classes (should not be used there). It probably crashed due some nullpointers caused if the annotation is missing but with view classed you don't get proper error logs.

PR #3661, opened 2019-11-22
Cleanup fmxlview and javax imports

bodymindarts:
Removes redundant `@FxmlView` from `abstract` view classes and consitently imports DI annotations from `javax` package.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/support/dispute/agent/DisputeAgentView.java
+++ b/desktop/src/main/java/bisq/desktop/main/support/dispute/agent/DisputeAgentView.java
@@ -42,7 +42,6 @@ import com.google.inject.name.Named;
 
 import javafx.scene.control.Button;
 
-@FxmlView
 public abstract class DisputeAgentView extends DisputeView {
 
     public DisputeAgentView(DisputeManager<? extends DisputeList<? extends DisputeList>> disputeManager,
--- a/desktop/src/main/java/bisq/desktop/main/support/dispute/agent/DisputeAgentView.java
+++ b/desktop/src/main/java/bisq/desktop/main/support/dispute/agent/DisputeAgentView.java
@@ -38,7 +38,7 @@ import bisq.core.util.coin.CoinFormatter;
 
 import bisq.common.crypto.KeyRing;
 
-import com.google.inject.name.Named;
+import javax.inject.Named;
 
 import javafx.scene.control.Button;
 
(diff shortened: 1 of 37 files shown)


lusarz (dismissed):
For me looks great :)

stejbac:
The `@Named` DI annotations on the constructor params of `AgentRegistrationView` and `DisputeAgentView` are presumably also not needed. (Also can remove redundant `FxmlView` imports.)

desktop/src/main/java/bisq/desktop/main/support/dispute/agent/DisputeAgentView.java
```diff
@@ -54,7 +50,7 @@ public DisputeAgentView(DisputeManager<? extends DisputeList<? extends DisputeLi
                             ContractWindow contractWindow,
                             TradeDetailsWindow tradeDetailsWindow,
                             AccountAgeWitnessService accountAgeWitnessService,
-                            @Named(AppOptionKeys.USE_DEV_PRIVILEGE_KEYS) boolean useDevPrivilegeKeys) {
```
cbeams, line 57: It's not obvious to me why injection of this value would still work without the explicit annotation. The value of `AppOptionKeys.USE_DEV_PRIVILEGE_KEYS` is in fact "useDevPrivilegeKeys", spelled exactly the same as the `useDevPrivilegeKeys` parameter, so in theory Guice could introspect and do the injection based on that, but this information is not inspectable at runtime unless the `-parameters` option has been passed to javac to keep the parameter symbol table available (and our build does not do this). What am I missing?

cbeams (commented):
utACK assuming a valid answer to my comment.

desktop/src/main/java/bisq/desktop/main/support/dispute/agent/DisputeAgentView.java
bodymindarts: The class is abstract and never instantiated directly. The annotation is legacy from a time before the type hierarchy was in place.

bodymindarts: The sub-classes should have the annotation.

cbeams: Roger that, and I see that the subclasses are in fact annotated. Thanks.

utACK

freimair (approved):
Ack

PR #3664, opened 2019-11-22
(5/6) Broadcast Remove messages seen before Adds

julianknutsen:
New commits 931c1f4

In the event that we receive an out of order RemoveData or RemoveMailboxData message, relay it to the P2P network. This should help with the propagation of these messages.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
+++ b/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
@@ -68,6 +68,8 @@ import com.google.inject.name.Named;
 import javax.inject.Inject;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Maps;
+
 
 import java.security.KeyPair;
 import java.security.PublicKey;
@@ -75,6 +77,7 @@ import java.security.PublicKey;
 import java.time.Clock;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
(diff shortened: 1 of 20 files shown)


p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -501,14 +501,14 @@ public boolean remove(ProtectedStorageEntry protectedStorageEntry,
...
-
-        // Valid remove entry, do the remove and signal listeners
-        removeFromMapAndDataStore(protectedStorageEntry, hashOfPayload);
+        if (storedEntry != null) {
+            // Valid remove entry, do the remove and signal listeners
+            removeFromMapAndDataStore(protectedStorageEntry, hashOfPayload);
+        } /* else {
+            // This means the RemoveData or RemoveMailboxData was seen prior to the AddData. We have already updated
+            // the SequenceNumberMap appropriately so the stale Add will not pass validation, but we still want to
+            // broadcast the remove to peers so they can update their state appropriately
```
freimair, line 510: The return value of the call is changed from `false` to `true` again. If I am not mistaken, we discussed this change (the other way around) quite recently. Do you see any issues that can arise from this change?

freimair (changes requested):

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
julianknutsen: I don't see any cases where this can cause issues for the client code.

The original change from true->false was in the equivalent sequence number case so it isn't related to this change.

In this case, the contract of the return value has stayed the same since it is based on the internal state changing. Previously, nothing changed so we returned false, but now that the SequenceNumberMap is changed in some cases it needs to return true.

In practice, this branch of the code is not relevant for return values. The `onMessage()` handler is the only caller we expect to receive `RemoveData` messages before `AddData` messages since the race is related to when the connections happen in relation to the original add/removes. It doesn't even look at the return values.

The P2PService caller is the second user and will only remove `ProtectedStorageEntry` items that it knows about which means they must have existed locally so the `remove()` call will always return true. Personally, I would like to add asserts around this to guarantee it, but they are risky for this type of legacy code where there are strange interactions I am not familiar with.

I also don't like returning `false` here in the add-remove-case because there is a guarantee being added "all future adds older than this operations will fail" and returning false is confusing because it makes it seem like nothing happened.

freimair (approved):

ripcurlx (approved):
utACK

PR #3667, opened 2019-11-23
(1/5) Test and Refactor P2PDataStorage Synchronization Path

julianknutsen:
## Motivation
The GetData synchronization sequence was currently untested and the code was tangled between the connection request handling logic and the `P2PDataStorage` update logic. In order to implement a new feature (**_replay removes during synchronization_**) I needed a way to verify current behavior and write tests for the changes I am making.

## Changes
This PR moves the `GetDataRequest`->`GetDataResponse`->`ProcessDataResponse` logic into the
P2PDataStorage where it can operate locally on the various data structures that need to be accessed
and updated. This has a few benefits:

- Easier testing: tests don't need to mock `Future` and `Connection` to validate behavior
- Public APIs can be removed that allow operations on internal data structures in favor of letting
  the P2PDataStorage handle the adds/removes/updates.
- Optimizations can be handled internally to the `P2PDataStorage` limiting external knowledge of implementation details
- Better encapsulation of the synchronization logic. The code is all about synchronizing two separate `P2PDataStorage` instances so letting the actual objects build and apply the synchronization messages makes more intuitive sense.

## Tests
Full unit test for each individual stage now exists as well as an integration test that creates two separate `P2PDataStorage` instances, runs the synchronization logic, and verifies they are synchronized. This type of "multiple node" testing will be used to validate the new feature development and will limit errors in development.

## Maintainability fixes
Since this patch tests the `GetData` messages sent back and forth, there was an opportunity to clean up the guarantees around null vs. empty w.r.t. sets and Capabilities. I think this is a good pattern moving forward to limit accidental NullPointerExceptions when we can do something less brittle.


p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -315,9 +308,7 @@ private boolean shouldTransmitPayloadToPeer(Capabilities peerCapabilities, Netwo
                 doAdd = true;
             }
             if (doAdd) {
-                boolean notContained = lookupSet.add(protectedStoragePayload.hashCode());
```
julianknutsen: @chimp1984 I could use your prior knowledge for the behavior. Do you know of any cases where we intentionally want/expect the payload hash and the hashCode to be different?

chimp1984: It deals with the case that there are multiple `ProtectedStorageEntry` objects with the same `ProtectedStoragePayload`. Not 100% sure if there are valid cases for that as the map in `P2PDataStorage` should not contain entries with the same `ProtectedStoragePayload`. There was the old `TradeStatistics` which was a `ProtectedStoragePayload` but that is not used anymore. The new is a `PersistableNetworkPayload`. For `TradeStatistics2` objects it is valid that we receive it 2 times (both traders are publishing it) in a different envelopes.

So I am not aware of any actual use case, but maybe it is better to keep that logic (maybe add a comment to make it easier to understand) to be able to support such a use case in future and to not break things if we have overlooked an existing use case. But I leave it to your consideration. You know that domain already better than myself... ;-)

EDIT: After a second look to map I am pretty sure there is no case were 2 different `ProtectedStoragePayload` could be delivered as we use the hash as key. So that check with the lookupMap is probably pointless.

julianknutsen:
Rebased against master and published now that v1.2.4 is branched.

freimair (approved):
utAck

ripcurlx (approved):
utACK

PR #3668, opened 2019-11-23
(2/5) P2PDataStorage Client API Sanitization

julianknutsen:
0649323

## Motivation
Minimizing the public API and public API parameters will limit the number of future bugs in development. It will also reduce the testing space of the API since the number of ways it can be used is reduced.

There was quite a bit of exiting technical debt around `isDataOwner` as well as unnecessary APIs and parameters now that the `GetData` synchronization is handled internally.

This cleans up the API and adds documentation to reduce the technical debt.


julianknutsen:

freimair (approved):
utAck

ripcurlx (approved):
utACK

PR #3669, opened 2019-11-23
(3/5) Make isDataOwner a private policy decision in BroadcastHandler

julianknutsen:
bfdb8f5

## Motivation
Clean up the API usages of broadcast and P2PDataStorage APIs, by removing an unneeded parameter.

##
`isDataOwner` is used when deciding how many peer nodes should receive
a BroadcastMessage. If the `BroadcastMessage` originated
on the local node it is sent to ALL peer nodes with a small delay.

If the node is only relaying the message (it originated on a different
node) it is sent to MAX(peers.size(), 7) peers with a delay that is
twice as long.

All the information needed to determine whether or not the
`BroadcastMessage` originated on the local node is available at the final
broadcast site and there is no reason to have callers pass it in.

In the event that the sender address is not known during broadcast (which
is only a remote possibility due to how early the local node address
is set during startup) we can default to relay mode.

This first patch just removes the deep parameters. The next will remove
everything else. There is one real change in `LiteNodeNetworkService.java`
where it was using the local node when it should have been using the
peer node. This was updated to the correct behavior.


julianknutsen:

freimair (approved):
utAck

ripcurlx (approved):
utACK

PR #3670, opened 2019-11-23
(4/5) Inline 1&2 line private functions in P2PDataStorage

julianknutsen:
6ff8756

One less jump you have to make in the editor to see what is actually happening.


julianknutsen:

freimair (approved):
utAck

ripcurlx (approved):
utACK

PR #3671, opened 2019-11-23
(5/5) [TESTS] Clean up mockito never() and eq(null) usages

julianknutsen:
17f4b70

Mockito's never() and any() don't play well together for nullable types. Change the broadcast mocks to user nullable() and fixup tests appropriately.


julianknutsen:

freimair (approved):
utAck

ripcurlx (approved):
utACK

Issue #3602, opened 2019-11-12
A timeout occurred.


vitormilagres1992:
I simply cannot complete any transactions. Right after I click on REVIEW TAKE OFFER TO BUY BITCOIN, I get the following message: A TIMEOUT OCCURRED, as you can see in the attached picture and in the log files.

I have tried to use it in two different locations (home and work), and I get the same error. On both instances I am running the same version 1.2.3 on Linux.

I have tried via Ethernet, wifi and even using my phone's hotspot. I have also tried it with two different sellers. Still the same error.

Thanks

![Screenshot from 2019-11-12 18-59-35](https://user-images.githubusercontent.com/44966647/68714769-4e3fa200-057f-11ea-8d85-b6c69bcfcd62.png)
[bisq.log](https://github.com/bisq-network/bisq/files/3838429/bisq.log)

captainreptile:
I experienced same error today while adding sell btc offer.  
After restarting the app i get a warning: 

> A transaction was rejected from the network for reason 'dust' (64).

as well as no connection to bitcoin network, offer fees were withdrawn and no offer or history in portfolio.

Using Linux and Bisq v1.2.3

vitormilagres1992:
This is so frustrating, it's happening to every single transaction I make. Yesterday I saw 7 different buy offers and I was willing to buy all of them, but guess what? None of them worked, not even a single one. What the hell is happening?

captainreptile:
I managed to solve it by forcing a resync: 

> Settings > NETWORK INFO

 and then hit on 

> "Delete SPV File and resync"

it took me ~45min to resync and since its working without any issue.

vitormilagres1992:
Tried it to resync it, but it didn't work. I've noticed I cannot do SAME BANK transactions, there's always this timeout error. The only ones that will go through are NATIONAL BANK and CASH DEPOSIT

stejbac:
I had a look through the attached logfile earlier. It appears that the issue is due to an inconsistency between the way `bisq.core.payment.ReceiptValidator` checks for a match between offer-taker bank account types (as used by the Offer Book & Take Offer views) and the validation later done in the constructor of `bisq.core.trade.Contract`. The latter is causing an error at the other end, as seen in the logfile:

[...]

PR #3673, opened 2019-11-24
Prevent taking of offers with unequal bank account types (excl. SEPA)

stejbac:
Use stricter criteria when deciding which of the taker's accounts (if any) are valid for a given offer. Specifically, prevent National Bank accounts from being used to take Same / Specific Bank(s) offers, so the three payment method types can never being mixed.

This prevents an error on the trading peer when the trade starts, due to enforcement of equal maker & taker payment method IDs (except for SEPA) in the Contract payload constructor.

This partially addresses #3602, where the erroneous peer response caused the taker to be presented with a confusing timeout, after they managed to take a _SAME_BANK_ offer using their _NATIONAL_BANK_ account. With this change, the taker would be directed to create a separate _SAME_BANK_ account if they don't have one already (which would then be automatically selected).

(Later, it might be worthwhile to make a more substantial change to allow National, Same or Specific Bank account types to be mixed under certain circumstances when trading, but that could be complicated by the peer running an older version.)

### DIFF
--- a/core/src/main/java/bisq/core/payment/ReceiptValidator.java
+++ b/core/src/main/java/bisq/core/payment/ReceiptValidator.java
@@ -66,10 +66,15 @@ class ReceiptValidator {
             return true;
         }
 
+        // Aside from Sepa or Sepa Instant, payment methods need to match
+        if (!isEqualPaymentMethods) {
+            return false;
+        }
+
         if (predicates.isOfferRequireSameOrSpecificBank(offer, account)) {
             return predicates.isMatchingBankId(offer, account);
         }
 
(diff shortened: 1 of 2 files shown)


freimair (approved):
utAck

ripcurlx:
> (Later, it might be worthwhile to make a more substantial change to allow National, Same or Specific 
Bank account types to be mixed under certain circumstances when trading, but that could be complicated by the peer running an older version.)

I agree to this patch for now like that, but we should have a look if we are able to solve this as mentioned in a future PR to make it more user friendly.

ripcurlx (approved):
utACK

PR #3680, opened 2019-11-25
Fix #3662: Malformed trade chat & dispute speech bubbles

### LINKED ISSUE #3662: Malformed trade chat & dispute speech bubbles
### Description

In the trade chat window and mediation chat pane, sometimes the "arrow" of a speech bubble is rendered on the wrong side (observed: left hand side instead of right), pointing into the bubble as seen in the screenshots below.

#### Version

1.2.3 and latest snapshot of master (2019/11/21).

### Steps to reproduce

(I haven't attempted to reproduce it yet - observed on one actual mediated trade.)

### Expected behaviour

Each speech arrow is placed correctly, such that it always points outwards.

### Actual behaviour

Arrows are intermittently placed on the wrong side of the bubble (although apparently always in the right orientation), so that they point into it. The rendering error is not consistent and sometimes a given individual speech bubble is rendered correctly when viewed later, e.g. after restarting the application. Also, I've observed consecutive speech bubbles from the same sender render correctly then incorrectly (or maybe it was the other way round?).

### Screenshots

Screenshot from 2019-11-21 18-10-30
Screenshot from 2019-11-21 19-25-57

#### Device or machine

AMD64 Debian Linux (HP laptop)

stejbac:
Prevent the 'arrow' of a message bubble from being sporadically anchored to the wrong side - appearing on the left instead of the right hand side of the bubble. This is due to the same `ListCell` object being reused by JavaFX for different bubbles as the user scrolls up and down the chat pane, which requires that the anchors of each arrow be properly cleared between `ListCell.updateItem(..)` calls.

To this end, move the block of `AnchorPane.clearConstraints(..)` calls to the beginning of the `updateItem(..) method`, as the apparent assumption that `updateItem(item, empty = true)` will always be called to clear the given `ListCell` before reusing it as a new bubble turns out to be wrong.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/shared/ChatView.java
+++ b/desktop/src/main/java/bisq/desktop/main/shared/ChatView.java
@@ -297,7 +297,7 @@ public class ChatView extends AnchorPane {
                     }
 
                     @Override
-                    public void updateItem(ChatMessage message, boolean empty) {
+                    protected void updateItem(ChatMessage message, boolean empty) {
                         super.updateItem(message, empty);
                         if (message != null && !empty) {
                             copyIcon.setOnMouseClicked(e -> Utilities.copyToClipboard(messageLabel.getText()));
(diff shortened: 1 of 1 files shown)


freimair (approved):
utAck

ripcurlx (approved):
utACK - Agree that it is safer to do it like that and that it will fix any constraints issues.

PR #3682, opened 2019-11-25
Improve account signing security

ripcurlx:


freimair (approved):
utAck. Did a small smoke test though.

PR #3665, opened 2019-11-22
(6/6) [BUGFIX] Fix duplicate sequence number use case (Non-persistent ProtectedStoragePayloads)

julianknutsen:
Fixes: #3679

3d571c4

Fix a bug introduced in d484617385276d033223ad8f36e821504e116f96 that
did not properly handle a valid use case for duplicate sequence numbers.

For in-memory-only ProtectedStoragePayloads, the client nodes need a way
to reconstruct the Payloads after startup from peer and seed nodes. This
involves sending a ProtectedStorageEntry with a sequence number that
is equal to the last one the client had already seen.

This patch adds tests to confirm the bug and fix as well as the changes
necessary to allow adding of Payloads that were previously seen, but
removed during a restart.

PR #3690, opened 2019-11-25
(7/7) FileManager.java Cleanup and Audit

julianknutsen:
2208003

### Motivation
There was suspicion of a corruption bug inside FileManager. In order to unblock a review, I went through and audited the code and cleaned it up to make the concurrency more obvious.

At the end of the analysis, there was no corruption bug regarding the TODO in FileManager.java:75, but the interactions between the `persistable` field and the `savePending` were hard to understand. This has been updated with comments and an `AtomicReference` usage to make it more clear.

In some interleavings, two `writeToFile` calls for the same exact data could have occurred and this remains the case since references are passed in whose underlying data can change while `saveToFile` is running on another thread. That is why many implementations use a `Concurrent` data structure. Although, it is worth point out that not all do and it _could_ be the source of some strange bugs.

### Analysis
The comment was concerned with a situation where an in-progress `saveFileTask` would allow a UserThread `saveLater` call to schedule another write. There are two reasons this is OK here:

1) Reference writes are atomic
2) `saveToFile` is `synchronized`

In the event that an in-progress `saveLater` call spawned another `saveFileTask`, only one would be allowed to write the file at a time.

It is possible that the second call could write the file first, but since all callers share the same persistable reference, **_both_** `saveToFile` calls will write the latest data.

### Bug Fix
There was one strange behavior that was fixed in this PR. If the first `saveLater` call had a large delay it would override any future `saveLater` delays until the original one was finished. This was because the first `saveLater` set `savePending` so all future `saveLater` calls returned early without scheduling a thread.

The update causes all requests to spawn a task so if the second `saveLater` call has a shorter delay, it will run and batch with the first call. The second task will finally get scheduled and it will immediately exit since there is no work to do.

### Future Work
I think if the end goal is to have all writes completely thread-safe. You would need to do something like have `PersistableEnvelope` define a `clone()` function that all subclasses implement. The FileManager could then just clone the object prior to passing it off to the writing thread and have some guarantees.

That is way outside the scope of this work, but it may be a good piece for someone else to pick up.

### DIFF
--- a/common/src/main/java/bisq/common/storage/FileManager.java
+++ b/common/src/main/java/bisq/common/storage/FileManager.java
@@ -36,20 +36,21 @@ import java.util.Random;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.ReentrantLock;
 
 import lombok.extern.slf4j.Slf4j;
+
+import static com.google.common.base.Preconditions.checkNotNull;
 
 @Slf4j
 public class FileManager<T extends PersistableEnvelope> {
     private final File dir;
     private final File storageFile;
     private final ScheduledThreadPoolExecutor executor;
-    private final AtomicBoolean savePending;
     private final long delay;
     private final Callable<Void> saveFileTask;
-    private T persistable;
+    private final AtomicReference<T> nextWrite;
     private final PersistenceProtoResolver persistenceProtoResolver;
     private final ReentrantLock writeLock = CycleDetectingLockFactory.newInstance(CycleDetectingLockFactory.Policies.THROW).newReentrantLock("writeLock");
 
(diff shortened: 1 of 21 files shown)


julianknutsen:
@freimair @chimp1984 This is what I found in the day of auditing.

I agree that some users pass in references that are not thread-safe, but fixing that requires a real design and implementation plan that is outside of the scope of my current work. I've outlined a potential solution in the `Future Work` section for anyone else who is interested or assigned to fix that technical debt.

Feel free to take this or trash it. I'm also happy to just close my open PR that fixes the persistence bug in `TempProposalStore`. The amount of overhead maintaining these patches for weeks isn't worth my time if they are known issues and we don't have the resources to fix them.

chimp1984:
Thanks @julianknutsen for looking into it and fixing issues there. I agree that a more extensive improvement is outside of the current scope. To clone each data structure might become expensive for larger data structures like DaoState or TradeStatistics. 

For a larger refactoring in that area we should probably consider to use a database solution instead of the simple file based approach.

freimair (approved):
utAck

ripcurlx (approved):
utACK

ripcurlx:
For documentation purpose. Reviews happened at:
- https://github.com/bisq-network/bisq/pull/3665
- https://github.com/bisq-network/bisq/pull/3664
- https://github.com/bisq-network/bisq/pull/3638
- https://github.com/bisq-network/bisq/pull/3636
- https://github.com/bisq-network/bisq/pull/3621
- https://github.com/bisq-network/bisq/pull/3628

PR #3691, opened 2019-11-26
Added a Pull Request template

freimair:
In order to streamline our reviewing process and thus, get more PR reviewed and subsequently merged in less time, I added a simple pull request template.

Main points:
- highlight to re-request a review after you addressed the requested changes in order for us to not loose track of your PR
- prefill github keywords for closing issues, as these are not always used correctly

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,0 +1,13 @@
+<!-- 
+- make yourself familiar with the CONTRIBUTING.md if you have not already (https://github.com/bisq-network/bisq/blob/master/CONTRIBUTING.md)
+- make sure you follow our [coding style guidelines][https://github.com/bisq-network/style/issues)
+- pick a descriptive title
+- provide some meaningful PR description below
+- create the PR
+- in case you receive a "Change request" and/or a NACK, please react within 30 days. If not, we will close your PR and it can not be up for compensation.
+- After addressing the change request, __please re-request a review!__ Otherwise we might miss your PR as we tend to only look at pull requests tagged with a "review required".
+-->
+
+Fixes #replaceWithIssueNr, fixes #replaceWithIssueNr
+
+Your PR description here.


chimp1984 (dismissed):
utACK

ripcurlx:
Maybe also add the signing of commits, with a link to the GitHub docu, in the comment block as it is a common problem for new contributors.

ripcurlx:
Maybe also add a hint that refactorings are fine, but it would be better to combine them together with changes that have direct user value (bug fixes, new features).

pull_request_template.md
```diff
@@ -0,0 +1,11 @@
+<!-- 
+- make sure you follow our [coding style guidelines][https://github.com/bisq-network/style/issues)
+- provide some meaningful info below
+- create the PR
+- in case you receive a "Change request" and/or a NACK, please react within 30 days. If not, we will close your PR and it can not be up for compensation.
+- After addressing the change request, __please re-request a review!__ Otherwise we might miss your PR as we tend to only look at the PR where there is a review required.
+-->
+
+Fixes #, fixes #
```
julianknutsen: Since this is targetted at new committers. It might be good to explain a bit about how GitHub uses the `fixes` keyword to onboard new users.

Maybe something similar to: `GitHub can automatically close or tag issues that are listed in the Pull Request. To automatically close a bug use fixes #<issue-number>. To reference an issue, but not close it, use: #<issue-number>`

ripcurlx:
Please also add instructions that the PR title should be communicating the user benefit as good as possible to make it easier to compile release notes.

freimair:
### General response

thanks for your suggestions and time spent thinking about that. I, myself, had a much bigger template in the first place and worked it down to just these points. After digging through the [github manual on pull request templates](https://help.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates#pull-request-templates) I realized that a bigger wall of text is going to be less read than just a few lines. The ultimate goal of this is to improve stuff in our workflow and not to teach people how to use github.

Imho, such a template will never be perfect and final. I suggest starting out with this small one and see if it actually has any impact on our workflow. If it does, great, we optimized our workflow. If it does not, there has not been a great waste of resources.

Besides that, I added and tweaked a few lines. We should probably talks about that in the ["Pull Requests" Dev-Call](https://github.com/bisq-network/events/issues/35) if I finally manage to hold that one.

### Onto your comments:

>  It might be good to explain a bit about how GitHub uses the fixes keyword to onboard new users.

As mentioned above, I already had a small wall of text linking to the github manual, explaining, given examples, etc with the same motivation you mention.

But is that really necessary to reach the ultimate goal?
- a template like `Fixes #replaceWithIssueNr` motivates to just fill in the issue number(s) your PR fixes,
- even as experienced devs occasionally miss include the reference, the simple line reminds them in the most unobtrusive way.

The ultimate goal of this line is to close issues which have been resolved and not to teach people how to use github.

> Please also add instructions that the PR title should be communicating the user benefit as good as possible to make it easier to compile release notes.

Again, I thought about including stuff like this. I ultimately decided against it as we already have a [database of do's and don'ts](https://github.com/bisq-network/style/issues) including all sorts of stuff (which, most probably, nobody reads anyways)

> Maybe also add the signing of commits, with a link to the GitHub docu, in the comment block as it is a common problem for new contributors.

Added the link to the CONTRIBUTING.md.

julianknutsen:
It seems like the focus of this patch was to give reviewers a better workflow that is more productive. So, just do the thing that you think will make your time more effective and productive. If you find that people are misusing it enough that you want to add text you can do it as an iterative step later.

ripcurlx (approved):
utACK

PR #3694, opened 2019-11-26
Use 2of2 multisig deposit tx version for manual payout

chimp1984:
With v1.2 we use 2of2 multisig for deposit tx. This commit changes the
manual payout window to reflect that.

- Remove unused code from legacy arbitration
- Fix comments


julianknutsen:
Do either of these need to make it into v1.2.4?

chimp1984:
> Do either of these need to make it into v1.2.4?

Not super important, but this one might be useful in case we need a manual payout with a 1.2. trade.

ripcurlx (approved):
utACK - I'll cherry-pick this into the release branch

PR #3695, opened 2019-11-26
Add shortcut list

chimp1984:
Should be merged after https://github.com/bisq-network/bisq/pull/3694 (not hard requirement though)

### DIFF
--- a/desktop/src/main/java/bisq/desktop/app/BisqApp.java
+++ b/desktop/src/main/java/bisq/desktop/app/BisqApp.java
@@ -275,7 +275,7 @@ public class BisqApp extends Application implements UncaughtExceptionHandler {
                     injector.getInstance(SendAlertMessageWindow.class).show();
                 } else if (Utilities.isAltOrCtrlPressed(KeyCode.F, keyEvent)) {
                     injector.getInstance(FilterWindow.class).show();
-                } else if (Utilities.isAltOrCtrlPressed(KeyCode.UP, keyEvent)) {
+                } else if (Utilities.isAltOrCtrlPressed(KeyCode.H, keyEvent)) {
                     log.warn("We re-published all proposalPayloads and blindVotePayloads to the P2P network.");
                     injector.getInstance(MissingDataRequestService.class).reRepublishAllGovernanceData();
                 } else if (Utilities.isAltOrCtrlPressed(KeyCode.T, keyEvent)) {
--- a/desktop/src/main/java/bisq/desktop/main/funds/transactions/TransactionsView.java
+++ b/desktop/src/main/java/bisq/desktop/main/funds/transactions/TransactionsView.java
@@ -41,8 +41,6 @@ import bisq.core.util.coin.CoinFormatter;
 
 import bisq.network.p2p.P2PService;
 
-import bisq.common.util.Tuple2;
-import bisq.common.util.Tuple4;
 import bisq.common.util.Utilities;
 
 import org.bitcoinj.core.Coin;
@@ -83,16 +81,8 @@ import javafx.collections.transformation.SortedList;
 
 import javafx.util.Callback;
 
-import java.text.DateFormat;
-
-import java.util.ArrayList;
 import java.util.Comparator;
-import java.util.Date;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.stream.Collectors;
 
 import javax.annotation.Nullable;
 
@@ -218,6 +208,7 @@ public class TransactionsView extends ActivatableView<VBox, Void> {
         };
 
         keyEventEventHandler = event -> {
+            // Not intended to be public to users as the feature is not well tested
             if (Utilities.isAltOrCtrlPressed(KeyCode.R, event)) {
                 if (revertTxColumn.isVisible()) {
                     confidenceColumn.getStyleClass().remove("last-column");
@@ -225,8 +216,7 @@ public class TransactionsView extends ActivatableView<VBox, Void> {
                     confidenceColumn.getStyleClass().add("last-column");
                 }
                 revertTxColumn.setVisible(!revertTxColumn.isVisible());
-            } else if (Utilities.isAltOrCtrlPressed(KeyCode.A, event))
-                showStatisticsPopup();
+            }
         };
 
         exportButton.updateText(Res.get("shared.exportCSV"));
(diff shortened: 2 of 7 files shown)


desktop/src/main/java/bisq/desktop/app/BisqApp.java
```diff
@@ -275,7 +275,7 @@ private void addSceneKeyEventHandler(Scene scene, Injector injector) {
                     injector.getInstance(SendAlertMessageWindow.class).show();
                 } else if (Utilities.isAltOrCtrlPressed(KeyCode.F, keyEvent)) {
                     injector.getInstance(FilterWindow.class).show();
-                } else if (Utilities.isAltOrCtrlPressed(KeyCode.UP, keyEvent)) {
```
julianknutsen, line 278: Comments should be updated https://github.com/bisq-network/bisq/blob/3854907c14357680038661c8153095a157efbc5d/core/src/main/java/bisq/core/dao/governance/voteresult/MissingDataRequestService.java#L81

desktop/src/main/java/bisq/desktop/main/funds/transactions/TransactionsView.java
```diff
@@ -218,15 +208,15 @@ public void onKeysAdded(List<ECKey> keys) {
...
         keyEventEventHandler = event -> {
+            // Not intended to be public to users as the feature is not well tested
             if (Utilities.isAltOrCtrlPressed(KeyCode.R, event)) {
                 if (revertTxColumn.isVisible()) {
                     confidenceColumn.getStyleClass().remove("last-column");
                 } else {
                     confidenceColumn.getStyleClass().add("last-column");
                 }
                 revertTxColumn.setVisible(!revertTxColumn.isVisible());
-            } else if (Utilities.isAltOrCtrlPressed(KeyCode.A, event))
```
julianknutsen, line 228: Is there an issue tracking the effort to update the website with the new fee information? For example, the Fees section of https://docs.bisq.network/exchange/whitepaper.html still lists that part of the fee goes to the arbitrator.

julianknutsen (commented):
NIT and questions

desktop/src/main/java/bisq/desktop/app/BisqApp.java
chimp1984: Thanks.

desktop/src/main/java/bisq/desktop/main/funds/transactions/TransactionsView.java
chimp1984: Oh the doc is not updated then. @m52go Could you have a look?

BTC fees go to the Bisq Donation address defined in the DAO (can be changed by voting). From that the owner of that address will buy BSQ with the BTC and burn the BSQ, so converting the fee back into the BSQ ecosystem.

m52go: That doc is obsolete [and archived](https://docs.bisq.network/archive.html).

See "Note on BTC revenues" [here](https://docs.bisq.network/user-dao-intro.html#traders-consume-burning-bsq).

julianknutsen: Maybe we should ```<noindex>``` it then? It is the second search result, and featured snippet, for "bisq fees" and when I was trying to understand the system more it was the first place I went to. I now see that the text at the top says it is archived, but with the search results jumping directly to sections it may cause confusion for new users. Just my 2c.

m52go: Good idea, will do...I had no idea that document was ranking so high in search results.

freimair (dismissed):
ahm I am not sure if I got your intentions but that is what caught my attention:

- in `AccountView` pressing `... + n` creates a new refund agent tab. Yet, the text says `arbitrator`.
- there is no code to create a new arbitrator-tab, can we get rid of the remnants?

It seems there is some tangling up of "arbitrator" and "refund agent". We should clear that up to avoid confusion.

minor stuff:
- `Navigate to account and click: {0}` shouldn't it be "press" instead of "click"

ripcurlx:
> It seems there is some tangling up of "arbitrator" and "refund agent". We should clear that up to avoid confusion.

Atm we used in all user facing views `arbitrator` instead of `refund agent` to reduce unnecessary concerns when we switched to the new trade protocol.
If someone tries to register an arbitrator (new refund agent) we labeled it refund agent to make it clearer to the new refund agents that they are in the correct registration view. So I don't think it is a good idea to change the label to arbitrator now to prevent confusion for the real refund agents.
I think maybe in a couple of releases when we get rid of most of the legacy arbitrator parts there might be the point also to rename the refund agents.

> * `Navigate to account and click: {0}` shouldn't it be "press" instead of "click"

NIT - Yes it would be good to be consistent in this view if we want to use click or press.

chimp1984:
@freimair 
> refund agent vs. arbitrator

We use refundAgent in the domain (code) but left arbitrator as the term for users (UI). There are some discussions for a better term, but so far arbitrator (UI) is refund agent and old arbitrator is "legacy arbitrator".

> click

Will change it.

freimair:
> * there is no code to create a new arbitrator-tab, can we get rid of the remnants?

what about dead code?

> but so far arbitrator (UI) is refund agent and old arbitrator is "legacy arbitrator".

seems reasonable. however, why is the refund agent tab then titled "refund agend registration"?

chimp1984:
> > * there is no code to create a new arbitrator-tab, can we get rid of the remnants?
> 
> what about dead code?

Removing the legacy arbitator domain is a bit of a bigger task. There might be some usage in messages, so simply deletion will not be enough.

> 
> > but so far arbitrator (UI) is refund agent and old arbitrator is "legacy arbitrator".
> 
> seems reasonable. however, why is the refund agent tab then titled "refund agend registration"?

"Arbitrator" is for users. "refund agend registration" is for refund agents to make it more clear and to avoid confusion for the refund agent.

chimp1984:
@freimair Can you review again?

freimair:
I am stepping back from reviewing this.

First, IMO, if we remove the short-cut to even trigger a "legacy arbitrator" tab, the code remnants should be removed. This is dead code. If we do not remove it right away, it might stay there until a newcomer asks about what that is.

Second, I do not like the dual vocabulary. It is only confusing.

chimp1984:
@freimair I only wanted to add a small feature as it is annoying to look up the shortcuts in the code each time one needs them. It was not intended to change existing naming schemes (refund agent vs. arbitrator) as there is anyway some discussion going on for a better term. As well it would be far beyond the scope of that PR to remove the domain code of legacy arbitration which carries also some risks and will require lots of testing with backward compatibility. As the risk/effort/benefit ration is pretty bad for that effort I also do not have intention to work on that any time soon. Too much other important stuff waiting to get implemented...

I understand that this discussion here becomes too much of an effort and I am ok with closing that PR as well. Not such an important feature to justify to burn so much time for reviewers and author.

sqrrm:
I will take a look at this as I want the short cut list added (too annoying looking through the code indeed)

ripcurlx (approved):
ACK

Tested on Regtest and everything is working as expected.

PR #3713, opened 2019-11-28
Update dao-setup doc to include bloom filter for newer bitcoincore v0.19.0.1

beingindot:
My dao setup was not synchronizing due to latest bitcoin core updation. 
using @ripcurlx's hint, i have added bloom filter then it worked. so proposed doc changes according to it.

### DIFF
--- a/docs/dao-setup.md
+++ b/docs/dao-setup.md
@@ -16,6 +16,8 @@ regtest=1
 # The default rpcPort for regtest from Bitcoin Core 0.16 and higher is: 18443
 # The default rpcPort for testnet is: 18332
 # For mainnet: 8332
+[regtest]
+peerbloomfilters=1
 rpcport=18443
 
 server=1
@@ -32,6 +34,8 @@ regtest=1
 # The default rpcPort for regtest from Bitcoin Core 0.16 and higher is: 18443
 # The default rpcPort for testnet is: 18332
 # For mainnet: 8332
+[regtest]
+peerbloomfilters=1
 rpcport=18443
 
 server=1
@@ -48,6 +52,8 @@ regtest=1
 # The default rpcPort for regtest from Bitcoin Core 0.16 and higher is: 18443
 # The default rpcPort for testnet is: 18332
 # For mainnet: 8332
+[regtest]
+peerbloomfilters=1
 rpcport=18443
 
 server=1


docs/dao-setup.md
```diff
@@ -56,6 +56,13 @@ rpcuser=YOUR_USER_NAME
 rpcpassword=YOUR_PW
 blocknotify="%AppData%\Bitcoin\blocknotify.bat" %s
 ```
+Note: If you are using bitcoin core version 0.19.0.1
+should add following two lines to the config file before rpcport option.
+```
+[regtest]
+peerbloomfilters=1
```
devinbileck: I believe this option has been available for some time now, but just recently changed to being disabled by default. Therefore, it should be safe to add this parameter regardless if you are still using an older bitcoin core. So rather than adding this as a separate note, just add it into the examples above.

beingindot: Have added into the examples above now. @devinbileck

ripcurlx:
@beingindot Could you please squash it into one commit and label the commit message based on https://chris.beams.io/posts/git-commit suggestions. Also give the PR a more speakable title so it is easier digest when creating a release log. Thanks!

beingindot:
pls check @ripcurlx

docs/dao-setup.md
```diff
@@ -57,6 +61,7 @@ rpcpassword=YOUR_PW
 blocknotify="%AppData%\Bitcoin\blocknotify.bat" %s
 ```
 
+
```
devinbileck: Just curious if this added blank line was intentional/necessary?

docs/dao-setup.md
```diff
@@ -16,6 +16,8 @@ regtest=1
 # The default rpcPort for regtest from Bitcoin Core 0.16 and higher is: 18443
 # The default rpcPort for testnet is: 18332
 # For mainnet: 8332
+[regtest]
```
devinbileck, line 19: Just wondering why this was added only to the Linux example and not the others? It should be fine to add it, but should be consistent. But not sure it is even necessary in the first place since the first line is regtest=1.

beingindot: without this i could not run in ubuntu. but i have not tested in other linux distros though.

docs/dao-setup.md
beingindot: resolved

docs/dao-setup.md
beingindot: pls let me know whether to keep or remove this.

beingindot: @ripcurlx @wiz @Emzy Could you check in your linux or other distros to confirm this option.

beingindot: on further testing, this [regtest] needed only when we set rpcport.

freimair: exactly. here on arch linux. if I want to set regtest-specific options, I have to use the '[regtest]' keyword to specify the config area.

devinbileck: I guess we should add it for the Windows and macOS examples as well.

beingindot: I have updated it now

freimair (approved):
Ack for linux.

ripcurlx (approved):
ACK - I also need for macOSX the `[regtest]` in the settings to be able to run 0.19.1

Bildschirmfoto 2019-12-06 um 16 21 37

PR #3714, opened 2019-11-28
Add zh_hant translations (Chinese Traditional)

wiz:
Screen Shot 2019-11-29 at 14 54 13
Screen Shot 2019-11-29 at 14 54 46
Screen Shot 2019-11-29 at 14 54 35
Screen Shot 2019-11-29 at 14 55 36
Screen Shot 2019-11-29 at 14 56 11
Screen Shot 2019-11-29 at 14 56 34
Screen Shot 2019-11-29 at 14 57 03
Screen Shot 2019-11-29 at 14 57 25
Screen Shot 2019-11-29 at 15 08 12
Screen Shot 2019-11-29 at 15 08 46
Screen Shot 2019-11-29 at 15 09 11
Screen Shot 2019-11-29 at 15 09 59
Screen Shot 2019-11-29 at 15 11 22

### DIFF
--- a/core/update_translations.sh
+++ b/core/update_translations.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 cd $(dirname $0)
-tx pull -l de,el_GR,es,ja,pt,ru,zh_CN,vi,th_TH,fa,fr,pt_BR
+tx pull -l de,el_GR,es,ja,pt,ru,zh_CN,zh_TW,vi,th_TH,fa,fr,pt_BR
 
 translations="translations/bisq-desktop.displaystringsproperties"
 i18n="src/main/resources/i18n"
@@ -13,6 +13,7 @@ mv "$translations/ja.properties" "$i18n/displayStrings_ja.properties"
 mv "$translations/pt.properties" "$i18n/displayStrings_pt.properties"
 mv "$translations/ru.properties" "$i18n/displayStrings_ru.properties"
 mv "$translations/zh_CN.properties" "$i18n/displayStrings_zh-hans.properties"
+mv "$translations/zh_TW.properties" "$i18n/displayStrings_zh-hant.properties"
 mv "$translations/vi.properties" "$i18n/displayStrings_vi.properties"
 mv "$translations/th_TH.properties" "$i18n/displayStrings_th.properties"
 mv "$translations/fa.properties" "$i18n/displayStrings_fa.properties"
--- a/core/src/main/java/bisq/core/locale/LanguageUtil.java
+++ b/core/src/main/java/bisq/core/locale/LanguageUtil.java
@@ -36,6 +36,7 @@ public class LanguageUtil {
             "pt", // Portuguese
             "pt-BR", // Portuguese (Brazil)
             "zh-Hans", // Chinese [Han Simplified]
+            "zh-Hant", // Chinese [Han Traditional]
             "ru", // Russian
             "fr", // French
             "vi", // Vietnamese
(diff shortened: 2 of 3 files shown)


core/update_translations.sh
```diff
@@ -12,11 +12,12 @@ mv "$translations/es.properties" "$i18n/displayStrings_es.properties"
 mv "$translations/ja.properties" "$i18n/displayStrings_ja.properties"
 mv "$translations/pt.properties" "$i18n/displayStrings_pt.properties"
 mv "$translations/ru.properties" "$i18n/displayStrings_ru.properties"
-mv "$translations/zh_CN.properties" "$i18n/displayStrings_zh.properties"
+mv "$translations/zh_CN.properties" "$i18n/displayStrings_zh-hans.properties"
+mv "$translations/zh_TW.properties" "$i18n/displayStrings_zh-hant.properties"
```
ripcurlx, line 16: To support that you need to update also `tx pull -l de,el_GR,es,ja,pt,ru,zh_CN,vi,th_TH,fa,fr,pt_BR` to `tx pull -l de,el_GR,es,ja,pt,ru,zh_CN,zh_TW,vi,th_TH,fa,fr,pt_BR`

ripcurlx (changes requested):
NACK - @wiz Thanks for the changes. I don't remember anymore if there was a reason not to support country specific language files, but I don't see a reason why not to do it. Please see my comment regarding one missing change.

ripcurlx:
@arunasurya @wiz We do have Chinese simplified already as a supported core language, but not traditional Chinese. How do we want to handle the adding of country variants of an existing language? I think it would be good to have something official (proposal or similar) to state the reasoning for adding this new language variant (as it also immediately makes it eligible for compensation)

arunasurya:
@ripcurlx , I agree with you that it is good to have clear reasoning for adding a new language. Right now, we don't have a clear policy but rather rough guidelines. Perhaps it is a good idea to have a more explicit policy. Meanwhile, I am communicating with the two translators who have done (ff98sha and wdljt) to figure out how important traditional Chinese is for Bisq, etc.

cryptosnip:
1.4 Billion people speak & write Mandarin - 1/3 the worlds population. So what is the the problem??Traditional Chinese is mostly written and is used in academic and traditional medicine and arts/music...

-- 
 Securely sent with Tutanota. Get your own encrypted, ad-free mailbox: 
 https://tutanota.com


Dec 3, 2019, 03:31 by notifications@github.com:

ripcurlx:
> 1.4 Billion people speak & write Mandarin - 1/3 the worlds population. So what is the the problem??

There is no problem with it, we just should have a transparent workflow when a new language or language variant is added as we only support a couple of core languages right now which are also up for compensation. Just to change it within a PR might go through unnoticed and might cause irritation by others who want to get their language added to Bisq as well.

ripcurlx:
> @ripcurlx , I agree with you that it is good to have clear reasoning for adding a new language. Right now, we don't have a clear policy but rather rough guidelines. Perhaps it is a good idea to have a more explicit policy. Meanwhile, I am communicating with the two translators who have done (ff98sha and wdljt) to figure out how important traditional Chinese is for Bisq, etc.

I think it is important, but it was only communicated through personal chat with me from @wiz. So I'm up for adding it in this case as it is just a language variant and it would add native language support for people based in Hongkong.

ripcurlx (dismissed):

ripcurlx (approved):
ACK

Tested it locally on Regtest and also tried to update translations from Transifex. Everything works as expected.

arunasurya:
I think that the translators should have made a proposal to add a new language before it was merged. I was looking into it and tried to evaluate if it should be  a core language. I don't think there was a consensus on adding traditional Chinese from all sides...Is it now officially a core language? 

P.S. I think we need to have a clearer process of adding a new language. I have stared writing it up in the Bisq translator documentation (on my github site) but it is not official since the documentation is not on Bisq. I guess there was a lack of clear communication on my part. I would be happy to talk to the translators and come up with a more concrete process.

PR #3716, opened 2019-11-29
Use correct language tags for script and regional language variants

wiz:
This PR fixes Bisq's language tags so Java's locale class will correctly display the name of script and regional variants of languages. There were previously hard-coded hacks to correctly display the language names correctly, and this PR removes those hacks since they are now properly displayed by Java.

It's actually a common misunderstanding in software localization:

* The language's regional variant, which is often a country, is separated with a dash.
* The user's country is separated with an underscore.

For some people, the difference is subtle:
"en-us" = American English
"en_US" = generic English in USA
"en-gb" = British English
"en-gb_US" = British English in USA

But for other people, the difference is very important:
"br" = Portugese
"br-pt" = Brazilian Portugese
"br-pt_JP" = Brazilian Portugese in Japan

In this case, to merge #3714, I need the following script variant:
"zh" = Chinese
"zh-hans" = Simplified Chinese
"zh-hant" = Traditional Chinese

Needs testing!


ripcurlx (approved):
ACK

Tested it locally on Regtest switching languages between pt, pt-BR and zh-Hans with app restart and everything worked as expected.

PR #3718, opened 2019-11-30
Add Makefile for automating localnet setup

cbeams:
Problem: contributors old and new must read and follow many manual steps
spread across three documents (docs/{build,dev-setup,dao-setup}.md) in
order to get up and running with a local regtest Bisq network deployment
suitable for isolated development and end-to-end testing. This process
is not only manual, but requires considerable trial and error for most
contributors, and can amount to hours of effort. Perhaps most
detrimental is that this friction makes it much less likely that we get
"all hands on deck" to cover test scenarios at release time. Getting up
and running with what this change refers to as a "localnet" should be
among the very first things a new contributor does. It should be fast
and easy, maximizing the contributor's ability to get productive right
away.

Solution: this commit introduces a simple and well-documented makefile
to the root of the source tree. It instructs the user to issue a series
of simple `make` commands, at the end of which they'll have a fully
functional localnet deployment.

Caveats:

 - No support for Windows unless the user is running Git Bash, Cygwin or
   similar. In any case, the makefile serves as clear documentation
   about what a Windows user would need to do manually, i.e. without the
   benefit of `make` automating it all.

 - The aforementioned setup documents should be updated to point to this
   makefile instead of explaining everything in prose. The dev-setup.md
   and dao-setup.md documents may actually be candidates for deletion if
   this new approach proves successful.

 - These changes do not include passing the new -peerbloomfilters=1
   option to bitcoin versions 0.19 and above. Those who have already
   upgraded should take care to add that option.

Notes:

 - The introduction of this makefile has no impact on Bisq's use of
   Gradle as a build system. Everything there is as it has been. This
   makefile is a completely optional convenience being added into the
   mix. It has the added benefit of being a "friendly face" to those not
   familiar with the Java / JVM ecosystem. Developers from many
   different backgrounds are familiar with make and makefiles, and they
   may find this one a pleasant and inviting surprise.

----

Special thanks to @bodymindarts for the inspiration to take this makefile-based approach. The [makefile in his bisq-workspace repo](https://github.com/bodymindarts/bisq-workspace/blob/master/Makefile) plus the pain I was experiencing trying to help out with v1.2.4 testing was what got this ball rolling.

For those interested in putting this to use in your current testing efforts, this PR is branched from the last common commit between the `master` and `release/v1.2.4` branches, so you can merge it into your own local `release/v1.2.4 branch` or just cherry-pick the single commit. Both should work cleanly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,10 +153,15 @@ deploy: setup
 	# create a new screen session named 'localnet'
 	screen -dmS localnet
 	# deploy each node in its own named screen window
-	targets=('bitcoind' 'seednode' 'seednode2' 'alice' 'bob' 'mediator'); \
-	for t in "$${targets[@]}"; do \
-		screen -S localnet -X screen -t $$t; \
-		screen -S localnet -p $$t -X stuff "make $$t\n"; \
+	for target in \
+			bitcoind \
+			seednode \
+			seednode2 \
+			alice \
+			bob \
+			mediator; do \
+		screen -S localnet -X screen -t $$target; \
+		screen -S localnet -p $$target -X stuff "make $$target\n"; \
 	done;
 	# give bitcoind rpc server time to start
 	sleep 5
--- a/Makefile
+++ b/Makefile
@@ -43,12 +43,12 @@
 #     bisq-seednode
 #     bisq-statsnode
 #
-#  - You will see a new 'localnet' directory containing the data dirs
+#  - You will see a new '.localnet' directory containing the data dirs
 #  for your regtest Bitcoin and Bisq nodes. Once you've deployed
 #  them in the step below, the directory will look as follows:
 #
-#     $ tree -d -L 1 localnet
-#     localnet
+#     $ tree -d -L 1 .localnet
+#     .localnet
 #     ├── alice
 #     ├── bitcoind
 #     ├── bob
@@ -101,6 +101,7 @@
 # cases.
 #
 
+STATE_DIR := .localnet
 
 # Set up everything necessary for deploying your localnet. This is the
 # default target.
@@ -112,7 +113,7 @@ clean-build:
 	./gradlew clean
 
 clean-localnet:
-	rm -rf localnet
+	rm -rf $(STATE_DIR)
 
 # Build all Bisq binaries and generate the shell scripts used to run
 # them in the targets below
@@ -127,16 +128,16 @@ localnet:
 	# and intuitive naming. This is a temporary measure until we clean these
 	# resources up more thoroughly.
 	unzip docs/dao-setup.zip
-	mv dao-setup localnet
-	mv localnet/Bitcoin-regtest localnet/bitcoind
-	mv localnet/bisq-BTC_REGTEST_Alice_dao localnet/alice
-	mv localnet/bisq-BTC_REGTEST_Bob_dao localnet/bob
+	mv dao-setup $(STATE_DIR)
+	mv $(STATE_DIR)/Bitcoin-regtest $(STATE_DIR)/bitcoind
+	mv $(STATE_DIR)/bisq-BTC_REGTEST_Alice_dao $(STATE_DIR)/alice
+	mv $(STATE_DIR)/bisq-BTC_REGTEST_Bob_dao $(STATE_DIR)/bob
 	# Remove the preconfigured bitcoin.conf in favor of explicitly
 	# parameterizing the invocation of bitcoind in the target below
-	rm -v localnet/bitcoind/bitcoin.conf
+	rm -v $(STATE_DIR)/bitcoind/bitcoin.conf
 	# Avoid spurious 'runCommand' errors in the bitcoind log when nc
 	# fails to bind to one of the listed block notification ports
-	echo exit 0 >> localnet/bitcoind/blocknotify
+	echo exit 0 >> $(STATE_DIR)/bitcoind/blocknotify
 
 # Deploy a complete localnet by running all required Bitcoin and Bisq
 # nodes, each in their own named screen window. If you are not a screen
@@ -163,8 +164,8 @@ bitcoind: localnet
 		-server \
 		-rpcuser=bisqdao \
 		-rpcpassword=bsq \
-		-datadir=localnet/bitcoind \
-		-blocknotify='localnet/bitcoind/blocknotify %s'
+		-datadir=$(STATE_DIR)/bitcoind \
+		-blocknotify='$(STATE_DIR)/bitcoind/blocknotify %s'
 
 seednode: build
 	./bisq-seednode \
@@ -176,7 +177,7 @@ seednode: build
 		--rpcPassword=bsq \
 		--rpcBlockNotificationPort=5120 \
 		--nodePort=2002 \
-		--userDataDir=localnet \
+		--userDataDir=$(STATE_DIR) \
 		--appName=seednode
 
 seednode2: build
@@ -189,7 +190,7 @@ seednode2: build
 		--rpcPassword=bsq \
 		--rpcBlockNotificationPort=5121 \
 		--nodePort=3002 \
-		--userDataDir=localnet \
+		--userDataDir=$(STATE_DIR) \
 		--appName=seednode2
 
 mediator: build
@@ -198,7 +199,7 @@ mediator: build
 		--useLocalhostForP2P=true \
 		--useDevPrivilegeKeys=true \
 		--nodePort=4444 \
-		--appDataDir=localnet/mediator \
+		--appDataDir=$(STATE_DIR)/mediator \
 		--appName=Mediator
 
 alice: setup
@@ -213,7 +214,7 @@ alice: setup
 		--rpcBlockNotificationPort=5122 \
 		--genesisBlockHeight=111 \
 		--genesisTxId=30af0050040befd8af25068cc697e418e09c2d8ebd8d411d2240591b9ec203cf \
-		--appDataDir=localnet/alice \
+		--appDataDir=$(STATE_DIR)/alice \
 		--appName=Alice
 
 bob: setup
@@ -222,7 +223,7 @@ bob: setup
 		--useLocalhostForP2P=true \
 		--useDevPrivilegeKeys=true \
 		--nodePort=6666 \
-		--appDataDir=localnet/bob \
+		--appDataDir=$(STATE_DIR)/bob \
 		--appName=Bob
 
 # Generate a new block on your Bitcoin regtest network. Requires that
--- a/Makefile
+++ b/Makefile
@@ -124,7 +124,7 @@ build:
 # Unpack and customize a Bitcoin regtest node and Alice and Bob Bisq
 # nodes that have been preconfigured with a blockchain containing the
 # BSQ genesis transaction
-localnet: clean-localnet
+localnet:
 	# Unpack the old dao-setup.zip and move things around for more concise
 	# and intuitive naming. This is a temporary measure until we clean these
 	# resources up more thoroughly.
--- a/Makefile
+++ b/Makefile
@@ -168,6 +168,14 @@ deploy: setup
 	# generate a block to ensure Bisq nodes get dao-synced
 	make block
 
+# Undeploy a running localnet by killing all Bitcoin and Bisq
+# node processes, then killing the localnet screen session altogether
+undeploy:
+	# kill all Bitcoind and Bisq nodes running in screen windows
+	screen -S localnet -X at "#" stuff "^C"
+	# quit all screen windows which results in killing the session
+	screen -S localnet -X at "#" kill
+
 bitcoind: .localnet
 	bitcoind \
 		-regtest \
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,242 @@
+#
+# INTRODUCTION
+#
+# This makefile is designed to help Bisq contributors get up and running
+# as quickly as possible with a local regtest Bisq network deployment,
+# or 'localnet' for short. A localnet is a complete and self-contained
+# "mini Bisq network" suitable for development and end-to-end testing
+# efforts.
+#
+#
+# REQUIREMENTS
+#
+# You'll need the following to proceed:
+#
+#  - Linux, macOS or similar *nix with standard tools like `make`
+#  - bitcoind and bitcoin-cli (`brew install bitcoin` on macOS)
+#  - JDK 10 to build and run Bisq binaries; see
+#    https://www.oracle.com/java/technologies/java-archive-javase10-downloads.html
+#
+#
+# USAGE
+#
+# The following commands (and a couple manual instructions) will get your
+# localnet up and running quickly.
+#
+# STEP 1: Build all Bisq binaries and set up localnet resources. This will
+# take a few minutes the first time through.
+#
+#     $ make
+#
+# Notes:
+#
+#  - When complete, you'll have a number of scripts available in the
+#  root directory. They will be used in the make targets below to start
+#  the various Bisq seed and desktop nodes that will make up your
+#  localnet:
+#
+#     $ ls -1 bisq-*
+#     bisq-desktop
+#     bisq-monitor
+#     bisq-pricenode
+#     bisq-relay
+#     bisq-seednode
+#     bisq-statsnode
+#
+#  - You will see a new 'localnet' directory containing the data dirs
+#  for your regtest Bitcoin and Bisq nodes. Once you've deployed
+#  them in the step below, the directory will look as follows:
+#
+#     $ tree -d -L 1 localnet
+#     localnet
+#     ├── alice
+#     ├── bitcoind
+#     ├── bob
+#     ├── mediator
+#     ├── seednode
+#     └── seednode2
+#
+# STEP 2: Deploy the Bitcoin and Bisq nodes that make up the localnet.
+# Run each of the following in a SEPARATE TERMINAL WINDOW, as they are
+# long-running processes.
+#
+#     $ make bitcoind
+#     $ make seednode
+#     $ make seednode2
+#     $ make mediator
+#     $ make alice
+#     $ make bob
+#
+#  Tip: Those familiar with the `screen` terminal multiplexer can
+#  automate the above by running the `deploy` target found below.
+#
+#  Notes:
+#
+#    - The 'seednode' targets launch headless Bisq nodes that help
+#    desktop nodes discover other peers, as well as storing and
+#    forwarding p2p network messages for nodes as they go on and
+#    offline.
+#
+#    - As you run the 'mediator', 'alice' and 'bob' targets above,
+#    you'll see a Bisq desktop node window appear for each. The Alice
+#    and Bob instances represent two traders who can make and take
+#    offers with one another. The Mediator instance represents a Bisq
+#    contributor who can help resolve any technical problems or disputes
+#    that come up between the two traders.
+#
+# STEP 3: Configure the mediator Bisq node. In order to make and take
+# offers, Alice and Bob will need to have a mediator and a refund agent
+# registered on the network. Follow the instructions below to complete
+# that process:
+#
+#  a) Go to the Account screen in the Mediator instance and press CMD+N
+#  and a popup will appear. Click 'Unlock' and then click 'Register' to
+#  register the instance as a mediator.
+#
+#  b) While still in the Account screen, press CMD+D and follow the same
+#  steps as above to register the instance as a refund agent.
+#
+# When the steps above are complete, your localnet should be up and
+# ready to use. You can now test in isolation all Bisq features and use
+# cases.
+#
+
+
+# Set up everything necessary for deploying your localnet. This is the
+# default target.
+setup: build localnet
+
+clean: clean-build clean-localnet
+
+clean-build:
+	./gradlew clean
+
+clean-localnet:
+	rm -rf localnet
+
+# Build all Bisq binaries and generate the shell scripts used to run
+# them in the targets below
+build:
+	./gradlew build
+
+# Unpack and customize a Bitcoin regtest node and Alice and Bob Bisq
+# nodes that have been preconfigured with a blockchain containing the
+# BSQ genesis transaction
+localnet:
+	# Unpack the old dao-setup.zip and move things around for more concise
+	# and intuitive naming. This is a temporary measure until we clean these
+	# resources up more thoroughly.
+	unzip docs/dao-setup.zip
+	mv dao-setup localnet
+	mv localnet/Bitcoin-regtest localnet/bitcoind
+	mv localnet/bisq-BTC_REGTEST_Alice_dao localnet/alice
+	mv localnet/bisq-BTC_REGTEST_Bob_dao localnet/bob
+	# Remove the preconfigured bitcoin.conf in favor of explicitly
+	# parameterizing the invocation of bitcoind in the target below
+	rm -v localnet/bitcoind/bitcoin.conf
+	# Avoid spurious 'runCommand' errors in the bitcoind log when nc
+	# fails to bind to one of the listed block notification ports
+	echo exit 0 >> localnet/bitcoind/blocknotify
+
+# Deploy a complete localnet by running all required Bitcoin and Bisq
+# nodes, each in their own named screen window. If you are not a screen
+# user, you'll need to run each of the make commands manually in a
+# separate terminal or as a background job.
+#
+# NOTE: You MUST already be attached to a screen session for the
+# following commands to work properly.
+deploy: setup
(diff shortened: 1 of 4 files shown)


cbeams:
I just realized that the commit comment and documentation in the makefile doesn't necessarily make clear how concise the process can be, especially for those comfortable using `screen`. These are the steps in a nutshell:

```
$ screen
$ make
$ make deploy
```

UPDATE: As of the latest changes, this is now as simple as:
```
$ make deploy
```

cbeams:
An additional note/caveat not mentioned in the original message. This implementation works with the `dao-setup.zip` file, but most if not all are agreed we'd like to see the full automation of a from-scratch genesis tx. @KanoczTomas has been working on this approach, as can be seen in https://github.com/bisq-network/bisq/tree/master/docs/autosetup-regtest-dao. If feasible, it would be desirable to eliminate the pre-fab zip altogether and replace it with equally automated steps for setting everything up from scratch. This may require Bisq's gRPC API to be in place for certain aspects; fortunately that work is underway as we speak!

Makefile
```diff
@@ -0,0 +1,242 @@
...
+# following commands to work properly.
+deploy: setup
+	screen -t bitcoin make bitcoind
+	sleep 2    # wait for bitcoind rpc server to start
+	make block # generate a block to ensure Bisq nodes get dao-synced
+	screen -t seednode make seednode
+	screen -t seednode2 make seednode2
+	screen -t alice make alice
+	screen -t bob make bob
+	screen -t mediator make mediator
```
julianknutsen: This is close to what I have locally just in a .rc file. Does bisq not handle bitcoind not having the rpc server up first? I've never run into a problem, but I always start bitcoind first so maybe I just got lucky?

Makefile
```diff
@@ -0,0 +1,242 @@
...
+		--useDevPrivilegeKeys=true \
+		--fullDaoNode=true \
+		--rpcUser=bisqdao \
+		--rpcPassword=bsq \
+		--rpcBlockNotificationPort=5120 \
+		--nodePort=2002 \
+		--userDataDir=localnet \
+		--appName=seednode
+
+seednode2: build
```
julianknutsen: Is the second seednode just for more testing of p2p forwarding? My setup has always been fine with just one so curious as to the extra benefit and if it is worth the resources.

Makefile
```diff
@@ -0,0 +1,242 @@
...
+		--baseCurrencyNetwork=BTC_REGTEST \
+		--useLocalhostForP2P=true \
+		--useDevPrivilegeKeys=true \
+		--nodePort=6666 \
+		--appDataDir=localnet/bob \
+		--appName=Bob
+
+# Generate a new block on your Bitcoin regtest network. Requires that
+# bitcoind is already running. See the `bitcoind` target above.
+block:
```
julianknutsen, line 252: I think this could use another sentence or so in the preamble. You end up needing to create blocks to test features like governance so helping new users fix common errors like "I created a proposal from Alice, but it isn't visible on Bob. Why not?" may help the onboarding.

julianknutsen (commented):
I'm actually of the opinion that it is OK if devs don't understand all of the internals of the dao setup and genesis transactions on their first day when they are just trying to get the software up and running to mess around with the features.

This is a great first step and the docs should be updated to recommend this as opposed to the list of manual commands. Scripts like these are easy to bikeshed about, but erring on the side of getting something committed that people can use to save time seems prudent. It is easy to add features later than can be driven by more dev use cases.

The next iteration should probably do everything from scratch so the zip file can be deprecated.  There are already issues with the zip file having different default accounts due to the pre-existing data. It is just error-prone to maintain default persistent data.

Going forward it seems like doing everything from scratch and utilizing the gRPC system to automate the "default" pieces that are time-consuming makes a lot of sense. It would be great to see things like default accounts added by a set of gRPC commands when that part of the API becomes available. Internal use cases are good drivers of features because the acceptance criteria is well-defined and the users are developers who can give feedback faster than typical Bisq users.

I've never seen a Makefile used in this type of manner, but as long as the docs help people unfamiliar with make know which commands to run it seems fine.

Makefile
chimp1984: Yes bitcoind must run first. If the seed runs as full DAO node bitcoind must run before the seed node.

Makefile
chimp1984: Only one seed node is ok.

bodymindarts: One seednode is okay but it leads to annoying error output in all the logs unless you remove one of [these lines](https://github.com/bisq-network/bisq/blob/8c8c7117d33c738bad86199a865d46bca7ddc9fd/core/src/main/resources/btc_regtest.seednodes#L16-L17)

Makefile
bodymindarts: Yeah that is why in the original [bisq-workspace](https://github.com/bodymindarts/bisq-workspace) I start bitcoind with a script that doesn't complete [until bitcoind is responsive](https://github.com/bodymindarts/bisq-workspace/blob/master/scripts/start_bitcoind.sh#L39-L49).

Makefile
cbeams: > One seednode is okay but it leads to annoying error output in all the logs

This is the main reason I spin it up. Gets everything closer to a no-broken-windows situation where the operator can assume there shouldn't be any (or many) errors / stack traces showing up in their Bitcoin and Bisq node logs.

Correct solution would perhaps be to do issuing a warning in the 2002 seednode log that the other well-known (3002) seednode cannot be found, and stopping trying after a reasonable several attempts instead of perpetually issuing error messages.

In the meantime, the `seednode2` target could be documented as optional and an explanatory note could be added about running to avoid the errors mentioned above.

The other reason I wanted a second seed node was to test what happens when one seednode is a dao fullnode and the other isn't. Seems to work fine, but the current configuration in the makefile spins them both up as fullnodes anyway.

Note that this is the same reason why `alice` is a dao fullnode but `bob` and `mediator` aren't: just to make sure we have this heterogenous setup in the out-of-the box localnet config. It better reflects the actual state of the production network, might help catch any issues.

Makefile
cbeams: Aside from @chimp1984's point that bitcoind should run before the seednode in any case, the actual reason I introduced the `sleep 2` and `make block` entries here is because if a new block (block 112) doesn't get created prior to starting the desktop nodes, they spin forever on DAO synchronization (which is annoying to look at but also eats up CPU, spins up fans and makes everything feel heavyweight). I believe this is because the desktop nodes need to actually receive a block notification, but it might also be that they need to have at least one confirmation on the genesis tx in block 111, maybe both. In any case, without block 112, there is a strange status message in `DAO->BSQ wallet->transactions` that reads "Awaiting synchronization... Validated 111 of 0 blocks". So something is off there. To see this behavior for yourself, just comment out the `sleep` and `make block` lines and run through the makefile instructions as usual.

bodymindarts: That makes a lot of sense. I think it should however be the responsibility of the `bitcoind` command to spin it up in a way that is consistent for the other commands to consume.

When adding an alternative `deploy-tmux` command for example that logic has to be duplicated.

Is it still a problem if `block 112` isn't present when the desktop client is spun up but appears very soon thereafter?

Makefile
```diff
@@ -0,0 +1,242 @@
...
+	screen -t bob make bob
+	screen -t mediator make mediator
+
+bitcoind: localnet
+	bitcoind \
+		-regtest \
+		-prune=0 \
+		-txindex=1 \
+		-server \
+		-rpcuser=bisqdao \
```
beingindot, line 186: dont we need to add rpcport here?

Makefile
```diff
@@ -0,0 +1,242 @@
...
+#  - You will see a new 'localnet' directory containing the data dirs
+#  for your regtest Bitcoin and Bisq nodes. Once you've deployed
+#  them in the step below, the directory will look as follows:
+#
+#     $ tree -d -L 1 localnet
+#     localnet
+#     ├── alice
+#     ├── bitcoind
+#     ├── bob
+#     ├── mediator
```
beingindot, line 55: I think mediator and seednode folders come after step 2.

julianknutsen (commented):
I pulled this down and used it for a bit of testing. Here are a few UX comments.

Not everyone has bitcoind and bitcoin-cli in the path. Fixed this locally with symlinks, but since this doesn't cover installing bitcoind it is a sharp edge.

`make clean && make build` should rebuild. I think Justin had comments on this too?

Having a way to reset the node state, but keep the build state, seems like a nice-to-have. Rebuilding everything to reset a test from scratch isn't very efficient.

Ctrl+C inside screen tab closes it. There are quite a few test cases where you just need to take down a node temporarily or restart it. I use the pattern below in my previous localnet .rc file. It may not be optimal but gave me what I needed. The new way requires `ctrl+c -> goto tab 0 -> type: screen -t alice make alice` instead of `ctrl+c -> up arrow -> enter`
```
screen -t bisq-bob
select 4
stuff "/home/julian/bisq/bisq-desktop --userDataDir=/home/julian/dao/dao-setup/ --daoActivated=true --genesisBlockHeight=111 --genesisTxId=30af0050040befd8af25068cc697e418e09c2d8ebd8d411d2240591b9ec203cf --baseCurrencyNetwork=BTC_REGTEST --useDevPrivilegeKeys=true --useLocalhostForP2P=true --nodePort=8888 --appName=bisq-BTC_REGTEST_Bob_dao --fullDaoNode=true --rpcUser=bisq --rpcPassword=bisq --rpcPort=1"443 --rpcBlockNotificationPort=5123
```

Makefile
cbeams: > dont we need to add rpcport here?

No, we're assuming the regtest default value of 18444 everywhere.

Makefile
cbeams: That's correct, and to this end, I mention in the documentation blurb directly above:

> Once you've deployed [all Bitcoind and Bisq nodes] in the step below, the directory will look as follows:

Makefile
cbeams: > Is it still a problem if `block 112` isn't present when the desktop client is spun up but appears very soon thereafter?

It is a problem in the sense that if 112 doesn't show up quickly, the contributor using the makefile is going to hear their fans spin up as the dao synchronization wait loop eats up CPU (we should obviously try to solve that problem at the root, but in the meantime...).

So if we're talking about the difference between generating block 112 just after deploying bitcoind vs. doing it just after deploying all the nodes, i.e. a difference measured in (sub)seconds, then it's not a problem in practice. Doing something like @KanoczTomas's start_bitcoind wrapper (which I believe you used elsewhere, @bodymindarts) might be a solution, but I've avoided any such shell scripts and/or variables thus far with the intention of making everything that's being done absolutely clear to the reader because it's all in one place (the makefile), free of any indirections or abstractions they need to dig into.

cbeams:
> `make clean && make build` should rebuild. I think Justin had comments on this too?

Yes. Changes are incoming that address this.

> Having a way to reset the node state, but keep the build state, seems like a nice-to-have. Rebuilding everything to reset a test from scratch isn't very efficient.

Agreed. You may have sync'd an earlier rev of the makefile, but now you'll see that in addition to the global `clean` that blows away both node and build state, there are also `clean-localnet` and `clean-build` that do the same for each.

> Ctrl+C inside screen tab closes it. There are quite a few test cases where you just need to take down a node temporarily or restart it.

Right, glad you caught this. I have `zombie` configured in my .screenrc such that when the process in a given window is killed, the window does not close, but is preserved in a "zombie" state such that `^[` kills it fully or pressing `@` resurrects it, in this case by re-running the original command, e.g. `make alice`, `make bitcoind` or whatever. This arrangement has been productive enough for me to roll with, but I forgot that this `zombie` configuration is not a default in screen. Here it is if you're interested:

    zombie "^["

More to the point, though, I'm working on improvements to the way screen is invoked by make that will naturally preserve the window when its process dies, allowing the user to get back to the natural `ctrl+c -> up arrow -> enter` workflow you mentioned and that we all likely want.

Makefile
bodymindarts: Yes I understand the intention with having it all self-contained in the makefile and like the approach.
It seems to be a trade trade-off between ease-of-use (via a wrapper script) and 'time searching for the exact commands being executed'. One is more focused on the developer who wants a streamlined workflow, the other on a first-timer just seeing how things work.
In general I like being explicit (ie. no wrapper or indirection) and try to achieve that whenever possible, but in my experience (and I use a Makefile in every single project I do) when there are more than a few commands being executed or there are other complications having 1 level of indirection, like:
```
something:
    scripts/do_something.sh
```
is an acceptable trade off, especially if it improves the dev-workflow (and active devs are probably the primary user of this interface). For first timers having to jump into the scripts folder for more information should be fine.

Another benefit is having the individual make commands closer together in the Makefile, which means not having to scroll so far when trying to assess what all the main commands are that are used to interface with the project as a dev. This also benefits first-time contributors.

bodymindarts:
I'd like to try and formulate an explicit set of use-cases and requirements.
This is just my intuition of how I would structure my dev-workflow:

- Bring up the a self-contained and consistent multi-node setup to test something
- Simulate passing of time (via blocks) when testing workflows that require confirmed transactions
- Be able to iterate on some feature or bug fix quickly
  This probably means, make a change to the source code, rebuild and then re-start just 1 of the nodes running the new source code. Usually when making changes I don't need all the nodes to be updated to try out if my change had the desired effect.
- Reset the state back to a some known situation for reproducing bugs (so somekind of snapshotting capability. After all the initial localnet setup is essentially loading a snapshot, perhaps this could be generalized).

I have not been able to determine if all of these are supported in a streamlined way and am just writting this as documentation (anyone else can add other workflows they think are important).

I like using tmux instead of screen and will add support for tmux once the screen workflows have stabalized.

cbeams:
Ok, I believe the latest commits here address most if not all of the feedback received so far. Please sync up, take it for another spin and let me know if you run into any problems. If everything works without error, I'd like to call this iteration "good enough", such that the PR can be merged and I can return to gRPC work. Further tweaks and improvements will no doubt be in order, but to the degree they're "nice-to-haves" let's try to manage them as subsequent PRs.

Here's what I consider the basic set of use cases that is supported right now, i.e. what should "just work" when you're using the makefile:

1. The _get started from scratch_ use case: `make deploy` in a fresh local bisq clone (or your freshly `clean`ed clone) should do everything from A to Z: it should build the desktop and seednode binaries, unpack and customize the dao-setup.zip file into the `.localnet` directory, and it should deploy bitcoind and all necessary Bisq nodes into a screen session named `localnet`.
2. The _I want to get started from scratch without `screen`_ use case: `make` followed by running each node deployment target, e.g. `make bitcoind`, `make seednode`, etc. in a separate window should get you up and running with everything.
3. The _I want to start over with the state of my nodes but I don't want to rebuild the binaries_ use case. `make clean-localnet deploy` will take care of your needs. Don't forget to kill all your existing node processes and `screen` session (if any) first, though.
4. The _I want to iterate on development and redeploy just one desktop node while leaving the rest of my localnet up and running_ use case. Kill the desktop process in question, e.g. 'alice', make your changes to desktop sources and then run `./gradlew :desktop:build` followed by `make alice`. This will trigger the desktop build and deploy the alice node. (NOTE: edited as per https://github.com/bisq-network/bisq/pull/3718#discussion_r353116322)
5. ... there are probably other nameable use cases that are supported in the current makefile, but these are the basics. Feel free to add to this list, but again, anything net new should probably be managed separately from this PR.

Remember, this stuff doesn't replace the gradle build and doesn't intend to become a comprehensive abstraction over it. It should handle the basics for new contributors who have enough to figure out without grokking gradle. Once you're into an actual development task, you should feel comfortable using gradle more surgically for whatever you need it to do, just as is expected today.

Makefile
cbeams: +1, thanks. Where stuff gets unwieldy and actually hurts comprehension, we should shell out to something under `scripts/`. The current `screen` voodoo is probably a good example of this. I'd like to try to keep the starting of bitcoind and Bisq nodes script-free, as I think there's value in seeing them all together, explicitly parameterized in one place as we have it now. Let's just do whatever makes sense over time though.

Makefile
```diff
@@ -0,0 +1,251 @@
...
+
+# Deploy a complete localnet by running all required Bitcoin and Bisq
+# nodes, each in their own named screen window. If you are not a screen
+# user, you'll need to manually run each of the targets listed below
+# commands manually in a separate terminal or as background jobs.
+deploy: setup
+	# create a new screen session named 'localnet'
+	screen -dmS localnet
+	# deploy each node in its own named screen window
+	targets=('bitcoind' 'seednode' 'seednode2' 'alice' 'bob' 'mediator'); \
```
julianknutsen: The Makefile didn't specify the SHELL variable so mine defaulted to /bin/sh which doesn't support the array syntax here. Not sure if there is a shell agnostic way to do it, but adding SHELL=/bin/bash to the Makefile fixed it locally.

```
julian@dev:~/bisq$ make deploy
# create a new screen session named 'localnet'
screen -dmS localnet
# deploy each node in its own named screen window
targets=('bitcoind' 'seednode' 'seednode2' 'alice' 'bob' 'mediator'); \
for t in "${targets[@]}"; do \
	screen -S localnet -X screen -t $t; \
	screen -S localnet -p $t -X stuff "make $t\n"; \
done;
/bin/sh: 1: Syntax error: "(" unexpected
make: *** [Makefile:156: deploy] Error 2

```

The iterative single-node behavior also doesn't seem to work quite right. I checked out an older version of the code, killed bob, ran `make bob` and it just restarted without a build.

I also ran `make desktop/build` and no changes, but `./gradlew :desktop:build` had something to do.

dependency_issue

Even making alice & bob have a dependency on `desktop/build` instead of just `setup` didn't work.

cbeams: > The Makefile didn't specify the SHELL variable so mine defaulted to /bin/sh which doesn't support the array syntax here. Not sure if there is a shell agnostic way to do it, but adding SHELL=/bin/bash to the Makefile fixed it locally.

Good catch, thanks @julianknutsen. Commit 234c228df removes the offending bashism such that `/bin/sh` should run without error.

> The iterative single-node behavior also doesn't seem to work quite right. I checked out an older version of the code, killed bob, ran `make bob` and it just restarted without a build.

You're right, and I should have mentioned this in https://github.com/bisq-network/bisq/pull/3718#issuecomment-560573961. The rationale for why things need to be this way is laid out in 5fb4b21. Search for the first occurrence of the word 'contention' there. I'm updating the comment above now.

cbeams:
FYI, there is now a `make undeploy` target as well. From commit message ed40afb:

```
$ git log -1 ed40afb
commit ed40afb1516a08aa651665ed1d61d60555543c5d
Author: Chris Beams <chris@beams.io>
Date:   Tue Dec 3 11:56:04 2019 +0100

    Add 'make undeploy' target to kill all running nodes
    
    Problem: previously, in order to completely shut down a running
    localnet, users had to attach to their 'localnet' screen and kill (^C)
    each process, then quit and kill the entire screen session.
    
    Solution: this change introduces an 'undeploy' target that automates
    sending the ^C to each screen window followed by sending screen's 'kill'
    command to any remaining windows, thus killing the entire 'localnet'
    screen session.
    
    The result is that users may now run the following two commands in
    succession any number of times to bring their localnet up and down (to
    'deploy' and 'undeploy' their localnet).
    
        # bring up localnet
        $ make deploy
    
        # use localnet to test, develop, etc...
    
        # bring down localnet
        $ make undeploy
```

Makefile
julianknutsen: Thanks for the commit link and it makes sense w.r.t. contention. Verified the bashism issue is fixed as well and will do 1.2.4 testing today with your latest code and call out any other glaring issues. So expect an ACK by EOD from me if everything works out.

bodymindarts: @cbeams thanks for your instructive commit messages!
I understand your reasoning behind removing `build` from PHONY. I'm just wondering wether there is a better solution. Doing a clean-rebuild takes a very long time and is not practical for quick iteration cycles. Essentially the problem you have solved is a race condition. Couldn't we ensure somehow that the build runs just once before deploying the individual nodes. Perhaps running the individual node commands shouldn't depend on the setup/build commands. That way we could have both a 1 time setup + iteration.
Is there a significant advantage of having all the individual node commands depend on setup/build that I'm missing?

julianknutsen (dismissed):
ACK

This is plenty good enough for the original intention. Chris has spent a lot of time iterating and I think we may be at the point of diminishing returns. Plenty of other high prio work to do.

Do any onboarding docs need to be updated to point to the make commands?

I was able to test this with quite a few node restart and recompile tests today without any issues other than a totally reasonable ubutntu popup message:
undeploy_error_popup

Makefile
cbeams: > Doing a clean-rebuild takes a very long time and is not practical for quick iteration cycles.

Here's what I'm doing in practice as I develop stuff. The following assumes I have already done a `make deploy`, and that now I'm iterating on something, using my `alice` desktop node to see my changes live:
```
# quit my running `alice` node, e.g. with CMD-Q in the UI or ^C on the running process
./gradlew :desktop:build && make alice
```
This picks up and rebuilds just the changes I've made and then deploys those new bits as `alice`.

> Couldn't we ensure somehow that the build runs just once before deploying the individual nodes.

It does effectively run just once. The individual node deployment targets, e.g. `alice` and `bob` depend on `build`. Note that build is, in the latest commits, actually included once again in PHONY, so it does run every time, but that target just depends on the more specific `desktop/build` and `seednode/build` targets, which in turn run only if their respective directories do *not* already exist. The effect here is that when we deploy more than one node or all of them via `make deploy`, the `{desktop|seednode}/build` targets get run once and only once, avoiding the above mentioned race condition and inefficient contention for Gradle resources. If you want to cause a fast (incremental) rebuild, it's simply necessary to drop down to calling gradle directly like I've shown above. calling `make clean-build` should not be necessary in any case, unless you actually want to blow away all the build directories.

> Is there a significant advantage of having all the individual node commands depend on setup/build that I'm missing?

It just ensures that deploying any given node causes `build` and `localnet` to run if they have not already done so. So someone can come along in a clean checkout and run only `make bitcoind` and `make seednode` and everything will work as they expect, meaning that the .localnet dir will get created and the seednode build will run. If that's already happened, then those targets are no-ops.

cbeams:
I think I've addressed pretty much everything in the feedback thus far and there've been multiple reports of things working as expected, so I think this initial cut is ready for a merge. @ripcurlx, your ACK is the required one. I know you were having some kind of issue with the Makefile on your side, I'd be happy to help you with that if you want to get it sorted before the merge.

Thanks again to everyone. The feedback has really helped improve and harden this.

cbeams:
@julianknutsen wrote:

> Do any onboarding docs need to be updated to point to the make commands?

Yes, and I've flagged this work to be done elsewhere. But I should probably make at least some basic change now. I'll do that real quick.

cbeams:
> I should probably make at least some basic change now. I'll do that real quick.

Commit 7d16890 has addressed updating documentation in a minimal but hopefully effective fashion. Most resources that new developers would see first, be it CONTRIBUTING.md, the Contributor Checklist or the root-level README.md point to the "developer docs" at docs/README.md where these changes have been made.

@ripcurlx, I think this PR is ready to go now.

ripcurlx:
> @ripcurlx, your ACK is the required one. I know you were having some kind of issue with the Makefile on your side, I'd be happy to help you with that if you want to get it sorted before the merge.

@cbeams It would be great to understand why it is not working with the `screen` multiplexer in my case on macOS Catalina with Z shell(zsh).
```
make deploy
# create a new screen session named 'localnet'
screen -dmS localnet
# deploy each node in its own named screen window
for target in \
			bitcoind \
			seednode \
			seednode2 \
			alice \
			bob \
			mediator; do \
		screen -S localnet -X screen -t $target; \
		screen -S localnet -p $target -X stuff "make $target\n"; \
	done;
# give bitcoind rpc server time to start
sleep 5
# generate a block to ensure Bisq nodes get dao-synced
make block
bitcoin-cli \
		-regtest \
		-rpcuser=bisqdao \
		-rpcpassword=bsq \
		getnewaddress \
		| xargs bitcoin-cli \
				-regtest \
				-rpcuser=bisqdao \
				-rpcpassword=bsq \
				generatetoaddress 1
error: Could not connect to the server 127.0.0.1:18443
Make sure the bitcoind server is running and that you are connecting to the correct RPC port.
```
As using the targets in regular opened tabs works without any problems, I'll merge this PR  anyways.

ripcurlx (approved):
ACK - Tested it locally and except for https://github.com/bisq-network/bisq/pull/3718#issuecomment-562583086 everything worked as expected.

cbeams:
> It would be great to understand why it is not working with the `screen` multiplexer in my case

@ripcurlx I'll ping you 1:1 to debug this, thanks.

PR #3726, opened 2019-12-02
Add handling of mediationResultState to isFundsLockedIn

### LINKED ISSUE #3727: Trades with a mediated payout end up in failed trades
If a trades gets payed out via a mediated payout the trade will be moved to failed trades after a restart. This happens with payout to both traders as well as with payout to only 1 trader.

chimp1984:
Fixes https://github.com/bisq-network/bisq/issues/3727
(part of the problem was that the trade ended up in failed trade)

Refactor method and add comments.
We did not handle the case of a mediated payout. isPayoutPublished() is
only reflecting non-disputed trade payouts.

@ripcurlx Please cherry pick to release branch.


ripcurlx (approved):
utACK

PR #3745, opened 2019-12-04
Rename "Spread" tab to a more general "Details"

### LINKED ISSUE #3376: Rename 'Spread' submenu to 'Details', since it shows much more than the spread
The *Spread* submenu (under `MARKETS`) shows the details of a market, which include the spread, but also many other info. I suggest to change the name from `SPREAD` to `DETAILS` to better reflect its content.

dmos62:
Fixes #3376 

Because spread is one of many columns in the table in the "Spread" tab, calling the whole tab "Spread" is not general enough. Calling the table "Details" is more general and is meant to signify that it offers more information than the "offer book" view.

A counter-argument would be that 1) the table does not actually offer new information, just new aggregates on the same information displayed under "Offer book"; and 2), "Details" is somewhat vague.

Currently, it is unclear what a more accurate title could be, and in the meantime "Details" is an improvement on "Spread".

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -287,7 +287,7 @@ mainView.version.update=(Update available)
 ####################################################################
 
 market.tabs.offerBook=Offer book
-market.tabs.spread=Spread
+market.tabs.spread=Details
 market.tabs.trades=Trades
 
 # OfferBookChartView


julianknutsen (commented):
**Testing**

The english version looks good.
verify_details

The Spanish version still has the old text which I think is to be expected. I have no experience with how the translations are done, so someone with a better understanding of the workflow can weigh in here.

verify_details_spanish

The fact that this was marked as a "Good First Issue" makes me think it is OK for just the english version to change, but @ripcurlx or @freimair can clear that up.

As far as your analysis on whether or not this is the "best" text. I think that this issue has been up for review for almost two months now so any sort of bikeshedding on it should have happened. The new text is definitely moving it in the right direction imo.

dmos62:
I was wondering about translation as well. In the localisation docs it says that the localised files should not be modified directly, but through the third-party translation platform.

ripcurlx:
@dmos62 @ julianknutsen Only the English translation is to be changed on master. If the value for one key is changing it is marked untranslated on Transifex and translators can pick it up for translation. Before each release I'm pulling the translations from Transifex and update it in the repository.

ripcurlx:
So we do update translations on master, but it is only the release manager (me) atm who is doing this to prevent manual changes of the translation files by mistake.

ripcurlx:
Regarding the new Details label: @m52go Do you have an opinion on this?

JFYI: @m52go is our proof reader when an English translation is added or changed.

sqrrm (approved):
ACK

I think this wording is an improvement. Still will let @m52go have a say before merging.

m52go:
Looks good to me 👍

julianknutsen:
@dmos62 Congrats on your first merged PR!

PR #3747, opened 2019-12-04
(6/6) Clean up technical debt in P2PDataStorage and ProtectedStorageEntry objects

julianknutsen:
b6b0026

## Motivation
This PR cleans up technical debt in the P2PDataStorage and the ProtectedStorageEntry objects that should be done before new features are added.

Future patches will change these objects quite a bit and getting them into a state that is easy to reason about will make the reviews easier.

## Testing
All existing unit tests continue to pass


p2p/src/main/java/bisq/network/p2p/storage/payload/ProtectedMailboxStorageEntry.java
```diff
@@ -107,7 +107,7 @@ public boolean isValidForAddOperation() {
         if (!result) {
             String res1 = this.toString();
             String res2 = "null";
-            if (mailboxStoragePayload != null && mailboxStoragePayload.getOwnerPubKey() != null)
+            if (mailboxStoragePayload.getOwnerPubKey() != null)
```
freimair, line 109: (applies to all removed null-checks)
you say in the commit comment that `fromProto` will throw an exception. I cannot see that. However, it seems like the protobuf implementation does supply a default value which is not null. So either way this change should be fine. Or did I miss something here?

freimair (changes requested):

p2p/src/main/java/bisq/network/p2p/storage/payload/ProtectedMailboxStorageEntry.java
julianknutsen: I could have been more clear with the distinction. Here is the relevant code for the payload itself. Since it is a submessage in proto3 it can technically be null, but the resolver handles those cases and raises an exception:

https://github.com/bisq-network/bisq/blob/76e8c5736a6c5f56cca83323c373d2f3eb3f2542/core/src/main/java/bisq/core/proto/network/CoreNetworkProtoResolver.java#L249-L278

The other non-null issues were due to the initialized value for proto3 not being null, this isn't one of those cases.

freimair: ah damn. missed that.

freimair (dismissed):
Ack

ripcurlx:
> Ack

@freimair Could you please add how you tested this code changes, so it easier to compare with other ACKs. Thanks!

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -440,46 +440,51 @@ public void onConnection(Connection connection) {
...
-                                    log.debug("Remove data ignored as we don't have an entry for that data.");
-                                }
-                            }
-                        }
-                    });
-        }
+        if (closeConnectionReason.isIntended)
+            return;
+
+        if (!connection.getPeersNodeAddressOptional().isPresent())
```
ripcurlx, line 446: Is there a reason why you prefer using directly `getPeersNodeAddressOptional().isPresent()` in comparison using `hasPeerNodeAddress()` ? It is used mixed in the original code as well, so we might use one way for the future.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -440,46 +440,51 @@ public void onConnection(Connection connection) {
...
+                        .filter(entry -> entry.getValue().getProtectedStoragePayload() instanceof RequiresOwnerIsOnlinePayload)
+                        .filter(entry -> ((RequiresOwnerIsOnlinePayload) entry.getValue().getProtectedStoragePayload()).getOwnerNodeAddress().equals(peersNodeAddress))
+                        .collect(Collectors.toCollection(ArrayList::new));
+
+        // Backdate each payload
+        toBackDate.forEach(mapEntry -> {
+            // We only set the data back by half of the TTL and remove the data only if is has
+            // expired after that back dating.
+            // We might get connection drops which are not caused by the node going offline, so
+            // we give more tolerance with that approach, giving the node the change to
```
ripcurlx: NIT: Small typo in the original code: `change` should be `chance`

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -448,43 +448,24 @@ public void onDisconnect(CloseConnectionReason closeConnectionReason, Connection
...
+        // Backdate all the eligible payloads based on the node that disconnected
+        map.values().stream()
+                .filter(protectedStorageEntry -> protectedStorageEntry.getProtectedStoragePayload() instanceof ExpirablePayload)
+                .filter(protectedStorageEntry -> protectedStorageEntry.getProtectedStoragePayload() instanceof RequiresOwnerIsOnlinePayload)
+                .filter(protectedStorageEntry -> ((RequiresOwnerIsOnlinePayload) protectedStorageEntry.getProtectedStoragePayload()).getOwnerNodeAddress().equals(peersNodeAddress))
+                .forEach(protectedStorageEntry ->  {
+                    // We only set the data back by half of the TTL and remove the data only if is has
+                    // expired after that back dating.
+                    // We might get connection drops which are not caused by the node going offline, so
+                    // we give more tolerance with that approach, giving the node the change to
```
ripcurlx: NIT: The `change` != `chance` typo in the original code is copied over to this place as well.

ripcurlx: NIT: The `change` != `chance` typo in the original code is copied over to this place as well.

ripcurlx (dismissed):
ACK

- Tested on local Regtest trading with each peer going offline after each step.
- Tested on local Regtest trader chat with each peer going offline after each message.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
julianknutsen: I wish there was a better reason than this one, but I used this pattern because it reduces the number of lint errors from IDEA. If you don't have an `isPresent()` call prior to a `get()` it will highlight the word and raise an alert. You can suppress it, but that is just more code to add.

The other benefit is that mocking is bit less error-prone because you just need to mock out the Optional and not the wrapper class that has the `hasPeerNodeAddress()` function.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
julianknutsen: Addressed in e8c8225

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
julianknutsen: Addressed in e8c8225

julianknutsen:
@ripcurlx This PR is in a strange GitHub state now. It has the "waiting for author" tag, but since it was ACKd and not "Request Changes" I don't have a button to push to move it back to your queue and can't remove the label myself. Not sure the right way to handle this type of workflow, but I wanted to give you a heads up.

In any event, I've addressed your comments and it is ready for another look.

ripcurlx:
> @ripcurlx This PR is in a strange GitHub state now. It has the "waiting for author" tag, but since it was ACKd and not "Request Changes" I don't have a button to push to move it back to your queue and can't remove the label myself. Not sure the right way to handle this type of workflow, but I wanted to give you a heads up.
> 
> In any event, I've addressed your comments and it is ready for another look.

It was probably not perfectly handled by myself, as I probably should have asked for change requested as status. In that case I was ok about everything, just had optional remarks.

Labels can only be changed by maintainers or triage permission owners. I'll have a look how to do it with GitHub actions automatically based on the PR state.

ripcurlx (approved):
ACK

PR #3748, opened 2019-12-04
Update data stores, translations and bitcoinj checkpoints

ripcurlx:
Here are the hash screenshots for the new data stores (from full node in dark mode). I've added the state from my light node (in light mode) for comparison.
Bildschirmfoto 2019-12-04 um 22 11 08
Bildschirmfoto 2019-12-04 um 22 11 16
Bildschirmfoto 2019-12-04 um 22 11 39
Bildschirmfoto 2019-12-04 um 22 11 24Bildschirmfoto 2019-12-04 um 22 11 55
Bildschirmfoto 2019-12-04 um 22 12 02


ripcurlx:
I'll merge this PR without review as it is mainly for documentation reason (screenshots).
@julianknutsen Maybe you could compare it with your locale node state to get another confirmation for this data stores.

julianknutsen:
Verified that the screenshots match my local state for the DAO, Proposal, and Blind Vote states.

core/src/main/resources/i18n/displayStrings_de.properties
```diff
@@ -1961,7 +1961,7 @@ showWalletDataWindow.includePrivKeys=Private Schlüssel einbeziehen
 tacWindow.headline=Nutzervereinbarung
 tacWindow.agree=Ich stimme zu
 tacWindow.disagree=Ich stimme nicht zu und beende
-tacWindow.arbitrationSystem=Vermittlersystem
+tacWindow.arbitrationSystem=Dispute resolution
```
arunasurya, line 1964: Why is the German translation (and others, see below) reverted to English?

arunasurya (commented):
I was looking at another Transifex issue and noticed that some strings seem to have reverted to English.

core/src/main/resources/i18n/displayStrings_de.properties
arunasurya: I looked into this briefly in Transifex and the string did change from "Vermittlersystem" to "Streitbeilegung" about a month ago but it seems that something went wrong when importing the new strings into the git repo.

PR #3749, opened 2019-12-05
Validate wallet date when restoring from seed

devinbileck:
Prevent a date that is older than the oldest wallet date, or a date that
is newer than the current date.


sqrrm (approved):
utACK

chimp1984 (commented):
For the oldest date I think we should use the latest v0.5 release date as that is when we changed wallet format.

devinbileck:
Though it may not be desired, I believe it is technically possible to restore prior to that. I figured it would be best to give the user the possibility to do that.
Perhaps it makes sense to implement feedback within the UI to inform/ask the user when they attempt to restore without specifying a date. At a minimum, inform them what happens when they dont specify a date, and inform them that any current Bisq wallet does not date older than the v0.5 release date.

PR #3751, opened 2019-12-05
Improve link to developer docs

cbeams:
The previous link format works fine while in the GitHub web interface,
but is not useful when trying to navigate from the filesystem. This
format works well in both contexts.

This minor issue was discovered in the course of documentation updates
for PR #3718.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -87,5 +87,4 @@ See the issues in the [bisq-network/style](https://github.com/bisq-network/style
 ## See also
 
  - [contributor checklist](https://docs.bisq.network/contributor-checklist.html)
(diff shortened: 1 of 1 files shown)


cbeams:
Closing as dropped. I remembered just now why I originally used the docs#readme link format. When clicked (via the web interface), it takes you to https://github.com/bisq-network/bisq/tree/master/docs#readme instead of https://github.com/bisq-network/bisq/blob/master/docs/README.md, and there's an important difference between the two. The former scrolls the user to the README text, but the user can also easily scroll up and see the list of all files in the docs/ directory. This is good for discoverability and context. The latter link format takes the user directly to the README.md file, and they'd have to navigate up to the docs/ directory to see the other files.

PR #3754, opened 2019-12-06
Fix dark mode slider switch not disabling dark mode on first try

### LINKED ISSUE #3447: Dark mode not disabled on 1st attempt to turn it off after restart
[How to Reproduce]
1. Start arbitration desktop on Ubuntu 18
2. Enable dark mode
3. Shut down desktop
4. Restart arbitration desktop
5. Disable dark mode 

[Expected Results]
Dark mode is disabled on 1st attempt to turn it off.

[Actual Results]
Dark mode is not disabled on 1st attempt to turn it off (see bottom left of screen shot below).
It is disabled on 2nd attempt to turn it off.
DarkModeNotDisabled

This issue was discovered during v1.2.0 [testing](https://bisq.ontestpad.com/script/103#/1) .

wiz:
Apparently the preferences property listener I was using doesn't call
the event handler on the first change for some reason. This is a
workaround to change the CSS theme directly from the slider switch.

Needs review and testing! Fixes #3447

### DIFF
--- a/core/src/main/java/bisq/core/user/Preferences.java
+++ b/core/src/main/java/bisq/core/user/Preferences.java
@@ -343,8 +343,8 @@ public final class Preferences implements PersistedDataHost, BridgeAddressProvid
         this.useAnimationsProperty.set(useAnimations);
     }
 
-    public void setCssTheme(boolean useDarkMode) {
-        this.cssThemeProperty.set(useDarkMode == true ? 1 : 0);
+    public void setCssTheme(int cssTheme) {
+        this.cssThemeProperty.set(cssTheme);
     }
 
     public void addFiatCurrency(FiatCurrency tradeCurrency) {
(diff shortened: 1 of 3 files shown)


ripcurlx:
I'll have a look at this behavior today and if it caused issues in other parts of the code as well.

ripcurlx:
@wiz The problem is that ```
private final IntegerProperty cssThemeProperty = new SimpleIntegerProperty(prefPayload.getCssTheme());``` gets initialized before the stored payload is initialized which causes the value of the `cssThemeProperty` always to set to `0`. That is also the reason why the listener is not triggered the first time when changing back to light mode after a restart.

ripcurlx:
You missed to update the value in readPersisted in the Preferences class (L:271+). But I also saw during investigation that we are still having properties there we don't use anymore so I'll create another PR cleaning up the unused properties and fixing this issue as well.

ripcurlx:
Closing as superseded by #3771 .

wiz:
> The problem is that ` private final IntegerProperty cssThemeProperty = new SimpleIntegerProperty(prefPayload.getCssTheme());` gets initialized before the stored payload is initialized which causes the value of the `cssThemeProperty` always to set to `0`

Ah that's why - good thing it was my bug and not something weird elsewhere :)
Thanks for fixing

PR #3757, opened 2019-12-06
Add script to create seednode jar archive without jar deps

wiz:
I use this script to maintain a FreeBSD seednode, might be handy for others like @Emzy


ripcurlx (approved):
utACK

PR #3762, opened 2019-12-08
Update contribution guidelines

freimair:
This is another step towards a more streamlined reviewing process and making better use of the resources we have. These changes are largely based on a recent maintainers call.

Key points of the re-design:
- added a "Are you the guy?" section. Let people know up front what they can expect.
- added a "getting started" section. Link (updated) communication channels, describe how to get going.
- refactored the rest:
  - refined some parts in order to make them more clear
  - added bits that might streamline the maintenance process of Bisq (testing, PR ttl, further reading)

Once this gets approved, I will schedule a dev-call on this very topic to get into the points a little more.


ripcurlx:
@cbeams Would be great to get your input on this, especially as a native speaker 😄

CONTRIBUTING.md
```diff
@@ -1,91 +1,101 @@
...
 
-Most communication about Bisq happens in the [Bisq Slack workspace](https://bisq.network/slack-invite).
+**What is there to work on?** You have to be able to find and work a problem on your own. In general, we do not assign stuff to people, there is no specification to be implemented, there is no given way forward. You yourself have to find, choose and pick something to work on. If you get stuck, we are keen to help, but please do not wait for us to take your hand. Also, you can spend as much (or little) time as you want and when you want to - Bisq still is an open source project.
 
-Discussion about code changes happens in GitHub issues and pull requests.
+**Will you get paid?** Well, yes and no. With Bisq, there is no classic funding or investors, so there is no payout of fiat money.
+- Yes, because there is the DAO. The DAO allows for compensation requests to be filed and voted on. If such a voting goes in your favor, you receive BSQ, a colored bitcoin token. Sell your BSQ and you get your fiat eventually.
```
ripcurlx: Maybe add a link to the DAO documentation for further reading.

CONTRIBUTING.md
```diff
@@ -1,91 +1,101 @@
...
+- No, because the market is still very small. We can sustain a few full-timers (paying rent, tax, small stuff) but do not expect to get rich fast and easy just now.
```
ripcurlx: IMO this sounds a little bit as if full timers are preferred in the market, which is not correct. Also it makes assumption about what we think is sustainable for contributors.
Maybe we could just point out the total numbers:
`No, because the market is still bootstrapping. Since the launch of the DAO in April 15th 2019 there were 623k BSQ sold for 68 BTC (average of 0.064 BTC per day) as of December 10th.`

CONTRIBUTING.md
```diff
@@ -1,91 +1,101 @@
...
 
-Discussion about larger changes to the way Bisq works happens in issues the [bisq-network/proposals](https://github.com/bisq-network/proposals/issues) repository. See https://docs.bisq.network/proposals.html for details.
+The more correct way to see payment is "compensation". By earning BSQ, you invest in the Bisq project (just like other startups do it). But make no mistake, the DAO has only been operating for a few months now. Eventually, Bisq will create serious revenue.
```
ripcurlx: I'm not so sure about the `revenue` term. Bisq had already serious volume (revenue) it is just more about the relationship of BSQ burnt and BSQ issued. Which would mean `profit` in regular terms. But again I'm not sure if it is wise to connect the dots that blunt using these terms. Maybe write something like:
`Although the platform was launched in April 2016, the DAO has only been operating for a few months now. Eventually, the amount of BSQ burnt (through trading fees or other DAO related utilities) in relation to BSQ issued will increase significantly over time.`

CONTRIBUTING.md
```diff
@@ -1,91 +1,101 @@
...
 
-Pull requests should be focused on a single change. Do not mix, for example, refactorings with a bug fix or implementation of a new feature. This practice makes it easier for fellow contributors to review each pull request on its merits and to give a clear ACK/NACK (see below).
 
+Still here? Nice. You are hired. On to the grind.
 
-## Reviewing Pull Requests
+# Getting started
 
-Bisq follows the review workflow established by the Bitcoin Core project. The following is adapted from the [Bitcoin Core contributor documentation](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#peer-review):
+Join Gibhub if you haven't already, fork the Bisq repository, configure your local git ([username](https://help.github.com/articles/setting-your-username-in-git/), [commit signing](https://help.github.com/articles/signing-commits-with-gpg/)), clone your Bisq-fork to disk, do a compile run `./gradlew build`.
```
ripcurlx:
```suggestion
Join Gibhub if you haven't already, fork the Bisq repository, configure your local git ([username](https://help.github.com/articles/setting-your-username-in-git/), [commit signing](https://help.github.com/articles/signing-commits-with-gpg/)), clone your Bisq-fork to disk and do a compile run `./gradlew build`.
```

CONTRIBUTING.md
```diff
@@ -1,91 +1,101 @@
...
 
-Anyone may participate in peer review which is expressed by comments in the pull request. Typically reviewers will review the code for obvious errors, as well as test out the patch set and opine on the technical merits of the patch. Project maintainers take into account the peer review when determining if there is consensus to merge a pull request (remember that discussions may have been spread out over GitHub, mailing list and IRC discussions). The following language is used within pull-request comments:
+Now you are ready to roll. Pick something off the [Good First Issue List](https://github.com/bisq-network/bisq/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22), pick any other issue, do some testing, spot a bug, report it and fix it, maybe there is something in the UI that needs cosmetic work.
```
ripcurlx: Although everyone can do whatever they want, maybe we should point the user more to the tasks we'd like to have them fulfilled. E.g. Take a good first issue or fix a bug.

CONTRIBUTING.md
```diff
@@ -1,91 +1,101 @@
...
 
-### Configure Git user name and email metadata
+Please honor the list of rules below for your pull request to get merged. Due to very limited resources, we are forced to enforce these quite rigorously.
 
-See https://help.github.com/articles/setting-your-username-in-git/ for instructions.
+## Honor Git best practice
+ - Create a topic branch off master
+ - Commit little, commit often
+ - Maintain a clean commit history (no merges commits!, use rebase and force-push if necessary)
+ - Use meaningful commit subjects AND description (then, now, good, bad, risks)
```
ripcurlx: > (then, now, good, bad, risks)
 
What do you mean with that?

ripcurlx (commented):
Concept ACK

I think the content we want to have in is there, but I'd like for more regular contributors to respond as this is a bigger change in our PR process as well.

@sqrrm @cbeams @julianknutsen @wiz @devinbileck @chimp1984

CONTRIBUTING.md
freimair: agreed, we should not make such assumptions. However, 0,xx BTC per day is hard to grasp. How about:

`- No, because the BSQ market is still bootstrapping. For the first months, we saw BSQ sold for a total of 8 BTC per month vs. about 20 compensation requests. So there is some payment happening but do not expect to get rich fast and easy just now.`

The numbers of course have to be properly fetched, I just eyeballed them. And, downside is that we might have to correct the numbers once in a while (I fear that this is not going to happen).

How about something like
`- No, because the BSQ market is still bootstrapping. Although BSQ already has a healthy sell rate, it is not yet enough to sustain all compensation requests. (So do not expect to get rich fast and easy just now.)`

CONTRIBUTING.md
freimair: You are correct of course. Revenue, profit, tricky stuff.

`Eventually, the BSQ market will be able to stay ahead of compensation requests.`

(native speaker needed...)

CONTRIBUTING.md
freimair: The first item is the good first issues, the second one is fix a bug. Please elaborate.

CONTRIBUTING.md
freimair: The link is provided in the compensation section down below.

Besides, this basically says that there is the "thing" and it can pay you. Is there need for more? If you choose to read on, you will hit the details eventually.

CONTRIBUTING.md
freimair: ah yes. more specific:

```suggestion
 - Use meaningful commit subjects AND description
   - how did stuff work before you touched it, what was bad back then
   - how does the stuff work now and why is it better now
```

CONTRIBUTING.md
julianknutsen: I would suggest an in-page link for the text "compensation requests" that go to the relevant section. Getting up to speed on all the logistics of a brand-new project can be overwhelming and it is likely that this document will be skimmed a few dozen times the first week someone is on the project. Quick navigation to the most important points that will help devs decide if this is a good project should be a priority.

CONTRIBUTING.md
julianknutsen: Full/part-time varies drastically across the world and may give the wrong impression. 

@ripcurlx  It also seems strange that full-timers wouldn't be preferred. Full-time developers are certainly more efficient, can work on more impactful projects, and can get up to speed much faster with fewer resources. If that isn't the goal, it may make more sense for this doc to call out small & part-time as the target contributor.

I think the best bet is to provide easy access to the metrics so devs can decide for themselves if it makes sense for their situation. I don't even think this exists in one-place right now, but the most useful item for me when analyzing this was to look at the total amount of BSQ purchased over the last 30 days and the average price. This gave me some understanding of how much demand there was for my future compensation request BSQ and how much would need to stay invested.

That, along with an easy way for people to view the end-of-cycle reports with the major inflation/deflation stats give a good glimpse of the state of BSQ.

Finally, this may exist in some other docs, but explaining that not only is there a free market for BSQ that traders use for reduced fees, but BSQ is also actively purchased from BTC fees and Arbitration every month. As the project grows, the liquidity will grow with it.

Not every dev is going to have a solid grasp on the economics so helping them get the high-level information is important so they don't get too invested with something that doesn't work for them.

I'm not a huge fan of the wording "So do not expect to get rich fast and easy just now.". Is the goal to discourage speculative investors? Devs asking for more than their worth?

CONTRIBUTING.md
julianknutsen: `But make no mistake, the DAO has only been operating for a few months now. Eventually, Bisq will create serious revenue.`

It isn't clear what `revenue` is referring to in the line. Does that translate to higher BSQ value or the capacity to "employ" more people at once?

I think @ripcurlx has a better phrasing here by suggesting that the gap between burnt and issued __could__ decrease as the project is more successful. But, the DAO may decide to use increased BSQ burn rate to employ more people instead of deflating supply so the future is really unknown.

If I were looking at this with fresh eyes again, I would want a pointer to the actual data that I could analyze and extrapolate. Is there a chart of earned vs. burned that is accessible since the lifetime of BSQ? What about the growth rate of total transaction volume or trading fees? Anything that helps forecast the future of Bisq's revenue vs expenses.

CONTRIBUTING.md
```diff
@@ -1,91 +1,101 @@
...
+- Yes, because there is the DAO. The DAO allows for compensation requests to be filed and voted on. If such a voting goes in your favor, you receive BSQ, a colored bitcoin token. Sell your BSQ and you get your fiat eventually.
+- No, because the market is still very small. We can sustain a few full-timers (paying rent, tax, small stuff) but do not expect to get rich fast and easy just now.
 
-Discussion about larger changes to the way Bisq works happens in issues the [bisq-network/proposals](https://github.com/bisq-network/proposals/issues) repository. See https://docs.bisq.network/proposals.html for details.
+The more correct way to see payment is "compensation". By earning BSQ, you invest in the Bisq project (just like other startups do it). But make no mistake, the DAO has only been operating for a few months now. Eventually, Bisq will create serious revenue.
 
+**Tools?** If you want to develop code, we expect you to know your ways around Git and Java. Then comes testing, JavaFX, Gradle, Github, ... As for non-functional skills, we love to see people knowing their ways around P2P networks, cryptography, security, privacy, ...
```
julianknutsen: `Then comes testing, JavaFX, Gradle, Github, ...`

I expected a list of testing frameworks or patterns after the first three words. I would suggest including either the specific testing framework that Bisq uses or split out the fact that strong candidates for Bisq know how to test code.

The non-functional skills part doesn't seem to fit with Tools. It might make sense to fold that into another section.

CONTRIBUTING.md
```diff
@@ -1,91 +1,101 @@
...
 
-## Contributor Workflow
+**Carreer?** Start small, go bigger. If we like your work, we might just ask you to take on a more official role. Maybe someday you are the guy improving this document (again).
```
julianknutsen: Who is "we"? I thought decisions were made from the DAO? What are "official roles"?

These are the first questions that came into my head when reading this. What is the intention of this section?

If it is to explain that Bisq is looking for developers to start small and part-time that might be good to add explicitly.

If it is to explain that Bisq only wants people familiar with open-source development structure it might be worth including explicitly.

If it is to explain that Bisq isn't a typical hierarchical organization with career paths it may make sense to include a section on meritocracy or the concept of distributed asynchronous dev orgs.

CONTRIBUTING.md
julianknutsen: When you think of the first commit by the perfect Bisq contributor candidate what is it and where is it?

This lists 5 different categories without any sense of preference. Each one is going to have a very different set of resources required by PRs, testing, etc.

It might make sense to shorten it up so that onboarding is easier for reviewers and contributors.

CONTRIBUTING.md
```diff
@@ -1,91 +1,101 @@
...
- - `NACK` means "I disagree this should be merged", and must be accompanied by sound technical justification. NACKs without accompanying reasoning may be disregarded;
- - `utACK` means "I have not tested the code, but I have reviewed it and it looks OK, I agree it can be merged";
- - `Concept ACK` means "I agree in the general principle of this pull request";
- - `Nit` refers to trivial, often non-blocking issues.
-
-Please note that Pull Requests marked `NACK` and/or GitHub's `Change requested` are closed after 30 days if not addressed.
-
-
-## Compensation
+Now that you have picked yourself something to work on, be vocal about it! Claim good first issues, or any issues, by simply leaving a comment saying that this is being worked on, join [Keybase](https://keybase.io/team/bisq). Once there, introduce yourself in *#introductions*, get help in *#dev-onboarding* if your setup fights you, use *#dev* to see if the task you picked is a good way forward (you do not have to do that for good first issues) and join other channels if you want to.
```
julianknutsen: and register your github with keybase

CONTRIBUTING.md
```diff
@@ -1,91 +1,101 @@
...
+# Pull request rules
 
-### Configure Git user name and email metadata
+Please honor the list of rules below for your pull request to get merged. Due to very limited resources, we are forced to enforce these quite rigorously.
 
-See https://help.github.com/articles/setting-your-username-in-git/ for instructions.
+## Honor Git best practice
+ - Create a topic branch off master
+ - Commit little, commit often
+ - Maintain a clean commit history (no merges commits!, use rebase and force-push if necessary)
```
julianknutsen: `no merge commits`

CONTRIBUTING.md
```diff
@@ -1,91 +1,101 @@
...
- 3. Capitalize the subject line
- 4. Do not end the subject line with a period
- 5. Use the imperative mood in the subject line
- 6. Wrap the body at 72 characters
- 7. Use the body to explain what and why vs. how
+## Test your code!
+ - Create automated tests for your work. JUnit is available. If you are fixing a bug, create a test that demonstrates the bug by failing. Then fix the bug.
+ - If automated testing is not feasible (we know it can be hard - working on it ...), provide a textual testing procedure (which we can include in our manual testing checklist): setup, prerequisits, steps, expected results
+ - If that is not possible either, provide some info (numbers, screenshots, testing procedure) on how you tested your stuff
+ - If all is lost, state why you have not been able to do the above.
```
julianknutsen, line 71: I like this section. The burden is on the contributor to convince the reviewers that the code is solid. Not the other way around.

CONTRIBUTING.md
```diff
@@ -1,91 +1,101 @@
...
-### Keep the git history clean
+In case you receive a NACK and/or `change requested`
+ - react within 30 days, or else, your pull request will be rejected automatically
+ - **re-request a review** after adressing a change request (because we almost always only look at "review required" pull request because of resource constraints)
 
-It's very important to keep the git history clear, light and easily browsable. This means contributors must make sure their pull requests include only meaningful commits (if they are redundant or were added after a review, they should be removed) and _no merge commits_.
+## Compensation
 
-### Additional style guidelines
+Your pull request got merged? great, congrats and thanks for making Bisq more better.
```
julianknutsen, line 96:
```suggestion
Your pull request got merged? Great! Congrats on your first contribution and thanks for making Bisq better.
```

CONTRIBUTING.md
julianknutsen: Link to the calendar of events for design, growth, dev-calls, and even daily standup?

julianknutsen (commented):
Small comments inline. I tried not to go too far in the weeds with the feedback I've already shared with my onboarding experience. Happy to take another look for more structure/grammar issues once a few more people weigh in.

devinbileck:
Overall this seems like good content to include and I agree with the feedback provided thus far. Some of the wording and terminology could be improved. @m52go would you be able to have a look at this?

CONTRIBUTING.md
freimair: > I expected a list of testing frameworks or patterns after the first three words.

I intentionally left that open.
- keep the text easy to read
- if a Java dev knows testing, we have already won. Testing (imo?) is not about a certain framework, testing is a philosophy. If there is a framework unknown to you, you just learn it (although JUnit can hardly be called unknown).
- maybe some dev comes along with a special set of skills we did not know we want and need. Not being too specific leaves some room for such a dev to fit in

Thoughts?

> The non-functional skills part doesn't seem to fit with Tools.

Agreed. Didn't like that title either. "Skills"?

CONTRIBUTING.md
freimair: > I thought decisions were made from the DAO?

Well that is a difficult topic. Imagine we want some dev to do something and we need the DAO to vote on it. The work has to be put on ice until the next voting happens (which takes about two weeks until a decision is visible).
In general, waiting for a DAO voting cycle kills the momentum of the project, progress, reactions to people, attacks, needs, changed requirements.

By "we" I am referring to the dev community behind Bisq. Be it long-time contributors, substantial stakeholders, people who have proven their worth and trustworthiness, "core-devs", stakeholders

> What are "official roles"?

I thought about https://github.com/bisq-network/roles/issues. But you are of course correct, this is not obvious as is.

> If it is to explain that Bisq isn't a typical hierarchical organization ...

How about something like 

`As said earlier, there is no boss in Bisq - there is only stakeholders. Stakeholders have invested their time, gained some experience with the project and thus, are granted influence on the project accordingly. As soon as you receive your first couple of BSQ, you are a stakeholder too.`

`Start with something small, get yourself familiar with how things are done. Like it? Go bigger. Take on bigger tasks, be it development work, [roles](https://github.com/bisq-network/roles/issues) or other things you believe Bisq is in need for.`

`All in all, there is no boss job, there is no annual raise of salary, there is just stakeholders and their passion for the Bisq project.`

CONTRIBUTING.md
freimair: "no merge commits" is right there. please elaborate.

CONTRIBUTING.md
freimair: I rather not have the "on your first contribution" phrase in there because this section is applicable for each and every pull request.

Thoughts?

freimair:
that is a first. Changing the CONTIBUTING.md broke the build with jdk12. Does master still build?

CONTRIBUTING.md
chimp1984: I think we should not "freeze" the currently bad market (caused by bad trade volume) in a doc like that. The first 4 months have been liquid enough to sell BSQ, just over the past 3-4 months trade volume and with that burned BSQ broke down for various reasons (Bitcoin bear market, our Monereo whales seems to have left, 1.2 update and the related problems and bugs,...). A neutral market overview would be better IMO or simply point to the markets page so people can draw their own picture.

CONTRIBUTING.md
cbeams: > just like other startups do it

replace this with:

> as you might in a more traditional startup

---

> Eventually, the BSQ market will be able to stay ahead of compensation requests.

I wouldn't make a definite statement like this. Qualify it with something like:

> Over time, the BSQ market should be able to stay ahead of compensation requests.

CONTRIBUTING.md
cbeams: It's a typo fix. `s/merges/merge/`. I'd like to discuss this rule anyway. @ripcurlx mentioned he'd put it on the agenda whenever we next do a call about these kinds of things.

cbeams (commented):
@freimair, I just had a moment to review this today, so haven't done a comprehensive job. If you can fold in the current batch of feedback, I'll commit to giving the doc a proper editing pass when you're done.

CONTRIBUTING.md
ripcurlx: > . I don't even think this exists in one-place right now, but the most useful item for me when analyzing this was to look at the total amount of BSQ purchased over the last 30 days and the average price.

I made a mistake on the average amount per day as I calculated the average trade amount per trade not the amount per day 😬 (0.285 BTC/day is the correct average)

sqrrm: I agree with @julianknutsen here, better to have less value judgement and more description of what is going on and how to interpret the BSQ value. We need new devs to understand what they're getting into for this to be sustainable. I think the 'no' section in particular would be better described by just saying the buy demand isn't big enough and a link to the BSQ stats.

CONTRIBUTING.md
sqrrm: Indeed, data is best. I would also be cautious to use firm statements about the future such as 'Eventually, Bisq will create serious revenue'. Better with something less firm but still positive, like 'Eventually, we believe that Bisq can be self sustained as volume increase'.

freimair:
will do

PR #3772, opened 2019-12-10
Add command for signing windows executables with the default code signing certificate

### LINKED ISSUE #1952: Investigate signing Windows and macOS binary executables

ripcurlx:
Fixes #1952.

In my role as Bisq desktop maintainer and release manager I applied for a personal code signing certificate for `DI(FH) Christoph Johann Atteneder` (yes they wanted all academic titles and names included) at Sectigo (former Comodo). I applied for a code signing certificate which will expire on December 10th 2020.

### Verification steps
To test the behavior of a signed Windows executable you'll actually need a Windows machine to do so.
For testing purpose I've created a signed version of our Windows installer for v1.2.4.
Executable: https://www.dropbox.com/s/ukrotb5neajgwwu/Bisq-1.2.4.exe?dl=0
Signature: https://www.dropbox.com/s/hasuwvch0lag7wx/Bisq-1.2.4.exe.asc?dl=0

### DIFF
--- a/desktop/package/windows/package.bat
+++ b/desktop/package/windows/package.bat
@@ -2,6 +2,9 @@
 ::   - Inno Setup unicode installed (http://www.jrsoftware.org/isdl.php)
 ::   - OracleJDK 10 installed
 ::     Note: OpenJDK 10 does not have the javapackager util, so must use OracleJDK
+::   - Sign Tool installed (https://docs.microsoft.com/en-us/windows/win32/seccrypto/signtool)
+::     Note: Sign Tool is part of Windows 10 SDK (https://go.microsoft.com/fwlink/?LinkID=698771)
+::   - Code signing certificate installed
 :: Prior to running this script:
 ::   - Update version below
 ::   - Ensure JAVA_HOME below is pointing to OracleJDK 10 directory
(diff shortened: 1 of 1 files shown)


ripcurlx:
@devinbileck Could you please verify if the behavior is now as expected for Windows OS?

devinbileck:
Posting my results here.

I still get the smartscreen dialog, although it does show you as the publisher:
image

As mentioned [here](https://blogs.msdn.microsoft.com/vsnetsetup/2013/01/09/windows-smartscreen-prevented-an-unrecognized-app-from-running-running-this-app-might-put-your-pc-at-risk):

> the SmartScreen will continue to warn about the application until the certificate develops a reputation. But it would display a valid publisher name instead of unknown publisher.

> Although not required, programs signed by an EV code signing certificate* can immediately establish reputation with SmartScreen reputation services even if no prior reputation exists for that file or publisher. EV code signing certificates also have a unique identifier which makes it easier to maintain reputation across certificate renewals.

I am not sure if the EV certificate is worth the hassle, nor am I sure how long it would take to develop reputation with the standard certificate.

sqrrm:
I don't feel well equipped to review this PR. @devinbileck could you do a proper review? Anyone else I can ask?

ripcurlx:
> I am not sure if the EV certificate is worth the hassle, nor am I sure how long it would take to develop reputation with the standard certificate.


Besides being more expensive an EV certificate has stronger KYC requirements. I have to have a look how difficult it would be to get one based on the existing code certificate I already have. 

Regarding reputation it seems you get it e.g. by getting signed applications installed. So I think maybe we could start with the current code signing certificate to see if it would gain enough reputation already after the next round of updates. @devinbileck What do you think?

devinbileck:
> So I think maybe we could start with the current code signing certificate to see if it would gain enough reputation already after the next round of updates.

Yes, that sounds good to me.

devinbileck (approved):
ACK

I am not able to test with a signing certificate, but when attempting to run the script without it, I encounter the following error and the script still completes.

`SignTool Error: No certificates were found that met all the given criteria.`

That seems reasonable to me. That way anyone can still run the script to generate an installer without requiring a signing certificate.

sqrrm (approved):
utACK

PR #3777, opened 2019-12-12
Revolut account should accept only phone number

### LINKED ISSUE #3728: Revolut account should be a phone number
### Description

I traded with people who had an email instead of a phone number in their revolut account.

I tested to create a new Revolut account in bisq. In the field "Phone number" I was able to put an email.

#### Version

1.2.3

### Steps to reproduce

Account => Add a new account => Revolut => Put an email

### Expected behaviour

When entering an email instead of a phone number, I should see an error message when trying to save.
Maybe limit the field to numbers, parenthesis and the + sign

### Actual behaviour

Email is accepted

#### Device or machine

Windows 10

beingindot:
Fixes #3728
Till now we used to accept email or phone number
Now we are accepting only phone number in input though
for display we will provide backward compatibility.

Based on the country selected, we will validate the given phone number input.

Added list of countries accepted from official revolut website
https://www.revolut.com/en-US/help/getting-started/verifying-identity/what-countries-are-supported

### DIFF
--- a/desktop/src/main/java/bisq/desktop/components/paymentmethods/RevolutForm.java
+++ b/desktop/src/main/java/bisq/desktop/components/paymentmethods/RevolutForm.java
@@ -23,6 +23,8 @@ import bisq.desktop.util.Layout;
 import bisq.desktop.util.validation.RevolutValidator;
 
 import bisq.core.account.witness.AccountAgeWitnessService;
+import bisq.core.locale.Country;
+import bisq.core.locale.CountryUtil;
 import bisq.core.locale.CurrencyUtil;
 import bisq.core.locale.Res;
 import bisq.core.payment.PaymentAccount;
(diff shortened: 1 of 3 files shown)


ripcurlx (changes requested):
NACK

I tested it on Regtest and when switching the country in the combobox the validation is not updated. See screenshots below.

Bildschirmfoto 2019-12-12 um 15 42 54
When I switch the country to German the validation condition should be fulfilled, but is still showing the Austrian country code error message
Bildschirmfoto 2019-12-12 um 15 43 05

beingindot:
Now I have added `updateAllInputsValid()` from combobox action event. that triggers validation and it enables save new account. 

but the error message around text field is not clearing on its own. only after we press tab it clears. I tried to use `accountIdInputTextField.resetValidation()` but not working so far. may be that's an existing bug? pls provide suggestions.

ripcurlx:
> resetValidation

It has to be used similar as in the Bankform

```
holderNameInputTextField.resetValidation();
holderNameInputTextField.validate();
```

desktop/src/main/java/bisq/desktop/components/paymentmethods/RevolutForm.java
```diff
@@ -84,6 +102,43 @@ public void addFormForAddAccount() {
...
+                return null;
+            }
+        });
+        return countryComboBox;
+    }
+
+    void setCountryComboBoxAction(ComboBox<Country> countryComboBox) {
+        countryComboBox.setOnAction(e -> {
+            selectedCountry = countryComboBox.getSelectionModel().getSelectedItem();
+            updateFromInputs();
```
ripcurlx, line 140: You need to reset the input field validation to make it work as expected. 
```
selectedCountry = countryComboBox.getSelectionModel().getSelectedItem();
accountIdInputTextField.resetValidation();
accountIdInputTextField.validate();
updateFromInputs();
```

beingindot: per your suggestion, added unfocus/refocus and this solved the problem. Pls check.

ripcurlx (approved):
ACK

Tested it on Regtest and it is working now as expected.

PR #3783, opened 2019-12-12
Not allow bitcoinj to auto connect to localhost when localhost was not detected by client

ripcurlx:
This fixes the problem if the local bitcoin core node is not detected by our client,
but bitcoinj is able to connect to it because of the auto connect to localhost behavior.
In that case the minimum required nodes to broadcast a transaction will be 4 (provided nodes settings),
but bitcoinj will only connect to one node. The requirement of 4 nodes will be never fulfilled and
the transaction never broadcasted.

### Testing
As this is a bug based on the default localhost connect behavior of bitcoinj it is difficult to simulate on regtest.
To simulate this behavior on mainnet you could try following steps:
- Use current master
- Skip the initial check for a localhost node in BisqSetup `step 3`
- Go to network settings and you'll see only one node is connected but the minimum peers for broadcast is still set to `4`
- Checkout this PR and skip the initial check for localhost again
- Go to network settings and you'll see that the client connects to the provided nodes and will be able to broadcast transactions

### DIFF
--- a/core/src/main/java/bisq/core/btc/setup/WalletConfig.java
+++ b/core/src/main/java/bisq/core/btc/setup/WalletConfig.java
@@ -216,8 +216,6 @@ public class WalletConfig extends AbstractIdleService {
         // no proxy case.
         if (socks5Proxy == null) {
             peerGroup = new PeerGroup(params, vChain);
-            // For dao testnet (server side regtest) we prevent to connect to a localhost node to avoid confusion
-            // if local btc node is not synced with our dao testnet master node.
         } else {
             // proxy case (tor).
             Proxy proxy = new Proxy(Proxy.Type.SOCKS,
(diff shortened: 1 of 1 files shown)


ripcurlx:
I found this issue when I experienced a trade failure myself where the logs showed ```Waiting for 4 peers required for broadcast, we have 1 ...``` which was never fulfilled as bitcoinj only connected to my local core node.

chimp1984 (commented):
utACK

chimp1984 (approved):

julianknutsen:
This would only fail for users that run full bitcoind nodes locally, right? 

What is the workaround if a user encounters this? restart?

Is waiting for the Jan release sufficient or is this a hotfix candidate?

ripcurlx:
> This would only fail for users that run full bitcoind nodes locally, right?

Yes

> What is the workaround if a user encounters this? restart?

Should solve the problem also if you have a transaction already in limbo. Unfortunately for me it wasn't solved with a restart and after a SPV resync the transaction was removed by bitcoinj.

> Is waiting for the Jan release sufficient or is this a hotfix candidate?

I don't think this is something new, so I think it should be fine to ship it in the next release.

chimp1984:
> Is waiting for the Jan release sufficient or is this a hotfix candidate?

I think that happens only in edge cases. At startup we connect to localhost (`BisqSetup.checkIfLocalHostNodeIsRunning`) on the default port and if that succeeds we set a flag to use local Bitcoin Core. There seems to be cases where that check fails even a Bitcoin Core is funning at localhost and we should find out what causes that. I could imagine some race conditions when both Bisq and Bitcoin Core are start up at same time could cause that as well if Bitcoin Core is syncing (this should be detected as well as BitcoinJ cannot sync reliably then).

sqrrm (approved):
utACK

PR #3790, opened 2019-12-14
Add report feature for disputes

### LINKED ISSUE #3788: Provide reporting feature to mediators
Mediators need to make a report of their cases to provide evidence for their work done and to give feedback to developers and UX experts to improve Bisq to lower the cases.
As @bisquser stated in https://github.com/bisq-network/roles/issues/83#issuecomment-565570073 it is not clear to them how to provide a report. We have a shortcut built in for copying all dispute cases to the clipboard but that contains all the communications and way too much private data to be suited to get added to the report. Creating manually such a report is quite a bit of effort.

We should add another shortcut (or button) for exporting the case in a format we want to see in the reports. Here is a rough suggestion.

Dispute nr: [using enumeration instead of trade ID should be better for privacy and enough to keep cases identifiable] 
Reason for dispute: [Bug, "future trade", bank problems, Usability, Other,...]
Summary: [Short summary notes as alreayd used. Mediators need to avoid to add private information in there and to make it informative enough to get an idea what the dispute was about]
Opening date:
Closinging date:

I think all that data is easy accessible inside the dispute domain and it should be a easy task to implement that and will help to improve and standardize reporting. i think we do not need a data filter as the mediator can easily delete the already reported past cases from the text.

A proof system for disputes might be tricky and would require probably some mechanism that traders are publishing information and that needs to match the information provided by mediators. I don't see this a high priority in the current state. Much more important is that feedback from mediators goes into improvements so that total dispute cases get reduced. 

Btw. All the above would be good for the refund agent as well. As it shares the same code structures it should be trivial to extend that.

chimp1984:
At the dispute views (mediator, refund agent both for trader and
dispute agents) the shortcut cmd+k (or crtl+k or alt+k) will open a
popup displaying all disputes as compact summary. A copy to clipboard
button make it easy to copy the text to a text editor and post the
relevant disputes for the mediators/refund agents report.

Fixes https://github.com/bisq-network/bisq/issues/3788

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeView.java
+++ b/desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeView.java
@@ -18,7 +18,6 @@
 package bisq.desktop.main.support.dispute;
 
 import bisq.desktop.common.view.ActivatableView;
-import bisq.desktop.common.view.FxmlView;
 import bisq.desktop.components.AutoTooltipButton;
 import bisq.desktop.components.AutoTooltipLabel;
 import bisq.desktop.components.AutoTooltipTableColumn;
@@ -41,6 +40,7 @@ import bisq.core.support.SupportType;
 import bisq.core.support.dispute.Dispute;
 import bisq.core.support.dispute.DisputeList;
 import bisq.core.support.dispute.DisputeManager;
+import bisq.core.support.dispute.DisputeResult;
 import bisq.core.support.dispute.DisputeSession;
 import bisq.core.trade.Contract;
 import bisq.core.trade.Trade;
@@ -96,6 +96,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import lombok.Getter;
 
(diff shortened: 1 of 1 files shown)


chimp1984:
Here is a example of the report format:

Dispute nr. 1
Opening date: 25 Nov 2019 11:52:08
Close date: 26 Nov 2019 12:04:54
Duration: 23 hours, 58 minutes
Summary notes: Buyer did not start payment.
Summary notes (trader 2): Buyer did not pay. Seller will get buyer's security deposit.
Reason: PROTOCOL_VIOLATION

Dispute nr. 2
Opening date: 12 Dec 2019 23:07:21
Close date: 13 Dec 2019 21:08:03
Duration: 22 hours, 58 minutes
Summary notes: Deposit tx was not published
Reason: BUG

Summary of reasons for disputes: 
PROTOCOL_VIOLATION: 1
BUG: 1


I did not include trade ID for privacy reasons, but that's up for discussion if that should be added.

The summary notes are usually the same but the mediator/refund agent can send different summary notes as well. Only in that case we show both ("Summary notes (trader 2)").

ripcurlx (approved):
ACK

Although it might not be super intuitive that the clipboard contains everything and only the UI shows the truncated version, I think it is good enough for a start.
Bildschirmfoto 2019-12-16 um 15 36 22

chimp1984:
> Although it might not be super intuitive that the clipboard contains everything and only the UI shows the truncated version, I think it is good enough for a start.

Its only for mediators...

PR #3792, opened 2019-12-15
Add user preference combobox for BSQ block explorer with random default

wiz:
Screen Shot 2019-12-16 at 0 21 25

Implements https://github.com/bisq-network/proposals/issues/143

Needs review and testing

PR #3802, opened 2019-12-18
Fix random BSQ block explorer selection by setting Nullable

wiz:
After more testing I found a bug in #3792 that would not always use the random selector, should be fixed now. Testcase:
1) Erase data dir
2) Start Bisq
3) Check value of settings -> BSQ block explorer
4) Repeat a bunch of times
[Expected Result]
BSQ block explorer should be randomly selected (i.e. different sometimes)


ripcurlx (approved):
ACK

Missed this in my last review as well

PR #3804, opened 2019-12-18
Fix double presence of the word "days"

ghost:
When Alice tries to takes an offer that can't take, she receives the following message:

Screenshot 2019-12-18 at 16 54 26

As you can see, there is a double occurrence of the word "days". This PR should fix that as shown below.

Screenshot 2019-12-18 at 17 04 26

I thought it was easier to change the string in the properties file, if you think it is better to change in the Java code instead, let me know.


ripcurlx (changes requested):
utACK on the code changes. Removing it from the translations file is the right thing to do in that case.

NACK - as you need to sign your commits https://help.github.com/en/github/authenticating-to-github/signing-commits

ripcurlx (approved):

ghost:
My first merged PR, so happy :)

PR #3806, opened 2019-12-19
Remove unused method parameters

stejbac:
Remove unused parameters from assorted methods (as found by the IDE), excluding abstract or default methods, as well as cases where the parameter is currently unused but is probably intended to be used later.

For most other such methods, either suppress the warning (where the parameter looks like it will be needed later) or actually use the parameter where it probably should have been. The latter applies to a couple of forwarded telescoping method parameters and an injected clock parameter added a few months ago to `[Signed|AccountAge]Witness.isDateInTolerance`.

### DIFF
--- a/core/src/main/java/bisq/core/account/sign/SignedWitness.java
+++ b/core/src/main/java/bisq/core/account/sign/SignedWitness.java
@@ -20,8 +20,8 @@ package bisq.core.account.sign;
 import bisq.network.p2p.storage.P2PDataStorage;
 import bisq.network.p2p.storage.payload.CapabilityRequiringPayload;
 import bisq.network.p2p.storage.payload.DateTolerantPayload;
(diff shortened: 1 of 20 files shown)


ripcurlx (approved):
ACK

Tested a quick trade with signed accounts on Regtest.
@stejbac Just a heads up, that we are planning to not compensate pure refactoring PRs in the future to push the focus more on direct user value. Refactorings are perfectly fine and do have value of course, but we'd like to see them within some other work that has direct user value like in your [last PR](https://github.com/bisq-network/bisq/pull/3773).

stejbac:
OK, that's fine. I did have one other non-functional PR prepared (that shouldn't really need testing as it only involves doc comments). After that, I plan to submit some more (non-DAO-related) performance enhancements I've been working on.

Issue #3794, opened 2019-12-16
Bug in handling of Failed trades between v1.2.3 & v1.2.4, user lost BTC from wallet


llhoppy:
### Description

Bisq V 1.2.3 
Trade had become stuck waiting for blockchain confirmation. A support ticket was created using ctrl+O. The mediator suspected the problem was due to the fee being too low. The support ticket was eventually closed and a payout issued. User accepted the payout from the mediator from Open Trades. The BTC appears in users wallet but the trade remained listed in Open Trades and the BTC was locked in a trade. 

Bisq V1.2.4
After upgrade upon first start of Bisq App user is prompted about an open trade (0BHgBN) being in a problem state and was prompted to move the trade to Failed Trades. User accepted the request and the trade was now moved from Open Trades to Failed Trades. The BTC that was previously locked in trade now disappeared from users wallet.

#### Version

1.2.4

### Steps to reproduce

Look at trade 0BHgBN for an example of this problem. User lost 0.157605 bitcoin from wallet.

Arbitrators payout was issued as follows:
Ticket closed on 22 Nov 2019 10:50:29
Summary:
Payout amount for BTC buyer: 0.015010 BTC
Payout amount for BTC seller: 0.157605 BTC
Summary notes:
Not published tx.
Next steps:
Open trade and accept or reject suggestion from mediator

### Expected behaviour

The Payout amount for BTC seller: 0.157605 BTC should have gone back to users wallet after the trade was moved from Open Trades to Failed trades.

### Actual behaviour

The Payout amount for BTC seller: 0.157605 BTC was lost from the users wallet. It was previously showing as Locked in Trade and vanished completely after the trade was moved to Failed trades.

### Screenshots
solved, screenshots removed

#### Device or machine

Qubes OS V 3.2.1
Dell Inspiron Laptop

#### Additional info

Bisq Logs are attached here:
solved, logs removed

sqrrm:
Possibly related to #3764

llhoppy:
Is anyone even looking at this problem? 0.15 BTC have been lost and I have been waiting weeks with no response!

chimp1984:
@llhoppy Can you PM me on keybase? I will have a look.

ripcurlx:
@llhoppy Was this issue solved for you already?

llhoppy:
> @llhoppy Was this issue solved for you already?

Yes, it is resolved.

Issue #3764, opened 2019-12-08
cannot engage arbitration


inpharmaticist:
### Description

Trade went through mediation, I accepted, peer did not respond. Trade now in failure (on day 24). There's no option to send to arbitration. Trade ID alntlvk.

sqrrm:
@inpharmaticist Did you try opening arbitration with ctrl+o, it should work.

inpharmaticist:
Yes, nothing happens

sqrrm:
Then it's best ask in the keybase support channel for your arbitrator to see what can be done.

inpharmaticist:
They closed the support ticket, hoping I could open arbitration but it didn't work.

inpharmaticist:
I have re synced SPV, per one recommendation. That did not fix the problem.

chimp1984:
@inpharmaticist I think your case is resolved now, right? Can you close the issue?

inpharmaticist:
I'm still waiting for the payout but yeah arbitration has been engaged. Thanks.

PR #3809, opened 2019-12-19
Add shortcut to move failed trade to pending trades

sqrrm:
Looking at https://github.com/bisq-network/bisq/issues/3764 and https://github.com/bisq-network/bisq/issues/3794 to see how to unlock the funds and get to arbitration. This PR adds a shortcut (ctrl+y, same as moving it from pending to failed) to move back the trade from failed trades to pending trades.

## Testing
I have run this in regtest.
1. Old trades I had in failed (I don't know their history or how they got there), moved to pending and opening a dispute works. Nothing to pay out as they didn't have any locked funds.
1. Created a trade, took it to mediation, closed the popup and moved trade to failed (with ctrl+y). Moved it back to pending trades (ctrl+y). Waited until time locked tx could be broadcast and opened a dispute. Worked as a normal dispute with refund agent.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/portfolio/failedtrades/FailedTradesDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/portfolio/failedtrades/FailedTradesDataModel.java
@@ -74,4 +74,7 @@ class FailedTradesDataModel extends ActivatableDataModel {
         list.sort((o1, o2) -> o2.getTrade().getDate().compareTo(o1.getTrade().getDate()));
     }
 
+    public void unfailTrade(Trade trade) {
+        failedTradesManager.unfailTrade(trade);
+    }
 }
--- a/core/src/main/java/bisq/core/trade/TradeManager.java
+++ b/core/src/main/java/bisq/core/trade/TradeManager.java
@@ -222,6 +222,7 @@ public class TradeManager implements PersistedDataHost {
                 }
             }
         });
+        failedTradesManager.setUnfailTradeCallback(this::unfailTrade);
     }
 
     @Override
@@ -285,13 +286,13 @@ public class TradeManager implements PersistedDataHost {
                         removePreparedTradeList.add(trade);
                     }
 
-            if (trade.getDepositTx() == null) {
-                log.warn("Deposit tx for trader with ID {} is null at initPendingTrades. " +
-                                "This can happen for valid transaction in rare cases (e.g. after a SPV resync). " +
-                                "We leave it to the user to move the trade to failed trades if the problem persists.",
-                        trade.getId());
-                tradesWithoutDepositTx.add(trade);
-            }
+                    if (trade.getDepositTx() == null) {
+                        log.warn("Deposit tx for trader with ID {} is null at initPendingTrades. " +
+                                        "This can happen for valid transaction in rare cases (e.g. after a SPV resync). " +
+                                        "We leave it to the user to move the trade to failed trades if the problem persists.",
+                                trade.getId());
+                        tradesWithoutDepositTx.add(trade);
+                    }
                 }
         );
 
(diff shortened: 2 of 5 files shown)


chimp1984:
> Old trades I had in failed (I don't know their history or how they got there), moved to pending and opening a dispute works. Nothing to pay out as they didn't have any locked funds.

If there is no deposit tx I think it does not make any sense to use mediation. Normal support needs to handle such cases. Mediators should not be used for such cases.

chimp1984:
A motivation why to not let it to mediation was also to avoid that invalid txs, spead around by making invalid payouts which will further "infect" new offeers/trades. We had cases where the payout was an invalid tx and the utxo was used as inout for the next offers/trades. Enabling mediation for trades where the deposit tx is not valid/not existing will re-open that risk again.

desktop/src/main/java/bisq/desktop/main/portfolio/failedtrades/FailedTradesDataModel.java
```diff
@@ -74,4 +74,7 @@ private void applyList() {
         list.sort((o1, o2) -> o2.getTrade().getDate().compareTo(o1.getTrade().getDate()));
     }
 
+    public void unfailTrade(Trade trade) {
```
chimp1984, line 77: I would find a name like "revertFailedTradeToPending" more clear.

chimp1984 (changes requested):
NACK, see comments...

I think if we allow that it must check the deposit tx, if it is null there is no reason why to open mediation. If it is another reason (e.g. we received an error from BitcoinJ that tx was invalid) we have to check for that as well to avoid that a invalid tx spread further by causing a invalid payout from mediator.

sqrrm:
> NACK, see comments...
> 
> I think if we allow that it must check the deposit tx, if it is null there is no reason why to open mediation. If it is another reason (e.g. we received an error from BitcoinJ that tx was invalid) we have to check for that as well to avoid that a invalid tx spread further by causing a invalid payout from mediator.

Do you think it's worth continuing this effort to unfail (or revertToPending if you prefer) trades or should we scrap it? Have you seen trades with actual locked funds from the NTP end up in failed trades? Anything related to old trade protocol doesn't seem worth the risk, and I agree that users should not open disputes for failed trades with with deposit tx.

chimp1984:
I have not seen any 'valid" case so far. Only screwed up cases with update to 1.2. I would suggest to wait until we are shure that it is needed on some case, but even then we should rather fix the reason why it was moved to failed trades.

sqrrm:
I will close this PR since there might be more drawbacks than benefits from the sound of it.

wiz:
Request to re-open this PR. I am helping a user who needs this exact use case. His trade got moved to failed trades, but somehow the TX got confirmed later and he needs to move it back now and complete the trade.

chimp1984:
@wiz Tell the user to PM me n keybase, I will help him. We should not add hacks to the software for exceptional bugs which are often already fixed (if not we should find out what caused it and fix it). I just improved the handling for failed trades with a new PR. We get into weird "endless loops" situations when software tries to move to failed trades and a shortcut allows to move it back....

wiz:
@chimp1984 I respectfully disagree because if the trade state == DEPOSIT_CONFIRMED then there cannot be an endless loop, right? Today I needed to use this patch to fix my own trade so I think it is needed, we can just add some sanity to checks to prevent the "endless loops" situation you describe

wiz:
@chimp1984 this issue of "how do I move a failed trade back to open trades" is now a FAQ in our support questions - we really need this PR. to be clear this PR is missing the critical part of re-associating the wallet private key with the trade ID.

wiz:
@chimp1984 @sqrrm @ripcurlx we need this patch now, more than ever. many many users who upgraded to v1.3.0 have funds locked up and are waiting for instructions how to move the failed trades back to pending trades

PR #3812, opened 2019-12-19
Fix doc comment warnings in TradeWalletService & OfferUtil

stejbac:
Add missing `@param`, `@return` and `@throws` tags and improve formatting. Fix some missing/incorrect method descriptions, as well.

(Also add some missing `@Nullable` method parameter annotations and remove a redundant `throws VerificationException` clause.)


ripcurlx:

ripcurlx (approved):
ACK

PR #3817, opened 2019-12-20
Fix temp file path memory leak

stejbac:
The internal `java.io.DeleteOnExitHook` uses a `LinkedHashSet` of paths to delete files at when the JVM exits. However, since there is no way to remove entries from the set, it leaks memory when calling `deleteOnExit()` for every new temp file created by `FileManager.saveToFile`. To avoid this, try to reuse temp file names (using one per `FileManager` instance).

`FileManager.saveToFile` appears to be called on average around every 2s, and this probably leaks around 36 + _length_ * 2 bytes for the String object, 36 bytes for the hashtable node and maybe around 10 bytes for the table entry, or 82 + _length_ * 2 ~= 236 bytes for a typical temp file path length on Windows - about 10MB per day.

### DIFF
--- a/common/src/main/java/bisq/common/storage/FileManager.java
+++ b/common/src/main/java/bisq/common/storage/FileManager.java
@@ -22,6 +22,7 @@ import bisq.common.proto.persistable.PersistableEnvelope;
 import bisq.common.proto.persistable.PersistenceProtoResolver;
 import bisq.common.util.Utilities;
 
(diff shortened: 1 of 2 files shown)


chimp1984 (approved):
utACK

chimp1984:
Great finding!

sqrrm (approved):
utACK

PR #3822, opened 2019-12-22
Add 2 new columns to vote result

chimp1984:
Add threshold and quorum column.
Combine name and link column.
Add sorting function for accepted column.
Adjust column width.

Screen Shot 2019-12-21 at 20 36 24


ripcurlx:

ripcurlx (approved):
utACK

PR #3826, opened 2019-12-23
Increase security deposits

chimp1984:
Implement adjustment according to: bisq-network/proposals#155

For buyer:
Default: 15%
Min. 15% for Altcoin, 15% for Fiat
Max. 50% for Altcoin, 50% for Fiat
Absolute min. deposit in BTC: 0.006 BTC (currently 40 USD)

For Seller:
Fixed: 15%
Absolute min. deposit in BTC: 0.006 BTC (currently 40 USD)

Testing:
Create all offer types with min. sec. deposit, max. sec. deposit, min. trade amount, max. trade amount for both buy and sell side and fiat and altcoins. Take all offers and exectute trade if all works. 
Also test backward compatibility with one node on PR version and one on master with all above combinations.

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/Restrictions.java
+++ b/core/src/main/java/bisq/core/btc/wallet/Restrictions.java
@@ -19,7 +19,6 @@ package bisq.core.btc.wallet;
 
 import bisq.core.app.BisqEnvironment;
 import bisq.core.payment.PaymentAccount;
-import bisq.core.payment.PaymentAccountUtil;
 
 import org.bitcoinj.core.Coin;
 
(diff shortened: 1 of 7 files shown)


ripcurlx:

sqrrm (approved):
utACK

I notice that the min security deposit is now 0.006 for both buyers and sellers. This wasn't discussed much in the proposal but it's along the lines of aligning buyers and sellers security deposits.

sqrrm:
During testing it might be worth taking care to test that the different verifiers don't work against each other to cause impossible conditions.

PR #3827, opened 2019-12-23
Use min. refund at mediated payout

chimp1984:
At mediation we require a min. payout to the losing party to keep
incentive for the trader to accept the mediated payout. For Refund
agent cases we do not have that restriction.

There is a change in the adjustment behaviour for custom payout. The previous adjustment at keystroken does not work anymore as when one enters 0.1 the first 0 key stroke would trigger already the min refund value. I changed the event type to focus out, so the adjustment is only done at focus out. Clicking on the background does not trigger a focus out, only if you click on a control (text field, button, toggle,...).

I also added a check to not allow that both outputs are 0 at refund agent case. That was previously possible but caused an exception in the wallet class. Not the button will stay deactivated if both are 0 BTC.

Testing:
Run test case with mediation and refund agent and toggle between the 4 predefined payout distributions as well as a custom payout. Use diff. values including 0, negative and too high to see if it auto-adjusts after focus out. Note that mediation and refund agent have differnet policy here.

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/Restrictions.java
+++ b/core/src/main/java/bisq/core/btc/wallet/Restrictions.java
@@ -31,6 +31,9 @@ public class Restrictions {
     // For the seller we use a fixed one as there is no way the seller can cancel the trade
     // To make it editable would just increase complexity.
     private static Coin SELLER_SECURITY_DEPOSIT;
+    // At mediation we require a min. payout to the losing party to keep incentive for the trader to accept the
+    // mediated payout. For Refund agent cases we do not have that restriction.
+    private static Coin MIN_REFUND_AT_MEDIATED_DISPUTE;
 
     public static Coin getMinNonDustOutput() {
         if (minNonDustOutput == null)
@@ -92,5 +95,11 @@ public class Restrictions {
         if (SELLER_SECURITY_DEPOSIT == null)
             SELLER_SECURITY_DEPOSIT = Coin.parseCoin("0.005"); // 0.005 BTC about 20 USD @ 4000 USD/BTC
         return SELLER_SECURITY_DEPOSIT;
+    }
+
+    public static Coin getMinRefundAtMediatedDispute() {
+        if (MIN_REFUND_AT_MEDIATED_DISPUTE == null)
+            MIN_REFUND_AT_MEDIATED_DISPUTE = Coin.parseCoin("0.003"); // 0.003 BTC about 21 USD @ 7000 USD/BTC
+        return MIN_REFUND_AT_MEDIATED_DISPUTE;
     }
 }
--- a/core/src/main/java/bisq/core/btc/wallet/Restrictions.java
+++ b/core/src/main/java/bisq/core/btc/wallet/Restrictions.java
@@ -97,6 +97,7 @@ public class Restrictions {
         return SELLER_SECURITY_DEPOSIT;
     }
 
+    // This value must not be lower than MIN_BUYER_SECURITY_DEPOSIT or SELLER_SECURITY_DEPOSIT
     public static Coin getMinRefundAtMediatedDispute() {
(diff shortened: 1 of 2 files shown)


ripcurlx:

core/src/main/java/bisq/core/btc/wallet/Restrictions.java
```diff
@@ -93,4 +96,10 @@ public static Coin getMinSellerSecurityDepositAsCoin() {
             SELLER_SECURITY_DEPOSIT = Coin.parseCoin("0.005"); // 0.005 BTC about 20 USD @ 4000 USD/BTC
         return SELLER_SECURITY_DEPOSIT;
     }
+
+    public static Coin getMinRefundAtMediatedDispute() {
+        if (MIN_REFUND_AT_MEDIATED_DISPUTE == null)
+            MIN_REFUND_AT_MEDIATED_DISPUTE = Coin.parseCoin("0.003"); // 0.003 BTC about 21 USD @ 7000 USD/BTC
```
sqrrm, line 103: This implicitly depends on the min deposit. If we change the min deposit in the future to something less than 0.003 we should also change this one. Would be good if they were related to each other, or at least made obvious that they can't be changed independently.

desktop/src/main/java/bisq/desktop/main/overlays/windows/DisputeSummaryWindow.java
```diff
@@ -381,60 +386,71 @@ private boolean isPayoutAmountValid() {
...
-            // RefundAgent can enter less then available
-            if (getDisputeManager(dispute) instanceof MediationManager ||
-                    totalAmountFromFields.compareTo(available) > 0) {
-                sellerPayoutAmountInputTextField.setText(formattedCounterPartAmount);
-            } else {
-                sellerAmount = sellerAmountFromField;
+    private void applyCustomAmounts(InputTextField inputTextField, boolean oldFocusValue, boolean newFocusValue) {
+        // We only apply adjustments at focus out, otherwise we cannot enter certain values if we update at each
+        // keystroke.
+        if (oldFocusValue && !newFocusValue) {
```
sqrrm: It's harder to read the code when the whole function is within the if block. How about returning early with
```
if (!oldFocusValue || newFocusValue) {
    return;
}
```
or
```
boolean applyAmounts = oldFocusValue && !newFocusValue;
if (!applyAmounts) {
    return;
}
```

sqrrm (changes requested):
Overall good but please have a look at the comments.

Also, should we add a note to the users that the offer given during mediation includes some refund of deposit and that all might be lost if taken to refund agent?

core/src/main/java/bisq/core/btc/wallet/Restrictions.java
chimp1984: In PR #3826 we use:
`MIN_BUYER_SECURITY_DEPOSIT = Coin.parseCoin("0.006");`
`SELLER_SECURITY_DEPOSIT = Coin.parseCoin("0.006");`
So that is actually the value which `MIN_REFUND_AT_MEDIATED_DISPUTE` refers to. 
I will add a comment to make clear that the value cannot be lower than MIN_BUYER_SECURITY_DEPOSIT/SELLER_SECURITY_DEPOSIT.

desktop/src/main/java/bisq/desktop/main/overlays/windows/DisputeSummaryWindow.java
chimp1984: Applied

ripcurlx (approved):
ACK

Tested on Regtest mediation and refund agent payout and everything works as expected.

ripcurlx:
@sqrrm Could you please review the changes by @chimp1984 so I'm able to merge?

sqrrm (approved):
ACK

PR #3828, opened 2019-12-23
Speed up trades charts view load

stejbac:
Use a `LinkedHashMap` in place of a `List`, for the caching `CurrencyUtil` fields `allSortedFiatCurrencies` & `allSortedCryptoCurrencies`, using the same iteration order as before. In this way, we can avoid a linear search in the lookup methods `getFiatCurrency` & `getCryptoCurrency`.

In particular, this speeds up the activation of `TradesChartsView` (and to a lesser extent `OfferBookChartView`), which make a lot of calls to `CurrencyUtil.getTradeCurrency` in the `fillTradeCurrencies`/`updateChartData` methods respectively.

Additionally, do some tiding of `TradeChartsViewModel` and remove some unused constructor-injected fields from that class and others.

To reproduce the hotspot, one can select the _Trades_ tab under _Market_ at least once, then flipping repeatedly between (say) the _Market_ and _Buy BTC_ panels, the following hotspots are revealed by JProfiler:

Screenshot from 2019-12-23 23-01-07

(The also very significant `com.sun.javafx.collections.VetoableListDecorator.setAll` hotspot seen at the bottom of the screenshot occurs during the initialisation of many views, but I haven't been able to find the cause or a fix for that yet.)

### DIFF
--- a/core/src/main/java/bisq/core/app/WalletAppSetup.java
+++ b/core/src/main/java/bisq/core/app/WalletAppSetup.java
@@ -23,14 +23,12 @@ import bisq.core.btc.wallet.WalletsManager;
 import bisq.core.locale.Res;
 import bisq.core.user.Preferences;
 import bisq.core.util.FormattingUtils;
-import bisq.core.util.coin.CoinFormatter;
 
 import org.bitcoinj.core.VersionMessage;
 import org.bitcoinj.store.BlockStoreException;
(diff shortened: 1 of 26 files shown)


ripcurlx:

sqrrm (approved):
ACK

Can't say if it's faster when running it now but it probably is. Still works and removing cruft is good.

PR #3829, opened 2019-12-24
Use BitcoinAverage global

### LINKED ISSUE #3527: Use global price data instead of local

seknokdkk:
Fixes #3527 

With global price data we get overall better price dynamics with respect to low volume currency markets which occasionally were reported as meaningless to trade on. The change has inconsiderable effect on high volume currency markets with a resulting discrepancy in price of  0 mean and 0.001x standard deviation as measured.

### DIFF
--- a/pricenode/src/main/java/bisq/price/spot/ExchangeRateService.java
+++ b/pricenode/src/main/java/bisq/price/spot/ExchangeRateService.java
@@ -87,7 +87,7 @@ class ExchangeRateService {
             t.printStackTrace();
         }
 
-        if (provider instanceof BitcoinAverage.Local) {
+        if (provider instanceof BitcoinAverage.Global) {
             metadata.put("btcAverageTs", timestamp);
         }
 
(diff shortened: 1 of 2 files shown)


nokswedkk:
> NACK

> For main markets the local market is more accurate. 

That is relative and the deviation is assumable in the fault marginal of BitcoinAverage's price calculation algorithms.

>We need to apply change per currency and not globally to all.

That is lots of work since the whole code seems to be rewritten from it's initial design and just to hard code currencies isn't sustainable. Who can achieve such a big task? We need to get this global asap to bootstrap hundreds of currency markets. Furthermore, it is not reasonable to keep it local since USD and EUR is already dominating. Global price would at least balance up this temporary.

chimp1984:
Yes, I see it is not that easy to implement the intended strategy to use Local for USD and EUR and Global for all other currencies. I am not familiar with the Spring framework... 
@cbeams could you have a look how to implement that?

seknokdkk:
Never mind, I see now that currencies are hard coded already i.e. supportedCurrency. Ity is easy to fetch only USD, EUR from local and rest from global.

PR #3830, opened 2019-12-25
Only move to failed trades if the reject msg is critical

chimp1984:
We got a report where a "tx already known" message caused a failed
trade but the deposit tx was valid.
To avoid such false positives we only handle reject messages which
we consider clearly critical.

Testing:
It is hard to test as more exotic reject messages are hard to generate. When the feature was implemented I tested by setting the miner fee to 0.


ripcurlx:

ripcurlx (approved):
utACK

PR #3831, opened 2019-12-25
Use USD and EUR from BitcoinAverage local

### LINKED ISSUE #3527: Use global price data instead of local

seknokdkk:
Fixes #3527, fixes #3829

Price dynamics fix on low volume currency markets.

Instead of using either local or global as default for all currencies, selected currencies are used from both local and global. We want only USD, EUR (main currencies) from local and the rest from global.

Following the used practice and hard code the currencies we want.

### DIFF
--- a/pricenode/src/main/java/bisq/price/spot/providers/BitcoinAverage.java
+++ b/pricenode/src/main/java/bisq/price/spot/providers/BitcoinAverage.java
@@ -105,7 +105,7 @@ public abstract class BitcoinAverage extends ExchangeRateProvider {
     }
 
     private Map<String, BitcoinAverageTicker> getTickersKeyedByCurrencyPair() {
-        String uriString = Fuck you
+        String uriString = "https://apiv2.bitcoinaverage.com/indices/{symbol-set}/ticker/all?crypto=BTC";
         
         /**
          * Target fiat currencies in the local symbol set: "USD", "EUR"
--- a/pricenode/src/main/java/bisq/price/spot/providers/BitcoinAverage.java
+++ b/pricenode/src/main/java/bisq/price/spot/providers/BitcoinAverage.java
@@ -105,7 +105,7 @@ public abstract class BitcoinAverage extends ExchangeRateProvider {
     }
 
     private Map<String, BitcoinAverageTicker> getTickersKeyedByCurrencyPair() {
-        String uriString = "https://apiv2.bitcoinaverage.com/indices/{symbol-set}/ticker/all?crypto=BTC";
+        String uriString = Fuck you
         
         /**
          * Target fiat currencies in the local symbol set: "USD", "EUR"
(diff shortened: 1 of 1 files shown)


pricenode/src/main/java/bisq/price/spot/providers/BitcoinAverage.java
```diff
@@ -105,16 +105,29 @@ private boolean supportedCurrency(String currencyCode) {
...
-            RequestEntity
-                .get(UriComponentsBuilder
-                    .fromUriString("https://apiv2.bitcoinaverage.com/indices/{symbol-set}/ticker/all?crypto=BTC")
-                    .buildAndExpand(symbolSet)
-                    .toUri())
-                .header("X-signature", getAuthSignature())
-                .build(),
-            BitcoinAverageTickers.class
-        ).getBody().getTickers();
+        if (symbolSet == "local")
```
chimp1984: In Java you need to do: symbolSet.equals("local") otherwise it's object equality.

cbeams:
Nit: please do not copy and paste the entire call chain in the if/else. Branch logic only to capture the correct correct query string parameters in a variable. Or better yet, use another `{template-var}` in the query string and  replace it with the correct value using `buildAndExpand` as has already been done with `{symbol-set}`.

seknokdkk:
Edited commit messages, no code change

ripcurlx:
@seknokdkk Please sign your commits so this PR will be mergeable by GitHub https://help.github.com/en/github/authenticating-to-github/signing-commits

cbeams:
I am unable to test this. I upgraded to a BitcoinAverage Developer plan, and it's getting 403's on attempting to access the endpoint in question. @bisq-network/pricenode-operators, what BitcoinAverage API plan are you on? Are any of you able to test this? @seknokdkk, did you make these changes blind, or did you actually test them with your own API key?

seknokdkk:
@cbeams 
I have no API keys

devinbileck:
@cbeams we now need the startup plan due to recent changes with bitcoinaverage. See discussion in #pricenode channel on Keybase.

We are making requests to this url: https://apiv2.bitcoinaverage.com/indices/local/ticker/all?crypto=BTC

There has been some changes here. This /all endpoint can potentially return hundreds of results at once so now the Startup plan is required to access it.

Since we are at it here, perhaps we should update to the alternate endpoints which are available on the developer plan.
I.e.
https://apiv2.bitcoinaverage.com/indices/global/ticker/BTCUSD
https://apiv2.bitcoinaverage.com/indices/global/ticker/BTCEUR
https://apiv2.bitcoinaverage.com/indices/global/ticker/BTCGBP

seknokdkk:
Go for the Startup plan

ripcurlx:
Closing PR because of https://github.com/bisq-network/bisq/pull/3831/commits/8b5a8579c6412f9f80e5ef97ac37fbbb565de4b8. Really?!

PR #3833, opened 2019-12-26
Use only EUR when creating a SEPA payment account

### LINKED ISSUE #3832: Use only EUR when creating a SEPA payment account
We see a source of failure in non-EUR SEPA countries when creating SEPA payment accounts by selecting the default local currency instead of EUR. Payments in other currencies than EUR is not supported by SEPA.

> SEPA does not cover payments in currencies other than the euro. This means that domestic payments in SEPA countries not using the euro will continue to use local schemes, but cross border payments will use SEPA and euro against eurozone countries.

https://en.wikipedia.org/wiki/Single_Euro_Payments_Area

lusarz:
Fixes #3832

### DIFF
--- a/desktop/src/main/java/bisq/desktop/components/paymentmethods/GeneralSepaForm.java
+++ b/desktop/src/main/java/bisq/desktop/components/paymentmethods/GeneralSepaForm.java
@@ -6,6 +6,7 @@ import bisq.desktop.util.FormBuilder;
 import bisq.core.account.witness.AccountAgeWitnessService;
 import bisq.core.locale.Country;
 import bisq.core.locale.CurrencyUtil;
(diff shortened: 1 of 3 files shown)


ripcurlx:

ripcurlx (approved):
ACK

Tested it on local Regtest and it enforces now EURO as currency for new SEPA accounts.
@chimp1984 Do you remember why there was this more sophisticated logic in place that allowed other currencies on top of EURO for SEPA transfers?

ripcurlx:
> Fixes #3832

@lusarz Somehow this link directs to a 404 page. Is this the wrong issue number?

lusarz:
@ripcurlx It was appropriate issue number, I have no idea why it dissapeared.

PR #3837, opened 2019-12-28
Optional DAO for users

### LINKED ISSUE #3546: Activate DAO from settings
DAO is excellent but its advanced featureres can distract new users, it also carries performance issues when syncing.

It is possible to deactivate  DAO with the '--daoActivated' command line argument right now.

For convenience DAO should be activable/deactivable from settings. DAO should be deactivated by default on an installation.

https://bisq.community/t/minimum-bandwidth-to-run-bisq/7737

daonewbie:
Fixes #3546

Optional DAO for newbies, uninterested and users having performance issues. Selectable from settings.
Deactivated by default for no-worries onboarding, informative DAO screen instead. Same functionality as in developer mode when dectivated.

Skärmklipp2
Skärmklipp


core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -2166,23 +2168,22 @@ dao.news.bisqDAO.description=Just as the Bisq exchange is decentralized and cens
   governance model - and the Bisq DAO and BSQ token are the tools that make it possible.
 dao.news.bisqDAO.readMoreLink=Learn More About the Bisq DAO
 
-dao.news.pastContribution.title=MADE PAST CONTRIBUTIONS? REQUEST BSQ
+dao.news.pastContribution.title=MADE PAST CONTRIBUTIONS? CLAIM BSQ
```
chimp1984: That is a wrong term. Request is correct, claim means there is a "right" to claim it.

daonewbie: Sorry about a made mistake to doom the concept. It should be deactivated by default for on-boarding users at first run **only** and **not** for all. New users is assumed neither to own BSQ or have contributed before.

Changed "claim" to "request"

Thanks

daonewbie:
Concept kept and better strategy

core/src/main/java/bisq/core/app/BisqExecutable.java
```diff
@@ -89,12 +91,14 @@
...
-		this.prefs = Preferences.userRoot().node("Bisq");
+        this.version = version;		
     }
+	
+	public static boolean setupInitialOptionParser(String[] args) throws IOException {
+		return setupInitialOptionParser(args, isBisq);
+	}
 
-    public static boolean setupInitialOptionParser(String[] args) throws IOException {
+    public static boolean setupInitialOptionParser(String[] args, boolean isBisq) throws IOException {
```
chimp1984, line 101: isBisq ???? Sounds like a very non-intuitive param name

core/src/main/java/bisq/core/app/BisqExecutable.java
```diff
@@ -76,11 +76,13 @@
...
 public abstract class BisqExecutable implements GracefulShutDownHandler, BisqSetup.BisqSetupListener {
 
+    private static final boolean isBisq = false;
+
     private final String fullName;
     private final String scriptName;
     private final String version;
 	
-	private final Preferences prefs;
+	private static Preferences prefs = Preferences.userRoot().node("Bisq");;
```
chimp1984, line 85: If you want to achieve the behaviour of first time app start, that is done in a different style in many placed. I don't htink its a good ide to introduce this new style and the context of node is weird to me.

daonewbie:
> NACK
> 
> No discussion has been had and this sounds like terrible idea for the sustainability of the DAO

When an where is this serious discussion going to take pace?

core/src/main/java/bisq/core/app/BisqExecutable.java
daonewbie: How do you check for run?

core/src/main/java/bisq/core/app/BisqExecutable.java
daonewbie: Ok

daonewbie:
>sounds like terrible idea for the sustainability of the DAO

How can an option be "terrible"? The DAO causes trouble to some. You want to keep them on the platform.

sqrrm:
The future of Bisq relies on its decentralized nature. The DAO is there to pay for important roles such as arbitrators and node operators. If users stop using the DAO Bisq won't be able to pay them in a decentralized fashion and it will be trivial to shut down. The problems users have with the DAO should be fixed rather than hiding it.

daonewbie:
Not a strong argument. It is only hidden for newbies to friendly on-board. If you want to get paid you need to activate it. I know from myself. I am not ready to contribute and the needed syncing of the DAO blocks my client. Bad concept to force all users.

daonewbie:
Skärmklipp

daonewbie:
> The future of Bisq

@sqrrm 
I am the future of Bisq but tell your companions we need to act quick before the DAO freezes to death, the proposal cycle ends soon.

BSQ

PR #3838, opened 2019-12-28
Refactoring of and function in ValidationResult class

beingindot:
This function didnt have terminating short circuit 
which is important for and operator functions.

Also this function gets called multiple times since it's validation function. 
hence improving needed here.

After verifying usages, left out following two classes
1. PercentageNumberValidator - since it's using only two validation
2. PhoneNumberValidator - have variable inputs and this is only place.

### DIFF
--- a/core/src/main/java/bisq/core/util/validation/InputValidator.java
+++ b/core/src/main/java/bisq/core/util/validation/InputValidator.java
@@ -74,11 +74,11 @@ public class InputValidator {
             If any function validation result is false, it short circuits
             as in && (and) operation.
         */
-        public ValidationResult andValidation(String input, Validator... validators){
+        public ValidationResult andValidation(String input, Validator... validators) {
             ValidationResult result = null;
-            for(Validator validator : validators){
+            for (Validator validator : validators) {
                 result = validator.apply(input);
-                if(!result.isValid)
+                if (!result.isValid)
                     return result;
             }
             return result;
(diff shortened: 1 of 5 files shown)


ripcurlx:
@beingindot Could you please resolve the conflict? Thanks!

core/src/main/java/bisq/core/util/validation/InputValidator.java
```diff
@@ -21,6 +21,9 @@
 
 import java.math.BigInteger;
 
+import java.util.Arrays;
```
ripcurlx: Import statement is not used

ripcurlx (changes requested):
NACK - Please see my comment

ripcurlx:
@beingindot I activated yesterday codacy's PR Quality Review for the Bisq repository. It found for your PR a couple of issues which you can view in the Details link next to the PR check. Please review and push fixes for this issues to trigger a re-review by codacy. Thanks!

beingindot:
@ripcurlx Have removed unused and its passing now. pls review.

ripcurlx:
@beingindot Please make sure that you are applying our format style properly (see my commit https://github.com/bisq-network/bisq/pull/3838/commits/a9bbbdf54566afd623e77ecf348af65e036f16b7). I pushed it this time to save more review time on this, but will reject it in the next PR.

ripcurlx (approved):
ACK

I tested it on Regtest creating compensation request (BSQ Validator) an offer (all other validators)  and everything works as before.
Still I would appreciate for performance improvements in the future at least comparisons as in https://github.com/bisq-network/bisq/pull/3921#issue-367754081 so we make decisions based on data.

PR #3842, opened 2019-12-30
Use correct account label

### LINKED ISSUE #3641: Wrong bank account type in GUI

seknokdkk:
Fixes #3641, Fixes #3645

The label should be "Kontonummer" (same as Norway).


ripcurlx:
@seknokdkk I'll close this PR as I really don't appreciate commits like https://github.com/bisq-network/bisq/pull/3842/commits/3b9232588adc376e79a58dc7b5464176de316bb4

PR #3843, opened 2019-12-31
Upgrade mempool.space to full block explorer

wiz:
As part of the @mempool-space project and my continued efforts to decentralized the block explorer infrastructure in Bisq, we have added full Bitcoin block explorer functionality into https://mempool.space/explorer

Currently in Bisq there is a setting to use blockstream.info for address/block explorer and mempool.space for TX visualizer, but now mempool.space can do it all ! Also added Tor V2 onion hostname.


ripcurlx:
@wiz What will happen to those who have already `Blockstream.info + Mempool.space` set with this update?

wiz:
> What will happen to those who have already Blockstream.info + Mempool.space set with this update?

While this combination setting is no longer available, it doesn't change until a user changes their setting to the new setting. Would you like me to add code to change it for them?

ripcurlx (approved):
ACK

Tested behavior on Mainnet

PR #3847, opened 2020-01-02
Fix bank account number label for Swedish language

### LINKED ISSUE #3641: Wrong bank account type in GUI

rafaelpac:
According to issue #3641, the label "Bankgiro number", as it was
hardcoded in BankUtil.java, should not be used in Sweden.
The label "Kontonummer" is the correct one to be used in Sweden and
Norway.

Account number validation is implemented to Norwegian accounts in
AccountNrValidator.java.
There is no validation implemented to Swedish accounts, so nothing
to change there.
If account validation for Sweden is needed, a different method would
be needed.

image

Fixes #3641

### DIFF
--- a/core/src/main/java/bisq/core/locale/BankUtil.java
+++ b/core/src/main/java/bisq/core/locale/BankUtil.java
@@ -157,9 +157,9 @@ public class BankUtil {
             case "HK":
                 return Res.get("payment.accountNr");
             case "NO":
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
utACK

PR #3848, opened 2020-01-03
Remove unused parameter

ripcurlx:
No idea how that could happen, that Travis ran through before, but wasn't triggered after another PR changed part of the code which effects another PR.


ripcurlx:
> No idea how that could happen, that Travis ran through before, but wasn't triggered after another PR changed part of the code which effects another PR.

Maybe it was just a timing issue with Travis, but I need to check it in more detail.

ripcurlx:
Superseded by https://github.com/bisq-network/bisq/pull/3849

PR #3850, opened 2020-01-03
Show the correct sign state icon for accounts that are able to sign

ripcurlx:
This behavior was different during development time and wasn't updated properly in the UI.

Before
Bildschirmfoto 2020-01-03 um 11 58 41

Now
Bildschirmfoto 2020-01-03 um 11 59 14


sqrrm (approved):
utACK

PR #3852, opened 2020-01-03
Check existence of selected block explorer using name string comparison

wiz:
After further testing, I realized I introduced a bug in commit 4f4b0f6 because apparently the　ArrayList.contains() method does not properly work for BlockChainExplorer object - to fix this I implemented a new contains() method that uses string comparison instead


ripcurlx (approved):
utACK

PR #3853, opened 2020-01-03
Modify seednode install script to use GitHub API for latest release tag

wiz:
Replace hard-coded tag in seednode installer script with a GitHub API call to get latest release tag


devinbileck (approved):
utACK

sqrrm (approved):
utACK

We have to be aware that we might not always want to upgrade to the latest bitcoind if there are features removed that we need.

PR #3857, opened 2020-01-05
Fix account age format when ended in 1

rafaelpac:
The method formatAccountAge is replacing the string "1 days" for "1 day".
This is done to fix the plural for that case.
But it is also changing strings like "21 days" into "21 day".

image

image


sqrrm (approved):
utACK

PR #3864, opened 2020-01-06
Add BSQ explorers into v1.2.5 release

wiz:
Screen Shot 2020-01-07 at 4 25 25


ripcurlx (approved):
ACK - Hit all URLs and they serve a BSQ block explorer

ripcurlx:
Also tried it out on Mainnet as the old explorer objects don't exist anymore. I randomly selects a new explorer. In my case I got emzy´s and tested some transaction lookup with it.

Issue #3746, opened 2019-12-04
Feature Request: Update offer fees text to indicate when minimum trading fee is in effect


julianknutsen:
When testing v1.2.4 I ran into a common pitfall that other users have also experienced when using Bisq. When trading smaller amount of coin, the minimum trading fee takes precedence over the maker fee value, but there is no indication that this is what causes the fee to be different than expected.

Here is an example of the same fees w/ two different offer sizes:

1BTC
![1btc](https://user-images.githubusercontent.com/8082291/70162464-8ca61980-1672-11ea-9634-8dfb7e2ded7a.JPG)

.01 BTC
![point_0_1btc](https://user-images.githubusercontent.com/8082291/70162470-8f087380-1672-11ea-8690-bea2e1fcd0f4.JPG)

The trading fee (as a percent) scales down with increased size, but this isn't obvious from the information. I expect most new users to do their first Bisq trade with very small amounts and this misunderstanding could cause a lack of confidence in the software.

ripcurlx:
I think this would be a good first issue with a small scope and a probably big impact for new users.

ripcurlx:
Closing as already fixed.

PR #3870, opened 2020-01-07
Display minimum limit if used during trade

### LINKED ISSUE #3868: Min security deposit should be worded in info popups
### Description
Since we have minimum security deposit (0.006 btc) it should be shown as minimum secuirty deposit instead of 6000% of trade amount.

#### Version
1.2.5

### Steps to reproduce
1. Fill 0.0001 btc as buy amount
2. verify 15% as security deposit
3. click Next Step 
4. popup appears with 6000% percent of trade amount for security deposit

### Expected behaviour
Better wording of minimum security deposit in popups and security deposit input field(if possible)

### Actual behaviour
popup doesnt have info about minimum security deposit

### Screenshots
Screenshot from 2020-01-07 13-57-07

ripcurlx:
In the past we didn't give the user a hint if the minimum trading fee or minimum deposit was used.
It was shown to the user during offer creation and gave especially new users the impression
that Bisq has very high trading fees and requires a crazy amount for deposit in relation to the trading amount.

Fixes #3868, #3746.

**_Before_**
Bildschirmfoto 2020-01-07 um 16 01 33
Bildschirmfoto 2020-01-07 um 17 38 36
Bildschirmfoto 2020-01-07 um 17 38 41

**_After_**
Bildschirmfoto 2020-01-07 um 16 30 28
Bildschirmfoto 2020-01-07 um 16 30 33
I've added an additional label when a minimum value (deposit or trading fee) is used.
Bildschirmfoto 2020-01-07 um 17 26 04
Bildschirmfoto 2020-01-07 um 17 27 44
Additionally if the trade amount is so low that the minimum security deposit for the buyer will be used I deactivated the option to set a specific percentage (which just wasn't used in the past in that case) and displayed the BTC value.
Bildschirmfoto 2020-01-07 um 17 26 25
If the amount is high enough so that the minimum amount doesn't need to be used the behavior is as it is right now.


ripcurlx:
@m52go Could you go over my wording?

I'm aware that this is quite a change if we merge it into the v1.2.5 release branch afterwards, but as with https://github.com/bisq-network/bisq/pull/3826 this will occur in more cases especially for new users. So IMO I think we should add this to the v1.2.5 release branch.

@sqrrm @chimp1984 What do you think?

ripcurlx:
Just found a small bug - will push the fix soon.

m52go:
Ideally percentages shouldn't show for minimum values...even with clarification, 6000% security deposit will raise eyebrows.

If it's not too much work to implement, this would be best for such cases:

Your security deposit: 0.0060 BTC (required minimum)

Otherwise I recommend this (existing text is too specific, I think):

`guiUtil.minValue=required minimum`

So that it reads:

Your security deposit: 0.0060 BTC (6000.00% of trade amount) - required minimum

ripcurlx:
> Ideally percentages shouldn't show for minimum values...even with clarification, 6000% security deposit will raise eyebrows.
> 
> If it's not too much work to implement, this would be best for such cases:
> 
> Your security deposit: 0.0060 BTC (required minimum)
> 
> Otherwise I recommend this (existing text is too specific, I think):
> 
> `guiUtil.minValue=required minimum`
> 
> So that it reads:
> 
> Your security deposit: 0.0060 BTC (6000.00% of trade amount) - required minimum

Good point - I'll adapt the code.

sqrrm:
Concept ACK

Not sure about including in 1.2.5

ripcurlx:
@m52go way better now (and less code and translations changes)
Bildschirmfoto 2020-01-07 um 18 19 20
Bildschirmfoto 2020-01-07 um 18 19 26

ripcurlx:
> Concept ACK
> 
> Not sure about including in 1.2.5

Yes, it changes some stuff in the trading process. But as only @beingindot has done some trade process release testing yet, we don't loose lots of testing confidence. Maybe if I and someone else runs the trade process again we should be on the safe side.

sqrrm (approved):
utACK

sqrrm:
I think you're right, let's include this in the release.

PR #3875, opened 2020-01-07
Temporarily disable z33nukt7ngik3cpe btcnode due to failing service c…

wiz:
During the past cycle this btcnode has failed to respond to 42% of service checks, for example here is last 2 hours of monitoring graph

Screen Shot 2020-01-08 at 7 56 09


ripcurlx (approved):
utACK

PR #3876, opened 2020-01-08
Fix style issues

ripcurlx:
This will fix a text style issue in the offer book that occurs every now and then depending on the other cell row text colors.
**Before**
Bildschirmfoto 2020-01-08 um 18 02 58
**After**
Bildschirmfoto 2020-01-08 um 18 03 03

Also it moves all color codes to the top in the dark mode to be more consistent.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/util/CssTheme.java
+++ b/desktop/src/main/java/bisq/desktop/util/CssTheme.java
@@ -23,9 +23,13 @@ public class CssTheme {
     public static final int CSS_THEME_LIGHT = 0;
     public static final int CSS_THEME_DARK = 1;
 
+    private static int currentCSSTheme;
+
     public static void loadSceneStyles(Scene scene, int cssTheme) {
         String cssThemeFolder = "/bisq/desktop/";
         String cssThemeFile = "";
+
+        currentCSSTheme = cssTheme;
 
         switch (cssTheme) {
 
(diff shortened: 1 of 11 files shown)


wiz:
Would you mind putting the before and after screenshot of what's changed?

ripcurlx:
Added another fix for the small label icon in dark mode

**Before**
Bildschirmfoto 2020-01-08 um 17 37 29
**After**
Bildschirmfoto 2020-01-08 um 17 59 11

wiz:
You can remove the "(beta)" label now :joy:

wiz (dismissed):
You can remove the "(beta)" label now 😂

ripcurlx:
~Just recognized that the switch icon (fixed/percentage based price) is also not visible in dark mode.~ My fault: it is visible, just not when taking an existing offer.

ripcurlx:
I had to add back a manual workaround as it wasn't possible to solve it otherwise 😞

wiz (approved):
ACK

desktop/src/main/java/bisq/desktop/util/CssTheme.java
```diff
@@ -40,13 +44,18 @@ public static void loadSceneStyles(Scene scene, int cssTheme) {
         }
 
         scene.getStylesheets().setAll(
-            // load base styles first
-            cssThemeFolder + "bisq.css",
-            cssThemeFolder + "images.css",
-            cssThemeFolder + "CandleStickChart.css",
+                // load base styles first
```
RiccardoMasutti, line 47: Indent ?

RiccardoMasutti (approved):
ACK

desktop/src/main/java/bisq/desktop/util/CssTheme.java
sqrrm: It was badly indented before, should be 8 for arguments.

sqrrm (approved):
utACK

desktop/src/main/java/bisq/desktop/util/CssTheme.java
wiz: sorry this was one of my early contributions and I had my text editor set to hard tabs or something

PR #3879, opened 2020-01-09
Remove "beta" label from Dark Theme slider switch text

wiz:


core/src/main/resources/i18n/displayStrings_el.properties
```diff
@@ -917,7 +917,7 @@ setting.preferences.addAltcoin=Πρόσθεσε altcoin
 setting.preferences.displayOptions=Προβολή επιλογών
 setting.preferences.showOwnOffers=Εμφάνισε τις προσφορές μου στο καθολικό προσφορών
 setting.preferences.useAnimations=Χρήση animations
-setting.preferences.useDarkMode=Χρήση σκοτεινής κατάστασης (beta)
+setting.preferences.useDarkMode=Χρήση σκοτεινής κατάστασης
```
sqrrm, line 920: There is no need to change any of the other displaystring files except displayStrings.properties as the others will be overwritten from the translation repo before release.

sqrrm (approved):
utACK

Issue #3753, opened 2019-12-05
[Feature] Improvements to BSQ supply charts and figures


dmos62:
This is a suggestion to make some improvements to the BSQ supply charts and figures that are found under DAO > Facts & Figures > BSQ Supply. For a quick and cheap introduction, at the bottom of the post there are two screenshots with captions.

Unresolved issue threads that talk about improvements to BSQ facts & figures panels: largely talk about how BSQ burn statistics are presented:
- [Feature] Provide a way to show burned BSQ in a specific timeframe #3303
- Improve burned BSQ chart #2990
- Show BSQ trade fee in % per day #3292

I decided to open a separate thread, because this proposal has a slightly larger scope than either of those.

### Suggestions
- use stacked area charts, because both charts in the "BSQ Supply" panel are already presented as sum totals of a few series (e.g. series in first chart are BSQ issued at genesis tx + BSQ issued for compensation + BSQ issued for reimbursement);
- in the "BSQ issued" chart increase detail with shorter intervals (currently the interval is 1 month);
- offer charts and figures for multiple time periods (more on that below);
- find a solution for outlier resilience (more on that below);
- consider using moving averages to improve chart readability (more on that below).

The last three suggestions are less clear-cut or obvious, so details below.

#### Multiple time periods

That's multiple time periods in the most general sense. It can either be the ability to select from a small set of time periods (last day, last 12 months, etc., like in Market > Trades) or it could be a way to specify a custom period (like a slider). I see two motivations for multiple time periods:
1. comfortabily examining the series on different timeframes; and,
2. seeing the aggregates (e.g. total BSQ issued) for different timeframes.

I think the optimal user experience and development effort would come from the simplest solution that works. So if a set of 3-4 timeperiods is enough (like in Market > Trades), that's preferable to a slider or some other "more interactive" solution.

#### Outlier resilience

Recently an outlier (point outside multiple standard deviations) appeared on the BSQ burnt chart (as seen in the screenshot at the bottom of the post), rendering the rest of the chart unreadably flat. Ways to offer outlier resilience:
- automatic explicit visual trimming of outliers (outliers drawn outside the chart, and thus invisible and not distorting);
- optional switch to logarithmic scale;
- ability to select which portion of y-scale to zoom in on;

I think automatic explicit visual trimming of outliers is by far the most user-friendly method and the easiest to implement.

#### Moving averages

[...]

PR #3890, opened 2020-01-11
Improve readability of the daily burnt BSQ chart

dmos62:
Relevant issue: #3753 

These commits are about improving readability of the daily burnt BSQ chart. It is done by introducing a 'Zoom to inliers' toggle, which when active zooms in on inliers, and by introducing a moving average. The zooming feature is automatic and should be at least somewhat robust to new trends. The moving average contextualizes the large peaks and gullies and generally improves interpretability of this noisy series. Also, the charts' Y axes are set to include 0, most notably in the monthly issued BSQ chart.

I have provided details in the commit messages.

The numerous changes to the SupplyView class are contained in a single commit, which I realise is not comfortable to review. Ideally I would have refactored in a separate commit, but I wasn't sure how I want to refactor until many of the changes were already implemented.

Below are screenshots illustrating changes to the daily burnt BSQ chart:

This is before changes:
bsq burnt before changes
This is a GIF slideshow (2 frames) of the zoom toggle being toggled in the new chart:
bsq burnt after changes animation

This took much longer to implement than expected. There were quite a few nuances I didn't anticipate.

Furthermore, some of the more minute cosmetic choices I made (or didn't make) might not be great, because since the holidays (and for a few weeks more) I'm using a pretty old machine, which makes the ROI of iterating less significant GUI changes prohibitvely low.

### DIFF
--- a/common/src/main/java/bisq/common/util/DoubleSummaryStatisticsWithStdDev.java
+++ b/common/src/main/java/bisq/common/util/DoubleSummaryStatisticsWithStdDev.java
@@ -1,0 +1,84 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.common.util;
+
+import java.util.DoubleSummaryStatistics;
+
+/* Adds logic to DoubleSummaryStatistics for keeping track of sum of squares
+ * and computing population variance and population standard deviation.
+ * Kahan summation algorithm (for `getSumOfSquares`) sourced from the DoubleSummaryStatistics class.
+ * Incremental variance algorithm sourced from https://math.stackexchange.com/a/1379804/316756
+ */
+public class DoubleSummaryStatisticsWithStdDev extends DoubleSummaryStatistics {
+    private double sumOfSquares;
+    private double sumOfSquaresCompensation; // Low order bits of sum of squares
+    private double simpleSumOfSquares; // Used to compute right sum of squares for non-finite inputs
+
+    @Override
+    public void accept(double value) {
+        super.accept(value);
+        double valueSquared = value * value;
+        simpleSumOfSquares += valueSquared;
+        sumOfSquaresWithCompensation(valueSquared);
+    }
+
+    public void combine(DoubleSummaryStatisticsWithStdDev other) {
+        super.combine(other);
+        simpleSumOfSquares += other.simpleSumOfSquares;
+        sumOfSquaresWithCompensation(other.sumOfSquares);
+        sumOfSquaresWithCompensation(other.sumOfSquaresCompensation);
+    }
+
+    /* Incorporate a new squared double value using Kahan summation /
+     * compensated summation.
+     */
+    private void sumOfSquaresWithCompensation(double valueSquared) {
+        double tmp = valueSquared - sumOfSquaresCompensation;
+        double velvel = sumOfSquares + tmp; // Little wolf of rounding error
+        sumOfSquaresCompensation = (velvel - sumOfSquares) - tmp;
+        sumOfSquares = velvel;
+    }
+
+    private final double getSumOfSquares() {
+        // Better error bounds to add both terms as the final sum of squares
+        double tmp = sumOfSquares + sumOfSquaresCompensation;
+        if (Double.isNaN(tmp) && Double.isInfinite(simpleSumOfSquares))
+            // If the compensated sum of squares is spuriously NaN from
+            // accumulating one or more same-signed infinite values,
+            // return the correctly-signed infinity stored in
+            // simpleSumOfSquares.
+            return simpleSumOfSquares;
+        else
+            return tmp;
+    }
+
+    private final double getVariance() {
+        double sumOfSquares = getSumOfSquares();
+        long count = getCount();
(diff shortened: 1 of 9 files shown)


common/src/main/java/bisq/common/util/DoubleSummaryStatisticsWithStdDev.java
```diff
@@ -0,0 +1,87 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.common.util;
+
+import bisq.common.util.Tuple2;
```
ripcurlx: Not used in code

common/src/main/java/bisq/common/util/DoubleSummaryStatisticsWithStdDev.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.util.DoubleSummaryStatistics;
+import bisq.common.util.DoubleSummaryStatisticsWithStdDev;
```
ripcurlx: Not used in code

desktop/src/main/java/bisq/desktop/main/dao/economy/supply/SupplyView.java
```diff
@@ -63,16 +70,26 @@
...
+
+import java.util.function.BiFunction;
+import java.util.function.Supplier;
+import java.util.function.Consumer;
+
+import javafx.collections.ListChangeListener;
 
 import static bisq.desktop.util.FormBuilder.addTitledGroupBg;
 import static bisq.desktop.util.FormBuilder.addTopLabelReadOnlyTextField;
+import static bisq.desktop.util.FormBuilder.addSlideToggleButton;
```
ripcurlx: Do you use IntelliJ for development? If yes, please use the formatting functionality that auto sorts and groups the different imports.

ripcurlx (changes requested):
NACK - Running `reformat code` and `optimize imports` in IntelliJ changes quite a lot in your files. Please adapt our code style.

Just for functionality your code does what it is expected to do. Could you please adapt the style of the line chart so the BSQ burnt has the same line thickness as the chart above. Atm it looks still a little bit developer style 😉 . 
Bildschirmfoto 2020-01-16 um 16 30 56

Thanks for this contribution 👍

desktop/src/main/java/bisq/desktop/main/dao/economy/supply/SupplyView.java
dmos62: I use vim and I do formatting and similar tasks manually. I'll ask for suggestions on #dev.

ripcurlx:
@dmos I activated yesterday codacy's PR Quality Review for the Bisq repository. It found for your PR a couple of issues which you can view in the Details link next to the PR check. Please review and push fixes for this issues to trigger a re-review by codacy. Thanks!

sqrrm:
Considering the weekly activity by the burning man it's known that every 7 days there will be a spike making the moving average a bit pointless. I suggest to not include the days with extreme outliers, which will be the burning days. The activity by the burning man is not that relevant as it's going to be correlating to the amount of bugs and failed trades.

Another thought could be to also add the moving average over the outliers to get a picture of how much is burnt due to failed trades, and thus show two moving averages.

dmos62:
@sqrrm if I understand correctly, your suggestion is to treat the very big spikes, like the ones on the right of the chart, as if it was a different data series. I.e. present it and the rest of daily burnt BSQ separately and analyse and reason about it separately as well.

I'm not sure I understand what the burning man is (I only know Burning Man, haha). I'll just suppose that it's something related to the 7 day cycle you mention.

So, if we were to do that, it would be much preferable to perform the "separation" earlier in the pipeline, ideally mirroring how it works with the issued BSQ series in `daoStateService.getIssuanceSet(issuanceType)`, `issuanceType` being either `REIMBURSEMENT` or `COMPENSATION`.

That said, as I see it, both these processes that we possibly would want to reason about separately are highly related in that they're burning BSQ, so being able analysing their sum total effect is desirable. The question would be if we want to also analyse them in isolation of each other.

As for the moving average, if we do want to analyse burning BSQ as a whole, it implies analysing what influence the inliers and outliers have on their total effect, and for that MA is important. It not only improves the readability of the noisy inliers, but it translates outliers and inliers to a "common denominator" for the lack of a better term.

sqrrm:
Yes, I consider them different data sets for most of our use cases. The problem is of course to do a proper separation, but we should in theory know exactly which coins were burnt by the burning man as this role has to report exactly what was burnt.

I don't know if it's possible to find this through the DAO, but I suspect not as there would be a hash in the opreturn but I don't think you have enough data to calculate that hash.

My suggestion was just the easiest solution to at least get some kind of usable moving average. I wouldn't spend too much time on this, it's probably more useful to manipulate this data outside of the app for any kind of analysis.

dmos62:
The more I think about it, the better I like the idea of tagging the inliers and outliers as explicit subseries within burnt BSQ. That would probably require tagging them as such on burning; I haven't looked at how issued BSQ is subdivided. If we had thought of this before, the automatic computation of the inlier range would not have been necessary.

Edit: oh actually, we probably can't tag retroactively, so something like this would still be necessary for analysing data accumulated until now.

This is all good food for thought. I wonder how the new priority based compensations will influence work on this.

This PR contains functionality to separate the inliers and outliers, so we have what we need for separating the series, and in turn having separate MAs, but doing it so far downstream has a hacky feel. Though this method would stop working if the clear distinction between the series disappeared in the future.

Could you describe the use cases you have in mind in the linked issue thread? That would be very useful.

dmos62:
Did a forced push.

Organized imports and reformated code with IntelliJ. Made CSS changes to the chart. Namely, the plot line width was reduced to 1px, and the plot line colors were changed to mirror more closely the perceived color of the stacked area chart above and, at the same time, they were also made to adapt to the different backgrounds under light and dark modes.

bsq burnt after style changes

ripcurlx:
@dmos62 I just pushed some minor code changes to get rid of as many IntelliJ warnings as possible. Please let me know if you are fine with those changes.

dmos62:
@ripcurlx Yes, I'm fine with the changes. Thanks for the improvements.

ripcurlx (approved):

Issue #3247, opened 2019-09-11
Incorrect account age used for trade amount limititation


littlebit2big:
I hope I'm in the right place. I tried in Slack and Bisq forum first but keep getting referred here. 

I'm getting an error message regarding trade amount limitation based on <2 months old which contradicts the actual age of my account which is displayed in the same message box.

v1.1.5

![Trade Limit and Account Age Error](https://user-images.githubusercontent.com/49798544/64723208-0d9fab80-d49e-11e9-99a0-ce615ecd7895.png)

wiz:
What payment method is this?

niyid:
Is the 2 month restriction period a blanket rule or does the period vary depending on the payment account?

I also noticed the extra "days" already handled by formatted account age in resource bundle key: **`offerbook.warning.tradeLimitNotMatching`**.

Who is assigning?

tc1176:
Same problem: https://www.reddit.com/r/bisq/comments/e5s86s/huh/

v1.2.3

sqrrm:
@Sethadon Since March of 2019 the limit for buying using most fiat methods has been 0.01 BTC. This change hasn't been reflected in all the popups and warnings.

With the 1.2.0 release there is a signing mechanism to raise the limits for buying BTC with fiat. After doing a successful buy trade with a counterparty that's a signer your account will get signed and 30 days later the limits will increase.

@littlebit2big For this particular issue it looks a bit strange if this is a fiat trade. Maximum for fiat trades is 0.25 BTC so the 0.49 BTC indicate it's probably an altcoin trade. The warning looks strange, that account should have a 2 BTC limit.

rafaelpac:
PR #3847 fixed that, this can be closed.

rafaelpac:
> 
> 
> Is the 2 month restriction period a blanket rule or does the period vary depending on the payment account?
> 
> I also noticed the extra "days" already handled by formatted account age in resource bundle key: **`offerbook.warning.tradeLimitNotMatching`**.
> 
> Who is assigning?

This was fixed by #3804, but now the message is all changed to reflect the account signing protocol.

PR #3897, opened 2020-01-13
Fix popup message about trade limits

### LINKED ISSUE #3885: Wrong popup message about trade limits
### Description

When a user tries to take an offer which is above his trade limits, the popup message says that he needs to wait 2 months.

But it is not the case anymore, since account signing was implemented.

image

#### Version

1.2.5

### Steps to reproduce

Create an unsigned account and try to take an offer above its limit.

### Expected behaviour

The message should say whether the account is signed or not and how long ago it happened.
There is a similar generic message in popup.warning.tradeLimitDueAccountAgeRestriction.buyer
But I think we could make it better, showing the user data.

Something like:
The payment method for this offer is considered risky for bank chargebacks.
Your allowed trade amount is limited because of these security restrictions.
With this version of the software, trading peers can verify and sign each others' payment accounts to create a network of trusted payment accounts.
After successfully trading with a peer with a verified payment account, your payment account will be signed and trading limits will be lifted after 30 days.
For more information on account signing, please see the documentation at https://docs.bisq.network/payment-methods#account-signing.

`Your account has not been signed.`
[or]
`Your account was signed XXX days ago (your limit will be lifted in (30-XXX) days).`

Your trade limit is not sufficient for that offer. You cannot take that offer at the moment.
`Your trade limit is: XXX.`
`The minimum trade amount of the offer is: XXX.`

### Actual behaviour

See message on the screenshot above.

#### Additional info

This could be easily fixed with minor changes to the strings file or the offerbookview file.
But I think that a message with more data like above would be better, which is more complex than my limited java abilities allow me to fix.

rafaelpac:
When a user tries to take an offer which is above his limit, the old popup is still being shown.
It says that the user limits will be raised when his account is 2 months old.
But that is not true anymore, now we have account signing etc...
This fix uses the string that is already used for offer creation, which is good since preserves translations.
But I think we would need a better message, which I am afraid I cannot get done, showing the sign state and age.
I am submitting this PR as is now since I think this is a big issue.
New users might try to use Bisq to accept an offer, get this message and then decide to wait 2 months, in vain.
So, it partially fixes #3885.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBookView.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBookView.java
@@ -580,10 +580,9 @@ public class OfferBookView extends ActivatableViewAndModel<GridPane, OfferBookVi
                 final long tradeLimit = model.accountAgeWitnessService.getMyTradeLimit(account.get(),
                         offer.getCurrencyCode(), offer.getMirroredDirection());
                 new Popup()
-                        .warning(Res.get("offerbook.warning.tradeLimitNotMatching",
-                                DisplayUtils.formatAccountAge(model.accountAgeWitnessService.getMyAccountAge(account.get().getPaymentAccountPayload())),
+                        .warning(Res.get("popup.warning.tradeLimitDueAccountAgeRestriction.buyer",
                                 formatter.formatCoinWithCode(Coin.valueOf(tradeLimit)),
-                                formatter.formatCoinWithCode(offer.getMinAmount())))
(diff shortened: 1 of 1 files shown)


ripcurlx:
@rafaelpac Thanks for catching this issue. I think it is good as suggested by you. But yes, both for offer creation and offer taking we could improve the popup taking the current account state into account and making it more tailored to the user account.

ripcurlx (approved):
ACK

Tested it locally on Regtest and it works now as expected.

rafaelpac:
Fixes #3247

PR #3904, opened 2020-01-16
Remove btcnodes not operated by Bisq btcnode team members

wiz:
### Background
Due to a change starting from Bitcoin 0.19.x, a non-default `bitcoin.conf` configuration is now required for usage with Bisq nodes: `peerbloomfilters=1` must be set.

### Summary
Because of the change in Bitcoin, a Bisq node can no longer use a Bitcoin node with default settings. Therefore, Bisq must ensure that all of its hard-coded Bitcoin nodes have the `peerbloomfilters=1` setting enabled.

Currently we have 2 hard-coded Bitcoin nodes whose operators are not members of the #btcnodes team/channel on Keybase or Github, and do not make regular reports to the Bisq DAO. Additionally, we have seen availability issues with these 2 btcnodes in the past, with no way to communicate or coordinate with these btcnode operators.

Because of the above concerns, this PR removes these 2 btcnodes. If the btcnode operators confirm they fulfill the above requirements, they can request to be re-added.

### DIFF
--- a/core/src/main/java/bisq/core/btc/nodes/BtcNodes.java
+++ b/core/src/main/java/bisq/core/btc/nodes/BtcNodes.java
@@ -70,18 +70,12 @@ public class BtcNodes {
                         // Devin Bileck
                         new BtcNode("btc1.dnsalias.net", "lva54pnbq2nsmjyr.onion", "165.227.34.198", BtcNode.DEFAULT_PORT, "@devinbileck"),
 
-                        // sgeisler - temporarily disabled due to recent uptime falling below acceptable threshold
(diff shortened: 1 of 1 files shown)


ripcurlx:
@jhoenicke @sgeisler I'm tagging you so you can give feedback on this PR before it is merged into master. I'll keep it open for a couple of days to give time to response.

ripcurlx:
Merging this PR because of no response by the node operators (@jhoenicke @sgeisler) for 6 days.

ripcurlx (approved):
utACK

PR #3907, opened 2020-01-18
List Liquid Bitcoin (L-BTC)

wiz:
### Asset name

Liquid Bitcoin

### Ticker

L-BTC

### Website

https://liquid.net

### Block explorer

https://blockstream.info/liquid/

### Additional technical requirements

Due to base layer privacy aka Confidential Transactions, the Bisq mediators / refund agents must obtain the Blinding Key from the L-BTC sender to verify transactions for any trade dispute, similar to Monero. 

While the [full documentation](https://docs.blockstream.com/liquid/developer-guide/confidential-transactions.html?highlight=blinding%20key) explains all the details, I will make a separate PR to docs repo for use by support agents, mediators, and refund agents with simple steps to verify trade funds were sent.

### DIFF
--- a/assets/src/main/java/bisq/asset/coins/LiquidBitcoin.java
+++ b/assets/src/main/java/bisq/asset/coins/LiquidBitcoin.java
@@ -17,9 +17,11 @@
 
 package bisq.asset.coins;
 
+import bisq.asset.AltCoinAccountDisclaimer;
 import bisq.asset.Coin;
 import bisq.asset.RegexAddressValidator;
 
+@AltCoinAccountDisclaimer("account.altcoin.popup.liquidbitcoin.msg")
 public class LiquidBitcoin extends Coin {
 
     public LiquidBitcoin() {
--- a/assets/src/main/java/bisq/asset/coins/LiquidBitcoin.java
+++ b/assets/src/main/java/bisq/asset/coins/LiquidBitcoin.java
@@ -1,0 +1,28 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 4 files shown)


ripcurlx:
> While the [full documentation](https://docs.blockstream.com/liquid/developer-guide/confidential-transactions.html?highlight=blinding%20key) explains all the details, I will make a separate PR to docs repo for use by support agents, mediators, and refund agents with simple steps to verify trade funds were sent.

@wiz I think it would be good to have this already before merging as we need to check the efforts with mediators and refund agents if it is ok with them as well. Just in case we get lots of mediation/refund agent cases this might cause lots of workload for verifying transactions.

wiz:
@ripcurlx When we listed other crypto assets with base layer privacy like Monero, Zcash, etc. did you require a separate bisq-docs contribution to document verification steps for mediators, or did they simply refer to the generic documentation for verification? If the former, could you point me to where this is currently documented so I can use that as a reference example for my bisq-docs PR?

m52go:
Since these docs are probably a better fit for the knowledge base, and since the knowledge base should be going live in the next 1-2 weeks (before v1.2.6 release), I think it's better if we wait to specify these verification steps there. 

I don't think there's an obvious place for them in the current documentation repository.

wiz:
After further discussion with @ripcurlx, I added the following popup dialog when adding an L-BTC payment account

Screen Shot 2020-01-22 at 4 03 55

wiz:
After further discussion with @ripcurlx, I added the following popup dialog message that will be displayed to users when configuring their L-BTC payment account

Screen Shot 2020-01-23 at 1 42 30

wiz:
After further discussion with @ripcurlx, I added the following popup dialog message that will be displayed to users when adding the L-BTC payment account

Screen Shot 2020-01-23 at 4 18 43

ripcurlx:
I'll merge this PR now after I've talked to mediators and refund agent about payment verification procedure.

ripcurlx (approved):
utACK

wiz:
test

ripcurlx:
~As I haven't ACKed asset PRs before I completely missed that the test for the asset was missing. So I wanted to create one quickly myself but I'm not able to validate addresses with the current validator implementation (e.g. [da369298d9bb312b636abd260ec1df31bfc513b2b059500879b3e95cad3b7519](https://blockstream.info/liquid/tx/da369298d9bb312b636abd260ec1df31bfc513b2b059500879b3e95cad3b7519?output:1))  should be a valid address. But when testing it the validator rejects this address. Do I understand something wrong how to get a valid L-BTC address?~ I was not looking properly and copied the tx id instead of the address. Still we found out that we only support non native segwit addresses, why we changed the error message.

wiz:
We can revise the popup message after https://github.com/mempool/mempool/issues/523 is implemented

PR #3909, opened 2020-01-19
Fix potential resource leak in AvoidStandbyModeService

stejbac:
Replace tail recursion of the `AvoidStandbyModeService.play` method with an ordinary loop, to prevent a new open JAR resource InputStream + sound file OutputStream (which were created every 4 minute playback) from accumulating on the stack, closing them inside the loop instead. (This also prevents eventual stack overflow.)

Also tidy up `FileUtil.resourceToFile` and put the JAR URL InputStream in a try-with-resources block, to ensure that it doesn't leak either.

--

Running _lsof_ does reveal an accumulation of file handles without this fix, although not as many as one would expect from a new stream being opened every 4 minutes - perhaps they were being closed somehow in a finaliser?

Screenshot from 2020-01-17 00-14-05

Also, JProfiler revealed an accumulation of `ZipInputStream` objects from loading the _.aiff_ JAR resource.

### DIFF
--- a/common/src/main/java/bisq/common/storage/FileUtil.java
+++ b/common/src/main/java/bisq/common/storage/FileUtil.java
@@ -22,6 +22,7 @@ import bisq.common.util.Utilities;
 import com.google.common.io.Files;
 
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.IOUtils;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -32,6 +33,7 @@ import java.io.IOException;
 import java.io.InputStream;
 
 import java.util.Arrays;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


ripcurlx (dismissed):
ACK - Tested it locally on Regtest. Toggling the avoid standy by feature (I can hear the sound behaving as before) everything works as before from a user perspective. Thanks for looking into this 👍

I just don't merge it immediately as I want to wait until https://github.com/bisq-network/bisq/pull/3889 gets merged so @cbeams doesn't need to do any unnecessary merges.

ripcurlx:
@stejbac Could you please resolve the conflict, caused by the recent gRPC PR merge?

ripcurlx:
Afterwards I'm happy to merge your PR

stejbac:
I've rebased onto master. (The conflict in `AvoidStandbyModeService` was caused by imports and a move from `BisqEnvironment.getAppDataDir()` to `config.appDataDir`.)

ripcurlx (approved):
ACK - https://github.com/bisq-network/bisq/pull/3909#pullrequestreview-345276132

PR #3912, opened 2020-01-22
Reduce codeowners to the required minimum

ripcurlx:
This is done to deactivate auto assignment of reviewers in most cases and should encourage everyone to start reviewing PRs on their own.


sqrrm (approved):
utACK

PR #3914, opened 2020-01-22
Prevent memory leak in CandleStickChart & VolumeChart

stejbac:
Ensure that the superclass methods `XYChart.removeDataItemFromDisplay` and `XYChart.removeSeriesFromDisplay` are always called from the implemented `dataItemRemoved` and `seriesRemoved` methods respectively, as specified by the API javadoc.

This prevents a leak of old `Candle` and `VolumeBar` objects every time the trades charts view is updated. The former is quite substantial, as each Candle object has a retained size of about 70kB and there are up to 90 candlesticks / volume bars leaked per chart update.

--

By cycling through the _Month_, _Week_ and _Day_ interval a few dozen times, the following heap usage stats were obtained using JProfiler:

Screenshot from 2020-01-22 10-42-19

I believe updates to the two charts happen periodically as long as the trades charts view is selected, as it appeared to leak while unattended as well. I haven't had a chance to measure the rate, though.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/market/trades/charts/price/CandleStickChart.java
+++ b/desktop/src/main/java/bisq/desktop/main/market/trades/charts/price/CandleStickChart.java
@@ -70,9 +70,6 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 /**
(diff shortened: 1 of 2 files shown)


ripcurlx (approved):
ACK

Tested PR against Mainnet switching between time frames and currencies and everything seems to be still working as before from a user perspective. Code changes looking fine to me as well, but I didn't verified the change using jProfiler. If anyone wants to review/verify this part as well feel free to do so.

PR #3922, opened 2020-01-29
Trade process refresh

### LINKED ISSUE #3905: Seller unable to confirm payment received
### Description
Many mediation cases lately are a result of seller not being able to confirm the payment has been received. On most cases the buyer confirms they have pressed ‘payment made’ but for some reason is not showing on the other end.

### How to fix
It's hard to reproduce and might be a result of a missed mailbox message. A simple option ask for a resend from the seller side might be enough. Thinking a bit further it would be nice to be able to confirm transmission failures of these messages. Not sure how that could be done though.

I can't think of any other reason that the seller wouldn't see the confirm payment button than a failed message.

sqrrm:
Fix #3905 

A seller can ask to refresh the trade process once every 24 hours. This step has been a problem causing a lot of mediation lately so this is a way to ask the buyer to resend the CounterCurrencyTransferStartedMessage

This fixes the problem when a mailbox message was lost. To test the seller need to not get the first CounterCurrencyTransferStartedMessage sent by the buyer, for example by letting the seednode drop it instead of sending to the seller. When button is pressed
- a RefreshTradeStateRequest is sent from seller to buyer
- the buyer receives the RefreshTradeStateRequest and
  - ignores it if it's not in FIAT_SENT phase
  - responds with a CounterCurrencyTransferStartedMessage if in FIAT_SENT phase and has already sent a CounterCurrencyTransferStartedMessage

### DIFF
--- a/common/src/main/proto/pb.proto
+++ b/common/src/main/proto/pb.proto
@@ -75,6 +75,8 @@ message NetworkEnvelope {
         DelayedPayoutTxSignatureResponse delayed_payout_tx_signature_response = 47;
         DepositTxAndDelayedPayoutTxMessage deposit_tx_and_delayed_payout_tx_message = 48;
         PeerPublishedDelayedPayoutTxMessage peer_published_delayed_payout_tx_message = 49;
+
+        RefreshTradeStateRequest refresh_trade_state_request = 50;
     }
 }
 
@@ -319,6 +321,12 @@ message MediatedPayoutTxSignatureMessage {
     string trade_id = 3;
     bytes tx_signature = 2;
     NodeAddress sender_node_address = 4;
+}
+
+message RefreshTradeStateRequest {
+    string uid = 1;
+    string trade_id = 2;
+    NodeAddress sender_node_address = 3;
 }
 
 // dispute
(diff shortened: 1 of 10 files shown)


ripcurlx:
I tested it locally on Regtest and everything seems to work as expected (changed `REFRESH_INTERVAL`for testing purpose). I tested it with both clients on PR and backwards comp test with v1.2.3 client where it fails silently if not supported.
**Before refresh trade state is enabled**
Bildschirmfoto 2020-02-03 um 11 22 22
**After refresh trade state is enabled**
Bildschirmfoto 2020-02-03 um 11 21 54

But I fear we need some extra text next to the button to explain its functionality. @m52go  Do you have a good and compact wording for it?

ripcurlx:
And as you just suggested in chat, probably don't show the inactive button if not needed. That way it sticks more out if it is displayed at some point.

m52go:
Suggestion:

_Sometimes P2P network messages acknowledging payment are not delivered, causing trades to get stuck. Hit the button below to resend the last message._

sqrrm:
Added text. Hide refresh section when not available

image
image

ripcurlx (approved):
ACK

Works as expected on Regtest. This should reduce our mediation/refund cases a lot (hopefully)!

sqrrm:
I'm not sure this is the only step that's losing messages. I didn't want to try to implement a generic refresh option as it's not clear in every step that it's safe to refresh.

Another option might be to make the mailbox messages more robust. I have a feeling it should be possible but I'd have to take a deeper look at how they work.

MwithM:
Why is it possible to referesh only every 24 hours? It's too long for fast payment methods like altcoins or sepa instant.

sqrrm:
It's a good point that it should probably be proportional to the trade duration.

PR #3932, opened 2020-02-02
Remove unused i18n strings

### LINKED ISSUE #3860: Remove unused i18n strings
It seems like this string is no longer referenced from the source code, so we should remove it (see the link below). It creates confusion for translators who think that it should be translated. We should probably review the strings to check that there are no unused ones in general to save unnecessary work for translators.

```
portfolio.pending.support.text.getHelp.arbitrator=If you have any problems you can try to contact the trade peer in the trade \
  chat or ask the Bisq community at https://bisq.community. \
  If your issue still isn't resolved, you can request more help from an arbitrator.
```

https://github.com/bisq-network/bisq/blob/567684172d49a89c90119f41a329c0170930e111/core/src/main/resources/i18n/displayStrings.properties#L793

niyid:
Fixes  #3860

To remove the unused key entries in the **`displayStrings.properties`** file, a JUnit Test class called **`bisq.core.util.DisplayPropertiesCheckTest`** was created which generates a new copy of the stripped file to **`displayStrings.pruned.properties`**. It also uses the DisplayPropertiesCheckTest.ignorelist file to provide a list (one per line) of regular expressions matching property keys to ignore.


ripcurlx (changes requested):
NACK

@niyid Thanks for looking into this, but your code needs to take care not to remove dynamically created keys in code as well. Atm it would remove used keys at well:
- dao.param.*
- dao.phase.*
- dao.bond.bondState.*
- dao.bond.lockupReason.*
- dao.bond.bondedRoleType.*
- dao.assetState.*
- dao.proposal.type.*
- dao.tx.type.enum.*
- *_SHORT
- NOT remove # comment lines
have to be excluded as they are used in code.

Also I don't think we should put this into the source directory. Normally this kind of tools are put into the test directory so it is not included in any source build process.

niyid:
- Converted to JUnit test class.
- Added ignore-list for dynamically created keys.
- Comment lines were **never** removed so no change required for this.

ripcurlx:
`message.state.*` has to be excluded as well.

`dao.phase.*` keys are still removed from the properties file by mistake.

Please remove the pruned properties file afterwards from this PR as I'll run it myself, verify and overwrite the English source file myself.

There is also an issue mentioned in codacy that needs to be fixed. Also please have a look if you have formatted everything correctly before requesting my review.

niyid:
@ripcurlx Please go ahead with your review. Regards.

ripcurlx (changes requested):
NACK

Running `testPrune` does seem to print correctly `UNUSED:...` properties, but it creates a `displayStrings.pruned.properties` file that is identically to `displayStrings.properties`

niyid:
> NACK
> 
> Running `testPrune` does seem to print correctly `UNUSED:...` properties, but it creates a `displayStrings.pruned.properties` file that is identically to `displayStrings.properties`

Apologies. Fixed. Please do not forget to remove the "waiting for author" label. Regards.

PR #3933, opened 2020-02-03
Accept old trade statistic object

ripcurlx:
The code didn't handle before the use case of new trade statistic objects
created by two old clients. This change make it independent of the cut off date
and allows us at a later point to update all trade statistics objects with
depositTxId value of null.

### DIFF
--- a/core/src/main/java/bisq/core/trade/statistics/TradeStatistics2.java
+++ b/core/src/main/java/bisq/core/trade/statistics/TradeStatistics2.java
@@ -53,6 +53,8 @@ import java.util.Optional;
 import lombok.Value;
 import lombok.extern.slf4j.Slf4j;
 
+import org.jetbrains.annotations.NotNull;
+
 import javax.annotation.Nullable;
 
 import static com.google.common.base.Preconditions.checkNotNull;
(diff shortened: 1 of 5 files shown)


ripcurlx:
Just to make the different use cases more clear:

1) Old clients
As they still have their local trade statistics they will still see the trade statistics.
- If a new trade is added and both parties are on the old version the trade will show up in the trade statistics for the clients.
- If a new trade is added and the client receives the trade statistics object from the seednode (or any other v1.2.6 node) it won't be shown in the trade statistics as it is seen as invalid trade statistics object (trade deposit tx id is null).

2) New clients v1.2.6+
Nothing changes for the new clients. They will see trades created by old clients and trades by new clients.

Seednodes:
Starting with v1.2.6 they will load the trade statistic object in memory and set the deposit tx id of the trade statistics objects to null (also for all old trades). We should update the seednodes first before releasing, so they are not missing out any trades by new clients which would be seen as invalid by the old seednodes.

ripcurlx:
As it doesn't give more privacy to set deposit transaction ids of old trade statistics to null (can be retrieved by data stores of old client binaries) and only would cause additional load for our seednodes I've reverted this change. Still as old clients will parse null to an empty string causes them to load trades by new clients (v1.2.6) to reload on restart as empty string != null.

core/src/main/java/bisq/core/trade/statistics/TradeStatistics2.java
```diff
@@ -217,7 +218,7 @@ public static TradeStatistics2 fromProto(protobuf.TradeStatistics2 proto) {
                 proto.getTradePrice(),
                 proto.getTradeAmount(),
                 proto.getTradeDate(),
-                null, // We don't want to expose this anymore
+                ProtoUtil.stringOrNullFromProto(proto.getDepositTxId()),
```
sqrrm, line 221: This would leave all new trades from older clients still exposed. That's perhaps the safest way to do it. Was hoping it would be possible to keep setting it to null for new version and just not serve the old version anything. Not sure if that would break things though.

ripcurlx: Yes, you only get the additionally privacy if you update or if you trade with a peer that has already updated. I had to do it like that to prevent the unnecessary load from old clients.

sqrrm (approved):
ACK

Works as planned and the deposittx is of course exposed as long as users don't upgrade. Nothing can be done about that.

sqrrm:
I verified that the old version did cause delivery of a lot of tradestats objects and this version doesn't.

PR #3934, opened 2020-02-03
Finalize multisig payout even on missing address entry

sqrrm:
The check for an address entry before finalizing the multisig payout
might prevent a valid payout when the initiation of the trade process
was interrupted. It's better to allow the completion of the trade
and just log a warning.


ripcurlx (approved):
utACK

Issue #3916, opened 2020-01-23
Create metric to monitor network size


freimair:
*Having this as an issue does not feel quite right, neither does having it as a proposal. After aligning with @ripcurlx, here it is anyways.*

How is our network? Well, the [monitor](https://monitor.bisq.network) provides some stats and we know a lot more about our network as we did before. However, we still do not know the size of the network (still an unchecked box in the [monitor proposal](https://github.com/bisq-network/proposals/issues/62). A recent investigation showed that a single price node encountered around 500 active clients per day which could mean that only roughly 500 * 5 price nodes = 2500 Bisq clients are active, which is a lot less than the number of downloads (around 10000) and may indicate that the active Bisq community is much smaller as we though it is. Plus, we do not know about the version spread of Bisq clients out there.

# Information gained

Creating a metric that can deliver a rough estimation about the network size might bring us base information like
- an estimation on how many Bisq clients are online at each time (= network size)
- ~~daily active users, weekly active users~~
- beyond that, see Bisq client uptime during a 24h period
  - see if Bisq clients are always on
  - or if there is a pattern during a 24h day (clients run by day, not by night, all over the world)
- and as a bonus, we might even get version spread info of active Bisq clients

Based on that data, we can derive more information like
- how does BTC/AltCoin volume and value affect the network size
- how do marketing measures affect the network size
- how do seed node/price node/... node issues affect the network size
- how does a (broken) release affect the network size
- see the version spread and thus, have info how a "forced/breaking" update might affect the user base

Having such information drastically changes the information base on which strategies are designed and decisions made. Thus, much more informed decisions might become possible in the fields of update/release strategies, marketing measures, and, among others, growth considerations.

# Required efforts

A lot of thought has been put into this and finally, we are confident that we can do it. My estimation of implementation efforts is as follows:
- coming up with the idea and concept (be my guest)
- design how to extract the data from price nodes and sending it to the monitor (500 USD)
- deploy to every price node (150 USD per price node = 750 USD)
- configure the monitor (500 USD)

makes a total of 1750 USD. However, getting price nodes up and connected one by one might take more than one cycle.

# Technical implementation details

- use price node log scraping
  - every client connects to a pricenode a known couple of times per timespan
  - the client reveals its version (this info is not used until now)
  - the client does not reveal its node addresses or anything else that would affect privacy
[...]

PR #3937, opened 2020-02-04
Remove unused uid in pricefeed service

freimair:
This came up during work on https://github.com/bisq-network/bisq/issues/3916.

Bisq frequently (once per minute) queries our price nodes for up-to-date
price information. It does so by HTTP GET request. However, it provided
a UID via the "User-Agent" HTTP header field. This UID has been a random
number which changed everytime Bisq got started up.

This UID has never been used. Thus, remove it.

### DIFF
--- a/core/src/main/java/bisq/core/provider/price/PriceProvider.java
+++ b/core/src/main/java/bisq/core/provider/price/PriceProvider.java
@@ -48,7 +48,7 @@ public class PriceProvider extends HttpClientProvider {
     public Tuple2<Map<String, Long>, Map<String, MarketPrice>> getAll() throws IOException {
         Map<String, MarketPrice> marketPriceMap = new HashMap<>();
         String json = httpClient.requestWithGET("getAllMarketPrices", "User-Agent", "bisq/"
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
utACK

wiz (approved):
This is great for privacy :+1:

wiz:
@freimair you missed this one https://github.com/bisq-network/bisq/blob/master/core/src/main/java/bisq/core/provider/fee/FeeProvider.java#L48

and this one https://github.com/bisq-network/bisq/blob/master/core/src/main/java/bisq/core/notifications/MobileNotificationService.java#L300

wiz:
pricenode still has UUID in logs
```
INFO  b.p.m.FeeRateController: Incoming /getFees request from: bisq/1.2.7, uid:f33c97d1-b065-48c4-bc84-eaaa680e4288
```

freimair:
I only focused on the stuff relevant to the networksize metric. I guess we can remove the uid for the getFee stuff as well. However, I will not touch the mobile notification stuff because I do not know how these things work in detail.

PR #3938, opened 2020-02-05
Temporarily deactivate public network option

ripcurlx:
This will be in place until we are able to check if a connected Bitcoin Core is configured properly for our requirements.

Bildschirmfoto 2020-02-05 um 10 32 56

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/settings/network/NetworkSettingsView.java
+++ b/desktop/src/main/java/bisq/desktop/main/settings/network/NetworkSettingsView.java
@@ -350,8 +350,10 @@ public class NetworkSettingsView extends ActivatableView<GridPane, Void> {
     }
 
     private boolean isPreventPublicBtcNetwork() {
-        return filterManager.getFilter() != null &&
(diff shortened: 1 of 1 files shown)


sqrrm (approved):
utACK

wiz (approved):
Thanks for addressing this critical issue

wiz:
It seems Bisq will fallback to using the public Bitcoin network if valid Bitcoin nodes are not configured, which is unreliable. Maybe you should fallback to the provided Bitcoin nodes instead?

cbeams:
Regarding re-enabling the public network option, even with pruning and bloom filter checks in place, see https://github.com/bisq-network/bisq/pull/3982#issuecomment-591881136.

dmos62:
To perform the checks on the local Bitcoin node, we initiate a Bitcoin connection to get information from the handshake and then we close it, just to have Bisq rebuild the connection in case the node is well configured. We might be able to perform these checks by hooking into the regular/existing connection process. In that case the checks would be free in terms of network IO. I might take this up, if there's demand.

PR #3939, opened 2020-02-05
Add additional information for a local Bitcoin Core v0.19+

ripcurlx:
Bildschirmfoto 2020-02-05 um 11 32 24

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -2661,6 +2661,7 @@ popup.securityRecommendation.msg=We would like to remind you to consider using p
 
 popup.bitcoinLocalhostNode.msg=Bisq detected a locally running Bitcoin Core node (at localhost).\n\
   Please make sure that this node is fully synced before you start Bisq and that it is not running in pruned mode.
+popup.bitcoinLocalhostNode.additionalRequirements=\n\nStarting with Bitcoin Core v0.19 you have to manually enable bloom filters by setting peerbloomfilters=1 in your configuration file.
 
 popup.shutDownInProgress.headline=Shut down in progress
 popup.shutDownInProgress.msg=Shutting down application can take a few seconds.\nPlease don't interrupt this process.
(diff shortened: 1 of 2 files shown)


core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -2661,6 +2661,7 @@ popup.securityRecommendation.msg=We would like to remind you to consider using p
 
 popup.bitcoinLocalhostNode.msg=Bisq detected a locally running Bitcoin Core node (at localhost).\n\
   Please make sure that this node is fully synced before you start Bisq and that it is not running in pruned mode.
+popup.bitcoinLocalhostNode.additionalRequirements=\n\nStarting with Bitcoin Core v0.19 you have to manually enable bloom filters by setting peerbloomfilters=1 in your configuration file.
```
wiz:
```suggestion
popup.bitcoinLocalhostNode.additionalRequirements=\n\nStarting with Bitcoin Core v0.19 you have to manually enable bloom filters by setting peerbloomfilters=1 in your bitcoin.conf configuration file.
```

wiz (approved):
utACK

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -2661,6 +2661,7 @@ popup.securityRecommendation.msg=We would like to remind you to consider using p
 
 popup.bitcoinLocalhostNode.msg=Bisq detected a locally running Bitcoin Core node (at localhost).\n\
   Please make sure that this node is fully synced before you start Bisq and that it is not running in pruned mode.
+popup.bitcoinLocalhostNode.additionalRequirements=\n\nStarting with Bitcoin Core v0.19 you have to manually enable bloom filters by setting peerbloomfilters=1 in your bitcoin.conf configuration file.
```
sqrrm, line 2664: I find it easier to read when the long lines are broken with a `\`

sqrrm (approved):
utACK

huey735:
@HiFi24Seven over at [reddit said this](https://www.reddit.com/r/bisq/comments/f0bekb/in_case_youre_having_trouble_running_bisq_with_a/):

> whitelist=bloomfilter@127.0.0.1/32 is a better way to do this, as it only enables bloomfilters on connections from your local machine. The reason this is a better approach is because bloomfilters have been found to be a security vulnerability and create an attack vector when enabled for all incoming connections, which is the reason they are now disabled by default.

I'm not sure of the merits but I thought it'd be worth mentioning it.

dmp1ce:
I connected to my node using the `.onion` address and didn't get this message. I didn't realize I needed to set `peerbloomfilters=1` and I was further confused in that I thought `peerbloomfilters=1` was the default for version `0.19.1`.

PR #3940, opened 2020-02-05
Use LocalBitcoinNode as singleton

ripcurlx:
This fixes an error when running a local Bitcoin Core for Mainnet
caused by a duplicate wallet setup


cbeams (approved):
utACK. @ripcurlx and I discussed this problem and fix offline, and it looks good to me. It was an oversight on my part that I didn't declare `LocalBitcoinNode` as a `@Singleton` in the first place.

PR #3942, opened 2020-02-06
Additional tweaks to seednode installation script for new use cases

wiz:
I'm making new one-command installers for other Bisq infrastructure which requires new configuration variables. This PR adds:
* Allows setting an external Bitcoin node not running on localhost
* Allows running a non-fullnode (for use cases such as Markets API)
* Allows enabling dump of Bisq Markets data to JSON files
* Renames the `bisq-seednode` service to `bisq` service


ripcurlx (dismissed):
utACK

sqrrm (approved):
utACK

PR #3944, opened 2020-02-06
Additional tweaks to seednode installation script for new use cases

wiz:
Replaces #3942 (sorry for the last minute change)

This script will now be able to start any Bisq node entrypoint, for use with Bisq seednode, Bisq pricenode, Bisq statsnode, etc. and adds necessary new configuration variables for these use cases:

* Renames the `bisq-seednode` service to generic `bisq` service
* Allows specifying Bisq entrypoint as a configuration setting
* Allows setting an external Bitcoin node not running on localhost
* Allows running a non-fullnode (for use cases such as Markets API)
* Allows enabling dump of Bisq Markets data to JSON files

### DIFF
--- a/seednode/README.md
+++ b/seednode/README.md
@@ -40,10 +40,10 @@ After installation, watch the Bitcoin blockchain sync progress
 sudo tail -f /bitcoin/debug.log
 ```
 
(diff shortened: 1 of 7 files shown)


ripcurlx (approved):
utACK

PR #3947, opened 2020-02-06
Fix dao bond table column label

m52go:
It seems this column label is misleading:

Screenshot from 2020-02-06 13-17-34-touched

What that column really shows is **unlock times**, i.e., the number of blocks it takes for bonds to become available once they're unlocked.

Proposed wording is consistent with the option available in the bonded reputation section.

Screenshot from 2020-02-06 13-23-44


ripcurlx (approved):
utACK

PR #3948, opened 2020-02-06
Add signing debug logs

sqrrm:
Fix warning about possible nullpointer exceptions. Better to have an
explicit check for debug purposes.

Smaller cleanups

I'm chasing a possible bug where some traders are not getting signed after a trade that looks like it should result in a signed account. These logs will help to add some more helpful info for getting .to the issue

### DIFF
--- a/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
+++ b/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
@@ -74,6 +74,8 @@ import java.util.stream.Stream;
 
 import lombok.extern.slf4j.Slf4j;
 
+import javax.annotation.Nullable;
+
 import static com.google.common.base.Preconditions.checkNotNull;
 
 @Slf4j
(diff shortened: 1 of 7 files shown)


ripcurlx:
@sqrrm Shall we add this to this release to get more information on this as soon as possible?

sqrrm:
Yes, would be good to get it in the release to start getting some hints of what's going on.

ripcurlx (approved):
ACK

Tested it on Regtest and it logged following during a test trade
```
Feb.-07 15:53:56.638 [JavaFX Application Thread] INFO  b.c.a.w.AccountAgeWitnessService: AccountSigning: 
tradeId: ZJEYLNV-7b505b13-6df6-4777-beb4-e18ee130a4cb-125
is buyer: true
buyer account age witness info: UNSIGNED Not signed yet
Optional[AccountAgeWitness{
     hash=b8d43b8509ee96bf1e8c82056fc5268ce3b645c1,
     date=2019-11-22T09:12:50.779Z
}]
seller account age witness info: ARBITRATOR signed by an arbitrator and can sign peer accounts
Optional[AccountAgeWitness{
     hash=85286fcdddb138bfebbd0abb5f87b95d04aff642,
     date=2019-11-22T09:03:30.483Z
}]
checking for sign trade: false
is myWitness signer: false
peer has signed witness: true
trade amount: 1000000
trade amount is sufficient: true
isSignWitnessTrade: false
```

PR #3951, opened 2020-02-08
Use `mkdir -p` in seednode installer when creating bitcoin/bisq homedirs

wiz:
Fixes an issue when mounting an SSD directly on `/bisq` or `/bitcoin` and the path already exists, reported by @m52go


ripcurlx (approved):
utACK

PR #3926, opened 2020-01-30
Reduce number of decimals shown for monetary fiat prices to 2

ghost:
Fixes #3449

Reduce number of decimals shown from 4 to 2 for fiat monetary values.

PR #3953, opened 2020-02-09
Fix incorrect rounding of BSQ dollar price to whole number

stejbac:
Add a new method to `DisplayUtils` to restore the old rounding behaviour of `formatVolumeWithCode` whenever a fractional volume is required. This fixes a regression caused by #3926 to remove unnecessarily displayed decimals for fiat volumes - it appears that in every case but the average dollar price on the BSQ dashboard a whole number should be shown.

Without this fix, the 30/90 day average price on the BSQ dashboard displays:
> 1 BSQ = 1 USD

(I believe the issue needs to be fixed in the 1.2.6 release branch as well.)


ripcurlx (approved):
ACK

Thanks for catching this one! I'll cherry-pick after the merge into our release branch.
Tested it on Mainnet and everything seems to be correct now.

PR #3956, opened 2020-02-10
Speed up BSQ supply view load

stejbac:
When displaying the monthly BSQ issued series, avoid repeatedly calling `DaoFacade.getBlockTime` for Issuance objects with the same chain height, as that method linearly scans the entire linked list of `DaoState` blocks, making it quite slow. Instead, memoise the mapping from chain height to block-time month, so that it is only computed once per graph point instead of once for every BSQ issuance.

Also do some minor cleanup of `SupplyView`, including a compilation error flagged incorrectly by IntelliJ, caused by the raw typed constructor reference `XYChart.Data::new`.

--

Repeatedly switching to the _BSQ Supply_ tab under _Facts & Figures_ revealed the following minor hotspot in JProfiler (fixed by the above):

Screenshot from 2020-02-10 05-55-53

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/dao/economy/supply/SupplyView.java
+++ b/desktop/src/main/java/bisq/desktop/main/dao/economy/supply/SupplyView.java
@@ -61,6 +61,7 @@ import javafx.collections.ListChangeListener;
 
 import javafx.util.StringConverter;
 
+import java.time.Instant;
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
ACK

Tested it no Mainnet and everything is working as before. Code changes are looking fine.

PR #3958, opened 2020-02-10
Revert setting of null to deposit tx id

ripcurlx:
As we saw some unexpected behavior during testing we decided to move this change to v1.2.7

Additionally we'll remove even more information from the trade statistics objects and will deploy it together with changes
in the offer objects.

### DIFF
--- a/core/src/main/java/bisq/core/trade/statistics/TradeStatisticsManager.java
+++ b/core/src/main/java/bisq/core/trade/statistics/TradeStatisticsManager.java
@@ -159,7 +159,6 @@ public class TradeStatisticsManager {
         private TradeStatistics2 tradeStatistics;
 
         public WrapperTradeStatistics2(TradeStatistics2 tradeStatistics) {
-
             this.tradeStatistics = tradeStatistics;
         }
 
--- a/core/src/main/java/bisq/core/trade/statistics/TradeStatisticsManager.java
+++ b/core/src/main/java/bisq/core/trade/statistics/TradeStatisticsManager.java
@@ -40,6 +40,7 @@ import java.io.File;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


core/src/main/java/bisq/core/trade/statistics/TradeStatisticsManager.java
```diff
@@ -150,4 +154,30 @@ private void dump() {
             jsonFileManager.writeToDisc(Utilities.objectToJson(array), "trade_statistics");
         }
     }
+
+    static class WrapperTradeStatistics2 {
+        private TradeStatistics2 tradeStatistics;
+
+        public WrapperTradeStatistics2(TradeStatistics2 tradeStatistics) {
+
```
sqrrm:
```suggestion
```

core/src/main/java/bisq/core/trade/statistics/TradeStatisticsManager.java
```diff
@@ -150,4 +154,30 @@ private void dump() {
...
+        public TradeStatistics2 unwrap() {
+            return this.tradeStatistics;
+        }
+
+        @Override
+        public boolean equals(Object obj) {
+            if (this == obj) return true;
+            if (obj == null || getClass() != obj.getClass()) return false;
+            var wrapper = (WrapperTradeStatistics2) obj;
```
sqrrm, line 173: Don't you need a check that obj is actually a `WrapperTradeStatistics2` here? Could cause an exception otherwise.

ripcurlx: Could add it to make it safer if it is reused somewhere.
Atm using it like that should not cause this case to ever happen.
```
.map(WrapperTradeStatistics2::new)
.distinct()
```

ripcurlx: But will add a check to be sure

ripcurlx: ah wait - I've added it already 😉 
```
 if (obj == null || getClass() != obj.getClass()) return false;
```

sqrrm: Oh, right, that's right, missed it.

sqrrm:
Perhaps we should not leave this one as `@Nullable` in `TradeStatistics2`
```
    @Nullable
    private final String depositTxId;
```

sqrrm (approved):
ACK

It's fine that `depositTxId` stays `@Nullable` since it's not set to null.

Tested with Alice and Bob on version 1.2.6, 1.2.5 respectively and Carol as a neutral spectator on both 1.2.5 and 1.2.6.

PR #3895, opened 2020-01-13
Validate ignored peers and BTC nodes

devinbileck:
In addition to adding input validation for ignored peers and BTC nodes, this PR also fixes #3137.

Since with input validation the user is no longer able to enter an invalid BTC host, to verify #3137 start the app with the following option:
`--btcNodes=32zzibxmqi2ybxpqyggwwuwz7a3lbvtzoloti7cxoevyvijexvgsfeid.onion:8333`

You will be presented with the following:
image

After shutting down the app, restart it without the previous option and observe the BTC config. You will see the invalid custom BTC config but it will be using the provided BTC nodes.

image

PR #3961, opened 2020-02-10
Fix address regex

devinbileck:
The original implementation of #3895 did not validate IPv6 nor FQDN addresses.


ripcurlx:
@devinbileck Could you please resolve the conflict caused by a merged PR? Thanks!

wiz (changes requested):
Thanks for fixing this so quickly. I'm testing it now but unfortunately my configuration still fails to pass your validation, I guess the following testcase is not handled by the FQDN regex:
```
foo.example.com:8333,bar.example.com:8333
```

wiz:
Also, probably unrelated to your change, but for some reason IPv6 address btcnodes work on the command line but don't work when set in the GUI - I always set my stuff from the command line but kinda surprised nobody else noticed that bug.

wiz (dismissed):
Actually will ACK for now and will submit a separate PR to fix the remaining FQDN regex case

ripcurlx:
☝️ resolved the conflict with master and force pushed it

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/3961#pullrequestreview-356553147

PR #3964, opened 2020-02-11
Fix btcnode validation regex to handle multiple FQDNs, add tests

wiz:
@devinbileck's PR #3961 fixed almost everything except one case for btcnode configuration, multiple FQDNs:
```
foo.example.com:8333,bar.example.com:8333
```

This PR fixes the FQDN regex to handle multiple cases and adds new testcases to verify


ripcurlx (approved):
utACK

PR #3965, opened 2020-02-11
Update bitcoinJ checkpoints and data stores

ripcurlx:
Light node in light mode and full node in dark mode.
Bildschirmfoto 2020-02-11 um 13 48 56
Bildschirmfoto 2020-02-11 um 13 48 52
Bildschirmfoto 2020-02-11 um 13 48 42
Bildschirmfoto 2020-02-11 um 13 48 46
Bildschirmfoto 2020-02-11 um 13 48 15
Bildschirmfoto 2020-02-11 um 13 48 35


sqrrm:
I get a difference in the TradeStatistics2Store when I run from the resources from last release. That doesn't sound right. I think yours is missing something (possibly one of the duplicates, but it should be the same running from yours and the one from previous release. Could you please check on your side again.

The other data stores look correct.

ripcurlx:
> I get a difference in the TradeStatistics2Store when I run from the resources from last release. That doesn't sound right. I think yours is missing something (possibly one of the duplicates, but it should be the same running from yours and the one from previous release. Could you please check on your side again.

You mean a difference in the number of entries? I'll debug a new live client from the old data store to check if I see different numbers in the trade statistic store compared to my v1.2.6 client

sqrrm (approved):
ACK

PR #3799, opened 2019-12-17
Add automation for new PRs and issues

ripcurlx:
This should be a first small step to automate our PR label handling

PR #3968, opened 2020-02-12
Change video link in greeting to blog link

m52go:
In case link to text post is preferred. The post has the video embedded in it.

See comments in [this thread](https://github.com/bisq-network/bisq/pull/3799#issuecomment-579700304) for context.


ripcurlx (approved):
ACK

I also saw that there is an issue with GitHub with non-repository owners so I have to change the auto PR reply system anyways

PR #3980, opened 2020-02-17
Allow 4 refreshes per trade

sqrrm:
Change refresh interval to be dependent on the trade period rather than a fix 24 hours (which is obviously not that helpful for a 1 hour trade.


ripcurlx (approved):
utACK

So we will have ~ 29h for SEPA and 12 min for a SEPA Instant trade until a refresh can be triggered

PR #3982, opened 2020-02-17
Add local Bitcoin node configuration detection

### LINKED ISSUE #3273: Show warning if user has a pruned Bitcoin node locally running
BitcoinJ requires a non-pruned Bitcoin node. We get frequently issues that users who don't know that run into problems. We should find a way to detect if a local node is running in pruned mode and if so show a popup to tell the user to either shut down that pruned node and connect to the provided bitcoin nodes or change their pruned node to a non-pruned node. 
Related is that a Bitcoin node cannot be used while it is syncing with the network. Not sure if that can be detected.

If there is no way to detect pruned mode from Bisq (we cannot use RPC for that), we could use the special exceptions thrown in BitcoinJ to catch those and react on those to show a popup.

### LINKED ISSUE #3935: Bisq doesn't connect to default Bitcoin Core version v0.19.0.1
### Description

The default Bitcoin Core version v0.19.0.1 can't connect to Bisq 

#### Version

Bisq v1.2.5 and Bitcoin Core version v0.19.0.1

### Steps to reproduce

Update to Bitcoin Core version v0.19.0.1

### Expected behaviour

Bisq should recognize that there's a local Bitcoin node and connect to it.

### Actual behaviour

Bisq does recognize that there's a local Bitcoin node running but can't connect to it and instead it shows "Bitcoin network peers: 0" and keeps printing the following in the command line:

``
Feb-03 22:36:08.775 [NioClientManager] INFO  org.bitcoinj.core.PeerGroup: [127.0.0.1]:8333: New peer      (1 connected, 0 pending, 1 max) 
Feb-03 22:36:08.775 [NioClientManager] INFO  org.bitcoinj.core.PeerGroup: Setting download peer: [127.0.0.1]:8333 
Feb-03 22:36:08.775 [NioClientManager] INFO  o.b.c.l.DownloadProgressTracker: Chain download switched to [127.0.0.1]:8333 
Feb-03 22:36:08.778 [NioClientManager] INFO  org.bitcoinj.core.PeerGroup: [127.0.0.1]:8333: Peer died      (0 connected, 0 pending, 1 max) 
Feb-03 22:36:08.778 [NioClientManager] INFO  org.bitcoinj.core.PeerGroup: Download peer died. Picking a new one. 
Feb-03 22:36:08.778 [NioClientManager] INFO  org.bitcoinj.core.PeerGroup: Unsetting download peer: [127.0.0.1]:8333 
Feb-03 22:36:08.778 [PeerGroup Thread] INFO  org.bitcoinj.core.PeerGroup: Peer discovery took 5.771 μs and returned 0 items 
Feb-03 22:36:08.779 [PeerGroup Thread] INFO  org.bitcoinj.core.PeerGroup: Waiting 1500 msec before next connect attempt to [127.0.0.1]:8333 
``

### Screenshots

bisq-alertb

#### Device or machine

64-bit Pop!_OS 18.04

#### Solution

This is solved by adding
``
peerbloomfilters=1
``
to the bitcoin.conf file. This info should be added to the "Background information" text box that alerts that Bisq has detected a local node and to make sure that it's not pruned. An additional note should warn Bitcoin Core 0.19 users of this issue.

dmos62:
Refactors LocalBitcoinNode and adds detection for local Bitcoin node's
configuration, namely, whether it is pruning and whether it has bloom
filter queries enabled.

The local node's configuration (and its presence) is retrieved by
performing a Bitcoin protocol handshake, which includes the local
Bitcoin node sending us its version message (VersionMessage in
BitcoinJ), which contains the information we're interested in.

Due to some quirky BitcoinJ logic, sometimes the handshake is
interrupted, even though we have received the local node's version
message. That contributes to the handshake handling in LocalBitcoinNode
being a bit complicated.

Refactoring consists of two principle changes: the public interface is
split into methods that trigger checks and methods that retrieve the
cached results. The methods that trigger checks have names starting
with "check", and methods that retrieve the cached results have names
that start with "is".

The other major refactor is the use of Optional<Boolean> instead of
boolean for storing and returning the results, an empty Optional
signifying that the relevant check was not yet performed. Switching to
Optionals has caused other code that queries LocalBitcoinNode to throw
an exception in case the query is made before the checks are. Before,
the results were instantiated to "false" and that would be returned
in case the query was made before the checks completed. This change has
revealed one occasion (Preferences class) where this happens.

This is shown in case a local Bitcoin node is found running, but misconfigured:
local node misconfigured popup

This is shown when setup with a local Bitcoin node is completed, to remind the user to have the node fully synced before using Bisq, and what the other requirements are: 
local node info popup

I hope to receive some phrasing corrections and maybe some UI ideas? I'm not totally happy with the "information" popup above: it's a bit redundant.

Possible next steps: 
- tell the user what exactly is misconfigured;
- option to retry the check.

I could have spent a bit more time and implemented these features too, but it's already two weeks that I'm doing this, so I'm just thrilled that I have something to PR. I can work on these later; I'll let you decide if these features are worthwhile.

Relevant issue: #3273


core/src/main/java/bisq/core/app/BisqSetup.java
```diff
@@ -559,7 +576,7 @@ else if (displayTorNetworkSettingsHandler != null)
 
         // We only init wallet service here if not using Tor for bitcoinj.
         // When using Tor, wallet init must be deferred until Tor is ready.
-        if (!preferences.getUseTorForBitcoinJ() || localBitcoinNode.isDetected()) {
```
dmos62, line 562: Removed redundant `LocalBitcoinNode` call, because `Preferences.getUseTorForBitcoinJ` makes the same call and it has the same effect.

desktop/src/main/java/bisq/desktop/main/MainViewModel.java
```diff
@@ -441,10 +451,12 @@ private void setupBtcNumPeersWatcher() {
                 checkNumberOfBtcPeersTimer = UserThread.runAfter(() -> {
                     // check again numPeers
                     if (walletsSetup.numPeersProperty().get() == 0) {
-                        if (localBitcoinNode.isDetected())
-                            getWalletServiceErrorMsg().set(Res.get("mainView.networkWarning.localhostBitcoinLost", Res.getBaseCurrencyName().toLowerCase()));
```
dmos62, line 445: Apart from the `.isDetected()` -> `.isUsable.get()` line, these are just formatting changes.

core/src/main/java/bisq/core/user/Preferences.java
```diff
@@ -736,11 +736,24 @@ public boolean showAgain(String key) {
...
+        // Sometimes this method, which queries LocalBitcoinNode for whether or not there's a
+        // usable local Bitcoin node, is called before LocalBitcoinNode has performed its
+        // checks. This was noticed when LocalBitcoinNode was refactored to return
+        // Optional<Boolean> istead of boolean, an empty Optional signifying that the relevant
+        // check has not yet been performed.
+        //
+        // To keep the method's behaviour unchanged, until a fix is implemented, we use
+        // Optional.orElse(false). Here 'false' normally means that the checks were performed
+        // and a suitable local Bitcoin node wasn't found.
+        var usableLocalNodePresent = localBitcoinNode.isUsable().orElse(false);
```
dmos62: This is the bug I mentioned.

sqrrm:
@dmos62 Both travis and codacy fail.

I will test this locally once it's building ok.

dmos62:
The implementation relies on BitcoinJ's NioClient attempting a handshake, but if that fails (for example when there isn't a locally running Bitcoin node), NioClient logs errors and exception to console, which is a bit annoying since we're already properly handling each case. Here's a screenshot:

nioclient exception spam

We could get the relevant BitcoinJ's logger and suppress it while we're doing these checks, but that feels a bit hacky (or does it?). Tell me how you feel about that.

dmos62:
I have made some changes to satisfy one of two Codacy complaints, and to make the other clearer. The other is a future callback that does nothing in case of failure, so `onFailure` is implemented, as required, but empty. So I put in a semi-colon to signify that it's meant to be empty:

            public void onFailure(Throwable thr) {  
                ;                                                
            }

I guess, I could have also did `public void onFailure(Throwable thr) {}`, and I think that was the original, but I passed the files through IntelliJ's reformatter and it forced a multiline declaration, so I thought I'd leave it at that.

dmos62 (commented):
I was surprised that the test suite for `LocalBitcoinNode` wasn't automatically executed, or compiled, when I was running `./gradlew desktop:build` and `desktop:run`.

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -229,6 +231,7 @@ public void onSuccess(Peer peer) {
                 peerVersionMessageFuture.set(peer.getPeerVersionMessage());
             }
             public void onFailure(Throwable thr) {
+                ;
```
sqrrm:
```suggestion
                // No action
```

sqrrm: I think a comment might work, codacy is complaining about the lone semicolon, which does look odd.

core/src/main/java/bisq/core/app/BisqSetup.java
```diff
@@ -482,14 +482,31 @@ private void maybeShowTac() {
...
         }
 
-        localBitcoinNode.detectAndRun(nextStep);
+        // Results of the check don't have to be passed to nextStep,
+        // because they're cached in LocalBitcoinNode and dependent routines query it themselves.
+        localBitcoinNode.checkUsable();
+
+        // Here we only want to provide the user with a choice (in a popup) in case a local node is
+        // detected, but badly configured.
+        var detectedButMisconfigured = localBitcoinNode.isDetectedButMisconfigured().get();
```
sqrrm: Since `localBitcoinNode.isDetectedButMisconfigured()` is an optional it would be prudent to check that it's present here. Although it might be certain for now that it's never null the characteristics of an optional is that it should be able to handle the null state.

core/src/main/java/bisq/core/app/BisqSetup.java
```diff
@@ -862,7 +879,7 @@ private void maybeShowSecurityRecommendation() {
     }
 
     private void maybeShowLocalhostRunningInfo() {
-        maybeTriggerDisplayHandler("bitcoinLocalhostNode", displayLocalhostHandler, localBitcoinNode.isDetected());
+        maybeTriggerDisplayHandler("bitcoinLocalhostNode", displayLocalhostHandler, localBitcoinNode.isUsable().get());
```
sqrrm: Same here with the usage of `Optional` without checking `isPresent()`.

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +49,272 @@
...
-        try (Socket socket = new Socket()) {
-            socket.connect(new InetSocketAddress("127.0.0.1", port), CONNECTION_TIMEOUT);
+
+    private static Peer createLocalPeer(
+            int port
+    ) throws UnknownHostException {
+        // TODO: what's the effect of NetworkParameters on handshake?
+        // i.e. is it fine to just always use MainNetParams?
+        var networkParameters = new MainNetParams();
+
```
sqrrm, line 133: So much space in this method...

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +49,272 @@
...
+    /**
+     * Initiates detection and configuration checks. The results are cached so that the public
+     * methods isUsable, isDetected, isWellConfigured don't trigger a recheck.
+     */
+
+    public boolean checkUsable() {
+        var optionalVersionMessage = attemptHandshakeForVersionMessage();
+        handleHandshakeAttempt(optionalVersionMessage);
+        // We know that the Optional/s will be populated by the end of the checks.
+        return isUsable().get();
```
sqrrm: Same with the checks for `Optional.isPresent()`.

sqrrm (changes requested):
Tested it and it works as expected, nice.

The errors logged by NioClientManager would be good to get rid of since most users won't be running with a local node so they would have big errors on startup. Not that familiar with logging, but if you have a way to do it I would say it's good.

I think the unchecked usages of `Optional` should be checked, perhaps use `orElse()`, didn't comment on all uses as there are quite a few.

NIT: Prefer no empty line between comment and method.

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +49,275 @@
...
-     * was successful. If the connection is successful, subsequent calls to
-     * {@link #isDetected()} will return {@code true}.
-     * @param callback code to run after detecting whether the node is running
+     * Creates an NioClient that is expected to only be used to coerce a VersionMessage out of a
+     * local Bitcoin node and be closed right after.
+     */
+
+    private static NioClient createClient(
+            Peer peer, int port, int connectionTimeout
+    ) throws IOException {
```
cbeams: To follow convention, formatting here should be:

```java
private static NioClient createClient(Peer peer, int port, int connectionTimeout) throws IOException {
```

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
cbeams: Yes, please apply same formatting advice here as I mentioned above on lines 65–67. And apply it consistently to any other similar changes in this PR as well. The general rule in effect here is to keep things on one line up to the 120-char right margin.

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +49,275 @@
...
+
+    /**
+     * Method backported from upstream bitcoinj: at the time of writing, our version is
+     * not BIP111-aware.
+     * Source routines and data can be found in Bitcoinj under:
+     * core/src/main/java/org/bitcoinj/core/VersionMessage.java
+     * and
+     * core/src/main/java/org/bitcoinj/core/NetworkParameters.java
+     */
+
```
cbeams: Nit: unnecessary newline here, and this need not be a (`/**`) Javadoc comment as it's on a private method and will never be published. Wrap comments at 90-char right margin, thanks.

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +49,275 @@
...
+    private static boolean isBloomFilteringSupportedAndEnabled(VersionMessage versionMessage) {
+        // A service bit that denotes whether the peer supports BIP37 bloom filters or not.
+        // The service bit is defined in BIP111.
+        int NODE_BLOOM = 1 << 2;
+
+        int BLOOM_FILTERS_BIP37_PROTOCOL_VERSION = 70000;
```
cbeams, line 192: Local variables should be camelCase.

cbeams (changes requested):
NACK on formatting and style issues. There are more than just a couple minor nits here, stuff that's worth getting right with a view toward future PRs being correct from the start.

This is by no means a complete review. I just took a quick look to understand what's changed with the new `LocalBitcoinNode` class and saw these superficial issues and commented on them. Unfortunately I don't have time to do more than that now.

core/src/main/java/bisq/core/app/BisqSetup.java
dmos62: I took your suggestion about getting rid of most of the Optional.get()s in other places, but in my opinion here a check would be superfluous, because it's easy to see that getting an empty Optional is impossible here with the current implementation of LocalBitcoinNode (`.checkUsable()` on the previous line guarantees that). That said if it were to change, all the calls would have to be reconsidered anyway, right?

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
dmos62: Do you feel strongly about this? These are constants, in the semantic sense, that I encapsulated in a single method with the code that uses them, and formating them as such helps single them out. I didn't want to separate them out to top-level static variables, because they're part of a small port from upstream BitcoinJ and I wanted to keep it completely separate from the rest of the class.

sqrrm: In my opinion this kind of containment is good. Camel case local variables would not be appropriate, but making them local seems right considering the situation.

sqrrm (dismissed):
utACK

I will test this before merging. @dmos62 if you feel like removing some of the extra empty lines in some of the methods meanwhile, please don't hesitate. I find the code harder to read with too much space as it breaks the flow compared with the other parts of the project.

dmos62:
> if you feel like removing some of the extra empty lines in some of the methods meanwhile, please don't hesitate.

@sqrrm could you point out examples of empty lines you would remove? I put some consideration into where I put empty lines. Usually to outline distinct steps in a method. Would appreciate feedback.

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +54,321 @@
...
+     * out of a local Bitcoin node and be closed right after.
+     */
+    private static NioClient createClient(Peer peer, int port, int connectionTimeout) throws IOException {
+        InetSocketAddress serverAddress =
+                new InetSocketAddress(InetAddress.getLocalHost(), port);
+
+        // This initiates the handshake procedure, which, if successful, will complete
+        // the peerVersionMessageFuture, or be cancelled, in case of failure.
+        NioClient client = new NioClient(serverAddress, peer, connectionTimeout);
+
```
sqrrm, line 88: Blank line

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +54,321 @@
...
+     */
+    private static Peer createLocalPeer(int port) throws UnknownHostException {
+        // TODO: what's the effect of NetworkParameters on handshake?
+        // i.e. is it fine to just always use MainNetParams?
+        var networkParameters = new MainNetParams();
+
+        // We must construct a BitcoinJ Context before using BitcoinJ. We don't keep a
+        // reference, because it's automatically kept in a thread local storage.
+        new Context(networkParameters);
+
```
sqrrm, line 121: Blank line

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +54,321 @@
...
+        var ourVersionMessage = new VersionMessage(networkParameters, 0);
+
```
sqrrm, line 133: Blank line

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +54,321 @@
...
+        var localPeerAddress = new PeerAddress(InetAddress.getLocalHost(), port);
+
```
sqrrm, line 159: Blank line

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +54,321 @@
...
+        AbstractBlockChain blockchain = null;
+
```
sqrrm, line 162: Blank line

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +54,321 @@
...
+     * Source routines and data can be found in Bitcoinj under:
+     * core/src/main/java/org/bitcoinj/core/VersionMessage.java
+     * and
+     * core/src/main/java/org/bitcoinj/core/NetworkParameters.java
+     */
+    private static boolean isBloomFilteringSupportedAndEnabled(VersionMessage versionMessage) {
+        // A service bit that denotes whether the peer supports BIP37 bloom filters or
+        // not. The service bit is defined in BIP111.
+        int NODE_BLOOM = 1 << 2;
+
```
sqrrm, line 191: Blank line

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +54,321 @@
...
+        int BLOOM_FILTERS_BIP37_PROTOCOL_VERSION = 70000;
+        var whenBloomFiltersWereIntroduced = BLOOM_FILTERS_BIP37_PROTOCOL_VERSION;
+
+        int BLOOM_FILTERS_BIP111_PROTOCOL_VERSION = 70011;
+        var whenBloomFiltersWereDisabledByDefault = BLOOM_FILTERS_BIP111_PROTOCOL_VERSION;
+
```
sqrrm, line 197: Blank line

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +54,321 @@
...
+     * core/src/main/java/org/bitcoinj/core/NetworkParameters.java
+     */
+    private static boolean isBloomFilteringSupportedAndEnabled(VersionMessage versionMessage) {
+        // A service bit that denotes whether the peer supports BIP37 bloom filters or
+        // not. The service bit is defined in BIP111.
+        int NODE_BLOOM = 1 << 2;
+
+        int BLOOM_FILTERS_BIP37_PROTOCOL_VERSION = 70000;
+        var whenBloomFiltersWereIntroduced = BLOOM_FILTERS_BIP37_PROTOCOL_VERSION;
+
```
sqrrm, line 194: Blank line

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +54,321 @@
...
+                    + " (before a version messages could be coerced)",
+                    port);
+        } else {
+            detected = Optional.of(true);
             log.info("Local Bitcoin node detected on port {}", port);
-            detected = true;
+
+            var versionMessage = optionalVersionMessage.get();
+            var configurationCheckResult = checkWellConfigured(versionMessage);
+
```
sqrrm: Blank line

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +54,321 @@
...
+        /* We temporarily silence BitcoinJ NioClient's and NioClientManager's loggers,
+         * because when a local Bitcoin node is not found they pollute console output
+         * with "connection refused" error messages.
+         */
+        var originalNioClientLoggerLevel = silence(NioClient.class);
+        var originalNioClientManagerLoggerLevel = silence(NioClientManager.class);
+
+        NioClient client;
+
```
sqrrm: Blank line

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +54,321 @@
...
+            client = createClient(peer, port, CONNECTION_TIMEOUT);
         } catch (IOException ex) {
-            log.info("No local Bitcoin node detected on port {}.", port);
+            log.error("Local bitcoin node handshake attempt unexpectedly threw: {}", ex);
+            return Optional.empty();
         }
-        callback.run();
+
+        ListenableFuture<VersionMessage> peerVersionMessageFuture = getVersionMessage(peer);
+
```
sqrrm: Blank line

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +54,321 @@
...
+        try {
+            var peerVersionMessage = peerVersionMessageFuture.get();
+            optionalPeerVersionMessage = Optional.of(peerVersionMessage);
+        } catch (ExecutionException | InterruptedException | CancellationException ex) {
+            optionalPeerVersionMessage = Optional.empty();
+        }
+
+        peer.close();
+        client.closeConnection();
+
```
sqrrm, line 255: Blank line

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +54,321 @@
...
+        getLogger(klass).setLevel(originalLevel);
+    }
+
+    private static ch.qos.logback.classic.Logger getLogger(Class klass) {
+        return (ch.qos.logback.classic.Logger) LoggerFactory.getLogger(klass);
+    }
+
+    private ListenableFuture<VersionMessage> getVersionMessage(Peer peer) {
+        SettableFuture<VersionMessage> peerVersionMessageFuture = SettableFuture.create();
+
```
sqrrm, line 309: Blank line

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +54,321 @@
...
+             * was before we switched to using Optionals. More specifically, the only query
+             * method at the time, isDetected(), would return false in such a case. We are
+             * relatively confident that the previous behaviour doesn't cause fatal bugs,
+             * so, in case LocalBitcoinNode is queried too early, we revert to it, instead
+             * of letting Optional.empty().get() throw an exception. The advantage over
+             * plain booleans then is that we can log the below error message (with
+             * stacktrace).
+             */
+            var whenChecksNotFinished = false;
+
```
sqrrm: Blank line

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -25,39 +54,321 @@
...
+            var throwable = new Throwable("LocalBitcoinNode was queried before it was ready");
+
```
sqrrm: Blank line

core/src/main/java/bisq/core/user/Preferences.java
```diff
@@ -736,11 +736,20 @@ public boolean showAgain(String key) {
...
         // On testnet there are very few Bitcoin tor nodes and we don't provide tor nodes.
+
+        // TODO bug. Non-critical, apparently.
+        // Sometimes this method, which queries LocalBitcoinNode for whether or not there's a
+        // usable local Bitcoin node, is called before LocalBitcoinNode has performed its
+        // checks. This was noticed when LocalBitcoinNode was refactored to return
+        // Optional<Boolean> istead of boolean, an empty Optional signifying that the relevant
+        // check has not yet been performed.
+        var usableLocalNodePresent = localBitcoinNode.safeIsUsable();
+
```
sqrrm, line 743: Blank line

sqrrm (commented):
I added a comment on some of the blank lines I would remove. Personally I would probably make it even tighter though.

I understand the motivation of separating some of the concepts with blank lines, but it doesn't really help with readability since the methods themselves already contain the scope of activity.

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
dmos62: Here rationale is to separate the creation of the future from the parts that use it. The next three parts of the method operate on this future.

dmos62:
@sqrrm thanks for the feedback. I appreciate it. In all your noted cases my motivation was to make visual some structure I saw in the methods. I commented under one of them detailing that a bit further. I would not feel comfortable removing those empty lines. We could write this off as stylistic differences. I'll say again that I do appreciate the feedback: I wasn't aware that this feature of my code was jarring to some.

Further thinking on this, I would say that I'm minimally sensitive to vertical verbosity (or horizontal for that matter). It might have to do with how I use my vim when working. I jump around the code a lot and information density on-screen at any given time is not really a direct factor for me.

dmos62:
I have two major changes in the works that could be merged in this PR. Related to #4005.

My initial approach was to make the bare minimum of changes to get these features, and I ignored a gut feeling that more profound changes could be useful. While thinking about #4005, I came full circle, and decided I'd like to change the following:

- set `LocalBitcoinNode` to trigger the checks automatically when it is queried for the first time;
- create a `LocalBitcoinNode::willUse` query method that would centralize logic that decides whether or not to use the local BTC node.

### Automatic triggering of checks

In my opinion, automatic triggering of checks is essential, as illustrated by #4005. We don't have a compile time assurance that there won't be inappropriate usages of LocalBitcoinNode (which currently requires the checks to be triggered before it can tell you anything about a locally running BTC node). My suggestion is to solve this by having the first query trigger the check.

### New query method `willUse`

`willUse` is clean up of equivalent logic that is littered throughout the codebase. The problem here is that we currently only centralize the questions "is there a local BTC node" and "is it usable/configured", but the codebase has to answer other related questions. Below I enumerate all related (implicit) questions: 

- "is there a local BTC node" in `NetworkSettingsView`;
- "should a usable local BTC node be ignored" in `BisqSetup` and `WalletConfig`;
- "are we going to use (are we using) a local BTC node" in `WalletSetup`, `Preferences`, `MainViewModel` and `NetworkSettingsView`.

These questions are all answered by combining `LocalBitcoinNode::isUsable` with inline logic. It results in code that is difficult to maintain and to read.

A `LocalBitcoinNode::willUse`, together with the already provided `isDetected` and `isUsable`, will cover all these cases, and will declutter local Bitcoin node related logic.

### Note

These two changes don't depend on each other.

dmos62:
Above commits, namely 0bbbe8c and 6b4878a, implement the changes I talk about in above post.

core/src/main/java/bisq/core/app/BisqSetup.java
```diff
@@ -482,14 +482,21 @@ private void maybeShowTac() {
...
             nextStep.run();
             return;
         }
 
-        localBitcoinNode.detectAndRun(nextStep);
+        // Here we only want to provide the user with a choice (in a popup) in case a
+        // local node is detected, but badly configured.
+        var detectedButMisconfigured = localBitcoinNode.isDetectedButMisconfigured();
+        if (detectedButMisconfigured) {
+            displayLocalNodeMisconfigurationHandler.accept(nextStep);
```
ripcurlx:
```suggestion
            if (displayLocalNodeMisconfigurationHandler != null) {
                displayLocalNodeMisconfigurationHandler.accept(nextStep);
            }
```

Add a check to prevent possible NullPointerException.

core/src/main/java/bisq/core/app/BisqSetup.java
ripcurlx: If `localBitcoinNode.willUse() == true` and `config.useTorForBtcOptionSetExplicitly == true` the new code will behave differently than before. So I don't think we should remove this from here.

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+     * could be noticably simplified, by turning connection failure callback into a
+     * future and using a first-future-to-complete type of construct, but I couldn't find
+     * a ready-made implementation.
+     */
+    private Optional<VersionMessage> attemptHandshakeForVersionMessage() {
+        Peer peer;
+        try {
+            peer = createLocalPeer(port);
+        } catch (UnknownHostException ex) {
+            log.error("Local bitcoin node handshake attempt unexpectedly threw: {}", ex);
```
ripcurlx: Using it like that the placeholder will never be filled. Maybe something like:
```suggestion
            log.error("Local bitcoin node handshake attempt was unexpectedly interrupted", ex);
```

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+
+        NioClient client;
+
+        try {
+            log.info("Initiating attempt to connect to and handshake with a local "
+                    + "Bitcoin node (which may or may not be running) on port {}.",
+                    port);
+            client = createClient(peer, port, CONNECTION_TIMEOUT);
+        } catch (IOException ex) {
+            log.error("Local bitcoin node handshake attempt unexpectedly threw: {}", ex);
```
ripcurlx: Using it like that the placeholder will never be filled. Maybe something like:
```suggestion
            log.error("Local bitcoin node handshake attempt was unexpectedly interrupted", ex);
```

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+        new Context(networkParameters);
+
+        var ourVersionMessage = new VersionMessage(networkParameters, 0);
+
+        var localPeerAddress = new PeerAddress(InetAddress.getLocalHost(), port);
+
+        AbstractBlockChain blockchain = null;
+
+        var peer = new Peer(networkParameters, ourVersionMessage, localPeerAddress, blockchain);
+        return peer;
```
ripcurlx: Not sure if that makes it so much more readable than simply:
```
return new Peer(networkParameters, ourVersionMessage, localPeerAddress, blockchain);
```

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+     */
+    private static NioClient createClient(Peer peer, int port, int connectionTimeout) throws IOException {
+        InetSocketAddress serverAddress =
+                new InetSocketAddress(InetAddress.getLocalHost(), port);
+
+        // This initiates the handshake procedure, which, if successful, will complete
+        // the peerVersionMessageFuture, or be cancelled, in case of failure.
+        NioClient client = new NioClient(serverAddress, peer, connectionTimeout);
+
+        return client;
```
ripcurlx: Same here:
```
return new NioClient(serverAddress, peer, connectionTimeout);
```

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+        // reference, because it's automatically kept in a thread local storage.
+        new Context(networkParameters);
+
+        var ourVersionMessage = new VersionMessage(networkParameters, 0);
+
+        var localPeerAddress = new PeerAddress(InetAddress.getLocalHost(), port);
+
+        AbstractBlockChain blockchain = null;
+
+        var peer = new Peer(networkParameters, ourVersionMessage, localPeerAddress, blockchain);
```
ripcurlx: I know that you do this because you use Vim, but everyone with a more sophisticated editor has this parameter information next to the null value in the constructor instantiation

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+                new InetSocketAddress(InetAddress.getLocalHost(), port);
+
+        // This initiates the handshake procedure, which, if successful, will complete
+        // the peerVersionMessageFuture, or be cancelled, in case of failure.
+        NioClient client = new NioClient(serverAddress, peer, connectionTimeout);
+
+        return client;
+    }
+
+    private static Level silence(Class klass) {
```
ripcurlx:
```suggestion
    private static Level silence(Class<?> klass) {
```

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+        var logger = getLogger(klass);
+        var originalLevel = logger.getLevel();
+        logger.setLevel(Level.OFF);
+        return originalLevel;
+    }
+
+    private static void restoreLoggerLevel(Class klass, Level originalLevel) {
```
ripcurlx:
```suggestion
    private static void restoreLoggerLevel(Class<?> klass, Level originalLevel) {
```

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+        getLogger(klass).setLevel(originalLevel);
+    }
+
+    private static ch.qos.logback.classic.Logger getLogger(Class klass) {
```
ripcurlx:
```suggestion
    private static ch.qos.logback.classic.Logger getLogger(Class<?> klass) {
```

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+        return (ch.qos.logback.classic.Logger) LoggerFactory.getLogger(klass);
+    }
+
+    private ListenableFuture<VersionMessage> getVersionMessage(Peer peer) {
+        SettableFuture<VersionMessage> peerVersionMessageFuture = SettableFuture.create();
+
+        var versionHandshakeDone = peer.getVersionHandshakeFuture();
+        FutureCallback<Peer> fetchPeerVersionMessage = new FutureCallback<Peer>() {
```
ripcurlx:
```suggestion
        FutureCallback<Peer> fetchPeerVersionMessage = new FutureCallback<>() {
```

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+            public void onSuccess(Peer peer) {
+                peerVersionMessageFuture.set(peer.getPeerVersionMessage());
+            }
+            public void onFailure(Throwable thr) {
```
ripcurlx:
```suggestion
            public void onFailure(@NotNull Throwable thr) {
```

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+            }
+        };
+        Futures.addCallback(
+                versionHandshakeDone,
+                fetchPeerVersionMessage
+        );
+
+        PeerDisconnectedEventListener cancelIfConnectionFails =
+                new PeerDisconnectedEventListener() {
+                    public void onPeerDisconnected(Peer peer, int peerCount) {
```
ripcurlx: Can be replaced with:
```
(disconnectedPeer, peerCount) -> {
```

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+                        if (peerVersionMessageAlreadyReceived) {
+                            // This method is called whether or not the handshake was
+                            // successful. In case it was successful, we don't want to do
+                            // anything here.
+                            return;
+                        }
+                        // In some cases Peer will self-disconnect after receiving
+                        // node's VersionMessage, but before completing the handshake.
+                        // In such a case, we want to retrieve the VersionMessage.
+                        var peerVersionMessage = peer.getPeerVersionMessage();
```
ripcurlx: That would need also to change this to:
```suggestion
                        var peerVersionMessage = disconnectedPeer.getPeerVersionMessage();
```
to prevent scope issues with the other peer passed as parameter. This makes it also easier to distinguish that they are different peers.

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+                        if (peerVersionMessage != null) {
+                            log.info("Handshake attempt was interrupted;"
+                                    + " however, the local node's version message was coerced.");
+                            peerVersionMessageFuture.set(peerVersionMessage);
+                        } else {
+                            log.info("Handshake attempt did not result in a version message exchange.");
+                            var mayInterruptWhileRunning = true;
+                            peerVersionMessageFuture.cancel(mayInterruptWhileRunning);
```
ripcurlx: Here again: This is something that helps only someone not using code editors like IntelliJ.

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+     */
+    public boolean willIgnore() {
+        BaseCurrencyNetwork baseCurrencyNetwork = config.baseCurrencyNetwork;
+
+        // For dao testnet (server side regtest) we disable the use of local bitcoin node to
+        // avoid confusion if local btc node is not synced with our dao testnet master node.
+        // Note: above comment was previously in WalletConfig::createPeerGroup.
+
+        return config.ignoreLocalBtcNode
+            || baseCurrencyNetwork.isDaoRegTest()
```
ripcurlx:
```suggestion
                || baseCurrencyNetwork.isDaoRegTest()
```

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+            || baseCurrencyNetwork.isDaoTestNet();
```
ripcurlx:
```suggestion
                || baseCurrencyNetwork.isDaoTestNet();
```

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+        var optionalVersionMessage = attemptHandshakeForVersionMessage();
+        handleHandshakeAttempt(optionalVersionMessage);
+    }
+
+    private void handleHandshakeAttempt(Optional<VersionMessage> optionalVersionMessage) {
+        if (!optionalVersionMessage.isPresent()) {
+            detected = false;
+            wellConfigured = false;
+            log.info("No local Bitcoin node detected on port {},"
+                    + " or the connection was prematurely closed"
```
ripcurlx:
```suggestion
                            + " or the connection was prematurely closed"
```

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+                    + " (before a version messages could be coerced)",
```
ripcurlx:
```suggestion
                            + " (before a version messages could be coerced)",
```

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
```diff
@@ -24,40 +53,321 @@
...
+         * with "connection refused" error messages.
+         */
+        var originalNioClientLoggerLevel = silence(NioClient.class);
+        var originalNioClientManagerLoggerLevel = silence(NioClientManager.class);
+
+        NioClient client;
+
+        try {
+            log.info("Initiating attempt to connect to and handshake with a local "
+                    + "Bitcoin node (which may or may not be running) on port {}.",
```
ripcurlx:
```suggestion
                            + "Bitcoin node (which may or may not be running) on port {}.",
```

ripcurlx (changes requested):
NACK

Please see my remarks in the code.

I tested the functionality on Regtest with a local Bitcoin Core and I don't get the informational popup anymore when running the client with a properly configured client
Bildschirmfoto 2020-02-26 um 11 29 59
Having peer bloom filters not configured and I also don't get the warning. Doesn't it work with the local Regtest setup?

dmos62:
> NACK
> 
> Please see my remarks in the code.
> 
> I tested the functionality on Regtest with a local Bitcoin Core and I don't get the informational popup anymore when running the client with a properly configured client
>
> Having peer bloom filters not configured and I also don't get the warning. Doesn't it work with the local Regtest setup?

Yes, if below expression evaluates to true, a local Bitcoin node is ignored:

```
config.ignoreLocalBtcNode
    || baseCurrencyNetwork.isDaoRegTest()
    || baseCurrencyNetwork.isDaoTestNet()
```

I didn't introduce this, though I centralized this logic, because it was repeated in several places around the codebase.

The "Background information" popup shouldn't be shown if you have bloom filters set to 1. That's weird. It is currently configured to show when a well configured local BTC node is found (LocalBitcoinNode.isUsable()).

I retested in mainnets and both pruning and disabled bloom filters are detected. I'll try and replicate your setup to see if I can understand what's happening.

core/src/main/java/bisq/core/app/BisqSetup.java
dmos62: You're right, I missed that. Thanks! My concern here is that this line decides whether to "initiate wallet" at this time, which implies that Tor will not be used. Ideally `Preferences.getUseTorForBitcoinJ()` would be a full authority on whether or not we're using Tor, but here it's only one half of the condition. So I can restore initial logic, but ideally I'd like to restructure this. I'll see what I can do.

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
dmos62: Is this a win/lose situation where we have to adopt to the majority? It does help me, as I'm not using IntelliJ. Does it disturb IntelliJ users?

dmos62: I've commited this and other IDE-related changes, since Github back-and-forths can be slow, but I'd like an answer to this.

dmos62:
@ripcurlx Pushed changes, including a major fix. I was probably wrong about why detection failed for you.

I was unable to replicate what you experienced. There was a bug preventing detection from working on other `baseCurrencyNetwork`s than mainnet, but it didn't seem to affect you. You managed to handshake with a regtest btc node and not get it detected as having bloom filters disabled. That should have been impossible for at least two reasons. I may have misunderstood what your setup was. I tested (successfully) with mainnet (no params to Bisq) and with following params (from dev-setup.md):

```
--baseCurrencyNetwork=BTC_REGTEST --useLocalhostForP2P=true
--useDevPrivilegeKeys=true --nodePort=5555 --appName=bisq-BTC_REGTEST_Alice
```
and `regtest=1` in bitcoin.conf.

Could you provide the Bisq arguments you used and the Bitcoin Core config as well?

core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
sqrrm: I agree with @ripcurlx here. In general I think the code should try to stick to the current style of the project. I did not choose this style but I follow it to make it easier for everyone to read the code. If everyone were to follow their own preferred style it would make the code much less readable.

This kind of extra variable comes as a surprise here since it's not following the common style and it makes me wonder why it's done in this particular way, as if it has some significance.

dmos62: @sqrrm the reason why I use these is to not have "anonymous" values in calls. Example: `buildHouse(false)` v. `buildHouse(withRoof)`.

cbeams:
I suggest disengaging from style discussions at this point. If the substantive changes in this PR are otherwise ACK-worthy, then merge it, and someone can issue a follow-up PR correcting the deviations. It’ll be faster that way. I’ll do it myself, even.

Alternatively, one of the maintainers could push those changes against the current PR branch, but it’s probably better in this situation to have a separate PR that documents the rework necessary.

@dmos, thanks for your contributions, but in a large established codebase like Bisq, a contributor’s goal should be to produce stylistically perfect patches, where perfect is defined by not causing reviewers to digress into style discussions or needing to clean up after you.

It takes some time to catch up with the nuances of style conventions in any large codebase and Bisq is far from perfectly consistent with its own conventions. As such, it’s expected that newcomers will make mistakes here and there; that’s not a problem. The main thing is to strive for patches that require as little overhead as possible for reviewers and maintainers, and style quibbles are overhead.

cbeams:
It looks like my previous message was sent too hastily. I sent it from my phone, where the last notifications I saw on this review thread were about style quibbles (explanatory variable declarations in particular). On returning to my laptop, I see that @dmos62 has followed up with commits that largely resolve the stylistic deviations.

I am wondering, however, why `LocalBitcoinNodeTests` was deleted with nothing seeming to replace it. Forgive me if that's been discussed somewhere already and I missed it.

cbeams:
I've just pushed a style polishing commit to my own fork at https://github.com/cbeams/bisq/commit/b93ca2b2b1dd24cd53014f9bd85269d572fde032. @dmos62, please consider pushing it to this PR branch.

I am also working on an additional commit that makes several more substantive changes. I'll wrap it up and push it it to my branch in my (European) AM tomorrow (Thursday). Please let me know if you have any significant local changes that haven't been pushed, so I'm aware whether or not conflicts are likely. Thanks.

dmos62:
@cbeams I don't have changes in the works. The test suite was removed because I don't have ideas about how to replace it. Before the detection mechanism was just checking if a socket can be opened; now a whole Bitcoin protocol handshake has to be performed.

I don't mean to be a nuisance with the stylistic discussions. I guess I didn't consider the additional effort the differences created. I'll try and emulate the rest of the codebase.

cbeams:
@dmos62, I've just submitted dmos62/bisq#1 with my changes, including the first polishing commit I mentioned above. Please merge and push that and the changes will show up here in this PR.

Note that I have NOT actually tested the functionality of this larger PR on my local machine, so I'm deferring to @ripcurlx's outstanding review comment regarding his current inability to get the behavior to work as expected.

ripcurlx:
Just tested the current state again on Regtest.

Background information for my local Bitcoin Core is shown again when configured properly
Bildschirmfoto 2020-02-27 um 10 10 35

Warning is now shown when peerbloomfilter is not enabled
Bildschirmfoto 2020-02-27 um 10 14 10

When trying to set e.g. `prune=550` somehow during debugging the app got stuck, but running it without breakpoints did work as expected.
Bildschirmfoto 2020-02-27 um 10 22 04

Following exceptions are logged in the command line by bitcoinJ which is expected I guess.

```
Feb.-27 10:21:12.035 [JavaFX Application Thread] INFO  b.c.btc.nodes.LocalBitcoinNode: Local Bitcoin node detected on port 18444 
Feb.-27 10:21:12.035 [JavaFX Application Thread] INFO  b.c.btc.nodes.LocalBitcoinNode: Local Bitcoin node badly configured (it is pruning and/or bloom filters are disabled) 
Feb.-27 10:21:12.037 [NioClientManager] WARN  o.b.core.PeerSocketHandler: [127.0.0.1]:18444 -  org.bitcoinj.core.ProtocolException: Received AlertMessage before version handshake is complete.
	at org.bitcoinj.core.Peer.processMessage(Peer.java:478)
	at org.bitcoinj.core.PeerSocketHandler.receiveBytes(PeerSocketHandler.java:184)
	at org.bitcoinj.net.NioClient$Handler.receiveBytes(NioClient.java:71)
	at org.bitcoinj.net.ConnectionHandler.handleKey(ConnectionHandler.java:223)

org.bitcoinj.core.ProtocolException: Received AlertMessage before version handshake is complete.
	at org.bitcoinj.core.Peer.processMessage(Peer.java:478)
	at org.bitcoinj.core.PeerSocketHandler.receiveBytes(PeerSocketHandler.java:184)
	at org.bitcoinj.net.NioClient$Handler.receiveBytes(NioClient.java:71)
	at org.bitcoinj.net.ConnectionHandler.handleKey(ConnectionHandler.java:223)
Feb.-27 10:21:12.039 [NioClientManager] WARN  o.b.net.ConnectionHandler: Error handling SelectionKey: java.lang.IllegalStateException  java.lang.IllegalStateException: null
	at com.google.common.base.Preconditions.checkState(Preconditions.java:429)
	at org.bitcoinj.net.ConnectionHandler.handleKey(ConnectionHandler.java:224)

java.lang.IllegalStateException: null
	at com.google.common.base.Preconditions.checkState(Preconditions.java:429)
	at org.bitcoinj.net.ConnectionHandler.handleKey(ConnectionHandler.java:224)

```

So ACK from my side now from a functionality point of view.  Please apply the [polishing commit](https://github.com/cbeams/bisq/commit/b93ca2b2b1dd24cd53014f9bd85269d572fde032) by @cbeams and I'm happy to merge this PR. Thanks for your contribution @dmos62 !

cbeams:
Did we consider at any point in this process doing these same pruning and bloom filter checks not only against a local Bitcoin node, but also against (a) Bisq's provided federation of nodes and (b) any custom nodes provided by the user? In the case of (a), it's probably a waste to do these checks, since our own federation nodes should always be properly configured, but case (b) is more important, especially for users who are pointing their local Bisq node at their own remote Bitcoin full node. See the last paragraph of the commit comment at https://github.com/dmos62/bisq/pull/1/commits/57b7041 for more on this.

In any case, I'm not suggesting that doing these checks against the federation or user-specified custom nodes should be done as part of this PR. I just want to see if it's been thought about at all.

cbeams (changes requested):
NACK per https://github.com/bisq-network/bisq/pull/3982#issuecomment-591869265

ripcurlx:
> In any case, I'm not suggesting that doing these checks against the federation or user-specified custom nodes should be done as part of this PR. I just want to see if it's been thought about at all.

Atm we temp [deactivated the option to connect to the public network](https://github.com/bisq-network/bisq/pull/3938) because of the lack of a node capabilities check. I think after this PR is merged this would be the next step if we are able to apply this checks also to our public network and user-specified custom node option.

cbeams:
> if we are able to apply this checks also to our public network

Did you mean "to our federation" here? I don't see a lot of value in re-enabling the public network option at this point. So long as Bisq relies on SPV, users are at risk on the public network because of chain surveillance. I think keeping it just to our own federation or a user's custom node(s) is sufficient. Doing checks against (at least) user-provided custom nodes would make sense to me for the same reasons doing it against their local node makes sense.

core/src/main/java/bisq/core/app/misc/ModuleForAppWithP2p.java
```diff
@@ -50,7 +50,6 @@
 import java.io.File;
 
 import static bisq.common.config.Config.*;
-import static bisq.core.btc.nodes.LocalBitcoinNode.LOCAL_BITCOIN_NODE_PORT;
```
cbeams, line 53: Good catch, thanks @dmos62.

cbeams (approved):

ripcurlx:
> Did you mean "to our federation" here? I don't see a lot of value in re-enabling the public network option at this point. So long as Bisq relies on SPV, users are at risk on the public network because of chain surveillance. I think keeping it just to our own federation or a user's custom node(s) is sufficient. Doing checks against (at least) user-provided custom nodes would make sense to me for the same reasons doing it against their local node makes sense.

I was talking about "public" public network. But yes, we can keep it deactivated for security reasons if on-one wants to explicitly use this feature because of specific reasons.

ripcurlx (approved):

Kixunil:
Wow, @dmos62 that wording is really clear! More projects should be like this!

PR #3996, opened 2020-02-21
Reference critical bug process contributing guide

ripcurlx:
Part of bisq-network/projects#16.

This change also replaces deprecated communication channel


cbeams (approved):
ACK

PR #3999, opened 2020-02-21
Add mynode.local as a block chain explorer option

rbrooklyn:
Add http://mynode.local:3002 as a URL for the user to select in the preferences for blockchain explorer to use.

This allows the user to use their own blockchain explorer running in a mynode installation on their own network. If not installation is present, then the link launched will not load, however, the user is unlikely to select this option without knowing what this is.

This is a trivial change.


RiccardoMasutti (approved):
ACK

ripcurlx:
@rbrooklyn Travis is detecting an abuse by your PR although it is probably absolut legit. We had this in the past already with one contributor. Back then someone else with a different account needed to cherry-pick the commits and create a PR themselves as there is no way to resolve this abuse warning manually.

rbrooklyn:
It's literally a one line change which I put in, which was a copy and paste from the line above it and changing a couple of strings. If that triggers some kind "abuse" detection then I'd say the abuse detection is broken, unless it's going by some kind of reputation scoring or something, as I probably don't have any on GitHub.

For what it's worth, this is the information I'm seeing. Seems to suggest that someone doing a "approving review" would push it through, although obviously I'm no expert:
image

It's up to you guys what you do with this. I wanted to contribute but I'm somewhat unwilling to keep putting effort in for a one line change. If this change is desired then it makes sense that someone with proper access to the repo do it rather than us wasting all this effort here.

PR #4002, opened 2020-02-24
Avoid startup failure when appDataDir is a symbolic link

### LINKED ISSUE #3998: Bisq fails to start because of data directory issue
### Description

Based on https://bisq.community/t/new-install-not-starting-up/8963/25

Users cannot start version v1.2.7 because of following exception being thrown:
```
$ /opt/Bisq/Bisq
error: Application data directory '/home/xxx/.local/share/Bisq' could not be created
```

#### Version

Regression bug affecting version v1.2.7.
Does work in v1.2.5.

### Steps to reproduce

Unfortunately not 100% clear why this is happening.
Every time this users start v1.2.7 with an existing data directory the app fails to start.
Reverting back to v1.2.5 solves the problem and app starts as expected.

### Expected behaviour

App starts.

### Actual behaviour

App fails to start.

#### Device or machine

We know about one affected user on Windows and one on Linux.

cbeams:
This PR fixes #3998 as explained in the commit comment for cbe6f19.

### DIFF
--- a/common/src/main/java/bisq/common/config/Config.java
+++ b/common/src/main/java/bisq/common/config/Config.java
@@ -798,10 +798,12 @@ public class Config {
      * @return the given directory, now guaranteed to exist
      */
     private static File mkAppDataDir(File dir) {
-        try {
-            Files.createDirectories(dir.toPath());
-        } catch (IOException ex) {
-            throw new UncheckedIOException(format("Application data directory '%s' could not be created", dir), ex);
+        if (!dir.exists()) {
+            try {
+                Files.createDirectories(dir.toPath());
+            } catch (IOException ex) {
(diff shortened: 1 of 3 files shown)


cbeams:
Heads up, I am about to push some additional commits to this PR branch that properly fix #3998. I'll update the title and description of the PR accordingly after doing so. Please don't merge in the meantime.

cbeams:
I've updated the title and description of this PR to reflect the actual fix. I've now gotten confirmation from both affected users that they were using symbolic links in their appDataDir paths, so we should be good to go to merge this, i.e. we should feel confident that these changes actually fix the bug.

ripcurlx (approved):
utACK

PR #4007, opened 2020-02-25
Add ARS as supported currency for MoneyGram

### LINKED ISSUE #4000: Moneygram Payment method.  ARS currency missing
It is not possible to post offers in ARS/BTC pair with Moneygram payment method, but Moneygram does support ARS currency (see picture below).   Argentina is very underveloped market pair in Bisq compared to the strong Bitcoin activity in that country.  I am willing to post offers in ARS/BTC pair, and payment methods such as Moneygram (or Western Union) are a very good option to bootstrap markets with no Bisq activity, but at this moment Bisq does not allow to use them. 

image

ripcurlx:
Fixes #4000.


sqrrm (approved):
utACK

PR #4008, opened 2020-02-25
Tableview multiple rows checkbox column issue

### LINKED ISSUE #3984: Utilizing the Enable Checkbox Results in Unexpected Results When Multiple Orders Exist
### Description
When multiple orders are open, when trying to utilize the "Enable" checkbox to deselect and select what order should be displayed, multiple orders appear to change their classification from Enabled to Disabled, or Disabled to Enabled, across multiple positions with a single click.

#### Version
1.2.7

### Steps to reproduce
Create multiple bids and offers across many currencies.  In my case, the number of open orders are above 30, and the list of open orders on the PORTFOLIO=>MY OPEN ORDERS extends below the bottom of the screen.  Scroll down below the bottom of the screen (this is important to cause the bug.)  Upon attempting to enable or disable a given transaction after scrolling down—particularly after resorting the list by buy/sell or by currency—the results are not as expected.

### Expected behaviour
Only the selected trade should be toggled between enabled and disabled.

### Actual behaviour
Multiple transactions toggle between enable and disable, sometimes not including the actual intended toggle target trade.  It is difficult to discern whether the list resorts, as well.

### Screenshots
Prior to scrolling down:
image

Slight scroll down:
image

Results after clicking the 7th trade on the list (one click); note the changes at the bottom:
image

#### Device or machine
PC Windows 10

beingindot:
Fixes #3984 

As mentioned in the issue #3984, Have created lot of buy/sell offers with different currencies spanning across around 30+ offers.
so that we need to use scroll bar in my open offers section and able to recreate the issue.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersView.java
+++ b/desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersView.java
@@ -406,8 +406,6 @@ public class OpenOffersView extends ActivatableViewAndModel<VBox, OpenOffersView
                                     if (checkBox == null) {
                                         checkBox = new AutoTooltipCheckBox();
                                         checkBox.setGraphic(iconView);
-                                        updateState(item.getOpenOffer());
-                                        setGraphic(checkBox);
                                     }
                                     checkBox.setOnAction(event -> {
                                         if (item.getOpenOffer().isDeactivated()) {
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
ACK

Tested it on local Regtest having more than 30 offers (scrolling is required) and when changing the enabled state of a single offer at the bottom the visual representation of other offers is changed as well. The state itself doesn't change just the rendering is wrong. Using the code in this PR fixes this problem

PR #4014, opened 2020-02-27
Make price providers cmdline param more readable

freimair:
We have a line in `bisq-desktop --help` saying:
```
  --providers=<host:port[,...]>
        List custom providers
```
Like so, it is not clear what providers you can set using that config option.

This PR only adds "price" to `List custom /price/ providers`.

Stumbled upon this during network size metric testing.

### DIFF
--- a/common/src/main/java/bisq/common/config/Config.java
+++ b/common/src/main/java/bisq/common/config/Config.java
@@ -373,7 +373,7 @@ public class Config {
                         .defaultsTo(false);
 
         ArgumentAcceptingOptionSpec<String> providersOpt =
-                parser.accepts(PROVIDERS, "List custom providers")
+                parser.accepts(PROVIDERS, "List custom price providers")
                         .withRequiredArg()
                         .withValuesSeparatedBy(',')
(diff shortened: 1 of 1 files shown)


common/src/main/java/bisq/common/config/Config.java
```diff
@@ -373,7 +373,7 @@ public Config(String defaultAppName, File defaultUserDataDir, String... args) {
                         .defaultsTo(false);
 
         ArgumentAcceptingOptionSpec<String> providersOpt =
-                parser.accepts(PROVIDERS, "List custom providers")
+                parser.accepts(PROVIDERS, "List custom price providers")
```
cbeams: I've always wanted to rename this option. Improving the description is a fine start.

```suggestion
                parser.accepts(PROVIDERS, "List custom pricenodes")
```

cbeams (changes requested):
Nit: "price provider" is old terminology, we're using pricenode pretty consistently everywhere now.

cbeams (approved):

PR #4015, opened 2020-02-27
Remove unused uid in fee provider service

freimair:
Bisq frequently (once per minute) queries our price nodes for up-to-date
fee information. It does so by HTTP GET request. However, it provided
a UID via the "User-Agent" HTTP header field. This UID has been a random
number which changed everytime Bisq gets started up.

This UID has never been used. Thus, remove it.

We had a similar case with the price-feed service. Now, pricenode logs should not contain any uid info anymore.

(this has been pointed out in https://github.com/bisq-network/bisq/pull/3937)


sqrrm (approved):
utACK

PR #4016, opened 2020-02-28
Fix nginx.conf for collectd so it only listens on 127.0.0.1

wiz:
Before:
```
tcp        0      0 0.0.0.0:2003            0.0.0.0:*               LISTEN     
```

After:
```
tcp        0      0 127.0.0.1:2003          0.0.0.0:*               LISTEN     
```


sqrrm (approved):
utACK

PR #3758, opened 2019-12-06
Vote Reveal - Silent Fail

niyid:
Fixes #3408

The log below shows where the vote reveal failed silently as a result of insufficient wallet balance.
```
Oct-15 06:43:21.835 [JavaFX Application Thread] INFO  org.bitcoinj.wallet.Wallet: Completing send tx with 1 outputs totalling 0.00000596 BTC and a fee of 0.00 BTC/kB 
Oct-15 06:43:21.837 [JavaFX Application Thread] WARN  org.bitcoinj.wallet.Wallet: Insufficient value in wallet for send: needed 0.00001606 BTC more 
Oct-15 06:43:21.845 [JavaFX Application Thread] ERROR b.c.d.g.v.VoteRevealService: VoteRevealException{
     voteRevealTx=null,
     blindVoteTxId='3cf435e275b8e1518ea969fe90e0504bbed1d68135310b3c24546244fca90b5d',
     myVote=null
} bisq.core.dao.governance.votereveal.VoteRevealException: Exception at calling revealVote. 
Oct-15 06:43:24.448 [JavaFX Application Thread] INFO  b.c.d.m.DaoStateMonitoringService: Passed checkpoint Checkpoint {
     height=586920,
     hash=523aaad4e760f6ac6196fec1b3ec9a2f42e5b272
} 
```

PR #4017, opened 2020-02-28
Revert "Vote Reveal - Silent Fail"

ripcurlx:
Based on our agreement that @ManfredKarrer needs to ACK all DAO related changes I'm reverting this PR because of https://github.com/bisq-network/bisq/pull/3758#issuecomment-592305052.


sqrrm (approved):
utACK

Diff is the same I get doing a revert.

PR #4018, opened 2020-02-28
Fix false deletion of hsdir when using --appDataDir

### LINKED ISSUE #3995: Bisq does not update onion address of offers after deleting Tor data
### Description

If a user with open offers presses the "Delete outdated Tor files and restart" button, their offers are not updated to the node's new onion address, so they cannot be taken by anyone

#### Version

v1.2.7

### Steps to reproduce

1) Create an offer
2) Note the onion address of your offer
3) Settings -> Network Info -> Open Tor Settings -> "Delete outdated Tor files and restart"
4) After restarting, note the onion address of your offer again
5) Create a 2nd offer
6) Note the onion address of your 2nd offer
7) Attempt to take your 1st offer from another node
8) Attempt to take your 2nd offer from another node

### Expected behaviour

1) The onion address returned in step 4 should be updated to your node's new onion address after resetting Tor, so it should not be the same as the onion address returned in step 2
2) The onion addresses of all your offers should be the same
3) All of your offers should be taken by the other node

### Actual behaviour

1) The onion addresses returned in steps 2 and 4 are the same
2) The onion addresses of 1st offer and 2nd offer are different
3) The 1st offer can not be taken

freimair:
Probably fixes https://github.com/bisq-network/bisq/issues/3995

There are circumstances where input via --appDataDir
will lead to the hiddenservice-directory to be deleted.
Successfully reproduced using

  --appDataDir=~/foo

Although the "~" does not get interpreted correctly on
my linux system, it does manage to throw off the
mechanics of sparing the hiddenservice-directory from
being deleted on startup.

### DIFF
--- a/common/src/main/java/bisq/common/storage/FileUtil.java
+++ b/common/src/main/java/bisq/common/storage/FileUtil.java
@@ -108,14 +108,14 @@ public class FileUtil {
             File[] files = file.listFiles();
             if (files != null)
                 for (File f : files) {
-                    if (!excludeFileFound)
-                        excludeFileFound = f.equals(exclude);
-                    if (!f.equals(exclude))
+                    boolean excludeFileFoundLocal = exclude != null ? f.getAbsolutePath().equals(exclude.getAbsolutePath()) : false;
+                    excludeFileFound |= excludeFileFoundLocal;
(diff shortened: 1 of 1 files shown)


sqrrm (approved):
utACK

PR #4023, opened 2020-03-03
Update title of new trade protocol warning

rafaelpac:
In fresh installs of Bisq, there is a warning about the changes that happened
to the trade protocol in v1.2.

image

But since we are already in v1.7, it might look confusing to a newbie.
He is installing v1.7 and get a message of v1.2.
This fix adds the word "Since", which clarifies it.


rafaelpac:
Sorry, it is actually 1.2.7, so message is still ok.

PR #4024, opened 2020-03-03
Changes to local BTC node reminder in Settings > Network info

dmos62:
Under "Settings" > "Network info" we have a list of connected Bitcoin nodes. Under this list we provide a reminder in small font about Bisq being able to use a local Bitcoin node. This makes changes to that reminder:

1) the text is changed to be more accurate;
2) the condition under which the reminder used to be disabled is removed (because it doesn't seem to be necessary);
3) the formatting is slighty changed by dropping the parantheses: the paratheses serve to visually show that the reminder is background information, but that is already achieved by using small print.

Screenshot of reminder after change:

small letters - local btc usage explanation


sqrrm (approved):
utACK

Issue #4003, opened 2020-02-24
High system load on Arch linux


freimair:
### Description

One of my two always-on bisq apps is sitting at 100% CPU.

#### Version

1.2.7

### Steps to reproduce

nothing special. start it up.

### Expected behaviour

not have that high of a CPU load

### Actual behaviour

Profiling the thing shows:
33.4% - bisq.common.storage.FileManager.saveToFile
21.7% - bisq.network.http.HttpClient.requestWithGETProxy
9.5% - com.google.protobuf.GeneratedMessageV3.parseDelimitedWithIOException

#### Device or machine

my dev machine, arch linux

stejbac:
I believe the biggest bottleneck in `saveToFile` (at least from what I've seen using sampling mode in JProfiler) is the `FileDescriptor::sync` call to try to force a write to disc, rather than protobuf conversion during persistence. I'm not sure if that's actual CPU usage though.

freimair:
I just found that Bisq writes 2GB of data to disk in just 2h of uptime.

I strongly recommend deciding on https://github.com/bisq-network/projects/issues/25 just to address this before it becomes an issue.

tenaciousRas:
CPU load is WAY too high for this application. Does this issue only affect arch linux, and not all platforms?

freimair:
just stumbled over that again and took the chance to take another snapshot:

![Screenshot from 2020-04-01 10-37-51](https://user-images.githubusercontent.com/1070734/78116814-1f7a7900-7405-11ea-8f41-45e8e20229e5.png)

this snapshot was taken 20 minutes after the dao sync process received its last block.

freimair:
@tenaciousRas after some investigating, I believe that is not only for arch linux but a general issue. However, I do have no means to verify that...

tenaciousRas:
@freimair Sweet snapshot - same here I cannot verify on Windows or Mac. My question is somewhat rhetorical - it probably isn't confined to *nix. Could be related to #3752. I haven't checked out the code, yet. Talk of thread-safe serialization within a FileManager class doesn't sound proper. "Create protobuf objects on the UserThread instead of the FileManager writer thread" --- sounds proper to me and points to something lacking in architecture.

freimair:
@stejbac what is your take on this? could your [fix](https://github.com/bisq-network/bisq/pull/4025) on #3752 enabled this/let this shine through/already fixed it?

freimair:
stejbac:
[...]

PR #4025, opened 2020-03-04
Make serialisation in FileManager::saveToFile thread-safe

### LINKED ISSUE #3752: ConcurrentModificationException FileManager::saveToFile
### Description
`BallotList` is not thread-safe, but is used by the `FileManager` to persist to disk. All users that extend
`PersistableList` can have the same race condition.

Regression: No

```
Dec-05 13:50:46.954 [Save-file-task-9892] ERROR b.common.storage.FileManager: Error at saveToFile, storageFile=/home/julian/.local/share/Bisq/btc_mainnet/db/BallotList java.lang.RuntimeException: java.util.ConcurrentModificationException
	at bisq.common.storage.FileManager.saveToFile(FileManager.java:200)
	at bisq.common.storage.FileManager.lambda$new$0(FileManager.java:83)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.util.ConcurrentModificationException: null
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1660)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
	at bisq.core.dao.state.model.governance.BallotList.getBuilder(BallotList.java:62)
	at bisq.core.dao.state.model.governance.BallotList.toProtoMessage(BallotList.java:55)
	at bisq.core.dao.state.model.governance.BallotList.toProtoMessage(BallotList.java:36)
	at bisq.common.storage.FileManager.saveToFile(FileManager.java:194)
	... 6 common frames omitted
```

#### Version

v1.2.4
v1.2.3 as well

### Steps to reproduce
Only seen it once when trying to reproduce

1. Clear local appData
2. `./bisk-desktop` 
3. See stack trace in logs

See comments for patch that can be applied to v1.2.3 to cause bug.

### Expected behaviour

All data structures used with FileManager are thread-safe

### Actual behaviour

Some data structures used with FileManager are not thread-safe

### Screenshots

saveToFileError

#### Device or machine

Ubuntu 19.10

#### Additional info

[bisq.log](https://github.com/bisq-network/bisq/files/3929287/bisq.log)

stejbac:
Add `toProtoMessageSynchronized()` default method to `PersistableEnvelope`, which performs (blocking) protobuf serialisation in the user thread, regardless of the calling thread. This should prevent data races like the `ConcurrentModificationException` observed in #3752, under the reasonable assumption that shared persistable objects are only mutated in the user thread.

Also add a `ThreadedPersistableEnvelope` sub-interface overriding the default method above, to let objects which are expensive to serialise (like `DaoStateStore`) be selectively serialised in the `save-file-task-X` thread as before, but directly synchronised with each mutating operation. As most objects are cheap to serialise, this avoids a noticeable performance drop without having to track down every mutating method for each store.

In all cases but one, classes implementing `ThreadedPersistableEnvelope` are stores like `TradeStatistic2Store`, with a single `ConcurrentHashMap` field. These require no further synchronisation, since the map entries are immutable, so the only mutating operations are `map.put(..)` calls which are already synchronised with map reads. (Even if `map.values().stream()` sees updates at different keys happen out-of-order, it should be benign, e.g. trade statistics might be persisted out-of-order if two trades come along at almost exactly the same time.)

The remaining case is `DaoStateStore`, which is only ever reset or modified via a single `persist(..)` call with a cloned `DaoState` instance and hash chain from `DaoStateSnapshotService`, so there is no aliasing risk from the various DAO state mutations done in `DaoStateService` and elsewhere.

This should fix #3752.

--

Adding some temporary benchmark logging to `PersistableEnvelope` shows the time spent in each `saveToFile` protobuf conversion forced to run on the user thread with the PR changes, from my Bisq node (running on mainnet) from startup to shutdown:

[Serialisation_Benchmarks.txt](https://github.com/bisq-network/bisq/files/4286961/Serialisation_Benchmarks.txt)

As can be seen in the filtered log, the serialisation of `PreferencesPayload`, `UserPayload` and `TradableList` took 100-140ms (which might cause noticeable jitter on the user thread) the first time they were persisted upon application startup. However, subsequent serialisations of the same objects seem to occur much faster (presumably after the protobuf code has had a chance to warm up), with no (single-threaded) `PersistableEnvelope` object taking longer than around 20ms to serialise once the application has started.

The assumed expensive-to-serialise objects like `DaoStateStore` and those managed by `MapStoreService` all implement `ThreadedPersistableEnvelope`, so don't appear in the above benchmark and are converted directly on each `save-file-task-X` thread as before.

### DIFF
--- a/common/src/main/java/bisq/common/proto/persistable/PersistableEnvelope.java
+++ b/common/src/main/java/bisq/common/proto/persistable/PersistableEnvelope.java
@@ -31,7 +31,7 @@ import java.util.concurrent.FutureTask;
  */
 public interface PersistableEnvelope extends Envelope {
 
-    default Message toProtoMessageSynchronized() {
+    default Message toPersistableMessage() {
         FutureTask<Message> toProtoOnUserThread = new FutureTask<>(this::toProtoMessage);
         UserThread.execute(toProtoOnUserThread);
         //noinspection UnstableApiUsage
--- a/common/src/main/java/bisq/common/proto/persistable/PersistableEnvelope.java
+++ b/common/src/main/java/bisq/common/proto/persistable/PersistableEnvelope.java
@@ -18,13 +18,8 @@
 package bisq.common.proto.persistable;
 
 import bisq.common.Envelope;
-import bisq.common.UserThread;
 
 import com.google.protobuf.Message;
-
-import com.google.common.util.concurrent.Futures;
-
-import java.util.concurrent.FutureTask;
 
 /**
  * Interface for the outside envelope object persisted to disk.
@@ -32,9 +27,6 @@ import java.util.concurrent.FutureTask;
 public interface PersistableEnvelope extends Envelope {
 
     default Message toPersistableMessage() {
-        FutureTask<Message> toProtoOnUserThread = new FutureTask<>(this::toProtoMessage);
-        UserThread.execute(toProtoOnUserThread);
-        //noinspection UnstableApiUsage
-        return Futures.getUnchecked(toProtoOnUserThread);
+        return toProtoMessage();
     }
 }
--- a/common/src/main/java/bisq/common/proto/persistable/NavigationPath.java
+++ b/common/src/main/java/bisq/common/proto/persistable/NavigationPath.java
@@ -45,7 +45,7 @@ public class NavigationPath implements PersistableEnvelope {
         return protobuf.PersistableEnvelope.newBuilder().setNavigationPath(builder).build();
     }
 
-    public static PersistableEnvelope fromProto(protobuf.NavigationPath proto) {
+    public static NavigationPath fromProto(protobuf.NavigationPath proto) {
         return new NavigationPath(new ArrayList<>(proto.getPathList()));
     }
 }
--- a/common/src/main/java/bisq/common/proto/persistable/NavigationPath.java
+++ b/common/src/main/java/bisq/common/proto/persistable/NavigationPath.java
@@ -21,7 +21,6 @@ import bisq.common.util.CollectionUtils;
 
 import com.google.protobuf.Message;
 
-import java.util.ArrayList;
 import java.util.List;
 
 import lombok.AllArgsConstructor;
@@ -36,7 +35,7 @@ import lombok.Setter;
 @Getter
 @Setter
 public class NavigationPath implements PersistableEnvelope {
-    private List<String> path = new ArrayList<>();
+    private List<String> path = List.of();
 
     @Override
     public Message toProtoMessage() {
@@ -46,6 +45,6 @@ public class NavigationPath implements PersistableEnvelope {
     }
 
     public static NavigationPath fromProto(protobuf.NavigationPath proto) {
-        return new NavigationPath(new ArrayList<>(proto.getPathList()));
+        return new NavigationPath(List.copyOf(proto.getPathList()));
     }
 }
--- a/common/src/main/java/bisq/common/proto/persistable/PersistableList.java
+++ b/common/src/main/java/bisq/common/proto/persistable/PersistableList.java
@@ -31,7 +31,7 @@ import lombok.Setter;
 import lombok.experimental.Delegate;
 
 @EqualsAndHashCode
-public class PersistableList<T extends PersistablePayload> implements PersistableEnvelope, Iterable<T> {
+public class PersistableList<T extends PersistablePayload> implements UserThreadMappedPersistableEnvelope, Iterable<T> {
     @Delegate(excludes = ExcludesDelegateMethods.class)
     @Getter
     @Setter
--- a/common/src/main/java/bisq/common/proto/persistable/PersistableList.java
+++ b/common/src/main/java/bisq/common/proto/persistable/PersistableList.java
@@ -31,7 +31,7 @@ import lombok.Setter;
 import lombok.experimental.Delegate;
 
 @EqualsAndHashCode
-public class PersistableList<T extends PersistablePayload> implements UserThreadMappedPersistableEnvelope, Iterable<T> {
+public class PersistableList<T extends PersistablePayload> implements PersistableEnvelope, Iterable<T> {
     @Delegate(excludes = ExcludesDelegateMethods.class)
     @Getter
     @Setter
--- a/common/src/main/java/bisq/common/proto/persistable/ThreadedPersistableEnvelope.java
+++ b/common/src/main/java/bisq/common/proto/persistable/ThreadedPersistableEnvelope.java
@@ -17,24 +17,28 @@
 
 package bisq.common.proto.persistable;
 
-import bisq.common.Envelope;
-import bisq.common.UserThread;
-
 import com.google.protobuf.Message;
 
-import com.google.common.util.concurrent.Futures;
-
-import java.util.concurrent.FutureTask;
-
 /**
- * Interface for the outside envelope object persisted to disk.
+ * Interface for the outer envelope object persisted to disk, where its serialization
+ * during persistence takes place on a separate thread (for performance).
+ * <p>
+ * To make the serialization thread-safe, all modifications of the object must be
+ * synchronized with it. This may be achieved by wrapping such modifications with the
+ * provided {@link ThreadedPersistableEnvelope#modifySynchronized(Runnable)} method.
  */
-public interface PersistableEnvelope extends Envelope {
+public interface ThreadedPersistableEnvelope extends PersistableEnvelope {
 
+    @Override
     default Message toProtoMessageSynchronized() {
-        FutureTask<Message> toProtoOnUserThread = new FutureTask<>(this::toProtoMessage);
-        UserThread.execute(toProtoOnUserThread);
-        //noinspection UnstableApiUsage
-        return Futures.getUnchecked(toProtoOnUserThread);
+        synchronized (this) {
+            return toProtoMessage();
+        }
+    }
+
+    default void modifySynchronized(Runnable modifyTask) {
+        synchronized (this) {
+            modifyTask.run();
+        }
     }
 }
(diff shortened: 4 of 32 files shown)


freimair:
the part about performance sounds awfully familiar to me. see https://github.com/bisq-network/bisq/issues/4003

sqrrm:
I would like to get this PR into the next release if possible. It touches core DAO code so @ManfredKarrer needs to review it. I think it looks good from what I've seen so far.

common/src/main/java/bisq/common/proto/persistable/PersistableEnvelope.java
```diff
@@ -18,9 +18,23 @@
...
+import com.google.common.util.concurrent.Futures;
+
+import java.util.concurrent.FutureTask;
 
 /**
  * Interface for the outside envelope object persisted to disk.
  */
 public interface PersistableEnvelope extends Envelope {
+
+    default Message toProtoMessageSynchronized() {
```
ManfredKarrer: I find the method name a bit confusing as it is only synchronized in `ThreadedPersistableEnvelope` but here we use mapping to userThread. I would recommend using `toPersistableMessage` as stated above.

common/src/main/java/bisq/common/proto/persistable/PersistableEnvelope.java
```diff
@@ -18,9 +18,23 @@
...
+
+import com.google.common.util.concurrent.Futures;
+
+import java.util.concurrent.FutureTask;
 
 /**
  * Interface for the outside envelope object persisted to disk.
  */
 public interface PersistableEnvelope extends Envelope {
+
```
ManfredKarrer, line 28: Maybe its better to use a specific interface here as well like in `ThreadedPersistableEnvelope`? Might make it more explicit that we alter the basic non-thread-safe but fast behaviour. We could use then `toPersistableMessage` as method name for all 3 cases and the interface marks which style is actually used. Would make it also a bit more distinct from the generic toProtoMessage method and shows its only relevant for persisting an envelope.

ManfredKarrer (commented):
utACK - but would prefer to see the suggested changes and want to emphasise that it has to be tested very well. Its certainly a high risk PR. But great to see some progress in that area! Thanks @stejbac !

I would suggest renaming `toProtoMessageSynchronized` to `toPersistableMessage`.
Also would suggest to make it more explicit and flexible by adding the new behaviour in a `UserThreadMappedPersistableEnvelope` interface and keep PersistableEnvelope as non-safe-thread but fast interface with the default method:
`
 default Message toPersistableMessage() {
        return toProtoMessage();
    }
`

If this is applied it has to be carefully checked for each persisted object which interface is most suitable and check with profiling how the changes compare. There are some generic classes like `PersistableList` which need to be considered as well (at the moment its all covered as `PersistableEnvelope` is used for the new method. 

There are several classes implementing `PersistableEnvelope` like `SignedWitness` but I think that is wrong as they are not persisted directly but wrapped in a storage class. I have not checked more closely but might be worth to clean that up in a separate PR.

I am not sure about the performance impacts when using UserThread mapping. Any delay > 20 ms is visually visible in the UI as lagging. 

I would suggest to create 3 categories for all persisted objects:
- Big objects -> performance is critical, use threaded version (DaoState,...)
- Small objects, correctness is critical -> use UserThread mapping (SequenceNrMap,...)
- Other uncritical -> use non-thread safe option in dedicated thread (NavigationPath)

The uncritical ones could be kept as non-threadsafe IMO. 

For further improvements the delays between write operations should be checked, there might be some optimisation possible, specially for those which have frequent write operations. 

From the DAO perspective I don't see any obvious risk with that change, but it should be tested very well specially with DaoStateHashes and the monitoring for the 3 different hash chains.

For the most problematic objects (DaoState, TradeStatistics, AccountAgeWitness) I suggest to start to look into a different design as suggested in the related issues/PRs (e.g. use a historical data structure which is not handled by the P2PDataStorage and only loaded in memory on demand). For DaoState that approach might not work, but some solution to make it more scalable is required here as well.

stejbac:
I've just pushed changes to address some of the issues pointed out above. I renamed `toProtoMessageSynchronized` and added `UserThreadMappedPersistableEnvelope` as suggested. There was another commit I intended to push to serialise `SequenceNumberMap` on the user thread, by making it implement the new interface instead of `ThreadedPersistableEnvelope`. However, benchmarking the code in the same way as above showed that it adds considerable overhead:

[Serialisation_Benchmarks_with_SequenceNumberMap.txt](https://github.com/bisq-network/bisq/files/4311586/Serialisation_Benchmarks_with_SequenceNumberMap.txt)

I think it would be better to make it continue to implement `ThreadedPersistableEnvelope` and rely on the concurrency guarantees of `ConcurrentHashMap` if possible. (That should effectively synchronise reads and updates of the optional map value, at each potential key independently.)

ManfredKarrer:
@stejbac Thanks for the implementation of the suggested changes!

Could we make a list of all potential persisted objects (some are only created in rare cases like FailedTrades) to check for each persisted object if the adequate interface is applied? Would be also great to see a performance comparison before/after change (hope I don't get too annoying with those requests but I think as its a bit of a risky change its better to be extra careful here). 

I just found another issue (no actual problem but potential risk for future changes):
`VoteWithProposalTxIdList` (as well as `MeritList`) is extending `PersistableList` which is now using the `UserThreadMappedPersistableEnvelope`. This list is using the PersistableList only for serialisation but not persistence (see comment in class). This should be fixed anyway in a more correct way (e.g. using a dedicated List only for that purpose). As we don't call the `toPersistableMessage` method here no behaviour change is introduced, but if anyone would start further changes to `PersistableList` we might inherit a potential critical change, which can easily lead then to a consensus problem (the serialised protoBuf data are used for consensus). I suggest to clean that up either with the PR (or better with a new one) so that this risk by abusing the `PersistableList` here is removed. A potential problematic change could be if one changes the list in `PersistableList` from `ArrayList` to any other list type which might produce a different protoBuf serialization result. As it is not obvious that `PersistableList` is used for consensus critical code here this should be changed to remove that risk.

stejbac:
Earlier, I (manually) worked out a type hierarchy of everything extending / implementing `PersistableEnvelope` after the proposed changes:

Screenshot from 2020-03-11 07-23-24
(Actuallty `PersistableList` isn't an abstract class, but probably could be.)

I was intending to clean up the classes incorrectly implementing `PersistableEnvelope` in a separate PR, as suggested.

It looks like every `toProtoMessage` and `fromProto` method of each class in the DAO package extending `PersistableList` collects items into a new list using a stream, anyway, when mapping the input list items to/from _proto.*_ objects. So I would have thought that only the iteration order of the `PersistableList.getList()` items matters for proto conversion.

ManfredKarrer:
Thanks for the list. 
Regarding `VoteWithProposalTxIdList` and `MeritList`: Yes I also assume there is no issue but I think its more safe to use a dedicated class to avoid potential future issues. Those classes implement also the `ConsensusCritical` marker interface. 

Also the `PersistableNetworkPayloadList` and its related classes (`PersistableNetworkPayloadListService`) could be cleaned up and removed (in separate PR). It was long ago used for converting db files. 
`PersistableHashMap` is also not used.

Thanks again for working on that!

ManfredKarrer:
Tx ID: d90a4cdf22719e0475b14d935156f796cb28ed8f113f0933be9bd43ffdaf685c

ripcurlx:
> Regarding `VoteWithProposalTxIdList` and `MeritList`: Yes I also assume there is no issue but I think its more safe to use a dedicated class to avoid potential future issues. Those classes implement also the `ConsensusCritical` marker interface.

@ManfredKarrer Is this a requirement for this PR to get merged? If not please NACK or if it is up to be merged ACK from your side to make it clear for the maintainers 😄 . Thanks!

ManfredKarrer:
@ripcurlx Thanks for reminding on that. I recommend to not change the behaviour of those 2 classes (not using `UserThreadMappedPersistableEnvelope` as they would do now from the base class). So until that is changed I need to NACK that PR. Beside that utACK.

stejbac:
I've just made another commit to remove the `UserThreadMappedPersistableEnvelope` interface from the two classes, as recommended by @ManfredKarrer. After the change, the interface still lies on all the remaining classes extending `PersistableList`, where I believe it's actually needed (except possibly `PaymentAccountList`, which appears to only have a hypothetical race setting `paymentAccount.selectedTradeCurrency` when using the non-thread-safe persistence interface).

This does, however, include three other `ConsensusCritical` classes, changing the original behaviour (to fix the bug): `MyProposalList`, `MyBlindVoteList` and `BallotList`.

ripcurlx (approved):
ACK - based on https://github.com/bisq-network/bisq/pull/4025#issuecomment-601981705

Tested it myself on Regtest (Arbitration, Voting Cycle, Trade, Account Creation) and I didn't see any issues so far.

PR #4031, opened 2020-03-06
Size the offer book on window activation

### LINKED ISSUE #4030: Offer Book screen does not reflow contents as expected
### Description
Sometimes when you switch to Market -> Offer Book, the list of open offers are displayed with only 5 rows visible leaving a large empty space below.  Once the window is manually resized the offer tables are adjusted to fill the available empty space.

#### Version
1.2.7

### Steps to reproduce
Navigate to any screen other than 'Market', for example 'Buy BTC'.
Close Bisq.
Re-start Bisq.
Enlarge the window in the vertical dimension (or maximize the window).
Navigate to the Market screen.

### Expected behaviour
The Market Offer Book tables should occupy the available empty space.

### Actual behaviour
The Market Offer Book tables are shown with only 5 rows, and an empty space below them (see screenshot below).

### Screenshots
This is how it looks when you first open the screen:
image

This is how it should look:
image

#### Device or machine
Linux

#### Additional info
Similar to https://github.com/bisq-network/bisq/issues/1587 which had been marked as fixed.

ghost:
Fixes #4030 

Currently the offer book tables are only being sized when the window proportions change.  However if the window is made bigger before the offer book is opened, then the tables are not resized from their default.

Other windows (for example Market -> Trades) solve this by make a sizing call in the `activate` method.    In order to do this the sizing code is separated into a `layout` method where it can be called from both `activate` and the existing `heightProperty` listener.  The same approach is taken here.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java
+++ b/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java
@@ -42,6 +42,7 @@ import bisq.core.util.coin.CoinFormatter;
 
 import bisq.network.p2p.NodeAddress;
 
+import bisq.common.UserThread;
 import bisq.common.config.Config;
 import bisq.common.util.Tuple3;
 import bisq.common.util.Tuple4;
@@ -91,6 +92,7 @@ import javafx.util.StringConverter;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 
 import static bisq.desktop.util.FormBuilder.addTopLabelAutocompleteComboBox;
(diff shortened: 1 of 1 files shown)


ripcurlx:
@jmacxx Thanks for your first PR! I'll review your PR now, but please adapt and force push your commit message based on https://github.com/bisq-network/style/issues/9. I don't want to be picky, but I think it is good to follow this guideline for your future contributions. Thanks!

ripcurlx (changes requested):
NACK

ACK from functionality and code changes. As soon as you have updated the commit message https://github.com/bisq-network/bisq/pull/4031#issuecomment-60011947 I'm happy to merge your PR.

ripcurlx (dismissed):

ripcurlx:
@jmacxx Please sign your commit. Thanks!

ripcurlx (approved):
ACK

Tested it on Regtest.

PR #4032, opened 2020-03-06
Pricenode crashes on startup due to CoinMarketCap retired API

wiz:
### Background
CoinMarketCap isn't actively used as a pricenode provider, because the only coins that used it have been delisted.

### Issue
CoinMarketCap has retired their v1 API, and the resulting `HTTP 410 Gone` causes the pricenode to crash on startup

### Summary
This PR simply removes CoinMarketCap as a PriceNodeProvider since it's apparently no longer used and the API has been retired


sqrrm:
From reading the comments on how the Providers are instantiated, this looks fine. I don't understand exactly how that's done though, would really prefer the review of @cbeams

wiz:
Again please note that this PR is an emergency measure since currently pricenodes crash on startup. Obviously we need to consider implementing one of the new CMC authenticated / paid APIs to replace the API that has been retired

cbeams (approved):
utACK, see my additional review commit.

PR #4035, opened 2020-03-09
Speed up DAO state monitor view load

stejbac:
Compute height cells lazily to speed up `DaoStateMonitorView`

Avoid a bottleneck computing the cycle index & calling `Res.get(..)` for every block since genesis in the DAO state monitor view, when building the `DaoStateBlockListItem` objects, by making the `height` field lazy. To do this, pass the cycle index into the constructor using an `IntSupplier` and make the height a memoised `Supplier<String>` with a custom getter.

Also add a unit test to check that the auto-generated `equals` & `hashCode` methods still work as expected, as it isn't totally clear what Lombok would do when a field type differs from its getter return type.

--

Flippling back and forth between the _Network monitor_ tab and the _Governance_ tab revealed the following call tree in JProfiler:

Screenshot from 2020-03-06 10-16-21

As can be seen above, most of the time is spent in the stream pipeline collecting new `DaoStateBlockListItem` objects, in `DaoStateMonitorView.onDataUpdate`. After the change, the main bottleneck appeared to be from internal JavaFX code, in the `listItems.setAll` call in `onDataUpdate` (perhaps from one of the listeners on `DaoStateMonitorView.sortedList`). However, I believe there was a noticeable speedup. Also, the slowdown over time is likely to be quadratic without the change, as the number of cycles and block grows linearly. (The main bottleneck appears to be computing the cycle index of each block.)

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/MainViewModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/MainViewModel.java
@@ -133,7 +133,7 @@ public class MainViewModel implements ViewModel, BisqSetup.BisqSetupListener {
     private Timer checkNumberOfP2pNetworkPeersTimer;
     @SuppressWarnings("FieldCanBeLocal")
     private MonadicBinding<Boolean> tradesAndUIReady;
-    private Queue<Overlay> popupQueue = new PriorityQueue<>(Comparator.comparing(Overlay::getDisplayOrderPriority));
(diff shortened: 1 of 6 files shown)


ripcurlx:
Waiting for ACK by @ManfredKarrer as it touches the DAO package

ManfredKarrer:
@ripcurlx It is only touching a view class, so I do not need to ACK that (only core dao package). Had a quick look and looks good to me, but don't consider it a review....

ripcurlx (approved):
ACK

Tested it on Mainnet and everything works as before. Code looks fine.

PR #4040, opened 2020-03-10
Log delayed payout tx bytes as soon as it's signed during trade init

sqrrm:
Log the delayed payout tx bytes before deposittx is broadcast as a desperate measure to at least have it logged at some point.


ripcurlx (approved):
utACK

PR #4041, opened 2020-03-10
Updated docker install of Pricenode

mrosseel:
- update of the docker installation for its new location (main bisq repo)
- adopted the builder/runner docker pattern to limit image size

PS: I'm aware there's a debian install script, but I want at least one guy running something else than debian + that script in case there are bugs/vulnerabilities specific to that combination of OS and install method.


ripcurlx:
@mrosseel Please have a look at the Codacy issues: https://app.codacy.com/gh/bisq-network/bisq/pullRequest?prid=5153686. Thanks!

mrosseel:
The pricenode scripts have been updated with collectd reporting, and since they heavily use systemd/journalctl which are hard to do in docker, I had to abandon the pricenode docker approach. It's possible to convert but too much work atm. 
These new scripts are easy to use so that's fine, only possible downside is that all pricenodes are running on debian (just a guess) and are one bad update/zero day away from all going down at once. If this becomes a priority ops issue I can look at it again, until then I'm going with the flow ;)

PR #4043, opened 2020-03-10
Add an empty dummy exchange rate to prevent warning logs

ripcurlx:


wiz (approved):
ACK
tested on http://gztmprecgqjq64zh.onion/getAllMarketPrices and no more errors in logs

sqrrm (approved):
utACK

PR #4045, opened 2020-03-11
Allow IPv6 connections to Bitcoin nodes

### LINKED ISSUE #3990: Configuring custom Bitcoin node in GUI settings by IPv6 address fails
### Description

The GUI settings fail to configure btcnodes by IPv6 address. However, passing an IPv6 address for the btcnode works fine when passed as a command line option.

#### Version

v1.2.7

### Steps to reproduce

1. Settings
2. Network Info
3. Uncheck "Use Tor for Bitcoin network"
4. Check "Use custom Bitcoin nodes"
5. Enter "[2401:b140::42:100]:8333" in the "Custom Bitcoin Nodes" field
6. Restart Bisq

### Expected behaviour

Bisq should connect to the configured Bitcoin node over IPv6

### Actual behaviour

Bisq loads random Bitcoin nodes from the public network

#### Additional info

Setting Bitcoin nodes by IPv6 works fine using command line option `--btcNodes=[2401:b140::42:100]:8333`

ghost:
Fixes #3990

Currently bisq desktop does not accept IPv6 addresses in the settings for custom nodes or via the `--btcNodes` command line option.  
While the address format is already validated in `desktop / GUIUtil`, the separation of address and port is handled incorrectly in `core / BtcNodes::fromFullAddress`.
This results in IPv6 addresses being ignored, regardless of whether they are entered via the GUI or command line.

This change fixes the code in `BtcNodes::fromFullAddress` to identify an IPv6 address as being surrounded by square brackets.

- IPv6 addresses are a sequence of hexadecimal and ':' characters enclosed by square brackets, followed by an optional colon and port number
- Onion addresses are a sequence of alphanumeric characters followed by `.onion` and an optional colon and port number
- IPv4 addresses are a sequence of numeric and '.' characters followed by an optional colon and port number

**Testing**:
the following use case tests were identified and checked:

- onion address no port
- onion address with port
- IPv4 address no port
- IPv4 address with port
- IPv6 address no port
- IPv6 address with port
- combination of all the above types
- invalid input

### DIFF
--- a/core/src/main/java/bisq/core/btc/nodes/BtcNodes.java
+++ b/core/src/main/java/bisq/core/btc/nodes/BtcNodes.java
@@ -113,13 +113,32 @@ public class BtcNodes {
          * @return BtcNode instance
          */
         public static BtcNode fromFullAddress(String fullAddress) {
-            String[] parts = fullAddress.split(":");
+            String[] parts = fullAddress.split("]");
             checkArgument(parts.length > 0);
-            final String host = parts[0];
+            String host = "";
             int port = DEFAULT_PORT;
-            if (parts.length == 2)
-                port = Integer.valueOf(parts[1]);
+            if (parts[0].contains("[") && parts[0].contains(":")) {
+                // IPv6 address and optional port number
+                // address part delimited by square brackets e.g. [2a01:123:456:789::2]:8333
+                host = parts[0].replace("[", "").replace("]", "");
+                if (parts.length == 2)
+                    port = Integer.parseInt(parts[1].replace(":", ""));
+            }
+            else if (parts[0].contains(":") && !parts[0].contains(".")) {
+                // IPv6 address only; not delimited by square brackets
+                host = parts[0];
+            }
+            else if (parts[0].contains(".")) {
+                // address and an optional port number
+                // e.g. 127.0.0.1:8333 or abcdef123xyz.onion:9999
+                parts = fullAddress.split(":");
+                checkArgument(parts.length > 0);
+                host = parts[0];
+                if (parts.length == 2)
+                    port = Integer.parseInt(parts[1]);
+            }
 
+            checkArgument(host.length()>0, "BtcNode address format not recognised");
             return host.contains(".onion") ? new BtcNode(null, host, null, port, null) : new BtcNode(null, null, host, port, null);
         }
 
--- a/p2p/src/main/java/bisq/network/DnsLookupTor.java
+++ b/p2p/src/main/java/bisq/network/DnsLookupTor.java
@@ -84,10 +84,10 @@ public class DnsLookupTor {
 
             byte[] hostBytes = host.getBytes(Charsets.UTF_8);
             buf = new byte[7 + hostBytes.length];
-            buf[0] = b('\u0005');
-            buf[1] = b('\u00f0');
-            buf[2] = b('\u0000');
-            buf[3] = b('\u0003');
+            buf[0] = b('\u0005');               // version SOCKS5
+            buf[1] = b('\u00f0');               // CMD_RESOLVE
+            buf[2] = b('\u0000');               // (reserved)
+            buf[3] = b('\u0003');               // SOCKS5_ATYPE_HOSTNAME
             buf[4] = (byte) hostBytes.length;
             System.arraycopy(hostBytes, 0, buf, 5, hostBytes.length);
             buf[5 + hostBytes.length] = 0;
@@ -113,13 +113,16 @@ public class DnsLookupTor {
                 throw new DnsLookupException(torStatusErrors.get(buf[1]) + "(host=" + host + ")");
             }
 
-            if (buf[3] != b('\u0001'))
+            final char SOCKS5_ATYPE_IPV4 = '\u0001';
+            final char SOCKS5_ATYPE_IPV6 = '\u0004';
+            final byte atype = buf[3];
+            if (atype != b(SOCKS5_ATYPE_IPV4) && atype != b(SOCKS5_ATYPE_IPV6))
                 throw new DnsLookupException(torStatusErrors.get(b('\u0001')) + "(host=" + host + ")");
 
-            buf = new byte[4];
+            final int octets = (atype == SOCKS5_ATYPE_IPV4 ? 4 : 16);
+            buf = new byte[octets];
             bytesRead = proxySocket.getInputStream().read(buf);
-
-            if (bytesRead != 4)
+            if (bytesRead != octets)
                 throw new DnsLookupException("Invalid Tor Address Response");
 
             return InetAddress.getByAddress(buf);


ghost:
While initially working ok in clearnet, I found that if '_Use Tor for Bitcoin network_' is enabled, IPv6 clearnet addresses are ignored.  `DnsLookupTor` is hard coded to expect 4 bytes in response from the proxy, however with IPv6, 16 bytes are returned.  I'm making a subsequent commit here to address this second issue.

wiz:
Can you fix the commit messages so they are capitalized and well documented like
```
Allow IPv6 connections to Bitcoin nodes (fixes #3990)
Allow IPv6 response (16 byte) from Tor DNS lookup
```

wiz (changes requested):
NACK

Test case 1 failed:
1) Set btcnode configuration to `node100.wiz.network,node130.wiz.network,node140.wiz.network,node150.wiz.network`
Expected) Seeing IPv6 addresses in settings -> network status
Actual) Seeing IPv4 addresses in settings -> network status

Test case 2 failed:
1) Set btcnode configuration to `2401:b140::42:100,2401:b140::42:130,2401:b140::42:140,2401:b140::42:150`
Expected) Connects to IPv6 btcnodes
Actual) Bisq fails to start due to "invalid configuration"

wiz:
I did get it working using the [aaaa::bbbb]:8333 notation, but I believe most users will do the above 2 notations so please fix those. Great work tho.

Screen Shot 2020-03-17 at 1 02 46

ghost:
The IP address is obtained from the user parameters by calling DNS lookup in `DnsLookupTor.java`.  So for example `node100.wiz.network` is passed to DNS, what is returned is either and IPv4 (4 byte buffer) or IPv6 (16 byte buffer).  For this example it resolves to 103.99.168.100.  Maybe there is a parameter that can be written to the DNS socket telling it that  should be IPv6 rather than IPv4?  I'll try to find out.

ghost:
DnsLookupTor.java is doing the same SOCKS5 network query as `tor-resolve`.  Namely a CMD_RESOLVE:

    buf[0] = b('\u0005');             // version SOCKS5
    buf[1] = b('\u00f0');             // CMD_RESOLVE
    buf[2] = b('\u0000');             // reserved 
    buf[3] = b('\u0003');             // SOCKS5_ATYPE_HOSTNAME
    [...]
    proxySocket.getOutputStream().write(buf);



[Unfortunately:](https://tor.stackexchange.com/questions/7212/how-do-i-resolve-dns-using-tor)
> Only drawback so far: You can't specify the query type (e.g. A, NS, MX, etc.) and also IPv6-only hosts (those with only AAAA records) can't (yet) be resolved at all and cause such a message: [warn] Got SOCKS5 status response '4': host is unreachable

What we would need is the tor DNS equivalent of 

    nslookup -q=AAAA node100.wiz.network
    node100.wiz.network	has AAAA address 2401:b140::42:100

But we get this

    tor-resolve node100.wiz.network
    103.99.168.100


From reading tor-resolve [code](https://gitweb.torproject.org/tor.git/tree/src/tools/tor-resolve.c), there's no parameter that can be written to the DNS socket telling it that we want IPv6 result rather than IPv4.  

I'm open to suggestions.

wiz:
Why you close this PR ? Don't give up so easily lol

> I'm open to suggestions.

Why are you using Tor for DNS if `Use Tor for Bitcoin network` is not checked?

ghost:
> Why are you using Tor for DNS if Use Tor for Bitcoin network is not checked?

Bisq is not using Tor when that option is unchecked.  However Java is defaulting to IPv4 for the name lookup.  Setting the following [option](https://stackoverflow.com/questions/20262587/where-to-set-java-net-preferipv6addresses-true#20262705) with the JavaVM is necessary to make Java default to IPv6 in the case where both address types are available.    `-Djava.net.preferIPv6Addresses=true`

image

Success!

image

wiz:
Good work. Perhaps the "prefer IPv6 addresses" should be a setting in the Bisq preferences GUI, and then you can set the JVM options appropriately at startup if the preference is set.

ripcurlx:
> Good work. Perhaps the "prefer IPv6 addresses" should be a setting in the Bisq preferences GUI, and then you can set the JVM options appropriately at startup if the preference is set.
 
~Could you please add this as suggested by @wiz to the preferences and set it manually (`System.setProperty("java.net.preferIPv6Addresses", "true");`)  if applicable during runtime?~

That was a lucky timing with your last commit 😉 Could you please adapt the wording of your last commit messages, so it speaks without needing the context of @wiz comment in this PR? Thanks.

common/src/main/java/bisq/common/setup/CommonSetup.java
```diff
@@ -38,6 +38,9 @@ public static void setup(UncaughtExceptionHandler uncaughtExceptionHandler) {
 
         if (Utilities.isLinux())
             System.setProperty("prism.lcdtext", "false");
+
+        // #3990 allow IPv6 custom Bitcoin node in settings by hostname
+        System.setProperty("java.net.preferIPv6Addresses", "true");
```
ripcurlx: Is there a drawback when we set this as default for all nodes?

ripcurlx: Compared to doing it based on a user preference.

ghost: I have not seen any drawback from setting this, even if IPv6 is not available.  Here's a [document](https://docs.oracle.com/javase/1.5.0/docs/guide/net/ipv6_guide/index.html) on the subject, which mentions setting the property.  And [another one](https://stackoverflow.com/questions/20262587/where-to-set-java-net-preferipv6addresses-true#20262705) which mentions that the setting has to be made very early in app startup.

If you think its too risky we can revert to choosing between IPv4 and IPv6 addresses in BtcNodeConverter where it creates a PeerAddress from hostname.

ripcurlx:
@jmacxx Please sign your commit. Thanks!

wiz (changes requested):
Tested the following in GUI preferences:
1) `node100.wiz.network,node130.wiz.network,node140.wiz.network,node150.wiz.network`
2) `node100.wiz.network:8333,node130.wiz.network:8333,node140.wiz.network:8333,node150.wiz.network:8333`
3) `2401:b140::42:100,2401:b140::44:130,2401:b140::44:140,2401:b140::44:150`
4) `[2401:b140::42:100]:8333,[2401:b140::44:130]:8333,[2401:b140::44:140]:8333,[2401:b140::44:150]:8333`

All passed except 2, can you fix it?

wiz (changes requested):
A safer way to implement this would be:
* if "Use Tor" is unchecked, enable a new radio button for "Prefer IPv4" or "Prefer IPv6"
* if "Prefer IPv6" is selected, then set the JVM option `java.net.preferIPv6Addresses=true`

wiz (approved):
ACK

After discussion, we concluded that "fixing ipv6" and "preferring ipv6" are very separate scopes, and since "preferring ipv6" has potential side effects, we agreed it would be safer to limit the scope of this PR to only "fixing ipv6", which means the expected case for testcases 3 and 4 below is not to connect to IPv6, but to use the JVM default setting, so now all testcases pass. Tested on macOS and Linux using single stack and dual stack configurations.

1) `2401:b140::42:100,2401:b140::44:130,2401:b140::44:140,2401:b140::44:150`
2) `[2401:b140::42:100]:8333,[2401:b140::44:130]:8333,[2401:b140::44:140]:8333,[2401:b140::44:150]:8333`
3) `node100.wiz.network,node130.wiz.network,node140.wiz.network,node150.wiz.network`
4) `node100.wiz.network:8333,node130.wiz.network:8333,node140.wiz.network:8333,node150.wiz.network:8333`

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/4045#pullrequestreview-378041639

wiz:
@devinbileck can you please make a note to test both IPv4 single-stack and IPv4/IPv6 dual-stack configurations in your next release testing?

PR #4046, opened 2020-03-11
Add holder name for new Faster Payments accounts

### LINKED ISSUE #3976: Update "Faster Payments" payment method to require trader's name
### Description

Currently the "Faster Payments" payment method requires only a "UK sort code" and a "Account number". This is no longer enough. [Banks now also require a name](http://www.fasterpayments.org.uk/press-release/new-name-check-safeguard-payments-revealed-october-2018).
[Example from a Bisq user.](https://bisq.community/t/having-trouble-with-faster-payments/8983)

#### Additional info

It's important that the traders go into a trade knowing what type of info they'll need to share so that they can make an informed decision.

stejbac:
Use `excludeFromJsonDataMap` to add an optional owner full name property to Faster Payments (GBP) accounts, without affecting the trade JSON contract (for backwards compatibility). This is to fix #3976, that some banks have started checking the recipient name.

This uses the new map key: `public static final String HOLDER_NAME = "holderName";`

Make the full name a non-optional field for _new_ accounts, but display the details for old Faster Payments accounts in exactly the same way as before, only showing the extra field in the info popup, account and trade step views when it is present.

Also update the now misleading payment instructions in the buyer popup. With the change, it shows as:
Screenshot from 2020-03-25 23-01-39

Also add a popup to advise the user to recreate their Faster Payments account.

Make an info popup open in the take/create offer view, upon choosing to take or make a new offer, to instruct the user to recreate their old Faster Payments account with an owner full name (and preserved salt). Also show the popup upon manual selection of any old (i.e. missing full name) Faster Payments account from the trading account combo box, analogously to the ClearXchange (Zelle) warning popup logic.
Screenshot from 2020-03-25 23-36-33

I've also copied/overwritten the two new/updated English resource bundle strings into the non-English resource files, so that the translations stay in sync with the updated information. (As it is a UK-only payment method, the non-English translations are probably not all that important anyway.)

--

Finally, I removed some redundant null-handling of the `excludeFromJsonDataMap` field
and constructor parameters. This touched all the `PaymentAccountPayload` subclasses, which was quite a lot of files (although fairly trivial changes).

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -3020,7 +3020,13 @@ If you have not fulfilled the above requirements you will lose your security dep
   For the {0} seller it is highly recommended \
 to get in contact with the {1} buyer by using the provided email address or mobile number to verify that he or she \
   is really the owner of the Zelle (ClearXchange) account.
-
+payment.fasterPayments.newRequirements.info=When using Faster Payments some banks have started verifying the receiver''s \
+  full name. Your current Faster Payments account only specifies a UK bank account number and sort code. This used to be \
+  sufficient, but may not be now for some sender''s banks. New Faster Payments accounts in Bisq now include an owner \
+  full name. Please consider recreating your account in Bisq to provide future {0} buyers with a recipient name.\n\n\
+  When recreating a Faster Payments account, take care to copy the account age verification salt from your old account \
+  and use the same sort code and account number. (The owner full name or account name need not be the same, however.) \
+  This preserves the public hash, so that the new account need not be aged or signed from scratch.
 payment.moneyGram.info=When using MoneyGram the BTC buyer has to send the Authorisation number and a photo of the receipt by email to the BTC seller. \
   The receipt must clearly show the seller's full name, country, state and the amount. The buyer will get displayed the seller's email in the trade process.
 payment.westernUnion.info=When using Western Union the BTC buyer has to send the MTCN (tracking number) and a photo of the receipt by email to the BTC seller. \
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -621,10 +621,9 @@ portfolio.pending.step2_buyer.halCashInfo.headline=Send HalCash code
 portfolio.pending.step2_buyer.halCashInfo.msg=You need to send a text message with the HalCash code as well as the \
   trade ID ({0}) to the BTC seller.\nThe seller''s mobile nr. is {1}.\n\n\
   Did you send the code to the seller?
-portfolio.pending.step2_buyer.fasterPaymentsHolderNameInfo=Some banks might require the receiver's name. \
-  The UK sort code and account number used to be sufficient for a Faster Payment transfer, but since then some banks \
-  have started verifying the receiver's name. For this reason, accounts created in old Bisq clients do not provide the \
-  trade partner with an owner full name. Please use trade chat to avoid any issues in such cases.
+portfolio.pending.step2_buyer.fasterPaymentsHolderNameInfo=Some banks might verify the receiver's name. \
+  Faster Payments accounts created in old Bisq clients do not provide the receiver's name, \
+  so please use trade chat to obtain it (if needed).
 portfolio.pending.step2_buyer.confirmStart.headline=Confirm that you have started the payment
 portfolio.pending.step2_buyer.confirmStart.msg=Did you initiate the {0} payment to your trading partner?
 portfolio.pending.step2_buyer.confirmStart.yes=Yes, I have started the payment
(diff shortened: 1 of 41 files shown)


ripcurlx:
@m52go Could you please review the wording?
I'll test it on Regtest after reviewing the code.

desktop/src/main/java/bisq/desktop/components/paymentmethods/FasterPaymentsForm.java
```diff
@@ -44,6 +44,10 @@
 
     public static int addFormForBuyer(GridPane gridPane, int gridRow,
                                       PaymentAccountPayload paymentAccountPayload) {
+        if (!((FasterPaymentsAccountPayload) paymentAccountPayload).getHolderName().isEmpty()) {
+            addCompactTopLabelTextField(gridPane, ++gridRow, Res.get("payment.account.owner"),
+                    ((FasterPaymentsAccountPayload) paymentAccountPayload).getHolderName());
+        }
```
ripcurlx, line 50: If this field is shown during the trade process it creates unnecessary scrolling.
Bildschirmfoto 2020-03-26 um 16 16 33
I'd just change line 54-55 to
```
addCompactTopLabelTextField(gridPane, gridRow, 1, Res.get("payment.accountNr"),
                ((FasterPaymentsAccountPayload) paymentAccountPayload).getAccountNr());
```
which puts sort code and account nr next to each other.
Bildschirmfoto 2020-03-26 um 16 19 57
That would work for both cases.

ripcurlx (changes requested):
NACK - please see my comment regarding the buyers UI during the trade process.
Also please only commit changes to the English source file all other files will be overwritten automatically by loading translations from Transifex before the next release.

ACK from a functionality perspective. Tested old and new Faster Payment Accounts in trades.

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -622,8 +622,9 @@ portfolio.pending.step2_buyer.halCashInfo.msg=You need to send a text message wi
   trade ID ({0}) to the BTC seller.\nThe seller''s mobile nr. is {1}.\n\n\
   Did you send the code to the seller?
 portfolio.pending.step2_buyer.fasterPaymentsHolderNameInfo=Some banks might require the receiver's name. \
-  The UK sort code and account number is sufficient for a Faster Payment transfer and the receivers name is not verified \
-  by any of the banks.
+  The UK sort code and account number used to be sufficient for a Faster Payment transfer, but since then some banks \
+  have started verifying the receiver's name. For this reason, accounts created in old Bisq clients do not provide the \
+  trade partner with an owner full name. Please use trade chat to avoid any issues in such cases.
```
```suggestion
  Faster Payment accounts created in old Bisq clients do not provide the receiver's name, \
  so please use trade chat to obtain it (if needed).
```

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -3019,7 +3020,13 @@ If you have not fulfilled the above requirements you will lose your security dep
...
 to get in contact with the {1} buyer by using the provided email address or mobile number to verify that he or she \
   is really the owner of the Zelle (ClearXchange) account.
-
+payment.fasterPayments.newRequirements.info=When using Faster Payments some banks have started verifying the receiver''s \
+  full name. Your current Faster Payments account only specifies a UK bank account number and sort code. This used to be \
+  sufficient, but may not be now for some sender''s banks. New Faster Payments accounts in Bisq now include an owner \
+  full name. Please consider recreating your account in Bisq to provide future {0} buyers with a recipient name.\n\n\
+  When recreating a Faster Payments account, take care to copy the account age verification salt from your old account \
+  and use the same sort code and account number. (The owner full name or account name need not be the same, however.) \
+  This preserves the public hash, so that the new account need not be aged or signed from scratch.
```
```suggestion
payment.fasterPayments.newRequirements.info=Some banks have started verifying the receiver''s \
  full name for Faster Payments transfers. Your current Faster Payments account does not specify a full name.\n\n \
  Please consider recreating your Faster Payments account in Bisq to provide future {0} buyers with a full name.\n\n\
  When you recreate the account, make sure to copy the precise sort code, account number, and account age \
  verification salt values from your old account to your new account. This will ensure your existing account's age and \
  signing status are preserved.
```

desktop/src/main/java/bisq/desktop/components/paymentmethods/FasterPaymentsForm.java
stejbac: OK, I'll apply the suggested change.

m52go (commented):
Existing text was already good. I just made it more concise, since there's already so much text in the pop-ups.

I did not suggest edits across languages, in case anyone would like to propose edits to these edits. If these edits are acceptable, I can do so.

stejbac:
I've just done a force push to drop the last commit which made the unwanted translation changes, and replaced it with a commit to apply @ripcurlx's suggestion.

I just noticed the text change suggestions - will push those shortly.

core/src/main/resources/i18n/displayStrings.properties
stejbac: OK, I'll apply that change. I think the text 

> Some banks might require the receiver's name ...

should possibly be changed to 

> Some banks might verify the receiver's name ...

with the suggested change, to emphasise that the issue is that they've started checking it. 

(Although my bank still accepts dummy names - I'm not sure how much longer, though.)

core/src/main/resources/i18n/displayStrings.properties
stejbac: OK, I'll apply the suggested simplification.

stejbac:
I've just force pushed again to include the popup text simplifications in the last commit.

ripcurlx (approved):

Issue #3918, opened 2020-01-27
Bisq using up too much memory on Linux even after closing


FKrauss:
### Description

Bisq will take more and more memory until the system freezes when none remains.
Additionally if shutdown before no memory is left, not all of it is released. About 4-7GB are shown to be additionally occupied, but no process seems to hold it. Logging out and in again does not fix it. A restart does.

#### Version

1.2.5

### Steps to reproduce

Unknown, just start the application for me.

### Expected behaviour

I expect bisq to take maybe 3GB of ram and released it after closing the application.

### Actual behaviour

More and more ram will be occupied until none is left and the system starts freezing. Sometimes bisq gets terminated by the system other times I just force reboot.

#### Device or machine

Ubuntu 18.04.3
Ryzen 5 2600X
G.Skill Aegis DIMM Kit 16GB

#### Additional info

The log contains a case where, after a reboot I started Bisq and got lucky. The application got terminated (I reacted too slow and could no longer move the mouse) after the ram was full. About 5.5GB of ram were shown to be still occupied after the termination.

[bisq.log](https://github.com/bisq-network/bisq/files/4117738/bisq.log)

ripcurlx:
Related to #3128, #3657.

ripcurlx:
Related to: #3128, #3657, #3918, #3917, #3787, #3786, #3686, #3677, #3343

ghubstan:
I think this is due to one or several JavaFx bugs related to bug [JDK-8188094](https://bugs.openjdk.java.net/browse/JDK-8188094)

During profiling, I noticed that the JavaFX Application Thread had been allocated 66.6 GiB
![JavaFX-App-Thread-Alloc-66Gib-2020-02-19 17-45-46](https://user-images.githubusercontent.com/36207203/74886556-397d6200-5357-11ea-94b2-be600a94b996.png)
This couldn't be right, as my vm is 8 GiB, and my bisq instance had -Xmx=4g.  I thought it was just a JDK Mission Control bug, but I've been reproducing the problem with a tiny javafx demo app that constantly leaks from the JavaFx app thread (the same demo app found in the OpenJDK bug report).  The OS does not get the RAM back until I quit the IDE too.

I have not found a solution.  I'm running Ubuntu 18 & 19(vm), using an nvidia gpu.

ghubstan:
[...]

Issue #3917, opened 2020-01-26
bisq freezes entire pc on arch linux


paju1986:
I installed bisq from the snap store on arch linux, when i open the application after a while it freezes the entire desktop and i have to reboot the machine.

My computer has 16gb of Ram so i dont expect to be a ram usage issue.
I can't provide logs since my machine freezes and i have to reboot.

ripcurlx:
Related to: #3128, #3657, #3918, #3917, #3787, #3786, #3686, #3677, #3343

chimp1984:
Can you try out the recommendations at https://github.com/bisq-network/bisq/issues/3918#issuecomment-607198677 ?

cd2357:
Should be fixed in the most recent release (v1.6.4) which brings several UI performance improvements and generally reduces system resource consumption. Please try it out and let us know if it's still an issue for you.

Issue #3686, opened 2019-11-25
Does not start in VirtualBox Debian (latest) guest


dimaatmelodromru:
### Description

Does not start in VirtualBox Debian (latest) guest

Since Bisq freezes my OS Ubuntu 19.10 I was trying to start it in VirtualBox Debian (latest) guest, 4GB RAM.
Nope. No messages. Nothing.

#### Version

1.2.3

dimaatmelodromru:
logs:

cl1p.net - The internet clipboard cl1p.net/rbtdn
Here is your stuff. Thanks for using cl1p. Tell your friends!

Nov-25 18:36:03.562 [main] INFO  b.c.s.CoreSetup: 


Log files under: /home/user/.local/share/Bisq/bisq 
Nov-25 18:36:03.620 [main] INFO  b.c.u.Utilities: System info: os.name=Linux; os.version=4.19.0-6-amd64; os.arch=amd64; sun.arch.data.model=64; JRE=10.0.2+13 (Oracle Corporation); JVM=10.0.2+13 (Java HotSpot(TM) 64-Bit Server VM) 
Nov-25 18:36:03.788 [main] INFO  b.c.s.CoreNetworkCapabilities: Set Capability.RECEIVE_BSQ_BLOCK 
Nov-25 18:36:04.138 [main] INFO  b.c.app.Version: Version{VERSION=1.2.3, P2P_NETWORK_VERSION=1, LOCAL_DB_VERSION=1, TRADE_PROTOCOL_VERSION=2, BASE_CURRENCY_NETWORK=0, getP2PNetworkId()=10} 
Nov-25 18:36:04.143 [main] INFO  b.c.s.CoreSetup: Path to Bisq jar file: /opt/Bisq/app/desktop-1.2.3-all.jar 
Nov-25 18:37:09.944 [main] INFO  b.c.s.CoreSetup: 


Log files under: /home/user/.local/share/Bisq/bisq 
Nov-25 18:37:10.017 [main] INFO  b.c.u.Utilities: System info: os.name=Linux; os.version=4.19.0-6-amd64; os.arch=amd64; sun.arch.data.model=64; JRE=10.0.2+13 (Oracle Corporation); JVM=10.0.2+13 (Java HotSpot(TM) 64-Bit Server VM) 
Nov-25 18:37:10.209 [main] INFO  b.c.s.CoreNetworkCapabilities: Set Capability.RECEIVE_BSQ_BLOCK 
[...]

Issue #3677, opened 2019-11-24
Investigate if -BjvmOptions for linux build is still needed


chimp1984:
We added the jvm options `-BjvmOptions=-Xss1280k` to the Linux build. I am not sure if that is still needed and if it might be related to the reports that on Linux some users have experienced very high RAM consumption.

HiDef888:
Hello Chimp,
I am one of the users having RAM consumption issues starting Bisq.  I commented out the line in Bisq.cfg `-Xss1280k` you refer to, to see if it affected my RAM usage, and it did not fix the issue unfortunately. I have also tried to force limiting the JVM to certain constraints and that doesn't work for me either.
Thanks

ripcurlx:
Related to: #3128, #3657, #3918, #3917, #3787, #3786, #3686, #3677, #3343

freimair:
maybe we should try adding a `-XX:MaxRAM=2147483648` to the java opts of the bisq-desktop binaries. @ghubstan had some success with that and I can confirm that this limits the RAM usage drastically even after days of always-on.

cd2357:
> I am one of the users having RAM consumption issues starting Bisq. I commented out the line in Bisq.cfg `-Xss1280k` you refer to, to see if it affected my RAM usage, and it did not fix the issue unfortunately. I have also tried to force limiting the JVM to certain constraints and that doesn't work for me either.

Should be fixed in the most recent release (v1.6.4) which brings several UI performance improvements and generally reduces system resource consumption (especially RAM). Please try it out and let us know if it's still an issue for you.

PR #4048, opened 2020-03-11
Limit system ram to 4GB

freimair:
Might treat some symptoms of #3128, #3657, #3918, #3917, #3686, #3677

OpenJDK assumes total system RAM of 128GB. Setting it to 4GB manually
reduces the greediness of the JVM drastically.

Only after setting the MaxRAM to 1.5GB it finally went into a heap space error. 2GB work fine on my machine.

Proposed strategy: do a full release test with this, see how it goes.

This find is courtesy of @ghubstan https://github.com/bisq-network/bisq/issues/3918#issuecomment-594236960

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -133,6 +133,10 @@ configure([project(':cli'),
             unixScriptFile.text = unixScriptFile.text.replace(
                 'cd "`dirname \\"$PRG\\"`/.." >/dev/null', 'cd "`dirname \\"$PRG\\"`" >/dev/null')
 
+            def script = file("${rootProject.projectDir}/bisq-$applicationName")
+            script.text = script.text.replace(
+                'DEFAULT_JVM_OPTS=""', 'DEFAULT_JVM_OPTS="-XX:MaxRAM=4GB"')
+
             if (osdetector.os != 'windows')
                 delete fileTree(dir: rootProject.projectDir, include: 'bisq-*.bat')
             else
--- a/build.gradle
+++ b/build.gradle
@@ -133,9 +133,11 @@ configure([project(':cli'),
             unixScriptFile.text = unixScriptFile.text.replace(
                 'cd "`dirname \\"$PRG\\"`/.." >/dev/null', 'cd "`dirname \\"$PRG\\"`" >/dev/null')
 
-            def script = file("${rootProject.projectDir}/bisq-$applicationName")
-            script.text = script.text.replace(
-                'DEFAULT_JVM_OPTS=""', 'DEFAULT_JVM_OPTS="-XX:MaxRAM=4GB"')
+            if (applicationName == 'desktop') {
+                def script = file("${rootProject.projectDir}/bisq-$applicationName")
+                script.text = script.text.replace(
+                    'DEFAULT_JVM_OPTS=""', 'DEFAULT_JVM_OPTS="-XX:MaxRAM=4g"')
+            }
 
             if (osdetector.os != 'windows')
                 delete fileTree(dir: rootProject.projectDir, include: 'bisq-*.bat')
--- a/desktop/package/linux/package.sh
+++ b/desktop/package/linux/package.sh
@@ -108,6 +108,7 @@ $JAVA_HOME/bin/javapackager \
     -srcfiles desktop-$version-all.jar \
     -appclass bisq.desktop.app.BisqAppMain \
     -BjvmOptions=-Xss1280k \
+    -BjvmOptions=-XX:MaxRAM=4g \
     -BjvmOptions=-Djava.net.preferIPv4Stack=true \
     -outfile Bisq-$version \
     -v
@@ -134,6 +135,7 @@ $JAVA_HOME/bin/javapackager \
     -srcfiles desktop-$version-all.jar \
     -appclass bisq.desktop.app.BisqAppMain \
     -BjvmOptions=-Xss1280k \
+    -BjvmOptions=-XX:MaxRAM=4g \
     -BjvmOptions=-Djava.net.preferIPv4Stack=true \
     -outfile Bisq-$version \
     -v


wiz:
@ripcurlx @sqrrm please also consider cherry-pick for v1.2.8 - many users have RAM issues and it goes towards our goal of "improve reliability"

ghubstan:
@freimair:  "Only after setting the MaxRAM to 1.5GB it finally went into a heap space error. 2GB work fine on my machine".   2 GB was the minimum MaxRAM for my machine too;  I got an OOM when setting MaxRAM=1.5GB.

ripcurlx:
> @ripcurlx @sqrrm please also consider cherry-pick for v1.2.8 - many users have RAM issues and it goes towards our goal of "improve reliability"

I am happy to cherry-pick this for v1.2.8 if someone with a linux system does the required testing for it.

freimair:
please specify "required testing". It has been developed on Linux. It worked.

I would rather see it tested on Windows.

ripcurlx:
This is just for the startup script when building from source, correct? So the created binaries won't have this change.

ghubstan (changes requested):
Setting this param value to "4GB" doesn't work on my machine;  it needs to be "4G" or "4g".

$ java -XX:MaxRAM=4GB -version
Improperly specified VM option 'MaxRAM=4GB'
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.

$ java -XX:MaxRAM=4g -version
openjdk version "11" 2018-09-25
OpenJDK Runtime Environment 18.9 (build 11+28)
OpenJDK 64-Bit Server VM 18.9 (build 11+28, mixed mode)

build.gradle
```diff
@@ -133,6 +133,12 @@ configure([project(':cli'),
             unixScriptFile.text = unixScriptFile.text.replace(
                 'cd "`dirname \\"$PRG\\"`/.." >/dev/null', 'cd "`dirname \\"$PRG\\"`" >/dev/null')
 
+            if (applicationName == 'desktop') {
+                def script = file("${rootProject.projectDir}/bisq-$applicationName")
+                script.text = script.text.replace(
+                    'DEFAULT_JVM_OPTS=""', 'DEFAULT_JVM_OPTS="-XX:MaxRAM=4GB"')
```
```suggestion
                    'DEFAULT_JVM_OPTS=""', 'DEFAULT_JVM_OPTS="-XX:MaxRAM=4g"')
```

of course... sorry.

freimair:
> This is just for the startup script when building from source, correct? So the created binaries won't have this change.

Yes, however, as discussed, I also changed the packager scripts.

ripcurlx (approved):
utACK

@freimair Let's give this a try as discussed pointing users in this issue to try out master (if they are able to build and run it). I'll create new snapshot Linux binaries after the release for testing.

cbeams:
I just ran into the following heap space error on my always-on Bisq node. I upgraded that node from v1.3.2 to v1.3.4 on March 1st, and the heap space error occurred on March 4th. I haven't had any such errors before, or at least not for a very long time. Makes me think this change is probably the culprit. I just asked @bisq-network/support in Keybase whether they've been seeing similar reports from users, no response yet.

image

cd2357:
> I haven't had any such errors before, or at least not for a very long time. Makes me think this change is probably the culprit.

Probably the culprit, or probably just didn't fix the underlying issue.

I [read somewhere](https://wiki.openjdk.java.net/pages/viewpage.action?pageId=20415996) that JavaFX also uses the system's video RAM, in addition to the normal RAM usually used by the Java VM.

This PR seems to restrict only the normal RAM usage -- we might have to investigate the VRAM side as well.

Btw, this means Bisq (or any JavaFX app) could behave differently on different systems, depending on that system's video settings (graphic card hardware? graphics drivers? amount of shared RAM between video and system? etc -- don't know what could play a role, but smth to keep in mind when looking at performance).

ghubstan:
> Btw, this means Bisq (or any JavaFX app) could behave differently on different systems, depending on that system's video settings (graphic card hardware? graphics drivers? amount of shared RAM between video and system? etc -- don't know what could play a role, but smth to keep in mind when looking at performance).

JavaFX does behave differently with different GPUs.  It does use native GPU drivers, and there have been many leak problems in JFX related to GPU and mesa drivers.

Here is a list of some [jfx leak related issues](https://bugs.openjdk.java.net/browse/JDK-8115293?jql=project%20%3D%20JDK%20AND%20component%20%3D%20javafx%20AND%20text%20~%20%22leak%22) -- old and new.  You have to hunt around for those related to GPUs, and for issues about leak related bugs found and fixed after JFX 11 was released.   

Issue [8188094](https://bugs.openjdk.java.net/browse/JDK-8188094) shows how to make JFX leak from a simple app running on Linux.  The unresolved issue reports the problem in JDK/JFX 8/9, and I reproduced it running JFX 11 on Ubuntu + NVIDIA GPU in Feb. 2020, when I started working on this problem.

JFX 14 has some leak bugfixes, but Bisq is holding back on upgrading to OpenJDK 14 until the new  [java packaging tool](https://openjdk.java.net/jeps/343) is released from incubation to GA.

I did a lot of profiling to find the leaks (maybe you saw some of my comments in issue [3918](https://github.com/bisq-network/bisq/issues/3918)) using JMC/JFR, JProfiler, async-profiler, bcc and bpftrace tools, under the assumption the problem was/is in Bisq and JFX.  I'm sure there are leaks to find and fix, but the only firm conclusion my analysis supported was that Bisq's JVM was reserving far more resident and virtual memory at startup than the Linux JVM needs.  (I could not find problems with Bisq's on-heap management.)  The problem was more serious when it starts with an empty data dir, syncs the entire chain, and the JVM never gives back any memory (to the OS).

cd2357:
Thanks for the detailed reply @ghubstan, that clarifies a lot.

> I'm sure there are leaks to find and fix, but the only firm conclusion my analysis supported was that Bisq's JVM was reserving far more resident and virtual memory at startup than the Linux JVM needs. (I could not find problems with Bisq's on-heap management.) The problem was more serious when it starts with an empty data dir, syncs the entire chain, and the JVM never gives back any memory (to the OS).

Have you tried using the `-XX:+UseG1GC` flag? According to [some reports](https://stackoverflow.com/a/40829444) it causes JavaFX to use more aggressive garbage collection, which reduces memory consumption and effectively give it back to the system.

freimair:
I have seen that too but only once for my always-on bisq. Seems like a legitimate Out-of-Heap error since Bisq is resource hungry beyond compare. Restart Bisq and everything is fine again.

There are multiple efforts and fixes in the pipeline, though.
- As @ghubstan already said, java10 is very old and outdated, yet, we refuse to upgrade, and that would fix a whole lot of issues (https://github.com/bisq-network/bisq/pull/4242).
- Plus, there are projects in line to help reduce RAM requirements (https://github.com/bisq-network/projects/issues/29, https://github.com/bisq-network/projects/issues/25) which are either waiting to be reviewed and merged or waiting to start (with a clear :+1: from the community but no budget for over a month now).

@cd2357 we played around with the -XX:+UseG1GC flag way back and it did not change anything. Only when I hunted down and found the faulty line of code that produced the resource leak, thing changed. Bisq still holds the whole data stores in RAM (not at first, but just click the "trades" tab and there you go) and it will do so until we change things.

As long as we do not fix the underlying issues, we will only treat single failures and waste time and resources over and over again.

PR #4049, opened 2020-03-11
Use Linux OS suspend inhibitors to avoid standby

ghubstan:
On linux hosts, AvoidStandbyService checks for executables
gnome-session-inhibit and systemd-inhibit in that order, then starts
the first found to prevent suspend or sleep while Bisq is running.

Although the hide avoid standby mode button is not displayed on linux,
the inhibitors can be toggled off/on when and if that button is
displayed.  An installed shutdown hook will turn the inhibitor off,
but they should always die when the parent process dies.

To check when an inhibitor is enabled or disabled from a linux terminal:

	If the desktop is gnome...

	$ dbus-send --session --print-reply --dest=org.gnome.SessionManager \
		/org/gnome/SessionManager org.gnome.SessionManager.GetInhibitors

	If Bisq falls back to systemd-inhibit on a non gnome desktop...

	$ systemd-inhibit --list
	$ systemd-inhibit --list | grep Bisq

If neither gnome-session-inhibit nor systemd-inhibit are present, the
standby service falls back to playing the silent audio file.

This does not cover all linux desktop types, but other inhibitors
exist for desktops I have not set up, such as mate, and it will not
be difficult to make changes to support them.

Memory use reduction varies from machine to (virtual) machine, but on my
Ubuntu 18 laptop, playing the audio file to block suspension is allocating
about 1 GB of native memory per hour -- via malloc() -- which this PR elminates.
On the same machine, virtual memory use goes from 10.4 GB to 10.1 GB.

This has been tested on Ubuntu 18 (gnome) and the following VMs:

	Debian 10.3 (gnome)
	Arch Linux 9.2.1 (gnome)
	Fedora 31.1.9 (cinnamon)

But, VMs not running an inhibitor or audio file did not always go into suspend mode
as expected.

This does not work on Windows Cygwin, but minor changes might make it work if
cygwin inhibitor tools like gnome-session-inhibit.exe and mate-session-inhibit.exe
are installed.

Addresses native / virtual memory allocation issues.

Your PR description here.

### DIFF
--- a/core/src/main/java/bisq/core/app/AvoidStandbyModeService.java
+++ b/core/src/main/java/bisq/core/app/AvoidStandbyModeService.java
@@ -22,12 +22,22 @@ import bisq.core.user.Preferences;
 import bisq.common.config.Config;
 import bisq.common.storage.FileUtil;
 import bisq.common.storage.ResourceNotFoundException;
+import bisq.common.util.Utilities;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+import java.nio.file.Paths;
+
 import java.io.File;
 import java.io.IOException;
+
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.CountDownLatch;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
 
 import lombok.extern.slf4j.Slf4j;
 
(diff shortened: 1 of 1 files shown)


ghubstan:
Closing this PR and will create another with OSX standby mode disabler functionality merged into  AvoidStandbyService.  

Right now, in master, it seems both caffeinate and audio player are running on OSX.  I will verify by testing on OSX.

PR #4050, opened 2020-03-11
Fixed wrongly removed listeners

### LINKED ISSUE #2779: GUI bug when editing offer
To reproduce:

0. Have an open offer
1. Click edit offer from the Portfolio tab and change the unit price. The change is reflected in the total amount - that's ok.
2. Without clicking on Confirm, select another tab at the top, e.g. Market.
3. Return to the tab Portfolio and observe two issues:
3.1. your changed unit price is reverted to the original unit price
3.2. entering another unit price does NOT change the total amount! Hopefully once I confirm the changed unit price (while seeing the wrong total amount), the correct amount will be broadcasted and not the one shown in the GUI.

BTW, this bug does not happen if in step 2 you select a subtab under Portfolio, e.g. History. The bug only happens if you select a top tab and then go back to Portfolio.

v1.0.1 on Windows, from downloaded binary.

freimair:
Fixes https://github.com/bisq-network/bisq/issues/2779

This is a 2 part fix:
- [x] prevent listener from being removed
- [ ] keep changes to EditOfferView even when navigating elsewhere

<details>

the second one is trickier, that is what I got so far:
- when switching to another tab of the main toolbar and returning, 

https://github.com/bisq-network/bisq/blob/0501b54633eb49d71f58bfbec1806ce83eb747be/desktop/src/main/java/bisq/desktop/main/offer/MutableOfferView.java#L225-L226

 gets called from outside. When negating the "if", values are kept.
- however, once that works and you switch to another subtab and back, everything is reset again
- going forth, the lines below disable the tab right before it is reenabled and therefore resets the model
https://github.com/bisq-network/bisq/blob/0501b54633eb49d71f58bfbec1806ce83eb747be/desktop/src/main/java/bisq/desktop/main/portfolio/PortfolioView.java#L105-L106

the thing is, this line received a change from @ManfredKarrer:

https://github.com/bisq-network/bisq/commit/3b91f08ce48042e237a8635f28a4616cd885c237#diff-8d842486c5cbb597c1a8c4ab576f274fL108-L109

with a comment saying that has been done because deactivated offers kept being republished.

</details>

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/offer/MutableOfferView.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/MutableOfferView.java
@@ -284,6 +284,9 @@ public abstract class MutableOfferView<M extends MutableOfferViewModel<?>> exten
     ///////////////////////////////////////////////////////////////////////////////////////////
 
     public void onTabSelected(boolean isSelected) {
+        if (isSelected == model.getDataModel().isTabSelected)
+            return;
+
         if (isSelected && !model.getDataModel().isTabSelected)
             doActivate();
         else


freimair:
@ripcurlx handing it over to you as discussed...

cd2357:
@freimair / @ripcurlx  -- is it a good idea to let a bot auto-close issues? Seems this one had recent activity is already about half-done.

Issue #3489, opened 2019-10-27
Bisq does not verify taker's funding TX is broadcast to Bitcoin P2P network


wiz:
### Description

Bisq allows its offers to be taken without verifying the taker's funding TX was actually broadcast to the Bitcoin P2P network.

#### Version

1.1.7

### Steps to reproduce

1) Configure Alice's Bisq node to only connect to a single a Bitcoin node that does not relay transactions to the Bitcoin P2P network (i.e. configure it to have zero peers)
2) Create an offer on Bob's node
3) Take the offer on Alice's node

### Expected behaviour

Alice's node should not attempt to take the offer if it cannot broadcast a valid funding TX to the Bitcoin P2P network.

Bob's node should not allow Alice to accept his offer if Bob cannot verify that Alice's funding TXID has been seen by at least 1 of his Bitcoin peers.

### Actual behaviour

Bob's offer gets accepted state, but the trade can never begin because Alice's funding TX was not broadcast to the Bitcoin P2P network. Additionally, Bob's funds are now stuck in limbo.

### Screenshots

<img width="1196" alt="Screen Shot 2019-10-27 at 19 38 13" src="https://user-images.githubusercontent.com/232186/67633385-a2495600-f8f2-11e9-8d86-645ca873db1b.png">

chimp1984:
The peer's trade fee tx is not knows to BitcoinJ as it is not related to the wallet. Beside that it is a valid situation that the peer's trade fee is not confirmed or not even found in the mempool, specailly for the taker fee tx that is likely the case as the time between publishing taker fee tx and deposit tx is very short and the tx is likely not propagated much in the network.

We have to find out what causes those bugs with not published txs. It might be related to the AddressEntry management, that there is a bug in some edge cases.

brincobt:
@wiz: I have put my log at a web URL. See bisq chat, please.

wiz:
@chimp1984 since BitcoinJ / SPV cannot verify this TX, we are discussing having all the btcnode operators start running a block explorer API like electrs on their servers - this will allow Bisq to query the full node if a TX exists in the mempool or not, which means the fullnode would have verified it. Users will be able to choose which node(s) to verify this TX using, including their own full node if they have one.

wiz:
[...]

PR #4053, opened 2020-03-12
Limit number of unconfirmed offers

### LINKED ISSUE #3705: Takers deposit transaction rejected for 'too-long-mempool-chain' causing issues
### Description

If a taker attempts to take too many offers that have not been confirmed, the deposit transaction will be rejected by the Bitcoin network for 'too-long-mempool-chain'. The affected trade will then be moved into his failed trades tab but remains open for the maker without any indication of an error.

The maker then likely waits for the trade timer to run out, with any chat attempts to contact the taker going unanswered (since chat messages are not shown to them for failed trades). Then the maker finally opens a support ticket and the mediator is able to communicate with both Alice and Bob (it will show up under the takers support tickets).

When the mediator suggests a payout, only the maker sees the mediation result in her open trades. The taker doesn't since the trade is in Failed trades. The maker is able to accept the result, but nothing happens since the taker is unable to accept as well. At this point, the maker is unable to contact the mediator (since the support ticket was closed), and is unable to contact the taker.

If the maker then tries to reject the result, she will encounter an error indicating the deposit transaction is not confirmed and unable to open an arbitration dispute [3]. And after doing an SPV resync, encounters an error that the trade has no deposit transaction set and to move it to failed trades [4].

#### Version

release/v.1.2.4 (commit hash 548a217751b44abab8452cef609ebefb420835f0)

### Steps to reproduce

Using a regtest setup, perform the following steps without generating any blocks:
1. Alice makes about 15 offers to buy BTC
2. Bob tries to take all the offers to sell BTC

### Actual behaviour

When attempting to take the 13th offer, Bob will encounter a transaction rejection error [1] and [2], and the trade will move into his Failed Trades tab. Meanwhile, the trade will show in Alice's open trades without any indication that an error occurred on the takers side. And even after generating several blocks, remains waiting for blockchain confirmation (note, I did see on a few attempts to reproduce this issue that it did eventually get confirmed after several blocks).

Note, the multisig deposit appears in both wallets, but remains unconfirmed.

### Expected behaviour

At a minimum, after Bob encounters the transaction rejection error, Alice's trade should update and indicate an error was encountered. Perhaps it needs to move to Failed Trades as well. Though ideally we should prevent this type of rejected transaction in the first place and perhaps block creating/taking too many offers without confirmation.

Additionally, do we need to allow chat for failed trades and potentially view mediator result for failed trades?

### Screenshots

[1]
image

[2]
image

[3] 
image

[4]
image

#### Environment

- Windows 10
- Regtest environment with Bitcoin core v0.18.1

#### Additional information

In Bitcoin Core if a transaction in the mempool has 25 descendants, or it and all of its descendants are over 101,000 vbytes, any newly-received transaction that is also a descendant will be rejected. So once you have 25+ UTXOs in a long unspent chain of unconfirmed transactions, you will encounter this issue. And since each take offer has 2 transactions (taker/tx fee and multisig deposit), that is why the issue was encountered when attempting to take the 13th offer.

freimair:
Fixes https://github.com/bisq-network/bisq/issues/3705

It seems that Bisq can reach a critical mass of unconfirmed transactions. Exceeding this mass causes all sorts of failures. In the corresponding issue it has been suggested to just limit the number of unconfirmed transactions.

After thorough discussion with @chimp1984, @sqrrm, @wiz and @ripcurlx, we agreed on a first step, namely, to introduce Bisqs own hard limit, get it into the next release and ultimately see if we can reduce support cases (currently 8-10 cases per month out of [2k5 trades](https://monitor.bisq.network/d/SzHBCGjmz/trades-playground?orgId=1&fullscreen&panelId=19))

This PR introduces Bisqs own hard limit of 20 unconfirmed transactions. If the limit is exceeded during
- offer creation at the makers side,
- taking the offer at the takers side,
- and taking the offer at the makers side,
processes are stopped so that no harm can be done.

# Considerations
- Although these errors are non-blocking and not meant to be reportet, the error handling framework does not allow for a more correct interaction with the user. We agreed on (@ripcurlx) creating an independent issue regarding the GUI error handling and keep this PR focused. The error handling is tracked in https://github.com/bisq-network/bisq/issues/4079.
- Note that there is hardly any synchronization in there and race conditions are to be expected. First, because of how Bisq uses BitcoinJ, there is no correct sync possible. Second, the hard limit is set to 20 which is less than the actual default hard limit of the bitcoin network (25) to allow for some race conditions. Third, this PR is designed to reduce the number of support requests. If time raises the need for more proper handling, we shall do it then.
- I considered extending the protocol to properly report the error. However, by doing so, newer clients cannot take offers of older clients because of how protobuf works. Again, if the need arises over time, we shall do it.

<details>

# Discussions, findings and strategies
To all the experts out there, here are some points to be discussed
- does this "hard limit" only affect offers (I guess the fee txs) or does it affect all txs. The question is should we create a more general safety net?
- is that something to be fixed in bitcoinj?

Please note that I am no expert in the BTC parts of the Bisq code plus the PR is currently largely untested. It is just there to get things moving.

EDIT: seems to be related to https://github.com/bisq-network/proposals/issues/132 and https://github.com/bisq-network/bisq/issues/3489

EDIT2: update after talking to @chimp1984 and @sqrrm 

- future impact
  - once the API can create offers and trading bots are used, this can get critical fast
  
- brainstormed strategies
  - report trade fail to trading partner
    - minor changes to the trade protocol implementation (check!)
    - what happens if the trade succeeds afterwards? because transactions are accepted later?
    - need for a cancellation feature
      - how to revert already committed transactions..
  - only publish offers once fee transaction is confirmed
    - hugely impacts usability because it can take a lot of time until transactions get confirmed (eg. 1h to confirmation, 5 offers take 5h to publish (implement a queue?)
    - might take way to long and thus, might break the trade protocol (does it during offer creation?)
    - does not prevent #3705 from happening
  - make best effort to prevent error
    - introduce Bisqs own hard limit for pending transactions
      - check before offer is created
      - check before offer is taken
        - on the takers side
        - on the makers side triggered by OfferAvailableRequest + error code in case we are close to the limit (maybe not necessary (yet), because maker only creates the fee transactions in sequence, multisig transactions coming in later might result in a tree (check!))
    - does not work with external wallet
    	- however, someone using an external wallet might not be using the same wallet for multiple trades for privacy
      - if the wallet funding transaction fails, it is out of scope for Bisq? (for now at least?)
  - others?

EDIT3: update after a call with @wiz:

https://github.com/bisq-network/bisq/issues/3705#issuecomment-600037511

</details>

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/WalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/WalletService.java
@@ -609,6 +609,15 @@ public abstract class WalletService {
         return wallet.getLastBlockSeenHeight();
     }
 
+    /**
+     * Check if there are more than 20 unconfirmed transactions in the chain right now.
+     *
+     * @return true when queue is full
+     */
+    public boolean isUnconfirmedTransactionsLimitHit() {
+        return 20 < getTransactions(true).stream().filter(transaction -> transaction.isPending()).count();
+    }
+
     public Set<Transaction> getTransactions(boolean includeDead) {
         return wallet.getTransactions(includeDead);
     }
--- a/core/src/main/java/bisq/core/offer/OpenOfferManager.java
+++ b/core/src/main/java/bisq/core/offer/OpenOfferManager.java
@@ -637,6 +637,12 @@ public class OpenOfferManager implements PeerManager.Listener, DecryptedDirectMe
                 availabilityResult = AvailabilityResult.OFFER_TAKEN;
             }
 
+            if (btcWalletService.isUnconfirmedTransactionsLimitHit()) {
+                errorMessage = "There are too many unconfirmed transactions at the moment. Please try again later.";
+                log.warn(errorMessage);
+                availabilityResult = AvailabilityResult.UNKNOWN_FAILURE;
+            }
+
             OfferAvailabilityResponse offerAvailabilityResponse = new OfferAvailabilityResponse(request.offerId,
                     availabilityResult,
                     arbitratorNodeAddress,
--- a/core/src/main/java/bisq/core/offer/OpenOfferManager.java
+++ b/core/src/main/java/bisq/core/offer/OpenOfferManager.java
@@ -638,7 +638,7 @@ public class OpenOfferManager implements PeerManager.Listener, DecryptedDirectMe
                 availabilityResult = AvailabilityResult.OFFER_TAKEN;
             }
 
-            if (btcWalletService.isUnconfirmedTransactionsLimitHit()) {
+            if (btcWalletService.isUnconfirmedTransactionsLimitHit() || bsqWalletService.isUnconfirmedTransactionsLimitHit()) {
                 errorMessage = Res.get("shared.unconfirmedTransactionsLimitReached");
                 log.warn(errorMessage);
                 availabilityResult = AvailabilityResult.UNKNOWN_FAILURE;
--- a/core/src/main/java/bisq/core/offer/OpenOfferManager.java
+++ b/core/src/main/java/bisq/core/offer/OpenOfferManager.java
@@ -22,6 +22,7 @@ import bisq.core.btc.wallet.BtcWalletService;
 import bisq.core.btc.wallet.TradeWalletService;
 import bisq.core.dao.DaoFacade;
 import bisq.core.exceptions.TradePriceOutOfToleranceException;
+import bisq.core.locale.Res;
 import bisq.core.offer.availability.DisputeAgentSelection;
 import bisq.core.offer.messages.OfferAvailabilityRequest;
 import bisq.core.offer.messages.OfferAvailabilityResponse;
@@ -638,7 +639,7 @@ public class OpenOfferManager implements PeerManager.Listener, DecryptedDirectMe
             }
 
             if (btcWalletService.isUnconfirmedTransactionsLimitHit()) {
-                errorMessage = "There are too many unconfirmed transactions at the moment. Please try again later.";
+                errorMessage = Res.get("shared.unconfirmedTransactionsLimitReached");
                 log.warn(errorMessage);
                 availabilityResult = AvailabilityResult.UNKNOWN_FAILURE;
             }
--- a/core/src/main/java/bisq/core/offer/placeoffer/PlaceOfferProtocol.java
+++ b/core/src/main/java/bisq/core/offer/placeoffer/PlaceOfferProtocol.java
@@ -18,6 +18,7 @@
 package bisq.core.offer.placeoffer;
 
 import bisq.core.offer.placeoffer.tasks.AddToOfferBook;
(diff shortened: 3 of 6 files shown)


huey735:
I'm not a coder let alone expert but echoing my comments on https://github.com/bisq-network/bisq/issues/3489#issuecomment-589155725
> I'm vehemently against relaying offers with not-broadcast or unconfirmed maker fee tx.

From my understanding this PR aims to reduce current issues but I as I see we need to first workout the current model in which the network takes the maker's messages for granted without any check that the make-offer transaction has even been properly broadcast.

freimair:
@huey735  that is a valid point. if I get you that would compile down to
- not broadcasting offers until the makertx has been confirmed
- not accepting offers that have unconfirmed makertx

the hard limit for the offer maker is still there but then, only the maker herself will get a too-long-mempool-chain-error when creating the umpteenth offer in sequence. (?)

huey735:
@freimair it would go a long way if the network could at least check that they saw the make-offer transaction on their mempools (this could be delegated to the seed nodes). But at the moment even that doesn't happen. So no for offers with unbroadcast make-offer transactions.

As for broadcast but unconfirmed make-offer transactions. This limit may not make sense for the maker?! They are responsible for 1 of the 3 transactions relevant to this issue. If they make 10 offers with all broadcast but unconfirmed transactions and 10 different traders take the offers before they've been confirmed would this issue affect the maker?
@devinbileck description only described the taker's side.

I don't know which transaction of the following the maker keeps track of:
1. make-offer tx
2. take-offer tx
3. multisig deposit

Assuming from the issue this PR is trying to address, the taker carries transactions 2 and 3.

freimair:
can it have some positive effect if we only broadcast the offer when we see our own maker fee confirmed? And shouldn't we do that anyways?

huey735:
@freimair Possibly. ping @wiz and @ripcurlx

wiz:
> can it have some positive effect if we only broadcast the offer when we see our own maker fee confirmed? And shouldn't we do that anyways?

Bisq was clearly written with the intention of immediately broadcasting offers with unconfirmed funding TX, which is totally fine as long as the funding TX is **actually in the mempool** and wasn't rejected by the network. IMO all Bisq nodes should verify this but since they are SPV that's not easy.

freimair:
due to deficiencies in the GUI error reporting framework, I can only get it to report like that:
Screenshot from 2020-03-20 13-06-23

wiz:
@freimair I have another suggestion. Have both parties send the full serialized Bitcoin TX to each other. at the time of offer taking and have both parties re-broadcast the counterparty's TX to their Bitcoin nodes - if they do not receive either a "broadcast successful" or "already broadcasted" response, they know the TX is invalid.

sqrrm (approved):
utACK

PR #4054, opened 2020-03-12
Fix using /bin/sh instead of /bin/bash in journalscraper.sh

wiz:
Fixes error @freimair 
```
Mar 12 13:43:19 gztmprecgqjq64zh collectd[1218]: exec plugin: exec_read_one: error = /journalreader/scraperscript.sh: 14: /journalreader/scraperscript.sh: Bad substitution
```

### DIFF
--- a/pricenode/journalscraper.sh
+++ b/pricenode/journalscraper.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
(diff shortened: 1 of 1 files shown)


freimair (approved):
jap. works.

sqrrm (approved):
utACK

Issue #4052, opened 2020-03-12
Bisq v1.2.8 fails to start on macOS catalina


wiz:
```
% /Applications/Bisq.app/Contents/MacOS/Bisq --appDataDir=foo
Mar-12 18:50:26.563 [main] INFO  bisq.core.setup.CoreSetup: 


Log files under: foo/bisq 
Mar-12 18:50:26.577 [main] INFO  bisq.common.util.Utilities: System info: os.name=Mac OS X; os.version=10.15.3; os.arch=x86_64; sun.arch.data.model=64; JRE=10.0.1+10 (Oracle Corporation); JVM=10.0.1+10 (Java HotSpot(TM) 64-Bit Server VM) 
Mar-12 18:50:26.580 [main] INFO  b.c.s.CoreNetworkCapabilities: Set Capability.RECEIVE_BSQ_BLOCK 
Mar-12 18:50:26.669 [main] INFO  bisq.common.app.Version: Version{VERSION=1.2.8, P2P_NETWORK_VERSION=1, LOCAL_DB_VERSION=1, TRADE_PROTOCOL_VERSION=2, BASE_CURRENCY_NETWORK=0, getP2PNetworkId()=10} 
Mar-12 18:50:26.669 [main] INFO  bisq.core.setup.CoreSetup: Path to Bisq jar file: /Applications/Bisq.app/Contents/Java/Bisq-1.2.8.jar 
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.inject.internal.cglib.core.$ReflectUtils$1 (file:/Applications/Bisq.app/Contents/Java/Bisq-1.2.8.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of com.google.inject.internal.cglib.core.$ReflectUtils$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
[...]

PR #4058, opened 2020-03-13
Give LocalBitcoinNode a timeout in case BitcoinJ hangs during handshake attempt

### LINKED ISSUE #4057: Bisq v1.2.8 fails to start on Linux about 25% of the time
Following up from #4052

If I start 4 nodes, only 3 will start, for some reason it hangs at this line
```
Mar-12 16:48:39.208 [JavaFX Application Thread] INFO  b.c.b.n.LocalBitcoinNode: Initiating attempt to connect to and handshake with a local Bitcoin node (which may or may not be running) on port 8333.
```

Adding the command line option `--ignoreLocalBtcNode=true` causes all 4 nodes to start correctly.

dmos62:
Fixes #4057 

@wiz discovered (#4057) than in some cases LocalBitcoinNode will hang during Bisq setup. This patch gives LocalBitcoinNode's handshake attempt a timeout. Specifically, the part where we block while waiting for BitcoinJ to attempt to return a local BTC node's VersionMessage. I had not originally implemented a timeout, because I was under the impression that in such a case the NioClient would timeout (which does have a timeout).

@wiz has tested this patch and verified that with this patch Bisq handles BitcoinJ hanging gracefully (by timing out the relevant future after 5000 ms).

This patch also prints an error in case a timeout happens, because it normally shouldn't.

The underlying cause is unknown, but seems to come from BitcoinJ.

### DIFF
--- a/core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
+++ b/core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
@@ -28,6 +28,8 @@ import java.io.IOException;
 import java.util.Optional;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.TimeUnit;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -49,6 +51,7 @@ public class LocalBitcoinNode {
 
     private static final Logger log = LoggerFactory.getLogger(LocalBitcoinNode.class);
     private static final int CONNECTION_TIMEOUT = 5000;
+    private static final int HANDSHAKE_TIMEOUT = CONNECTION_TIMEOUT;
 
     private final Config config;
     private final int port;
@@ -244,10 +247,15 @@ public class LocalBitcoinNode {
 
         // block for VersionMessage or cancellation (in case of connection failure)
         try {
-            var peerVersionMessage = peerVersionMessageFuture.get();
+            var peerVersionMessage = peerVersionMessageFuture.get(HANDSHAKE_TIMEOUT, TimeUnit.MILLISECONDS);
             optionalPeerVersionMessage = Optional.of(peerVersionMessage);
-        } catch (ExecutionException | InterruptedException | CancellationException ex) {
+        } catch (ExecutionException | InterruptedException | CancellationException | TimeoutException ex) {
             optionalPeerVersionMessage = Optional.empty();
+            if (ex instanceof TimeoutException) {
+                log.error("Exploratory handshake attempt with a local Bitcoin node (that may not be there)" +
+                        " unexpectedly timed out. This should never happen; please report this. HANDSHAKE_TIMEOUT" +
+                        " is {} ms. Continuing as if a local BTC node was not found.", HANDSHAKE_TIMEOUT);
+            }
         }
 
         peer.close();
--- a/core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
+++ b/core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
@@ -249,13 +249,13 @@ public class LocalBitcoinNode {
         try {
             var peerVersionMessage = peerVersionMessageFuture.get(HANDSHAKE_TIMEOUT, TimeUnit.MILLISECONDS);
             optionalPeerVersionMessage = Optional.of(peerVersionMessage);
-        } catch (ExecutionException | InterruptedException | CancellationException | TimeoutException ex) {
+        } catch (ExecutionException | InterruptedException | CancellationException ex) {
             optionalPeerVersionMessage = Optional.empty();
-            if (ex instanceof TimeoutException) {
-                log.error("Exploratory handshake attempt with a local Bitcoin node (that may not be there)" +
-                        " unexpectedly timed out. This should never happen; please report this. HANDSHAKE_TIMEOUT" +
-                        " is {} ms. Continuing as if a local BTC node was not found.", HANDSHAKE_TIMEOUT);
-            }
+        } catch (TimeoutException ex) {
+            optionalPeerVersionMessage = Optional.empty();
+            log.error("Exploratory handshake attempt with a local Bitcoin node (that may not be there)" +
+                    " unexpectedly timed out. This should never happen; please report this. HANDSHAKE_TIMEOUT" +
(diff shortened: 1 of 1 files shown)


dmos62:
The fix is such that if BitcoinJ hangs (like @wiz was sometimes experiencing), a 5 second timeout will get triggered. This 5 second prolonging isn't ideal, but the other alternative is diving into BitcoinJ for patching. That would be difficult, because I can't reproduce the issue myself. Another alternative is reducing the timeout. Even a 2 seconds timeout shouldn't get triggered under normal circumstances.

wiz (commented):
I've tested this PR and it does prevent my issue #4057 from reproducing. However, I feel I should NACK this PR because of the multiple issues this feature has introduced (including #4052), and I personally feel we should revert the feature for now and re-evaluate it (with heavy testing) before merging it back into master. The most important thing is to keep Bisq stable and we must be risk averse when it comes to these types of changes.

ripcurlx:
> and I personally feel we should revert the feature for now and re-evaluate it (with heavy testing) before merging it back into master. The most important thing is to keep Bisq stable and we must be risk averse when it comes to these types of changes.

That was my first intention, but the problem with reverting this feature is that unfortunately too much code touching this parts of the codebase occurred, so I can't just revert it by the GitHub feature. So it would also need lots of manual code changes which are also error-prone as well. I agree that keeping Bisq stable is key, so I think for the future we should also discuss having more devs than just two involved in release testing. In our last three releases we always had issues based on very specific OS setups. That can only be detected if lots of people are at least starting and smoke testing new releases. In short for the v1.2.9 update I think we need to go with the best local bitcoin node detection solution we can come up with.

sqrrm (approved):
utACK

sqrrm:
I think, like @ripcurlx, that we should merge this PR since it's fixing something that's wrong in the current code base. If we decide to not release the local bitcoin node detection we'll make a separate revert commit later on. I won't merge until at least one more person gives an ACK or @wiz changes his mind though.

@wiz this PR doesn't affect the inclusion of local bitcoin node in the next release. If that's your reason for NACK I think that's the wrong way to suggest reverting the entire feature.

ripcurlx:
@dmos62 When running with a local full node I recognized that I always the following exception log in the console:
```
März-14 11:54:29.683 [NioClientManager] WARN  o.b.net.ConnectionHandler: Error handling SelectionKey: java.nio.channels.CancelledKeyException  java.nio.channels.CancelledKeyException: null
	at java.base/sun.nio.ch.SelectionKeyImpl.ensureValid(SelectionKeyImpl.java:73)
	at java.base/sun.nio.ch.SelectionKeyImpl.readyOps(SelectionKeyImpl.java:87)
	at java.base/java.nio.channels.SelectionKey.isWritable(SelectionKey.java:310)
	at org.bitcoinj.net.ConnectionHandler.handleKey(ConnectionHandler.java:229)

java.nio.channels.CancelledKeyException: null
	at java.base/sun.nio.ch.SelectionKeyImpl.ensureValid(SelectionKeyImpl.java:73)
	at java.base/sun.nio.ch.SelectionKeyImpl.readyOps(SelectionKeyImpl.java:87)
	at java.base/java.nio.channels.SelectionKey.isWritable(SelectionKey.java:310)
	at org.bitcoinj.net.ConnectionHandler.handleKey(ConnectionHandler.java:229)
```
The connection is established successfully to the node, but I did get this exception.

wiz:
@sqrrm yes good points, you're right. I spoke with @ripcurlx and he told me the release plans for monday

dmos62:
> @dmos62 When running with a local full node I recognized that I always the following exception log in the console:
> 
> ```
> März-14 11:54:29.683 [NioClientManager] WARN  o.b.net.ConnectionHandler: Error handling SelectionKey: java.nio.channels.CancelledKeyException  java.nio.channels.CancelledKeyException: null
> 	at java.base/sun.nio.ch.SelectionKeyImpl.ensureValid(SelectionKeyImpl.java:73)
> 	at java.base/sun.nio.ch.SelectionKeyImpl.readyOps(SelectionKeyImpl.java:87)
> 	at java.base/java.nio.channels.SelectionKey.isWritable(SelectionKey.java:310)
> 	at org.bitcoinj.net.ConnectionHandler.handleKey(ConnectionHandler.java:229)
> 	at org.bitcoinj.net.NioClientManager.handleKey(NioClientManager.java:86)
> 	at org.bitcoinj.net.NioClientManager.run(NioClientManager.java:122)
> 	at com.google.common.util.concurrent.AbstractExecutionThreadService$1$2.run(AbstractExecutionThreadService.java:66)
> 	at com.google.common.util.concurrent.Callables$4.run(Callables.java:122)
> 	at org.bitcoinj.utils.ContextPropagatingThreadFactory$1.run(ContextPropagatingThreadFactory.java:49)
> 	at java.base/java.lang.Thread.run(Thread.java:844)
> 
> java.nio.channels.CancelledKeyException: null
> 	at java.base/sun.nio.ch.SelectionKeyImpl.ensureValid(SelectionKeyImpl.java:73)
> 	at java.base/sun.nio.ch.SelectionKeyImpl.readyOps(SelectionKeyImpl.java:87)
> 	at java.base/java.nio.channels.SelectionKey.isWritable(SelectionKey.java:310)
> 	at org.bitcoinj.net.ConnectionHandler.handleKey(ConnectionHandler.java:229)
> 	at org.bitcoinj.net.NioClientManager.handleKey(NioClientManager.java:86)
> 	at org.bitcoinj.net.NioClientManager.run(NioClientManager.java:122)
> 	at com.google.common.util.concurrent.AbstractExecutionThreadService$1$2.run(AbstractExecutionThreadService.java:66)
> 	at com.google.common.util.concurrent.Callables$4.run(Callables.java:122)
> 	at org.bitcoinj.utils.ContextPropagatingThreadFactory$1.run(ContextPropagatingThreadFactory.java:49)
> 	at java.base/java.lang.Thread.run(Thread.java:844)
> ```
> 
> The connection is established successfully to the node, but I did get this exception.

Those should be some of the expected exceptions I have not muted. I'll double check and submit PR to mute them.

dmos62:
So the misguiding exceptions pointed out by @ripcurlx are a more difficult problem than I expected. To be clear, they are harmless. They are thrown because we close the Peer while it's still doing something in parallel. The "in parallel" bit is what complicates things. I'll continue investigating. Should also lead to a better understanding of the edge cases. Can't guarantee a solution for Monday.

We can consider, as a temporary measure, a log entry preceding these exceptions that explains them.

Edit: solved; see mentioned PR.

ripcurlx:
I've now also tested the two configuration errors  again on Regtest and it is working as before 👍

ripcurlx:
Ups wrong PR 😉

Issue #3657, opened 2019-11-21
Hangs the whole OS on startup. Ubuntu 19.10


dimaatmelodromru:
### Description

Hangs the **whole fucking OS** on startup @ "Connecting to Tor 1/4". Ubuntu 19.10. *Even mouse pointer hardly moves*.

#### Version

1.2.3

### Steps to reproduce

Just start on my system. Every time.

### Expected behaviour

App opens

### Actual behaviour

System freezes

### Screenshots

none.

#### Device or machine

AMD CPU
[bisq.log](https://github.com/bisq-network/bisq/files/3875146/bisq.log)

#### Additional info

[bisq.log](https://github.com/bisq-network/bisq/files/3875157/bisq.log)

HiDef888:
Same.  
Ubuntu 19.10 

Slowly eats all RAM then moves to SWAP and eventually crashes machine.

clutwo:
Same here... Running Manjaro 18.1.3 with kernel 5.3.11-1.

System completely frozen... no mouse movement, can't even change to TTY. 

This has never happened to me before on this system. Tried starting Bisq twice... had to punch the reset button both times.

devinbileck:
I just tested on Ubuntu 19.10 and did not encounter an issue.
Are you able to try with a new data dir? Maybe your data dir is in a bad state.
Just curious, how many CPU cores and RAM does your machine have?

HiDef888:
> I just tested on Ubuntu 19.10 and did not encounter an issue.
> Are you able to try with a new data dir? Maybe your data dir is in a bad state.
> Just curious, how many CPU cores and RAM does your machine have?

Ryzen 2200g, 16gb RAM, 2gb Swap space.

What type information would be helpful to you?  Do i need to build from source to use a different data directory? 
I'm on kernel 5.3.0-23

Is there perhaps some other logging I could paste in that'd be more helpful?  I watched my resources and it seemed to be a type of memory leak - RAM continued to fill rapidly and then grind to a halt.  My system hovers around 4gb of RAM usage until I open bisq, then it consumes it ALL.

devinbileck:
Yes there is a known memory leak. See #3128 for more details.

RobotDyn:
8GB RAM, AMD Ryzen

dimaatmelodromru:
8GB RAM, AMD CPU

ripcurlx:
[...]

Issue #3128, opened 2019-08-22
Memory Leak - Ubuntu 18.04 LTS


bettyvschmartz:
For the last couple of months Bisq has become unusable. Upon load, memory consumption increases until it hits 100% (32GB!) of RAM and the application crashes. This is true for all versions from 0.9.8 to 1.1.5.

I have tried installing Oracle Java 8 and 12, as well as using OpenJDK 8 and 11 but none of these have any effect. Have also tried removing the Bisq directory and starting afresh but with no joy. 

Things were working well previously so assuming maybe an OS package update somewhere. 

Strange issue. Any help appreciated. 

OS: Using Ubuntu 18.04 LTS
RAM: 32GB

-- BVS

huey735:
I have no idea what may be causing this but is this an old wallet? If so it may be worth it moving into a new directory with a new fresh wallet.
I have an old laptop with 8GB RAM and old Intel CPU and both my local Bitcoin node and Bisq running don't go over 2GB.

christophsturm:
can you please create a memory dump? https://www.baeldung.com/java-heap-dump-capture

https://visualvm.github.io

bettyvschmartz:
No problem. I am using VisualVM; what specifically would you like from the memory dump as aware this could contain sensitive information or private keys? 

--- BVS

christophsturm:
You could just create a memory snapshot with visualvm and take a screenshot of it

bettyvschmartz:
What is strange @christophsturm is that the Bisq process only consumes just over 1GB but the system memory whilst running is fully consumed. As soon as Bisq is closed the RAM usage return to normal or Bisq exits firstly (probably out of memory error). See below:

_Total Memory Usage_

![Screenshot-20190827200354-736x160](https://user-images.githubusercontent.com/10772184/63801812-12465a80-c909-11e9-984e-e4506a21bce2.png)

_Bisq Memory Usage_

![Screenshot-20190827200242-737x101](https://user-images.githubusercontent.com/10772184/63801873-3efa7200-c909-11e9-811a-dc6a3c93cebb.png)

_VisualVM:_

![Screenshot-20190827203114-819x364](https://user-images.githubusercontent.com/10772184/63802104-c5af4f00-c909-11e9-80dd-4ee152004398.png)

No other processes are using any significant amount of memory. 

-- BVS

bettyvschmartz:
[...]

PR #4060, opened 2020-03-13
Improve AvoidStandyModeService

ghubstan:
This change substitutes use of memory efficient, OS supplied sleep/suspend
inhibitor utilities for the silent audio file player on Linux, merges
OSXStandbyModeDisabler functionality into AvoidStandyModeService, and
removes that class.

This change also stops Bisq from running the audio file on OSX;  it is
currently running both caffeinate and the silent audio file, and the avoid
standby mode button in the the OSX settings view is currently toggling
the audio player on and off, but not caffeinate.  (The only way to shut
down caffeinate is by shutting down Bisq.)

The OSX avoid standby mode button button has not been hidden so a cached
'do not avoid standby mode' preference does not leave the user stuck without
a caffeinate service the next time they start Bisq. (They can use it to
turn caffeinate on, but it can't be used to turn it off without shutting
down Bisq too.)

The avoid standy mode button is now displayed on Linux because the native
inhibitor can be toggled on and off.

There is no change to the avoid shutdown service on Windows and Unix.

OSX and UI related changes are described in the commit message above.  An 
explanation of Linux OS related changes follows.

On linux hosts, AvoidStandbyService checks for executables
gnome-session-inhibit and systemd-inhibit in that order, then starts
the first found to prevent suspend or sleep while Bisq is running.

To check if an inhibitor is enabled or disabled in a gnome terminal...

	$ dbus-send --session --print-reply --dest=org.gnome.SessionManager \
		/org/gnome/SessionManager org.gnome.SessionManager.GetInhibitors

If Bisq falls back to systemd-inhibit on a non gnome desktop...

	$ systemd-inhibit --list
	$ systemd-inhibit --list | grep Bisq
	
If neither gnome-session-inhibit nor systemd-inhibit are found, the
standby service falls back to playing the silent audio file.

This does not cover all linux desktop types, but other inhibitors
exist for desktops I have not set up, such as mate.  It should not
be difficult to support them.

Off heap memory use reduction varies from machine to (virtual) machine.
On my Ubuntu 18 laptop, playing the audio file to block sleep/suspend
is  allocating about 1 GB of native memory per hour via malloc() calls.
On the same machine after warm-up, Bisq's  virtual memory usage goes from 
10.4 GB to 10.1 GB.  (This figure was tracked before the MaxRAM=4g setting 
was merged from PR #4048.)

This has been tested on OSX Catalina 10.15, Ubuntu 18 (gnome) and the following VMs

	Debian 10.3 (gnome)
	Arch Linux 9.2.1 (gnome)
	Fedora 31.1.9 (cinnamon)

However, VMs not running an inhibitor or audio file do not always go into suspend mode
as schedules.

This does not work on Windows Cygwin, but minor changes might make it work if
cygwin inhibitor tools like gnome-session-inhibit.exe and mate-session-inhibit.exe
are installed.

Paritally addresses off-heap memory consumption issues #3128, #3657, 
#3918, #3917, #3686, #3677

This PR replaces #4049


ripcurlx:
@wiz @freimair Could you give this a spin?

freimair (approved):
here on a freshly upgraded Arch linux with gnome shell desktop.

I can confirm that the inhibitor works, although it seems that at least arch linux relies on the systemd mechanics as a primary tool for inhibiting shutdown/hibernate/sleep/...

I further can confirm that disk I/O has dropped by about 50%, at least the read part. Previously, I have seen that the sound file got read very often, it is gone now.

This is a green from me! Thanks for your efforts!

ripcurlx (approved):
ACK - Tested it on macOS 10.15.3 and it worked as well.

ripcurlx:
@ManfredKarrer Do you remember why you used the audio workaround and not caffeinate on macOS? Or was it just the best cross-platform solution to do it like that?

ripcurlx:
Checked back with Manfred and he doesn't have specific concerns about this change as he didn't use caffeinate in his original solution.

ghubstan:
Trying to reopen this PR to force a new Travis CI build and see if it fixes  JDK12 build failure:  
`Could not resolve com.google.protobuf:protoc:3.10.0`

See https://travis-ci.org/github/bisq-network/bisq/jobs/669606673?utm_medium=notification&utm_source=github_status

The JDK 10 build passed, and it also passed on my own JDK11.

cbeams:
I just reviewed this change now, and was initially wondering why we didn't delete the `prevent-app-nap-silent-sound.aiff` file. Looking deeper, I see that it's still used on Windows. This is mentioned above:

> There is no change to the avoid shutdown service on Windows and Unix.

Just wondering: was any investigation done as to how to inhibit sleep on Windows?

ghubstan:
> Just wondering: was any investigation done as to how to inhibit sleep on Windows?

@cbeams, I did not investigate how to suspend sleep on Win because don't have a windows OS.

PR #4062, opened 2020-03-14
Mute unnecessary logging and minor improvements

dmos62:
Replaces NioClient and NioClientManager with BlockingClient, which
gets rid of misguiding log entries as reported in #4058 PR's comments by
@ripcurlx. Also introduces a few minor changes including better logs and
clearer code.

### DIFF
--- a/core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
+++ b/core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
@@ -8,11 +8,13 @@ import org.bitcoinj.core.Peer;
 import org.bitcoinj.core.PeerAddress;
 import org.bitcoinj.core.VersionMessage;
 import org.bitcoinj.core.listeners.PeerDisconnectedEventListener;
-import org.bitcoinj.net.NioClient;
-import org.bitcoinj.net.NioClientManager;
+
+import org.bitcoinj.net.BlockingClient;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
+
+import javax.net.SocketFactory;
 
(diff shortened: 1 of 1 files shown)


ripcurlx (changes requested):
NACK

I just tested it on Mainnet with and without a local Bitcoin Core.
Running a well configured Bitcoin Core everything looks fine, but when I start the node as a light client without a Bitcoin Core running I'll always get this warning log:
```
März-15 11:44:21.452 [JavaFX Application Thread] ERROR b.c.btc.nodes.LocalBitcoinNode: Exploratory handshake attempt with a local Bitcoin node (that may not be there) unexpectedly timed out. This should never happen; please report this. Continuing as if a local BTC node was not found. 
```
If that will be triggered all the time no matter if there is a local bitcoin core node running or not, I think the text should be changed, that this is expected if no local BTC node is running on this system.

ripcurlx:
I've now also tested the two configuration errors again on Regtest and it is working as before 👍

dmos62:
@ripcurlx that log error is about BitcoinJ hanging, which is not normal. I've not experienced this. @wiz reported this first, but in his case it was when a local BTC node wasn't running. You experiencing it when there is a local BTC node running is troubling.

Could you clarify your second comment? Did the bug go away on its own?

Starting to think this implementation is cursed. The bright side is that if we generalize configuration checking to remote nodes as well, we'll be able to do away with all this additional IO. I've realised that this implementation turned out like an awkward appendage, because I felt intimidated hooking into Bisq's regular BTC networking. A lot would have been simpler.

What's kind of troubling is that if LocalBitcoinNode's usage of BitcoinJ is not somehow incorrect, these weird BitcoinJ hangs might be happening in other parts of Bisq too.

PR #4074, opened 2020-03-17
Update ops team bonded role titles to match current names

wiz:
* Rename "Data relay node operator" to "Price node operator"
* Rename "Markets operator" to "Markets API operator"
* Rename "BTC node operator" to "Bitcoin node operator"


ripcurlx (approved):
utACK

Issue #4080, opened 2020-03-21
Our BitcoinJ fork sometimes fails to connect because of deprecated BTC alert system


dmos62:
In some cases BitcoinJ throws a `ProtocolException`, crashing the connection, because it receives an alert message during the handshake [0]. The alert message is sent because Bitcoin Core >= 0.14.0 is telling us to upgrade to atleast version 70012 [1][2].

It would be difficult to predict the consequences of changing the advertised version. We could just have BitcoinJ always ignore alert messages (which are deprecated and only useful/used for locking out very old clients). It seems that our fork already is attempting to do that, and the handshake case is an oversight.

I've discovered this when testing a local Bitcoin node that is set to prune. So far this is the only setup where I've received an alert message during handshake. I couldn't discover why.

[0] https://github.com/bisq-network/bitcoinj/blob/bisq_0.14.7/core/src/main/java/org/bitcoinj/core/Peer.java#L474
[1] https://github.com/bitcoin/bitcoin/commit/8ff8d219c3d9a0b6f604bb99ca4a27ad77cdb69b
[2] https://en.bitcoin.it/wiki/Alert_system

wiz:
Yeah, it also reproduces if you `./bisq-desktop --baseCurrencyNetwork=BTC_TESTNET`

PR #4081, opened 2020-03-21
Check BTC node config without IO overhead

dmos62:
Edit: not proof of concept anymore; this initial post is outdated. TLDR is that this PR has Bisq check if nodes it connects to are suitable for use by Bisq (e.g. that the node isn't pruning). That's currently most useful with a local BTC node or custom nodes (because otherwise Bisq uses our own federated nodes that are known to be well configured), but it checks all BTC nodes it connects to, with negligible overhead. This PR is currently stagnating in large part because it depends on a (minor) BitcoinJ PR.

This is a proof of concept for checking BTC node's configuration. It's an alternative approach to what I had proposed earlier (that was merged and reverted due to reliability issues). In contrast to the previous attempt, this approach doesn't incur additional IO and doesn't use blocking. Whereas the previous attempt initiated a short lived BitcoinJ connection to a node, this approach just listens into the connections we are already establishing. Also, while previously the checking applied only to a local BTC node, this approach generalizes checking to all BTC nodes we connect to.

Currently, it has two problems:

a) connecting to a pruning node, on my setup at least, causes the connection to crash, and the listener never gets triggered; this is a rare BitcoinJ edge case bug and I opened an issue for it https://github.com/bisq-network/bisq/issues/4080 ;

and, b) this WIP implementation just does `peer.close()`, which causes the PeerGroup to just retry the connection, in case we're in local mode, or possibly retry later in case we're in non-local mode (didn't explore non-local behaviour yet); I'm pondering if this is indeed undesirable or if we would want to use this to get free auto-retry.

Another open question is what to do in case a remote node is found to be misconfigured. Just an error log?

### DIFF
--- a/core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
+++ b/core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
@@ -46,6 +46,7 @@ import org.bitcoinj.core.Context;
 import org.bitcoinj.core.NetworkParameters;
 import org.bitcoinj.core.Peer;
 import org.bitcoinj.core.PeerAddress;
+import org.bitcoinj.core.VersionMessage;
 import org.bitcoinj.core.PeerGroup;
 import org.bitcoinj.core.RejectMessage;
 import org.bitcoinj.core.listeners.DownloadProgressTracker;
@@ -91,10 +92,12 @@ import java.util.Date;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
 import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 
 import lombok.Getter;
+import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 
 import org.jetbrains.annotations.NotNull;
@@ -140,6 +143,10 @@ public class WalletsSetup {
     public final BooleanProperty shutDownComplete = new SimpleBooleanProperty();
     private final boolean useAllProvidedNodes;
     private WalletConfig walletConfig;
+
+    @Setter
+    @Nullable
+    private Consumer<String> displayUserBtcNodeMisconfigurationHandler;
 
     ///////////////////////////////////////////////////////////////////////////////////////////
     // Constructor
@@ -214,6 +221,10 @@ public class WalletsSetup {
                 // We don't want to get our node white list polluted with nodes from AddressMessage calls.
                 if (preferences.getBitcoinNodes() != null && !preferences.getBitcoinNodes().isEmpty())
                     peerGroup.setAddPeersFromAddressMessage(false);
+
+                peerGroup.addVersionMessageReceivedEventListener((peer, versionMessage) -> {
+                    UserThread.execute(() -> handleConfiguration(peer, versionMessage));
+                });
 
                 peerGroup.addConnectedEventListener((peer, peerCount) -> {
                     // We get called here on our user thread
@@ -409,6 +420,64 @@ public class WalletsSetup {
         networkConfig.proposePeers(peers);
     }
 
+    ///////////////////////////////////////////////////////////////////////////////////////////
+    // Peer configuration check
+    ///////////////////////////////////////////////////////////////////////////////////////////
+
+    /* Provided a Peer and its VersionMessage, will check and handle Peer's advertised
+     * configuration. The VersionMessage is provided separately, because it is not
+     * always available and this is meant to be called from a listener that is triggered
+     * when the VersionMessage becomes available.
+     */
+    private void handleConfiguration(Peer peer, VersionMessage versionMessage) {
+        var wellConfigured = isWellConfigured(versionMessage);
+        if (wellConfigured) {
+            log.info("Peer well configured: {}", peer);
+        } else {
+            log.error("Peer _badly_ configured (pruning and/or bloom filters disabled): {}", peer);
+            var usingLocalNode = localBitcoinNode.shouldBeUsed();
+            var usingCustomNodes =
+                BtcNodes.BitcoinNodesOption.CUSTOM.ordinal() == preferences.getBitcoinNodesOptionOrdinal();
+            if (usingLocalNode || usingCustomNodes) {
+                displayUserBtcNodeMisconfigurationHandler.accept(peer.toString());
+            }
+        }
+    }
+
+    private static boolean isWellConfigured(VersionMessage versionMessage) {
+        var notPruning = versionMessage.hasBlockChain();
+        var supportsAndAllowsBloomFilters =
+            isBloomFilteringSupportedAndEnabled(versionMessage);
+        return notPruning && supportsAndAllowsBloomFilters;
+    }
+
+    /**
+     * Method backported from upstream bitcoinj: at the time of writing, our version is
+     * not BIP111-aware. Source routines and data can be found in bitcoinj under:
+     * core/src/main/java/org/bitcoinj/core/VersionMessage.java
+     * and core/src/main/java/org/bitcoinj/core/NetworkParameters.java
+     */
+    @SuppressWarnings("UnnecessaryLocalVariable")
+    private static boolean isBloomFilteringSupportedAndEnabled(VersionMessage versionMessage) {
+        // A service bit that denotes whether the peer supports BIP37 bloom filters or
+        // not. The service bit is defined in BIP111.
+        int NODE_BLOOM = 1 << 2;
+
+        int BLOOM_FILTERS_BIP37_PROTOCOL_VERSION = 70000;
+        var whenBloomFiltersWereIntroduced = BLOOM_FILTERS_BIP37_PROTOCOL_VERSION;
+
+        int BLOOM_FILTERS_BIP111_PROTOCOL_VERSION = 70011;
+        var whenBloomFiltersWereDisabledByDefault = BLOOM_FILTERS_BIP111_PROTOCOL_VERSION;
+
+        int clientVersion = versionMessage.clientVersion;
+        long localServices = versionMessage.localServices;
+
+        if (clientVersion >= whenBloomFiltersWereIntroduced
+                && clientVersion < whenBloomFiltersWereDisabledByDefault)
+            return true;
+
+        return (localServices & NODE_BLOOM) == NODE_BLOOM;
+    }
 
     ///////////////////////////////////////////////////////////////////////////////////////////
     // Backup
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -2786,6 +2786,7 @@ Please check if you have the latest version of Bisq installed.\n\
 You can download it at: [HYPERLINK:https://bisq.network/downloads].\n\n\
 Please restart the application.
 popup.warning.startupFailed.twoInstances=Bisq is already running. You cannot run two instances of Bisq.
+popup.warning.userBtcNodeMisconfigured.explanation=The user-provided Bitcoin node {0} was found to be misconfigured. A user-provided BTC node is either a locally running node that was automatically detected and used, or a node explicitly specified via the --btcNodes command-line option. Bisq requires BTC nodes to have bloom filters enabled and pruning disabled. Automatic use of a local Bitcoin node can be disabled by using the --ignoreLocalBtcNode option.
 popup.warning.tradePeriod.halfReached=Your trade with ID {0} has reached the half of the max. allowed trading period and is still not completed.\n\nThe trade period ends on {1}\n\nPlease check your trade state at \"Portfolio/Open trades\" for further information.
 popup.warning.tradePeriod.ended=Your trade with ID {0} has reached the max. allowed trading period and is not completed.\n\n\
   The trade period ended on {1}\n\n\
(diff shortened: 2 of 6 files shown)


ripcurlx:
@dmos62 Your build is broken right now: https://travis-ci.org/github/bisq-network/bisq/builds/665235608?utm_source=github_status&utm_medium=notification

> a) connecting to a pruning node, on my setup at least, causes the connection to crash, and the listener never gets triggered; this is a rare BitcoinJ edge case bug and I opened an issue for it #4080 ;

Yes, I get following exception in the logs
```
März-23 15:46:57.444 [NioClientManager] WARN  o.b.core.PeerSocketHandler: [127.0.0.1]:18444 -  org.bitcoinj.core.ProtocolException: Received AlertMessage before version handshake is complete.
	at org.bitcoinj.core.Peer.processMessage(Peer.java:478)
	at org.bitcoinj.core.PeerSocketHandler.receiveBytes(PeerSocketHandler.java:184)
	at org.bitcoinj.net.ConnectionHandler.handleKey(ConnectionHandler.java:223)
```
> 
> and, b) this WIP implementation just does `peer.close()`, which causes the PeerGroup to just retry the connection, in case we're in local mode, or possibly retry later in case we're in non-local mode (didn't explore non-local behaviour yet); I'm pondering if this is indeed undesirable or if we would want to use this to get free auto-retry.

You mean a retry to the same peer?

> Another open question is what to do in case a remote node is found to be misconfigured. Just an error log?

It depends. If it would be a misconfigured provided node/public node I would just ignore it and not connect to it. If it is a node the user has configured by using a config/start parameter I think we should display a warning to the user. Also if it is a misconfigured local node we want to notify the user that something is not correct.

In general if we do this check after the connection I think it should be fine if we block the setup process at that point (e.g. when a local bitcoin node is misconfigured) and do it as with your previous implementation to allow the user to shutdown and fix this issue.

dmos62:
>> and, b) this WIP implementation just does peer.close(), which causes the PeerGroup to just retry the connection, in case we're in local mode, or possibly retry later in case we're in non-local mode (didn't explore non-local behaviour yet); I'm pondering if this is indeed undesirable or if we would want to use this to get free auto-retry.
> 
>You mean a retry to the same peer?

Yes. I didn't double-check, but I think BitcoinJ uses a FILO queue for keeping addresses of nodes. It pops and connects to them until wanted number of connections is established. In case we're in local-mode, then that queue has size 1 and contains localhost. `peer.close()` probably causes the node to go to the back of the queue.

> block the setup process [when a local bitcoin node is misconfigured] and do it as with your previous implementation to allow the user to shutdown and fix this issue

I'd be in favor of letting an unclosable error popup be displayed asynchroniously. That would involve less fiddling with the setup process (less likely to break something).

ripcurlx:
> I'd be in favor of letting an unclosable error popup be displayed asynchroniously. That would involve less fiddling with the setup process (less likely to break something).

I think this should be fine as well.

dmos62:
This PR now depends on this PR on our BitcoinJ fork: https://github.com/bisq-network/bitcoinj/pull/35. For development I used a BitcoinJ build from a local Maven repository. When that PR for BitcoinJ is merged, relevant changes to `build.gradle` should be made here.

Screenshot:

user provided node misconfigured

As seen in the screenshot, we're not providing a button to revert to not using a user-provided node (don't see how that can be done easily). The reason why before we could offer that option before is that we were blocking on the check and we were only handling the local bitcoin node.

It can also be seen that Bisq is actually loaded, and after you click "shutdown" the informational local BTC node popup will appear briefly before Bisq closes.

In the background, the misconfigured peers are not explicitly closed (BitcoinJ closes them automatically if they're pruning). The errors are logged and if custom BTC nodes are used, or a local BTC node, the above error popup locks the user out.

In other words, advantage over previous tries at getting this implemented, in terms of features, is that now we get errors when remote nodes are misconfigured, and the user is locked out if the misconfigured node is local or custom. Before we were only checking the local Bitcoin node.

dmos62:
Bump

chimp1984:
@dmos62 Sorry that the PR got delayed so long... I asked @oscarguindzberg to review your BitcoinJ PR. If that is done I think we should get that PR in as well asap.

core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
```diff
@@ -224,11 +228,12 @@ protected void onSetupCompleted() {
                 if (preferences.getBitcoinNodes() != null && !preferences.getBitcoinNodes().isEmpty())
                     peerGroup.setAddPeersFromAddressMessage(false);
 
+                peerGroup.addVersionMessageReceivedEventListener(WalletsSetup.this::handleConfiguration);
```
chimp1984: We get called from the BitcoinJ thread. We should map it to our userThread to avoid threading issues.
```
peerGroup.addVersionMessageReceivedEventListener((peer, versionMessage) ->
                        UserThread.execute(handleConfiguration(peer, versionMessage)));
```

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -2527,6 +2527,7 @@ It will make debugging easier if you include the bisq.log file by pressing "Open
 
 popup.error.tryRestart=Please try to restart your application and check your network connection to see if you can resolve the issue.
 popup.error.takeOfferRequestFailed=An error occurred when someone tried to take one of your offers:\n{0}
+popup.error.userBtcNodeMisconfigured.explanation=The user-provided Bitcoin node {0} was found to be misconfigured. A user-provided BTC node is either a locally running node that was automatically detected and used, or a node explicitly specified via the --btcNodes command-line option. Bisq requires BTC nodes to have bloom filters enabled and pruning disabled. Automatic use of a local Bitcoin node can be disabled by using the --ignoreLocalBtcNode option.
```
chimp1984: I think we should not show an Error popup as it comes with the bug-report text and is confusing for that use case, even it is somehow an error as Bisq cannot be used in that case. I would suggest to use warning popup and to reflect that in the display string we should use `popup.warning.userBtcNodeMisConfigured.explanation`. Also changed to camelCase as IDEA static code inspector complains about a typo otherwise. I think misconfigured should be mis-configured in the text.

desktop/src/main/java/bisq/desktop/main/MainViewModel.java
```diff
@@ -401,6 +401,15 @@ private void setupHandlers() {
         daoPresentation.getBsqSyncProgress().addListener((observable, oldValue, newValue) -> updateBtcSyncProgress());
 
         bisqSetup.setFilterWarningHandler(warning -> new Popup().warning(warning).show());
+
+        walletsSetup.setDisplayUserBtcNodeMisconfigurationHandler(
+                (String peer) ->
+                    new Popup()
+                        .hideCloseButton()
+                        .error(Res.get("popup.error.userBtcNodeMisconfigured.explanation", peer))
```
chimp1984: As per prev. comment:
-> ` .warning(Res.get("popup.warning.userBtcNodeMisConfigured.explanation", peer))`

core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
```diff
@@ -397,45 +402,27 @@ private void configPeerNodes(@Nullable Socks5Proxy proxy) {
...
-      return Futures.transform(versionMessageFuture, handler);
-    }
-
-    private static Void handleConfiguration(Peer peer, VersionMessage versionMessage) {
+    /* Provided a Peer and its VersionMessage, will check and handle Peer's advertised
+     * configuration. The VersionMessage is provided separately, because it is not
+     * always available and this is meant to be called from a listener that is triggered
+     * when the VersionMessage becomes available.
+     */
+    private Void handleConfiguration(Peer peer, VersionMessage versionMessage) {
```
chimp1984: The uppercase Void is probably a leftover from the previous code. Should be `void`.

core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
```diff
@@ -397,45 +402,27 @@ private void configPeerNodes(@Nullable Socks5Proxy proxy) {
...
-        peer.close();
+        log.error("Peer _badly_ configured (pruning and/or bloom filters disabled): {}", peer);
+        var usingLocalNode = localBitcoinNode.shouldBeUsed();
+        var usingCustomNodes =
+            BtcNodes.BitcoinNodesOption.CUSTOM.ordinal() == preferences.getBitcoinNodesOptionOrdinal();
+        if (usingLocalNode || usingCustomNodes) {
+            displayUserBtcNodeMisconfigurationHandler.accept(peer.toString());
+        }
       }
       return null;
```
chimp1984: If `void` is used above the return need to be removed.

chimp1984 (changes requested):

core/src/main/resources/i18n/displayStrings.properties
dmos62: Warning doesn't seem entirely appropriate, giving the event a milder air than it might deserve (event being not something to keep in mind, but something to fix), but it might be more appropriate than Error, which seems to be dedicated to events that should be bug-reported. I'll switch to Warning.

I think the spelling of "misconfigure" is correct. "Mis" + "construct" is a rarer combination, so it's not found on many dictionaries, but you can find more common uses of the prefix mis in entries like "misplace" [0] or "misconduct" [1]. If you take a look at the descriptions of the prefix mis- in popular dictionaries [2], it's normal to use it without a hyphen. I think a hyphen is used uniquely when connecting entire words (something like the hyphenation in post-modernism [3] is a discouraged spelling: redirects to postmodernism).

> IDEA static code inspector complains about a typo

Can words be added to its dictionary, like you would in a spellchecking text editor? It would be awkward to alter writing to match a spellchecker.

[0] https://www.thefreedictionary.com/misplace
[1] https://www.thefreedictionary.com/misconduct
[2] https://www.thefreedictionary.com/mis-
[3] https://www.thefreedictionary.com/post-modernism

chimp1984: You can set the header message as well, so the title is not warning but something more appropriate. The error type popup carries the automatic support for bug report...

chimp1984: Re misconfigure: Ah ok. I am not native speaker and IDEA suggestions is my teacher ;-).
> Can words be added to its dictionary
Yes.

chimp1984 (dismissed):
utACK

Thanks for the update!

chimp1984:
@ripcurlx This is tagged as 1.4.0 can u remove that? Can u remove also the dopped tag. It is still an important PR. I guess we are now ready from the bitcoinj side to complete it. 
@dmos62  can you check with @oscarguindzberg if there is anything missing?

chimp1984:
@dmos62 Do you have time to continue work on that?

dmos62:
@chimp1984 Yep. I'll get in touch with @oscarguindzberg.

dmos62:
I've ported this PR and [a BitcoinJ PR](https://github.com/bisq-network/bitcoinj/pull/36) that this depends on to BitcoinJ 15.8. No noteworthy changes required.

This PR currently needs a locally installed BitcoinJ library instance, for development purposes (you have to `git checkout` the above mentioned BitcoinJ PR and run `gradle install` on it). Hence Travis failing. Once the BitcoinJ PR is merged, we can update this PR to use our Github hosted fork.

core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
```diff
@@ -409,6 +420,64 @@ private void configPeerNodes(@Nullable Socks5Proxy proxy) {
...
+    ///////////////////////////////////////////////////////////////////////////////////////////
+
+    /* Provided a Peer and its VersionMessage, will check and handle Peer's advertised
+     * configuration. The VersionMessage is provided separately, because it is not
+     * always available and this is meant to be called from a listener that is triggered
+     * when the VersionMessage becomes available.
+     */
+    private void handleConfiguration(Peer peer, VersionMessage versionMessage) {
+        var wellConfigured = isWellConfigured(versionMessage);
+        if (wellConfigured) {
```
kaiwolfram, line 434: A temp variable is not necessary imo, the method call is already very concise and readable
```suggestion
        if (isWellConfigured(versionMessage)) {
```

core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
```diff
@@ -409,6 +420,64 @@ private void configPeerNodes(@Nullable Socks5Proxy proxy) {
...
+            log.info("Peer well configured: {}", peer);
+        } else {
```
kaiwolfram, line 436: Redundant `else` clause bc the `if` can return. Makes it more readable
```suggestion
        if (wellConfigured) {
            log.info("Peer well configured: {}", peer);
            return;
        }
```

core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
dmos62: A variable defining the condition is a style I use (and propagate) and, even if the called method is descriptive as well, I still use it. One reason is consistency of style. Another, the if statement doesn't care that the node being well configured is predicated on the version message (that's an implemenation detail), so there's no reason to put that in the if statement. I would be against your suggestion, because it optimizes for terseness at the cost of readability.

core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
dmos62: If you don't remove the `wellConfigured` intermediary variable as you suggested above (which I am against), this originally reads "if the node is well configured log that, else perform following steps", which is favourable over your suggetion: "if the node is well configured, given provided version message, exit the method, foregoing steps that would only be executed in case the node were not well configured." Your edits have a negative effect on readability and a very minimal effect on terseness, which, in my opinion, should not be a consideration anyway (but that's besides the point).

Issue #4086, opened 2020-03-21
Transitive dependencies creating version conflicts & missing jar verification


ghubstan:
### Description

The gradle build needs to be more strict about excluding transitive dependencies that create version conflicts.  (This implies some dependencies might need to be upgraded, and some even downgraded, to align versions of direct and transitive dependencies.)

#### Version

Commit 17bb7b4

### Steps to reproduce

Build Bisq and see multiple versions of some jars in the lib dir.

### Expected behaviour

There should not be any version conflicts, and every jar needs to be verified by the gradle-witness.gradle file.

### Actual behaviour

There are several transitive dependencies downloaded during the build that conflict with direct dependencies specified in gradle.build, and _in the case of the desktop build_, they are not being verified by gradle-witness.  

Here are duplicate jars with the letters 'grpc' in their name:

    grpc-context-1.10.1.jar
    grpc-context-1.25.0.jar
    grpc-core-1.10.1.jar
    grpc-core-1.25.0.jar
    grpc-netty-shaded-1.10.1.jar
    grpc-netty-shaded-1.25.0.jar
    grpc-protobuf-1.10.1.jar
    grpc-protobuf-1.25.0.jar
    grpc-protobuf-lite-1.10.1.jar
    grpc-protobuf-lite-1.25.0.jar
    grpc-stub-1.10.1.jar
    grpc-stub-1.25.0.jar
    opencensus-contrib-grpc-metrics-0.11.0.jar
    opencensus-contrib-grpc-metrics-0.21.0.jar

This may not be an urgent problem because the latest jar is loaded at run time. However, that also means that a transitive dependency with a larger version than is specified in the build file is loaded, not the jar which has passed gradle witness verification.  For example:  the build file specifies

    slf4jVersion = '1.7.22'

but 

    lsof -Pan -p $(pgrep -f BisqAppMain)        ( on linux )

shows me that the transitive dependency slf4j-api-1.7.25.jar is being loaded instead.

Maybe building the shipped, shaded jar eliminates most of this problem, but I new devs like me may spend a lot of time trying to figure out why jar version conflicts like the grpc example I described above exist.

ghubstan:
I don't think it will be possible to resolve all version conflicts because too many transitive dependencies have too many of their own version conflict problems in their own dependency graphs.  But it would be good to minimize them in Bisq as much as practicable.

cd2357:
@ghubstan is this still an issue?

ghubstan:
[...]

PR #4087, opened 2020-03-21
Upgrade to slf4j 1.7.25

ghubstan:
Align Bisq's slf4j version requirement with that of other dependencies

This removes a duplicated dependency, and corrects the gradle-witness.gradle
slf4-api entry, which verified slf4-api v1.7.22, not the transitive dependency
loaded at runtime.

Partial fix for #4086


ghubstan:
slf4j upgrade included in pr 4088

PR #4088, opened 2020-03-21
Upgrade firebase-admin, avoid gRPC version conflict

ghubstan:
The latest firebase-admin (v6.12.2) depends on grpc-api v1.24.1,
which is as close as we can get to the grpc v1.25.0 specified in
build.gradle.  Currently, the firebase-admin version requirement
is causing the build to download grpc v1.10.1 jars, but at runtime,
grpc 1.25.0 jars are loaded.

Transitive dependencies grpc-netty-shaded, grpc-protobuf, grpc-protobuf-lite, 
grpc-protobuf-stub, grpc-protobuf-grpclb (v1.23.0) are also excluded.

Eliminating version conflicts saves a dev time thinking about which
source jar to choose from while examining exception stack traces.

Partial fix for #4086

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,7 +34,7 @@ configure(subprojects) {
         easybindVersion = '1.0.3'
         easyVersion = '4.0.1'
         findbugsVersion = '3.0.2'
-        firebaseVersion = '6.2.0'
+        firebaseVersion = '6.12.2'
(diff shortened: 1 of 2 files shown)


ghubstan:
This has to be cancelled because it breaks bisq-daemon.  

PR #4089 is OK (it just upgrades apache and logback deps.)

PR #4090, opened 2020-03-23
Only use InputValidation when formatting voting stake

ripcurlx:
As this is annoying me every time after voting I've added this no-priority fix.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/dao/governance/proposals/ProposalsView.java
+++ b/desktop/src/main/java/bisq/desktop/main/dao/governance/proposals/ProposalsView.java
@@ -54,9 +54,10 @@ import bisq.core.dao.state.model.governance.Vote;
 import bisq.core.locale.Res;
 import bisq.core.user.Preferences;
 import bisq.core.util.FormattingUtils;
-import bisq.core.util.coin.BsqFormatter;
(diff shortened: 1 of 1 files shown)


desktop/src/main/java/bisq/desktop/main/dao/governance/proposals/ProposalsView.java
```diff
@@ -554,6 +555,7 @@ private void updateViews() {
                 if (optionalMyVote.isPresent()) {
                     MyVote myVote = optionalMyVote.get();
                     Coin stake = Coin.valueOf(myVote.getBlindVote().getStake());
+                    stakeInputTextField.setValidator(new InputValidator());
```
sqrrm, line 558: Shouldn't this be set at line 676 instead of
```
        stakeInputTextField.setValidator(new BsqValidator(bsqFormatter));
```

ripcurlx: No the validator there is fine. Just after submitting the vote the value entered is formatted with `BSQ` in the end which causes the original validator to fail. I think it is good that it is formatted with the code in this case so I didn't remove the formatting, but decided to adapt the validator.

sqrrm (approved):
utACK

PR #4091, opened 2020-03-23
Suppress netty.shaded debug stack traces in cli

ghubstan:
Configure cli loggers to avoid scary but ignorable netty debug
log messages with stack traces:  "Reflective setAccessible(true)",
and those showing unsuccessful netty native lib load attempts.
It may be preferable to not use logback in cli, and achieve the
same result programmatically with slf4j, but this shows how to
hide this ugly log output in the meantime.

Remove comment about why these messages were logged, and switch
to non-deprecated ManagedChannelBuilder.forAddress(host, port)
		.usePlaintext().build()
in BisqCliMain.

### DIFF
--- a/cli/src/main/java/bisq/cli/app/BisqCliMain.java
+++ b/cli/src/main/java/bisq/cli/app/BisqCliMain.java
@@ -57,10 +57,6 @@ import static java.lang.String.format;
 
 /**
  * gRPC client.
- *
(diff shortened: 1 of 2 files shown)


cbeams (dismissed):
utACK

ripcurlx:
@ghubstan Could you please resolve the conflict caused by a recent merge. Thanks!

ghubstan:
@ripcurlx , merge conflict resolved

ghubstan:
Closing because merged PR #4097 has this change.

Issue #4039, opened 2020-03-10
Bisq does not add "dust" outputs to miner fee


wiz:
### Background

According to Bitcoin consensus rules, no output can be less than 546 sats. If a transaction with such an output is broadcast, it will be rejected.

### Description

Bisq attempts to broadcast Bitcoin transactions containing "dust" outputs, instead of simply adding the dust to the miner fee like most wallets do.

#### Version

v1.2.7

### Steps to reproduce

This is the difficult part, as users randomly do this if their transaction coincidentally results in an output of less than 546 sats. However, since Bisq assumes transactions it creates are valid (without checking the dust rule), this results in triggering other random Bisq issues. Sometimes it results in a failed trade, sometimes it results in a corrupted wallet.

### Expected behaviour

The dust output should not be created, and instead added to fees.

### Actual behaviour

The dust output is created, and the Bitcoin transaction is rejected, causing further issues in Bisq depending on the situation.

ripcurlx:
> The dust output is created, and the Bitcoin transaction is rejected, causing further issues in Bisq depending on the situation.

So you had actual transactions with outputs below the dust limit? I though we had some dust check in place already during transaction creation, so I just wanted to make sure before someone digs into this codebase.

One more thing: to prevent confusion what is up for compensation and what isn't please don't add issues to the `critical bugs board` on your own. Otherwise devs might pick it up by mistake and not getting compensated if the issue is not getting the `is:critical` label in the end.

wiz:
Yes, this is a common recurring issue in Bisq, @huey735 how many cases would you say you've seen caused by "the dust issue" ?

huey735:
@wiz recently I can at least think of these two:
* https://github.com/bisq-network/support/issues/337
* https://www.reddit.com/r/bisq/comments/fgwu59/tx_rejected_for_dust_bisq_not_connecting_to_main/

The first one is slightly different from what this issue is about but it also demonstrates Bisq inability to manage dust values.

ghost:
### example in P2P wallet

1) Bob chooses send from a single UTXO containing 0.1 BTC
2) he sends 0.099995 BTC to Alice, with "amount includes mining fee" set.
3) A transaction is generated and written to the log file:
   0.09993624 to Alice
   0.00005876 mining fee
   0.00000500 to Bob's change address
4) no error is given about dust amounts.
5) the transaction is posted on Bob's wallet and his available balance goes to zero.
[...]

Issue #3688, opened 2019-11-25
Create offer where BSQ change output is below threshold shows misleading error message


devinbileck:
### Description

When attempting to create an offer and using BSQ to pay for fees, if the change output is below the min dust value, the error message indicates to report a bug but it is not a bug.

#### Version

1.2.3

### Steps to reproduce

1. Fund your BSQ wallet with exactly 5.46 BSQ.
2. Attempt to create an offer, paying for fees with BSQ.

### Expected behaviour

A simple error message indicating the change output is below the min threshold.

### Actual behaviour

An error message leading the user to believe it is a bug in the software.

### Screenshots

![image](https://user-images.githubusercontent.com/603793/69575970-39500f00-0f80-11ea-84a5-af78d23ca21d.png)

#### Device or machine

Windows 10

PR #4094, opened 2020-03-24
Prevent dust outputs from being created during the trade process

ghost:
This change fixes an issue whereby dust change outputs are inadvertently created during the trading process, unbeknownst to the user.  The dust outputs cause the Bitcoin node to reject the
transaction and the trade then becomes stuck.

The solution taken here is to detect a dust TXO during the trade process and remove it from the transaction before broadcasting.  This applies in both making an offer and/or taking an existing offer.  When a dust output is prevented, it will be noted in the log as follows:

image

Related to #4039

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
@@ -177,6 +177,9 @@ public class TradeWalletService {
 
             checkNotNull(wallet, "Wallet must not be null");
             wallet.completeTx(sendRequest);
+            if (removeDust(tradingFeeTx)) {
+                wallet.signTransaction(sendRequest);
+            }
             WalletService.printTx("tradingFeeTx", tradingFeeTx);
 
             if (doBroadcast && callback != null) {
@@ -1215,5 +1218,26 @@ public class TradeWalletService {
         checkArgument(!tx.getInputs().isEmpty(), "The tx must have inputs. tx={}", tx);
         tx.getInputs().forEach(input -> input.setSequenceNumber(TransactionInput.NO_SEQUENCE - 1));
         tx.setLockTime(lockTime);
+    }
+
+    private boolean removeDust(Transaction tx) {
+        List<TransactionOutput> txos = tx.getOutputs();
+        List<TransactionOutput> toKeep = new ArrayList<TransactionOutput>();
+        for (TransactionOutput zz: txos) {
+            if (zz.getValue().value >= 546) {
+                toKeep.add(zz);
+            }
+            else {
+                log.info("removing dust TXO = {}", zz.getValue().toFriendlyString());
+            }
+        }
+        if (toKeep.size() != txos.size()) {
+            tx.clearOutputs();
+            for (TransactionOutput zz : toKeep) {
+                tx.addOutput(zz);
+            }
+            return true;    // dust was removed
+        }
+        return false;       // no action necessary
     }
 }
--- a/core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
@@ -1220,21 +1220,29 @@ public class TradeWalletService {
         tx.setLockTime(lockTime);
     }
 
-    private boolean removeDust(Transaction tx) {
-        List<TransactionOutput> txos = tx.getOutputs();
-        List<TransactionOutput> toKeep = new ArrayList<TransactionOutput>();
-        for (TransactionOutput zz: txos) {
-            if (zz.getValue().value >= 546) {
-                toKeep.add(zz);
+    // BISQ issue #4039: prevent dust outputs from being created.
+    // check all the outputs in a proposed transaction, if any are below the dust threshold
+    // remove them, noting the details in the log. returns 'true' to indicate if any dust was
+    // removed.
+    private boolean removeDust(Transaction transaction) {
+        List<TransactionOutput> originalTransactionOutputs = transaction.getOutputs();
+        List<TransactionOutput> keepTransactionOutputs = new ArrayList<>();
+        for (TransactionOutput transactionOutput: originalTransactionOutputs) {
+            if (transactionOutput.getValue().value >= preferences.getIgnoreDustThreshold()) {
+                keepTransactionOutputs.add(transactionOutput);
             }
             else {
-                log.info("removing dust TXO = {}", zz.getValue().toFriendlyString());
+                log.info("your transaction would have contained a dust output of {}", transactionOutput.toString());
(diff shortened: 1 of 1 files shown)


ghost:
~~I neglected to mention that this initial commit only handles the path taken when BTC is used for the trading fee; when BSQ is used the transaction is built by a different routine so some work will be needed to check for dust in the same way there.~~

~~Also, a bit sloppy that I check the dust limit using a hard coded 546 sats value; `preferences.getIgnoreDustThreshold()` could be used which equates to 600 sats currently.~~

[edit] both these issues have now been addressed.

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
```diff
@@ -1216,4 +1219,25 @@ private void applyLockTime(long lockTime, Transaction tx) {
         tx.getInputs().forEach(input -> input.setSequenceNumber(TransactionInput.NO_SEQUENCE - 1));
         tx.setLockTime(lockTime);
     }
+
+    private boolean removeDust(Transaction tx) {
+        List<TransactionOutput> txos = tx.getOutputs();
+        List<TransactionOutput> toKeep = new ArrayList<TransactionOutput>();
```
sqrrm:
```suggestion
        List<TransactionOutput> toKeep = new ArrayList<>();
```

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
```diff
@@ -1216,4 +1219,25 @@ private void applyLockTime(long lockTime, Transaction tx) {
         tx.getInputs().forEach(input -> input.setSequenceNumber(TransactionInput.NO_SEQUENCE - 1));
         tx.setLockTime(lockTime);
     }
+
+    private boolean removeDust(Transaction tx) {
+        List<TransactionOutput> txos = tx.getOutputs();
```
sqrrm: More descriptive variable names are preferred...

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
```diff
@@ -1216,4 +1219,25 @@ private void applyLockTime(long lockTime, Transaction tx) {
...
+        List<TransactionOutput> toKeep = new ArrayList<TransactionOutput>();
+        for (TransactionOutput zz: txos) {
```
sqrrm: ...especially names like `zz` are very hard to follow except in the most simple cases. Calling it `output` or `txOutput` to make it more readable is preferred.

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
```diff
@@ -1216,4 +1219,25 @@ private void applyLockTime(long lockTime, Transaction tx) {
...
+            if (zz.getValue().value >= 546) {
+                toKeep.add(zz);
+            }
+            else {
+                log.info("removing dust TXO = {}", zz.getValue().toFriendlyString());
```
sqrrm: A more verbose log will help log readers.

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
```diff
@@ -1216,4 +1219,25 @@ private void applyLockTime(long lockTime, Transaction tx) {
...
+            }
+        }
+        if (toKeep.size() != txos.size()) {
+            tx.clearOutputs();
+            for (TransactionOutput zz : toKeep) {
```
sqrrm: Please use a more descriptive variable name

sqrrm (changes requested):
Looks functionally correct and easy to follow. Would be helpful with a comment on how this works and what the effects are, perhaps around the `removeDust` method

Please review inline comments

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
ghost: Thankyou for the review!  I have made the changes you suggested.  Here is an example of the more verbose logging as a result of this change:

image

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
```diff
@@ -257,6 +260,7 @@ public Transaction completeBsqTradingFeeTx(Transaction preparedBsqTx,
         checkNotNull(wallet, "Wallet must not be null");
         wallet.completeTx(sendRequest);
         Transaction resultTx = sendRequest.tx;
+        removeDust(resultTx);
```
sqrrm, line 263: I think this will cause BSQ change that lower than dust to be used as extra fee. That could be a non trivial amount (5.46 BSQ > USD3). While that's not a huge amount it's still too big to be an acceptable loss for users. If BSQ prices rise it will become even less so.

I think we'll have to fail transactions that try to pay the fee using BSQ if the change is less than dust. Best would be to show something at the time the fee is selected that it's likely to result in a dust change output and thus not be possible to continue. I think that should be doable, the fees are known at the time of setting up the trade so we can calculate the BSQ dust. One thing we need to do is make sure that if there is enough BSQ but split over 2 or more outputs, it's aggregated to avoid dust outputs. Example, counting satoshis:

Fee: 1000
BSQ wallet outputs:
1. 1200
2. 900

I think the tx might be constructed using only outputs 1. leaving a change output of 200, but it would be possible to construct the tx to take both 1. and 2. to pay the 1000 in fee and get the change 1100.

sqrrm (changes requested):

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
ghost: @sqrrm From what I've seen there is existing code which checks the BSQ for dust before building the transaction (see https://github.com/bisq-network/bisq/issues/3688).  The `removeDust` insertion here looks at the transaction after it is built, just before broadcasting, to remove any dust change BTC.  I will look more into how it works though.

sqrrm: It's possible the check during tx build is not there but rather later, causing it to look like a bug. Still, that's at least better than burning BSQ unwittingly. If you could check that it's handled properly and not as a bug report that would be best of course.

ghost: At the moment there is only one way to get messages back to the GUI - via the errorMessage property on the Offer which is used by numerous exceptional events.  I looked into reporting warnings back via adding a warningMessage property and will propose that as a new PR to address #3688.  (Its a non-trivial change and I don't want to complicate this PR by adding it in here).

sqrrm (approved):
ACK

PR #4096, opened 2020-03-26
Move command & core proto defs to new subproject

ghubstan:
Protobuf definition files were moved from common and core to a new
protodefinition subproject.

The two main reasons for doing this are to speed up builds by not
having to regenerate common and core protobuf classes
every time a change is made in those subprojects, and to remove
the grpc cli's direct dependency on core and the transitive dependency
on common.

In order to accomplish this, cli's BisqCliMain was stripped of
its dependencies on common and core.  Cli can only get the version
and balance now.

gRPC stub boilerplate was moved from BisqCliMain to a CliCommand
class to avoid some of the bloat that is going to happen as the
read-response loop supports more rpc commands.


cbeams:
This is looking good, @ghubstan. I'm going to push one or more commits to this branch as my review.

cbeams (approved):
ACK, please see my two review commits.

I didn't mention this in the commit comments, but I also removed the FIXMEs from grpc.proto about IDEA complaining about not having references to imported proto definitions from pb.proto. I'm not getting any errors any more, and I can navigate from grpc.proto to pb.proto by clicking, e.g. `CMD-B` on the imported definition. Find Usages on, e.g. the `TradeStatistics2` proto def shows its usage in grpc.proto, etc, so I think everything is working as expected. This is probably because of some bug fix in a recent version of the IDEA protobuf plugin.

ghubstan:
I knew something was wrong with my codeStyleSettings, and there still is after setting editor code style scheme to "default", formatting imports with that, then resetting it back to "project".   
My import organizer still doesn't match yours.  Not sure what's going on with it.

I let ^ALT-L organize imports, but doing it through the reformat dialog gives the same result.
And when I do it, idea edits .idea/codeStyles/codeStyleConfig.xml

ghubstan:
Thanks for reviewing this so quickly.

cbeams:
You might want to `rm -rf .idea` followed by `git checkout .idea` to make sure all settings are in sync / correct. Also, in my experience it usually takes a restart of IDEA to pick up changes to the import table.

ghubstan:
`rm -rf .idea` followed by `git checkout .idea` solved it

PR #4097, opened 2020-03-27
Remove read stdin loop

ghubstan:
Replaced the Scanner input read loop with upgraded joptsimple
dependency. Cli now takes a single, non-option program argument, runs
it and exits.  Also removed the "stop client" command because there is
no input loop, but shutdown() is called for orderly channel shudown
before the jvm terminates.  Also changed cmd syntax from camel case
to lowercase, mimicking bitcoin-cli.

Configured logback to supress all debug & info level netty output, and
bypassed logback to print results to System.out (replaces #4091).

### DIFF
--- a/cli/src/main/java/bisq/cli/app/BisqCliMain.java
+++ b/cli/src/main/java/bisq/cli/app/BisqCliMain.java
@@ -20,18 +20,22 @@ package bisq.cli.app;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 
-import java.util.ArrayList;
-import java.util.Arrays;
+import joptsimple.OptionParser;
+import joptsimple.OptionSet;
+import joptsimple.OptionSpec;
+
 import java.util.List;
-import java.util.Scanner;
 import java.util.concurrent.TimeUnit;
 
 import lombok.extern.slf4j.Slf4j;
 
+import static bisq.cli.app.CommandParser.GETBALANCE;
+import static bisq.cli.app.CommandParser.GETVERSION;
+import static bisq.cli.app.CommandParser.STOPSERVER;
 import static java.lang.String.format;
 import static java.lang.System.currentTimeMillis;
 import static java.lang.System.exit;
-import static java.lang.System.in;
+import static java.lang.System.out;
 
 /**
  * gRPC client.
@@ -41,60 +45,28 @@ public class BisqCliMain {
 
     private final ManagedChannel channel;
     private final CliCommand cmd;
+    private final OptionParser cmdParser;
 
     public static void main(String[] args) {
-        new BisqCliMain("localhost", 8888);
+        new BisqCliMain("localhost", 8888, args);
     }
 
-    private BisqCliMain(String host, int port) {
+    private BisqCliMain(String host, int port, String[] params) {
         // Channels are secure by default (via SSL/TLS);  for the example disable TLS to avoid needing certificates.
         this(ManagedChannelBuilder.forAddress(host, port).usePlaintext().build());
 
-        // Simple input scanner
-        // TODO use some more sophisticated input processing with validation....
-        try (Scanner scanner = new Scanner(in)) {
-            while (true) {
-                long startTs = currentTimeMillis();
+        long startTs = currentTimeMillis();
 
-                String[] tokens = scanner.nextLine().split(" ");
-                if (tokens.length == 0) {
-                    return;
-                }
-                String command = tokens[0];
-                if (tokens.length > 1) {
-                    List<String> params = new ArrayList<>(Arrays.asList(tokens));
-                    params.remove(0);
-                }
-                String result;
+        String command = parseCommand(params);
+        String result = runCommand(command);
 
-                switch (command) {
-                    case "getBalance":
-                        long satoshis = cmd.getBalance();
-                        // TODO mimic bitcoin-cli?  Depends on an error code: Loading block index... Verifying blocks...
-                        result = satoshis == -1 ? "Server initializing..." : cmd.prettyBalance.apply(satoshis);
-                        break;
-                    case "getVersion":
-                        result = cmd.getVersion();
-                        break;
-                    case "stop":
-                        result = "Shut down client";
-                        try {
-                            shutdown();
-                        } catch (InterruptedException e) {
-                            log.error(e.toString(), e);
-                        }
-                        break;
-                    case "stopServer":
-                        cmd.stopServer();
-                        result = "Server stopped";
-                        break;
-                    default:
-                        result = format("Unknown command '%s'", command);
-                }
+        // First response is rather slow (300 ms) but following responses are fast (3-5 ms).
+        // log.info("{}\t{}", result, cmd.responseTime.apply(startTs));
+        out.println(result + "\t" + cmd.responseTime.apply(startTs));
 
-                // First response is rather slow (300 ms) but following responses are fast (3-5 ms).
-                log.info("{}\t{}", result, cmd.responseTime.apply(startTs));
-            }
+        try {
+            shutdown(); // Orderly channel shutdown
+        } catch (InterruptedException ignored) {
         }
     }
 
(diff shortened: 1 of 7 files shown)


cli/src/main/java/bisq/cli/app/BisqCliMain.java
```diff
@@ -41,60 +45,28 @@
 
     private final ManagedChannel channel;
     private final CliCommand cmd;
+    private final OptionParser cmdParser;
```
cbeams:
```suggestion
    private final OptionParser parser;
```

cli/src/main/java/bisq/cli/app/BisqCliMain.java
```diff
@@ -41,60 +45,28 @@
...
 
     public static void main(String[] args) {
-        new BisqCliMain("localhost", 8888);
+        new BisqCliMain("localhost", 8888, args);
```
cbeams:
```suggestion
        new BisqCliMain("localhost", 9998, args);
```

8888 is a poor number as it's frequently used by other services. 9998 fits with bisq's p2p port of 9999 and follows suit with bitcoind's 8333 p2p port and 8332 rpc port. I realize you did not introduce this change and that it's somewhat out of scope for this PR, but please make it anyway.

@freimair, please note that the choice of 9998 here overlaps with the same port used in `monitor/src/main/java/bisq/monitor/metric/TorHiddenServiceStartupTime.java`. Please consider changing that to something else if you see it being a conflict.

cli/src/main/java/bisq/cli/app/BisqCliMain.java
```diff
@@ -41,60 +45,28 @@
...
     }
 
-    private BisqCliMain(String host, int port) {
+    private BisqCliMain(String host, int port, String[] params) {
```
cbeams:
```suggestion
    private BisqCliMain(String host, int port, String[] args) {
```

For consistency and conventionality.

cli/src/main/java/bisq/cli/app/BisqCliMain.java
```diff
@@ -41,60 +45,28 @@
...
-                        result = "Server stopped";
-                        break;
-                    default:
-                        result = format("Unknown command '%s'", command);
-                }
-
-                // First response is rather slow (300 ms) but following responses are fast (3-5 ms).
-                log.info("{}\t{}", result, cmd.responseTime.apply(startTs));
-            }
+        long startTs = currentTimeMillis();
```
cbeams: Please remove this.

cli/src/main/java/bisq/cli/app/BisqCliMain.java
```diff
@@ -41,60 +45,28 @@
...
+
+        String command = parseCommand(params);
+        String result = runCommand(command);
+
+        // First response is rather slow (300 ms) but following responses are fast (3-5 ms).
+        // log.info("{}\t{}", result, cmd.responseTime.apply(startTs));
+        out.println(result + "\t" + cmd.responseTime.apply(startTs));
```
cbeams: And this as well. There is no requirement at present to measure the response time, and output from a unix utility should not include anything other than the data requested, i.e. such that it can be piped to the next utility. This sort of output breaks that unix idiom. Again, I realize you did not introduce this. I'm just asking you to remove it and start thinking about things this way, making similar decisions as you go.

cli/src/main/java/bisq/cli/app/CommandParser.java
```diff
@@ -0,0 +1,27 @@
...
+    OptionParser configure() {
+        OptionParser parser = new OptionParser();
+        parser.allowsUnrecognizedOptions();
+        parser.nonOptions(GETBALANCE).ofType(String.class).describedAs("get btc balance");
+        parser.nonOptions(GETVERSION).ofType(String.class).describedAs("get bisq version");
+        return parser;
+    }
+
+    static void printUsage() {
+        err.println("Usage:  bisq-cli getbalance | getversion");
```
cbeams: Please study the way help output is printed in `BisqExecutable#execute` and note the use of `BisqHelpFormatter` which we may want to reuse (we would need to move it in order to do so, don't worry about it now, just use the default help formatter implementation in the meantime).

In any case, when help is explicitly requested, it shouldn't be printed to stderr, but stdout.

Please also explicitly `System.exit` 0 or 1 accordingly. You'll see how this is done in `BisqExecutable` as well.

cbeams (changes requested):
See my comments and suggested changes. There are additional changes I'd make as well, but time is limited and this is good enough to go forward. Thanks.

cli/src/main/java/bisq/cli/app/BisqCliMain.java
ghubstan: Changed in cli, but still need to decide how to resolve conflict with port number(s) in `monitor/src/main/java/bisq/monitor/metric/TorHiddenServiceStartupTime.java`

cli/src/main/java/bisq/cli/app/CommandParser.java
ghubstan: Prints usage message to stdout, and uses EXIT_SUCCESS and EXIT_FAILURE arguments in System.exit(). 

What's not done:  help formatting.  I need to discuss Bisq + `joptsimple.HelpFormatter` issues with @cbeams.  Right now, the only options passed to Cli are of type `NonOptionArgumentSpec`, which did not work as expected with `joptsimple.HelpFormatter`.  It looks like Bisq Config only uses `ArgumentAcceptingOptionSpec` options, which do work with `joptsimple.HelpFormatter`.  It did not seem correct to change non-option cli arguments to argument accepting options just to make them work with the `joptsimple.HelpFormatter` interface, which is why I checked in the simple 'print usage' print statement.

cli/src/main/java/bisq/cli/app/BisqCliMain.java
ghubstan: TODO  change BisqGrpcServer's listening port.

ghubstan: BisqGrpcServer now listens on port 9998.

ghubstan: A reminder to @freimair, re `TorHiddenServiceStartupTime`
```
int localPort = Integer.parseInt(configuration.getProperty(LOCAL_PORT, "9998"));
int servicePort = Integer.parseInt(configuration.getProperty(SERVICE_PORT, "9999"));
```
now conflicts with the `BisqGrpcServer` listening port `9998` in this `remove-read-stdin-loop` branch (not merged).   I am sure I should not be changing server + listening ports in `TorHiddenServiceStartupTime`.

cbeams:
Please have another look at how help / usage output is handled in `BisqExecutable` and co. I’m referring to hooking into JOptSimple’s built-in support here.

cli/src/main/java/bisq/cli/app/CommandParser.java
cbeams: Ok, that's fine then. We can come back to it later, thanks.

cbeams:
> Please have another look at how help / usage output is handled in `BisqExecutable` and co. I’m referring to hooking into JOptSimple’s built-in support here.

Please disregard this message. I wrote it via mobile, where I hadn't seen @ghubstan's https://github.com/bisq-network/bisq/pull/4097#discussion_r400246596.

cbeams:
I would generally recommend against force pushing on PR branches, especially where feedback is being folded in. It destroys history / information, and messes with merging / rebasing if someone (like myself) is coming along with review commits as well. I don't know why you did these force pushes, perhaps there was a great reason, just mentioning.

cbeams (approved):
ACK, the changes look good. Thanks @ghubstan.

ghubstan:
> I would generally recommend against force pushing on PR branches, especially where feedback is being folded in. It destroys history / information, and messes with merging / rebasing if someone (like myself) is coming along with review commits as well. _I don't know why you did these force pushes_, perhaps there was a great reason, just mentioning.

Sorry.  No good enough reason for squashing too many intermediate commits, just a lot of interruptions today, preventing me from correctly implementing all of your requested changes in one push before the end of your, @cbeams' day.  I thought one clean (2nd) commit message was more important.  I'll be more careful and patient next time.

cbeams:
All good, thanks!

cli/src/main/java/bisq/cli/app/BisqCliMain.java
freimair: I do not think so. the monitor is quite separate from everything else, plus, this is a default fallback value - it can and should be configured via the monitors properties file.

PR #4102, opened 2020-03-30
Tidy PersistableEnvelope inheritance

stejbac:
This refactoring is a continuation from the concurrency fixes in #4025, to remove all the remaining superfluous `PersistableEnvelope` interfaces on classes implementing `Proto`.

- Remove an unused `PersistableEnvelope` interface from the following five `PersistableNetworkPayload` implementations:

  >  AccountAgeWitness, BlindVotePayload, ProposalPayload, SignedWitness, TradeStatistics2

  These already have corresponding `*Store` envelope classes which correctly implement the interface.

- Additionally, delete the unused class `PersistableHashMap`, make `(UserThreadMapped)PersistableList` abstract and remove some unused functionality from the base class.

- The last commit in this PR touches the DAO packages, as the consensus-critical classes `MeritList` and `VoteWithProposalTxIdList` needlessly implement the `PersistableEnvelope` interface prior to this change, via the base class `PersistableList`. I inlined the `list` field, instead of inheriting it from the base class, and made them standalone classes instead. This shouldn't observably alter behaviour and would hopefully also reduce the chance of future inadvertent changes in behaviour (as `PersistableList` lies outside the DAO packages).


sqrrm (commented):
I like the first two commits, looks good to me but I haven't tested.

The third commit, touching the DAO classes makes me worried. I think it's fine and that it won't cause any trouble for the consensus but I'm not confident. Considering the possible trouble I would hesitate to merge without more thorough review and testing. We also still need the review of @ManfredKarrer

ripcurlx:
> We also still need the review of @ManfredKarrer

Just pinged him, so we get this into master for the next release.

ManfredKarrer (approved):
utACK

I recommend to test a few cycles in regtest where the changed classes are used to see if it does not break consensus. Also maybe better to merge it after the next release so we get more time for dev testing to detect potential issues. I dont see any but the DAO consensus is tricky.

ripcurlx:
Just tested a couple of cycles on Regtest.
The only thing I recognized was, that although the consensus hashes were fine, the voting results were only shown for one party. After a restart the results were shown correctly to the other party as well. 

**Before restart**
Bildschirmfoto 2020-04-27 um 15 50 26
**After restart**
Bildschirmfoto 2020-04-27 um 15 51 25

@stejbac Do you think this can be caused by this changes? Could be an existing bug, that I just never recognized during testing.

ripcurlx:
> Just tested a couple of cycles on Regtest.
> The only thing I recognized was, that although the consensus hashes were fine, the voting results were only shown for one party. After a restart the results were shown correctly to the other party as well.
> 
> **Before restart**
> Bildschirmfoto 2020-04-27 um 15 50 26
> **After restart**
> Bildschirmfoto 2020-04-27 um 15 51 25
> 
> @stejbac Do you think this can be caused by this changes? Could be an existing bug, that I just never recognized during testing.

As this just happened on Mainnet as well, I have no concerns about this anymore 😬

ripcurlx (approved):
ACK

PR #4107, opened 2020-03-31
Set expiry date to years in the future

freimair:
Belongs to https://github.com/bisq-network/projects/issues/13

The default expiry date of the certificate has been quite short with only 30 days. Changed it to years in the future. Please note that these certificates are used for client authentication and are pinned on the monitoring server. Therefore, in case a certificate gets compromised, we can just remove it from the server and we are save again.


wiz (approved):
ACK

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/4107#pullrequestreview-384849716

PR #4027, opened 2020-03-04
Report HS version to pricenode

freimair:
Part of https://github.com/bisq-network/projects/issues/23

In order to evaluate progress on https://github.com/bisq-network/projects/issues/23,
the Bisq app reports its hiddenservice version.

This change is going to be undone as soon as we do not need the
info anymore.

PR #4112, opened 2020-04-02
Add parsing to pricenode cmdline inputs

freimair:
Stumbled over during https://github.com/bisq-network/bisq/pull/4027

This PR adds basic input validation and correction for the custom pricenode command line parameter:
- `--help` says
```
  --providers=<host:port[,...]>
        List custom pricenodes
```
- yet, setting the value to `44mgyoe2b6oqiytt.onion:80` fails.
- because the correct formatting requires "http" in front and a "/" at the end


ripcurlx (approved):
utACK

PR #4114, opened 2020-04-02
Track p2p data store files using Git LFS

cbeams:
The large binary objects in p2p/src/main/resources/ are updated on every
Bisq release with the latest network data to avoid the need for new Bisq
clients to download all of this information from the network, which
would easily overload seed nodes and generally bog down the client.

This approach works well enough for its purposes, but comes with the
siginficant downside of storing all of this binary data in Git history
forever. The current version of these binary objects total about 65M,
and they grow with every release. In aggregate, this has caused the
total size of the repository to grow to 360M, making it cumbersome to
clone over a low-bandwith connection, and slowing down various local Git
operations.

To avoid further exacerbating this problem, this commit sets these files
up to be tracked via Git LFS. There's nothing we can do about the 360M
of files that already exist in history, but we can ensure it doesn't
grow in this unchecked way going forward. For an understanding of how
Git LFS works, see the reference material at [1], and see also the
sample project and README at [2].

We are using GitHub's built-in LFS service here, and it's important to
understand that there are storage and bandwidth limits in place. We have
1G total storage and 1G per month of bandwidth on the free tier. If we
exceed this, we will need to purchase a "data pack" from GitHub, which
will get us to 50G storage and bandwith. These are reasonably priced and
not the end of the world if it becomes necessary. In an attempt to avoid
this, however, the Travis CI build configuration has been updated to
cache Git LFS files, such that they are not re-downloaded on every CI
build, as this would very quickly cause us to exceed the free tier
bandwith limit (see [3] and [4]). With that out of the way, the variable
determining whether we exceed the monthly limit is how many clones we
have every month. Only the latest version of LFS-tracked files are
downloaded on a given clone, and at the current 65M that means we have
about 15 clones per month. This is not very many, and will almost
certainly mean that we exceed the limit. Our options at that point are
to buy a data pack or to run our own LFS server. We would almost
certainly do the former to start.

Tracking these files via LFS means that developers will need to have Git
LFS installed in order to properly synchronize the files. If a developer
does not have LFS installed, cloning will complete successfully and the
build would complete successfully, but the app would fail when trying to
actually load the p2p data store files. For this reason, the build has
been updated to proactively check that the p2p data store files have
been properly synchronized via LFS, and if not, the build fails with a
helpful error message. The docs/build.md instructions have also been
updated accordingly.

It is important that we make this change now, not only to avoid growing
the repository in the way described above as we have been doing now for
many releases, but also because we are now considering adding yet more
binary objects to the repository, as proposed at
https://github.com/bisq-network/projects/issues/25.

[1]: https://git-lfs.github.com
[2]: https://github.com/cbeams/lfs-test
[3]: https://docs-staging.travis-ci.com/user/customizing-the-build/#git-lfs
[4]: https://github.com/travis-ci/travis-ci/issues/8787#issuecomment-394202791


freimair:
so we can publish new "database updates" between releases?

cbeams:
> so we can publish new "database updates" between releases?

This approach would allow us to add LFS-tracked binaries at any time, just like any other commit. If what you mean is can we cut releases of these new "database updates" outside of the normal release cadence that's more of a process question. I wouldn't want to mix up the semantics of what a tag / release means in this repository, but we could discuss further what's actually needed and how to get there.

cbeams:
When I put together this PR earlier today, I was grossly underestimating the number of clones that this repository gets every month. This is the Clones graph in `Insights > Traffic`:

image

We averaged roughly 28 clones per day from 14 unique cloners over the last 14 days. I'm assuming that most of those cloners are not just different Travis CI machines. If in fact most of them are Travis CI, then we might be OK from a bandwidth perspective because of how we're caching LFS data in Travis as mentioned above. But let's just assume for a moment that all of those clones are normal, non-Travis clones. That would mean 28 (clones/day) x 30 (days/month) x 65 (MB of LFS data per clone) / 1024 (MB/GB) = `53 GB per month` of LFS bandwidth. So that would already put us over the 1 data pack (50 GB/month) limit. 1 data pack is $60 / year, and we'd need at least 2, meaning $120 per year. It's not prohibitively expensive to consider the DAO paying for this, but it does make me pause and think before pushing to execute this.

Let's step back for a moment. I think we have the following options:

1. **Do nothing.** Stay with the status quo, let the repository size continue to grow with every release, keep all data stores n Git history forever. We've been tracking these binaries since July of 2018. In roughly 20 months, they've grown the repository from around 20M to 360M. This rate of growth will only continue from here out, assuming that Bisq continues to succeed and therefore have more of this data (e.g. historical trade data) accumulating more rapidly.

2. **Use Git LFS as proposed here.** Bite the bullet, pay the $120 per year and stop the growth of the repository now.

3. **Run our own Git LFS server.** This wouldn't necessarily be a cost savings, as we'd have to maintain the instance and pay for the bandwidth anyway. It's also not obvious what the go-to implementation would be. https://github.com/git-lfs/git-lfs/wiki/Implementations.

4. **Re-think our approach to distributing data stores entirely.** This is a larger conversation, but now may be the right time to have it. I was absent when the decision was made to start tracking and distributing these data stores, but I believe it was a pragmatic decision made to alleviate growing load on our seed nodes. So the question arises, why put all of this on the seed nodes in the first place? Why not have the entire network of Bisq nodes share this data with one another, more or less the way that Bitcoin full nodes service new or recently offline nodes in catching up with the blockchain? If we combined this approach with relaxing the need for new Bisq clients to have all historical data right out of the gate, we could even make it efficient. Certain kinds of data, e.g. account signing data would need to be comprehensive, i.e. new clients would want and need to get all that data up front, but they don't necessarily need to have the full 56,000-trade history (at time of writing) just to get going. They could request the last N trades up front, and then slowly catch up over time, or only catch up with the full history on request from the user. So the basic idea here is why don't we take this load off our seed nodes and distribute it to the whole network? This may well have been considered and dismissed for good reason, my apologies if this is rehashing old territory, but the refresher might be good to have anyway.

Discussion welcome, please. We don't have to pull the trigger on this PR urgently (though I would like to do it before we add or update any more data stores). Better to get this right. Having considered all the above, I would probably still go with option (2) above, i.e. merging this PR and just paying for GitHub's LFS service, as it's the easiest thing to do and will just work. But I'd like to hear others' thoughts.

freimair:
I believe we need some more context to fully grasp the topic:

## Why are these datastores in place?

Given a fresh install of Bisq, Bisq needs to sync up its local copy of our distributed database. It could do so by just asking the network and download the data from there. We do not have the bandwidth, as our P2P network is still very small.

Thus, every release comes with a snapshot of the distributed database. Thus, fresh Bisq installs do not have to download huge amounts (we might just sit at the 100MB mark) of data just to get started.

(Alas, the user just downloaded the data via the installer, so technically, it does not make a difference at the user level. However, given a too-small network and generally, upstream << downstream, it could take a while. So we went for hosting it via the installers.)

## Does changing the storage location alter Bisq functionality?

No, the database snapshots are still going to be included in the release binaries. Only builds from source might have to be adjusted. There is no need to change Bisqs functionality and inner workings.

## Further options I can think:

5. only add the updated data as proposed as a side effect in https://github.com/bisq-network/projects/issues/25. Of course this is a short term countermeasure and only slows the problem down. But that could be substantial as well: going from adding a database with `size(t) = size(t-1) + newdata` to adding a database with `size(t) = newdata` reduces the order of size growth significantly.
Again, this is NOT a solution, but it might buy us time to think about the issue properly.

6. use git releases to host the database snapshots
7. use git packages to host the database snapshots

## Historical data vs. live data

The database (and its snapshots) evolves like (just made up the names though):

`hd(release) = hd(release-1) + livedata`

with `release` being a point in time, where a new bisq version has been released (eg. v1.2.9). This release ships historical data `hd(bisq version)`. `livedata` is data coming in in between releases when Bisq business commences. Trades are being conducted, new accounts are created (Note, offers are not included as they will be gone after they have been taken). Before `livedata` gets compiled into `historical data`, it only lives in the P2P network

For example:
- historical data `hd(v1.2.9)` is shipped with release v1.2.9
- `hd(v1.2.9)` includes `hd(v1.2.8) + livedata` that accumulated in the time between the releases

(Please note that upgrading a bisq version does not use the shipped database snapshot of the new version as for now)

That being said, I understand your (4.) is aimed at `livedata`:

> 4. **Re-think our approach to distributing data stores entirely.** So the question arises, why put all of this on the seed nodes in the first place? Why not have the entire network of Bisq nodes share this data with one another

- Well, the entire network is sharing data with one another. That is how it works and how it is designed.
- seednodes do NOT distribute the database snapshots, these snapshots are shipped with the installer binaries.

however:
- as time goes on, a database snapshot shipped with the installer start missing objects immediately (because life goes on and trades happen and ...)
- the Bisq app, however, needs all the data, especially the latest, to work properly
- thus, bisq does the following steps before it shows the actual GUI to the user
  1. ask 2 seednodes for preliminary data (> 4MB request)
  2. ask 2 seednodes for updated data (4MB again) after hidden service is online in case there has been something happening between preliminary request and the hidden service coming online (note that having v3 HS now and their very low publishing time, we might think of loosing the preliminary request (see what I did there? it finally comes together :1st_place_medal: ) )
  3. ask 2 seednodes for peers
- by doing it like so, the Bisq network eliminates issues with source file availability. Just think back to the good ol' days with eMule when you had downloaded 99% of the ubuntu4.iso and there just hasn't been a source around for the very last 4k block of data you have been missing. As our P2P network is still very small, we might encounter that as well.

- other approaches have been suggested, one of the most promising imho is to scramble up the current steps and split the data load.
  1. ask for peers
  2. ask these peers + seed nodes for data, but do not ask all data from all of them
which of course is not trivial, because we do not know which data we are missing...


## Dev Call

I have a feeling we should discuss this on a dev call someday. Needs some more info though.

freimair:
ahm in case this hasn't been made obvious, I 100% :+1: git lfs.

sqrrm:
I support the git lfs solution for now.

One point to make on the DaoStateStore is that it's not the size of the store that's the issue. If a new user didn't have that data store the first thing they would have to do is wait for the DAO to sync from genesis before doing anything else, that's a long wait that would ruin the user experience.

ripcurlx:
I'm also fine merging this PR, as long it covers all required changes for the existing release process.

cbeams:
I just need to buy the data packs and update the PR to resolve the conflict. Will do that shortly.

cbeams:
I've just updated this PR to resolve merge by force pushing newly LFS-tracked binaries. I've also modified the commit message in several places for clarity / accuracy. Everything else in the PR remains the same.

This is ready to go from my perspective. It should "just work" for everyone, assuming they have git-lfs installed, as detailed in the changes to `docs/build.md` (you just need to e.g. `brew install git-lfs`).

Also note that I'm going to wait to buy the first data pack just to see what our actual usage looks like. I'll check it every day so we don't run into any problems getting rate limited.

ripcurlx (approved):
utACK

cd2357:
You mentioned the free tier offers 1GB storage and 1GB / month bandwidth.

Have you guys considered zipping the datastore files before storing them on LFS? That would reduce storage + bandwidth used.

Did a quick test, the files uncompressed are about 75MB, compressed 35MB => savings of 50%.

cbeams:
I've just purchased 1 data pack for Git LFS, as we were approaching our yearly limits.

This PR was merged 19 days ago at time of writing:

```
$ git show 2e3e811
commit 2e3e811bf6447bb1995134c3d6d79479a4ad65cb
Merge: 5fd05cf66 51fc2710a
Author: Christoph Atteneder <christoph.atteneder@gmail.com>
Date:   Wed Apr 29 16:15:06 2020 +0200

    Merge pull request #4114 from cbeams/git-lfs

    Track p2p data store files using Git LFS
```

And in that time, we have used 10.2 GB of bandwidth, bringing us close to the 12 GB per year limit on the free LFS plan:

image

That means we're averaging 0.5 GB / day of bandwidth. 10.2/19×365 = 196 GB per year, meaning that **purchasing 1 data pack, with 600 GB per year should be plenty.** I've just done this, and here is what our LFS billing section looks like now:

image

I will request reimbursement for the 60 USD charge once yearly as part of my regular compensation request.

- [x] I will also update the GitHub Admin role documentation to reflect that renewing this data pack is a duty. (done; see [this change](https://bisq.wiki/index.php?title=GitHub_Admin&type=revision&diff=1245&oldid=1244) to the wiki)

cbeams:
@cd2357 wrote in https://github.com/bisq-network/bisq/pull/4114#issuecomment-621330732

> Have you guys considered zipping the datastore files before storing them on LFS?

That's a good idea, and no, I don't recall it ever coming up. We're good to go now for quite a while, as we're only using 0.2 GB of storage and we have 600 GB available. Currently, each new release adds around 70 MB, and this number grows slightly with every release, so we have on the order of thousands of releases to go before we hit the limit. Certainly no rush from that perspective, but it would be nice to reduce the size now to speed up (shallow) cloning. If you're interested in putting together a PR for that, @cd2357, go ahead, but (a) I think it's pretty low priority, and (b) it needs to be comprehensive, e.g. release documentation / scripts need to be updated to ensure the files get zipped correctly instead of added in their raw form. @ripcurlx / @sqrrm (release managers) would need to be informed, and you'd want to get their blessing about doing it up front.

cbeams:
Oops, I just realized I made a mistake. We don't have 600 GB of storage, we only have 50. That's still hundreds of releases we can do at the current rate before we run into limits, but it does bump the priority of compressing the files at least somewhat.

freimair:
ad zipping stuff: please don't rush things. I am in the process of restructuring the data distribution mechanisms and this should already give us a smaller data footprint.

Ideas I am currently thinking about are
- removing the data stores from the installer packages completely and even from the main repository
- instead offer an external URL where these stores can be fetched if needed
- a separate Git repository probably with only binaries
- zipping them there would be a lot easier
- needs some cryptographic means of making sure the stores are sound of course
- and I am not quite sure if that is an entirely good plan at all

Issue #4110, opened 2020-04-01
Seednode logs unhandled exception 10 times per second running master


wiz:
Logs are filled with this message 10 times per second:

```
Apr-01 16:20:54.588 [SeedNodeMain] ERROR b.c.s.CommonSetup: Uncaught Exception from thread SeedNodeMain
Apr-01 16:20:54.588 [SeedNodeMain] ERROR b.c.s.CommonSetup: throwableMessage= com.google.common.util.concurrent.Futures.addCallback(Lcom/google/common/util/concurrent/ListenableFuture;Lcom/google/common/util/concurrent/FutureCallback;)V
Apr-01 16:20:54.588 [SeedNodeMain] ERROR b.c.s.CommonSetup: throwableClass= class java.lang.NoSuchMethodError
Apr-01 16:20:54.589 [SeedNodeMain] ERROR b.c.s.CommonSetup: Stack trace:
java.lang.NoSuchMethodError: com.google.common.util.concurrent.Futures.addCallback(Lcom/google/common/util/concurrent/ListenableFuture;Lcom/google/common/util/concurrent/FutureCallback;)V
        at bisq.network.p2p.network.NetworkNode.sendMessage(NetworkNode.java:288)
        at bisq.network.p2p.peers.BroadcastHandler.sendToPeer(BroadcastHandler.java:172)
        at bisq.network.p2p.peers.BroadcastHandler.lambda$broadcast$2(BroadcastHandler.java:157)
        at bisq.common.FrameRateTimer.run(FrameRateTimer.java:50)
        at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
        at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:658)
        at bisq.common.MasterTimer$1.lambda$run$0(MasterTimer.java:38)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1135)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at java.base/java.lang.Thread.run(Thread.java:844)

```

freimair:
I tried to reproduce your issue with the latest master and have not been able to do so.

did you do a `./gradlew clean :seednode:installDist`? because your installer does a `build` only, which might not be enough...

if yes, may I ask for the logs from the startup to the error case you already postet?

[...]

PR #4116, opened 2020-04-02
Add guava dependency to seednode

ghubstan:
Merging PR #4096, which moved protobuf defs out of core and common,
left :seednode without its required dependency on guava, causing
NoSuchMethodErrors.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -488,6 +488,7 @@ configure(project(':seednode')) {
     dependencies {
         compile project(':core')
         compileOnly "org.projectlombok:lombok:$lombokVersion"
+        implementation "com.google.guava:guava:$guavaVersion"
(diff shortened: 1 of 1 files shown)


wiz:
@ghubstan would this have caused https://github.com/bisq-network/bisq/issues/4110 ?

ripcurlx:
> @ghubstan would this have caused #4110 ?

Yes, I think that is it.

ripcurlx (approved):
ACK

Solved the problem for me

ghubstan:
> @ghubstan would this have caused #4110 ?

@wiz, yes... profuse apologies

wiz:
No need to apologize bro we were just trying to figure out if it was a problem with my seednode installation script. Thanks for fixing :+1:

PR #4122, opened 2020-04-06
Make UserThread::run* methods thread safe

### LINKED ISSUE #4055: Bisq sometimes fails to prompt user for password to unlock wallet
### Description

Bisq will very sometimes (but very rarely) start up without asking the user for their wallet password. This causes many other random bugs to occur, as the wallet is locked and cannot be used to do any trade protocol functions, etc.

#### Version

v1.2.8 on Ubuntu Linux

### Steps to reproduce

1) Encrypt wallet with a password
2) Start Bisq over and over, eventually it will neglect to ask you to unlock the wallet
It only occurs once every 20 ~ 30 restarts or so and I have no idea what causes it.

### Expected behaviour

Bisq should always prompt user for password to unlock wallet

stejbac:
The `UserThread::run*` methods delegate to `UITimer::run(Later|Periodically)` in the case of the desktop application. These use the JavaFX `TimeLine` API (via `bisq.common.reactfx.FXTimer`) to schedule future events. However, this API isn't thread safe and isn't meant to be called outside the FX application thread. This causes occasional task misfires and out-of-order scheduling when `UserThread::runAfter` is called outside the user thread.

Make the `UITimer::run*` methods safe to call from any thread by checking we are in the application thread and delegating to `UserThread::execute` otherwise. This also improves consistency between the contracts of the `run*` and `execute` methods. As the former has many call sites, this is safer than trying to track down all the non-thread-safe uses.

(The `Timer` used in the headless app already appears to be thread-safe.)

This fixes #4055, caused by out-of-order scheduling of the `execute` and `runAfter` tasks in the `WalletConfig.onSetupCompleted` anonymous class method in `bisq.core.btc.setup.WalletsSetup.initialize`.

Also prevent an exception caused by non-thread-safe calls into JavaFX during the shutdown of `OpenOfferManager`, which was uncovered by the above, by adding a missing `UserThread::execute` call.

### DIFF
--- a/core/src/main/java/bisq/core/offer/OpenOfferManager.java
+++ b/core/src/main/java/bisq/core/offer/OpenOfferManager.java
@@ -159,9 +159,8 @@ public class OpenOfferManager implements PeerManager.Listener, DecryptedDirectMe
         openOfferTradableListStorage = storage;
 
         // In case the app did get killed the shutDown from the modules is not called, so we use a shutdown hook
-        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
-            UserThread.execute(OpenOfferManager.this::shutDown);
-        }, "OpenOfferManager.ShutDownHook"));
+        Runtime.getRuntime().addShutdownHook(new Thread(() ->
+                UserThread.execute(OpenOfferManager.this::shutDown), "OpenOfferManager.ShutDownHook"));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


freimair (approved):
> Also prevent an exception caused by non-thread-safe calls into JavaFX during the shutdown of OpenOfferManager, which was uncovered by the above, by adding a missing UserThread::execute call.

I was going to address that and other things with https://github.com/bisq-network/projects/issues/28. But that is a good fix until then.

to the topic:
- reviewed the code
- did a local build
- did a smoke test checking for any irregularities

all went good. Its a :+1: from me.

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/4122#pullrequestreview-391452374

Issue #3768, opened 2019-12-09
Filter signed accounts


sqrrm:
### Description

Add option to filter signed account. Will invalidate signing and the entire signed tree below this AccountAgeWitness.

Filtering probably best done on pubkey. All AccountAgeWitnesses using the same pubkey will be filtered out.

ripcurlx:
This feature is for the case that a scammer account got signed by mistake (e.g. bank transfer from stolen bank accounts is not detected within 30 days and a trade with a signer account took place) and started to sign afterwards other stolen bank accounts to be able to lift the 0.01 BTC limit. As soon as the scammer account is detected (a chargeback happens) we should be able to filter during runtime this account an all accounts that got signed by it to prevent further damage to the network. As always users can ignore this filter by using a `--ignoreDevMsg` startup parameters. Doing so the user uses Bisq at its own risk and will not be able for compensation if anything unexpected happens.

ripcurlx:
Closing as complete.

Issue #3931, opened 2020-02-02
Account signing issue with Interac e-Transfer


ghost:
### Description

I have a completed trade where I purchased 0.01 BTC with Interac e-Transfer and the seller has an account age of 86 days and "Signed by peer and can sign peer accounts (limits lifted). The trade started Jan 26 and completed Jan 27. The Interac e-Transfer account under Account > National Currency Accounts still shows "Not signed yet".

Two others have also reported this on the forums: https://bisq.community/t/my-account-didnt-get-signed/8837

#### Version

v1.2.5

### Steps to reproduce

Only tried once with a real trade.

### Expected behaviour

Account should be signed once the trade completes. 

### Actual behaviour

Account remains unsigned after one week.

#### Device or machine

Ubuntu 18

sqrrm:
There are a few cases like this and I haven't been able to figure why it's failing yet. I'm looking into it.

huey735:
Another case - https://bisq.community/t/account-not-getting-signed/9053/4

Bayernatoor:
I've been seeing alot of reports of this happening so i tested it out myself. 

I completed a CAD/BTC trade yesterday using the e-transfer payment method. The seller's account has been signed for 143 days and could sign peers (limits lifted). My account was not signed prior to the trade and had been created 79 days ago. 

Upon completion of the trade my account was not signed - If i look under Account Signing Status it comes up as not signed. 

This also happened a long time ago to another one of my addresses.

Bayernatoor:
I went looking through my logs for any clues as to what may be causing this. The fields indicating signing are showing  as False. 

![image](https://user-images.githubusercontent.com/55212954/78504880-5a4a1d00-7735-11ea-8900-ea1254e8a615.png)





Perhaps, the peer is unable to sign even though their account shows them as being capable of it.

sqrrm:
This is a known issue with one of the signedwitnesses in the CAD market. I don't have a good solution yet. So far I can only think of invalidating all CAD witnesses that depend on that account, but that might invalidate the whole CAD market which would be quite intrusive.

Bayernatoor:
@sqrrm i see...is there anything i can do on my end to test?

sqrrm:
At this point there is nothing to do, unless you happen to have an original `SignedWitnessStore` file from October last year. I doubt anyone has it though.

Bayernatoor:
[...]

PR #4124, opened 2020-04-06
Add signed witness filter

sqrrm:
- Add a filter to pubkeys used in AccountAgeWitness signing
- Fix inverted arbitrator signing of initial account age witnesses from
disputes
- Add test to verify that signed witness filter works
- Add test to verify that the arbitrator signing was fixed

This PR adds a filter to signed witness pubkeys, which are used as the identifier in the signing tree. If  a pubkey is filtered the witnesses dependent on that key will also be invalidated.

Example:
```
  Arbitrator
  |        |
  w1'      w2
 /  \     /  \
w3' w4'  w5  w6'
|   |    |   |  \
w7' w8'  w9  wA' wB'
```
w1 and w6 are filtered, causing all dependent witnesses to be filtered as well (all the prime `'` witnesses in the graph). In this example only w2, w5 and w9 are still considered signed.

#### Bad data fix
A fix to how already signed witnesses are retrieved (`SignedWitnessService.getVerifiedWitnessDateList`) also affects cases where some data in the signed tree is missing, as seen in #3931.

The extra check
```
        if (!isSignedAccountAgeWitness(accountAgeWitness)) {
            return new ArrayList<>();
        }
```
will cause a verification each time rather than allowing to start a trade without checking and later being surprised that the counterparty isn't signed.

For now this will likely cause some issues for the CAD market for people that have already signed  accounts but that are missing an ancestor signed witness data item. It will avoid any surprises with users not getting signed where they expect it though.

### DIFF
--- a/core/src/main/java/bisq/core/account/sign/SignedWitnessService.java
+++ b/core/src/main/java/bisq/core/account/sign/SignedWitnessService.java
@@ -18,6 +18,7 @@
 package bisq.core.account.sign;
 
 import bisq.core.account.witness.AccountAgeWitness;
+import bisq.core.filter.FilterManager;
 import bisq.core.support.dispute.arbitration.arbitrator.ArbitratorManager;
 import bisq.core.user.User;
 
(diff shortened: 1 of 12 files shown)


ripcurlx (approved):
ACK

Code looks fine and I tested multiple cases on Regtest filtering out malicious signed witnesses.

ripcurlx:
Fixes #3768.

PR #4134, opened 2020-04-07
Add BuyerVerifiesDonationAddress task

chimp1984:
Buyer was not verifying if the donation address is either the default
DAO param address or the recent one changed by DAO voting. We do not
support past DAO param addresses to avoid that past addresses receive
funds.


sqrrm (approved):
ACK

PR #4135, opened 2020-04-07
Add donation address check for open trades

chimp1984:
Depends on https://github.com/bisq-network/bisq/pull/4134

### DIFF
--- a/core/src/main/java/bisq/core/trade/DonationAddressValidation.java
+++ b/core/src/main/java/bisq/core/trade/DonationAddressValidation.java
@@ -1,0 +1,86 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.core.trade;
+
+import bisq.core.btc.wallet.BtcWalletService;
+import bisq.core.dao.DaoFacade;
+import bisq.core.dao.governance.param.Param;
+
+import org.bitcoinj.core.Address;
+import org.bitcoinj.core.NetworkParameters;
+import org.bitcoinj.core.Transaction;
+import org.bitcoinj.core.TransactionOutput;
+
+import lombok.Value;
+import lombok.extern.slf4j.Slf4j;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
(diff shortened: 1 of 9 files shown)


core/src/main/java/bisq/core/trade/TradeManager.java
```diff
@@ -297,6 +297,15 @@ private void initPendingTrades() {
                                 trade.getId());
                         tradesWithoutDepositTx.add(trade);
                     }
+
+            try {
+                DonationAddressValidation.validate(trade.getDelayedPayoutTx(),
```
sqrrm, line 302: This causes problems at startup. I think it's because it's throwing a runtimeexception due to `        checkNotNull(delayedPayoutTx, "delayedPayoutTx must not be null");` in `DonationAddressValidation.validate`

core/src/main/java/bisq/core/trade/TradeManager.java
```diff
@@ -297,6 +297,15 @@ private void initPendingTrades() {
...
+                        daoFacade,
+                        btcWalletService);
```
sqrrm, line 304:
```suggestion
                        if (trade.getDelayedPayoutTx() != null) {
                            DonationAddressValidation.validate(trade.getDelayedPayoutTx(),
                                    daoFacade,
                                    btcWalletService);
                        }
```

core/src/main/java/bisq/core/trade/DonationAddressValidation.java
```diff
@@ -0,0 +1,86 @@
...
+        // In case the seller has deactivated the DAO the default address will be used.
+        String defaultDonationAddressString = Param.RECIPIENT_BTC_ADDRESS.getDefaultValue();
+
+        TransactionOutput output = delayedPayoutTx.getOutput(0);
+        NetworkParameters params = btcWalletService.getParams();
+        Address address = output.getAddressFromP2PKHScript(params);
+        if (address == null) {
+            // The donation address can be as well be a multisig address.
+            address = output.getAddressFromP2SH(params);
+            checkNotNull(address, "address must not be null");
```
sqrrm, line 80: This might also cause trouble for `TradeManager` at startup

core/src/main/java/bisq/core/trade/TradeManager.java
chimp1984: The delayedPayoutTx should be available any time. Did you get some error in logs? Can you reproduce it? Was it on regtest or mainnet?

core/src/main/java/bisq/core/trade/DonationAddressValidation.java
chimp1984: Wallet should be initiatialized at that moment. Why do you think it can cause an issue?

core/src/main/java/bisq/core/trade/TradeManager.java
sqrrm: Regtest.

You get that if you test this PR. It will create a trade that doesn't have a delayedPayoutTx

core/src/main/java/bisq/core/trade/DonationAddressValidation.java
sqrrm: If you ever had a malformed tx you wouldn't be able to start up. In general I think it should work.

chimp1984:
Replaced by https://github.com/bisq-network/bisq/pull/4138

PR #4136, opened 2020-04-07
Add check for disputes for donation address

chimp1984:
Depends on https://github.com/bisq-network/bisq/pull/4134


ripcurlx:
@chimp1984 Please have a look at the conflicts, so I can review and merge this PR for the next release. Thanks!

sqrrm:
The first commit is no longer needed as it's now handled in the trade protocol.

The second commit I'm not sure it's a good idea to have. It adds complexity and doesn't really add much value. The main thing would be to find out that there is a bad address in use, but that could be done easier I think.

chimp1984:
> doesn't really add much value. The main thing would be to find out that there is a bad address in use, but that could be done easier I think.

I think it does add value as it adds another protection level at the moment when the refund agent would do the payout to avoid invalid payouts.

ripcurlx:
@chimp1984 If you resolve the conflicts I'm happy to merge this PR for v1.3.7. Thanks!

chimp1984:
I will try to find time....

chimp1984:
I did not had time for testing. It would require a bit of testing before release....

chimp1984:
Replaces by #4516

PR #4140, opened 2020-04-08
Fix range delayedpayout

sqrrm:

Your PR description here.


ripcurlx (approved):
ACK

Tested on Regtest the cases of taking malicious offers and how behavior is if you are already in a trade .

PR #4143, opened 2020-04-08
Fix issue with connected txoutput

chimp1984:


sqrrm (approved):
ACK

chimp1984:
Testing scenarios:
Alice is buyer, Bob is seller.

Test address validation:
- Alice has Dao activated, Bob activated.
- Alice has Dao deactivated, Bob activated.
- Alice has Dao activated, Bob deactivated.
- Alice has Dao deactivated, Bob deactivated.

Bob takes offer from Alice and both restart afterwards. Expected: No error or popup.

Change DAO param for donation address. Adjust new address in hardCodedAddresses at DelayedPayoutTxValidation. Repeat above tests.

Test payout tx is valid for publishing:
Do trades and open mediation. Close mediation ticket and open arbitration. Btc core should receive the exact trade amount from delayed payout tx (default address is taken from btc core as well as dao param change address). Use range trades with less then max. amount to see if all is working correctly as well.

PR #4157, opened 2020-04-10
Unfail with reattach

sqrrm:
Unfail trades from failed tab with ctrl+y

A check is done if the trade can be unfailed, first by checking that it has the required txs (deposittx and delayedpayouttx) and that the addresses needed are available. If they are the trade is moved to pending trades and the addresses tagged appropriately to allow for mediation or normal completion of the trade.


sqrrm:
This PR is now ready for review

wiz (approved):
ACK
tested OK with v1.3.1 tag + this PR to resolve multiple failed trades with real users
@ripcurlx please cherry-pick for v1.3.2 :heart:

ripcurlx (approved):
ACK

Tested it on Regtest with one of my failed trades and was able to complete the trade after unfailing.

PR #4159, opened 2020-04-11
Remove PersistableNetworkPayloadList & TradeStatisticsList

stejbac:
This is another refactoring PR to continue from the work in #4025, by removing `PersistableNetworkPayloadList` as suggested in the review. This together with #4102 removes all the remaining unnecessary `PersistableEnvelope` implementations.

This class is dead code, together with its store service, as they were only referenced from `CorePersistenceProtoResolver::fromProto`, the binding logic and from `AppendOnlyDataStoreService` by orphaned migration code. However, migration from the old persisted data was completed long ago and the store file is no longer being read or written from anywhere in the codebase.

Also remove its associated `PersistableEnvelope` proto message type, along with the `TradeStatisticsList` message type. The latter is long deprecated and has no corresponding Java class implementing `PersistableEnvelope`, so removing it won't change behaviour (outside the exception message thrown when attempting to resolve it).


freimair (approved):
- reviewed the code. nothing surpising there.
- compiled and ran it, no surprises there either
- started a mainnet client, no previously unseen exceptions
- be aware that I only did a limited smoke test
- but it looks good overall

its an ACK from me

sqrrm (approved):
utACK

PR #4168, opened 2020-04-14
Secondary sort order in Market Offer Lists

### LINKED ISSUE #4167: Secondary sort order in Offer lists
### Description

There is no way to manually sort offers in [Market]->[Offer Book].

Being that way, it limits the use of this view because one cannot see at first glance what is up for offer in terms of volumes available.

#### Version

all

### Steps to reproduce

start up Bisq.

### Expected behaviour

have a secondary sort order sorting offers by volume if there are offers with the same price

### Actual behaviour

The sort order is kind of random. But not in a good way. I have seen sort orders where one trader is on top, another one is below. They both had same price and similar volumes. This favors the "top" trader.

### Screenshots

Screenshot from 2020-04-14 15-06-21

freimair:
Fixes #4167

before: 
Screenshot from 2020-04-14 15-06-21

after:
Screenshot from 2020-04-14 16-35-39

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartViewModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartViewModel.java
@@ -354,10 +354,18 @@ class OfferBookChartViewModel extends ActivatableViewModel {
             }
         });
 
+        // 1st comparator: sort by price
+        // - ascending, if showing offers to buyers
+        // - descending, if showing offers to sellers
         if (reversePrimarySortOrder)
             primary = primary.reversed();
 
-        return primary.thenComparing(Offer::getAmount);
+        // 2nd comparator: sort by amount, in descending order
+        // The goal is to show the more attractive offers at the top
+        // Both buyers and sellers would prefer to see higher amounts first (for offers with the same price)
+        Comparator<Offer> secondary = Comparator.comparing(Offer::getAmount, Comparator.reverseOrder());
+
+        return primary.thenComparing(secondary);
     }
 
     private void buildChartAndTableEntries(List<Offer> sortedList,
--- a/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartViewModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartViewModel.java
@@ -280,18 +280,7 @@ class OfferBookChartViewModel extends ActivatableViewModel {
                 .map(OfferBookListItem::getOffer)
                 .filter(e -> e.getCurrencyCode().equals(selectedTradeCurrencyProperty.get().getCode())
                         && e.getDirection().equals(OfferPayload.Direction.BUY))
-                .sorted((o1, o2) -> {
-                    long a = o1.getPrice() != null ? o1.getPrice().getValue() : 0;
-                    long b = o2.getPrice() != null ? o2.getPrice().getValue() : 0;
-                    if (a != b) {
-                        if (CurrencyUtil.isCryptoCurrency(o1.getCurrencyCode()))
-                            return a > b ? 1 : -1;
-                        else
-                            return a < b ? 1 : -1;
-                    } else {
-                        return 0;
-                    }
-                })
+                .sorted(getComparator(false))
                 .collect(Collectors.toList());
 
         final Optional<Offer> highestBuyPriceOffer = allBuyOffers.stream()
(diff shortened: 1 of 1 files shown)


desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartViewModel.java
```diff
@@ -355,6 +333,26 @@ private void updateChartData() {
         buildChartAndTableEntries(allSellOffers, OfferPayload.Direction.SELL, sellData, topSellOfferList);
     }
 
+    private Comparator<Offer> getComparator(boolean reversePrimarySortOrder) {
```
ripcurlx: Maybe let the method name speak a little bit more for itself.

ripcurlx:
ACK from a functionality point of view. Could you add this to the actual Offerbook as well?

desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartViewModel.java
```diff
@@ -355,6 +333,26 @@ private void updateChartData() {
...
+                    return a < b ? 1 : -1;
+            } else {
+                return 0;
+            }
+        });
+
+        if (reversePrimarySortOrder)
+            primary = primary.reversed();
+
+        return primary.thenComparing(Offer::getAmount);
```
cd2357: Minor suggestion to ensure orders with higher amounts are displayed first:

```suggestion
        // 1st comparator: sort by price
        // - ascending, if showing offers to buyers
        // - descending, if showing offers to sellers
        if (reversePrimarySortOrder)
            primary = primary.reversed();

        // 2nd comparator: sort by amount, in descending order
        // The goal is to show the more attractive offers at the top
        // Both buyers and sellers would prefer to see higher amounts first (for offers with the same price)
        Comparator<Offer> secondary = Comparator.comparing(Offer::getAmount, Comparator.reverseOrder());

        return primary.thenComparing(secondary);
```

freimair:
> Could you add this to the actual Offerbook as well?

Actually, I would prefer it that is a separate issue. Reasons are that it is not trivial to code that into Bisq, because the other lists can be sorted by clicking the column headers.

> ensure orders with higher amounts are displayed first

I chose this representation because it is better than having it all scrambled up and because more BSQ trades are done with small amounts at the moment.

desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartViewModel.java
sqrrm: I think this is the more natural way to show the order book, larger orders first. That way more of the book is shown sooner.

desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartViewModel.java
```diff
@@ -333,7 +333,14 @@ private void updateChartData() {
...
 
-    private Comparator<Offer> getComparator(boolean reversePrimarySortOrder) {
+    /**
+     * Returns a comparator to be used for Offers. Sorts primarily for price and
+     * secondarily for offer volume.
+     *
+     * @param reversePrimarySortOrder
+     * @return
+     */
+    private Comparator<Offer> getComparatorWithSecondarySortOrder(boolean reversePrimarySortOrder) {
```
dmos62, line 343: I think this name is preferable: `getOfferComparator(...)`. The problem with `getComparator` was just that it didn't mention the object of comparison.

dmos62 (changes requested):
NACK

The routine is very complicated, especially the primary comparator, because we're checking inside the comparator (twice, for some reason) for whether we're comparing sell offers or buy offers and then we're reversing order inside it as well. That shouldn't happen in the comparator. The comparator in both sell and buy cases should be `Comparator.comparing(Offer::getPrice).thenComparing(Offer::getAmount)`, with a `.reversed.()` inserted where needed, for example: `Comparator.comparing(Offer::getPrice).reversed().thenComparing(Offer::getAmount)`

The rest of the functionality should be achieved by filtering the stream on whether it's a buy/sell (which is already being done at https://github.com/bisq-network/bisq/blob/f3dc42e2ed1fc2f2560b541a18cd996a7d23b9ed/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartViewModel.java#L282 and https://github.com/bisq-network/bisq/blob/f3dc42e2ed1fc2f2560b541a18cd996a7d23b9ed/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartViewModel.java#L311); if nulls have to be handled (can there be offers without prices?), that should happen there as well (presuming we're not showing offers without prices).

desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartViewModel.java
dmos62: Not requesting this change, since the whole method should be unnecessary.

freimair:
@dmos62 originally, there is some double-check whether the Price is a crypto currency or not. Idk where that came from or what it is good for. But surely, there is a reason for that. So I kept it and just made it work. To your suggestion, of course, in a perfect world, that would be what we should do. Fully agree on that.

Please go ahead an open a PR doing it just like that.

I cannot not spend more time on that.

dmos62:
Edit: posted link to branch with my changes, but then realised it will be easier to review if I squash to one commit and post to a second PR that supersedes this one.

dmos62:
@ripcurlx @sqrrm superseded by https://github.com/bisq-network/bisq/pull/4259.

sqrrm:
Can this PR be closed?

PR #4173, opened 2020-04-15
Add second donation address as valid

sqrrm:
Allow using old donation address until we have a better solution. Avoid sending trades to failed when they have properly formed delayed payout txs.


ripcurlx (approved):
utACK

PR #4174, opened 2020-04-15
Update hostname and IP address of BTC node lva54pnbq2nsmjyr.onion

devinbileck:
I deployed a new instance and as a result have a new IP address.
Also, while I was at it I added it to my bisq.services domain.


ripcurlx:
> I deployed a new instance and as a result have a new IP address.

So I guess this means I should filter out your current IP address, right?

devinbileck:
Sure. I guess for those that don't use tor?

ripcurlx:
> Sure. I guess for those that don't use tor?

Yes, I just filtered `165.227.34.198:8333, btc1.dnsalias.net:8333`

ripcurlx (approved):
ACK - Checked IP on https://bitnodes.io and everything ` 172.105.21.216:8333 /Satoshi:0.19.1/` seems to be fine.

ripcurlx:
@devinbileck Please also ping @wiz in that case that he can update his alert script. Thanks!

PR #4178, opened 2020-04-18
When accepting an offer do not round the BTC amount outside range

### LINKED ISSUE #3871: Price rounding with fixed price
* The price for an offer is fixed e.g. 6452.3042
* The maximum amount for this offer has been chosen: 0.05
* The fiat price results in having decimal places e.g. 6452.3042 * 0.05 = 322.61521
* The fiat price is rounded up 323.00
* The bitcoin amount is adjusted 323.00 / 6452.3042 = 0.050059636
* The bitcoin amount is now higher than the maximum amount to be traded
* The validation fails and the offer can not be taken

bisq

**Workaround**: Adjust the amount to buy manually 
**Improvement**: The final fiat amount should be rounded down in this case e.g. 322.00

ghost:
Fixes #3871.  The BTC amount of an offer was being adjusted up in certain cases due to a rounding calculation from the fiat value and price.  This would prevent some offers from being taken.

There's an existing check to ensure that the amount is not adjusted below the minimum, here we add a check to ensure that the amount is not adjusted higher than the maximum defined in the offer.


sqrrm (approved):
utACK

PR #4186, opened 2020-04-21
Limit MaxRAM for all java apps

ghubstan:
Setting DEFAULT_JVM_OPTS="-XX:MaxRAM=4g" in bisq-desktop has not
resulted in any reported issues, so this jvm option can be used in other
launch scripts to reduce the amount of memory reserved by headless
apps.  This change is based on the assumption none of the headless
apps require as much memory as desktop.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -137,11 +137,9 @@ configure([project(':cli'),
             unixScriptFile.text = unixScriptFile.text.replace(
                 'cd "`dirname \\"$PRG\\"`/.." >/dev/null', 'cd "`dirname \\"$PRG\\"`" >/dev/null')
 
-            if (applicationName == 'desktop') {
-                def script = file("${rootProject.projectDir}/bisq-$applicationName")
-                script.text = script.text.replace(
-                    'DEFAULT_JVM_OPTS=""', 'DEFAULT_JVM_OPTS="-XX:MaxRAM=4g"')
-            }
+            def script = file("${rootProject.projectDir}/bisq-$applicationName")
+            script.text = script.text.replace(
+                'DEFAULT_JVM_OPTS=""', 'DEFAULT_JVM_OPTS="-XX:MaxRAM=4g"')
 
             if (osdetector.os != 'windows')
                 delete fileTree(dir: rootProject.projectDir, include: 'bisq-*.bat')


cbeams:
This PR essentially reverts @freimair's commit https://github.com/bisq-network/bisq/pull/4048/commits/f127c28d1d887bd065d856598f3bf179382644e3 in PR #4048. Just to be clear, was the 4g limitation narrowed down to desktop there just to be conservative, or was there a more specific reason?

cbeams:
It is also the case that the `-XX:MaxRAM=4g` option added in PR #4048 has only been out there for a few days with the release of v1.3.2. Granted, we had the pre-release period before that, but just for completeness sake, there could be memory / performance issues with long-running instances that haven't cropped up or been reported yet. I'm not saying we shouldn't merge this, but that maybe we want to wait a while, closer to the next release to do so.

To be really safe, we should probably have one or two of our seed node operators run with this flag for a while and make sure nothing adverse happens.

ghubstan:
> This PR essentially reverts @freimair's commit [f127c28](https://github.com/bisq-network/bisq/commit/f127c28d1d887bd065d856598f3bf179382644e3) in PR #4048.

@cbeams, I can't speak for @freimair, but I think he wanted to be careful, and experiment with the setting only on desktop where it was needed most.

While working on :cli, I noticed :daemon was using about as much resident memory as :desktop, and realized this setting is needed for all apps (and I wanted to get this PR out of the way before returning to work on :cli).

It was a mistake to imply the setting had been in use by everyone running :desktop during the last few weeks, and I also prefer being "really safe" by asking some node operators to try it before it's merged.

freimair:
> @cbeams, I can't speak for @freimair, but I think he wanted to be careful, and experiment with the setting only on desktop where it was needed most.

@ripcurlx and I agreed on this strategy yes. We wanted to see if it actually makes a difference and if problems arise. I would have gone all-in but that is what we settled on.

Also, we have no data on whether it is necessary or even has an effect on eg. Windows platforms.

dmos62:
Others might already be aware of this, but I'll point it out for the sake of completeness that `MaxRAM` is meant to specify the total/real memory size, and `ErgoHeapSizeLimit` is for saying "I know there's more memory, but you should only use this much" [0]. And, as I understand (docs are hard to find), JVM will try to satisfy these constraints, but will violate them if need be; i.e. setting either of them too low won't cause JVM to kill itself or anything like that.

For example setting `MaxRAM` to total RAM in machine and `ErgoHeapSizeLimit` to something reasonable like 4GB would be the intended (not necessarily best or even effective) way to limit RAM usage.

If anyone finds the actual docs for this, I would be interested in reading those.

[0] https://chriswhocodes.com/hotspot_options_jdk10.html?s=MaxRam

ghubstan:
not stale

sqrrm:
@ghubstan @freimair @ripcurlx It there some kind of conclusion to what we should do here?

ghubstan:
> Is there some kind of conclusion to what we should do here?

My 2 sats:  I think the riskiest aspect of this change was adding the `MaxRAM` setting to `bisq-desktop`.  Since being tried out by devs and end users of the released change:
* Chris reported 1 `:desktop` OOME some months ago.
* I saw one after running a `:desktop` instance for a ~week on a 10+ year old 8 Gb Dell server.  My OOME happened while resident memory was still < 2 Gb, so I can't be sure it was due to the 4g MaxRAM setting.

That's 2 OOMEs that I know of being reported since the change has been in use by Linux/OSX devs and users, and not enough to warrant backing out the prior `MaxRAM` setting in `:desktop`.

My conclusion for this PR:  JavaFX's appetite is why the change was needed, so I'm now in favor of letting this PR go stale unless others want to see it merged.

PR #4192, opened 2020-04-23
Add BTC node qxjrxmhyqp5vy5hj.onion

devinbileck:
This is a newly deployed instance intended on replacing one of @Emzy's
nodes in an attempt to reduce the number of nodes under the control of a
single contributor.


ripcurlx (approved):
ACK - checked 173.255.240.205:8333 running Satoshi:0.19.1

PR #4193, opened 2020-04-23
Display "Payment Method" column in Open Offers view

### LINKED ISSUE #3902: Portfolio My open offers Instant altcoins not flagged
### Description

Expect to be able to easily identify offers that are instant or not.
only way to identify them is clicking info, or memorizing the trade info.

#### Version

1.2.5
### Expected behaviour

view My Open Offers

### Actual behaviour

altcoin offers and instant altcoin offers appear the same.

#### Device or machine

Mac OS 10.15.2
Ubuntu 18.4, 19.10

#### Additional info

would also be helpful to be able to set a schedule for instant alts.  or have a button to turn them all on/off.

### LINKED ISSUE #4125: Feature Request: add column to Portfolio -> My Open Offers
### Description

I trade in multiple Bisq payment methods. The current view of `My Open Offers` gives me a good overview of the spreads I'm asking around, but a critical piece is left out: the Payment Method.

So I'd like to propose that `My Open Offers` gets a bit closer to `Open Trades` and have Payment method there too

#### Version

1.2.9

ghost:
Adjust widths of existing columns to fit in a column for payment method.

Fixes #3902
Fixes #4125
implementation screenshot: https://github.com/bisq-network/bisq/issues/4125#issuecomment-618663526

image

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersView.java
+++ b/desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersView.java
@@ -69,7 +69,7 @@ public class OpenOffersView extends ActivatableViewAndModel<VBox, OpenOffersView
     @FXML
     TableColumn<OpenOfferListItem, OpenOfferListItem> priceColumn, amountColumn, volumeColumn,
             marketColumn, directionColumn, dateColumn, offerIdColumn, deactivateItemColumn,
-            removeItemColumn, editItemColumn;
+            removeItemColumn, editItemColumn, paymentMethodColumn;
     private final Navigation navigation;
     private final OfferDetailsWindow offerDetailsWindow;
     private SortedList<OpenOfferListItem> sortedList;
@@ -84,6 +84,7 @@ public class OpenOffersView extends ActivatableViewAndModel<VBox, OpenOffersView
 
     @Override
     public void initialize() {
+        paymentMethodColumn.setGraphic(new AutoTooltipLabel(Res.get("shared.paymentMethod")));
         priceColumn.setGraphic(new AutoTooltipLabel(Res.get("shared.price")));
         amountColumn.setGraphic(new AutoTooltipLabel(Res.get("shared.BTCMinMax")));
         volumeColumn.setGraphic(new AutoTooltipLabel(Res.get("shared.amountMinMax")));
@@ -101,6 +102,7 @@ public class OpenOffersView extends ActivatableViewAndModel<VBox, OpenOffersView
         setPriceColumnCellFactory();
         setAmountColumnCellFactory();
         setVolumeColumnCellFactory();
+        setPaymentMethodColumnCellFactory();
         setDateColumnCellFactory();
         setDeactivateColumnCellFactory();
         setEditColumnCellFactory();
@@ -116,6 +118,7 @@ public class OpenOffersView extends ActivatableViewAndModel<VBox, OpenOffersView
         priceColumn.setComparator(Comparator.comparing(o -> o.getOffer().getPrice(), Comparator.nullsFirst(Comparator.naturalOrder())));
         volumeColumn.setComparator(Comparator.comparing(o -> o.getOffer().getVolume(), Comparator.nullsFirst(Comparator.naturalOrder())));
         dateColumn.setComparator(Comparator.comparing(o -> o.getOffer().getDate()));
+        paymentMethodColumn.setComparator(Comparator.comparing(o -> o.getOffer().getPaymentMethod().getId()));
 
         dateColumn.setSortType(TableColumn.SortType.DESCENDING);
         tableView.getSortOrder().add(dateColumn);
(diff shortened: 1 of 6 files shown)


desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersView.java
```diff
@@ -116,6 +118,10 @@ public void initialize() {
         priceColumn.setComparator(Comparator.comparing(o -> o.getOffer().getPrice(), Comparator.nullsFirst(Comparator.naturalOrder())));
         volumeColumn.setComparator(Comparator.comparing(o -> o.getOffer().getVolume(), Comparator.nullsFirst(Comparator.naturalOrder())));
         dateColumn.setComparator(Comparator.comparing(o -> o.getOffer().getDate()));
+        paymentMethodColumn.setComparator(Comparator.comparing(
+                o -> o.getOffer() != null ? o.getOffer().getPaymentMethod().getId() : null,
```
dmos62: Is the null check necessary? The above calls to result of `o.getOffer()` don't null check.

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersViewModel.java
```diff
@@ -19,6 +19,7 @@
 
 import bisq.desktop.common.model.ActivatableWithDataModel;
 import bisq.desktop.common.model.ViewModel;
+import bisq.desktop.main.portfolio.pendingtrades.PendingTradesListItem;
```
dmos62: Unused import.

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersViewModel.java
```diff
@@ -118,6 +121,23 @@ String getMarketLabel(OpenOfferListItem item) {
         return CurrencyUtil.getCurrencyPair(item.getOffer().getCurrencyCode());
     }
 
+    String getPaymentMethod(OpenOfferListItem item) {
+        String result = "";
+        if (item != null) {
+            Offer offer = item.getOffer();
+            checkNotNull(offer);
+            checkNotNull(offer.getPaymentMethod());
+            String method = Res.get(offer.getPaymentMethod().getId() + "_SHORT");
```
dmos62: Can we guarantee that `Res.get(paymentMethod.getId() + "_SHORT")` will always find the translation? If unsure, should wrap in try-catch.

dmos62 (changes requested):
NACK due to a couple minor things. Changes seem in good shape otherwise. Haven't tested.

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersView.java
ghost: I agree, and will modify to bring it into line with the code above it.

Just to note, this line was copied from the existing PendingTradesView.java in which there is a payment method column.  Should I also make a change to do the same in PendingTradesView.java or would that need to be a separate code-cleanup type PR?

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersViewModel.java
ghost: This method is copied directly from PendingTradesViewModel.java (line 273).  I presume since the code is already working, no need to change it.  I'm open to suggestions of changing both though, if you have strong worries about it.

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersViewModel.java
ghost: :+1: will remove.

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersViewModel.java
dmos62: It would be good to move this duplicating logic to [`Offer`](https://github.com/bisq-network/bisq/blob/master/core/src/main/java/bisq/core/offer/Offer.java) itself (`Offer.getPaymentMethodNameWithCountryCode()` or something better). Put in a try-catch in that method for `java.util.MissingResourceException` (that's what `Res.get` might throw) and use an Option return type to distinguish whether or not the translation was found (again, since we can't guarantee (I presume) that someone won't forget to put in a translation when adding a new payment method). Extra points for extracting the `Res.get(paymentMethod().getId() + "_SHORT")` part into something like `PaymentMethod.getShortName()` (returns an Option). If you do less than I suggest, no one will probably mind, but it's great to leave the code just a little better.

Edit: at first suggested moving to `PaymentMethod`, but then realised the logic relies on `Offer`. Corrected.

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersView.java
dmos62: I think it's fine if you leave the other one as is.

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersViewModel.java
ghost: :+1: Done / refactored.  `Res.get` handles `MissingResourceException` itself and in that case returns the key.  I think that is a good way of handling the case of missing translations since it alleviates the caller from having to write custom error handling everywhere.

dmos62: @jmacxx I missed that, good catch.

dmos62:
untested ACK. Changes super tidy.

ghost:
Squashed-in the changes.  @dmos62 thanks for the review!

dmos62:
@jmacxx thanks for the refactor!

sqrrm (approved):
utACK

PR #4195, opened 2020-04-25
Fix Trader Chat row selection and unread message count

### LINKED ISSUE #4172: Clicking on the trader chat icon should select the trade
### Description

With multiple open trades, clicking on the trader chat icon of a trade that is not selected should select that row in the trade list.  It does not.  This causes confusion as it gives misleading cues as to which trader you are communicating with.  

#### Version

1.3.1

### Steps to reproduce

You have two open trades (A and B).
Select trade A in the table.  Its details are shown in the main part of the screen.
Then click on the trader chat icon of trade B in the table.

### Expected behaviour

Trade B should be highlighted in the table, the details of trade B should be displayed in the trade details section, and trader chat screen for trade B should open.

### Actual behaviour

Trade A remains highlighted in the table, details of trade A are still displayed in the trade details section, and trader chat screen opens for trade B.

ghost:
Clicking on the trader chat icon of a trade that is not selected should select that row in the trade list. It does not. This causes confusion as it gives misleading cues as to which trader you are
communicating with. Issue a call to select the row of the button when clicked.

The trader chat view can be opened either by clicking on the chat icon of a trade in the list, or clicking on "OPEN TRADER CHAT" button for the currently displayed trade.  In the latter case, the count of new messages displayed on the chat icon was not cleared even though trader chat was shown.  The solution is to move the scope of the routine that updates trader chat message count from within the table cell handler to the PendingTradesView class so that it can be called whenever
onChat() is invoked (i.e. when the Trader Chat screen is shown).

Fixes bisq-network#4172

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/PendingTradesView.java
+++ b/desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/PendingTradesView.java
@@ -442,8 +442,26 @@ public class PendingTradesView extends ActivatableViewAndModel<VBox, PendingTrad
         // Delay display to next render frame to avoid that the popup is first quickly displayed in default position
         // and after a short moment in the correct position
         UserThread.execute(() -> chatPopupStage.setOpacity(1));
+        updateChatMessageCount(trade, badgeByTrade.get(trade.getId()));
     }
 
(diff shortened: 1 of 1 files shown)


ghost:
Still relevant.

sqrrm (approved):
utACK

PR #4209, opened 2020-04-29
Migrate expect based rpc test suite to bats

ghubstan:
The [bats](https://github.com/sstephenson/bats) automated bash testing system is more concise and easier to work with than the older `tcl - expect` based test program.


cbeams (approved):
ACK, with a couple polishing commits added.

PR #4211, opened 2020-04-29
Add link to Lombok plugin in IDEA setup guide

cd2357:
In the latest version of IDEA (2020.1), the Lombok plugin could not be found via the Plugins menu, as per the setup guide.

Apparently there were some incompatibilities between the latest versions of the plugin / of IDEA, which lead to it appearing as incompatible with this version.

Therefore, I added a link to the plugin's page (JetBrains Plugin Repository). The linked page contains the plugin changelog (describing the version incompatibility + fix). It also contains a top-right button called "Install to IDE", which properly installs the plugin in IDEA 2020.1.

### DIFF
--- a/docs/idea-import.md
+++ b/docs/idea-import.md
@@ -4,7 +4,7 @@ Most Bisq contributors use IDEA for development. The following instructions have
 
  1. Follow the instructions in [build.md](build.md) to clone and build Bisq at the command line.
  1. Open IDEA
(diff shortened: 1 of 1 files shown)


cd2357:
Seems this is not necessary after all.

The plugin incompatibility problem I mentioned in the commit was because of an issue with my local setup.

I tried it now in a new environment, fresh IDEA install, and the original steps in the tutorial work fine.

No need to add the link to the plugin homepage anymore, therefore closing the PR.

PR #4212, opened 2020-04-29
Bugfix and typofix

dmos62:
Two omitted breaks in switch statements and one typo. One of the omitted breaks is serious-ish, info in commit messages.

### DIFF
--- a/core/src/main/java/bisq/core/dao/governance/proposal/param/ChangeParamValidator.java
+++ b/core/src/main/java/bisq/core/dao/governance/proposal/param/ChangeParamValidator.java
@@ -144,6 +144,7 @@ public class ChangeParamValidator extends ProposalValidator implements Consensus
             case REIMBURSEMENT_MIN_AMOUNT:
                 checkArgument(inputValueAsCoin.value >= Restrictions.getMinNonDustOutput().value,
                         Res.get("validation.amountBelowDust", Restrictions.getMinNonDustOutput().value));
+                break;
             case COMPENSATION_REQUEST_MAX_AMOUNT:
             case REIMBURSEMENT_MAX_AMOUNT:
                 checkArgument(inputValueAsCoin.value >= Restrictions.getMinNonDustOutput().value,
--- a/core/src/main/java/bisq/core/dao/node/parser/TxOutputParser.java
+++ b/core/src/main/java/bisq/core/dao/node/parser/TxOutputParser.java
@@ -271,12 +271,12 @@ class TxOutputParser {
 
                         // If it is the vote stake output we return false.
                         if (index == 0) {
-                            return false;
+                            break;
                         }
 
                         // There must be a vote fee left
                         if (availableInputValue <= 0) {
-                            return false;
+                            break;
                         }
 
                         // Burned BSQ output is last output before opReturn.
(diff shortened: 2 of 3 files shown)


sqrrm (approved):
utACK

sqrrm:
@ManfredKarrer I merged too soon. I didn't look at the two commits touching the DAO until after merging. I think it's correct but sensitive stuff.

core/src/main/java/bisq/core/dao/governance/proposal/param/ChangeParamValidator.java
```diff
@@ -144,6 +144,7 @@ private void validateBsqValue(Coin currentParamValueAsCoin, Coin inputValueAsCoi
             case REIMBURSEMENT_MIN_AMOUNT:
                 checkArgument(inputValueAsCoin.value >= Restrictions.getMinNonDustOutput().value,
                         Res.get("validation.amountBelowDust", Restrictions.getMinNonDustOutput().value));
+                break;
```
ManfredKarrer, line 147: I think the correct version would be:
```
            case REIMBURSEMENT_MIN_AMOUNT:
            case COMPENSATION_REQUEST_MAX_AMOUNT:
            case REIMBURSEMENT_MAX_AMOUNT:
                checkArgument(inputValueAsCoin.value >= Restrictions.getMinNonDustOutput().value,
                        Res.get("validation.amountBelowDust", Restrictions.getMinNonDustOutput().value));
                checkArgument(inputValueAsCoin.value <= 200000000,
                        Res.get("validation.inputTooLarge", "200 000 BSQ"));
                break;
```
Also MinAmount must not exceed the 200 000 BSQ.

core/src/main/java/bisq/core/dao/node/parser/TxOutputParser.java
```diff
@@ -285,14 +285,15 @@ private boolean isBtcOutputOfBurnFeeTx(TempTxOutput tempTxOutput) {
                         // We always have the BSQ change before the burned BSQ output if both are present.
                         checkArgument(optionalOpReturnIndex.isPresent());
                         if (index != optionalOpReturnIndex.get() - 1) {
-                            return false;
+                            break;
```
ManfredKarrer, line 288: I don't see any problem with that change but as it is a pure "cosmetic" change I prefer to not touch the code. Some could argue the early return is more readable. I don't have a strong opinion and the break is more consistent but evaluating benefit / risk just does not justify the change IMO even it seems a trivial and very low risk change.

ManfredKarrer (commented):
NACK

core/src/main/java/bisq/core/dao/node/parser/TxOutputParser.java
```diff
@@ -285,14 +285,15 @@ private boolean isBtcOutputOfBurnFeeTx(TempTxOutput tempTxOutput) {
...
                         }
 
                         // Without checking the fee we would not be able to distinguish between 2 structurally same transactions, one
                         // where the output is burned BSQ and one where it is a BSQ change output.
                         long blindVoteFee = daoStateService.getParamValueAsCoin(Param.BLIND_VOTE_FEE, tempTxOutput.getBlockHeight()).value;
                         return availableInputValue == blindVoteFee;
                     }
+                    break;
```
dmos62, line 296: If the semantic changes in this file are reverted, take care not to revert adding this `break`.

sqrrm: I'm doing a revert of the whole PR. There was an easy to use button to do the revert.

Best make a new PR with the good changes.

PR #4214, opened 2020-04-29
Add rpc wallet protection endpoints

### LINKED ISSUE #4198: Implement wallet protection rpc methods
After implementing gRPC authentication in [4189](https://github.com/bisq-network/bisq/pull/4189), it makes sense to implement wallet encryption / decryption endpoints in the next development phase.

A new PR will will close this issue after the following tasks are completed.

- [x] Add error handling to `getbalance` method.
      Balances are not available while :daemon is initializing or when a wallet is locked.  Currently, the user sees `Error: UNKNOWN` for any server-side error.  Tell users to wait or how to unlock their wallet.

- [x] Implement `setwalletpassword` method:  `setwalletpassword "password" "newpassword"` with error handling.  There are two use cases:
          * A single `"password"` argument is provided to encrypt an unencrypted wallet.
          * An optional second `"newpassword"` argument is is provided to change the password on an encrypted wallet.
      This is the analog to bitcoin core's [encryptwallet](https://bitcoincore.org/en/doc/0.19.0/rpc/wallet/encryptwallet)  and [walletpassphrasechange](https://bitcoincore.org/en/doc/0.19.0/rpc/wallet/walletpassphrasechange) methods, combined into one.

- [x] Implement `removewalletpassword` method:  `removewalletpassword "password"` with error handling.
      Automated testing with ([bats](https://github.com/sstephenson/bats)) is a necessity for :cli, and despite the lack of a use case for a `removewalletpassword` method in Bisq, and the security risks it exposes, a  `removewalletpassword` method will allow devs to quickly run the test suite without having to wait for :daemon's full initialization to complete before any wallet related operations can be performed.  This method should not be mentioned in help text.

- [x] Implement `unlockwallet "password" timeout` method, where `timeout` is in seconds.
      This is the analog to bitcoin core's [walletpassphrase](https://bitcoincore.org/en/doc/0.19.0/rpc/wallet/walletpassphrase)  method.

- [x] Implement `lockwallet` method.
      This is the analog to bitcoin core's [walletlock](https://bitcoincore.org/en/doc/0.19.0/rpc/wallet/walletlock)  method.  After calling this method, users will need to call `unlockwallet` again before being able to call any methods which require the wallet to be unlocked.

- [x] Port the `expect` based test.sh script in the cli module to [bats](https://github.com/sstephenson/bats) (easier to read and write).

ghubstan:
Implemented `lockwallet`, `unlockwallet`, `removewalletpassword`, and `setwalletpassword` methods with basic error handling.

Also added basic error handling to the existing `getbalance` method, and removed unused `BalancePresentation` from `CoreAPI`.

Resolves #4198

#### Notes

I didn't spend time second-guessing maintainers about the ordering of rpc service defs & methods in grpc.proto and GrpcServer, so changes might need to be made in that regard.  It makes sense to group wallet protection related methods together, but that is inconsistent with other alphabetically ordered service and method names.

There is also a `getbalance` error handling bug that needs fixing:  `Error: null` is printed to the :cli console when the daemon  is losing peers (SOCKET_TIMEOUT).

Bats test cases for these methods are ready to push in a new PR after (if) [4209](https://github.com/bisq-network/bisq/pull/4209) is merged.


ghubstan:
There is a bug needing a fix and test case for `CliMain` at line 174:
`if (nonOptionArgs.size() < 2)` should be if `(nonOptionArgs.size() < 3)`.
FIXED

cbeams:
Note that I changed the description from "To close task list in #4198" to "Resolves #4198" so that this PR shows up as "linked" in #4198 and so that it will in fact close #4198 when merged. See https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword.

cbeams (changes requested):
NACK. @ghubstan, I have not comprehensively reviewed this yet, but it's clear so far that error handling will need some rework. You've built in `success` and `error_message` fields into the new `*Reply` definitions, but this is not idiomatic gRPC usage. gRPC provides a first-class `Status` object, which should be accessed via `StatusRuntimeExceptions` thrown from the server side and caught on the client. We're doing this already with invalid passwords, but we now need to do it more consistently from our services.

See https://grpc.io/docs/tutorials/basic/java/#calling-service-methods and https://github.com/grpc/grpc/blob/master/doc/statuscodes.md for details.

cbeams:
Also, and it's my bad for not specifying this, we probably should have cut this up into multiple PRs implementing one endpoint at a time, or at least separate commits for each. I'm not suggesting that you re-do everything in separate PRs/commits here, but it does mean more re-work when we find something like the `Status` issue above. If we'd caught that in a single endpoint implementation, the other's wouldn't have needed rework. Another reason to do one PR per endpoint is from a security review perspective. The less code there is to look at, the more likely it is that we notice subtle problems.

ghubstan:
Notes about commit 2a9d1f6 (WIP)

@cbeams, 

I know you will look at all of the changes, but please review my use of the Tuple2 return type used in CoreApi?  Wrapping the return value and an ApiStatus code in a single return value -- for void return types too, unfortunately -- seemed like a better idea than throwing Exceptions to be caught by GrpcServer, to then be transformed into gRPC StatusRuntimeExceptions for the client.  I don't know what your preference is.  I prefer to make the changes after you ask for them.


And there is a lot of redundant error handling code in GrpcServer:

```
if (!result.second.equals(ApiStatus.OK)) {
	StatusRuntimeException ex = new StatusRuntimeException(result.second.getGrpcStatus()
	        .withDescription(result.second.getDescription()));
	responseObserver.onError(ex);
	throw ex;
}

```
There are only service classes in GrpcServer, no private methods yet.  I don't know if you want these duplicate code blocks factored out.

cbeams:
@ghubstan, in preparation to more fully review the use of `Tuple2` and the other things you've asked about above, I've committed 39c868f which refactors the way we're declaring wallet-related grpc services. Have a look, and I'll get back to my review here as soon as I can. In short, I do want to eliminate `Tuple2` and go with exceptions, but I need to think about it / try things out a bit further myself.

cbeams:
I doubt, by the way, @ghubstan, that the way we want to implement `unlockwallet` is by actually unencrypting the wallet with `removewalletpassword`. If the Bisq node were to be shut down before the timeout expires, it would be without a wallet password at all when it starts up again. I am not sure exactly how unlocking the wallet works when using the GUI, but I would guess it doesn't work this way. Perhaps you could look into that in the meantime, while I continue my review?

cbeams:
@ghubstan, regarding error handling, I've landed on an exception-based solution I'm happy with so far, but not quite ready to commit. Just FYI.

cbeams:
@ghubstan, have a look at what I've done in 163061a to change the return type of `CoreWalletService#getAvailableBalance` from `Tuple2<Long, ApiStatus>` back to `long` and  how throwing `IllegalStateException` is now favored. The status of all `StatusRuntimeExceptions` is now set to `UNKNOWN`, as I don't see a requirement at this point for the client to differentiate between different gRPC status codes. We can introduce that later if need be. For now, this approach keeps `CoreWalletService` completely unaware of gRPC concepts (as if it were really located in `bisq.core` where we want it to be) and it its methods are left returning values and throwing exceptions in an idiomatic / plain old java fashion.

Could you refactor the remaining methods in `CoreWalletService` to follow suit? When finished, `ApiStatus` should be empty and therefore deleted, and there should no longer be any usage of `Tuple2`. Preferred is to refactor each method in its own commit. Thanks.

ghubstan:
RE commit ab17b67:  I was catching exceptions (indicating a bug or bugs to be fixed?) I've seen during testing this method.  Otherwise they will bubble up to client.

RE commit 9164579:  I don't think an extra line for an explicit `return;` is needed for methods that return void.

cbeams:
> RE commit [ab17b67](https://github.com/bisq-network/bisq/commit/ab17b672294011b766da3ab6e02a349c01c4f7e0): I was catching exceptions (indicating a bug or bugs to be fixed?) I've seen during testing this method. Otherwise they will bubble up to client.

I don't see the value in wrapping and rethrowing there. I'd rather see the exception bubble up as-is to the client and then see specific error handling code put wherever it's most appropriate. Wrapping it in an IllegalStateException may not really reflect the nature of the problem, e.g. just a programmer error.

> RE commit 9164579: I don't think an extra line for an explicit return; is needed for methods that return void.

If you don't explicitly return there, then the next`if` block will return `true` and throw because you just finished encrypting the wallet. Prior to your refactoring, you did explicitly return from that block with a `Tuple2`. You either need to return or make the subsequent `if` an `else if` block; your call.

core/src/main/java/bisq/core/grpc/ApiStatus.java
```diff
@@ -56,5 +53,5 @@ public String toString() {
                 ", description='" + description + '\'' +
                 '}';
     }
-}
+    }
```
cbeams: Whitespace. Doesn't matter as `ApiStatus` will be removed shortly enough, but it makes me wonder whether or not you're using Editorconfig as suggested at bisq-network/style#10?

ghubstan:
@cbeams, the `Tuple2<Obj, ApiStatus>` return type is no longer used in `CoreWalletService`, and `ApiStatus` has been deleted from the repo.

There have been some changes in error messages, so the old cli/test.sh needs work, and there are the test cases for the wallet protection methods I haven't checked in yet.

ghubstan:
@cbeams,  And I know there are still problems with the lock/unlock wallet methods.  I should have looked at the bitcoin core source implementations of these methods before asking for help...  I don't think I can look at that until Tuesday.  (Just letting you know.)

core/src/main/java/bisq/core/grpc/ApiStatus.java
ghubstan: My Intellij IDE is using the project scheme defined in the repo's `.editorconfig` file:

```
root = true

[*]
indent_style = space
indent_size = 4
continuation_indent_size = 8
end_of_line = lf
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true

[*.diff]
trim_trailing_whitespace = false

[Makefile]
indent_style = tab

[*.bat]
end_of_line = crlf

[build.gradle]
continuation_indent_size = 4

[.idea/codeStyles/*.xml]
indent_size = 2
insert_final_newline = false

```
It could have been a fat finger error when I was reformatting the class after removing some lines with ^ALT-L.  But I'm really not sure if that is the cause.  I'll keep an eye on it.

ghubstan:
> If you don't explicitly return there, then the next`if` block will return `true` and throw because you just finished encrypting the wallet. Prior to your refactoring, you did explicitly return from that block with a `Tuple2`. You either need to return or make the subsequent `if` an `else if` block; your call.

@cbeams,
I was way off (wrong!) about that.  Sorry.   Fixed in commit f5a4ca5.

cbeams:
Thanks, @ghubstan. I think we're looking pretty good here. I'll give it a final pass and see about merging it as soon as possible.

cbeams:
> I think we're looking pretty good here.

I take that back, actually. There will probably need to be another substantive set of changes to tighten up the various issues you reference in your comment above:

> there are still problems with the lock/unlock wallet methods.... I don't think I can look at that until Tuesday.

No worries on waiting until Tuesday. I'll have limited time tomorrow as well.

ghubstan:
@cbeams , I am still figuring out how to ensure the wallet is encrypted in the event of a crash before  `unlockpassword timeout` expiration.  I think the intermediary commits  fbb025a - 24248d4 above were necessary, but `CoreWalletService` still decrypts/encrypts in the unlock/lock wallet methods.  This will be fixed.

I think an intermediate step in that direction is to decrypt and re-encrypt the wallet only inside the methods that need to temporarily unlock the wallet, instead of  decrypting it in the `unlockwallet` method.  `unlockwallet` would just cache the crypter+key, which would be used by `getbalance` to:  `decrypt wallet -> getbalance -> encrypt wallet -> return balance`.  The  `unlockwallet timeout` task or the `lockwallet` method, whichever happens first,  would clear the temp crypter+key variables (see commit 4262f29).  

I don't think `getbalance` even requires wallet decryption to display an updated balance, but it would be the place to see if this is feasible until a better solution to the wallet protection unlock/lock problem is found, and new methods are implemented.  (Then the wallet decrypt/encrypt code would be removed from `getbalance`.)

Tomorrow, I will run :desktop in regtest mode to see how wallet decryption/encryption works while trading and sending btc/bsq to and from other wallets.  I realize maybe all that's needed to unlock the wallet is the aesKey, but _so far_, bitcoin core src tells me it decrypts the in-memory wallet in its `walletpassphrasechange` method, and Bisq / BitcoinJ src tells me `unlockwallet` will have to decrypt the wallet and save modified wallet files to disk (I could be and hope I'm wrong). 

(On a related matter:  the [UI does have a removepassword function](https://github.com/bisq-network/bisq/issues/4198#issuecomment-624832555).)

EDIT:  This comment is obsolete.  See commit a79ae57.

ghubstan:
The correctness of commit a79ae57 needs to be verified by implementing and testing wallet related methods requiring the aesKey cached in `unlockwallet`.  

For example, a method analogous to bitcoin-cli's `sendtoaddress "addr" amt`, that performs the same operation as the Bisq UI's  `WithdrawalView` # 
        `doWithdraw(amount, fee, callback)` -> `sendFunds(amount, fee, aesKey, callback)`

cbeams:
@ghubstan, I just ran into the following scenario when testing everything locally:

1. spin up a daemon against a new data directory, e.g: `./bisq-daemon --appDataDir=/tmp/newbisqdatadir --apiPassword=xyz`
1. run `./bisq-cli --password=xyz getbalance` until you get a valid response of `0.00000000`
1. set a wallet password with `./bisq-cli --password=xyz setwalletpassword foo`
1. notice that `./bisq-cli --password=xyz getbalance` errors out as expected
1. unlock the wallet with `./bisq-cli --password=xyz unlockwallet foo 60`
1. notice that `./bisq-cli --password=xyz getbalance` returns 0.00000000 once again
1. lock the wallet with `./bisq-cli --password=xyz lockwallet`
1. notice that `./bisq-cli --password=xyz getbalance` errors out as expected
1. unlock the wallet once again with `./bisq-cli --password=xyz unlockwallet foo 60`
1. notice that `./bisq-cli --password=xyz getbalance` does NOT return 0.00000000 as expected, but rather fails with a "balance is not yet available message"
1. notice that no matter how long you wait, the command above always fails with the same message.

It appears there is no way to recover from this state via the cli, and that the only way out is to spin up a fresh data directory. Stopping and starting the server makes no difference either.

ghubstan:
> It appears there is no way to recover from this state via the cli

@cbeams , I haven't been able to reproduce the error yet.  I don't suppose you still have the server log, and could check if the `balance is not yet available` msgs on the cli correspond to lost peer connections and Tor layer log.errors? 

I've seen this before and should have mentioned it.  Fixing tor binaries is out of this PR's scope, but the daemon needs to return a meaningful msg to the cli when this happens.

I've been turning my net connection off & on trying to force the tor errors I think are related to the daemon bug.  No success so far, but I'll leave daemon on in hopes tor starts failing.

cbeams:
@ghubstan, I just tried reproducing this on my side and recorded my efforts in the video at keybase://team/bisq/2020-05-12%20repro%20grpc-api%204214%20bug. In the end, I was unable to reproduce it exactly as I reported it. Please have a look at the video, but I'm not sure if it makes sense to focus on this more if I can't reliably reproduce it myself.

ghubstan:
@cbeams, 

I will see the keybase video later today.  But today I am seeing numerous unrecoverable `balance is not yet available` errors and am working on it now.

An interval of tor circuit breakdown during startup prevented the wallet balance from being updated.  
When I see a series of tor debug statements like these:

	CircuitStatus: 92 CLOSED
	CircuitStatus: 92 CLOSED
	CircuitStatus: 144 FAILED
	CircuitStatus: 148 LAUNCHED
	CircuitStatus: 94 CLOSED
	CircuitStatus: 149 EXTENDED
	CircuitStatus: 148 BUILT
	CircuitStatus: 153 EXTENDED
	.....
	CircuitStatus: 153 BUILT
	CircuitStatus: 134 CLOSED
	...
	...
	Ending in many CLOSED statements
	...
	...
	CircuitStatus: 134 CLOSED

the initial null wallet balance cannot be updated.  

But you saw your balance change from a non-null value to null (then you saw `balance is not yet available`).    I suspect the balance is set to null while the tor circuit is broken, but I haven't reproduced that scenario yet.

When the circuit if finally reconstructed -- however long it takes -- the wallet balance is updated (confirmed by hitting my breakpoint @ `Balances#updateBalance`), and the cli can show the balance.  I think this is the problem you found.  If it is, you would have seen your balance after the tor circuit problems resolved themselves.   Eventually.  (EDIT:  _But maybe not.  Ever. The recovery of the balance after circuit rebuild I saw earlier is not happening now.  I have to take back what I said about "eventual" recovery_.)

 I am discussing with @freimair about how this problem is or might be categorized, and how this state could be communicated to the :cli user.

This problem is related and possibly identical to the **"3/4" problem** described in issue [2547](https://github.com/bisq-network/bisq/issues/2547).

chimp1984:
@ghubstan 
Regarding wallet encryption (not sure if already resolved, have not read the full thread...):

If wallet is encrypted you need the aesKey (or generate it from the password) for certain wallet methods. Bisq does not decrypt/encrypt but requests from the user the password to get the aesKey. Encrypting and decrypting is only done when setting or removing the wallet password. 
The wallet write operation has a certain risk for disk failure and when setting the  password we delete all the rolling backups (as seed would be visible in unencrypted files), so that could be a critical problem if the write operation fails and causes a corrupted wallet (I read on BitcoinJ mailinglist that this happened sometimes). So the encryption/decryption operations should be done only when absolutely needed IMO.

Side note:
If the wallet is encrypted Bisq keeps the aesKey in memory to be able to sign a tx in case a taker takes an offer (which happenes non-interactive). This might be a bit of a security risk and it would be better if we do that only when absolutely needed (e.g. in case a maker has open offers).

I am not sure about your use cases for the API but passing the wallet password if required for the wallet operation seems to me the right way to deal with it.

ghubstan:
@chimp1984 ,
The full thread is tedious, so thanks for the comment.  Please do read this comment and help me clarify something I think needs to be changed?

To be direct, my question is:
Should `CoreWalletService` inject `TradeWalletService` so `CoreWalletService` can use the cached `TradeWalletService#aesKey`, instead of it's own (duplicate) copy of the `aesKey`?

Some background...  

Parts of this PR relevant to your comments are (1) an `unlockwallet` method that takes a password and caches the derived aesKey, and (2) a `lockwallet` method that clears the cached aesKey.  

Before I figured out all I needed for most encrypted wallet operations was the aesKey, I made sure kludged unlock/lock wallet methods worked for me by removing and re-setting the password (both causing a wallet re-write).  Seeing this in the thread surely caused worry for some, not to mention
> (not sure if already resolved, have not read the full thread...)

Then I changed the `unlockwallet` and `lockwallet` methods so they do not trigger wallet re-writes -- after looking more closely at bitcoin-cli and the bisq UI events that trigger `onAesKey()`.  

In the final merge, `CoreWalletService#unlockWallet()` caches the password arg derived aesKey, and `CoreWalletService#lockWallet()` clears that cached aesKey.  Right now, there are no gRPC endpoints that need that cached aesKey, but my assumption is operations such as funds withdrawal and tx-signing will need the key, and not the password used to derive it.  

Keeping that key in memory between "unlock" and "lock" commands is a security risk, but maybe less of one than typing the password in the console for every operation on an encrypted wallet?  

If you think my assumptions behind the `unlockwallet` and `lockwallet` impls are incorrect, let me know...  If not, I repeat the question:  should `CoreWalletService` inject `TradeWalletService` so `CoreWalletService` can use the cached `TradeWalletService#aesKey`, instead of it's own (duplicate) copy of the `aesKey`?

I did implement it then backed it out, waiting for an endpoint that actually needs the aesKey to work (or your recommendation, whichever came first).

PR #4215, opened 2020-04-29
Reject expired data

### LINKED ISSUE #4026: Bisq does not verify maker node is online when displaying offers aka ghost offers
### Description

After killing a Bisq node, its offers are still visible on the network for several days

#### Version

v1.2.7

### Steps to reproduce

1) Create an offer
2) kill -9 bisq

### Expected behaviour

Offer should disappear from the offer book after a few minutes

### Actual behaviour

Offer is still visible on the market after several days

chimp1984:
If a new received protectedStorageEntry is expired we do not store it
and do not broadcast.

It is unclear why we receive expired data (some are very old), but a
manipulated node might produce that and as it only removed at each
batch process running each minute to clean out expired data it still
could propagate. Is an attack vector also to flood the network with
outdated offers where the maker is likely not online.

Should fix https://github.com/bisq-network/bisq/issues/4026

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
+++ b/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
@@ -38,10 +38,9 @@ import bisq.network.p2p.storage.messages.RemoveDataMessage;
 import bisq.network.p2p.storage.messages.RemoveMailboxDataMessage;
 import bisq.network.p2p.storage.payload.CapabilityRequiringPayload;
 import bisq.network.p2p.storage.payload.DateTolerantPayload;
-import bisq.network.p2p.storage.payload.ExpirablePayload;
-import bisq.network.p2p.storage.payload.ProcessOncePersistableNetworkPayload;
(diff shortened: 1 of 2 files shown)


p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -580,6 +577,15 @@ private boolean addProtectedStorageEntry(ProtectedStorageEntry protectedStorageE
             return false;
         }
 
+        // To avoid that expired data get stored and broadcast we check early for expire date.
+        if (protectedStorageEntry.isExpired(clock)) {
+            log.warn("We received an expired protectedStorageEntry from ppper {}. getCreationTimeStamp={}, protectedStorageEntry={}",
```
ripcurlx: I guess you meant peer.
```suggestion
            log.warn("We received an expired protectedStorageEntry from peer {}. getCreationTimeStamp={}, protectedStorageEntry={}",
```

sqrrm (approved):
NACK

Logging too much at warning level.

Let's fix the logging before merging.

PR #4221, opened 2020-05-01
Fix proof-of-burn signing for password protected wallets

### LINKED ISSUE #3836: Proof of burn signing does not work if wallet password is set
### Description

Signing with proof of burn only works if wallet password is removed

#### Version

1.2.4

stejbac:
Add a missing `bsqWalletService.isEncrypted()` check and optional AES key argument to the `signMessage(..)` call in `ProofOfBurnService`, analogous to the BitcoinJ call to sign with an EC key in `MyBlindVoteListService`.

This fixes #3836 (Proof of burn signing does not work if wallet password is set).

--

Unfortunately this PR touches the core DAO packages again, although it is fairly trivial.


sqrrm (approved):
ACK

@ManfredKarrer Please review

ManfredKarrer (approved):
utACK

PR #4222, opened 2020-05-01
BondsView: Fix sorting of numeric columns

### LINKED ISSUE #3231: Table view columns should use a numeric-sort algorithm

cd2357:
The table in the BondsView uses string sorting by default. This results in unexpected behavior when sorting non-string columns.

This commit adds custom comparators to the numeric columns to address that.

Fixes #3231


sqrrm (approved):
utACK

For UI fixes it's always a good idea to include before and after screen shots to show the fix. Really helps the reviewer.

PR #4223, opened 2020-05-02
Fix QR code display of offer funding amount

### LINKED ISSUE #3666: Wrong miner fee displayed for offer maker
I've created an offer to sell BTC, I then scanned the QR code with my external wallet and the amount to send showed as less than what was displayed in the BISQ app.
It looks like this deviation was caused by a wrong miner fee being displayed.
I then somehow made this issue go away by either:
a). withdrawing all my funds from the BISQ wallet to the external wallet OR
b). selecting a custom withdrawal transaction fee OR
c). pure coincidence that the actions of a). and/or b). fixed the issue.

### LINKED ISSUE #4194: QR code not matching BTC amount when using BSQ for trading fee
When making an offer and selecting to pay the trading fee with BSQ the QR code will lead to a different BTC amount than what is printed in the transaction summary in Bisq.

ghost:
When creating an offer the small QR code displayed for funding was off by an amount that represent the mining fees.  This was due to the QR code being displayed simultaneously with the wallet calulating the fee.

We fix the issue by listening to the data model `missingCoin` value of the offer object, and updating the QR code image appropriately.

Fixes #3666
Fixes #4194


sqrrm (approved):
utACK

Seems this would first show the wrong QR-code and then update it to the correct one.

PR #4226, opened 2020-05-02
Add mention `peerbloomfilters=1` attibute

ghost:
`peerbloomfilters` needs to be set in order to synchronize with local regtest bitcoind chain

---------------

During my setup of local Bisq dev I have encountered this issue, and I believe it should be mention in this doc. It is indeed mentioned here (https://github.com/bisq-network/bisq/blob/master/docs/dao-setup.md) but in dev-setup process it is needed sooner. 

I have also mentioned on Keybase that I am not the first who encountered this during onboarding.


sqrrm (approved):
utACK

PR #4228, opened 2020-05-02
UI: In OfferBook, show tooltip on data point hover

cd2357:
A tooltip is shown when the user hovers over data points in the chart.

The tooltip text contains the coordinates of the hovered point: its price (X axis) and the cumulative offer volume at that price point (Y axis).

This helps users better read the chart data.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java
+++ b/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java
@@ -52,6 +52,7 @@ import javax.inject.Named;
 
 import com.jfoenix.controls.JFXTabPane;
 
+import javafx.scene.Cursor;
 import javafx.scene.chart.AreaChart;
 import javafx.scene.chart.NumberAxis;
 import javafx.scene.chart.XYChart;
@@ -64,10 +65,12 @@ import javafx.scene.control.TableCell;
 import javafx.scene.control.TableColumn;
 import javafx.scene.control.TableView;
 import javafx.scene.image.ImageView;
+import javafx.scene.input.MouseEvent;
 import javafx.scene.layout.AnchorPane;
 import javafx.scene.layout.HBox;
 import javafx.scene.layout.Priority;
 import javafx.scene.layout.Region;
+import javafx.scene.layout.StackPane;
 import javafx.scene.layout.VBox;
 
 import javafx.geometry.Insets;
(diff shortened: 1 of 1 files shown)


sqrrm (changes requested):
Again, please add before and after screen shots of UI changes for easier review.

cd2357:
Adding screenshots of the offerbook:

Before (when hovering over graph datapoints, nothing happens) :

Screenshot from 2020-05-06 22-08-19

After (when hovering over datapoints, a tooltip is shown with the data point coordinates) :

Screenshot from 2020-05-06 22-11-44

desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java
```diff
@@ -421,9 +427,47 @@ private void updateChartData() {
 
         seriesBuy.getData().addAll(model.getBuyData());
         seriesSell.getData().addAll(model.getSellData());
+
+        // Before adding the points to the chart, set a custom node used for showing a label on hover
```
ghost, line 431: Instead of a comment, I would prefer the named method (with same information on the name). Comments tend to age and becoming inconsistent with the code. Best documentation is the code itself.

desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java
```diff
@@ -421,9 +427,47 @@ private void updateChartData() {
...
+        // Add both series to the chart
         areaChart.getData().addAll(List.of(seriesBuy, seriesSell));
     }
 
+    /**
+     * A node which displays a label on hover, but is otherwise empty
+     *
+     * Based on https://gist.github.com/jewelsea/1422628
+     */
+    protected class HoveredLabelNode extends StackPane {
```
ghost, line 449: I would move this Node to the separate file and pass data via constructor (the price ->`xAxis.getTickLabelFormatter().toString(price)`. 

This file is quite long already and this is separable responsibility.

ghost:
Just small UI glitch, nodes on the edges will not be fully visible: 
image

ghost:
Concept ACK

sqrrm:
No strong opinion on the HoverableNode to a separate class, but I think fixing the edge condition should be done. A small thing but it makes a big difference in feel to have visual issues like that.

Issue #1387, opened 2018-02-22
Add memo field for withdraw BTC funds


ManfredKarrer:
At the "send funds" screen we should add a new input field for a user message/memo so part transfers can be easier assigned.
In the "transactions" screen we should add the truncated memo text to the "Withdrawn from wallet" text. The tooltip should display the full text. Maybe "Withdrawn from wallet" can be replaced by a shorter text...

ManfredKarrer:
Still relevant

niyid:
Anybody taken this yet?

cd2357:
Solved by @petrhejna in #4229

PR #4229, opened 2020-05-03
Add memo field to withdrawal transaction

ghost:
- "Memo" field is modeled as property of the new object Transaction which is stored in persitant storage.
- Transaction object is modeled in a way that allows extension in the furure for more persisted attributes.

Will solve #1387

---------------

~This is a Draft intended for early review because it is my first contribution during onboarding.~

Preview:

image


sqrrm:
I notice you have added a new storage class for transactions. There is already the bitcoinj Transaction which stores the transactions and has a memo field. What was the reasons for not using that? I suppose those transactions would be lost in an SPV resync. Just feels like quite a bit of extra work to store the memo field separately like this.

ghost:
Thank you very much for direction. I have updated the code in  this PR: 

For now, I'll keep commits (to allow anyone to see the hisstory in this PR) but I will of course squash it before the merge.

desktop/src/main/java/bisq/desktop/main/funds/withdrawal/WithdrawalView.java
```diff
@@ -164,7 +163,8 @@ private WithdrawalView(BtcWalletService walletService,
                            @Named(FormattingUtils.BTC_FORMATTER_KEY) CoinFormatter formatter,
                            Preferences preferences,
                            BtcAddressValidator btcAddressValidator,
-                           WalletPasswordWindow walletPasswordWindow) {
+                           WalletPasswordWindow walletPasswordWindow
+    ) {
```
sqrrm: This is not customary. Not sure if it's in the style guide, but better keep it in line with other code and not line break before the parenthesis.

sqrrm (commented):
Tested it. The popup is not following dark mode scheme I think.
image

desktop/src/main/java/bisq/desktop/main/funds/withdrawal/WithdrawalView.java
ghost: :+1: this looks like some forgotten change after reverting my previous changes.

Maybe question/suggestion - is there any good automated style check (ideally with fixer) to solve issues like this to save some time? I am not really experienced in Java ecosystem, but I believe there is something like https://eslint.org/ but for Java.

ghost:
@sqrrm About the dark mode tooltip. I am not sure if its a bug in my PR. As I understand it, I am using standard `AutoTooltipLabel` component. This component is used for other fields as well. 

From this PR: 
image

Am I missing something? If not, I suggest to create a special Issue for this as bug in the `AutoTooltipLabel` in DarkMode.

sqrrm (approved):
utACK

I agree, the tooltip you use should be fixed as a separate issue.

PR #4230, opened 2020-05-03
More readable tables and lists: Alternate row background colors

cd2357:
Alternate row background colors are now available, regardless of theme.

A few CSS selectors affecting table and list row coloring were moved the dark theme stylesheet (theme-dark.css) to the common stylesheet (bisq.css).

These selectors are theme-independent, since they re-use variables defined in each theme stylesheet (like -bs-background-color). Therefore a more appropriate place for them is in the common stylesheet.

This move means that alternating row background colors are now available for all tables and lists, in all views, in both dark and light themes.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/bisq.css
+++ b/desktop/src/main/java/bisq/desktop/bisq.css
@@ -248,7 +248,9 @@
 .list-view .list-cell:even {
     -fx-background-color: derive(-bs-background-color, 5%);
 }
+.list-view .list-cell:hover,
 .list-view .list-cell:selected,
+.table-view .table-cell:hover,
 .table-view .table-cell:selected {
     -fx-background: -fx-accent;
     -fx-background-color: -fx-selection-bar;
@@ -897,6 +899,7 @@ textfield */
     -fx-background-color: derive(-bs-background-color,-5%);
     -fx-border-color: derive(-bs-background-color,-5%);
 }
+.table-view .table-row-cell:hover .table-cell,
 .table-view .table-row-cell:selected .table-cell {
     -fx-background: -fx-accent;
     -fx-background-color: -fx-selection-bar;
--- a/desktop/src/main/java/bisq/desktop/bisq.css
+++ b/desktop/src/main/java/bisq/desktop/bisq.css
@@ -237,6 +237,28 @@
     -fx-background-color: -fx-selection-bar;
 }
 
+
+/* list view */
+.list-view .list-cell {
+    -fx-background-color: -bs-background-color;
+}
+.list-view .list-cell:odd {
+    -fx-background-color: derive(-bs-background-color, -5%);
+}
+.list-view .list-cell:even {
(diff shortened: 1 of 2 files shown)


sqrrm (changes requested):
Please add screenshots.

cd2357:
Here are some screenshots.

Before:

Screenshot from 2020-05-06 22-03-27

Screenshot from 2020-05-06 22-04-27

After:

Screenshot from 2020-05-06 21-56-48

Screenshot from 2020-05-06 21-57-07

ghost:
For "not full tables" it look in my opinion little bit akward 

image

Maybe not to use stripes for empty rows?

desktop/src/main/java/bisq/desktop/bisq.css
```diff
@@ -867,6 +891,32 @@ textfield */
  * Table                                                                       *
  *                                                                             *
  ******************************************************************************/
+.table-view .table-row-cell:even .table-cell {
+    -fx-background-color: derive(-bs-background-color, 5%);
+    -fx-border-color: derive(-bs-background-color,5%);
```
ghost, line 896: Maybe nitpicky, but there is whitespace inconsistency :)

ghost:
I would personally make it less "sharp" by lowering `+-5%` to just `+-3%`

Here is how it looks like with `3`. (Note: I have also found a UI glitch that lines have different size (probably only problem when they are empty.)

image

ghost:
Concept ACK

cd2357:
> For "not full tables" it look in my opinion little bit akward
> 
> Maybe not to use stripes for empty rows?

Would potentially make sense, but it's not related to my change.

To achieve that, table model logic would have to be changed to remove specific CSS classes from rows if they are empty. That' most likely custom java logic that has to be done for every table.

This PR is only about bringing a dark-theme behavior (showing odd/even rows as having different colors) also into the light-theme.

So I would say, to keep PRs conceptually separated, that change should be covered in a different PR (especially since it affects both dark and light themes -- this PR only fixes some CSS tags in the light theme).

cd2357:
> I would personally make it less "sharp" by lowering `+-5%` to just `+-3%`

Yes, would be worth exploring.

But perhaps in a separate PR (see comment above).

I tried to keep this PR specifically about bringing a dark-theme-only behavior (alterate row colors) to both dark and light themes.

desktop/src/main/java/bisq/desktop/bisq.css
sqrrm: I agree, just run some auto formatter on the code to get spacing uniform. It's a small thing but someone will do it later and there will be changes in an unrelated PR.

Apart from that I'm good to merge this PR.

sqrrm (approved):
utACK

ripcurlx:
In that case it would have been great to get @pedromvpg's feedback before merging, as he did this by intent as far as I remember. Anyways I'll leave it in for this release and we might make an adaption in the next release.

PR #2772, opened 2019-04-23
[WIP] Use bisq's bitcoinj 0.15.1

oscarguindzberg:

Issue #1139, opened 2017-12-31
Support for Bitcoin Bech32 SegWit addresses


rex4539:
https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki

mrosseel:
bisq depends on bitcoinj for its wallet/bitcoin transactions, and bitcoinj has no segwit/bech32 support as of today. If it's in bitcoinj, we will support it asap!

sneurlax:
See https://github.com/bitcoinj/bitcoinj/pull/1441 ... so we will get BIP49 (P2SH-P2WKPH, 3...) addresses before BIP173 (P2WKPH, bc1...) addresses

ManfredKarrer:
I will close for now to keep GH organized with issues which are subject to be implemented in a reasonable time scope. I mark it as deferred.

sneurlax:
See https://github.com/sipa/bech32/pull/40#issuecomment-365921271

ghost:
@mrosseel: segwit support has been merged into bitcoinj master since june 2017: https://github.com/bitcoinj/bitcoinj/pull/1334

ManfredKarrer:
Thats not full segwit support. See BitcoinJ mailing list....

ghost:
@mrosseel: thanks for clarifying; I wasn't aware that the current bitcoinj implementation is not complete, but in my understanding the existing code is enough to generate compatible segwit addresses that start with `3` as can be seen in the unit test at https://github.com/bitcoinj/bitcoinj/blob/segwit/core/src/test/java/org/bitcoinj/core/AddressTest.java#L178 (outdated `segwit` branch). I don't dominate the topic so I apologize in case if my assumptions are wrong... [SamouraiWallet](https://github.com/Samourai-Wallet/samourai-wallet-android) apparently relies on bitcoinj for segwit and they provide segwit support in their android wallet: https://github.com/Samourai-Wallet/samourai-wallet-android

ManfredKarrer:
SamouraiWallet uses its own fork. I have not followed the details but I am pretty sure that BitcoinJ at the current state is not fully supporting SegWit.

ghost:
[...]

Issue #1077, opened 2017-12-09
Improve fee estimation


ManfredKarrer:
As we have seen in the recent spam attack and blockchain congestion the fee estimation is difficult to get right.
We use an API from https://bitcoinfees.earn.com but it was usually overpaid quite a lot. So we reduced a while back to get more to an optimum. That worked pretty well until recently when in 1 or 2 days the mempool grew from 20k to 200k unconfirmed transactions. That rapid rise was not handled well by the current model.
One reason was the average function. We used the average of the last hour, which kicked us always to the lower end and with rising fees the chance to get those txs confirmed decreased.

We changed now to more aggressive and expensive fees (lower max blocks param and only last 10 min taken for average) but I think that will lead to overpay again. Why we should pay the Bitmain and co more as absolutely needed? They do a great disservice for Bitcoin since long and it really hurts to pay them extra. So we should find a way to get back to the min. fee required to get the tx confirmed in < 3 blocks.

We could manually adjust the parameters and once we observe mempool getting filled up we have to be more aggressive again. But that manual adjustment is not a good solution of course. 
So we should find a way how to influence the parameters based on those observations we would do manually, namely the size of the mempool as well the dynamics (fast rise or more organic behavior).

First approach could be to simply request the mempool size  and use that as influence for adjusting our parameters. 

Anyone who is expert in that field would be very welcome to give input. Alternative estimations services would be great as well as ear.com does not work well from my experience (e.g. I often observed the they suggest a high fee and I got txs with much less confirmed immediately).  

io.bisq.provider.fee.providers.BtcFeesProvider is the class which implements our estimation.
It simply takes the maxDelay as param and adds the average function to not be too much exposed to spikes.

ripcurlx:
Besides optimizing the fee adjustments, we should also add the information within the trading process of the expected confirmation time to reduce the number of disputes opened.

montvid:
I use https://jochen-hoenicke.de/queue/#2h I enter manually 300 200 140 or 100 depending on the chart - works for me. Now it would be 300.

mrosseel:
@montvid based on what do you take 300? You just take the 'second most expensive' fee interval from that page?

montvid:
500 - 300 tx 300 - 15000 tx 200 - 25000 tx so i choose the best for me. If I want to wait i choose 200 if i want it quite fast i choose 300 if i want instant i can do earn.com or 500.

ManfredKarrer:
[...]

PR #4231, opened 2020-05-03
User can set custom tx fee to override FeeService estimated fee

cd2357:
The user preferences already contained an option to set a custom withdrawal tx fee.

With this change, the scope of that field has been extended to cover all callers of the FeeService, which means every place where a tx fee is needed. If the preferences field is unset, the FeeService behavior falls back to the existing one (uses fee estimates from the earn.com API).

In summary, two things were done in this commit:
1. the FeeService logic was extended (if user set a custom fee, uses that instead of earn.com API estimate; otherwise queries fee API endpoint for estimations, as usual)
2. relevant fields, labels and methods have been renamed to reflect the change in scope (customWithdrawalTxFee -> customTxFee)

I tried to keep the change as non-invasive as possible. From what I can tell, the change is minimal (consisting mostly of renamed fields and methods).

The protobuf change is backward compatible with previously serialized user preference protobufs, because the "custom fee" fields already existed before. Its only the metadata for addressing the fields that has changed, which does not end up in the serialization. Only the field values are serialized, which remained the same.

Several local tests were successful:
- loading previously stored preferences (persisted before change, loaded after change)
- persisting preferences across Bisq restarts
- changing custom fee preference + navigating around UI + opening new trades

One known minor issue remains open, which is that of adjusting translations for other languages ("Withdrawal transaction fee" -> "Transaction fee"). Currently I have only changed the EN version.

Please let me know if you see any issues with the general direction or with the implementation.

Workaround for https://github.com/bisq-network/projects/issues/27 until a better fee estimation service is used.


cbeams (changes requested):
NACK.

We allow users to set a custom withdrawal transaction fee rate because it's their problem if they set it too low and end up queued in the mempool for an extended period.

We do not allow users to set a custom trading transaction fee rate because it creates a problem for their trading counterparty if they set it too low. The whole trade will get held up in such a case, and that's unfair to the counterparty who used the default fee or chose a more reasonable custom value.

Nit: for future PRs, please follow the commit guidelines at https://github.com/bisq-network/style/issues/9.

sqrrm:
I'm not sure about extending the custom fee all over the place, but perhaps it could be used a bit more and only forcing higher fees for the most sensitive parts of the trade protocol. It's not obvious that's a good idea though. Any stuck tx adds to the support load, even if it's due to a user setting the custom fee too low.

huey735:
@cbeams I understand that your argument applies to offer-taking, but what about offer-making and proposals?

cbeams:
> @cbeams I understand that your argument applies to offer-taking, but what about offer-making [...]?

Offer making is a different story as it's theoretically only the maker's problem, and they're not externalizing anything on their counterparty, but as @sqrrm mentions, in practice it's probably still going to land in support. Supporting RBF (replace-by-fee) in Bisq would help solve this as it would allow makers to self-serve solving the problem. And there's nothing we can do in support to help them anyway. We would have to be very clear that in the UI that changing the recommended transaction fee is to be done at the users' risk, that there's nothing we can do to help their transaction if it's stuck, etc. Would need to think about how offer publication happens as well. I think right now, the offer gets published to the offer book and becomes visible even before the first confirmation occurs. We wouldn't want that to be the case if this change were made, because again an unwitting taker could get hurt for no fault of their own if the maker fee tx gets stuck.

> what about [...] proposals?

This would be less risky to allow it here. It's fully the problem of the proposer if the tx gets stuck, and we could be clear that there is no recourse for this, just do it at your own risk.

---

With the above said, I don't see this as a priority now that we've solved the problem of too-high fee rate recommendations out of earn.com with @cd2357's PR at https://github.com/bisq-network/projects/issues/27.

As far as I'm aware, users haven't been complaining about the tx fee rate in Bisq except during a couple incidents where rates were way too high and got stuck there. This happened in 2017, just recently, and IIRC there was one other time too. If the new mempool.space service proves reliable in producing reasonable estimates, my preference would be to leave this area alone for now, and not add more complexity and risk leading to potentially more support incidents and poor user experiences.

cd2357:
> As far as I'm aware, users haven't been complaining about the tx fee rate in Bisq except during a couple incidents where rates were way too high and got stuck there. This happened in 2017, just recently, and IIRC there was one other time too.

That is a rather charitable interpretation I'd say, considering that
* people only realize what mining fees they pay if they read the fine print in the trade popup
* of those who do, few know what an acceptable mining fee is at the moment of the trade
* of those who know, few get upset enough to want to report it
* of those who want to report it, few know exactly how to do it
* of those, even fewer have the patience and willingness to setup keybase, or setup a github account + create an issue

So there are many stages to go through, and only the last group are "visible" to the Bisq team as having made a complaint.

Its also a question of which metric to use when deciding how to prioritize limited dev resources:
* minimize amount of BTC unnecessarily "lost" by traders in every trade?
  * then prioritize fee estimation (done), bech32 support (#1139) + upgrade BitcoinJ (#2772), possibly RBF (https://github.com/bisq-network/bisq/issues/1077#issuecomment-357526156), etc
* minimize amount of open tickets and complaints?
  * then prioritize those things which most amount of people complain about

> With the above said, I don't see this as a priority now that we've solved the problem of too-high fee rate recommendations out of earn.com with @cd2357's PR at bisq-network/projects#27.

I agree with @cbeams. From a *"should Bisq merge this change"* perspective, this PR is probably not necessary anymore, now that the fee estimation is more accurate. So I'll just go ahead and close the PR.

However, looking from a *"should Bisq eventually support this"* sort of perspective, I'd say yes. After all, Bisq is open source and if the protocol allows such a change, sooner or later someone will patch their local client and do it, cause it would be in their best interest to save some sats on mining fees.

PR #4232, opened 2020-05-04
Rename "Rocket Chat Admin" role to "Keybase Admin"

cbeams:
Per https://github.com/bisq-network/roles/issues/79#issuecomment-618483481

Note that the ROCKET_CHAT_ADMIN enum label has not been renamed for
the backward compatibility reasons discussed in BondedRoleType Javadoc.

See also bisq-network/roles#60.


sqrrm (approved):
utACK

There is no need to change all the displayStrings language files, just changing the main one is enough.

sqrrm:
I think renaming the enum shouldn't be a problem either.

@ManfredKarrer this touch DAO code, just a comment, but still.

ManfredKarrer (approved):
utACK

ManfredKarrer:
> I think renaming the enum shouldn't be a problem either.

I assume as well, but not sure if testing effort is worth it.

PR #4233, opened 2020-05-04
Reduce initial request size

### LINKED ISSUE #2547: Startup problem on slow/degraded networks (the "3/4" problem)
Possibly related to: #2474, #2327. Uses #2278.

This is a slightly edited re-post of the issue reported at:
https://bisq.community/t/issue-bisq-startup-problem-on-slow-networks-3-4-problem/7233

**Summary**: 
On slow/congested networks, Bisq never progresses past step 3/4 of the startup sequence, due to a timeout occurring in the initial handshake to seed nodes.

**Details**: 
One of the compulsory steps in the Bisq client startup sequence is a special handshake with a seed node, consisting in an exchange of large data packets in the 1.5 MB size range (PreliminaryGetDataRequest in the debug log). On slow networks, specifically when the network *output* channel is degraded, this handshake fails systematically with a timeout error, resulting in an inability to start up (the "3/4" problem). 

Bisq versions tested: 0.8.1, 0.9.3, 0.9.4, 0.9.5
Test platform: Tails 3.10, 3.11, 3.12 (Debian 9); Compatibility workarounds from #2278 

To reproduce the bug, use a working Bisq installation and limit the network output bandwidth to a small value (100 kbps in my example) using the Linux TBF qdisc. 

**Steps to reproduce**:
- start with a correctly installed and tested operational copy of Bisq
- close the application
- in a terminal, execute (replace eth0 with name of default route interface):
`sudo tc qdisc add dev eth0 root tbf rate 100kbit burst 1540 latency 50ms`
- start the application as usual

**Expected result**:
Bisq should complete the startup sequence and operate (with some lag due to network bandwidth limitation). 

**Actual result**:
Bisq never progresses past step 3/4 of startup sequence, resulting in complete no-op.

**Log fragment**
```
Feb-26 12:16:38.651 [JavaFX Application Thread] INFO b.n.p.p.g.RequestDataHandler: We send a PreliminaryGetDataRequest to peer fl3mmribyxgrv63c.onion:8000.
Feb-26 12:17:38.555 [JavaFX Application Thread] WARN b.n.p.p.g.RequestDataManager: We have started already a requestDataHandshake to peer. nodeAddress=fl3mmribyxgrv63c.onion:8000
Feb-26 12:17:38.652 [JavaFX Application Thread] INFO b.n.p.p.g.RequestDataHandler: A timeout occurred at sending getDataRequest:PreliminaryGetDataRequest(supportedCapabilities=[0, 1, 2, 7]) on nodeAddress:fl3mmribyxgrv63c.onion:8000
Feb-26 12:18:44.754 [InputHandler-fl3mmribyxgrv63c.onion:8000] INFO b.n.p.n.Connection$SharedModel: Shut down caused by exception java.net.SocketTimeoutException: Read timed out on connection=Connection{peerAddress=Optional[fl3mmribyxgrv63c.onion:8000], peerType=SEED_NODE, uid=‘1c112047-32de-4631-93e9-5c6a175a30ee’}
Feb-26 12:18:44.774 [JavaFX Application Thread] INFO b.n.p.p.PeerManager: onDisconnect called: nodeAddress=Optional[fl3mmribyxgrv63c.onion:8000], closeConnectionReason=CloseConnectionReason{sendCloseMessage=false, isIntended=false} SOCKET_TIMEOUT
Feb-26 12:18:44.822 [JavaFX Application Thread] INFO b.n.p.p.p.PeerExchangeManager: onDisconnect closeConnectionReason=CloseConnectionReason{sendCloseMessage=false, isIntended=false} SOCKET_TIMEOUT, nodeAddressOpt=Optional[fl3mmribyxgrv63c.onion:8000]
```

### LINKED ISSUE #2549: Unhandled unrecoverable program state following failed handshake with seed nodes
Related to: #2547, possibly #2474, #2327 

This is a slightly edited re-post of the issue reported at:
https://bisq.community/t/issue-bisq-startup-problem-on-slow-networks-3-4-problem/7233/3

**Details:**
After a network congestion/degradation during startup (issue #2547), Bisq remains in a dead-end state that doesn't recover even after the network performance is restored. The only way to recover is restarting the program. 
This condition is not properly handled nor reported in the UI, causing confusion to unaware users. 

**Steps to reproduce**
- same as issue #2547, start with a working installation
- close application
- degrade network: `sudo tc qdisc add dev eth0 root tbf rate 100kbit burst 1540 latency 50ms`
- start application
- wait 3 hours, per issue #2547 there should be no progress past step 3/4
- remove network degradation: `sudo tc qdisc del ...`
- wait indefinitely, there will still be no progress past step 3/4, even 12 hours later

**Expected result**
The program should attempt to recover seamlessly from the degraded state when the network conditions allow it.

**Alternative expected result**
The program should detect the dead-end state and report it in the UI along with some information about its probable cause, then exit or restart in order to attempt to recover from this state. 

**Actual result**
The program keeps running, displays a progress indicator and shows the offers, trades, wallet contents, exchange rates etc, creating a false appearance of almost-operational and a false expectation of further progress, when in reality it is in an unrecoverable no-op state.

### LINKED ISSUE #3763: Application won't progress past Step 3/4 Hidden Service published
### Description
Application is stuck on step 3/4.

#### Version

1.2.4 Mac Os

### Screenshots
Screen Shot 2019-12-08 at 12 12 56 PM

#### Device or machine
Mac Os Mojave 10.14.6

freimair:
Fixes https://github.com/bisq-network/bisq/issues/2547, fixes https://github.com/bisq-network/bisq/issues/2549, fixes https://github.com/bisq-network/bisq/issues/3763, part of https://github.com/bisq-network/projects/issues/25

This PR reduces the initial request size to a O(1). By numbers, the number of historical keys to be sent to the seednodes is 1 now. With 1.3.4, that is 1 vs. 100k+ object keys and 2x1kB vs. 2x2500kB at startup.

## Details
This is a big step towards a more pleasant user experience and due to changes to parts of the data store implementation it is also high risk.
I see the split data store implementations as a temporary solution until I can proceed to https://github.com/bisq-network/projects/issues/29. Hence, even as there could be more tests, there aren't.

## Testing efforts
I created a number of test classes handling the basics of the migration and update scenarios, everyday use cases plus a few special cases. Additionally, there is a shell script running a kind of manual integration test which helped in finding more issues. The "manual" test takes care of setting up a testing testnet and through the steps upgrades pieces of software as well. Take a look at it if you so please. The results of the test have to manually observed although there is a `result.log` once finished which lists all the request and response sizes. (please not that some logs are put there twice because that is simply how it is).

Please give this a thorough test. If there are any questions, please feel free to contact me on keybase.

## EDIT:

as always when trying to test disk IO, travis complains and fails the tests. Please try the PR locally anyways. I will deactivate the tests for travis before the PR is merged.


p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
```diff
@@ -0,0 +1,190 @@
...
+
+    @Override
+    protected PersistableNetworkPayload putIfAbsent(P2PDataStorage.ByteArray hash,
+                                                    PersistableNetworkPayload payload) {
+        // make sure we do not add data that we already have (in a bin of historical data)
+        if (getMap().containsKey(hash))
+            return null;
+
+        PersistableNetworkPayload previous = store.getMap().putIfAbsent(hash, payload);
+        persist();
```
sqrrm, line 75: Seems excessive to persist if nothing was written, but I notice that's what MapStoreService does.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
```diff
@@ -0,0 +1,190 @@
...
+        return previous;
+    }
+
+
+    @Override
+    public Map<P2PDataStorage.ByteArray, PersistableNetworkPayload> getMap() {
+        HashMap<P2PDataStorage.ByteArray, PersistableNetworkPayload> result = new HashMap<>();
+        result.putAll(store.getMap());
+        history.forEach((s, store) -> result.putAll(store.getMap()));
```
sqrrm, line 84: getMap() seems to be called a fair bit and it's recreating the map every time. Not sure how expensive that is, but perhaps some caching should be considered.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
```diff
@@ -0,0 +1,190 @@
...
+    public Map<P2PDataStorage.ByteArray, PersistableNetworkPayload> getMap(String filter) {
+        HashMap<P2PDataStorage.ByteArray, PersistableNetworkPayload> result = new HashMap<>();
+        result.putAll(store.getMap());
+
+        // TODO do a proper language, possibly classes
+        if (filter.startsWith("since ")) {
+            filter = filter.replace("since ", "");
+            if (!filter.equals(Version.VERSION)) {
+                String finalFilter = filter;
+                history.entrySet().stream().filter(entry -> Integer.valueOf(entry.getKey().replace(".", "")) > Integer.valueOf(finalFilter.replace(".", ""))).forEach(entry -> result.putAll(entry.getValue().getMap()));
```
sqrrm: Try to keep lines < 120 chars

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        return result;
+    }
+
+    @Override
+    protected void readFromResources(String postFix) {
+        // check Version.VERSION and see if we have latest r/o data store file in working directory
+        if (!new File(absolutePathOfStorageDir + File.separator + getFileName() + "_" + Version.VERSION).exists())
+            makeFileFromResourceFile(postFix); // if we have the latest file, we are good, else do stuff // TODO are we?
```
sqrrm: This checks if the latest file exists...

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
```diff
@@ -0,0 +1,190 @@
...
+        else {
+            // load stores/storage
+            File dbDir = new File(absolutePathOfStorageDir);
+            List<File> resourceFiles = Arrays.asList(dbDir.list((dir, name) -> name.startsWith(getFileName() + "_"))).stream().map(s -> new File(s)).collect(Collectors.toList());
```
sqrrm: ... if it does exist we make a list of all the files we have.

Also, line breaking streams make them a lot easier to follow.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
```diff
@@ -0,0 +1,190 @@
...
+
+            history = new HashMap<>();
+            store = readStore(getFileName());
+            resourceFiles.forEach(file -> {
+                SplitStore tmp = readStore(file.getName().replace(postFix, ""));
+                history.put(file.getName().replace(postFix, "").replace(getFileName(), "").replace("_", ""), tmp);
```
sqrrm: Then we add them to our history. What is the key, not easy to follow. Do we depend on having all the previous stores since we're only checking for the last one?

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
```diff
@@ -0,0 +1,190 @@
...
+            log.warn("make dir failed.\ndbDir=" + dbDir.getAbsolutePath());
+
+        // check resources for files
+        List<String> versions = new ArrayList<>();
+        versions.add(Version.VERSION);
+        versions.addAll(Version.history);
+        List<String> resourceFiles = versions.stream().map(s -> getFileName() + "_" + s + postFix).collect(Collectors.toList());
+
+        // if not, copy and split
+        resourceFiles.forEach(file -> {
```
sqrrm: `file` is a filename string...

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
```diff
@@ -0,0 +1,190 @@
...
+            final File destinationFile = new File(Paths.get(absolutePathOfStorageDir, file.replace(postFix, "")).toString());
+            final String resourceFileName = file;
```
sqrrm: ...making this var a bit excessive.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
```diff
@@ -0,0 +1,190 @@
...
+            }
+        });
+
+        // split
+        // - get all
+        history = new HashMap<>();
+        store = readStore(getFileName());
+        resourceFiles.forEach(file -> {
+            SplitStore tmp = readStore(file.replace(postFix, ""));
+            history.put(file.replace(postFix, "").replace(getFileName(), "").replace("_", ""), tmp);
```
sqrrm: A helper method to do these string manipulations would help readability.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
```diff
@@ -0,0 +1,190 @@
...
+            // - subtract all that is in resource files
+            store.getMap().keySet().removeAll(tmp.getMap().keySet());
```
sqrrm: With this change I think we're making the resource files the final arbiter of what is the distributed data. Is that what we want? Even for full nodes that are always on.

From what I understood so far, nodes will keep getting data from the network. With new versions all new data since the last version will be stored in a new `resource_x.x.x_MAINNET` file and all nodes will restore this data from resources. It looks like it will just throw away its own local data on new versions though, that seems incorrect.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -205,24 +213,47 @@ public GetUpdatedDataRequest buildGetUpdatedDataRequest(NodeAddress senderNodeAd
...
+        System.arraycopy(Version.VERSION.getBytes(), 0, result, 0, Version.VERSION.length());
+        return result;
+    }
+
+    private String containsSpecialKey(Set<P2PDataStorage.ByteArray> collection) {
+        Optional<String> result = collection.stream().map(byteArray -> new String(byteArray.bytes).trim()).filter(s -> s.matches("^[0-9]\\.[0-9]\\.[0-9]$")).findFirst();
+        if (result.isPresent())
+            return result.get();
+        else
+            return "";
```
sqrrm: Easier to read would be `        return result.orElse("");`
or even
```
        return collection.stream()
                .map(byteArray -> new String(byteArray.bytes).trim())
                .filter(s -> s.matches("^[0-9]\\.[0-9]\\.[0-9]$"))
                .findFirst()
                .orElse("");
```

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -205,24 +213,47 @@ public GetUpdatedDataRequest buildGetUpdatedDataRequest(NodeAddress senderNodeAd
...
 
+    private byte[] getSpecialKey() {
+        byte[] result = new byte[20];
+        Arrays.fill(result, (byte) 0);
+        System.arraycopy(Version.VERSION.getBytes(), 0, result, 0, Version.VERSION.length());
+        return result;
+    }
+
+    private String containsSpecialKey(Set<P2PDataStorage.ByteArray> collection) {
+        Optional<String> result = collection.stream().map(byteArray -> new String(byteArray.bytes).trim()).filter(s -> s.matches("^[0-9]\\.[0-9]\\.[0-9]$")).findFirst();
```
sqrrm: Is String(byte[]) safe enough? I suspect so but I'm not sure.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -205,24 +213,47 @@ public GetUpdatedDataRequest buildGetUpdatedDataRequest(NodeAddress senderNodeAd
...
+
         // We collect the keys of the PersistableNetworkPayload items so we exclude them in our request.
         // PersistedStoragePayload items don't get removed, so we don't have an issue with the case that
         // an object gets removed in between PreliminaryGetDataRequest and the GetUpdatedDataRequest and we would
         // miss that event if we do not load the full set or use some delta handling.
-        Set<byte[]> excludedKeys = this.appendOnlyDataStoreService.getMap().keySet().stream()
-                .map(e -> e.bytes)
-                .collect(Collectors.toSet());
+        Set<byte[]> excludedKeys;
+        if (seedNodeRepository != null && seedNodeRepository.isSeedNode(networkNode.getNodeAddress()))
```
sqrrm: Braces around multi-line blocks really helps readability in my opinion.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -268,9 +299,18 @@ public GetDataResponse buildGetDataResponse(
...
                 P2PDataStorage.ByteArray.convertBytesSetToByteArraySet(getDataRequest.getExcludedKeys());
 
+        String specialKey = containsSpecialKey(excludedKeysAsByteArray);
+
+        Map<ByteArray, PersistableNetworkPayload> tmp;
+        if ("".equals(specialKey))
+            tmp = this.appendOnlyDataStoreService.getMap();
+        else {
+            tmp = this.appendOnlyDataStoreService.getMap("since " + specialKey);
+        }
```
sqrrm: Is this the separation of old and new requests for data? More comments would be helpful for understanding.

Variable name `tmp` isn't really descriptive, better naming would help.

sqrrm (commented):
I've read through the core parts of the PR. For now just adding comments to gain a better understanding. This PR makes me concerned that we might lose data during upgrades or otherwise, but I don't have a full understanding yet.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
freimair: agreed. However, it is not trivial to find whether there has been something added to the map or not since the nested `put(...)` only returns the old value.

I believe that is why MapStoreService does this like so (and that might be one source of our IO issues).

I decided to stick with the "proven" way of doing things until we can address the whole disk IO issue in a more general way https://github.com/bisq-network/projects/issues/29

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
freimair: I thought about that. The only thing that can be cached is loading the historical data stores into the result. And since these are read-only anyways, there is nothing that changes. So why not cache it?

Well, the maps of the stores are in memory as well. So the data is already there and ready to be transferred. The only redirection that happens when we do it like this is that we iterate through all stores and add the data to the result. Compare that to a "cached" map of historical data, the data has to be transferred as well. The only thing we gain by caching the data is we need more RAM while saving on a few instructions to iterate through history.

IMO The small computational overhead does outweigh adding the complexity of another caching layer that needs to be maintained.

Anyways, this piece of code is going to be deprecated by https://github.com/bisq-network/projects/issues/29 anyways. Then is the time IMO to really dial down performance.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
freimair: > Do we depend on having all the previous stores since we're only checking for the last one?

exactly. And you may have spotted something here.

When we have the last one, we already did the transformation to the new split store system. Hence, I assumed that from now on we already have the historical stores. On a Bisq update, only the newest store is checked and if it isn't there, we need to copy it from resources and un-duplicate things in our live data store.

But you are correct, that might need a bit more effort to make the whole system more robust.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
freimair: > With this change I think we're making the resource files the final arbiter of what is the distributed data. Is that what we want? Even for full nodes that are always on.

This seems to be the only way of introducing "checkpoints" so that we do not have to query all the data.

> It looks like it will just throw away its own local data on new versions though, that seems incorrect.

It does not throw away any data. It only de-duplicates the data. If we have a data store "incoming" from resources, we remove the data that is contained in the resource store from our live data store.

An example:
Given _ResourceStore={1,3}_ and _LiveStore(t)={1,2,3,4,5}_ and we enter the copy and split business logic, the outcome would be having the _ResourceStore_ being copied to the working directory and the new _LiveStore(t+1)=LiveStore(t) \ ResourceStore = {2,4,5}_. Together, they hold {1,2,3,4,5}.

Bringing it all together:
In terms of initial data request size, we can now ask for all data except _ResourceStore_ simply by telling other that we have the "ResourceStore_1.3.4" -> sending {"1.3.4"} instead of {1,3}. And because in real live, the data stores count >100k objects, we save a whole lot of data. Additionally, we actually ask for all data "since" (instead of "except") and thus keep the request object count at O(1) in the future.

Again, and hopefully, the file operations will soon be superseded by https://github.com/bisq-network/projects/issues/29.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
freimair: > Is this the separation of old and new requests for data?

Correct. If there is a special key incoming we prefilter the data, if not, we do what we did before.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
freimair: at least I haven't found an issue yet

freimair:
just rebased to master and folded in the first batch of feedback. (Thanks @sqrrm for your efforts)

Build is failing because JDK12 does not support removing final-modifiers through reflections. If I remove the tests that rely on that everything would be green. Not sure how to handle that. And no, there is no other way to test this except loosening something that clearly is final to being not final in the first place.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
freimair: just added some additional logic to address this issue.

I could not stay true to the rest of the coding style and did my own design. And by doing that, I saved on 30 loc.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
sqrrm: I ask since it might act differently on different platforms. I think it would never be an issue for version values, but perhaps other values could cause problems, deterministic is nice.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
```diff
@@ -121,39 +119,28 @@ private int parseSpecialKey(String specialKey) {
...
+        store = readStore(getFileName());
+
+        // create file list of files we should have by now
+        List<String> versions = new ArrayList<>();
+        versions.add(Version.VERSION);
+        versions.addAll(Version.history);
+
+        // go through the list one by one
+        versions.forEach(version -> {
+            String filename = getFileName() + "_" + version + postFix; // postFix has a preceding "_"
```
sqrrm: There aren't any postfixes in the db dir, should the postFix be added here?

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
```diff
@@ -121,39 +119,28 @@ private int parseSpecialKey(String specialKey) {
...
+            if (new File(absolutePathOfStorageDir, filename).exists()) {
+                // if it is there already, load
+                history.put(version, readStore(getFileName() + "_" + version));
+            } else {
+                // either copy and split
+                history.put(version, copyAndSplit(version, postFix));
```
sqrrm, line 139: This will be done for all historical versions but copyAndSplit doesn't know about versions so it will make an identical copy of the resource data and put a version number on it.

I think there is something not right with this. We should be shipping this version with a split store (which will just be one old version resource). As time goes we will add more versioned resource files. It might be the old storage format doesn't have to ported in code at all. If we just rename the resource files we have to name_1.3.4_BTC_MAINNET this version is already good to go.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -213,19 +213,33 @@ public GetUpdatedDataRequest buildGetUpdatedDataRequest(NodeAddress senderNodeAd
...
+     * @param knownPayloadHashes
+     * @throws NoSuchElementException if there is no "special key" in the list
+     * @return the "special key"
+     */
+    private String containsSpecialKey(Set<P2PDataStorage.ByteArray> knownPayloadHashes) {
+        return knownPayloadHashes.stream()
+                .map(byteArray -> new String(byteArray.bytes).trim())
+                .filter(s -> s.matches("^[0-9]\\.[0-9]\\.[0-9]$"))
+                .findFirst()
+                .orElseThrow();
```
sqrrm: Please don't use exceptions as part of the expected execution path. This is not how Bisq uses exceptions so it causes confusion.

sqrrm (commented):
Thinking a bit about it I guess there is conceptually the same model we have now with one store, just split in many.

I am having a hard time getting a good overview of what's going on, but I think there is some muddling of concerns. A new client would not have to know about the old format as I see it. Only those nodes delivering data (seednodes typically) need to separate the kind of request they're getting and deliver data accordingly.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
freimair: oh my. sorry. bug. fix is coming up.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
freimair: > make an identical copy of the resource data and put a version number on it.

I believe you got it already. We rename the resource files and stick a "version" to it manually. So eg. with 1.3.6 we ship 3 stores: name_1.3.4_BTC_MAINNET, name_1.3.5_BTC_MAINNET, name_1.3.6_BTC_MAINNET. Then, as copyAndSplit knows about the version, each file is copied if necessary.

- the storage format remains the same. It is still Protobuf and files.
- `name_1.3.4_BTC_MAINNET` yes, exactly (I will be renaming the files in the PR once all is said and done)
- but in the future, there is not one big data store, but multiple - one for each release of Bisq
- why multiple?
  - because then we get some time relations and thus, chronological order. Once we have some order, we can start deciding which parts we keep in RAM (not in this PR, but down the line)
  - reduce the amount of data added to the repository by each release (deprecated by https://github.com/bisq-network/bisq/pull/4114)
  - in the long run, keep small chunks if possible

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
freimair: Just played around with it a bit. Doing an old-fashioned for-switch-if routine is about 10 times faster. However, 10 times faster means it takes an average of 2,6ms vs. 23,4ms (60k objects, average over 100 tries) for an operation that is performed twice during the runtime of Bisq (file loading time is about 3 seconds). In contrast, it takes 20 LOC more to achieve this. 

ad deterministic behavior: I haven't found any hint on non-deterministic behavior. Creating a String from byte[] is pretty standard. I, however, added a charset for good measure. Although "\d.\d.\d" contains nothing another charset would handle differently.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
dmos62: Reading this is a bit difficult, because `getMap()` is easy to confuse with `store.getMap()`.

`getMap().containsKey(hash) == false` implies `store.getMap().containsKey(hash) == false`, because `getMap()` is a union of `store` and `store`s inside `history`, so L76 here will always insert a new entry. I'd like to point out that these `.put` and `.putIfAbsent` implementations (their returns) have different semantics than in `java.util.Map` (`.put` returns `void`). Same in `MapStoreService`.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -205,24 +214,61 @@ public GetUpdatedDataRequest buildGetUpdatedDataRequest(NodeAddress senderNodeAd
...
+
         // We collect the keys of the PersistableNetworkPayload items so we exclude them in our request.
         // PersistedStoragePayload items don't get removed, so we don't have an issue with the case that
         // an object gets removed in between PreliminaryGetDataRequest and the GetUpdatedDataRequest and we would
         // miss that event if we do not load the full set or use some delta handling.
-        Set<byte[]> excludedKeys = this.appendOnlyDataStoreService.getMap().keySet().stream()
-                .map(e -> e.bytes)
-                .collect(Collectors.toSet());
+        Set<byte[]> excludedKeys;
+        if (seedNodeRepository != null && seedNodeRepository.isSeedNode(networkNode.getNodeAddress())) {
```
dmos62: Meaning of this predicate not immediately obvious: maybe `var weAreASeedNode = seedNodeRepository != null && seedNodeRepository.isSeedNode(networkNode.getNodeAddress())`?

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -205,24 +214,61 @@ public GetUpdatedDataRequest buildGetUpdatedDataRequest(NodeAddress senderNodeAd
...
+        } else {
+            excludedKeys = this.appendOnlyDataStoreService.getMap("since " + Version.VERSION).keySet().stream()
+                    .map(e -> e.bytes)
+                    .collect(Collectors.toSet());
+        }
 
         Set<byte[]> excludedKeysFromPersistedEntryMap = this.map.keySet()
                 .stream()
                 .map(e -> e.bytes)
                 .collect(Collectors.toSet());
```
dmos62: `.keySet().stream().map(e -> e.bytes).collect(Collectors.toSet())` is applied 3 times in this method. Moving it to a method (like `getKeySetInBytes(Map)`) would improve readability.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -205,24 +214,61 @@ public GetUpdatedDataRequest buildGetUpdatedDataRequest(NodeAddress senderNodeAd
         return new GetUpdatedDataRequest(senderNodeAddress, nonce, this.getKnownPayloadHashes());
     }
 
+    /**
+     * Create the special key. <br><br>
```
dmos62, line 218: "Special key" is a problem semantically, in that it doesn't describe how it is special. Better to be concrete: it's the release version string in bytes. That's not my main point here, but I noticed a few phrasings like this so felt that it should be mentioned.

The actual problem is that there's string and byte manipulations in various files in this PR. You can put all this logic in `bisq/common/app/Version.java` under `toP2PDataStorageKey`, `static fromP2PDataStorageKey`, etc. Most places where strings are being passed around, `Version`s should be passed around instead. For example, `Version.history` would do better as `List<Version>`, than `List<String>`.

p2p/src/main/java/bisq/network/p2p/storage/persistence/AppendOnlyDataStoreService.java
```diff
@@ -55,6 +55,18 @@ public void readFromResources(String postFix) {
         services.forEach(service -> service.readFromResources(postFix));
     }
 
+    /**
+     * TODO to be implemented
```
dmos62: Probably forgot to remove this.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
```diff
@@ -0,0 +1,201 @@
...
+ * </ul>
+ * </p>
+ *
+ * <h3>Further reading</h3><p><ul>
+ *     <li><a href="https://github.com/bisq-network/projects/issues/25">project description</a></li>
+ * </ul></p>
+ */
+@Slf4j
+public abstract class SplitStoreService<T extends SplitStore> extends MapStoreService<T, PersistableNetworkPayload> {
+    protected HashMap<String, SplitStore> history;
```
dmos62, line 51: `HashMap<Version, SplitStore>` would be better.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
```diff
@@ -0,0 +1,201 @@
...
+            String finalFilter = filter.replace("since ", "");
+            if (!finalFilter.equals(Version.VERSION)) {
+                history.entrySet().stream()
+                        .filter(entry -> parseSpecialKey(entry.getKey()) > parseSpecialKey(finalFilter))
+                        .forEach(entry -> result.putAll(entry.getValue().getMap()));
+            }
+        }
+
+        return result;
+    }
```
dmos62, line 105: `getMap(String)` is cryptic. `getObjectsSince(Version)` would be much better. Notice that gets rid of the language/DSL. Doing a language is not necessary, because currently there's only one usage for these version keys here: getting objects from newer releases than the one specified.

`parseSpecialKey(finalFilter)` is called a lot, could be cached.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
```diff
@@ -0,0 +1,201 @@
...
+        SplitStore historicalStore = readStore(destinationFile.getName());
+        // - subtract all that is in resource files
+        store.getMap().keySet().removeAll(historicalStore.getMap().keySet());
+
+        // - create new file with leftovers
+        storage.initAndGetPersisted(store, 0);
+        storage.queueUpForSave();
+
+        return historicalStore;
+    }
```
dmos62, line 201: Why are we copying historical datastore checkpoints/splits out of resources? Because that's what the live-store-only solution we currently use does? For troubleshooting purposes? Seems unnecessary, though I may be missing something. I presume we keep the live store in the working directory, because we can't mutate resource files (is that even true?).

dmos62 (changes requested):
Requested changes. Mostly related to clean up, but still significant. I found readability difficult. Would appreciate more helpful names and more intermediary variables.

I wrote up a description of the PR and the problem it's solving. I'll include it below. Some of the things there may not be correct; it represents my ongoing learning about Bisq's P2P (and this PR). @freimair, after corrections, if they're necessary, would you consider including some of this information in an appropriate place in the code? I think future devs will appreciate an explanation of why this datastore design was chosen. I would appreciate feedback if I got something wrong.

------------------------

A datastore is a store of Bisq p2p objects. It is synced between peers-peers and peers-seednodes, 
and this way makes up a distributed database.

I presume that a peer's datastore is synced in this way:
  - a peer/client starts up;
  - it starts a "bulk sync" procedure:
    - gathers up UIDs (or hashes) of the objects it already has in its datastore;
    - uploads those UIDs to a seed node;
    - seed node takes a set difference of the passed UIDs and the UIDs of the objects it has;
    - seed node uploads to the peer/client the objects corresponding to that difference;
  - peer then enters p2p mode, listening for and publishing updates.

The problem is that this "bulk sync" procedure includes the peer sending a large number of UIDS
to the seed node, which can timeout, in case of insufficient network throughput, causing the
syncing to unrecoverably terminate.

That shows two problems:
  - Bisq does not recover from throughput issues;
  - the bulk sync is bandwidth-intensive.

This spec addresses the bandwidth-intensity.

Chronological information in the p2p objects would allow querying it by comparing objects to a
timestamp-like value; however, the p2p objects do not contain chronological information. It is
unclear whether introducing a timestamp would have any negative effects, but it would involve
making sensitive changes.

Freimair's proposal is about using chronologically-indexed datastore "checkpoints" that are 
distributed with releases
  - (and thus identical across clients and seednodes)
  - each checkpoint contains objects generated since the last checkpoint
    - the first checkpoint contains all prior objects
    - this effectively assigns chronological information to objects
      - precision of which, for the purposes of bulk sync, corresponds to frequency of releases;
  - this way a bulk sync can be initiated by identifying
    - the latest checkpoint that the client has
    - and the objects it has stored since then (objects not in any checkpoint known to him)
    - which reduces the bandwidth requirement
      - to the equivalent of uploading UIDs that are newer than the release used
        - because identifying the checkpoint is negligible bandwidth-wise.

Requirements for the checkpoint datastore:
  - initializable from the previous monolithic datastore or from scratch;
  - a "running" checkpoint has to be maintained;
    - which holds objects generated since the last "release" checkpoint;
  - at no point can the datastore be mutated, except via appends;
  - "release" checkpoints must be identical across peers and seednodes.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
freimair: I see this whole PR a small-as-can-be bugfix for the 3/4 problem. The "followup"-project would be to remove the protobuf-files altogether and make everything work with a proper database (https://github.com/bisq-network/projects/issues/29). When we approach that project, all the file-handling will be removed and cleaned up.

I agree that the implementation is not as clean as could be, but I rather focus my efforts and not do stuff twice.

freimair:
@dmos62 thanks for your review. Appreciate it.

First of all, your PR description is correct.

Furthermore, I believe I can address all of your comments in summary:
I see this PR as an as-small-as-possible fix for the 3/4 problem. Yes, files are copied around where they probably do not need to be, APIs could be more readable, return values are ignored or misused, and a lot of other stuff. Thing is, most of the issues you raised have been in the code already and I do not want to touch them. They have proven themselves. Keep the changes small, keep the isolated. Its a minefield.

That being said, https://github.com/bisq-network/projects/issues/29 is in the queue. That project will get rid of the copying files around because there will be no more individual files, the API will need a major overhaul, the filter might get enhanced, the `isSeednode` code will be gone because it prevents our network from falling apart while upgrading from single data stores (now) to multiple (then) -After that is done, it is no longer needed.

Now to your obvious question: why not do it properly? I rather not spend time on reworking the same thing twice :smile: because I am a) a lazy bastard, and b) we need to use the resources we have efficiently. And of course, coding styles and our brains differ, so there is no perfect PR anyways.

dmos62:
@freimair followed up above post with a Keybase message and we had a brief exchange there. I'll summarize how I see this situation and what my concerns are:

- P2P datastore code has a fair amount of technical debt, which makes making changes to it without compromises slow;
- we want to introduce changes quickly, which encourages us to make compromises;
- compromising changes tend to degrade reliability and are more likely to introduce mistakes;
- compromising changes require more resources to maintain: freimair points out that there are plans to replace large portions of this code entirely, which means that if all goes well those portions will not require maintenance;
- compromising changes require more resources to review.

My concerns:

- I've spent a long time on the PR (~2.5 days, I realise this says as much about my reviewing ability as it does about the PR), but I'm still not confident that all edge cases are covered; I don't feel comfortable seeing this merged at this time;
- I think that reliability, maintainability and reviewability are preferable to speed, especially in sensitive code, though speed has its importance too; it's a balancing act;
- introducing changes that are difficult to maintain, because if all goes well they won't have to be maintained, is planning for the best outcome; I don't think that's prudent;
- I think it's good to address this human element: PRs that make hasteful changes, or side-step technical debt, often require the reviewer to address that complexity anyway, so that he can make sure that the changes are correct, which can be frustrating for the reviewer (who usually always has his plate full as is).

Idea:

Stack a backup system on top of this in such a way that the mutations we perform cannot lose data. Of course we have to be aware that this is addressing complexity with more complexity. Maybe this is an example of how _not_ to solve this situation.

chimp1984:
A few "high level" comments/thoughts:

By splitting up the data into groups we are creating "blocks" (I use that terminology with intention). How to uniquely identify a block in a distributed system is a challenge. 

The approach used here makes the shortcut via the release maintainer to be the single entity to define what is the right "block". This might not be an issue (or at least does not make things worse as it is) but I am not 100% sure. The checkpoint terminology reveals the centralisation issue.

Is there another way to do it?
We have date fields in most (all?) of the relevant data structures and could use that for sorting the items but there could be multiple items with the same date. To get unique IDs ans sorting we could add the hash fields and create a tuple {date, hash} as ID  and sort primarily by date, secondarily by hash. 
But there can be a late insertion as well which could mutate our data blocks. For certain data we don't allow insertion outside a defined time period tolerance window (`DateTolerantPayload`). So we could consider blocks of a certain age as immutable at least. Newer blocks might get "reorged". Handling "reorgs" sounds like looking for toubles though...

Another approach could be to use a fixed time period window (e.g. 10 days) and fill up all data whose date falls into that period. The GetData request would send the index of the time window and the hash of its content. If the Seed node has a different hash it will send the whole package. For older data that should be very rare and the additional network overhead might be a good trade off with the more simple approach and low processing costs. The hash gets calculated once and is stored with the package. A problem is how to handle the case if a user has more data items than the seed node. The it would request each time again the package. If a seed is missing data it would get dossed by the network....

Another approach might be to use Bloom filters for the data request, but that is likely expensive for the seed node to process. A probably bigger problem than the 2.5 MB request is the load on the seed node to process the response.

dmos62:
The main problem that we're trying to get around is that p2p objects are meant to be indexed by hash and don't have properties that would allow them to be indexable by something lightweight (like time). Bloom filters are normally the solution for these situations, but their processing is memory intensive, which opens up Bisq to denial-of-service attacks. It might be worthwhile to look at academic literature on the topic.

I can't dig deeper now, but I found some interesting articles following these links. Proabilistic set membership data structures are actively researched:
- https://scholar.google.com/scholar?q=bloom+filters+alternative+memory
- https://scholar.google.com/scholar?q=bloom+filters+memory

chimp1984:
@dmos62 Thanks for your input!

After a discussion with @sqrrm we agreed the best way forward is to limit a first mitiagtion to trade statistic objects only. This will reduce bandwith requirement by 50% as well as seed node processing load by 50% (as its about 50% of the hashes). It will also reduce required RAM for the Bisq app if it is only loaded on demand.

Rough idea:

Prune existing trade statistics by removing all data older than a cut off date - e.g. 2 months (TBD). Beside that handling of trade statistics data is exactly as it is now but instead of 4 years of data we only deal with 2 month. I assume that will be about 1-5% of current data.
At each release we add a snapshot named with the release version which contains all historical data until a defined cutoff date (can have an overlap with the in-memory data to be on the safe side - details have to be defined once we work on it).
This historical data is available for new users as its in the binary. For not updated users they will request it with a new request call from the seed nodes. They pass the versions they have and receive the missing ones. This will be only done once the data is actually needed (user goes on historical trade statistics UI). We could even spit that UI in recent trades which covers the in-memory data and historical data which makes the request and/or loads the shipped resource files. After leaving this UI we unload that data from RAM. Loaded data and in-memory data might overlap but as its hashmaps those will get canceld out. 

This approach reduces a lot the risks as if anything goes wrong we only screw up trade statistics which are non-critical data.
If the concept of using version based buckets works out well it can be extended to the next data type like account witness objects (here we dont have the concept of on-deman data as all data are required).
 
Backward  compatibility:
We need to take care that not updated nodes still work properly. I guess there are no big issues beside potential extra load when a updated seed node request data from a not updated seed. Here we get a bit of extra load as the not updated seed node would deliver all the missing data (up to the hard cap limit). The updated node would ignore those historical data as they will not fit the date check predicate. With customizing the update of seeds by assigning a seed to connect with (one which has alreayd updated) at startup we can minimize that load. Beside that we could use other tools like capabilities, verion numbers, new data fields as flags, activation date to make the transition smooth.
Users who have not updated longer than 2 months would see gaps in the trade statistics, but that is not a major issue as that data is only used for informational purpose. We can even require an update via the filter after a certain time.

I might work on that myself in a couple of weeks if I find time. Otherwise if anyone wants to pick it up, please leave a message.

sqrrm:
Having discussed it and considering the risks it seems much safer to only focus on the tradestats to begin with. The risks are very high if any of the DaoState data or account witness data is lost. Losing some tradestats data is not critical in the same way but could still yield results. It would also help to prove the concept.

chimp1984:
I started working on it: https://github.com/bisq-network/bisq/pull/4405 Lets move discussion over there....

freimair:
A bit late, I know, but I can be online again, local infrastructure is fixed, and I take my time to comment on your discussions. Although I would suggest a call sometime, just to get somewhere.

### ad rough idea
pruning data does not solve the problem.
- it lowers usability
- we fix a piece of code that is at its limits with throwing even more code on it, thus increasing complexity instead of decreasing it
- it adds a lot of complexity down the road as we have to handle missing data throughout Bisq
- once Bisq gets more successful, we need to start pruning data a lot more rigorously (one month, 2 weeks?)

### ad another way to do it
- we cannot use time (as you guys already realized)
- the only thing we control is releases. someone throws a number of messages into a bucket and put a stamp on it. Then we ship this bucket. No issues with later additions, not even incomplete buckets are an issue because the network keeps the databases synced anyways.
- `The checkpoint terminology reveals the centralisation issue.` That is correct in a couple of ways:
  - `reveals`, because the issue is there already, cannot be discussed away
  - we have to deal with it eventually
  - by leaving it untouched, we do not move and cannot deal with "it"
  - we cannot solve all and everything now

### ad backwards compatibility
- seed nodes communicate using the old mechanics. nothing changed there. Because of all the reasons you mentioned and more.

### ad `if any of the DaoState data is lost`
- DaoState data isn't touched by this PR at all, and although the massive data store causes real issues, it does not have anything to do with the 3/4 problem

### ad loosing any data in general
- no data is deleted, no pruning, no shuffling around.
- the only thing that happens is to move data from the live database once a new bucket arrives (through app update) and thus, reduce the Disk IO load

### ad `Having discussed it and considering the risks it seems much safer to only focus on the tradestats to begin with. `
- focusing on tradestats is possible if you guys so desire
- however, I do not understand why there is this fear of loosing data because of this PR. Yet, pruning data - ie. intentionally loosing data, seems to be a way forward?
- I do however, fully understand, and have seen it happen, that we loose data once in a while because the current implementation cannot cope with the sheer amount of data anymore (because eg. it takes too long to write to disk and files get clipped [which by the way is addressed by this PR as well, at least for non-dao stuff])
- I do fully understand that we deny users access to Bisq which do not own a powerful dev machine
- I do fully understand that we deny users access to Bisq which do not have decent broadband internet access

### and finally
I am fully aware that there is a lot of technical dept in these parts of the code. And the dept needs to go. But we cannot eliminate this dept in one step (well, actually, we might be able to pull it off, but that would be a complete rewrite of the p2p module and all the risks that comes with touching this part of the code). And it seems like we are discussing stuff over and over again, not coming to a consensus for months. Meanwhile, the actual request size has grown beyond 6MB by now and we are denying more and more users access to the bisq network (and thereby loosing them).

The actual bad thing here is that the code stayed as is, no technical dept added, non removed, nothing changes. Just a massive waste of time.

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
freimair: I somehow can agree on 'special key' to not be ideal. Fix coming up...

On the other hand, I fully agree on manipulating Version strings is downright bad and not desirable. However, refactoring `version` format would affect big portions of Bisq. Thus, it would directly violate our doctrine of keeping PR small and right to the point. One has to draw the line somewhere, here it is.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
freimair: again, not touching the `version` format in order to keep the PR small and down to the point.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
freimair: > Why are we copying historical datastore checkpoints/splits out of resources? Because that's what the live-store-only solution we currently use does? For troubleshooting purposes? Seems unnecessary, though I may be missing something. 

the existing file storage mechanisms are all tailored to working in the working directory. Changing that would be equivalent to replacing the file storage mechanisms with something else entirely. Hence, leaving it as is (until https://github.com/bisq-network/projects/issues/29).

> I presume we keep the live store in the working directory, because we can't mutate resource files (is that even true?).

Well, obviously, the "live store" is something that is created and maintained by the Bisq app per installation and even per (configurable) app name and path. Placing that in the resources would only allow for a single instance of the data store AND would be overwritten and therefore invalidated on every app upgrade.

p2p/src/main/java/bisq/network/p2p/storage/persistence/SplitStoreService.java
freimair: you are correct, YAGNI right now. However, that will change swiftly once we get to reducing the memory footprint (which will start with https://github.com/bisq-network/projects/issues/29)

chimp1984 (changes requested):
NACK

Basic concept is ok but many impl. elements are not acceptable IMO. Specially the hack to encode the version into the payload of the hashes instead of adding a new protobug field to the request msg.

I suggest that we use #4519 which is based on this PR one but has a lot of changes. It is still WIP and not much tested, should be ready for review in a few days...

chimp1984:
@sqrrm @ripcurlx Can you close that as it was replaced by a merged PR

PR #4247, opened 2020-05-09
Extend FeeRateService to support multiple API endpoints

### LINKED ISSUE #27: Migrate Bisq pricenodes from earn.com fee estimation API to our own self-hosted mempool fee estimation API service
> _This is a Bisq Network project. Please familiarize yourself with the [project management process](https://bisq.wiki/Project_management)._

## Description

This is a project proposal to migrate Bisq pricenodes away from using the earn.com fee estimation API in favor of an open-source self-hosted mempool fee estimation API backend, that we can decentralize further by having all Bisq btcnode operators run.

## Rationale

Whenever the Bitcoin market sees major price action, trading volume suddenly spikes, which results in a huge rush of new transactions into the Bitcoin mempool, and "next block" fees rise very quickly:

Screen Shot 2020-03-12 at 14 09 39

This sudden rise in mempool fees causes a major issue in Bisq whereby any trades taken during the time when the mempool fees start to rise use far too low of a fee, and they do not get confirmed for several hours or even days later, until the mempool clears out. The cause of this issue is that Bisq currently uses Earn.com fee estimation API which is slow and not well maintained. The https://mempool.space backend uses realtime data from Bitcoin nodes, so it is extremely fast to return realtime fee estimates, compared to the Earn.com API which is always lagging behind:

Screen Shot 2020-01-22 at 1 47 36

This issue was first reported by @ManfredKarrer in https://github.com/bisq-network/bisq/issues/1077 but at the time, a better mempool fee estimation API service did not exist. But now we have a great open-source and self-hosted solution, the [mempool project](https://github.com/mempool/mempool). Of course I am bias, since I am a maintainer/contributor to this project, but IMO it really is the best mempool project for Bitcoin and we have worked very hard making it so.

Recently during the past few days we have seen this issue reproduce again, and now users are complaining in support that their deposit TX have still not confirmed even after waiting for 12 hours. So therefore I propose that we pull the trigger on ditching Earn.com API once and for all.

## Criteria for delivery

After all Bisq pricenodes upgrade to the new code that will query our federated btcnode mempool fee estimation backend service, we can consider it delivered.

## Measures of success

After all Bisq nodes start receiving the self-hosted mempool fee data, and users no longer experience this issue due to sudden fee spikes, we can consider it a success.

## Risks

I've been running the mempool backend and hosting https://mempool.space for over 6 months, and it has never crashed, so I feel it is quite stable to run in production. However, we could keep Earn.com API as a backup data source to mitigate the risk of any issues with our own self-hosted API service.

## Tasks

- [x] Modify the Bisq pricenode code to use the mempool fee estimation backend
- [x] Have all Bisq btcnode operators are running the mempool fee estimation backend
- [x] Have all Bisq pricenodes upgrade to the new code that will query our federated btcnode mempool fee estimation backend service

## Estimates

Dev: $2000 USD
Ops: $1000 USD

cd2357:
Extend `FeeRateService` and related classes to allow querying multiple fee estimation API endpoints, instead of just one.

The API endpoints are queried independently and in parallel. Each will recover gracefully from any connection issues or timeouts, not influencing the others.

The `FeeRateService` exposes the average of the retrieved fee rates to the Bisq clients. Alternatively, if no fee estimate could be queried, it will return `BitcoinFeeRateProvider.MIN_FEE_RATE`.

Up to 5 API endpoints are supported. They can be configured in the file `application.properties`.

When no configured API endpoint is reachable, or none is configured, the logic falls back to asking the https://mempool.space endpoint. By default, 2 endpoints are configured.

Fixes bisq-network/projects#27


cd2357:
cc @wiz

cd2357:
FYI for reviewers: I realise its a relatively big change, so here's a short "changelog" to help with the review:

* `build.gradle`
  * added test dependencies for Jupiter
* `README.md`
  * added short description of how the endpoints can be configured
* `FeeRateService`
  * before: get fee rate from provider
  * now
     * get fee rates from all available providers + calculate average
     * fallback to `BitcoinFeeRateProvider.MIN_FEE_RATE` if providers are (temporarily) unreachable
* `BitcoinFeeRateProvider`
  * convert to abstract class
  * define five subclasses of it, each with its own endpoint API, which represent the diffeent providers
  * annotate subclasses so that they are only instantiated when endpoint URL is defined
  * first subclass defaults to mempool.space in case nothing else is defined (or config is corrupted)
* `application.properties`
  * extend file with 5 properties (API endpoints for supported fee providers)
* the rest are tests

cbeams (approved):
ACK including my review commits. Please see each for details. I've tested this locally running against both the default mempool.space and mempool.emzy.de instances, and saw that everything works as expected. After starting up my pricenode locally and getting a couple successful polling results from both endpoints, I configured Little Snitch to deny connections to mempool.emzy.de, and saw the following error show up in the logs:

```
May-11 11:42:37.189 [Timer-0] ERROR b.p.m.p.MempoolFeeRateProvider$First: Error retrieving bitcoin mining fee estimation: I/O error on GET request for "https://mempool.space/api/v1/fees/recommended": Host is down (connect failed); nested exception is java.net.SocketException: Host is down (connect failed)
```

After removing the rule to deny connections, polling started working as per usual once again. So the implementation seems to tolerate intermittent failures of an individual endpoint well.

One thing that's not ideal about the implementation is that it doesn't log explicitly what the calculated average fee rate is. It only logs the result of each request against the individual mempool api endpoints. This makes sense, because the implementation only calculates the average on calls to `/getFees`, and it would create a lot of noise in the logs to log that each time. It's probably better to leave things as-is for now and if debugging errant fee rates becomes a problem in the future, we can add more sophistication to the logging.

Beyond what's mentioned above, I haven't done what I'd consider exhaustive testing of the implementation, but I have looked at everything closely and I think this is in good enough shape to merge and to give a careful trial deployment in production. @bisq-network/pricenode-operators should just be ready to roll back if something unexpected comes up.

Issue #2869, opened 2019-06-04
Add option to make trade 'reason' less obvious


rexi1e:
Its pretty easy for a bank that has recognized the format of a Bisq trade "reason" to identify which its customers are trading bitcoin with Bisq.

Could you make an option for the user to pick a randomly-generated trade reason that's less obvious?

Maybe source a couple hundred reasons from Venmo's public data or something and then randomly pick form that, or something...I dunno

peterzen:
I think this would be an important feature. Even something basic generating a separate ID for this field which would be varying in length, character composition and maybe a couple of static words added randonly, would go a long way.

rexi1e:
A presenters at HCPP mentioned Bisq as great KYC-free option to trade bitcoin, but their reason for not using was this.. Also many users having issues with Revolut, probably for same reason

You know [this endpoint](https://venmo.com/api/v5/public?limit=1000) show latest public Venmo txs. Check `message` values for some ideas ;)

peterzen:
I was there also :) Got inspired to pick this up if noone else is working on it yet.

ghost:
Here is the export of messages from the mentioned Venmo endpoint:
```
"Miss you",
"Groceries",
"Virtual coffee date!!!",
"To help w/ your bills this month",
"Thanks for everything you do",
"Rent's due, dude",
"Takeout > Going out",
"Thank you, friend",
"Pizza for din. And bfast too.",
"You got this",
"Dinner",
"Treat yo self",
"Grab a snack on me!",
```

petrhejna:
I have found also this source for messages: https://github.com/fafrd/venmoscraping/blob/master/example.json

ghost:
And another one https://github.com/sa7mon/venmo-data it looks juicy, but I was not able to download the dataset. 

Also there is the study where they have extracted the data for: https://www.rpubs.com/toehmler/559525

-----------

Another one is: https://github.com/erikgregorywebb/datasets/blob/master/venmo-public-payments-raw-2019-01-01.csv

But there is a lot of emojis, that are maybe hard to type (I would like to stick to plaintext). And also there are some quite akward messages like:
 `"Best head of my life✊🏾🍆💦😛"`

Here is polished list of transaction messages found in this dataset: https://gist.github.com/petrhejna/84a5718ae1426448b851b8091cfd87c8

-------- 

I would even say, we need to generate our own dataset (do some copywriting). Not sure if data from Venmo can be used.

[...]

PR #4249, opened 2020-05-09
Reason for payment is now list of fake reasons instead of random string ID to obfuscate fact its from the Bisq

ghost:
Solves https://github.com/bisq-network/bisq/issues/2869

image

----

For now, the list of reasons is very limited, and I need to find some more (like hundreds more) reasons. But from the code perspective it's working.

I did some research in the https://github.com/bisq-network/bisq/issues/2869 issue, and I doubt that Venmo is usable.

### DIFF
--- a/common/src/main/java/bisq/common/util/ReasonsForPayment.java
+++ b/common/src/main/java/bisq/common/util/ReasonsForPayment.java
@@ -1,0 +1,30 @@
+package bisq.common.util;
+
+public class ReasonsForPayment {
+
+    private static final String[] REASONS = {
+            "Miss you",
+            "Groceries",
+            "Virtual coffee date!!!",
+            "To help w/ your bills this month",
+            "Thanks for everything you do",
+            "Rent's due, dude",
+            "Takeout > Going out",
+            "Thank you, friend",
+            "Pizza for din. And bfast too.",
+            "You got this",
+            "Dinner",
+            "Treat yo self",
+            "Grab a snack on me!"
+    };
+
+    public static String getReason(String id) {
+        int hash = 7;
+        for (char c : id.toCharArray()) {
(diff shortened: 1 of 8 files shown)


common/src/main/java/bisq/common/util/ReasonsForPayment.java
```diff
@@ -0,0 +1,30 @@
+package bisq.common.util;
+
+public class ReasonsForPayment {
+
+    private static final String[] REASONS = {
+            "Miss you",
```
ghost, line 6: Should we translate this via `resources/i18n`?

cd2357:
NACK.

Interesting concept, but I think it adds complexity and might actually make it easier to match such a payment description to Bisq.

The random string used now is used a lot in other places as well (some merchants use it as Order ID, or Purchase ID, to match it with orders in their own payment systems).

Also, it is sometimes a random alpha-numeric string, sometimes only numeric. Which obfuscates it even more.

So I'd say it's good the way it is now.

ghost:
True, I agree with your points. The only thing I like to solve is how to have issues ready for implementation, that were **accepted** and are actually **wanted** before anyone picks them up and spend time on them. 

I'll open this on Keybase chat as well.

wiz:
NACK, mediators need the trade ID in the payment details field to resolve disputes

dmos62:
NACK, small dictionary size means that this is private by obscurity (that's bad). If we wanted to increase dictionary size, we would have to use some kind of dynamic generation, but then how to make it resilient to heuristics? Plus what @wiz said about needing trade ID in the payments field. @petrhejna sorry for the bad experience. You happened upon an issue whose solution is easy to implement (most likely), but hard to find. @sqrrm I think there's a concensus that this PR should be closed.

sqrrm:
Closing as not accepted.

PR #4250, opened 2020-05-10
Bugfixes related to switch statements

dmos62:
This is a second go at https://github.com/bisq-network/bisq/pull/4212, which was reverted. The previous PR was fixing some problems with switch statements and a typo where a short-circuit boolean operator was used without need. This PR is the same as the last one, but the semantic/cosmetic changes (why the original PR was reverted) are removed and it also implements a suggestion by @ManfredKarrer to add a (forgotten) check.


ghost:
For the `ChangeParamValidator` file, there is `ChangeParamValidatorTest` already created. I believe it is *very* useful to add tests for bugs to prevent re-occurrence. 

Also adding a test for validator is (I believe) technically easy (in contrast to some UI changes).

sqrrm (commented):
utACK

@ManfredKarrer I still need your review on this to be able to merge, but I think this is a reasonable and safe change.

ManfredKarrer (approved):
utACK

sqrrm (approved):

PR #4256, opened 2020-05-12
Display correct trade and tx fees in CSV export

### LINKED ISSUE #4207: Exporting trade history to CSV has incorrect fee values
### Description

Exporting the trade history CSV using the EXPORT TO CSV button in PORTFOLIO -> HISTORY produces incorrect fee values ("Transaction Fee" and "Maker Fee").

#### Version

1.3.2

### Steps to reproduce

See description.

### Expected behaviour

EXPORT TO CSV creates a CSV with correct values for columns "Transaction Fee" and "Maker Fee". 

### Actual behaviour

EXPORT TO CSV creates a CSV with incorrect values for columns "Transaction Fee" and "Maker Fee". For example, my CSV says Transaction Fee: 0.00006080; Maker fee: 0.00043750; but that trade's JSON data is "txFee": 6560, "makerFee": 36. The fees for this trade were paid in BSQ.

#### Device or machine

Linux Mint 19.3 Tricia x86_64
MATE

ghost:
When exporting trade history to CSV, users would expect to see their own transaction and trade fees listed.  Maker fees are stored on the Offer, Taker fees are stored on the Trade object.
The code was not properly checking whether the user was the Offer Maker or Taker, and as such the report was always displaying the Taker information for Tx and Trade fee.  Therefore any trades where the user was Maker showed the wrong fees.

The column title `Maker Fee` is incorrect and misleading since the user doing the export is not necessarily the Maker.  We rename this column to `Trade Fee` to indicate that it is the fee the user paid for trading, regardless of Maker/Taker status.

Fixes #4207


ghost:
### Testing

- I generated the CSV export against my trade history in `pre` and `post` environments, then compared the files with Meld Diff Viewer.  Noted that the trades in which I was Maker the fees changed to reflect the correct values from the Offer object.

ghost:
Code :+1:, I only add that I strongly suggest to write unit-test for this. I believe its easy to write and it will help in the future to a) explain why it is this way, b) prevent part of code from braking in the future.

sqrrm (approved):
utACK

PR #4258, opened 2020-05-15
Prevent transient negative missingCoin amount sent to QRCode display

ghost:
In bisq-network/bisq/pull/4223 a fix was made to the QRCode display of amount needed to fund an offer.  This accidentally introduced a bug whereby if the user overpaid the funding, then the QRCode would throw an error `Coin must be positive`.  This is due to a race condition in OfferDataModel which was setting a property first to a negative amount before checking if it was negative and if so then setting it to zero.  The fix here is to check for negative before setting the property so that any listeners do not get an event (negative) followed by another event (zero).

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/offer/OfferDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/OfferDataModel.java
@@ -73,9 +73,10 @@ public abstract class OfferDataModel extends ActivatableDataModel {
             balance.set(tradeWalletBalance);
         }
         if (totalToPayAsCoin.get() != null) {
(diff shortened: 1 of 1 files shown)


ghost:
### Testing

To reproduce the error: create an offer and fund it from your external wallet.  If you send more than the required amount, an error message will pop up.

image

### Verification

Make the same test (overfund your trade from an external wallet).  With the fix, no error message will be displayed, and the QRCode will indicate the amount remaining = 0.

image

sqrrm (approved):
utACK

Good fix. Reminder to not set properties that are observed to intermediary values.

PR #4260, opened 2020-05-15
Fix trade withdraw to external wallet step 4

### LINKED ISSUE #3592: UTXO combinations resulting in dust outputs
I came across the following error when trying to withdraw my security deposit to an external wallet after a successful trade.

Screen Shot 2019-11-09 at 10 41 43 AM

The workaround for this was to send it first to my Bisq wallet and from there send it to my external wallet.

According to Manfred...."this is likely a bug in certain utxo combinations. fee estimation is tricky and can become an infinite task (we break after certain iterations) if utxos are unfortunate... e.g. if u add a input the size is bigger and require more fee, so u need to add another input -> can continue....dust outputs are not allowed so it can become a bit tricky. hard to test those edge cases but we should fix that issue..."

### LINKED ISSUE #4205: Final step of trade (withdraw BTC to external wallet) fails
### Description

I went through a trade to buy BTC (as a market taker).

In the last step I was told the trade is complete, I can now withdraw the BTC to the Bisq wallet, or to an external wallet.

The amount available to withdraw was 0.016 BTC (amount bought + deposit).

I chose the "External wallet" option, entered the destination address (P2SH, starting with 3..).

Attempting to withdraw failed with

> The fees for that transaction exceed the available funds or the resulting output value is below the min. dust value:
> Missing 0.016 BTC 

Withdrawing to the Bisq wallet went through though.

#### Version

1.3.2

### Steps to reproduce

In the last step of the trade, when withdrawing BTC, attempt to withdraw to an external wallet (P2SH address).

### Expected behaviour

The same behavior as with withdrawing to Bisq wallet. When initiating withdrawal to external wallet, it should:

* show overview of prepared tx (fee, amount, destination address) + ask for final confirmation
* on confirm, successfully broadcast tx

### Actual behaviour

Clicking the "Withdraw to external wallet" fails with the error above, before even the confirmation popup is shown.

### Screenshots

Didn't take a screenshot, but see log snippet below.

#### Device or machine

Fedora 30

#### Additional info

```
Apr-28 21:01:11.592 [JavaFX Application Thread] INFO  org.bitcoinj.wallet.Wallet: Completing send tx with 1 outputs totalling 0.016 BTC and a fee of 0.00 BTC/kB 
Apr-28 21:01:11.595 [JavaFX Application Thread] WARN  org.bitcoinj.wallet.Wallet: Insufficient value in wallet for send: needed 0.016 BTC more 
Apr-28 21:01:11.596 [JavaFX Application Thread] ERROR b.d.m.p.p.s.b.BuyerStep4View: The fees for that transaction exceed the available funds or the resulting output value is below the min. dust value:
Missing 0.016 BTC 
bisq.core.btc.exceptions.InsufficientFundsException: The fees for that transaction exceed the available funds or the resulting output value is below the min. dust value:
Missing 0.016 BTC
	at bisq.core.btc.wallet.BtcWalletService.getFeeEstimationTransaction(BtcWalletService.java:923)
	at bisq.desktop.main.portfolio.pendingtrades.steps.buyer.BuyerStep4View.reviewWithdrawal(BuyerStep4View.java:182)
	at bisq.desktop.main.portfolio.pendingtrades.steps.buyer.BuyerStep4View.lambda$onWithdrawal$2(BuyerStep4View.java:166)
...
```

ghost:
The routine `cleanUpAddressEntries` in TradeManager was prematurely releasing funds associated with trades when they reach the last step of the process (Step 4, Withdraw to External Wallet). 

Usually, with a single trade active this would not be an issue because `cleanUpAddressEntries` is called after the withdrawal, but if you have more than one trade at Step 4 then the first withdrawal would go though, but all the others would fail with an error box "Missing x.xxx BTC" (the trade proceeds or deposit amount).  Alternatively, if you restart the bisq app with a
trade in Step 4, the same cleanup will occur and the attempt to withdraw to external wallet will fail.

The change here considers any trade held by the TradeManager to be in use and therefore will not have their associated address entries freed up.  After Step 4 has passed, the trade is no
longer held by the TradeManager, and so `cleanUpAddressEntries` will at that point free up the address.

Fixes #4205
Fixes #3592

### DIFF
--- a/core/src/main/java/bisq/core/trade/TradeManager.java
+++ b/core/src/main/java/bisq/core/trade/TradeManager.java
@@ -360,12 +360,12 @@ public class TradeManager implements PersistedDataHost {
     private void cleanUpAddressEntries() {
         // We check if we have address entries which are not in our pending trades and clean up those entries.
         // They might be either from closed or failed trades or from trades we do not have at all in our data base files.
-        Set<String> tradesIdSet = getTradesStreamWithFundsLockedIn()
+        Set<String> activeTrades = getTradableList().stream()
                 .map(Tradable::getId)
                 .collect(Collectors.toSet());
 
         btcWalletService.getAddressEntriesForTrade().stream()
-                .filter(e -> !tradesIdSet.contains(e.getOfferId()))
+                .filter(e -> !activeTrades.contains(e.getOfferId()))
                 .forEach(e -> {
                     log.warn("We found an outdated addressEntry for trade {}: entry={}", e.getOfferId(), e);
                     btcWalletService.resetAddressEntriesForPendingTrade(e.getOfferId());


ghost:
### Test scenario (issue, prior to fix):
- Create a trade, take it through to Step 4.
- Shut down bisq and restart it, to trigger cleanUpAddressEntries.
- Click Withdraw to External Wallet.  It fails with an warning box `Missing x.xxxx BTC` and in the logs `Insufficient Funds` exception.


### Test scenario (with fix applied):

- Create a trade, take it through to Step 4.
- Shut down bisq and restart it, to trigger cleanUpAddressEntries.
- Click Withdraw to External Wallet and verify that it succeeds.
- Verify that log messages are logged like this, AFTER the withdrawal succeeds.
`We found an outdated addressEntry for trade .. context=TRADE_PAYOUT`
`swap addressEntry .. offerId from context TRADE_PAYOUT to available`

sqrrm (approved):
utACK

More people testing this during release testing would be good.

PR #4262, opened 2020-05-16
standardize tense in trade process wording

tylerchambers:
Adjust the wording in the trade process to all have a matching tense.

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -564,9 +564,9 @@ portfolio.pending.invalidDelayedPayoutTx=Please do NOT send the Altcoin or Fiat 
 portfolio.pending.step1.waitForConf=Wait for blockchain confirmation
 portfolio.pending.step2_buyer.startPayment=Start payment
 portfolio.pending.step2_seller.waitPaymentStarted=Wait until payment has started
-portfolio.pending.step3_buyer.waitPaymentArrived=Wait until payment arrived
+portfolio.pending.step3_buyer.waitPaymentArrived=Wait for payment to arrive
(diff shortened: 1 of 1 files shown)


core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -564,9 +564,9 @@ portfolio.pending.invalidDelayedPayoutTx=Please do NOT send the Altcoin or Fiat
 portfolio.pending.step1.waitForConf=Wait for blockchain confirmation
 portfolio.pending.step2_buyer.startPayment=Start payment
 portfolio.pending.step2_seller.waitPaymentStarted=Wait until payment has started
-portfolio.pending.step3_buyer.waitPaymentArrived=Wait until payment arrived
+portfolio.pending.step3_buyer.waitPaymentArrived=Wait for payment to arrive
 portfolio.pending.step3_seller.confirmPaymentReceived=Confirm payment received
-portfolio.pending.step5.completed=Completed
+portfolio.pending.step5.completed=Complete
```
ghost, line 569: I suggest to align the key `portfolio.pending.step5.completed` to match the english wording -> `portfolio.pending.step5.complete`

ghost: (in both cases)

ghost: Um.. @petrhejna I suggest leaving the code tag `portfolio.pending.step5.completed` alone.  Its hardly any different from the proposed text and the ripple effects in the translation files would be quite large.

ghost:
NACK.  I can see that the tense does not match, but really don't think the original wording is so bad it justifies a change.

sqrrm:
I cannot really see the big difference in these formulations. Not worth the pain to change.

Issue #4255, opened 2020-05-12
Step 4 of trade: Rename "Move funds to Bisq wallet" -> "Keep funds in Bisq wallet"


cd2357:
### Description

The final step of a successful trade offers 2 buttons:
* Move funds to Bisq wallet
* Withdraw to external wallet

Judging by their wording, both imply an on-chain transaction. However, only the 2nd one actually causes a transaction, the 1st one simply marks the trade as complete and keeps the funds in the same (Bisq wallet) account.

My suggestion is: rename this button to "Keep funds in Bisq wallet"

That would more accurately describe what it does and avoid confusion for those who want to minimize transactions.

#### Version

1.3.4

PR #4264, opened 2020-05-16
Rename "Move funds to Bisq wallet" -> "Keep funds in Bisq wallet" to …

ghost:
Just renaming a text according to the proposal from the issue.

Fixies: #4255


ghost (changes requested):
Just change the text that is displayed to the user.  No need to change the code tag.  One line, one file change and no horrendous ripple effects.

ghost:
@jmacxx I understand your point, but I think the consistency is good and it helps to better communicate via the code. I also acknowledge it is a small thing, but still, the consistency of the code with the way how its meant for the user helps to communicate the reason why its there.

sqrrm:
@petrhejna No need to change the specific language files, only displayStrings.properties needs to be changed. The other ones are handled by the translation team and pulled at release time.

I agree with @jmacxx here, better not change the code tag.

ghost:
Closing in favor of https://github.com/bisq-network/bisq/pull/4313

PR #4275, opened 2020-05-26
POC Bisq gRPC for gRPC and REST clients

ghubstan:
This is an experiment to see how the Bisq daemon could work with REST clients.  

There is a single gRPC `CallService` that takes a random string (method name + params) from either gRPC or REST clients, and returns a gRPC (HTTP 2) or REST (HTTP 1.1 via HTTP proxy) response.  All method/parameter validation happens on the server.  

For REST clients, there is one `POST /v1/call` endpoint.  Service method names and parameters are sent in the HTTP request body, for example:

	curl -v -X POST http://host:port/v1/call -H "Authorization:xyz"  -H "Content-Type: application/json" -d '{"params": "method arg1 \"arg2\" arg3"}'

Like in the gRPC client/server implementation in the main branch, gRPC server authentication is as simple as possible -- no TLS.
	
Since there is no `grpc-web` or `grpc-gateway` Java implementation yet, I created a separate [Golang based HTTP proxy]( https://github.com/ghubstan/bisq-grpc-gateway) to forward REST requests.  To be explicit, REST requests must go through the [bisq-grpc-gateway](https://github.com/ghubstan/bisq-grpc-gateway) prototype.  Not having a Java/HTTP gateway may result in an immediate `NACK`, but `proxy.go` should never grow too complex, and it can be replaced when/if `grpc-web` for Java is released.  

One of the primary motivations for experimenting with this now has been @cbeams's discussion about the need to serve REST clients, and getting Bisq deployed in [myNode](https://mynodebtc.com).

Also included in this PR is a possible solution to the `server not available` problem.  The idea is that if the server cannot access a wallet's balance for any reason, it cannot do anything.  See commit 1011809.

Hopefully, commits examined in order best explain how this POC works.


ghubstan:
I was semantically wrong to characterise this gateway proxy as something that enables Bisq to work with RESTful clients.  So... oops.  Doing so was a time eater.  

What it does is expose one web endpoint mapped to the single gRPC service.  This single gRPC service performs all implemented Bisq functions, whether requests come from a browser (HTTP 1.1) or gRPC client (HTTP2).

Instead of using URL path elements to pass method names & parameters, use the request body.  That's not RESTful.  I was trying to combine most of the thoughts @cbeams recorded for the #grpc-api channel ( keybase://team/bisq/2020-05-18%20api%20next%20steps ) into one experiment:  just have a single pass-through gRPC service that takes method names + parameters, and expose the one gRPC service to RESTful web clients (but that's not a RESTful web client).

Right now (if using [go grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway)), I can only think of one way of implementing a single pass-through gRPC service that could serve gRPC and RESTful clients:

1. Use the POC's `CallService#call(String[] params)`, or something similar, to serve the gRPC client.  (Keep the client simple.)

2. For RESTful clients, define a gRPC service for each method, and use those gRPC services as pass throughs to the above gRPC `CallService#call(String[] params)`.

    `/method1/param1/val1` maps to gRPC service `Method1Service#call(param1, ..., paramN)`, `Method1` returns `CallService#call(String[] params)`

    `/method1/param1/val1/paramN/valN` maps to gRPC service `Method1Service#call(param1, ..., paramN)`, `Method1` returns `CallService#call(String[] params)`
  	
    `/method2/param/val` maps to gRPC service `Method2Service#call(param)`, `Method2` returns `CallService#call(String[] params)`
	
      etc...
	
[Go grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway) protos can be used to map multiple URLs with the same base path element to a single gRPC service, but unique URL base path elements need to be mapped to unique gRPC services.

I didn't do that because I wanted to see how exposing one gRPC service to gRPC and web clients might work.

PR #4280, opened 2020-06-03
Arbitrator witness signing

sqrrm:
This is a way for legacy arbitrators to repair some of the lost data in the signedwitness store. With ctrl+O from legacy arbitrator support view `SignUnsignedPubKeysWindow` is opened.

There is no change to current behavior of Bisq for a normal user in this PR.

Some more debug log shortcuts have been added as ctrl+shift+key under `PaymentAccountsView`

At the moment `SignSpecificWitnessWindow` is not that interesting, but it might be a good thing to have in the future. It's a way for arbitrators to sign specific `AccountAgeWitnesses`.

Some logging code from `AccountAgeWitnessService` was moved to a separate class.

`cleanSignedWitnesses` in `SignedWitnessService` is not yet in use, but it should probably be used soon. I don't want to enable it in this PR as it would change current behavior, and it also needs persistence to be fixed.


ripcurlx:
@sqrrm Could you please resolve the conflict? Thanks!

ripcurlx (approved):
ACK - tested this PR in conjunction with https://github.com/bisq-network/bisq/pull/4314

PR #4283, opened 2020-06-04
Warn when restoring wallet seedwords without date

### LINKED ISSUE #3766: Provide feedback when attempting to restore a wallet from seed words without specifying a date
Add feedback within the UI to inform/ask the user when they attempt to restore a wallet from seed words without specifying a date. At a minimum, inform them what happens when they dont specify a date, and inform them that any current Bisq wallet does not date older than the v0.5 release date.

This is in reponse to feedback received to #3749.

ghost:
When the user restores a BIP39 wallet from seed words there is an optional 'wallet date' field.  If the user does not enter that field the chain scan will be a lengthy process. Issue a popup message to the user in this case, informing them that the synch will take a long time and can be reduced by either specifying the wallet creation date or the date when Bisq introduced BIP39 seed words.

Fixes #3766


sqrrm (approved):
utACK

PR #4294, opened 2020-06-11
Add multiple fee receivers

sqrrm:
This is a rebased version of #4150 with @stejbac's patch applied.

I still worry about the filter distribution but it might be a minor issue compared to adding some kind of cutoff date and extra complexity that way. I have tested this and it seems to work. The only think I can't get to work locally is the propagation of remove filter. It could be an issue of applying recently received filters though, need to look a bit more at it.

@stejbac Do you have any other issues with this PR. Simple is good to adding more stuff indeed less desirable unless we really do need it.

### DIFF
--- a/core/src/main/java/bisq/core/filter/Filter.java
+++ b/core/src/main/java/bisq/core/filter/Filter.java
@@ -105,6 +105,10 @@ public final class Filter implements ProtectedStoragePayload, ExpirablePayload {
     @Nullable
     private final List<String> bannedSignerPubKeys;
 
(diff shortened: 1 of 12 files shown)


ripcurlx (approved):
ACK - tested it locally on Regtest with 40+ trades and the random and weighted selection seems to work. I also tested removing addresses from the filter after a couple of transactions. The only thing I recognized is an existing issue with the Filter propagation right now which is addressed in https://github.com/bisq-network/bisq/pull/4335.

PR #4296, opened 2020-06-12
Add wallets protection api tests

ghubstan:
(This PR should be merged after PR [4295](https://github.com/bisq-network/bisq/pull/4295).  Its branch is based on PR 4295's branch.)

This commit includes the following changes:

 * New tests for methods `lockwallet`, `unlockwallet`,
   `removewalletpassword`, and `setwalletpassword`.

 * New `getbalance` method error handing tests to verify
   error message correctness when wallet is locked.

 * Update to `getversion` method test -- now expects `1.3.4`.

 * Check for new `[params]` column header in help text.

This PR replaces [4243](https://github.com/bisq-network/bisq/pull/4243), which had not been kept in sync with commits to tested methods in [4214](https://github.com/bisq-network/bisq/pull/4214).  It also replaces [4268](https://github.com/bisq-network/bisq/pull/4268), which was based on a stale main branch.

Running the test suite will verify all tests pass

(1)  Clone and build branch `wallets-protection-api-tests`

      $ git clone -b wallets-protection-api-tests https://github.com/ghubstan/bisq.git wallets-protection-api-tests

      $ cd wallets-protection-api-tests

      $ ./gradlew clean build -x test

(2)  Run the server on Linux or OSX:  

      $ ./bisq-daemon --apiPassword=xyz --appDataDir=/tmp/newbisqdatadir

(3)  Let the server initialize (sync) for at least 2 minutes.

(4)  Run the bats test suite: 

      $ bats cli/test.sh

All tests should pass.

Issue #4257, opened 2020-05-14
Implement simplest possible trading script


ghubstan:
After implementing rpc wallet protection methods ([ISSUE 4214](https://github.com/bisq-network/bisq/issues/4198) + [PR 4214](https://github.com/bisq-network/bisq/pull/4214)), we need a  minimum set of endpoints to enable trading from the gRPC API, and a bats script for test / demo purposes.

The goal of the set of PRs that will accompany this issue is to allow a user to perform all of the following functions with the gRPC API  (the UI should not be needed for any setup).

- [x] Get Funding Addresses

  Method Signature:	`getfundingaddresses`

  Show existing unused BTC addresses in the wallet.
  If there are no unused addresses in the wallet, the server will create one.

  Analog to bitcoin-cli method [getnewaddress](https://bitcoincore.org/en/doc/0.19.0/rpc/wallet/getnewaddress) 

- [x] Get Address Balance

  Method Signature:	`getaddressbalance address`

  Show BTC balance for given address in the wallet, with number of confirmations of the most recent transaction included.  (TODO  Show list of txIds included as well?)

   Analog to bitcoin-cli method [listreceivedbyaddress](https://bitcoincore.org/en/doc/0.19.0/rpc/wallet/listreceivedbyaddress) 

- [x] Fund a BTC wallet

  No new method is needed to fund a wallet.  User sends BTC from an external wallet to a Bisq "Fund your wallet" address, then awaits confirmations.

  Use Case:
      * User finds an unused funding address by calling `getfundingaddresses`.
      * User sends BTC via `bitcoin-cli sendtoaddress "addr" amt`.
      * User polls status of the funding TX by calling `getaddressbalance address` and/or `getbalance`.

- [x] Create a Payment Account

  Method Signature:	`createpaymentacct acct-name paymentmethod acct-number currency`

  (The different payment methods will have different attributes, the above signature is specific to PerfectMoney)

  In the interest in keeping it simple at this stage, the 1st impl can only create a "PerfectMoney dummy" account, mimicking the default payment method created by the UI when running Bisq in regtest / dev mode with the DAO setup used to test every new release.

- [x] Get Payment Accounts

  Method Signature:	`getpaymentaccts`

  Show list of all payment methods created by the user in the UI or `createpaymentacct`.  The returned id is a required parameter of the `createoffer` method.
  
- [x] Get Existing Offers

  Method Signature:	`getoffers buy | sell, currency` 

[...]

PR #4299, opened 2020-06-13
Add rpc method 'getfundingaddresses'

ghubstan:
This addresses task 1 in issue [4257](https://github.com/bisq-network/bisq/issues/4257).

This new gRPC Wallet service method displays the BTC wallet's list of receiving addresses.  The balance and number of confirmations for the most recent transaction is displayed to the right of each address.  Instead of returning a gRPC data structure to the client, the service method returns a formatted String.

If the BTC wallet has no unused addresses, one will be created and included in the returned list, and it can be used to fund the wallet.

The new method required injection of the `BtcWalletService` into `CoreWalletsService`, and the usual boilerplate changes to `grpc.proto`, `CliMain`, and `GrpcWalletService`.

Some of the next PRs (for issue [4257](https://github.com/bisq-network/bisq/issues/4257)) will require some common functionality within `CoreWalletsService`.  These additional changes were made:

  * a private, class level `formatSatoshis` function
  * a public `getNumConfirmationsForMostRecentTransaction` method
  * a public `getAddressBalance` method
  * a private `getAddressEntry` method

A unit test that verifies a successful return status was added to `cli/test.sh`.

This PR should be reviewed/merged after PR [4296](https://github.com/bisq-network/bisq/pull/4296).

### DIFF
--- a/core/src/main/java/bisq/core/grpc/CoreWalletsService.java
+++ b/core/src/main/java/bisq/core/grpc/CoreWalletsService.java
@@ -19,7 +19,7 @@ import javax.annotation.Nullable;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 @Slf4j
-class CoreWalletService {
+class CoreWalletsService {
 
     private final Balances balances;
     private final WalletsManager walletsManager;
@@ -31,7 +31,7 @@ class CoreWalletService {
     private KeyParameter tempAesKey;
 
     @Inject
-    public CoreWalletService(Balances balances, WalletsManager walletsManager) {
+    public CoreWalletsService(Balances balances, WalletsManager walletsManager) {
         this.balances = balances;
         this.walletsManager = walletsManager;
     }
--- a/core/src/main/java/bisq/core/grpc/CoreWalletsService.java
+++ b/core/src/main/java/bisq/core/grpc/CoreWalletsService.java
@@ -1,21 +1,33 @@
 package bisq.core.grpc;
 
 import bisq.core.btc.Balances;
+import bisq.core.btc.model.AddressEntry;
+import bisq.core.btc.wallet.BtcWalletService;
 import bisq.core.btc.wallet.WalletsManager;
 
+import org.bitcoinj.core.Address;
+import org.bitcoinj.core.TransactionConfidence;
 import org.bitcoinj.crypto.KeyCrypterScrypt;
 
 import javax.inject.Inject;
 
 import org.spongycastle.crypto.params.KeyParameter;
 
+import java.text.DecimalFormat;
+
+import java.math.BigDecimal;
+
+import java.util.List;
+import java.util.Optional;
 import java.util.Timer;
 import java.util.TimerTask;
+import java.util.function.Function;
 
 import lombok.extern.slf4j.Slf4j;
 
 import javax.annotation.Nullable;
 
+import static java.lang.String.format;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 @Slf4j
@@ -23,6 +35,7 @@ class CoreWalletsService {
 
     private final Balances balances;
     private final WalletsManager walletsManager;
+    private final BtcWalletService btcWalletService;
 
     @Nullable
     private TimerTask lockTask;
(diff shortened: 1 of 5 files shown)


ghubstan:
**TODO** Re-format the result string as a table, e.g.  
```
Address      Balance          Confirmations
=======      ===========      =============
nM5Y...      ##.########                ###
```

core/src/main/java/bisq/core/grpc/CoreWalletsService.java
```diff
@@ -50,6 +75,68 @@ public long getAvailableBalance() {
...
+        Address address = getAddressEntry(addressString).getAddress();
+        return btcWalletService.getBalanceForAddress(address).value;
+    }
+
+    public String getFundingAddresses() {
+        if (!walletsManager.areWalletsAvailable())
+            throw new IllegalStateException("wallet is not yet available");
+
+        if (walletsManager.areWalletsEncrypted() && tempAesKey == null)
+            throw new IllegalStateException("wallet is locked");
```
dmos62, line 88: Checks of this type will probably be performed for most API endpoints. It would be nice to eventually see them moved into their own methods (`checkThatWalletsAreAvailable()`, etc.). Not requesting this change at this point.

core/src/main/java/bisq/core/grpc/CoreWalletsService.java
```diff
@@ -50,6 +75,68 @@ public long getAvailableBalance() {
...
+            var numConfirmations = getNumConfirmationsForMostRecentTransaction(a.first);
+            String addressInfo = "" + a.first
+                    + "  balance: " + format("%13s", btcBalance)
+                    + ((a.second > 0) ? ("  confirmations: " + format("%6d", numConfirmations)) : "")
+                    + "\n";
+            addressInfoBuilder.append(addressInfo);
+        });
+
+        return addressInfoBuilder.toString().trim();
+    }
```
dmos62, line 132: The Tuple3 feels unnecessary. Its main use seems to be to cache the balances, which are needed for checking if we need to generate a new address, otherwise we could get all that while formatting and a structure to hold these intermediary results wouldn't be necessary. Since the problem is basically caching, we can just do that explicitly by memoizing `getAddressBalance`:
```
var balances = CacheLoader.from(getAddressBalance);
// Usage:
var someBalance = balances.load(someAddress);
```
https://guava.dev/releases/20.0/api/docs/com/google/common/cache/CacheLoader.html

That will make the rest of the method simpler and shorter.

core/src/main/java/bisq/core/grpc/CoreWalletsService.java
ghubstan: How about `verifyEncryptedWalletIsUnlocked()` , throwing the exception if the wallet is encrypted, but not temporarily unlocked by a call to `unlockWallet(String password, long timeout)` ?

"Available" is taken;  it means the daemon is sufficiently warmed up and the btc & bsq wallets are not `null`. See `walletsManager.areWalletsAvailable()`.

Good call.  I'll create a new PR for this today.

dmos62 (changes requested):

core/src/main/java/bisq/core/grpc/CoreWalletsService.java
dmos62: What I see as a reusable and complex unit is not the predicate, but the pair of a predicate and the throwing of the exception. For example, checking if a wallet is locked and, if so, throwing an exception saying that it is locked. So the method I'm suggesting  would look like this:
```
private void aGoodName() {
  if (walletsManager.areWalletsEncrypted() && tempAesKey == null
    throw new IllegalStateException("wallet is locked");
}
```

Does that make sense?

ghubstan: I'm preparing to test this now:

```
   // Throws a RuntimeException if wallets are encrypted and locked.
    private void verifyEncryptedWalletIsUnlocked() {
        if (walletsManager.areWalletsEncrypted() && tempAesKey == null)
            throw new IllegalStateException("wallet is locked");
    }

```

ghubstan: Requested change is in PR [4312](https://github.com/bisq-network/bisq/pull/4312).

core/src/main/java/bisq/core/grpc/CoreWalletsService.java
ghubstan: Resolved in PR [4322]( https://github.com/bisq-network/bisq/pull/4322).

dmos62 (approved):
utACK. Some of the suggested changes have been introduced in later PRs.

sqrrm (approved):
ACK

PR #4301, opened 2020-06-14
Update install_collectd_debian.sh to work on Ubuntu 20.04 LTS

wiz:
@freimair 
* Use `sudo -H` instead of `sudo -H -i` since it seems sudo works differently now


devinbileck:
I was able to run the original script in Ubuntu 20.04 with `sudo -H -i`. What is the issue?

wiz:
@devinbileck not sure, I'll close for now, can re-open later if necessary

PR #4302, opened 2020-06-14
Add V3 seednodes wizseed3 and wizseed7, retire seednode ef5qnzx6znifo3df

wiz:
As part of the Tor V3 upgrade, and the effort to increase the number of
Bisq seednodes from 8 to 12, this PR adds 2 new V3 seednodes, and will
start the process to retire the old V2 seednode ef5qnzx6znifo3df.

The ef5qnzx6znifo3df seednode will continue operating for 2-3 months
during the retirement phase-out period, after which time we can filter
it out from the network for clients who have not upgraded if necessary.


sqrrm (approved):
utACK

PR #4305, opened 2020-06-14
Sign same name

sqrrm:
Accounts that can sign automatically sign other accounts with the same name. This allows users to add new bank accounts and get them signed without doing trades.

The logic is that it's the trustworthiness of the user that's verified with the signed witnesses.


ripcurlx (approved):
ACK - Tested on Regtest and Mainnet. Works as expected.

PR #4304, opened 2020-06-14
Add rpc method 'getaddressbalance'

ghubstan:
This PR addresses task 2 in issue [4257](https://github.com/bisq-network/bisq/issues/4257)  
	
This new gRPC Wallet service method displays the balance and number of confirmations for the most recent transaction for the given BTC  wallet address.

The new method required the usual boilerplate changes to `grpc.proto`, `CliMain`, and `GrpcWalletService`.

Two unit tests to check error msgs was added to `cli/test.sh`.

This PR should be reviewed/merged after PR [4299](https://github.com/bisq-network/bisq/pull/4299).

PR #4308, opened 2020-06-15
Add rpc method 'createpaymentacct'

ghubstan:
This addresses task 4 in issue [4257](https://github.com/bisq-network/bisq/issues/4257).
This PR should be reviewed/merged after PR [4304](https://github.com/bisq-network/bisq/pull/4304).

This new gRPC `PaymentAccounts` service method creates a dummy `PerfectMoney` payment account for the given name, number and fiat currency code, as part of the required "simplest possible trading API" demo. An implementation supporting all payment methods is not in the scope.

Changes specific to the new rpc method implementation are:

* New `createpaymentacct` method + help text was added to `CliMain`.
  Help text format specifiers were also changed to make room for larger method names and longer argument lists.

* The `GetPaymentAccounts` proto service def was renamed `PaymentAccounts` to avoid a name collision in the `grpc.proto` file, and the new rpc `CreatePaymentAccount` was made part of the newly named `PaymentAccounts` service def. 

* New `GrpcPaymentAccountsService` (gRPC boilerplate) and `CorePaymentAccountsService` (method implementations) classes were added.

* The gRPC `GetPaymentAccountsService` boilerplate code was moved from `GrpcServer` to the new `GrpcPaymentAccountsService` class, and `GrpcPaymentAccountsService` is injected into `GrpcServer`.

* A new `createpaymentacct` unit test was added to the bats test suite `cli/test.sh` (checks for successful return status code).

Maybe bit out of scope...  some small changes were made towards making sure the entire API is accessible from `CoreApi`, which is used in this PR as a pass-through object to the new `CorePaymentAccountsService`.  In the next PR, similar refactoring will be done to make `CoreApi` the pass-through object for all of the existing `CoreWalletsService` methods.  (`CoreWalletsService` will be injected into `CoreApi`.)  In the future, all `Grpc*Service` implementations should call core services through `CoreApi` for the sake of consistency.

### DIFF
--- a/cli/src/main/java/bisq/cli/CliMain.java
+++ b/cli/src/main/java/bisq/cli/CliMain.java
@@ -17,6 +17,7 @@
 
 package bisq.cli;
 
+import bisq.proto.grpc.GetAddressBalanceRequest;
 import bisq.proto.grpc.GetBalanceRequest;
 import bisq.proto.grpc.GetFundingAddressesRequest;
 import bisq.proto.grpc.GetVersionGrpc;
@@ -59,6 +60,7 @@ public class CliMain {
     private enum Method {
         getversion,
         getbalance,
+        getaddressbalance,
         getfundingaddresses,
         lockwallet,
         unlockwallet,
@@ -154,6 +156,16 @@ public class CliMain {
                     out.println(btcBalance);
                     return;
                 }
+                case getaddressbalance: {
+                    if (nonOptionArgs.size() < 2)
+                        throw new IllegalArgumentException("no address specified");
+
+                    var request = GetAddressBalanceRequest.newBuilder()
+                            .setAddress(nonOptionArgs.get(1)).build();
+                    var reply = walletService.getAddressBalance(request);
+                    out.println(reply.getAddressBalanceInfo());
+                    return;
+                }
                 case getfundingaddresses: {
                     var request = GetFundingAddressesRequest.newBuilder().build();
                     var reply = walletService.getFundingAddresses(request);
@@ -230,6 +242,7 @@ public class CliMain {
             stream.format("%-19s%-30s%s%n", "------", "------", "------------");
             stream.format("%-19s%-30s%s%n", "getversion", "", "Get server version");
             stream.format("%-19s%-30s%s%n", "getbalance", "", "Get server wallet balance");
+            stream.format("%-19s%-30s%s%n", "getaddressbalance", "", "Get server wallet address balance");
             stream.format("%-19s%-30s%s%n", "getfundingaddresses", "", "Get BTC funding addresses");
             stream.format("%-19s%-30s%s%n", "lockwallet", "", "Remove wallet password from memory, locking the wallet");
             stream.format("%-19s%-30s%s%n", "unlockwallet", "password timeout",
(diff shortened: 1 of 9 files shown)


cli/src/main/java/bisq/cli/CliMain.java
```diff
@@ -152,6 +160,46 @@ public static void run(String[] args) {
...
+
+                    var accountName = nonOptionArgs.get(1);
+
+                    if (nonOptionArgs.size() < 3)
+                        throw new IllegalArgumentException("no account number specified");
+
+                    var accountNumber = nonOptionArgs.get(2);
+
+                    if (nonOptionArgs.size() < 4)
+                        throw new IllegalArgumentException("no fiat currency specified");
```
dmos62, line 191: These size-based checks don't actually know which argument was ommitted. We might (or might not) go for more interesting interfaces in the future, but in this case we only know that not all arguments were passed.

cli/src/main/java/bisq/cli/CliMain.java
```diff
@@ -152,6 +160,46 @@ public static void run(String[] args) {
...
+
+                    var fiatCurrencyCode = nonOptionArgs.get(3).toUpperCase();
```
dmos62, line 193: `.toUpperCase()` seems misplaced here. It would soon get repetive. Whether the underlying logic differentiates between capitalizations is a low-level implementation detail and would do better at the lowest practical level.

dmos62 (changes requested):

cli/src/main/java/bisq/cli/CliMain.java
ghubstan: The original motivation for doing it this way was to keep it as simple as possible because this is/was _a reference implementation only_, to show others how to build their own gRPC CLI.  This isn't the final word on the issue, so let's leave it open.

cli/src/main/java/bisq/cli/CliMain.java
ghubstan: Agreed.  I will fix this.

ghubstan: This is fixed in commit 61285a7, part of PR [4329](https://github.com/bisq-network/bisq/pull/4329).

cli/src/main/java/bisq/cli/CliMain.java
dmos62: I'm not sure I follow. I'll try to rephrase. My concern here is that this behaviour is incorrect in some cases. Suppose you call `createpaymentacct` with arguments `account name`, `account number` (`fiat currency` omitted). It will output `"no fiat currency specified"`, which is correct. But, if you call it with arguments `account name`, `fiat currency` (`account number` omitted), it will output `"no fiat currency specified"`, which is incorrect. Or am I mistaken?

ghubstan: I will replace this code that assumes correct param order, with a simple argument count check.   The change will be made in the latest PR of this long PR chain:  PR [4329](https://github.com/bisq-network/bisq/pull/4329).

ghubstan: The requested change is in commit 9691f35 , part of PR [4329](https://github.com/bisq-network/bisq/pull/4329)

dmos62 (approved):

dmos62:
utACK

sqrrm (approved):
utACK

PR #4295, opened 2020-06-12
Rename CoreWalletService -> CoreWalletsService

ghubstan:
This change fixes the ambiguity in the original class name, which
implied it was a btc wallet service, not a bsq and btc wallets service.

PR #4309, opened 2020-06-15
Call core wallets service methods from CoreApi

ghubstan:
This change is a refactoring of the gRPC `Wallets` service for the purpose of making `CoreApi` the entry point to  all core implementations. 

Seeing `GrpcWalletsService` using `CoreWalletsService` (that's sensible) made me hesitate to make this change, but if `GrpcWalletsService` doesn't use `CoreApi` instead of `CoreWalletsService`, then we set up an inconsistent situation where some `Grpc*Service` impls use `CoreApi` and some do not.  I think it is better to be consistent for myself other devs, and consistently inject only `CoreApi` into each `Grpc*Service`, instead of a variable number of core services.

Also, the gRPC `Wallet` proto def name was changed to `Wallets` because this service manages BSQ and BTC wallets;  `GrpcWalletService` was renamed to `GrpcWalletsService` for the same reason.   (This change should have been made in PR [4295](https://github.com/bisq-network/bisq/pull/4295)).

This PR should be reviewed/merged after PR [4308](https://github.com/bisq-network/bisq/pull/4308). 
 (This PR's branch was created from the PR 4308 branch.)

### DIFF
--- a/cli/src/main/java/bisq/cli/CliMain.java
+++ b/cli/src/main/java/bisq/cli/CliMain.java
@@ -28,7 +28,7 @@ import bisq.proto.grpc.PaymentAccountsGrpc;
 import bisq.proto.grpc.RemoveWalletPasswordRequest;
 import bisq.proto.grpc.SetWalletPasswordRequest;
 import bisq.proto.grpc.UnlockWalletRequest;
-import bisq.proto.grpc.WalletGrpc;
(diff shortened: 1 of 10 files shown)


dmos62 (approved):

dmos62:
utACK

sqrrm (approved):
utACK

PR #4311, opened 2020-06-16
Increase price precision in Offer Book chart (altcoins)

### LINKED ISSUE #4251: Market Offer book graph needs higher precision x-axis values
### Description

In the Bisq app on the Market/Offer Book tab, the graph showing the current offers needs higher precision x-axis values to be able to distinguish the difference in price between various offers. 
#### Version

1.3.4
### Steps to reproduce

For currency combinations like XMR/BTC, the range of x-axis values at time of writing are 0.006, 0.006, 0.006, 0.006, 0.006, 0.006, 0.006, 0.007. These values are largely useless for reading the graph, and needs to be at least one more decimal place higher in precision to be able to read the offer prices.
### Expected behaviour

The x-axis has higher precision values, making it easy to determine the price of various offers on the graph at a glance.
### Actual behaviour

The user has no idea what the actual prices are, only that there is some difference, and has to use the mouse to the read the value from a pop-up dialog box.
### Screenshots

graph

#### Device or machine

Linux Mint 19.3

ghost:
In the Offer Book Chart, altcoins were displayed with prices having only 3 decimal places of precision.  This resulted in not being able to see any meaningful prices across the x-axis because they were all the same, e.g. ETH=0.025

Instead it should show 8 digits of precision for altcoins.  So for example ETH=0.02456914
Change the rule for display of altcoin Offer Book chart from:
    `if price begins 0.000 then display higher precision`
to:
    `if price begins 0.0 then display higher precision`
Fiat currency chart not affected by this rule.

Fixes #4251

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java
+++ b/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java
@@ -226,7 +226,7 @@ public class OfferBookChartView extends ActivatableViewAndModel<VBox, OfferBookC
                             final double doubleValue = (double) object;
                             if (CurrencyUtil.isCryptoCurrency(model.getCurrencyCode())) {
                                 final String withCryptoPrecision = FormattingUtils.formatRoundedDoubleWithPrecision(doubleValue, cryptoPrecision);
-                                if (withCryptoPrecision.equals("0.000")) {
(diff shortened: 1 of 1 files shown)


ghost:
XMR before
image
XMR after
image
ZEC before
image
ZEC after
image
LTC before
image
LTC after
image
ETH before
image
ETH after
image

sqrrm (approved):
utACK

PR #4313, opened 2020-06-16
Change "Move funds to Bisq wallet" to "Keep funds in Bisq wallet"

dcod3d:
Fixies: #4255

Comments from PR #4264 suggested only changing the displayed text rather than the tags. This PR just changed the text


ghost (approved):
uACK

sqrrm (approved):
utACK

PR #4314, opened 2020-06-16
Send signed witness to counterparty

sqrrm:
Signer sends SignedWitness to signee after trade is completed. This allows the signee to make sure their witness is properly published in case the signer isn't well connected. An extra measure to assure that signed witnesses are well distributed upon signing.

I made this on top of PR #4280 as it's touching a lot of the same code. That PR should be merged before this one.

### DIFF
--- a/common/src/main/java/bisq/common/util/Utilities.java
+++ b/common/src/main/java/bisq/common/util/Utilities.java
@@ -42,10 +42,6 @@ import javafx.scene.input.KeyCodeCombination;
 import javafx.scene.input.KeyCombination;
 import javafx.scene.input.KeyEvent;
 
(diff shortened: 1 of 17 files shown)


ripcurlx:
@sqrrm Please resolve the conflict - thanks!

sqrrm:
I think this merge conflict should be resolved once #4280 is merged since I fixed the same conflict there

ripcurlx (approved):
ACK

Tested it on regtest with both clients on the new version and a regression test with the buyer on the old version. During the regression test an error is logged that has no effect on the trade process as the new message type is not known by the client. So everything is working as expected.

sqrrm:
Travis issues, will create new PR

PR #4319, opened 2020-06-18
Update seednode/pricenode install scripts

devinbileck:
While deploying a new seednode instance on Ubuntu 20.04, I ran into a few issues with the install script and README. This PR addresses those issues.

### DIFF
--- a/pricenode/install_pricenode_debian.sh
+++ b/pricenode/install_pricenode_debian.sh
@@ -34,6 +34,11 @@ echo "[*] Upgrading apt packages"
 sudo -H -i -u "${ROOT_USER}" DEBIAN_FRONTEND=noninteractive apt-get update -q
 sudo -H -i -u "${ROOT_USER}" DEBIAN_FRONTEND=noninteractive apt-get upgrade -qq -y
 
+echo "[*] Installing Git LFS"
+sudo -H -i -u "${ROOT_USER}" curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash
+sudo -H -i -u "${ROOT_USER}" apt-get install git-lfs
+sudo -H -i -u "${ROOT_USER}" git lfs install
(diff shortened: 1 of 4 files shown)


cbeams (approved):
utACK. Thanks, Devin!

Before merging, I'd recommend at least one of the other @bisq-network/pricenode-operators try these changes out for themselves.

ripcurlx:
@wiz Could you have a look at this changes before I merge this based on https://github.com/bisq-network/bisq/pull/4319#pullrequestreview-433045580

wiz:
utACK
Devin, thanks for fixing this!

pricenode/install_pricenode_debian.sh
```diff
@@ -34,6 +34,11 @@ echo "[*] Upgrading apt packages"
 sudo -H -i -u "${ROOT_USER}" DEBIAN_FRONTEND=noninteractive apt-get update -q
 sudo -H -i -u "${ROOT_USER}" DEBIAN_FRONTEND=noninteractive apt-get upgrade -qq -y
 
+echo "[*] Installing Git LFS"
+sudo -H -i -u "${ROOT_USER}" curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash
```
wiz, line 38: Why is this needed? Doesn't ubuntu have a git-lfs package by default?

Emzy: I also think ```apt install git-lfs``` should work just fine. No need for external packages.

devinbileck: This is used to ensure the latest git-lfs package is installed. I don't recall now, but perhaps the latest isn't necessary for our needs and it should be fine to remove this line.
https://github.com/git-lfs/git-lfs/wiki/Installation#debian-and-ubuntu

PR #4323, opened 2020-06-19
Add rpc method 'getpaymentaccts'

ghubstan:
This PR addresses task 5 in issue [4257](https://github.com/bisq-network/bisq/issues/4257)  

This new gRPC PaymentAccounts service method displays the user's stored payment accounts.
A unit test to check a successful return status code was added  to cli/test.sh.

This PR should be reviewed/merged after PR [4322](https://github.com/bisq-network/bisq/pull/4322).

PR #4324, opened 2020-06-19
Return protos from funding address methods

ghubstan:
The `getaddressbalance` and `getfundingaddresses` methods now send new `AddressBalanceInfo` proto messages instead of a formatted String to the client.  The `AddressBalanceInfo` message contains address string, balance, and # of confirmations (transaction confidence) fields.

Changes include:
* A new `AddressBalanceInfo` proto message
* A wrapper class for the new `AddressBalanceInfo` proto
* New `getaddressbalance` and `getfundingaddresses` signatures in server
* `AddressBalanceInfo` display logic in client
* Removal of balance formatting logic in server
* Refactoring of balance formatting logic in client

This PR should be reviewed/merged after PR [4323](https://github.com/bisq-network/bisq/pull/4323).

The first cut of these methods methods were hacked because of uncertainty about whether or not we were going to ditch gRPC for an exclusively RESTful API.  But it seems likely we will serve both kinds of clients from the gRPC server.

### DIFF
--- a/cli/src/main/java/bisq/cli/CliMain.java
+++ b/cli/src/main/java/bisq/cli/CliMain.java
@@ -28,7 +28,7 @@ import bisq.proto.grpc.PaymentAccountsGrpc;
 import bisq.proto.grpc.RemoveWalletPasswordRequest;
 import bisq.proto.grpc.SetWalletPasswordRequest;
 import bisq.proto.grpc.UnlockWalletRequest;
-import bisq.proto.grpc.WalletGrpc;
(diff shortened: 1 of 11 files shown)


ghubstan:
@dmos62, can you take a look at this to see if it resolves your [requested change in PR 4304](https://github.com/bisq-network/bisq/pull/4304#pullrequestreview-434170634)?

dmos62 (approved):
utACK. I'm really liking these changes.

sqrrm (approved):
ACK

PR #4329, opened 2020-06-22
Add rpc method 'getoffers'

ghubstan:
The new method returns current buy or sell offers for a fiat ccy.

This PR should be reviewed/merged after PR [4324](https://github.com/bisq-network/bisq/pull/4324).

Changes include:

* New core.grpc classes
CoreOffersService
GrpcOffersService
model.OfferInfo

* CoreApi -- The new CoreOffersService is injected into CoreApi and
the old getOffers() and placeOffer() impls were moved into the
new CoreOffersService.  The getOffers implementation was re-done.
Other changes are just rearranging location of core method calls.

* GrpcServer -- The new GrpcOffersService replaced the old
GetOffersService and PlaceOfferService.

* grpc.proto -- The old GetOffers and PlaceOffer services were combined
into a single Offers service, and the PlaceOffer rpc was renamed
as CreateOffer.  These are the only substantive changes; the rest
is just rearranging location of the service defs in the file.
Also created a lighterweight OfferInfo proto message wrapper to
be passed between server & client (client has no access to core's
Offer and OfferPayload).

* OfferInfo -- A new wrapper around the OfferInfo proto message.

* CliMain -- The new GetOffers service stub was added.
Some (maybe too much) number and ccy formatting logic was
copied & modified from core.  Some tedius string formatting
was added too (needs to be tidied up).

* License comments were also copied to several classes, and I
made a mistake in reverting changes to the wrong file.

* Added unit tests to cli/test.sh.

### DIFF
--- a/core/src/main/java/bisq/core/grpc/CoreApi.java
+++ b/core/src/main/java/bisq/core/grpc/CoreApi.java
@@ -19,16 +19,12 @@ package bisq.core.grpc;
 
 import bisq.core.grpc.model.AddressBalanceInfo;
 import bisq.core.monetary.Price;
-import bisq.core.offer.CreateOfferService;
 import bisq.core.offer.Offer;
-import bisq.core.offer.OfferBookService;
 import bisq.core.offer.OfferPayload;
-import bisq.core.offer.OpenOfferManager;
 import bisq.core.payment.PaymentAccount;
 import bisq.core.trade.handlers.TransactionResultHandler;
 import bisq.core.trade.statistics.TradeStatistics2;
 import bisq.core.trade.statistics.TradeStatisticsManager;
-import bisq.core.user.User;
 
 import bisq.common.app.Version;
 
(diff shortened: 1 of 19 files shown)


core/src/main/java/bisq/core/grpc/CoreApi.java
```diff
@@ -48,35 +44,95 @@
...
+                price,
+                useMarketBasedPrice,
+                marketPriceMargin,
+                amount,
+                minAmount,
+                buyerSecurityDeposit,
+                paymentAccount,
+                useSavingsWallet,
+                resultHandler);
+    }
```
dmos62, line 122: It took me a minute to realise that the difference between these methods is that one of them takes primitives and the other takes higher level abstractions. I'm leaving this comment as an annotation for myself.

core/src/main/java/bisq/core/grpc/CoreOffersService.java
```diff
@@ -0,0 +1,138 @@
...
+        this.createOfferService = createOfferService;
+        this.offerBookService = offerBookService;
+        this.openOfferManager = openOfferManager;
+        this.user = user;
+    }
+
+    public List<Offer> getOffers(String direction, String fiatCurrencyCode) {
+        List<Offer> offers = offerBookService.getOffers().stream()
+                .filter(o -> !o.getDirection().name().equalsIgnoreCase(direction)
+                        && o.getOfferPayload().getCounterCurrencyCode().equalsIgnoreCase(fiatCurrencyCode))
```
dmos62: Would like to see these predicates moved into descriptive intermediate variables for the sake of readability.
```
o -> {
  var offerOfWantedDirection = !o.getDirection().name().equalsIgnoreCase(direction);
  var offerInWantedCurrency = o.getOfferPayload().getCounterCurrencyCode().equalsIgnoreCase(fiatCurrencyCode);
  return offerOfWantedDirection && offerInWantedCurrency;
}
```

core/src/main/java/bisq/core/grpc/CoreOffersService.java
```diff
@@ -0,0 +1,138 @@
...
+                             User user) {
+        this.createOfferService = createOfferService;
+        this.offerBookService = offerBookService;
+        this.openOfferManager = openOfferManager;
+        this.user = user;
+    }
+
+    public List<Offer> getOffers(String direction, String fiatCurrencyCode) {
+        List<Offer> offers = offerBookService.getOffers().stream()
+                .filter(o -> !o.getDirection().name().equalsIgnoreCase(direction)
```
dmos62: There is an inconsistency here. If we call with `direction = sell`, this will return offers with `direction = buy`. Not yet sure how to fix it.

dmos62 (changes requested):

core/src/main/java/bisq/core/grpc/CoreOffersService.java
ghubstan: Requested change is in commit 51d82b1.

core/src/main/java/bisq/core/grpc/CoreOffersService.java
ghubstan: Requested change is in commit 51d82b1.

dmos62 (approved):

dmos62 (changes requested):
Oh, I forgot that the locale/timezone changes are pending.

dmos62 (approved):
utACK

sqrrm (approved):
ACK

PR #4330, opened 2020-06-23
[WIP] Increase deposit to 30% for all trades

sqrrm:
This is an increase across the board for all trades. There has been some talk about keeping altcoins at 15%, I'm not sure. Perhaps we should have a smaller percentage deposit for trades over 1 BTC?

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/Restrictions.java
+++ b/core/src/main/java/bisq/core/btc/wallet/Restrictions.java
@@ -54,11 +54,11 @@ public class Restrictions {
     }
 
     public static double getDefaultBuyerSecurityDepositAsPercent() {
-        return 0.15; // 15% of trade amount.
+        return 0.30; // 30% of trade amount.
     }
 
     public static double getMinBuyerSecurityDepositAsPercent() {
-        return 0.15; // 15% of trade amount.
+        return 0.30; // 30% of trade amount.
     }
 
     public static double getMaxBuyerSecurityDepositAsPercent() {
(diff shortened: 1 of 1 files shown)


ripcurlx:
> This is an increase across the board for all trades. There has been some talk about keeping altcoins at 15%, I'm not sure. Perhaps we should have a smaller percentage deposit for trades over 1 BTC?

Do we know the rough percentage of future trades on altcoins vs. fiat? If fiat trades are our biggest issue it might make sense to start trying it out there and compare the refund agent cases afterwards to check if it helped.

leo816:
I agree with this measure. Hope that this will be incentive enough for users to conclude trades correctly

wiz:
Can you please link to the DAO proposal where it was discussed and decided to make this change? I searched a bit but couldn’t find it. 

My first thought after seeing this for the first time is that since this will lockup 30% more capital for security deposits, won’t it also have an effect of reducing Bisq trade volume of around 30%? Obviously this change would also significantly add more friction for Bisq traders, so aside from the reduced availability of trade capital, it may actually cause Bisq trade volume to drop by more than 30%.

sqrrm:
There is no DAO proposal that I know of. It's been discussed on keybase.

wiz:
I feel the DAO proposal process would be the appropriate way to propose, discuss, and decide changing such an important Bisq network parameter - raising 15% to 30% is a big change, let's talk about this first?

sqrrm:
Withdrawing this PR after discussion and instead proposing #4347

PR #4121, opened 2020-04-05
Dark mode, flat UI, alignment fixes

pedromvpg:
UI improvements and consistency pass:

- a darker Dark Mode 🦇
- reduce contrast in active button on main nav
- add hover states in main nav
- adjustments on UI to make the design more minimal and flat
  - remove shadows and gradients
  - remove bottom lines from fields
  - standardize colors of divider lines
- alignment issue fix in settings page in national currencies table
- alignment issue in slide toggles
- fix color disparity on support chat (dark mode)

Bisq_UI_Changes

Screen Shot 2020-04-05 at 6 54 58 PM

chat_arrows

PR #4334, opened 2020-06-24
Update Dark Theme to V2 with new CSS and images by Pedro

wiz:
* Increase overall darkness
* Clean up tables and increase background contrasts
* Add background color to disabled buttons
* Remove large padding on read only fields to fix Create Offer scroll
* Darken color of dark theme grey speech bubble
* Remove negative padding from toggle switches
* Add background to password field
* Closes #4121


ripcurlx (changes requested):
NACK - Tested a trade on Regtest and found this issue (see amount fields):

Bildschirmfoto 2020-06-25 um 13 38 43

m52go:
The red field background seems to stop abruptly. Is this intentional? It looks to me like it should fill the whole field:

Screenshot from 2020-06-26 12-55-08

There is missing padding on the bottom of the screen for funding a trade (see light mode on the right for comparison):

Screenshot from 2020-06-26 12-57-06

It seems there's extra padding on this screen for taking an offer (can scroll, but there is no need to scroll):

Screenshot from 2020-06-26 12-58-08

The grey used for the bubble body is different from the grey for the bubble tip. Also (to my eyes) the blue looks out-of-place:

Screenshot from 2020-06-26 13-00-34

pedromvpg:
Corrected all of this except the invalid fields on Create Order.

Impossible to do without changing something in the java code. 
The outer element needs to be able to detect that the element inside of it is invalid.

Group 11


Also made other small adjustments like making the theme generally darker and revised a bit of the colors on the header (they had some kind of leftover blueish gray).

Group 10

Adjustments here
https://github.com/pedromvpg/bisq/tree/dark-theme-v2

ripcurlx (changes requested):
NACK - slider for fee is not positioned correctly

Bildschirmfoto 2020-07-03 um 21 22 26

wiz:
@ripcurlx AFAIK that issue was already fixed, can you double check you are running the correct commit?

m52go:
Running 82baa86b163ff23dfa7f45493b04b1b2ab557bfb...I'm seeing it correctly positioned.

dark

ripcurlx:
Probably my bad. I checked it out clean once again and now it seems to be correct.

ripcurlx (approved):

PR #4335, opened 2020-06-25
Add equals check to prevent resetting of newly set Filter

ripcurlx:
While testing #4294 I recognized that when adding a new filter object to the network one of my regtest nodes got its filter reset. This was caused by a timing issue that the added filter message was received before the removed filter message. In that case the filter was reset and the new filter never applied. This is only a problem during runtime when a new filter is set. After a restart the correct filter is applied. This PR adds a equals check to prevent the reseting of a filter by mistake. As the added filter message overwrites all fields anyways not having the removed filter message is not essential.

### DIFF
--- a/core/src/main/java/bisq/core/filter/FilterManager.java
+++ b/core/src/main/java/bisq/core/filter/FilterManager.java
@@ -17,7 +17,6 @@
 
 package bisq.core.filter;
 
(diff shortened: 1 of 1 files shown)


sqrrm (approved):
utACK

PR #4339, opened 2020-06-26
Upgrade grpc & gson dependencies

ghubstan:
`gRPC` from v1.25.0 to v1.30.2        (commit msg incorrectly says upgrade was to v1.30.3)
  `gson` from v2.8.5  to v2.8.6


sqrrm:
I'm inclined to approve of these library upgrades since they're google libraries and thus less likely to include malicious code. That's the policy we have had previously. Would be good to have some more opinions on this though.

ghubstan:
> Would be good to have some more opinions on this though.

I created the PR so I'm sure my opinion is not what you need here.  But I want you to know I have not seen any protobuf related problems while simultaneously running `:seednode`, `:desktop` (as arb), `:daemon` and `:cli` on Linux, which I frequently do as part of my work on a new `:apitest` subproject.  (I mention this because the transitive dependency `com.google.protobuf:protobuf-java` was also upgraded in this commit.)

chimp1984:
> I'm inclined to approve of these library upgrades since they're google libraries and thus less likely to include malicious code. That's the policy we have had previously. Would be good to have some more opinions on this though.

I agree but I think any update should include a "justification" as it adds some risk. If the update is not required, we should be conservative IMO.

ghubstan:
I had no need to push this into release v1.3.6, but I would like to get this into master after more analysis and experimentation with the changes by myself and others -- before I create another long series of PR + branches for the API project.

And @chimp1984 is right, this change involves a number of upgrades to transitive dependencies that impact Bisq;  they need more careful examination and consensus among core devs.  I've done some of that research today, but I'm going back to work on testing release v1.3.6 now.

I will add more detail about this matter after release testing is done.

ghubstan:
I'm closing this PR because it would force an upgrade of too many important transitive dependencies.  I am convinced it would not create new problems because I have done 2+ weeks of testing this change on desktop and API, but other developers do not have enough time to review and test the changes themselves.  

I think it is important to build the gRPC based API on the latest grpc-* libraries, and will push for this change again at a later stage of API development, but not now.

chimp1984:
@ghubstan Thanks for your considerations! The main issue I see is not that any updated dependency could break something, but that it was used as attack to introduce some vulnerability. You know probably about the hack of some exchanges due some nodes dependencies which got exploited by a hacker....

dmos62:
@chimp1984 we were talking with @ghubstan yesterday and I came to think that this is about whether we do the updates early or postpone them. Provided we're talking about updates that don't provide new functionality that's needed at the moment. As a counter-example, updating Guava would provide new utilities that would translate to some additional functionality, as compared to updating GSON or gRPC, which is mostly about transparent things like bugfixes or optimizations.

So thinking that way, we postpone updating GSON and gRPC. That makes sense, in that it's not an essential update right now, and by being conservative we expose ourselves less to malicious updates, as you pointed out. There's also a downside, in that postponed upgrades accumulate a bit like debt and when we finally do upgrade, the batches of upgrades will be larger and thus will bear more risk.

The conclusion I'm coming to is that if there are two extremes in here, rejecting every non-essential upgrade vs. accepting any upgrade, the optimal position is somewhere between the two. It's a balancing act. And, that said, this upgrade proposal seems a pretty safe bet. So is it a good time to make this upgrade? It could be.

I've spelled out my thinking about upgrade policy, because I'm still figuring out what to think about it and what the consesus is.

PR #4341, opened 2020-06-28
Remove @cbeams as build and pricenode code owner

cbeams:
This is to avoid automatically requesting a review from me on all PRs that touch the Gradle build and pricenode codebase, e.g. as just happened in #4339.


sqrrm (approved):
utACK

PR #4343, opened 2020-06-29
Clean up TOR status icon

pedromvpg:
Removed some left-over white pixels from the tor status icon that were still visible from a past white background circle.

image


ripcurlx (approved):
utACK

ripcurlx:
@pedromvpg You need to sign your commit. Otherwise I'm not able to merge it. Thanks!

PR #4346, opened 2020-06-30
Fix tor v3 address validation

ghost:
The regex validation for addresses was not accepting tor v3 format.  This patch allows tor v3 addresses which are always 56 chars in length (whereas tor v2 addresses are always 16 chars in length).  Tests have been included.

before, does not allow tor v3 address:
image

after, allows tor v3 address:
image


ripcurlx (approved):
utACK - Thanks! 👍

PR #4349, opened 2020-07-01
Improve account limits popup text

m52go:
Changed docs link to wiki link, and do some minor wording tweaks.

Screenshot from 2020-07-01 14-48-44


ripcurlx (approved):
utACK

PR #4350, opened 2020-07-01
Improve zelle payment account info text

m52go:
This PR updates the popup with Zelle guidance. The list of banks currently in the program no longer applies...Zelle now works with hundreds of banks, and there's now a website with more thorough guidance.

Other notable changes to address common questions:
- point out that transfer limits vary by bank, so users should check
- mention that zelle can be used without a bank, but limits are much lower
- specify that names MUST match

Screenshot from 2020-07-01 14-53-39


ripcurlx (approved):
utACK

PR #4353, opened 2020-07-03
Signer send witness to signee

sqrrm:
Same as #4314


ripcurlx (approved):
ACK

Tested it on regtest with both clients on the new version and a regression test with the buyer on the old version. During the regression test an error is logged that has no effect on the trade process as the new message type is not known by the client. So everything is working as expected.

PR #4354, opened 2020-07-06
Fix taker use all BSQ for fee payment

sqrrm:
Fixes https://github.com/bisq-network/security/issues/4

Issue: if a taker used exactly all BSQ from the BSQ inputs to pay the
trading fee, there was no BSQ change in the takeOfferFeeTx. It was
assumed that the second output was the reservedForTrade output, but in
the case of missing BSQ change it was the first output.

Fix: added a check to make sure the value of the inputs to the deposit
tx match the expected inputAmount.

Added a check that if there is no BSQ outputs in the bsqTradingFeeTx a
change output is added of value 1 satoshi more than the BSQ input value.
This ensures that the second output is always the reservedForTrade
output. It also ensures that the BSQ is burnt, even in the very unlikely
case that the amount of BSQ burnt is larger than the reservedForTrade
amount.

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
@@ -694,6 +694,7 @@ public class BsqWalletService extends WalletService implements DaoStateListener 
         coinSelection.gathered.forEach(tx::addInput);
         try {
             Coin change = bsqCoinSelector.getChange(fee, coinSelection);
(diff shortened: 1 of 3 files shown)


ripcurlx (approved):
ACK

I tested the use-case described on regtest spending all remaining BSQ in my wallet and everything worked as expected. I also tested buyer as taker with BTC and BSQ fee without spending all BSQ and all worked as expected as well.

PR #4355, opened 2020-07-06
Increase contrast of date picker colors in Dark Theme v2

wiz:
Before:
Screen Shot 2020-07-07 at 3 18 39

After:
Screen Shot 2020-07-07 at 3 28 24

@pedromvpg are you sure this is the correct CSS? it doesn't look right


ripcurlx (approved):
utACK

PR #4365, opened 2020-07-09
Fix NullPointerException with external tor usage

ddawson:
If `torControlPort` is specified, but neither `torControlPassword` nor `torControlCookieFile` are specified, we have `cookieFile == null` in `bisq.network.p2p.network.RunningTor`, but `RunningTor.getTor()` assumes a cookie file has been specified and tries to check that the file exists, causing the thread to crash. Added a check for `null` to fix this.


mrosseel (approved):
utAck

ripcurlx (approved):
utACK

PR #4366, opened 2020-07-09
Add API test harness for Linux & OSX

ghubstan:
This PR adds a new `:apitest` subproject to Bisq with the following goals:

* Enable test driven development of the API project
* Automate end to end testing of the API 
* Simplify `:desktop` testing in regtest/full-dao mode

It requires the installation of bitcoin-core v0.19 or v0.20 binaries in the $PATH, or the path to them defined in the subproject's config file `src/main/resources/apitest.properties`:

    bitcoinPath=/home/you/bitcoin/v0.19.1/src
	

The `:apitest` build downloads and installs dao-setup files for `bitcoind`, bob & alice datadirs in `apitest/build/resources/main`, and test case `@BeforeAll` methods start `bitcoind`, `seednode`, `arbitration` node (`:desktop` or `:daemon`), plus bob & alice nodes (`:desktop` or `:daemon`) as Linux background processes.  Currently, the only testing mode supported is regtest / full-dao mode, without Tor.

To run a full clean build, and install dao setup files:

    ./gradlew clean build :apitest:installDaoSetup

To install or re-install dao setup files:

    ./gradlew :apitest:installDaoSetup

To run API `@Test` cases:

    ./gradlew :apitest:test -DrunApiTests=true

A gradle Test Summary containing any initialization errors and test failures can be found in

    apitest/build/reports/tests/test/index.html

This test harness works on a 2014 Mac mini with a Dual-Core i5 and 8 Gb RAM.

`[EDITED  27-July-2020]`


wiz:
Does this require adding new jar deps?

ghubstan:
> Does this require adding new jar deps?

`[EDITED 19-July-2020]`

The `org.junit.jupiter` dependendency was upgraded to v5.6.2 -- only in the `:apitest:test` dependency definitions, nowhere else in the bisq main or test dependency definitions.  The upgrade is necessary for running ordered tests.

build.gradle
```diff
@@ -143,6 +144,16 @@ configure([project(':cli'),
                     'DEFAULT_JVM_OPTS=""', 'DEFAULT_JVM_OPTS="-XX:MaxRAM=4g"')
             }
 
+            if (applicationName == 'apitest') {
+                // Give apitest more MaxRAM, and pass the logback config file as a system
+                // property to avoid chatty logback startup due to multiple logback.xml
+                // files in the classpath (:daemon & :cli).
+                def script = file("${rootProject.projectDir}/bisq-$applicationName")
+                script.text = script.text.replace(
+                    'DEFAULT_JVM_OPTS=""', 'DEFAULT_JVM_OPTS="-XX:MaxRAM=6g' +
```
dmos62: 6GB MaxRAM gives me pause.

- Would `ErgoHeapSizeLimit` [0] be a more direct way to limit RAM usage? `MaxRAM` is meant to specify real RAM size (all available memory).
- In the comment it says that we're increasing RAM allotment, so we're doing this because otherwise MaxRAM would be set to something lower?
- What if we're running this on a memory restricted environment (< 6GB RAM)?
- Is this to do with bisq using too much RAM in some cases? But wasn't that only relevant for long-running processes?

[0] http://hg.openjdk.java.net/jdk9/jdk9/hotspot/file/b756e7a2ec33/src/share/vm/runtime/globals.hpp#l2034

apitest/src/main/java/bisq/apitest/linux/BitcoinDaemon.java
```diff
@@ -0,0 +1,91 @@
...
+    public BitcoinDaemon(ApiTestConfig config) {
+        super("bitcoind", config);
+    }
+
+    @Override
+    public void start() throws InterruptedException, IOException {
+        String bitcoindCmd = "export LD_LIBRARY_PATH=" + config.berkeleyDbLibPath + ";"
+                + " " + config.bitcoinPath + "/bitcoind"
+                + " -datadir=" + config.bitcoinDatadir
+                + " -daemon";
```
dmos62: `export LD_LIBRARY_PATH=` doesn't seem to be necessary with my distribution of `bitcoind`. Also, `config.bitcoinPath` seems to expect to be pointed to the `bin` sub-directory of bitcoin-core's home (the one that contains the `bitcoind` executable). That's a bit unexpected. I think it's more intuitive for it to point to the executable directly, or alternatively to bitcoin-core's home directory.

dmos62 (changes requested):

apitest/src/main/java/bisq/apitest/linux/BitcoinDaemon.java
ghubstan: `export LD_LIBRARY_PATH` is necessary for my bitcoin-core installation.  This issue is something we'll need to sort out for an unknown variety of setups.   

I think `config.bitcoinPath` should point to the directory containing bitcoin-core binaries, not the `bitcoind` and `bitcoin-cli` binaries themselves -- requiring two `apitest` config options instead of one.   On the other hand, my `bitcoin-core` binaries are isolated in my `bitcoin-core` src directory (I never ran `make install`), not a default system `bin` directory where many unrelated executables are installed.  Let's keep discussing...

ghubstan: You no longer need to configure a `BerkeleyDb` lib path if your bitcoin-core binary is statically linked to `BerkeleyDb`.  The `berkeleyDbLibPath` option now defaults to an empty string.  If not set, no `export LD_LIBRARY_PATH` will not be pre-pended to the `bitcoind` command.

apitest/src/main/java/bisq/apitest/config/BisqAppConfig.java
```diff
@@ -0,0 +1,123 @@
...
+            8888,
+            5123,
+            9999),
+    bobdesktop("bisq-BTC_REGTEST_Bob_dao",
+            "bisq-desktop",
+            "\"-XX:MaxRAM=4g -Dlogback.configurationFile=apitest/build/resources/main/logback.xml\"",
+            BisqAppMain.class.getName(),
+            8888,
+            5123,
+            -1);
```
dmos62, line 84: How do these MaxRAM settings relate to our discussion about similar settings in build.gradle?

ghubstan: Limiting RAM for `bisq-apitest` was not necessary -- and removed from the `bisq-apitest` script -- because it is a headless app.  But we need to limit RAM for `:desktop` apps started by `bisq-apitest`, like we do for the Linux Bisq UI.

Above, you mentioned:
>> Is this to do with bisq using too much RAM in some cases? But wasn't that only relevant for long-running processes?

The Bisq UI is using too much without this limit, especially when starting it on an empty datadir.  This is why `BisqAppConfig` has a larger limit for `bobdesktop` than `bobdaemon`.  It's not really necessary for daemons at this stage, but it might be useful for profiling long running tests in the future.

apitest/src/main/java/bisq/apitest/Scaffold.java
```diff
@@ -0,0 +1,445 @@
...
+    }
+
+    private void installBitcoinConf() {
+        // We write out and install a bitcoin.conf file for regtest/dao mode because
+        // the path to the blocknotify script is not known until runtime.
+        String bitcoinConf = "\n"
+                + "regtest=1\n"
+                + "[regtest]\n"
+                + "peerbloomfilters=1\n"
+                + "rpcport=18443\n"
```
dmos62: rpcport is configured in ApiTestConfig, but here it's hardcoded.

ghubstan: Oops.  Fixed it and two other hardcoded configs in commit cf3b545.

mrosseel:
some small questions from my side:
- how hard is it to make it cross-platform? Just converting the .sh scripts or something more?
- how hard is it to integrate this into travis ?

dmos62:
@mrosseel native Windows compatibility would take a fair amount of work and was a non-goal from the start. Running Linux environments under Windows is getting easier every day though, so I'm sure one could find a way around this without significant modifications.

ghubstan:
> * how hard is it to make it cross-platform? Just converting the .sh scripts or something more?

1-2 days of work, I estimate -- for someone with a Windows machine.   Right now, there are 22 in-line bash commands executed from Java (only 1 separate `.sh` script ).   To work on Windows, the `bisq.apitest.linux.BashCommand` would need to be ported to `bisq.apitest.windows.WindowsCommand`, and 22 `BashCommands` would need to be ported to `WindowsCommands`.  

 I would be surprised if this did not already work on OSX if bash is on the system. 

> how hard is it to integrate this into travis ?

`[EDITED 19-July-2020]`

Right now, travis would need to run a normal build + a task to download & extract dao-setup.zip files:

    ./gradlew clean build :apitest:installDaoSetup

Then run API `@Test` cases:

    ./gradlew :apitest:test -DrunApiTests=true

ghubstan:
> @mrosseel native Windows compatibility would take a fair amount of work and was a non-goal from the start. Running Linux environments under Windows is getting easier every day though, so I'm sure one could find a way around this without significant modifications.

@dmos62 may be more realistic in his estimation of the work effort due to the differences between linux bash and windows system commands.

apitest/src/main/java/bisq/apitest/scenario/FundWalletScenarioTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+        // Given the default tx fee rate, we want to send 2.5 + the fee,
+        // so the new wallet balance will be exactly 12.5 btc.
+        // We should calculate the fee based on the fee rate and tx size
+        // instead of hard coding the fee amount.
+        double btc = parseDouble("2.5") + parseDouble("0.0000336");
+        bitcoinCli.sendToAddress(unusedAddress, btc);
+
+        bitcoinCli.generateBlocks(1);
+        sleep(1500);
```
dmos62: `bitcoind` has the `blocknotify` callback for this. Not really a problem now, but something to consider if the number of "sleeps" grows cumbersome.

apitest/src/main/java/bisq/apitest/method/GetBalanceTest.java
```diff
@@ -0,0 +1,68 @@
...
+            fail(ex.getMessage());
+        }
+    }
+
+    @Test
+    @Order(1)
+    public void testGetBalance() {
+        var balance = grpcStubs.walletsService.getBalance(GetBalanceRequest.newBuilder().build()).getBalance();
+        assertEquals(1000000000, balance);
+    }
```
dmos62: The wallet being funded with 10 BTC is an implicit step in the setup, which feels wrong. Preferably it would be explicit. I couldn't track down where it happens. Comment?

apitest/src/main/java/bisq/apitest/scenario/FundWalletScenarioTest.java
ghubstan: This will become a nuisance, but right now I don't know of another way to give `:daemon` some time to parse new blocks.  (The callback does not reach the `:cli`.)

dmos62: Oh right, we're waiting for `:daemon`, not `bitcoind`. `:daemon` will probably have a gRPC endpoint for getting notifications about funding events.

apitest/src/main/java/bisq/apitest/scenario/FundWalletScenarioTest.java
```diff
@@ -0,0 +1,78 @@
...
+        long balance = getBalance();  // bisq wallet was initialized with 10 btc
+        assertEquals(1000000000, balance);
+
+        String unusedAddress = getUnusedBtcAddress();
+
+        // Given the default tx fee rate, we want to send 2.5 + the fee,
+        // so the new wallet balance will be exactly 12.5 btc.
+        // We should calculate the fee based on the fee rate and tx size
+        // instead of hard coding the fee amount.
+        double btc = parseDouble("2.5") + parseDouble("0.0000336");
```
dmos62: We could use the `subtractfeefromamount` parameter set to false to not have to calculate a fee.

apitest/src/main/resources/bitcoin.conf
```diff
@@ -0,0 +1,16 @@
...
+
+server=1
+txindex=1
+
+# TODO migrate to rpcauth
+rpcuser=apitest
+rpcpassword=apitest
+
+# The blocknotify path will be defined and at runtime.
+# blocknotify=bash ~/.bitcoin/blocknotify %
```
dmos62: It might be worth it to not use a bitcoin.conf at all and pass all parameters inline, like this: https://gist.github.com/dmos62/20981dae799e2fc31b62dd443c44c7ae#file-tests-bats-L139

Since your setup already works, this is low priority.

apitest/src/main/java/bisq/apitest/method/GetBalanceTest.java
ghubstan: It is an important part of the `dao-setup` for regtest/dev mode testing.     Bob & Alice are given initial BTC & BSQ balances and dummy payment accounts so they can trade without coders / testers having to go through the motions of funding wallets from `bitcoin-cli -regest`, setting up payment accounts, etc.  Having to do that would make manual  release testing even more time consuming that it already is.  

There are more details [here](https://github.com/bisq-network/bisq/blob/master/docs/dao-setup.md).

Eventually, we want to create and test the [DAO setup environment](https://github.com/bisq-network/bisq/blob/master/docs/dao-setup.md#setup-a-custom-dao-genesis-transaction), but I don't think that's in scope for quite awhile.  For now, we use the pre-fab [dao-setup.zip](https://github.com/bisq-network/bisq/blob/master/docs/dao-setup.zip) -- what sets up the pre-funded wallets for Bob & Alice.

apitest/src/main/java/bisq/apitest/scenario/FundWalletScenarioTest.java
ghubstan: This is fixed in commit 5d7133a.

apitest/src/main/resources/bitcoin.conf
ghubstan: Passing all `bitcoind` parameters inline is cleaner.  It's one less file the test harness has to write & install, and the `bitcoind` configuration is clearly visible in the 'apitest' logfile.   Fixed in commit f7d8c0e.

apitest/src/main/java/bisq/apitest/method/GetBalanceTest.java
dmos62: Do you think it would make automated tests more time consuming as well. Feels dirty not starting with a clean slate. As an alternative, we could be diligent about commenting in the test suites what the initial state is.

apitest/src/main/resources/blocknotify
```diff
@@ -0,0 +1,20 @@
...
+echo $1 | nc -w 1 127.0.0.1 5121
+
+# Alice Node listens on port 5122
+echo $1 | nc -w 1 127.0.0.1 5122
+
+# Bob Node listens on port 5123
+echo $1 | nc -w 1 127.0.0.1 5123
+
+# Some other node listens on port 5124, etc.
+# echo $1 | nc -w 1 127.0.0.1 5124
```
dmos62, line 20: Not requesting this change, but note that you probably can pass a blocknotify script to bitcoind inline, like this: https://gist.github.com/dmos62/20981dae799e2fc31b62dd443c44c7ae#file-tests-bats-L150 (untested).

ghubstan: I had thought of that;  we could pass an inline script for notifying only the ports being used by the scaffolding to avoid errors in the bitcoind log.  It's on the TODO list.

apitest/src/main/java/bisq/apitest/method/GetBalanceTest.java
ghubstan: I don't know what the cost would be until I could compare.  There would be a trade off because it would skip the dao-setup files refresh in `@BeforeClass`.

I added a javadoc comment explaining why Bob & Alice have non-zero initial BTC & BSQ balances and a dummy payment account before tests start.  (See commit b4d3ea7.)

dmos62 (dismissed):
ACK.

We had a lot of back and forth over this PR. At 3k lines the PR isn't small. Part of the complexity is the shell-level interaction between bitcoind, bitcoin-cli and Bisq daemons. I had some reservations about doing everything in Java, since that required implementing shell-type functionality. I was floating an idea to use Java routines for things that require Java API access and Bash routines for the rest (like setup, teardown and e2e tests), hoping that that would cut down on the boilerplate (and would be possible at all). I wrote up a [Bash e2e testing harness](https://gist.github.com/dmos62/20981dae799e2fc31b62dd443c44c7ae) for comparison, but it was moot point, because ghubstan's PR was almost done by then. Some of the ideas there were implemented in this PR. It seems to work pretty well right now. Some kinks may (or may not) appear when the number of tests gets high, but we'll cross that bridge when we get there.

Thank you @ghubstan for the work!

ripcurlx:
@sqrrm Could you please have a look at this PR?

ghubstan:
@dmos62 , sorry for the late check-in (dismissing your ACK), but it was important.   See commit 27ee4b8.

ghubstan:
> semantic duplication

Fixed in commit 8bb7e12

dmos62 (dismissed):
ACK

apitest/src/main/java/bisq/apitest/linux/BitcoinDaemon.java
```diff
@@ -55,6 +55,7 @@ public void start() throws InterruptedException, IOException {
                 + " -txindex=1"
                 + " -peerbloomfilters=1"
                 + " -debug=net"
+                + " -fallbackfee=0.0002"
```
dmos62, line 58: This change seems to change the behaviour of bitcoind in mainnet mode. Before 0.20.0 bitcoind had different default `fallbackfee`s for mainnet and test modes (0 and 0.0002 respectively), which was then fixed by setting the defaults to 0 in all cases.

By setting `fallbackfee=0.0002` we're setting it on all net modes. I can't tell if that's a problem.

dmos62: Oh wait. This bitcoind will never run in mainnet, will it? Then the concern is that the apitest testing harness will have a different bitcoind environment than a production daemon would.

Again, I don't fully understand the significance of the `fallbackfee` option. It's used when the bitcoind can't estimate a "floating fee", right? And it only applies when sending funds from a wallet managed by the bitcoind instance, right? If both answers are yes, then I don't see this causing differences between prod and dev environments, but I'd like someone else to sanity check this too.

ghubstan: This new fallbackfee setting would change the behaviour of bitcoind in mainnet mode -- from the default 0.00 to 0.0002.  But `apitest` cannot run on mainnet unless the code is hacked, and `apitest` should never run on mainnet -- it should never spend mainnet BTC, do a trade on mainnet, etc.

>> Then the concern is that the apitest testing harness will have a different bitcoind environment than a production daemon would.

Yes, it does mean that the `apitest` harness will have a different `bitcoind` environment than a production daemon would, but that is what we want.  The bitcoin-core change ([PR 16524](https://github.com/bitcoin/bitcoin/pull/16524)) was to make the fallbackfee the same on all chains -- to fix bitcoin-core's inconsistency.  It was 0 (disabled) on mainnet, but 0.0002 for testnet/regtest chains.  Now (v0.20+), it is the same for all chains: 0 (disabled) by default for all chains.

We want bitcoin-core v0.20 (regtest) to behave like it did before the bitcoin-core PR 16524 change.

In the context of `apitest`, the fallbackfee affects the behaviour of the bitcoin-cli `sendtoaddress` command, where we use the `subtractfeefromamount=false` parameter.  (See `FundWalletScenarioTest#testFundWallet()`)  This `bitcoin-cli` command errors if the fallbackfee is disabled on bitcoin-core v0.20.  Enabling it on `bitcoind` ensures this command behaves the same way it does on v0.19.

apitest/src/test/java/bisq/apitest/method/BitcoinCliHelper.java
```diff
@@ -40,9 +39,12 @@ public BitcoinCliHelper(ApiTestConfig config) {
 
     public String getNewBtcAddress() {
         try {
-            String newAddress = new BitcoinCli(config, "getnewaddress").run().getOutput();
-            assertNotNull(newAddress);
-            return newAddress;
+            BitcoinCli newAddress = new BitcoinCli(config, "getnewaddress").run();
+
+            if (newAddress.isError())
+                fail(format("Could generate new bitcoin address:%n%s", newAddress.getErrorMessage()));
```
dmos62: Typo? "Could _not_ generate ..."?

ghubstan: Oops.  Fixed in commit 9637cc0.

sqrrm:
Finally managed to get the tests to pass. There is a conflict with current running bitcoin core processes it seems. Once my local bitcoin-qt regtest instance was shut down the tests passed.

ghubstan:
> Finally managed to get the tests to pass. There is a conflict with current running bitcoin core processes it seems. Once my local bitcoin-qt regtest instance was shut down the tests passed.

I intend to fix this in one of the next PRs.  

When `apitest` cannot start `bitcoind` because a host is already running one or more bitcoin-core procs -- bitcoind, bitcoin-qt, ... -- an error message describing the problem and workaround should be loggged.  Passing alternate port params to bitcoind may solve this.  New ApiTestConfig options for alternate bitcoind parameters might also be necessary.

ghubstan:
> There is a conflict with current running bitcoin core processes it seems. Once my local bitcoin-qt regtest instance was shut down the tests passed.

Commit 72ff4dc resolves the `-rpcport` conflict, but the bind address:port conflict is still unresolved because `bitcoinj` hardcodes the regtest bind address:`port` in `RegTestParams`.  Commit e88bdb9 adds a doc explaining how to work around the conflict by starting your bitcoin-qt app with a `-port` option:  `$ bitcoin-qt -regtest -port=20444`

apitest/src/main/java/bisq/apitest/linux/BitcoinDaemon.java
sqrrm: I think this is reasonable. These tests should never be used to start a live bitcoind.

apitest/src/test/java/bisq/apitest/method/MethodTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+    protected final String getUnusedBtcAddress() {
+        return grpcStubs.walletsService.getFundingAddresses(createGetFundingAddressesRequest())
+                .getAddressBalanceInfoList()
+                .stream()
+                .filter(a -> a.getBalance() == 0 && a.getNumConfirmations() == 0)
+                .findFirst()
+                .get()
+                .getAddress();
+
```
sqrrm:
```suggestion
```

apitest/src/test/java/bisq/apitest/method/WalletProtectionTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Test
+    @Order(6)
+    public void testLockWalletWhenWalletAlreadyLockedShouldThrowException() {
+        var request = createLockWalletRequest();
+
+        Throwable exception = assertThrows(StatusRuntimeException.class, () ->
+                grpcStubs.walletsService.lockWallet(request));
+        assertEquals("UNKNOWN: wallet is already locked", exception.getMessage());
+
```
sqrrm:
```suggestion
```

sqrrm (dismissed):
ACK

sqrrm:
I had some suggested empty line removals, not critical. The codacy complaints don't seem legit. We have been requiring non-empty blocks, but I think that's not a good policy so hope to remove that together with the fix of bad indentation rules.

apitest/src/test/java/bisq/apitest/method/MethodTest.java
ghubstan: Extra while line removed in commit c3abd4e.

apitest/src/test/java/bisq/apitest/method/WalletProtectionTest.java
ghubstan: Extra while line removed in commit c3abd4e.

ghubstan:
Closing/reopening PR to force travis build after [pricenode unit test failure](https://travis-ci.org/github/bisq-network/bisq/builds/718958332) .

ghubstan:
Closing/reopening PR again to force codacy check after @ripcurlx made some codacy config changes.

ghubstan:
> Closing/reopening PR again to force codacy check...

Closing/reopening PR does not force codacy check.

ghubstan (commented):
RE 2ba0ee9:  the private method was also moved to the bottom of the class.

This change did not force a codacy check.

sqrrm (dismissed):
ACK

ghubstan:
This PR should be merged before [PR 4417](https://github.com/bisq-network/bisq/pull/4417).

ripcurlx:
@ghubstan I contacted Codacy on this issue. As the configuration problem should be solved (I've even deactivated it for testing), but it sill doesn't green light this PR.

ghubstan:
> @ripcurlx (& @sqrrm):  ghubstan I contacted Codacy on this issue. As the configuration problem should be solved (I've even deactivated it for testing), but it sill doesn't green light this PR.

I pushed a change to shorten some too-long lines in one class, and codacy passed it.   But I don't know if the problems are fixed yet, since codacy  analyzed only the one changed class(?)  I may push a small change to the next [PR](https://github.com/bisq-network/bisq/pull/4417) , based off this one, to see what codacy does with it.

By the way, the failed Travis CI JDK12 test failed to install JDK 12, so I won't close/re-open this PR to force another  Travis build.    I don't want to risk losing the Codacy Review PASS ;)

sqrrm (approved):
ACK

PR #4373, opened 2020-07-23
Change IP for local node connection to Loopback address

QuantumRipple:
If the local Bitcoin full node is bound to only listen on the loopback interface (127.0.0.1), attempting to open a socket to `InetAddress.getLocalHost()` - the return of which is variable but usually NOT 127.0.0.1 - will not work. Changing to `InetAddress.getLoopbackAddress()` resolves this.

### DIFF
--- a/core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
+++ b/core/src/main/java/bisq/core/btc/nodes/LocalBitcoinNode.java
@@ -81,7 +81,7 @@ public class LocalBitcoinNode {
      */
     private static boolean detect(int port) {
         try (Socket socket = new Socket()) {
-            var address = new InetSocketAddress(InetAddress.getLocalHost(), port);
+            var address = new InetSocketAddress(InetAddress.getLoopbackAddress(), port);
             socket.connect(address, CONNECTION_TIMEOUT);
             log.info("Local Bitcoin node detected on port {}", port);
             return true;


ripcurlx:
@dmos62 Could you please have a look at this? Thanks!

dmos62:
> If the local Bitcoin full node is bound to only listen on the loopback interface (127.0.0.1), attempting to open a socket to `InetAddress.getLocalHost()` - the return of which is variable but usually NOT 127.0.0.1 - will not work. Changing to `InetAddress.getLoopbackAddress()` resolves this.

Researched this a bit, but I'm still a bit fuzzy. As far as I can tell `localhost` normally does resolve to the loopback interface. Under what conditions doesn't it?

QuantumRipple:
On my linux system, getLocalHost() returns 127.0.1.1 - the hosts file alias for my hostname. The typical usecase for getLocalHost() is to get the LAN address of the host (e.g. 192.168.1.x) but even that it does unreliably as evidenced by it giving 127.0.1.1 on my system. All of these addresses will connect to local machine so long as the listener is bound to all incoming addresses (0.0.0.0) rather than a specific interface.

dmos62:
> On my linux system, getLocalHost() returns 127.0.1.1 - the hosts file alias for my hostname. The typical usecase for getLocalHost() is to get the LAN address of the host (e.g. 192.168.1.x) but even that it does unreliably as evidenced by it giving 127.0.1.1 on my system. All of these addresses will connect to local machine so long as the listener is bound to all incoming addresses (0.0.0.0) rather than a specific interface.

Hm, I'll lay out how I understand things, which conflict with some of the things you said:
- `0.0.0.0` doesn't, as far as ipv4 is concerned, signify "all interfaces". It signifies an invalid address and participants are free to interpret it as they like (the "all interfaces" interpretation is a common convention, but not the only one): 
    > In the Internet Protocol Version 4, the address 0.0.0.0 is a non-routable meta-address used to designate an invalid, unknown or non-applicable target. This address is assigned specific meanings in a number of contexts, such as on clients or on servers. [Wikipedia](https://en.wikipedia.org/wiki/0.0.0.0)
- `localhost`, as defined by ipv4, is meant to resolve to your default loopback interface, not the IP address of your physical interface; systems that can have multiple loopback interfaces (e.g. 127.0.0.1, 127.0.1.1, ...), like Linux, have ways to specify which is the default; sounds like on yours it's the hosts file.

In your situation, it seems like, the problem is that bitcoind is binding to a non-default loopback interface. Either reconfigure bitcoind, or change the default loopback interface by editing the hosts file.

I am against the suggestion to query for a loopback address directly in Bisq, because it's undefined what loopback address it should return. The purpose of localhost is to define which loopback address is the default, so it's idiomatic to use it.

QuantumRipple:
Perhaps I have not been clear about what this PR fixes - my `localhost` _is_ 127.0.0.1. `InetAddress.getLocalHost()` does _not_ (always) return the address for `localhost`, and is thus not the idiomatic way to get the default loopback address. On my system, for example, it returns the address for my computer's hostname (just like the docs say it is intended to do). `InetAddress.getLoopbackAddress()` returns the IP for `localhost`. Furthermore, `InetAddress.getLocalHost()` can fail to resolve and thow an exception. `InetAddress.getLoopbackAddress()`cannot.
```
jshell> InetAddress.getLocalHost()
$1 ==> computername/127.0.1.1

jshell> InetAddress.getLoopbackAddress()
$2 ==> localhost/127.0.0.1
```
`0.0.0.0` in the context of the bitcoind configuration file's bind address means listen on all addresses. Other definitions are not applicable.

dmos62:
What is the output of `dig +short localhost` on that system? Btw, could you drop a link to the docs you mentioned?

QuantumRipple:
```
$ dig +short localhost
127.0.0.1
```
```
$ cat /etc/hosts
# Host addresses
127.0.0.1  localhost
127.0.1.1  computername
::1        localhost ip6-localhost ip6-loopback
ff02::1    ip6-allnodes
ff02::2    ip6-allrouters
```
API Docs: http://cr.openjdk.java.net/~iris/se/11/build/latest/api/java.base/java/net/InetAddress.html 
Note that "the local host name" is not the same as `localhost`.

dmos62:
> Note that "the local host name" is not the same as localhost.

That's the part I missed. I was confused why there's both a resolve localhost method and a get loopback interface method.

dmos62 (approved):
ACK. Behaviour being fixed is indeed a bug.

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/4373#pullrequestreview-456736695

PR #4377, opened 2020-07-26
Clean up TOR status icon

pedromvpg:
Removed some left-over white pixels from the tor status icon that were still visible from a past white background circle.
85822841-de66dc00-b749-11ea-8cb2-2fc789b50982


ripcurlx (approved):
utACK - :+1:

PR #4382, opened 2020-07-28
Force non-interactive install of git lfs

cd2357:
Update install command to not need user interaction

### DIFF
--- a/pricenode/install_pricenode_debian.sh
+++ b/pricenode/install_pricenode_debian.sh
@@ -36,7 +36,7 @@ sudo -H -i -u "${ROOT_USER}" DEBIAN_FRONTEND=noninteractive apt-get upgrade -qq 
 
 echo "[*] Installing Git LFS"
 sudo -H -i -u "${ROOT_USER}" curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash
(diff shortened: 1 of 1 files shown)


pricenode/install_pricenode_debian.sh
```diff
@@ -36,7 +36,7 @@ sudo -H -i -u "${ROOT_USER}" DEBIAN_FRONTEND=noninteractive apt-get upgrade -qq
 
 echo "[*] Installing Git LFS"
 sudo -H -i -u "${ROOT_USER}" curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash
-sudo -H -i -u "${ROOT_USER}" apt-get install git-lfs
+sudo -H -i -u "${ROOT_USER}" apt-get -y install git-lfs
```
ripcurlx, line 39: @cbeams Do you remember if there was a reason doing it as an interactive install before?

devinbileck: That was my error. I forgot to include the non-interactive flag.

devinbileck (approved):
utACK

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/4382#pullrequestreview-457968351

PR #4383, opened 2020-07-28
Maintain floor amount of 5.46 BSQ to prevent dust errors

### LINKED ISSUE #4372: Error when placing an offer
### Description

An error pops up when I form an offer and try to submit it and the offer has a "minimum BTC" value.

#### Version

1.3.6

### Steps to reproduce

1. Create New Offer to Buy BTC with CAD
2. Fill out the form - set amount and price - I'm setting 0.07 BTC with a 0% spread and a minimum BTC of 0.015
3. Funding with 0.46 BSQ
4. Next Step: Fund your offer (works)
5. Transfer Funds from Bisq Wallet (works)
6. Review: Place Offer to Buy Bitcoin (works)
7. Confirm: Place Offer to Buy Bitcoin (ERROR)

### Expected behaviour

Offer should be posted

### Actual behaviour

Offer isn't posted

### Screenshots

bisqerror

#### Device or machine

Ubuntu Linux 64-bit on Intel x86_64

#### Additional info

Scenario works if I make another offer with no minimum BTC, so I suspect there's a bug in the minimum BTC feature

ghost:
When an offer is made using BSQ for trade fee, the BSQ amount is burnt by doing a send-to-self.  However if the BSQ change
is below the bitcoin dust limit this causes an error.  We fix this by maintaining a floor amount of 5.46 BSQ.

Fixes #4372, more info here https://github.com/bisq-network/bisq/issues/4372#issuecomment-664615771

image

### DIFF
--- a/core/src/main/java/bisq/core/offer/OfferUtil.java
+++ b/core/src/main/java/bisq/core/offer/OfferUtil.java
@@ -140,6 +140,10 @@ public class OfferUtil {
         if (makerFee == null)
             return true;
 
+        Coin surplusFunds = availableBalance.subtract(makerFee);
+        if (Restrictions.isDust(surplusFunds)) {
+            return false; // we can't be left with dust
+        }
         return !availableBalance.subtract(makerFee).isNegative();
     }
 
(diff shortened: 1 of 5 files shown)


ripcurlx (approved):
utACK

Code is looking fine - I'll test it as part of the release testing for v1.3.7

PR #4391, opened 2020-07-31
Tweak popup text and trade rules for USPMO payment method

wiz:
Screen Shot 2020-08-01 at 4 10 10

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -3117,12 +3117,15 @@ payment.cashDeposit.info=Please confirm your bank allows you to send cash deposi
 payment.revolut.info=Please be sure that the phone number you used for your Revolut account is registered at Revolut \
   otherwise the BTC buyer cannot send you the funds.
 
-payment.usPostalMoneyOrder.info=Money orders are one of the more private fiat purchase methods available on Bisq.\n\n\
-  However, please be aware of potentially increased risks associated with their use. Bisq will not bear any \
-  responsibility in case a sent money order is stolen, and the mediator or arbitrator will in such cases award the BTC \
-  to the sender of the money order, provided they can produce tracking information and receipts. \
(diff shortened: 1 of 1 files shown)


m52go:
Looks good - we came up with the text together.

cd2357:
Is this shown only to buyers, or to both buyers and sellers?

**If only to buyers:**
* `BTC buyers must ...` => `You (as a BTC buyer) must...`
  * seems more user-friendly and less "legalese"
* `... if there is a trade dispute, you will be required to send ..` => `... if there is a trade dispute, you (as a BTC Buyer) will be required to send..`
  * same reason as above, seems more user-friendly

**If shown to both:**
* `... if there is a trade dispute, you will be required to send ..` => `... if there is a trade dispute, the BTC Buyer (USPMO sender) will be required to..`
  * clarifies even more which of the two (buyer or seller) the "you" referes to

Just a few suggestions, I'm not a native EN speaker, and have no idea how USPMO works, so take this with a grain of salt :)

wiz:
It's shown when creating a payment account, so all users see the terms, but these terms only apply for trades where you are the buyer, so I think the usage of "you" makes sense here.

And since this establishes the official Bisq policy for the USPMO payment method, it's intended to be "legalese", and follows the similar style used in terms for Monero and other payment methods that have special terms.

ripcurlx (approved):
utACK

leo816:
It looks good to me, the only thing that I would add is to warn the seller not to confirm he received the payment until he has cashed the USPMO. USPS requires a US Money Order to be at least 60 days old before someone may submit the refund paperwork. 
So two things:

- We should remind the BTC Seller that a USPMO is eligible for cancellation 60 days after printed, and that the Seller should at least check the date before closing the trade. I have seen some examples where the dates are unclear (attaching a picture for reference)

- Not to confirm with the "payment received" button until they've fully cashed it.

US Postal Money Order

PR #4047, opened 2020-03-11
Added graceful shutdown hook

freimair:
Fixes https://github.com/bisq-network/bisq/issues/3884, fixes https://github.com/bisq-network/bisq/issues/2465

Graceful shutdown has only been done in case of an error or when
using the GUI. A regular eg. seednode shutdown is not covered
though.

Now, a `kill -TERM ...` or a `kill -INT` or `[Ctrl+C]` triggers a graceful shutdown procedure.

Tested on a Bisq app and seednode service launch by firing up and then
- `kill -TERM javapid`
- `kill -INT javapid`
- run from terminal followed by hitting `[Ctrl+C]`

PR #4393, opened 2020-08-01
Don't use `kill -9` to restart seednodes now that SIGTERM is handled

wiz:
Now that we can gracefully restart from #4047


ripcurlx (approved):
utACK

PR #4396, opened 2020-08-03
Only show signing state if signing for payment method is required

ripcurlx:
While testing the USPMO payment account for v1.3.7 I recognized that there is a display issue that surfaced because of https://github.com/bisq-network/bisq/pull/4305. We are signing all payment accounts with the exact same name to have already a couple of seeds if we need to enable account signing for other payment methods as well.

This wasn't handled properly in the offerbook.

Displaying `0 days` for USPMO offer 
Bildschirmfoto 2020-08-03 um 11 55 40

Should be `N/A`
Bildschirmfoto 2020-08-03 um 12 00 25


sqrrm (approved):
utACK

PR #4402, opened 2020-08-06
Move merit and stake column to the right

chimp1984:
For better usability I moved the column next to the result.


chimp1984:
Screen Shot 2020-08-06 at 11 41 25

chimp1984:
@sqrrm Could you do a review? Is a trivial one...

sqrrm (approved):
utACK

PR #4315, opened 2020-06-16
PriceNode: Add support for multiple ExchangeRateProviders

cd2357:
Extend ExchangeRateService in pricenode to add basic support for multiple providers per currency.

When the service detects that multiple providers have exchange rates for the same currency, it automatically aggregates (averages) these rates into a single ExchangeRate.

The client thus receives a single ExchangeRate per currency, regardless of the number of providers which gathered exchange rates for it (be it a single provider, or multiple). This ensures compatibility with previous and existing clients.

The purpose of this WIP is to test and review whether:
* this approach to support multiple ExchangeRateProviders is viable
* the client transparently handles both normal and aggregate rates "as is" (so without any client modifications)
* the pricenode correctly handles multiple providers (multi-threading, caching of rates, calculation of aggregate price, etc)
* there are any fundamental showstoppers with this approach (risks, incompatibilities, etc)

Addresses bisq-network/projects#35

----
_(Update 21.07.2020: Add overview of supported currencies + providers)_

# Supported Currencies and Exchange Providers

In this PR, rates for the following currencies are retrieved from the following exchange providers:

**Table 1: Fiat currencies and corresponding exchange providers supported by this PR**

```
AED=[BITPAY, COINGECKO]
AFN=[BITPAY]
ALL=[BITPAY]
AMD=[BITPAY]
ANG=[BITPAY]
AOA=[BITPAY]
ARS=[BITPAY, COINGECKO, COINPAPRIKA]
AUD=[BITPAY, BTCMARKETS, COINGECKO, COINPAPRIKA, IndependentReserve, KRAKEN, QUOINE]
AWG=[BITPAY]
AZN=[BITPAY]
BAM=[BITPAY]
BBD=[BITPAY]
BDT=[BITPAY, COINGECKO]
BGN=[BITPAY]
BHD=[BITPAY, COINGECKO]
BIF=[BITPAY]
BMD=[BITPAY, COINGECKO]
BND=[BITPAY]
BOB=[BITPAY, COINPAPRIKA]
BRL=[BITPAY, COINGECKO, COINPAPRIKA, MercadoBitcoin]
BSD=[BITPAY]
BTN=[BITPAY]
BWP=[BITPAY]
BYN=[BITPAY]
BZD=[BITPAY]
CAD=[BITPAY, COINGECKO, COINPAPRIKA, KRAKEN]
CDF=[BITPAY]
CHF=[BITPAY, COINGECKO, COINPAPRIKA, KRAKEN]
[...]

PR #4406, opened 2020-08-08
Update GUI labels to reflect multiple rate providers

cd2357:
Update the Bisq GUI labels to better indicate the source of the exchange rate data.

The areas changed are:

1) labels in the top-right box showing the exchange rate
2) label in the About view

**1) Top right box / The shown text for fiat and alts:**

Screenshot from 2020-08-09 13-09-40

**1) Top right box / The tooltip for fiat and alts:**

Screenshot from 2020-08-09 13-58-55

Screenshot from 2020-08-09 13-59-11

**2) About view / Source of market prices under Data Providers**

Screenshot from 2020-08-09 15-32-50

---

Addresses bisq-network/projects#35

Related to #4315


cd2357:
@wiz this is the label and tooltip change, in a separate PR.

core/src/main/java/bisq/core/provider/price/PriceFeedService.java
```diff
@@ -400,7 +384,12 @@ public void onSuccess(@Nullable Tuple2<Map<String, Long>, Map<String, MarketPric
                 UserThread.execute(() -> {
                     checkNotNull(result, "Result must not be null at requestAllPrices");
                     timeStampMap = result.first;
-                    epochInSecondAtLastRequest = timeStampMap.get("btcAverageTs");
+
+                    // Each currency rate has a different timestamp, depending on when
+                    // the pricenode aggregate rate was calculated
+                    // However, the request timestamp is when the pricenode was queried
+                    epochInMillisAtLastRequest = System.currentTimeMillis();
```
wiz, line 391: Hold on a second, this isn't just changing labels, you're actually changing code here. It looks like you're replacing the timestamp of the actual price data, with the timestamp Bisq updated it from the Pricenode, which is incorrect. This behavior needs to be preserved, for example if a pricenode has an issue and doesn't get data updates, the local Bisq node needs to be able to detect that it's old data and switch to a different Pricenode with recently updated data.

cd2357: Since there is no more BA provider, the `btcAverageTs` field is basically useless.

It is however used here to populate the `epochInMillisAtLastRequest` timestamp -- which is used in the UI to show when the fiat rates were retrieved.

Before, when using BA: this made sense, cause BA was the "single source of truth" for all fiat rates, and Poloniex was the source of alt rates.

Now, when using multiple providers: basically the pricenode itself is the "provider", because it aggregates rates from multiple sources for both fiat and alts.

So, the closest "timestamp" that shows the last time the "provider" (pricenode) was polled, is the one in the code snippet above.

Alternative would be to extract and index every timestamp per exchange rate, then selectively show the right one depending on what currency pair the user selected in the UI. That seems like overkill for what this change tries to achieve. Especially since those "per exchange rate" timestamps are very likely the same, cause they all show the point in time when the avg rate was calculated, and they're all calculated when the pricenode is polled (the avg is freshly calculated based on newest data).

So I would say, this is the simplest and most pragmatic approach (`epochInMillisAtLastRequest` = `timestamp_when_pricenode_last_successfully_queried`).

wiz: The timestamp is not for when the data was updated from the pricenode, that's irrelevant. The timestamp is for the price itself, i.e. when the price was obtained from the provider, or in the case of a weighted average aka index, the time when the index was last updated. Can you please use the `timestampSec` field for each asset's index?

cd2357: Done. This field now just stores the "pricenode query timestamp", but the `timestampSec` of each asset is shown in the tooltip (see PR screenshots).

wiz (dismissed):
ACK ef5b80440cda2eee7de2f5c155eedfaf042cb3dd
looks great!
Screen Shot 2020-08-10 at 1 32 14
Screen Shot 2020-08-10 at 1 34 23
Screen Shot 2020-08-10 at 1 33 07

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1166,7 +1165,6 @@ setting.about.providers=Data providers
 setting.about.apisWithFee=Bisq uses 3rd party APIs for Fiat and Altcoin market prices as well as for mining fee estimation.
 setting.about.apis=Bisq uses 3rd party APIs for Fiat and Altcoin market prices.
```
wiz:
```suggestion
+setting.about.apisWithFee=Bisq uses Bisq Price Indices for Fiat and Altcoin market prices, and Bisq Mempool Nodes for mining fee estimation.
+setting.about.apis=Bisq uses Bisq Price Indices for Fiat and Altcoin market prices.
```

wiz: Trusted Third Parties are security holes :wink:

sqrrm (commented):
How do we handle roll back of the Bisq index in case of issues? Might be better to wait with merging this PR.

cd2357:
> How do we handle roll back of the Bisq index in case of issues? Might be better to wait with merging this PR.

Yes, I'd also suggest to wait with this PR. At least until the pricenode PR runs for a while and you're confident it doesn't need a rollback.

wiz:
just-merge-it

sqrrm (approved):
utACK

Issue #4300, opened 2020-06-14
Peers disconnecting with UNKNOWN_EXCEPTION


eigentsmis:
**TESTING UPDATE: PLEASE SEE MY LAST COMMENT IN THIS THREAD**

After 15-20min from opening the Bisq app on OSX 10.12.6 I see very long roundtrip times (see attached snippet) and also most of the peers disconnected with UNKNOWN_EXCEPTION (see attached log)

Did a lot of testing and my conclusion so far is that it's directly related to the "Turn display off after" setting in the Energy Saver window on my Mac. It seems that once the display goes off it also puts the system in some sort of idle/sleep mode and I believe this impacts Bisq's network connectivity.

Two ways which I've so far tested successfully as a workaround to this issue are:

1. Set the "Turn display off after" setting to "Never" OR
2. Add the "-d" option to the existing caffeinate command which Bisq issues upon starting.
(Example: /usr/bin/caffeinate -w 4137 -d)

<img width="87" alt="huge_roundtrip_times" src="https://user-images.githubusercontent.com/57468555/84581979-0f432a80-adb4-11ea-8844-3c3fba7e0109.png">

[UNKNOWN_EXCEPTION.log](https://github.com/bisq-network/bisq/files/4775689/UNKNOWN_EXCEPTION.log)

dmos62:
You experience these connection issues if you start Bisq after the display had turned itself off, right? If the display turns itself off (for the first time) while a Bisq instance is running (without connection issues until this point), does the connection degrade immediately? Does it ever recover without rebooting your Mac?

eigentsmis:
> You experience these connection issues if you start Bisq after the display had turned itself off, right? If the display turns itself off (for the first time) while a Bisq instance is running (without connection issues until this point), does the connection degrade immediately? Does it ever recover without rebooting your Mac?

The steps are:
1. Open Bisq (with display on and set for 5min to go in standby if no activity)
2. Wait about 20-30min
3. Login back to computer and see that Bisq connections are either very low in number or have very high roundtrip times. Also, number of offers present goes way down (most of the time to no offers)
4. After waiting a while and not letting the display go off again I do see that Bisq is trying to recover itself and I see an improvement in roundtrip times but it seems only some offers show up and not all of them.
5. To see all offers again a Bisq restart is required.

eigentsmis:
BIG TESTING UPDATE:
The display going to sleep and causing this issue seems to be a byproduct of the following root issue:

[...]

PR #4407, opened 2020-08-09
Revert usage of caffeinate on OSX

chimp1984:
It seems caffeinate is not preventing that Bisq gets throttled resources
once the OS switches to hibernate.
See: https://github.com/bisq-network/bisq/issues/4300#issuecomment-670997004

### DIFF
--- a/core/src/main/java/bisq/core/app/AvoidStandbyModeService.java
+++ b/core/src/main/java/bisq/core/app/AvoidStandbyModeService.java
@@ -173,7 +173,9 @@ public class AvoidStandbyModeService {
 
     private File getSoundFile() throws IOException, ResourceNotFoundException {
         File soundFile = new File(config.appDataDir, "prevent-app-nap-silent-sound.aiff");
-        if (!soundFile.exists()) {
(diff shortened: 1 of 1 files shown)


chimp1984:
Just tested master version again. Bisq binary has about 50 offers after a few hours hibernate. Once activated it speeds up to > 400 offers. If I start the app from the IDE/src code it seems it does not fall into hibernate. Might be that the IDE keeps it alive for debugger or that it behaves differently if run as application.

chimp1984:
After more testing and feedback from @ripcurlx it seems the issues happens only if run as binary and not on all OSX version. I could reproduce it always with 10.12.6. Even without going to sleep mode the inactive app got throttled and lost most offers. As we do not know which versions are affected I recommend to revert to the old sound file solution.

chimp1984:
I will try to get the sound file smaller and the noise level lower as well.

ripcurlx (approved):
utACK - as I never experienced the issue on my macOS version (Catalina) I can't test it properly. But I'm fine with rolling back to the previous solution which proved to work across versions.

PR #4408, opened 2020-08-12
Update seednodes

devinbileck:
As part of the Tor V3 upgrade, this PR adds 2 new V3 seednodes, and will
start the process to retire older V2 seednodes.

The V2 seednodes will continue operating for 2-3 months during the
retirement phase-out period, after which time we can filter it out from
the network for clients who have not upgraded if necessary.

See https://github.com/bisq-network/ops/issues/4


chimp1984 (approved):
utACK

ripcurlx (approved):
utACK

PR #4409, opened 2020-08-12
Optimize AssetService

chimp1984:
Previous version had several performance issues:
- We kept tradeStatistics in a lookup map in memory (about 40 MB)
- We built up the StatefulAssets even if it was not used/needed.
Building the StatefulAssets was an expensive call.
- AssetTradeActivityCheck was only used for printing out statistics and
called on startup causing performance cost.

Optimized version:
- We call build the StatefulAssets only on demand (user activates
view). Method call takes now 22 ms.
- We create a optimized lookup map with a tuple of the data we need
and keep it as local variable.
- We removed AssetTradeActivityCheck.

Thanks to the profiling reporting from:
https://github.com/bisq-network/proposals/issues/199


core/src/main/java/bisq/core/dao/governance/asset/AssetService.java
```diff
@@ -144,52 +117,137 @@ public void addListeners() {
...
+                errorMessageHandler.handleErrorMessage(exception.getMessage());
+            }
+        });
+    }
+
+
+    ///////////////////////////////////////////////////////////////////////////////////////////
+    // Private
+    ///////////////////////////////////////////////////////////////////////////////////////////
+
```
sqrrm, line 240:
```suggestion

    // Get the trade volume from lookupDate until current date
```

desktop/src/main/java/bisq/desktop/main/dao/burnbsq/assetfee/AssetFeeView.java
```diff
@@ -261,17 +287,20 @@ private long getDays() {
         return getListingFee().value / assetService.getFeePerDay().value;
     }
 
+    // We only update on new BSQ blocks and at view activation. We do not update at each trade statistics change as
+    // that would cause too much CPU load. The assetService.updateAssetStates() call takes about 22 ms.
     private void updateList() {
+        log.error("updateList");
```
sqrrm:
```suggestion
```

sqrrm (changes requested):
Looks good. Would prefer to have pure refactoring as separate commits for easier review.

Added a couple of suggestions before merging.

Needs review from @ManfredKarrer since it touches core/dao

chimp1984:
> Would prefer to have pure refactoring as separate commits for easier review.

Agree. Sorry.

ManfredKarrer:
utACK

sqrrm (approved):
ACK

PR #4410, opened 2020-08-13
Optimize application shutdown

chimp1984:
There have been several long delays as well a wrong order of the
shutdown process (wallet got shutdown after network shutdown.

Shutdown is now pretty fast, but depends on open offers and connections.


sqrrm (approved):
ACK

PR #4414, opened 2020-08-17
[WIP] Atomic trade

sqrrm:
This PR implements a new account type, Atomic BSQ with a separate trade protocol, see https://github.com/bisq-network/proposals/issues/50

Atomic trade protocol is completed in one request -> response interaction between taker and maker.
1. Taker sends a CreateAtomicTxRequest to maker. This includes all the data needed to create and sign the atomic tx from maker's side.
1. Maker verifies all inputs and answers with CreateAtomicTxResponse that includes an atomic tx with maker's inputs signed.
1. Taker verifies the inputs and that their own outputs are paid as expected, signs and publishes the completed atomic tx.
1. Fin

image
There is still a minimum 15% deposit to avoid messing around too much with current trade protocol. It's not actually needed since the trade is an atomic transaction, but I leave that as a future improvement.

image
Can't be taken by an old client since the account type doesn't exist

image
Take with new client, requires to have the deposit, but it's not used apart from checking that it's available during the take offer process.

image
Trade completes immediately.

image
image
Shows in transactions as atomic

image
image
Need a confirmation before it can show the type of BSQ tx

image
image
After confirmation

### DIFF
--- a/core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
+++ b/core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
@@ -44,7 +44,10 @@ public class TxFeeEstimationService {
     public static int TYPICAL_TX_WITH_1_INPUT_SIZE = 260;
     private static int DEPOSIT_TX_SIZE = 320;
     private static int PAYOUT_TX_SIZE = 380;
+    private static int ATOMIC_TX_SIZE = 975;
     private static int BSQ_INPUT_INCREASE = 150;
+    private static int EXTRA_INPUT_INCREASE = 225;
+    private static int EXTRA_OUTPUT_INCREASE = 70;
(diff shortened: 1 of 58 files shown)


sqrrm:
When the makerAsBuyer (selling BSQ) doesn't have enough BSQ at the time their offer is taken they get a warning message popup
image

The taker gets an error popup
image

This could be handled better. First, a hold on BSQ could be added to make sure the offer is funded while it's available. Then there is old time out error that's not very pretty. We could improve on the message and make it look a bit less threatening, but the effect is still going to be a failed take offer attempt no matter how we spin it.

flix1:
This looks amazing. Looking forward to seeing it implemented!

pazza83:
This is great thanks for this. Looking forward to try it out.

PR #4415, opened 2020-08-18
Add option to use different css for dev mode

sqrrm:
This is to help differentiate dev instances and live ones.
This can help devs avoiding silly mistakes.

Small thing, but a different color really helps sometimes.

### DIFF
--- a/common/src/main/java/bisq/common/config/Config.java
+++ b/common/src/main/java/bisq/common/config/Config.java
@@ -68,6 +68,7 @@ public class Config {
     public static final String BASE_CURRENCY_NETWORK = "baseCurrencyNetwork";
     public static final String REFERRAL_ID = "referralId";
     public static final String USE_DEV_MODE = "useDevMode";
+    public static final String USE_DEV_MODE_HEADER = "useDevModeHeader";
     public static final String TOR_DIR = "torDir";
     public static final String STORAGE_DIR = "storageDir";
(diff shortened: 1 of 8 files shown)


wiz:
Cool idea, obviously Testnet should be a different color. To take it even further tho, is the additional command line option even necessary? Seems like you could just check isDevMode() in the loadSceneStyles() and load the theme-dev.css

sqrrm:
It would be a bit annoying to do UI development if the UI isn't looking like prod so I figured it's better to have it as a separate option.

sqrrm:
@ripcurlx The codacy settings for css indentation seems off as well, it's expecting 2 characters rather than 4.

ripcurlx:
> @ripcurlx The codacy settings for css indentation seems off as well, it's expecting 2 characters rather than 4.

I already updated this some time ago. I tried to re-analyze, but if it doesn't work I'll ignore this pattern here for now as well.

ripcurlx:
@sqrrm I removed the pattern for now. Could you please force push to this branch so it is getting rebuilt in Codacy? Thanks!

sqrrm:
@ripcurlx thanks, at least codacy is not complaining. Let's re-enable stuff if we notice there is a need for it to keep the balance of helpfulness and annoyance.

ripcurlx (approved):
utACK

PR #4420, opened 2020-08-20
Fix intermittent blank price cells in offer book view

stejbac:
Replace faulty cell update logic, which uses a `ChangeListener<Scene>`, added in July 2017 (#73f21399) to keep the price column in the offer book table up to date, as it appears to occasionally result in blank cells. Also it seems only the prices, not the volumes, were being kept in sync with the market price feed.

Make the price and volume cells stateless and keep them in sync with the market feed by adding it as a dependency of each `OfferBookListItem` Observable generated by the cell value factory, instead of directly attaching listeners to it. In this way, `TableCell::updateItem` will be called by the framework whenever the price/volume needs updating.

(This does have the disadvantage that if the price feed is unavailable, causing `Offer::getPrice` to return null, then the cells will reflect that immediately instead of showing any old, stale values, but that is necessary for the UI to behave consistently anyway.)

--

This is to fix issues like the following, that I have occasionally observed in the Sell Bitcoin offer book, filtered for BSQ (though never in the Buy Bitcoin offer book or for other counter currencies for some reason):

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartViewModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartViewModel.java
@@ -50,7 +50,6 @@ import javafx.beans.property.ObjectProperty;
 import javafx.beans.property.SimpleIntegerProperty;
 import javafx.beans.property.SimpleObjectProperty;
 import javafx.beans.value.ChangeListener;
-import javafx.beans.value.ObservableValue;
 
 import javafx.collections.FXCollections;
 import javafx.collections.ListChangeListener;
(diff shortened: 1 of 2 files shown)


stejbac:
(I seem to be having some difficulty attaching any PNGs or editing the above description in my current browser for some reason - perhaps a bug in the GitHub Javascript?)

stejbac:
(I seem to be having some difficulty attaching any PNGs or editing the above description in my current browser for some reason - perhaps a bug in the GitHub Javascript?)

stejbac:
(I seem to be having some difficulty attaching any PNGs or editing the above description in my current browser for some reason - perhaps a bug in the GitHub Javascript?)

stejbac:
(I seem to be having some difficulty attaching any PNGs or editing the above description in my current browser for some reason - perhaps a bug in the GitHub Javascript?)

stejbac:
(I seem to be having some difficulty attaching any PNGs or editing the above description in my current browser for some reason - perhaps a bug in the GitHub Javascript?)

sqrrm (approved):
ACK

I think showing nothing when there is no feed is better than showing an old price, so that's also an improvement in my opinion.

stejbac:
Here is a screenshot of the problem I meant to attach earlier (but couldn't as I was using an old unsupported browser):

Screenshot from 2020-07-24 15-30-49

PR #4421, opened 2020-08-20
Autoconfirm XMR trade if tx check was validated by proof service

ghost:
This is a follow-up PR to #4378 by @chimp1984

Implements bisq-network/proposals#86

### Features:

* XMR seller is prompted to enter txId and viewkey.

On the XMR buyer side,
* looks up the XMR transaction to verify that it has confirmed (using [onion-monero-blockchain-explorer](https://github.com/moneroexamples/onion-monero-blockchain-explorer))
* user can run their own validating service, or use the ones provided
* 100% agreement of all chosen services is necessary for confirmation
* feature can be configured and turned on/off from settings screen
* feature can be globally turned off via admin filter
 

### Implementation notes

Auto confirmation entails API requests to multiple autoconf service providers.  If the responses all show that the required number of confirmations have been reached the trade will be auto confirmed.

The user still has the option to confirm manually just as if auto confirm were not enabled.

Any service failure will abort the attempt at auto confirming.
- if a service is down, we fail with CONNECTION_FAIL
- if a service returns garbage response, we fail with API_INVALID
- if a service returns a different address, we fail with ADDRESS_INVALID
- if a service returns a different txId, we fail with TX_HASH_INVALID
- if a service returns a different viewkey, we fail with TX_KEY_INVALID
- if a service returns a match, but the amount is different we fail with AMOUNT_NOT_MATCHING
- if a service returns transaction not found we wait then re-try
- if a service returns less than the required confirm count, we wait then re-try
- if a service returns >= the required confirm count, TradeManager decrements its count of outstanding service calls.  Once the count reaches zero it transitions the trade to confirmed.

In any of these cases the status is shown on screen.  It also shows the progress of the confirmation count.

The general approach to implementing this feature was to keep it as simple as possible.  

Auto confirmation is a one-shot process triggered by receipt of `counterCurrencyExtraData` in the `CounterCurrencyTransferStartedMessage` from the counterparty.  

If Bisq is closed while auto confirmation was pending, when it re-starts the trade will show auto conf status disabled.  
~~Auto-confirmed status of a trade is not persisted.~~ Auto-confirmed status is persisted since for trade history it might be important to have some "proof" if it was used or not, as well in dispute cases. [requested by @chimp1984].

### Testing

In regtest use a clearnet service address, e.g. `78.47.61.90:8081` (from @Emzy)

In mainnet over tor this would be `monero3bec7m26vx6si6qo7q7imlaoz45ot5m2b5z2ppgoooo6jx2rqd.onion` which is included as a default setting.

To try out this feature without making a live XMR trade, the following example TX proofs may be used.

    Sent 0.12121212 XMR (for sell 0.0012 BTC @ price 0.0099)
    to:  4ATyxmFGU7h3EWu5kYR6gy6iCNFCftbsjATfbuBBjsRHJM4KTwEyeiyVNNUmsfpK1kdRxs8QoPLsZanGqe1Mby43LeyWNMF
    tx:  0d0d5e74e56c9060095e62b92bce2c7fe8da43a28d3e11a7c51ffe0749fe4dd4
    key: 443263d31c8b05f5dd5734128fcb3cac83d0fd3c90b03ec19d5131216a5e3505

    Sent 0.1 XMR (for sell 0.001 BTC @ price 0.01)
    to:  4ATyxmFGU7h3EWu5kYR6gy6iCNFCftbsjATfbuBBjsRHJM4KTwEyeiyVNNUmsfpK1kdRxs8QoPLsZanGqe1Mby43LeyWNMF
    tx:  e8dcd8160aee016d8a0d9c480355d65773dc577313a0af8237c35f9d997b01c0
    key: 300fa18ff99b32ff097d75c64d62732bdb486af8c225f558ee48c5f777f9b509

Automated testing routines were created to verify the API JSON response handling.

Checking iteroperability between new and old clients.
- If BTC Seller = new BTC buyer = old, the buyer is not prompted for txId and key and the new proto field is not sent to the BTC seller.  Therefore auto-confirmation will not be started and the trade will operate in the old way, manual confirmation.
- If BTC Seller = old BTC buyer = new, the buyer provides txId and key and sends the new proto field to the BTC seller.  The BTC seller does not handle the new field and the trade will operate in the old way, manual confirmation.

**Test cases**

- New user (preferences)
- Auto-confirm success
- Premature manual confirm by user
- XMR Tx Id / key not provided
- XMR Tx not found
- Re-used XMR TxId
- Feature disabled in settings
- Feature disabled by filter
- Transaction limit exceeded
- Invalid amount
- XMR Trade date too old
- XMR Trade date newer than Bisq trade date (ok)
- Bisq shut down during Auto-confirmation procedure
- XMR API service not reachable

### Discussion

* Perhaps some XMR sellers may not want (or be able) to provide the txId and viewkey?  In interest of flexibility, the UI allows those values to be optional and simply disables autoconf for the trade if the keys are empty or do not pass validation.


core/src/main/java/bisq/core/trade/asset/xmr/XmrProofInfo.java
```diff
@@ -0,0 +1,152 @@
...
+                }
+            }
+
+            // reaching this point means there was no matching amount
+            return new XmrProofResult(XmrProofResult.State.AMOUNT_NOT_MATCHING, null);
+
+            // TODO check if date of tx is after tradeDate (allow some tolerance to avoid clock sync issues).
+            //  Otherwise a scammer could send an old txKey from a prev trade with same amount before seller has
+            //  updated to new feature, thus the check for duplication would not detect the scam.
+            //  >>>>> the API does not provide date of tx <<<<<<
```
ghost: @chimp1984 how important is it to get tradeDate from the monero API, as it is not currently there?

chimp1984:
@jmacxx 
Great to see the final PR! I will try to review on the weekend as soon I find time.

> Perhaps some XMR sellers may not want (or be able) to provide the txId and viewkey?

It is a requirement for XMR that they use a wallet capable of providing the tx key for the BTC buyer to be able to proof the XMR transfer. In a dispute the mediator would not be able to verify if the XMR transfer was done otherwise. So we just make it more safe in the way that if user ignore the info in the popup and use a not capable wallet they would risk to lose the dispute case. Now they cannot complete the trade, or at least they get aware of the problem very early.

core/src/main/java/bisq/core/trade/asset/xmr/XmrProofInfo.java
chimp1984: I think its just some edge case at update where this could be a risk. Does the API not provide the date? If not would you mind to ask the Monero dev if he can add it? I guess it would be a small change and would remove the described risk.

chimp1984:
@boring-cyborg Wrong labelling!

core/src/main/java/bisq/core/trade/asset/xmr/XmrProofInfo.java
ghost: Created PR  https://github.com/moneroexamples/onion-monero-blockchain-explorer/pull/214

ghost: Incorporated trade date verification, I chose a tolerance window of 1 day.  https://github.com/bisq-network/bisq/pull/4421/commits/9a2e28cbb3abf6b0ba408f1c97c04845ea485ee5  Now the example trades won't auto-confirm since they are too old!  :wink:

image

assets/src/main/java/bisq/asset/CryptoNoteAddressValidator.java
```diff
@@ -57,6 +60,19 @@ public AddressValidationResult validate(String address) {
             return AddressValidationResult.invalidStructure();
         }
     }
+
+    // obtain the raw hex version of a monero address
+    public static String addressToHex(String address) {
```
chimp1984: I would suggest to name it more specifically. As we remove type and checksum it is not really converting but we use it for verifying if it is a valid XMR address. I think the hex check might be good enough for now but maybe there is a more exact validation available. If we rename the method to 'isAddressValid' and return a boolean I think it gets more clean.

core/src/main/java/bisq/core/filter/Filter.java
```diff
@@ -109,6 +109,9 @@
     @Nullable
     private final List<String> btcFeeReceiverAddresses;
 
+    // added after v1.3.7
+    private final boolean disableAutoconf;
```
chimp1984: I would suggest to stick with camelCase -> `disableAutoConf`

core/src/main/java/bisq/core/trade/asset/xmr/XmrProofInfo.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    // something to uniquely identify this object by
+    public String getKey() {
+        return txHash + "|" + serviceAddress;
+    }
+
+    public XmrProofResult checkApiResponse(String jsonTxt) {
+        try {
+            final JsonObject json = new Gson().fromJson(jsonTxt, JsonObject.class);
```
chimp1984: We don't use `final` for local variables.

core/src/main/java/bisq/core/trade/asset/xmr/XmrProofInfo.java
```diff
@@ -0,0 +1,152 @@
...
+            } else if (!jsonStatus.matches("success")) {
+                return new XmrProofResult(XmrProofResult.State.API_FAILURE, "unhandled status value");
+            }
+
+            // validate that the address matches
+            final JsonElement jsonAddress = jsonData.get("address");
+            if (jsonAddress == null) {
+                return new XmrProofResult(XmrProofResult.State.API_INVALID, "missing address field");
+            } else {
+                String expectedAddressHex = CryptoNoteAddressValidator.addressToHex(this.recipientAddress);
```
chimp1984: Ah it seems you need the trimmed address for the comparision, if so ignore my comment earlier. Still might be good to have a validate method and a convert method to make it more explicit.

core/src/main/java/bisq/core/trade/asset/xmr/XmrProofInfo.java
```diff
@@ -110,6 +117,20 @@ public XmrProofResult checkApiResponse(String jsonTxt) {
                 }
             }
 
+            // validate that the txDate matches within tolerance
+            final JsonElement jsonTimestamp = jsonData.get("tx_timestamp");
+            if (jsonTimestamp == null) {
+                return new XmrProofResult(XmrProofResult.State.API_INVALID, "missing tx_timestamp field");
+            } else {
+                long tradeDateSeconds = tradeDate.getTime() / 1000;
+                long difference = abs(jsonTimestamp.getAsLong() - tradeDateSeconds);
```
chimp1984: I think the `abs` call here is leading to undesired behaviour. We only want to protect against old txs, so then the trade data  > tx_timestamp. If we allow both directions we limit the confirmation time to 1 day, which might be mostly ok for XMR but not sure... E.g. if tradeDateSeconds is today 18:00 and tx_timestamp is tomorrow 19:00 because miner fee was low and tx got included very late it would fail, but should be still valid.
I think correct impl would be:
`long difference = tradeDateSeconds - jsonTimestamp.getAsLong();
    if (difference > 60 * 60 * 24) {....
`
Not sure if jsonTimestamp is from confirmed block or from mempool, but should not make a difference.

core/src/main/java/bisq/core/trade/asset/xmr/XmrProofResult.java
```diff
@@ -17,15 +17,81 @@
...
+        API_FAILURE,
+        API_INVALID,
+        TX_KEY_REUSED,
+        TX_HASH_INVALID,
+        TX_KEY_INVALID,
+        ADDRESS_INVALID,
+        AMOUNT_NOT_MATCHING;
+    }
+
+    @Getter
```
chimp1984: Annotating the class with @Value would make it more explicit that its a immutable class. @Getter annotations can be removed then as @Value included those.

core/src/main/java/bisq/core/trade/asset/xmr/XmrProofResult.java
```diff
@@ -17,15 +17,81 @@
...
+        this.state = state;
+        if (isErrorState())
+            log.error(errorMsg != null ? errorMsg : state.toString());
+    }
+
+    public String getTextStatus() {
+        switch (state) {
+            case TX_NOT_CONFIRMED:
+                return Res.get("portfolio.pending.autoConfirmPending")
+                        + " " + Integer.toString(confirmCount)
```
chimp1984: Integer.toString is not needed.

core/src/main/java/bisq/core/trade/asset/xmr/XmrTransferProofRequester.java
```diff
@@ -17,158 +17,109 @@
...
 
 import org.jetbrains.annotations.NotNull;
 
 @Slf4j
-@Singleton
-class XmrTransferProofRequester {
+public class XmrTransferProofRequester {
 
     private final ListeningExecutorService executorService = Utilities.getListeningExecutorService(
             "XmrTransferProofService", 3, 5, 10 * 60);
```
chimp1984: Better use class name 'XmrTransferProofRequester'

core/src/main/java/bisq/core/trade/asset/xmr/XmrTransferProofRequester.java
```diff
@@ -17,158 +17,109 @@
...
-         */
-
+        if (terminated) {
+            // the XmrTransferProofService has asked us to terminate i.e. not make any further api calls
+            // this scenario may happen if a re-request is scheduled from the callback below
+            log.info("request() aborted, this object has been terminated: {}", httpClient.toString());
+            return;
+        }
         ListenableFuture<XmrProofResult> future = executorService.submit(() -> {
             Thread.currentThread().setName("XmrTransferProofRequest-" + this.toString());
```
chimp1984: this.toString() would not result it any readable meaningful data (shows java internal id). Also if someone adds a toString mehtod we would get all that into the thread name rendering it unreadable...
Better use the tx hash + service url, that should be unique enough

core/src/main/java/bisq/core/trade/asset/xmr/XmrTransferProofRequester.java
```diff
@@ -17,158 +17,109 @@
...
     private final Consumer<XmrProofResult> resultHandler;
     private final FaultHandler faultHandler;
-
+    @Setter
+    private boolean terminated;
     private long firstRequest;
-    //todo dev settings
-    private long REPEAT_REQUEST_SEC = TimeUnit.SECONDS.toMillis(5);
+    // these settings are not likely to change and therefore not put into Config
+    private long REPEAT_REQUEST_PERIOD = TimeUnit.SECONDS.toMillis(30);
```
chimp1984: Is 30 seconds not a bit too aggressive for the nodes? Not sure whats the best value here. What are average block confirmation times on XMR? I guess some value in between 1-3 min. might be better.

core/src/main/java/bisq/core/trade/asset/xmr/XmrTransferProofRequester.java
```diff
@@ -17,158 +17,109 @@
...
-                    UserThread.execute(() -> resultHandler.accept(result));
+                if (terminated) {
+                    log.info("API terminated from higher level: {}", httpClient.toString());
+                    return;
                 }
+                if (System.currentTimeMillis() - firstRequest > MAX_REQUEST_PERIOD) {
+                    log.warn("we have tried this service API for too long, giving up: {}", httpClient.toString());
+                    return;
+                }
+                if (result.isPendingState())
```
chimp1984: Better use alwqays curly brackets also for one-liners. We had some bugs related to that in the past and changed then code style (but still in the code from old days).

core/src/main/java/bisq/core/trade/asset/xmr/XmrTransferProofRequester.java
```diff
@@ -17,158 +17,109 @@
...
-        this.httpClient = httpClient;
-        this.tradeDate = tradeDate;
-        this.txHash = txHash;
-        this.txKey = txKey;
-        this.recipientAddress = recipientAddress;
-        this.amount = amount;
+        this.httpClient = new XmrTxProofHttpClient(socks5ProxyProvider);
+        this.httpClient.setBaseUrl("http://" + xmrProofInfo.getServiceAddress());
+        if (xmrProofInfo.getServiceAddress().matches("^192.*")) {
+            log.info("ignoring Socks5 proxy for local net address: {}", xmrProofInfo.getServiceAddress());
```
chimp1984: I think there is another use case where we dont want to use the onion address. For regtest we usually dont use tor, so for dev testing its more convenient to use the clearnet IP of a service (at least one should provide clearnet access for that reason).

core/src/main/java/bisq/core/trade/asset/xmr/XmrTransferProofService.java
```diff
@@ -17,65 +17,69 @@
...
     }
 
-    private void cleanup(String tradeId) {
-        map.remove(tradeId);
+    public void terminateRequest(XmrProofInfo xmrProofInfo) {
+        final String key = xmrProofInfo.getKey();
+        XmrTransferProofRequester requester = map.getOrDefault(key, null);
+        if (requester != null) {
+            log.info("Terminating API request for {}", key);
+            requester.setTerminated(true);
```
chimp1984: I would find it better to call a method like stop() where the property is set instead of exposing the property itself. Small issue but makes API of the class more clear IMO.

core/src/main/java/bisq/core/trade/Trade.java
```diff
@@ -465,6 +475,7 @@ protected Trade(Offer offer,
         processModel = new ProcessModel();
         lastRefreshRequestDate = takeOfferDate;
         refreshInterval = Math.min(offer.getPaymentMethod().getMaxTradePeriod() / 5, MAX_REFRESH_INTERVAL);
+        xmrProofResult = new XmrProofResult(0, 0, XmrProofResult.State.FEATURE_DISABLED);
```
chimp1984: Why is xmrProofResult set here? It should be marked as a @Nullable fields and left null until used. Also we should use the setXmrProofResult method only to avoid that the xmrProofResultProperty gets out or sync.

core/src/main/java/bisq/core/trade/Trade.java
```diff
@@ -428,12 +429,21 @@ public Phase getPhase() {
...
 
+    @Getter
+    private XmrProofResult xmrProofResult;
+    public void setXmrProofResult(XmrProofResult xmrProofResult) {
+        this.xmrProofResult = xmrProofResult;
+        xmrProofResultProperty.setValue(xmrProofResult);
+    }
+    @Getter
+    // This observable property can be used for UI to show a notification to user of the XMR proof status
+    transient final private ObjectProperty<XmrProofResult> xmrProofResultProperty = new SimpleObjectProperty<>(xmrProofResult);
```
chimp1984: I think we should not pass xmrProofResult here as it is null at construction time.

core/src/main/java/bisq/core/trade/Trade.java
```diff
@@ -1170,6 +1181,7 @@ public String toString() {
                 ",\n     takerPaymentAccountId='" + takerPaymentAccountId + '\'' +
                 ",\n     errorMessage='" + errorMessage + '\'' +
                 ",\n     counterCurrencyTxId='" + counterCurrencyTxId + '\'' +
+                ",\n     counterCurrencyExtraData='" + counterCurrencyExtraData + '\'' +
```
chimp1984, line 1197: We could add xmrProofResult as well

core/src/main/java/bisq/core/trade/TradeManager.java
```diff
@@ -932,65 +935,118 @@ public void processCounterCurrencyExtraData(Trade trade) {
...
+
+        log.warn("Tx Proof Failure {}, shutting down all open API requests for this trade {}",
+                result.getState(), trade.getShortId());
+        trade.setXmrProofResult(result);         // this updates the GUI with the status..
+        resultsCountdown = -1;  // signal all API requestors to cease
+        txProofResultsPending.put(trade.getId(), resultsCountdown);   // track proof result count
+        return failure;
+    }
+
+    private boolean isDisableAutoconf() {
```
chimp1984: Maybe rename to `isAutoConfDisabledByFilter` to make it more distinct to preferences value.

core/src/main/java/bisq/core/trade/TradeManager.java
```diff
@@ -932,65 +935,118 @@ public void processCounterCurrencyExtraData(Trade trade) {
...
-                                    if (!walletsSetup.isDownloadComplete()) {
-                                        return;
-                                    }
+        if (!p2PService.isBootstrapped()) {
+            return failure;
+        }
+        if (!walletsSetup.hasSufficientPeersForBroadcast()) {
+            return failure;
+        }
+        if (!walletsSetup.isDownloadComplete()) {
```
chimp1984: I am not sure if that handling is ok (same for hasSufficientPeersForBroadcast). I guess the message which triggers the processCounterCurrencyExtraData can be received while user is syncing wallet and we dont want to stop the service in such a case but delay the requests. As it cannot be handled here I would suggest to move the 
```
 if (!walletsSetup.hasSufficientPeersForBroadcast()) {
            return failure;
        }
        if (!walletsSetup.isDownloadComplete()) {
            return failure;
        }
```
code to the 
```
result -> {
                            if (!handleProofResult(result, trade))
                                xmrTransferProofService.terminateRequest(xmrProofInfo);
                        }
```
part of the calling method.

But not sure if that is really enough as it need to be handled in other places as well where the result gets processed. Need to get a bit more overview to be able to make more meaningful suggestions....

core/src/main/java/bisq/core/trade/TradeManager.java
```diff
@@ -932,65 +935,118 @@ public void processCounterCurrencyExtraData(Trade trade) {
...
+            return failure;
+        }
 
-                                    if (!trade.isPayoutPublished()) {
-                                        trade.setState(Trade.State.SELLER_CONFIRMED_IN_UI_FIAT_PAYMENT_RECEIPT);
-                                    }
+        // here we count the Trade's API results from all
+        // different serviceAddress and figure out when all have finished
+        Integer resultsCountdown = txProofResultsPending.getOrDefault(trade.getId(), 0);
```
chimp1984: We dont use boxed types if not needed.

core/src/main/java/bisq/core/trade/TradeManager.java
```diff
@@ -932,65 +935,118 @@ public void processCounterCurrencyExtraData(Trade trade) {
...
-                        });
+        if (trade.isPayoutPublished()) {
+            log.warn("trade payout already published, shutting down all open API requests for this trade {}",
+                    trade.getShortId());
+            txProofResultsPending.remove(trade.getId());
+            return failure;
+        }
+
+        if (result.isPendingState()) {
+            log.info("Received a {} message for tradeId {}, retry will happen automatically",
```
chimp1984: This log is too generic. it should contain more specific info about the actual use case/feature.

core/src/main/java/bisq/core/trade/TradeManager.java
```diff
@@ -932,65 +935,118 @@ public void processCounterCurrencyExtraData(Trade trade) {
...
+            if (resultsCountdown > 0) {
+                txProofResultsPending.put(trade.getId(), resultsCountdown);   // track proof result count
+                return success; // not all APIs have confirmed yet
+            }
+            // we've received the final PROOF_OK, all good here.
+            txProofResultsPending.remove(trade.getId());
+            trade.setXmrProofResult(result);            // this updates the GUI with the status..
+            log.info("Auto confirm was successful, transitioning trade {} to next step...", trade.getShortId());
+            if (!trade.isPayoutPublished()) {
+                trade.setState(Trade.State.SELLER_CONFIRMED_IN_UI_FIAT_PAYMENT_RECEIPT);
```
chimp1984: Maybe we should add a comment here and to the `Trade.State.SELLER_CONFIRMED_IN_UI_FIAT_PAYMENT_RECEIPT` state that this can be also set by the autoConf feature and not only by UI activity. Renaming the state might be another option but not sure if its used in protobuf as well and if it would have any backward compatibility issues (likely not, but prob. not worth the risk and effort).

core/src/main/java/bisq/core/trade/TradeManager.java
```diff
@@ -932,65 +935,118 @@ public void processCounterCurrencyExtraData(Trade trade) {
...
             }
+            accountAgeWitnessService.maybeSignWitness(trade);
+            // transition the trade to step 4:
+            ((SellerTrade) trade).onFiatPaymentReceived(() -> { },
+                    errorMessage -> { });
+            return success;
         }
+
+        log.warn("Tx Proof Failure {}, shutting down all open API requests for this trade {}",
```
chimp1984: Can you comment more details when this case is expected?

core/src/main/java/bisq/core/user/Preferences.java
```diff
@@ -120,6 +120,10 @@
             new BlockChainExplorer("bsq.bisq.cc (@m52go)", "https://bsq.bisq.cc/tx.html?tx=", "https://bsq.bisq.cc/Address.html?addr=")
     ));
 
+    // list of XMR proof providers : this list will be used if no preference has been set
+    public static final List<String> DEFAULT_XMR_PROOF_PROVIDERS = new ArrayList<> (Arrays.asList(
+            "monero3bec7m26vx6si6qo7q7imlaoz45ot5m2b5z2ppgoooo6jx2rqd.onion"));
```
chimp1984: We should get at least 2 at release time.

core/src/main/java/bisq/core/user/Preferences.java
```diff
@@ -234,6 +238,9 @@ public void readPersisted() {
             prefPayload.setPreferredTradeCurrency(preferredTradeCurrency);
             setFiatCurrencies(CurrencyUtil.getMainFiatCurrencies());
             setCryptoCurrencies(CurrencyUtil.getMainCryptoCurrencies());
+            // default values for autoconfirmsettings when persisted payload is empty:
+            prefPayload.setAutoConfirmSettings(new AutoConfirmSettings(
+                    false, 5, 10000000, DEFAULT_XMR_PROOF_PROVIDERS));
```
chimp1984: Instead of 10000000 better use Coin.COIN.value

core/src/main/java/bisq/core/user/PreferencesPayload.java
```diff
@@ -127,8 +127,8 @@
     private int blockNotifyPort;
     private boolean tacAcceptedV120;
 
-    // Added with 1.3.7 false be default
-    private boolean autoConfirmXmr;
+    // Added after 1.3.7
+    private AutoConfirmSettings autoConfirmSettings;
```
chimp1984: Annotate with @Nullable

core/src/main/java/bisq/core/user/PreferencesPayload.java
```diff
@@ -190,7 +190,7 @@ public Message toProtoMessage() {
                 .setBuyerSecurityDepositAsPercentForCrypto(buyerSecurityDepositAsPercentForCrypto)
                 .setBlockNotifyPort(blockNotifyPort)
                 .setTacAcceptedV120(tacAcceptedV120)
-                .setAutoConfirmXmr(autoConfirmXmr);
+                .setAutoConfirmSettings((protobuf.AutoConfirmSettings) autoConfirmSettings.toProtoMessage());
```
chimp1984: Has to be handled as nullable value: Optional.ofNullable(

core/src/main/java/bisq/core/user/PreferencesPayload.java
```diff
@@ -279,6 +280,6 @@ public static PreferencesPayload fromProto(protobuf.PreferencesPayload proto, Co
                 proto.getBuyerSecurityDepositAsPercentForCrypto(),
                 proto.getBlockNotifyPort(),
                 proto.getTacAcceptedV120(),
-                proto.getAutoConfirmXmr());
+                AutoConfirmSettings.fromProto(userAutoConfirmSettings));
```
chimp1984: Has to be handled as nullable value

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -569,6 +569,11 @@ portfolio.pending.step2_buyer.startPayment=Start payment
 portfolio.pending.step2_seller.waitPaymentStarted=Wait until payment has started
 portfolio.pending.step3_buyer.waitPaymentArrived=Wait until payment arrived
 portfolio.pending.step3_seller.confirmPaymentReceived=Confirm payment received
+portfolio.pending.step3_seller.autoConfirmStatus=Auto-Confirm Status
+portfolio.pending.autoConfirmTxNotFound=Tx Not Found
```
chimp1984: If space in UI permits better use full term `transaction` instead of tx.

chimp1984: also lower case...
portfolio.pending.autoConfirmTxNotFound=Transaction not found

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -569,6 +569,11 @@ portfolio.pending.step2_buyer.startPayment=Start payment
 portfolio.pending.step2_seller.waitPaymentStarted=Wait until payment has started
 portfolio.pending.step3_buyer.waitPaymentArrived=Wait until payment arrived
 portfolio.pending.step3_seller.confirmPaymentReceived=Confirm payment received
+portfolio.pending.step3_seller.autoConfirmStatus=Auto-Confirm Status
```
chimp1984: portfolio.pending.step3_seller.autoConfirmStatus=Auto-confirm status

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -718,8 +723,8 @@ portfolio.pending.step3_seller.amountToReceive=Amount to receive
 portfolio.pending.step3_seller.yourAddress=Your {0} address
 portfolio.pending.step3_seller.buyersAddress=Buyers {0} address
 portfolio.pending.step3_seller.yourAccount=Your trading account
-portfolio.pending.step3_seller.xmrTxHash=Tx hash
-portfolio.pending.step3_seller.xmrTxKey=Tx private key
+portfolio.pending.step3_seller.xmrTxHash=Tx id
```
chimp1984: We use `Transaction ID` if space permits, otherwise `Tx ID`

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1232,6 +1241,9 @@ setting.about.shortcuts.sendFilter=Set Filter (privileged activity)
 setting.about.shortcuts.sendPrivateNotification=Send private notification to peer (privileged activity)
 setting.about.shortcuts.sendPrivateNotification.value=Open peer info at avatar or dispute and press: {0}
 
+setting.info.headline=New XMR Autoconfirm feature
+setting.info.msg=When selling BTC for XMR you can use the autoconfirm feature to verify that the correct amount of XMR has been deposited to your wallet and automatically confirm the trade, releasing BTC to the counterparty.\n\n\
+  This can make the trade process more efficient for both peers.  Limits can also be set on the max amount of BTC that can be auto-confirmed.  See the auto-confirm section on this page for more info and to enable the feature.
```
chimp1984: We should request from m52go to review/improve the text.

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -2493,8 +2506,8 @@ showWalletDataWindow.walletData=Wallet data
 showWalletDataWindow.includePrivKeys=Include private keys
 
 setXMRTxKeyWindow.headline=Prove sending of XMR
-setXMRTxKeyWindow.txHash=Transaction hash
-setXMRTxKeyWindow.txKey=Tx private key
+setXMRTxKeyWindow.txHash=Transaction id
+setXMRTxKeyWindow.txKey=Tx key
```
chimp1984: We should stay consistent with the term. I think `Transaction private key` is used in monero.

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -718,8 +723,8 @@ portfolio.pending.step3_seller.amountToReceive=Amount to receive
 portfolio.pending.step3_seller.yourAddress=Your {0} address
 portfolio.pending.step3_seller.buyersAddress=Buyers {0} address
 portfolio.pending.step3_seller.yourAccount=Your trading account
-portfolio.pending.step3_seller.xmrTxHash=Tx hash
-portfolio.pending.step3_seller.xmrTxKey=Tx private key
+portfolio.pending.step3_seller.xmrTxHash=Tx id
+portfolio.pending.step3_seller.xmrTxKey=Tx key
 portfolio.pending.step3_seller.xmrTxVerificationError=The XMR transfer validation for trade with ID ''{0}'' failed.\nReason: ''{1}''
```
chimp1984: Seems xmrTxVerificationError is not used anymore

desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/steps/seller/SellerStep3View.java
```diff
@@ -76,8 +76,8 @@
     private BusyAnimation busyAnimation;
     private Subscription tradeStatePropertySubscription;
     private Timer timeoutTimer;
-    private final ChangeListener<XmrProofResultWithTradeId> xmrProofResultWithTradeIdListener;
-
+    TextFieldWithCopyIcon autoConfirmStatusField;
```
chimp1984: should be private

desktop/src/main/java/bisq/desktop/main/settings/preferences/PreferencesView.java
```diff
@@ -895,9 +915,63 @@ private void activateDaoPreferences() {
...
+        {
+            AutoConfirmSettings x = preferences.getAutoConfirmSettings();
+            autoConfirmXmr.setSelected(x.enabled);
+            autoConfRequiredConfirmations.setText(String.valueOf(x.requiredConfirmations));
+            autoConfTradeLimit.setText(formatter.formatCoin(Coin.valueOf(x.tradeLimit)));
+            autoConfServiceAddress.setText(String.join(", ", x.serviceAddresses));
+        }
+
+        autoConfirmXmr.setOnAction(e -> {
+            boolean bEnabled = autoConfirmXmr.isSelected();
```
chimp1984: We dont use type prefixes

core/src/main/java/bisq/core/trade/asset/xmr/XmrTransferProofRequester.java
chimp1984: See ProvidersRepository how it uses @Named(Config.USE_LOCALHOST_FOR_P2P) boolean useLocalhostForP2P to determine if a clearnet address need to be used.

chimp1984: The regex with `^192.*` would not match if a user sets `localhost`.

core/src/main/java/bisq/core/user/Preferences.java
chimp1984: This code need to be moved outside the if/else clause. 
```
 if (getAutoConfirmSettings() == null) {
            prefPayload.setAutoConfirmSettings(new AutoConfirmSettings(
                    false, 5, Coin.COIN.value, DEFAULT_XMR_PROOF_PROVIDERS));
        }
```
Otherwise it will be only set for new users but not for those who have already the prefPayload but not the autoConfirmSettings

desktop/src/main/java/bisq/desktop/main/settings/preferences/PreferencesView.java
```diff
@@ -895,9 +915,63 @@ private void activateDaoPreferences() {
         blockNotifyPortTextField.textProperty().addListener(blockNotifyPortListener);
     }
 
+    private void activateAutoConfirmPreferences() {
+        {
```
chimp1984: Why you use an initializer block here?

core/src/main/java/bisq/core/trade/asset/xmr/XmrProofInfo.java
```diff
@@ -0,0 +1,152 @@
...
+        this.serviceAddress = serviceAddress;
+    }
+
+    // something to uniquely identify this object by
+    public String getKey() {
+        return txHash + "|" + serviceAddress;
+    }
+
+    public XmrProofResult checkApiResponse(String jsonTxt) {
+        try {
```
chimp1984: I think it would be useful to log the full jsonTxt.

chimp1984:
@jmacxx Thanks for the work on the feature! Looks pretty good already.

Some comments: 
- I would recommend to use Capitalistion in comments and log messages. Proper sentences makes logs easier to read.  
- Could you merge current master into your branch, should not cause any conflicts.
- Translation strings should be lower case inside sentence...
- Hex validation is done in several places. Would be better to move to Hex util class. Regex is not needed as encode/decode throws if input is wrong. Just need additional lenght check.
- Maybe we should consider to move the processCounterCurrencyExtraData method in TradeManager to a new class to not over burden the TradeManager too much with this specialized feature. It is quite large already...
- We mix a bit between generice feature and XMR specific. Even there are no concrete plans to add the autoconf feature to other altcoins we might add it (e.g. if it get sponsored by the altocin), so better to have all prepared to not get issues when we add more. UI can easily be changed, but data in protobuf, trade, messages are more problematic. 
- In preferences view the auto conf feature is at the very bottom. We  could decrease vertical space for currency selection to make it more prominent or even move it in front of currency selection. 
- The addressRegexValidator does not allow http:// only the IP or onion. 
- Deleting the service address does not refill it with default only at UI screen change or restart. Its a bit tricky to do that. Usually we used a focusOutHandler on the inputfield and apply it when the user leaves focus. 
- The Auto-confirmed badge label is not veritcally centered. Might be a rounding issue with layout...seems to be 1 pixel off.
- Not sure if needed, but might be nice to see the status information at the sellers pending trade view. e.g. how many confirmations from which services. But not sure if UI space allows that. Can be added also later if we see demand for it.
- For dev testing it would be good to deactivate certain checks, like that the key was not used in the past, the date check and the amount check. Otherwise testers need to create a XMR tx for each test run... The DevEnv.devMode property could be used to ignore those checks. Also would be convenient to autofill the buyers tx id and txkey fields with valid data.
- We should consider to allow emty tx ID and tx key input but show a warning if user does not want to send those data. I am not aware of any privacy issue by revealing those data, but it could be that there are some or at least some traders think there are some. Beside that if the users used a wallet which does not provide those data it would lead to a dispute case as he cannot confirm that he sent the xmr. It is mandatory to use a wallet which supports those data, but not sure if we should enforce disputes where otherwise if the seller recieved it there would not be a dispute.
- We should also state in the info popup that it is best for security and privacy that the user runs his own service node and should provide some easy setup instructions (e.g. install script) . beside that we should make clear that there might be reasons why the severice does not provide the result and the user still need to check in the permitted trade time the state of the trade.
- We should add some more detailed info text outside of the app (e.g. blog, wiki, docs). @m52go what do you think? We should make clear the limitations and context of the service. E.g. security is based on assumption that 2 bonded role owners do not collude to defraud traders. There is a privacy loss as the service operators learn about the matching of an onion address and a XMR tx id and amount. All that can be avoided if user runs own service. And of course it is an optional feature if traders dont feel comfortable with it.
- It would be nice to get some statistics how much the feature is used. We could get number of trades using the feature from the node operators (excluding those who run their own service).  
- Once we are in a more final state we should think about the edge cases like when wallet is not synced.... also to define test use cases would help ourself for dev testing and later the testers for release testing.

chimp1984 (changes requested):

core/src/main/java/bisq/core/trade/asset/xmr/XmrProofInfo.java
ghost: Thanks for the detailed review!  :pray: It'll probably take some time to address all the issues, I'll take care of the smaller nitpicks first and then focus on the few larger ones.

Regarding jsonTxt, its already logged as soon as its received by XmrProofRequester, here's an example:

    Aug-18 09:48:53.116 [XmrTransferProofRequest-bisq.core.trade.asset.xmr.XmrTransferProofRequester@4aa8a732] INFO  b.c.t.a.x.XmrTransferProofRequester: HttpClient{socks5ProxyProvider=bisq.network.Socks5ProxyProvider@19a0e182, baseUrl='http://78.47.61.90:8081', ignoreSocks5Proxy=false} 
    Aug-18 09:48:53.116 [XmrTransferProofRequest-bisq.core.trade.asset.xmr.XmrTransferProofRequester@4aa8a732] INFO  b.c.t.a.x.XmrTransferProofRequester: /api/outputs?txhash=488e48ab0c7e69028d19f787ec57fd496ff114caba9ab265bfd41a3ea0e4687d&address=4ATyxmFGU7h3EWu5kYR6gy6iCNFCftbsjATfbuBBjsRHJM4KTwEyeiyVNNUmsfpK1kdRxs8QoPLsZanGqe1Mby43LeyWNMF&viewkey=6c336e52ed537676968ee319af6983c80b869ca6a732b5962c02748b486f8f0f&txprove=1 
    Aug-18 09:48:54.568 [XmrTransferProofRequest-bisq.core.trade.asset.xmr.XmrTransferProofRequester@4aa8a732] INFO  b.c.t.a.x.XmrTransferProofRequester: {"data":{"address":"e957dac72bcec80d59b2fecacfa7522223b6a5df895b7e388e60297e85f3f867b42f43e8d9f086a99a997704ceb92bd9cd99d33952de90c9f5f93c82c62360ae","outputs":[{"amount":0,"match":false,"output_idx":0,"output_pubkey":"3136c340c7aa017175014a3756d91106395f8db970dd76039fa87bab63e8ccaa"},{"amount":100000000000,"match":true,"output_idx":1,"output_pubkey":"cffdc3ab0c0962e56cef363656be61db585c53cb1750a37e241ced16b4d50a26"}],"tx_confirmations":10437,"tx_hash":"488e48ab0c7e69028d19f787ec57fd496ff114caba9ab265bfd41a3ea0e4687d","tx_prove":true,"viewkey":"6c336e52ed537676968ee319af6983c80b869ca6a732b5962c02748b486f8f0f"},"status":"success"} 
    Aug-18 09:48:59.568 [XmrTransferProofRequest-bisq.core.trade.asset.xmr.XmrTransferProofRequester@4aa8a732] INFO  b.c.t.a.x.XmrTransferProofRequester: confirmations: 10437, xmr txid: 488e48ab0c7e69028d19f787ec57fd496ff114caba9ab265bfd41a3ea0e4687d 
    Aug-18 09:48:59.574 [JavaFX Application Thread] INFO  b.c.t.TradeManager: Received a PROOF_OK message, remaining proofs needed: 0, tradeId 64667 
    Aug-18 09:48:59.574 [JavaFX Application Thread] INFO  b.c.t.TradeManager: Auto confirm was successful, transitioning trade 64667 to next step...

core/src/main/java/bisq/core/user/PreferencesPayload.java
ghost: ~~The check for null settings is handled by protobuf boilerplate code.  It creates a default instance of the settings object.~~

[edit] : never mind, I see your point now.

core/src/main/java/bisq/core/trade/asset/xmr/XmrProofInfo.java
chimp1984: Cool!

ripcurlx:
The boring bot unfortunately will always tag it as `no-priority` as it touches the altcoin repo. Just ignore this for now as it has `priority`.

m52go:
> We should add some more detailed info text outside of the app (e.g. blog, wiki, docs).

Sure, I can write draft points to cover in the next day or two for feedback to make sure material is covered completely and correctly.

ghost:
> We mix a bit between generice feature and XMR specific. Even there are no concrete plans to add the autoconf feature to other altcoins we might add it (e.g. if it get sponsored by the altocin), so better to have all prepared to not get issues when we add more. UI can easily be changed, but data in protobuf, trade, messages are more problematic.

Changed proto definition to allow a list of AutoConfirmSettings.
Commit https://github.com/bisq-network/bisq/pull/4421/commits/4dde9e19cf518af5cd2609cf893d0e03f0b29564

> Maybe we should consider to move the processCounterCurrencyExtraData method in TradeManager to a new class to not over burden the TradeManager too much with this specialized feature. It is quite large already...

Moved the new code out to AutoConfirmManager.java
Commit https://github.com/bisq-network/bisq/pull/4421/commits/13978e33c2e0882ebb291f3a713bdc5d612ef500

> Hex validation is done in several places. Would be better to move to Hex util class. Regex is not needed as encode/decode throws if input is wrong. Just need additional lenght check.

Is this referring to the Transaction ID and key input fields?  I should not restrict them to hex input?

> The addressRegexValidator does not allow http:// only the IP or onion.

`address:port` is all that is needed.  Bisq chooses the protocol (http).

> Not sure if needed, but might be nice to see the status information at the sellers pending trade view. e.g. how many confirmations from which services. But not sure if UI space allows that. Can be added also later if we see demand for it.

It shows the number/expected confirmations.
image

> The Auto-confirmed badge label is not veritcally centered. Might be a rounding issue with layout...seems to be 1 pixel off.

I intentionally oriented the badge slightly above and to the right of the title to make it appear similar to other badges shown on menu items.  I think it looks great that way, but could certainly move it down a bit if that is what you want.

> We should also state in the info popup that it is best for security and privacy that the user runs his own service node and should provide some easy setup instructions (e.g. install script) . beside that we should make clear that there might be reasons why the severice does not provide the result and the user still need to check in the permitted trade time the state of the trade.

Done.  Commit https://github.com/bisq-network/bisq/pull/4421/commits/99146009b6bfdb22e333d11f99c0110436166f71
Will post a screenshot so @m52go can review.

> We should consider to allow emty tx ID and tx key input but show a warning if user does not want to send those data. I am not aware of any privacy issue by revealing those data, but it could be that there are some or at least some traders think there are some. Beside that if the users used a wallet which does not provide those data it would lead to a dispute case as he cannot confirm that he sent the xmr. It is mandatory to use a wallet which supports those data, but not sure if we should enforce disputes where otherwise if the seller recieved it there would not be a dispute.

It allows empty entry; I've also added the warning in that case.  Will post a screenshot so @m52go can review.

> For dev testing it would be good to deactivate certain checks, like that the key was not used in the past, the date check and the amount check. Otherwise testers need to create a XMR tx for each test run... The DevEnv.devMode property could be used to ignore those checks. Also would be convenient to autofill the buyers tx id and txkey fields with valid data.

Done.

> Deleting the service address does not refill it with default only at UI screen change or restart. Its a bit tricky to do that. Usually we used a focusOutHandler on the inputfield and apply it when the user leaves focus.

Done.

> I would recommend to use Capitalistion in comments and log messages. Proper sentences makes logs easier to read.
> Translation strings should be lower case inside sentence...

Duly noted; I have attempted to correct all those instances of mis-capitalization.

ghost:
@m52go 

New settings feature text

image

Warning when XMR trans keys not provided

image

chimp1984:
@jmacxx Thanks for your fast response and implementations! I will try to find time for updated review tomorrow.

core/src/main/java/bisq/core/trade/Trade.java
ghost: It needs an initial value on the trade for the case when auto confirmation is disabled and there will be no event fired.  Clients such as SellerStep3View display the status on screen.

core/src/main/java/bisq/core/user/Preferences.java
ghost: @Emzy sorry to bother you again - any chance you can upgrade your testing service to latest master?  We did another update a few days ago.  https://github.com/moneroexamples/onion-monero-blockchain-explorer/commit/38f2700fd506596c531dee8572338d066255fed9

core/src/main/resources/i18n/displayStrings.properties
ghost: How about "Transaction key" instead?  The term "private key" always raises red flags because we are conditioned to never reveal our private keys.  

- https://www.getmonero.org/resources/user-guides/prove-payment.html refers to it as "the transaction's key".  Their web form calls it a "secret transaction key", slightly less intimidating without the word private.
- monero-wallet-cli also calls it tx key: `get_tx_key TXID`
- https://monero.stackexchange.com/questions/1860/how-do-i-prove-that-a-payment-has-been-sent-to-a-particular-address/1861 they call it transaction key.
- https://github.com/moneroexamples/onion-monero-blockchain-explorer calls it `viewkey`.
- Bisq's contract text for XMR account creation calls it `Transaction key` three times, and also gives synonyms `Tx Key, Tx Secret Key or Tx Private Key`
- https://xmr.llcoins.net/checktx.html does use "private key" though.

assets/src/main/java/bisq/asset/CryptoNoteAddressValidator.java
```diff
@@ -62,7 +62,7 @@ public AddressValidationResult validate(String address) {
     }
 
     // obtain the raw hex version of a monero address
-    public static String addressToHex(String address) {
+    public static String convert(String address) {
```
chimp1984: if u name method more specific it does not require the comment. e.g. convertToRawHex()

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -642,7 +642,11 @@ portfolio.pending.step2_buyer.fasterPaymentsHolderNameInfo=Some banks might veri
 portfolio.pending.step2_buyer.confirmStart.headline=Confirm that you have started the payment
 portfolio.pending.step2_buyer.confirmStart.msg=Did you initiate the {0} payment to your trading partner?
 portfolio.pending.step2_buyer.confirmStart.yes=Yes, I have started the payment
-
+portfolio.pending.step2_buyer.confirmStart.warningTitle=You have not provided proof of payment
+portfolio.pending.step2_buyer.confirmStart.warning=You have not entered the transaction ID and the transaction key.\n\n\
```
chimp1984: Suggestion: portfolio.pending.step2_buyer.confirmStart.warning=You have not entered the transaction ID and the transaction key.\n\n\
  By not providing this data the peer cannot use the auto confirm feature to release the BTC as soon the XMR has been received.\n\
  Beside that, Bisq requires that the sender of the XMR transaction is able to provide this information to the mediator or arbitrator in case of a dispute.

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1240,9 +1244,11 @@ setting.about.shortcuts.sendFilter=Set Filter (privileged activity)
...
-  This can make the trade process more efficient for both peers.  Limits can also be set on the max amount of BTC that can be auto-confirmed.  See the auto-confirm section on this page for more info and to enable the feature.
+setting.info.headline=New XMR Auto confirm feature
+setting.info.msg=When selling BTC for XMR you can use the auto confirm feature to verify that the correct amount of XMR has been deposited to your wallet and automatically confirm the trade, releasing BTC to the counterparty.\n\n\
+  This can make the trade process more efficient for both peers.  Limits can also be set on the max amount of BTC that can be auto-confirmed.  See the auto-confirm settings on this page for more info and to enable the feature.\n\n\
+  Auto confirmation relies on Bisq checking the XMR transaction on an explorer API.  For maximum security it is recommended that you point Bisq to your own explorer service, but there are also services available run by Bisq (the default setting).\n\n\
+  For more information and details on how to setup, see the blog post here: [.....]
```
chimp1984: Suggestions:
setting.info.msg=When selling BTC for XMR you can use the auto confirm feature to verify that the correct amount of XMR has been deposited to your wallet and automatically confirm the XMR receipt and releasing the BTC to the counterparty.\n\n\
  This can make the trade process more efficient for both peers.  Limits can also be set on the max amount of BTC per trade that can be auto-confirmed as well as the number of required confirmations. \n\n\
  Auto confirmation relies on Bisq checking the XMR transaction on an XMR explorer node using the private transaction key provided by the XMR sender. Bisq requests by default from at least 2 nodes operated by Bisq contributors. Both need to provide the same result for triggering the confirmation.
  For maximum security and privacy it is recommended that you run your own XMR explorer node and set that address in the settings.\n\n\
  More information and details on how to setup your own node can be found in the Bisq wiki at: [.....]

@m52go Could you review that?

chimp1984 (changes requested):
We need at least 2 providers for DEFAULT_XMR_PROOF_PROVIDERS. Can you push wiz and emzy that they have the updated versions running? One need to run with clearnet.

Clearnet handling of provider for non-tor dev setup is missing.

In Trade:
Do not set xmrProofResult in constructor as it is only used in special circumstances. Keeping it null is more clear. (-> remove xmrProofResult = new XmrProofResult(0, 0, XmrProofResult.State.FEATURE_DISABLED);)
Makr xmrProofResult transient and @Nullable
setXmrProofResult can be package scope.

  	// xmrProofResult is not persisted yet
    @Getter
    @Nullable
    private transient XmrProofResult xmrProofResult;

    void setXmrProofResult(XmrProofResult xmrProofResult) {
        this.xmrProofResult = xmrProofResult;
        xmrProofResultProperty.setValue(xmrProofResult);
    }
    
We should consider to persist XmrProofResult in trade. for trade history it might be important to have some "proof" if autoconfirm was used or not, as well in dispute cases.
if we use propbug for XmrProofResult the enum should not be used as enum in protobuf as protobuf enums are cumbersome (have global scope). use the name instead. see other enum serialisations...(ProtoUtil.enumFromProto)

Still capitalized:
setting.preferences.autoConfirmXMR=XMR Auto Confirm
setting.preferences.autoConfirmEnabled=Enabled
setting.preferences.autoConfirmRequiredConfirmations=Required Confirmations
setting.preferences.autoConfirmMaxTradeSize=Max Trade Size (BTC)
setting.preferences.autoConfirmServiceAddresses=Service Addresses
setting.info.headline=New XMR Auto confirm feature
filterWindow.disableAutoConf=Disable Auto-confirmation (altcoins)


setXMRTxKeyWindow.txHash=Transaction id -> setXMRTxKeyWindow.txHash=Transaction ID
setXMRTxKeyWindow.txKey=Tx key -> setXMRTxKeyWindow.txKey=Transaction key

The second popup at xmr key input is removing the shaded background. You need to start that second popup with a short delay (UserThread.runAfter with 50 ms or so).
Auto confirm settings should be better visible. Either move it on top or reduce number of entries in currency list.

handleProofResult:
This handling need more thought/work:
 if (!walletsSetup.hasSufficientPeersForBroadcast()) {
            return failure;
        }
        if (!walletsSetup.isDownloadComplete()) {
            return failure;
        }

chimp1984:
@m52go Can we add a short description and install help to the wiki?

@Emzy @wiz Can you run the updated verison of the services (with tx date delivered)? one service need to provide clearnet as well (i think emzy has that already) - for dev testing in non-tor setup.
Could you provide a installation script so users can easily install the service themself? I got lots of troubles on Debian with compiling the tool...Seems on Ubuntu its easier...

I think it should be handled as bonded role. We could reuse DATA_RELAY_NODE_OPERATOR as role for that.

Emzy:
> @Emzy @wiz Can you run the updated verison of the services (with tx date delivered)? one service need to provide clearnet as well (i think emzy has that already) - for dev testing in non-tor setup.

I just build from master.
If we will use this in production, I need to move the explorer to another machine. It's running on a old pc at home.

ghost:
Hi @chimp1984 thank you for taking time to review again.  

> Clearnet handling of provider for non-tor dev setup is missing.

Clearnet in dev mode is by default - no special handling needed.  The SOCKS5 proxy provider is null in dev mode so HttpClient uses clearnet.  To attest to that I've been using emzy's clearnet service for the past few weeks.  The relevant line in HttpClient:
https://github.com/bisq-network/bisq/blob/f45a26ca54b8fcecdc8b071f0f1eca7d042cb4c4/p2p/src/main/java/bisq/network/http/HttpClient.java#L97-L100

If I have misunderstood please correct me.  

I am working on all other points you brought up.  Then I'll merge in latest from master as requested (after squashing my commits).

Cheers.

core/src/main/resources/i18n/displayStrings.properties
m52go: This text assumes it will be shown on the settings screen:
```
setting.info.msg=When selling BTC for XMR you can use the auto-confirm feature to verify that the correct amount of 
XMR was sent to your wallet so that Bisq can automatically mark the trade as complete, making trades quicker for everyone.\n\n
Auto-confirm checks the XMR transaction on at least 2 XMR explorer nodes using the private transaction key provided 
by the XMR sender. By default, Bisq uses explorer nodes run by Bisq contributors, but we recommend running your
own XMR explorer node for maximum privacy and security.\n\n
You can also set the maximum amount of BTC per trade to auto-confirm as well as the number of required 
confirmations here in Settings.\n\n
See more details (including how to set up your own explorer node) on the Bisq wiki: [.....]
```
I expect the wiki entry to be fairly short, mainly covering:
- what the auto-confirm feature does
- how to set up an xmr explorer node
- relevant settings: max btc, number of confirmations, and node address

@jmacxx @chimp1984 is there a reference somewhere for how to set up an XMR explorer node?

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1051,6 +1062,11 @@ setting.preferences.explorer=Bitcoin block explorer
 setting.preferences.explorer.bsq=BSQ block explorer
 setting.preferences.deviation=Max. deviation from market price
 setting.preferences.avoidStandbyMode=Avoid standby mode
+setting.preferences.autoConfirmXMR=XMR Auto Confirm
```
m52go: Small but I suggest using a hyphen, e.g. "XMR Auto-Confirm".

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1228,6 +1244,11 @@ setting.about.shortcuts.sendFilter=Set Filter (privileged activity)
 setting.about.shortcuts.sendPrivateNotification=Send private notification to peer (privileged activity)
 setting.about.shortcuts.sendPrivateNotification.value=Open peer info at avatar or dispute and press: {0}
 
+setting.info.headline=New XMR Auto confirm feature
```
m52go: Small but I suggest using a hyphen here too, e.g. "New XMR auto-confirm feature" or "New XMR Auto-Confirm Feature".

core/src/main/resources/i18n/displayStrings.properties
ghost: The service requires building first [monero](https://github.com/monero-project/monero) and then the [blockchain explorer](https://github.com/moneroexamples/onion-monero-blockchain-explorer) from source.  (Looking in the issues list, many people have had problems building due to the C++ boost library).

The only sources of information I know of are:
https://github.com/moneroexamples/onion-monero-blockchain-explorer#compilation-on-ubuntu-16041804
which further references:
https://github.com/moneroexamples/monero-compilation/blob/master/README.md

I had a terrible time trying to build monero on Ubuntu 16.04 (& failed), but it worked ok on 18.04 following the above instructions.  Once monero was built, the explorer part was a breeze.  To run, start `monerod` wait for it to sync, then run the explorer: `xmrblocks --enable-json-api`

One good point is all that hard work encourages XMR users to become sovereign, running their own node and explorer!

m52go: Thanks! I've collected what I could in this article:
https://bisq.wiki/Draft:Auto-confirming_trades

It's in the draft namespace so it shouldn't be easily visible just yet.

Feel free to suggest any changes, missing details, etc (something I wondered: how is explorer address found/determined for putting into Bisq?).

ghost: :+1:  Looks good to me.  The explorer address will be the address:port of the machine you are running it on, for example `192.168.1.111:8081` if on your LAN, or `localhost:8081` if running on the same machine as Bisq.

chimp1984:
@jmacxx But the onion address is used or at least if you have persisted it before. I will test again later, but when i tested it i got connection to the onion address. We should add the IP address of the clearnet.

ghost:
If the user is in non-tor dev mode they should enter clearnet addresses in the settings.  It is not up to Bisq to pick and choose which addresses are to be used.

core/src/main/resources/i18n/displayStrings.properties
m52go: @jmacxx for what it's worth you can probably include the non-draft wiki URL [in the strings](https://github.com/bisq-network/bisq/pull/4421/files#diff-2b16c2ca0e92ce28cd7268b07a143647R1253) for this PR and I'll make it live when it's merged.

https://bisq.wiki/Auto-confirming_trades

ghost:
> Once we are in a more final state we should think about the edge cases like when wallet is not synced.

I've moved the precondition checks such that if they are not met then auto-confirmation is not started.


> .... also to define test use cases would help ourself for dev testing and later the testers for release testing.

Here is my current list of test use cases.  I'll also put this into the main topic so it is more visible.

- New user (preferences)
- Auto-confirm success
- Premature manual confirm by user
- XMR Tx Id / key not provided
- XMR Tx not found
- Re-used XMR TxId
- Feature disabled in settings
- Feature disabled by filter
- Transaction limit exceeded
- Invalid amount
- XMR Trade date too old
- XMR Trade date newer than Bisq trade date (ok)
- Bisq shut down during Auto-confirmation procedure
- XMR API service not reachable


TBD:

> It would be nice to get some statistics how much the feature is used. We could get number of trades using the feature from the node operators (excluding those who run their own service).

> We should consider to persist XmrProofResult in trade. for trade history it might be important to have some "proof" if autoconfirm was used or not, as well in dispute cases.
> if we use propbug for XmrProofResult the enum should not be used as enum in protobuf as protobuf enums are cumbersome (have global scope). use the name instead. see other enum serialisations...(ProtoUtil.enumFromProto)

core/src/main/resources/i18n/displayStrings.properties
ghost: Done, thanks!

chimp1984:
> If the user is in non-tor dev mode they should enter clearnet addresses in the settings. It is not up to Bisq to pick and choose which addresses are to be used.

Its for devs running the regtest setup. It is just more convenient if the clearnet IP is set instead that devs need to find out where to find that...

ghost:
@wiz :kissing:  Any chance you can start to set up and host an XMR proof service? 
https://bisq.wiki/Draft:Auto-confirming_trades#Set_up_your_own_node

chimp1984:
When setting up a XMR account the popup is very large (exceeding the main window). I guess we should shorten the content as well as add information about the auto-confirm feature.

Screen Shot 2020-08-29 at 17 58 01

m52go:
I re-did the wiki page and made it more broad so that it is now a general reference for Monero trading (i.e., proving and verifying transactions), but that required me to change the page's title and location. 

@jmacxx this means https://bisq.wiki/Auto-confirming_trades doesn't exist anymore...please change that link to https://bisq.wiki/Trading_Monero#Auto-confirming_trades.

New page will be at https://bisq.wiki/Trading_Monero. See the draft at https://bisq.wiki/Draft:Trading_Monero.

Suggestion for pop-up:

```
account.altcoin.popup.xmr.msg=Trading XMR on Bisq requires that you understand the following requirement:\n\n\
If selling XMR, you must be able to provide the following information to a mediator or arbitrator in case of a dispute:\n\
- the transaction key (Tx Key, Tx Secret Key or Tx Private Key)\n\
- the transaction ID (Tx ID or Tx Hash)\n\
- the destination address (recipient's address)\n\n\
See the wiki for details on where to find this information on popular Monero wallets:\n\
https://bisq.wiki/Trading_Monero#Proving_payments\n\n\
Failure to provide the required transaction data will result in losing disputes.\n\n\
Also note that Bisq now offers automatic confirming for XMR transactions to make trades quicker,\
but you need to enable it in Settings.\n\n\
See the wiki for more about the auto-confirm feature:\n\
https://bisq.wiki/Trading_Monero#Auto-confirming_trades
```

This text skips the sections on verifying transactions and wallet-specific details, but this information is now on the wiki.

For some reason my regtest setup is only letting me add BSQ so I cannot show a graphical rendering of the new text box.

ghost:
@m52go Done! https://github.com/bisq-network/bisq/pull/4421/commits/d9d0814064c214d79da64ca90a23aeaf7e331360
image

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1308,35 +1332,18 @@ mediator or arbitrator in case of a dispute.\n\n\
...
-- Monero GUI: change wallet to Advanced mode and go to Advanced > Prove/check > Check Transaction\n\
-- Monero CLI: use the command check_tx_key TXID TXKEY ADDRESS\n\
-- XMR checktx tool (https://xmr.llcoins.net/checktx.html)\n\
-- Explore Monero website (https://www.exploremonero.com/receipt)\n\n\
-If you are still not sure about this process, visit (https://www.getmonero.org/resources/user-guides/prove-payment.html) \
-to find more information or ask a question on the Monero support subreddit (https://www.reddit.com/r/monerosupport/).
+See the wiki for details on where to find this information on popular Monero wallets:\n\
+https://bisq.wiki/Trading_Monero#Proving_payments\n\n\
+Failure to provide the required transaction data will result in losing disputes.\n\n\
+Also note that Bisq now offers automatic confirming for XMR transactions to make trades quicker,\
```
```suggestion
Also note that Bisq now offers automatic confirming for XMR transactions to make trades quicker, \
```

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1308,35 +1332,18 @@ mediator or arbitrator in case of a dispute.\n\n\
...
 or the ArQmA forum (https://labs.arqma.com) to find more information.
-account.altcoin.popup.xmr.msg=Trading XMR on Bisq requires that you understand and fulfill \
-the following requirements:\n\n\
-Prove payments: since Monero is a private coin, some transaction details aren't publicly available \
-in the blockchain, and, in case of a dispute, the mediator or arbitrator needs them to check if the \
-transaction was really made. In Bisq, the sender of the XMR transaction is the one responsible for \
-providing this information to the mediator or arbitrator in case of a dispute. In order to do that, \
-you must send XMR using a wallet that provides the information required to prove the payment was made, \
-which includes:\n\n\
+account.altcoin.popup.xmr.msg=Trading XMR on Bisq requires that you understand the following requirement:\n\n\
```
```suggestion
account.altcoin.popup.xmr.msg=Trading XMR on Bisq requires that you understand the following requirement.\n\n\
```

chimp1984:
@m52go 
> For some reason my regtest setup is only letting me add BSQ

You need to pay the asset listing fee. luckily on regtest its just virtual money ;-). need  a block conf to get active. so  pay fee for xmr, create a block -> u can use xmr.

chimp1984:
@jmacxx Made a PR to your branch.... https://github.com/jmacxx/bisq/pull/1

ghost:
Thanks @chimp1984, I'm AFK for the next few hours, will look at this when I return.

ghost:
@chimp1984 I attempted to merge your pull request on top of my branch but got stuck in too many merge conflicts.  It seems we have been refactoring the code at the same time.  I think the best option would be to close this PR and you open a new one with your current snapshot of the code.

ghost:
I've merged your latest @chimp1984 .  Will test it out.

chimp1984:
Thanks @jmacxx ! I will continue soon as well testing. best to use 2 services for testing and add some artificial delay at the repsonses so u can see how ui behaves... i used same service twice with some small hacks to not ignore it as it has same uid..

I pinged @wiz how his progress is with getting a node. To have 2 nodes is a mandatory requirement for release IMO.

chimp1984:
@jmacxx I tried to do a pR to your branch but it did not work. So I did a new PR to Bisq. https://github.com/bisq-network/bisq/pull/4458

ghost:
Closing as replaced by #4458

PR #4422, opened 2020-08-21
Pricenode tweak: completely remove Coinmate and cex.io as data providers

wiz:
Based on user feedback, these exchanges are too illiquid and do not accurately reflect the true price of the market. After research, I found the user's feedback to be correct and this PR removes both exchanges.


mrosseel:
Wouldn't it be cleaner to set the exchange weights to zero? In this way when the exchange gains relevance or is useful for another crypto pair it's still available. Don't want to block this PR, just something to keep in mind for the future

wiz:
@mrosseel Yes, @cd2357 will likely re-add these exchanges with a new configuration in another PR later, this PR was urgently done to fix $DASH price from being 1% off which caused one user to notice

sqrrm (approved):
utACK

sqrrm:
I think this can be deployed pretty fast as long as price node operators coordinate.

cd2357:
How about just disabling them, instead of removing?

They can be easily disabled by commenting out the `@Component` annotation + evtl adding a short comment hinting that they're temporarily disabled until rate weights are avialable.

cd2357:
Too late, I didn't notice its already merged.

sqrrm:
If you want to, just revert this PR and disable as you suggest. Sounds more appropriate.

PR #4423, opened 2020-08-23
Modify pricenode install script to use master instead of release tag

wiz:
Since the Pricenode Operators always run master and don't care about the main Bisq release tags, the install script should use master as well


ripcurlx (approved):
utACK

PR #4425, opened 2020-08-23
Reduce seednode maxconnections setting from 50 to 40 in bisq.env

wiz:
Due to a potential Tor rate limiting issue, it would be safer to reduce the `--maxconnections` settings on all seednodes from 50 to 40.


ripcurlx (approved):
utACK

PR #4417, opened 2020-08-19
Move gRPC boilerplate from :core to :daemon

ghubstan:
This change moves gRPC boilerplate classes from the `:core.grpc` pkg into a new `:daemon.grpc` pkg.

* The `:core.grpc `pkg was renamed `:core.api`, and no longer has any dependencies on `gRPC` libraries.

* All core service classes in the `:core.api` pkg were made package protected, excepting `CoreApi`, making `CoreApi` the only possible entry point for all `Grpc*Service` -> -`Core*Service` calls.
  
* All `Grpc*Service` classes in the `:daemon.grpc` pkg were made package protected, excepting `GrpcServer`;  the only class depending on the Grpc*Service classes is GrpcServer.
  
* gRPC dependencies were moved from the `gradle.build` file's `:core` sub-project to `:daemon`.  

No changes were made to `apitest`, but this PR's branch is based on the [PR 4366](https://github.com/bisq-network/bisq/pull/4366) branch, so [PR 4366](https://github.com/bisq-network/bisq/pull/4366) ought to be merged before this one.

PR #4427, opened 2020-08-24
Move gRPC boilerplate from :core to :daemon

ghubstan:
This change moves `gRPC` boilerplate code from the `:core.grpc` pkg into a new `:daemon.grpc` pkg.

* The `:core.grpc` pkg was renamed `:core.api`, and no longer has a dependency on `gRPC` libs.

* All core service classes in the `:core.api` pkg are now package protected, excepting `CoreApi`, making `CoreApi` the only possible entry point for all `Grpc*Service -> Core*Service` calls.

* All `gRPC` service classes in the `:daemon.grpc` pkg are now package protected, excepting `GrpcServer`;  the only class depending on `Grpc*Service` class is `GrpcServer`.

* `gRPC` dependencies were moved from the `gradle.build` file's `:core` subproject to `:daemon`.

This PR replaces [4417](https://github.com/bisq-network/bisq/pull/4417).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -328,19 +328,6 @@ configure(project(':core')) {
             exclude(module: 'jackson-annotations')
         }
         implementation "com.google.protobuf:protobuf-java:$protobufVersion"
-        implementation("io.grpc:grpc-protobuf:$grpcVersion") {
-            exclude(module: 'guava')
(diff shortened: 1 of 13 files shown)


sqrrm (approved):
utACK

PR #4428, opened 2020-08-24
Remove duplicate CompositeOptionSet class

ghubstan:
This commit changes the access modifier on `bisq.common.config.CompositeOptionSet` to make it visible to `bisq.apitest.config.ApiTestConfig`, and the duplicate `bisq.apitest.config.ApiTestConfig` was deleted from 'apitest'.

This PR replaces [4417](https://github.com/bisq-network/bisq/pull/4417).


sqrrm (approved):
utACK

PR #4429, opened 2020-08-25
Fix setUpScaffold() signature

ghubstan:
Adds the missing String[] params to the method signature, so test cases can pass any needed combination of options to the scaffolding setup from a `@BeforeAll` method.

_PR 1 of 5, to be reviewed/merged in PR number order._


sqrrm (approved):
utACK

PR #4430, opened 2020-08-25
Move bats based mainnet test from :cli to :apitest

ghubstan:
Move `:cli test.sh` to `:apitest mainnet-test.sh`

* The bats test script was moved to the apitest subproject and renamed.

* Version tests were updated for release 1.3.7.

* The duplicated "test getoffers buy eur check return status" was 
  replaced by a new "test getoffers sell eur check return status" test.

* The bats dependency was switched to bats-core because development
  has halted on https://github.com/sstephenson/bats/tree/master.
  The new bats repository is https://github.com/bats-core/bats-core/tree/master.

_PR 2 of 5, to be reviewed/merged in PR number order._

### DIFF
--- a/apitest/scripts/mainnet-test.sh
+++ b/apitest/scripts/mainnet-test.sh
@@ -1,11 +1,11 @@
 #!/usr/bin/env bats
 #
-# Integration tests for bisq-cli running against a live bisq-daemon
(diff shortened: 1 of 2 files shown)


sqrrm (approved):
utACK

PR #4431, opened 2020-08-25
Remove duplicated grpc stub creation logic

ghubstan:
The `:apitest GrpcStubs` class was removed and recreated in the `:cli`  subproject, to be used by both `:cli` and `:apitest`.  `CliMain` was changed to use the new `GrpcStubs`.

_PR 3 of 5, to be reviewed/merged in PR number order._


sqrrm (approved):
utACK

PR #4432, opened 2020-08-25
Remove redundant ApiTestConfig options

ghubstan:
This change removes three `:apitest` scaffold setup options: `runArbNodeAsDesktop`, `runAliceNodeAsDesktop`, and `runBobNodeAsDesktop`, which should have  been deleted when the `supportingApps` option was added.  The comma delimited list of apps passed with the `supportingApps` option now determines whether `arbitration` / `bob` / `alice` nodes are started as desktops or daemons.

_PR 4 of 5, to be reviewed/merged in PR number order._


sqrrm (approved):
utACK

Issue #2369, opened 2019-02-05
Add stable coins


ripcurlx:
As I plan to add a couple of stable coins for v0.9.4 I'd love to get the feedback of people who are actually using them right now.

Based on my short research based on volume following coins would be on my list:
- Tether (ERC-20 and Omni layer based token (Bitcoin))
  - Official project URL: https://tether.to/
  - Official block explorer URL: https://omniexplorer.info (omni layer), https://etherscan.io/token/0xdac17f958d2ee523a2206206994597c13d831ec7 (ERC-20)
- DAI (ERC-20 based token)
  - Official project URL: https://makerdao.com/en
  - Official block explorer URL: https://etherscan.io/token/0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359
- TrueUSD (ERC-20 based token)
  - Official project URL: https://www.trusttoken.com/trueusd/
  - Official block explorer URL: https://etherscan.io/token/0x0000000000085d4780B73119b644AE5ecd22b376
- USD Coin (ERC-20 based token)
  - Official project URL: https://www.centre.io/usdc
  - Official block explorer URL: https://etherscan.io/token/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48

Is there a coin in the list that shouldn't be added? If yes, why? Is there an important stable coin missing that you're using? Please reply in the comments below - Thanks!

phraktle:
Hi @ripcurlx,

I would love to see Augmint EUR (see https://github.com/bisq-network/bisq/pull/2345). It provides a decentralized, collateralized euro-targeted stablecoin.

ripcurlx:
@phraktle If you change the commit message I'll merge your PR.

ripcurlx:
I've updated the list of stable coins with the requirement for our asset listing. Atm the only problematic one is Tether (surprise, surprise). I'm not sure if we should support both address formats (Omni Layer and ERC-20) as it might cause confusion and arbitration cases in the end. I guess I'll stick to the newer ERC-20 implementation, although the Omni Layer (Bitcoin) based implementation might be the more "secure" one. I start to hate this stable coins already 😉

PR #4434, opened 2020-08-25
Add Tether token support via ERC20 and Omni

pingiun:
Related: https://github.com/bisq-network/proposals/issues/170, #2369

From the proposal and the discussion in Keybase it is clear that the USDT token is wanted on Bisq. This PR adds basic support for the ERC20 based token and the Omni based token. 

I think a discussion should be had about the best way to make this implementation better. As of now, it adds two altcoins and thus two new markets are possible. I think it can be improved by joining these accounts or these markets in some way, for when users want to be able to trade both versions of USDT. On centralised markets there is only one USDT/XBT market, and the two tokens are considered equal. I think this behaviour is desirable.

### DIFF
--- a/core/src/main/java/bisq/core/provider/price/PriceProvider.java
+++ b/core/src/main/java/bisq/core/provider/price/PriceProvider.java
@@ -70,7 +70,12 @@ public class PriceProvider extends HttpClientProvider {
                 final double price = (Double) treeMap.get("price");
                 // json uses double for our timestampSec long value...
                 final long timestampSec = MathUtils.doubleToLong((Double) treeMap.get("timestampSec"));
-                marketPriceMap.put(currencyCode, new MarketPrice(currencyCode, price, timestampSec, true));
+                if (currencyCode.equals("USDT")) {
+                    addPrice(marketPriceMap, "USDT-O", price, timestampSec);
+                    addPrice(marketPriceMap, "USDT-E", price, timestampSec);
+                    addPrice(marketPriceMap, "L-USDT", price, timestampSec);
+                }
+                addPrice(marketPriceMap, currencyCode, price, timestampSec);
             } catch (Throwable t) {
                 log.error(t.toString());
                 t.printStackTrace();
@@ -78,6 +83,10 @@ public class PriceProvider extends HttpClientProvider {
 
         });
         return new Tuple2<>(tsMap, marketPriceMap);
+    }
+
+    private void addPrice(Map<String, MarketPrice> marketPriceMap, String currencyCode, double price, long timestampSec) {
+        marketPriceMap.put(currencyCode, new MarketPrice(currencyCode, price, timestampSec, true));
     }
 
     public String getBaseUrl() {
--- a/desktop/src/main/java/bisq/desktop/main/MainView.java
+++ b/desktop/src/main/java/bisq/desktop/main/MainView.java
@@ -534,11 +534,10 @@ public class MainView extends InitializableView<StackPane, MainViewModel>
 
         String selectedCurrencyCode = model.getPriceFeedService().getCurrencyCode();
         MarketPrice selectedMarketPrice = model.getPriceFeedService().getMarketPrice(selectedCurrencyCode);
-
         return Res.get("mainView.marketPrice.tooltip",
                 "Bisq Price Index for " + selectedCurrencyCode,
                 "",
-                DisplayUtils.formatTime(new Date(selectedMarketPrice.getTimestampSec())),
+                selectedMarketPrice != null ? DisplayUtils.formatTime(new Date(selectedMarketPrice.getTimestampSec())) : Res.get("shared.na"),
                 model.getPriceFeedService().getProviderNodeAddress());
     }
 
--- a/assets/src/main/java/bisq/asset/LiquidBitcoinAddressValidator.java
+++ b/assets/src/main/java/bisq/asset/LiquidBitcoinAddressValidator.java
@@ -1,0 +1,12 @@
+package bisq.asset;
+
+public class LiquidBitcoinAddressValidator extends RegexAddressValidator {
+    static private final String REGEX = "^([a-km-zA-HJ-NP-Z1-9]{26,35}|[a-km-zA-HJ-NP-Z1-9]{80}|[a-z]{2,5}1[ac-hj-np-z02-9]{8,87}|[A-Z]{2,5}1[AC-HJ-NP-Z02-9]{8,87})$";
+    public LiquidBitcoinAddressValidator() {
+        super(REGEX);
+    }
+
+    public LiquidBitcoinAddressValidator(String regex, String errorMessageI18nKey) {
+        super(REGEX, "validation.altcoin.liquidBitcoin.invalidAddress");
+    }
+}
--- a/assets/src/main/java/bisq/asset/coins/LiquidBitcoin.java
+++ b/assets/src/main/java/bisq/asset/coins/LiquidBitcoin.java
@@ -19,12 +19,13 @@ package bisq.asset.coins;
 
 import bisq.asset.AltCoinAccountDisclaimer;
 import bisq.asset.Coin;
+import bisq.asset.LiquidBitcoinAddressValidator;
 import bisq.asset.RegexAddressValidator;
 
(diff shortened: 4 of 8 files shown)


chimp1984 (changes requested):
I guess the price feed will not work with the postfixes. We need some extra handling there...

chimp1984:
You need to add the assets to bisq.asset.Asset file.
USDT seems to be not included in the price feed which is weird as it is a major coin.

chimp1984:
I looked into ways how to support the 2 flavors. I think easiest is to add special hanlding to PriceProvider.getAll method. Check for "USDT" result and add then 2 entries one with USDT-O and one with USDT-E, both same price. So there is a adjusted ticker entry for the price. Atm there are no prices but should be available. 
@wiz Any idea why USDT is not included?

desktop/src/main/java/bisq/desktop/main/MainView.java
```diff
@@ -535,6 +535,10 @@ private String getPriceProviderTooltipString() {
         String selectedCurrencyCode = model.getPriceFeedService().getCurrencyCode();
         MarketPrice selectedMarketPrice = model.getPriceFeedService().getMarketPrice(selectedCurrencyCode);
 
+        if (selectedMarketPrice == null) {
```
chimp1984: I think that is better:

```
private String getPriceProviderTooltipString() {
        String selectedCurrencyCode = model.getPriceFeedService().getCurrencyCode();
        MarketPrice selectedMarketPrice = model.getPriceFeedService().getMarketPrice(selectedCurrencyCode);
        return Res.get("mainView.marketPrice.tooltip",
                "Bisq Price Index for " + selectedCurrencyCode,
                "",
                selectedMarketPrice != null ? DisplayUtils.formatTime(new Date(selectedMarketPrice.getTimestampSec())) : Res.get("shared.na"),
                model.getPriceFeedService().getProviderNodeAddress());
    }```

**Comment by chimp1984** [comment]
Tests are also required.

**Review by chimp1984 — CHANGES_REQUESTED** [review]

**Comment by pingiun** [comment]
> Tests are also required.

What should be tested? There is no real new functionality. Or do you mean the tooltip string?

**Comment by pingiun** [comment]
> I looked into ways how to support the 2 flavors. I think easiest is to add special hanlding to PriceProvider.getAll method. Check for "USDT" result and add then 2 entries one with USDT-O and one with USDT-E, both same price. So there is a adjusted ticker entry for the price. Atm there are no prices but should be available.
> @wiz Any idea why USDT is not included?

I implemented this and amended my commit

**Comment by flix1** [comment]
I suppose that just as in the USD/BTC market you see all offers, including multiple payment methods (Bank wire, Zelle, US Postal Money Order, etc)... you could have a single market for USDT and incluye variants as payment methods (L-USDT, USDT-O, USDT-E). That would be the ideal.

However I would be happy to see a gradual approach in which we follow the path of least resistance and just add 1 as a regular altcoin (I would suggest USDT-Omni). IF there is demand/activity/volume we can improve on it in later releases.

The advantages of a gradual approach with the minimum amount of work are several:

-We increase the chances of having this for the next release.
-We don't waste time and effort before seeing if there is any actual demand for thiss market from Bisq users or new users.
-Devs don't become frustrated if this attempt fails. 
-We allow moderators, etc to have a learning curve for USDT.

**Comment by pingiun** [comment]
@flix1 do you suggest removing USDT-E from this PR and renaming USDT-O to USDT?

**Comment by flix1** [comment]
@pingiun I don't really know how much work each of the options requires. It's your call.

But if it's simpler, faster to just add one option as a regular altcoin... I would suggest USDT-O. That leaves the option open to add USDT-E in the future and creates the least amount of confusion.

**Comment by pingiun** [comment]
With the code in this PR I believe we can just enable USDT-O in the shipping client, leaving the code for the USDT-E token. I'm not sure how the tokens are filtered in the shipped client, does that work via the DAO asset listing?

**Comment by chimp1984** [comment]
@flix1 That is not possible. The payment method for all altcoins is the same. That would be a major refactoring. If there is more need to group assets with diff. sub-flavors we could extend the Assets framework to support that. But I guess it first need to be proven that there is real demand.

**Comment by chimp1984** [comment]
> > Tests are also required.
> 
> What should be tested? There is no real new functionality. Or do you mean the tooltip string?

The addresses. See other assets.

**Review by chimp1984 — DISMISSED** [review]
utACK
Besider missing tests (address validation) it looks good to me. Price providers do not provide yet USDT so price handling cannot be tested yet. @wiz Can we add USDT already to providers?

**Comment by pingiun** [comment]
> > > Tests are also required.
> > 
> > 
> > What should be tested? There is no real new functionality. Or do you mean the tooltip string?
> 
> The addresses. See other assets.

The addresses are already tested, as they are ethereum and bitcoin addresses

**Comment by wiz** [comment]
For starters, the name of the coin isn't "Tether" but "Tether USD" which should be reflected in the class names, filenames, etc. in case we want to add Tether EUR or something in the future. Also, can you please add the Liquid Tether USD token under L-USDT ticker?

As for the price feed, any new assets need to implement their own price feeds (if they want it) since we maintain our own price feeds now.

**Comment by pingiun** [comment]
> For starters, the name of the coin isn't "Tether" but "Tether USD" which should be reflected in the class names, filenames, etc. in case we want to add Tether EUR or something in the future. Also, can you please add the Liquid Tether USD token under L-USDT ticker?

Done, I guessed that L-USDT uses the same addresses as Liquid Bitcoin.

> As for the price feed, any new assets need to implement their own price feeds (if they want it) since we maintain our own price feeds now.

I will look at this in a new PR

File: core/src/main/java/bisq/core/provider/price/PriceProvider.java [hunk 2]
```diff
@@ -70,7 +70,12 @@ public PriceProvider(HttpClient httpClient, String baseUrl) {
                 final double price = (Double) treeMap.get("price");
                 // json uses double for our timestampSec long value...
                 final long timestampSec = MathUtils.doubleToLong((Double) treeMap.get("timestampSec"));
-                marketPriceMap.put(currencyCode, new MarketPrice(currencyCode, price, timestampSec, true));
+                if (currencyCode.equals("USDT")) {
+                    addPrice(marketPriceMap, "USDT-O", price, timestampSec);
+                    addPrice(marketPriceMap, "USDT-E", price, timestampSec);
+                    addPrice(marketPriceMap, "L-USDT", price, timestampSec);
+                }
+                addPrice(marketPriceMap, currencyCode, price, timestampSec);
```
chimp1984, line 78: We are adding USDT additionally as well. So we will end up with 4 price entries for USDT. I am not sure if L-USDT is treated as a seperate altcoin and if preice might differ slightly. I assume so. We should consider to either filter out the USDT-O and USDT-E in the price display code or to not add USDT (so users need to look up USDT-O or USDT-E). As both are the same price I guess its better to filter then out and use USDT.

chimp1984: As this is getting more messy with sepcial cases I would suggest to add generic support.  Add a new field in the asset to indicate the ticker currency for price feed (USDT). That will be used then for picking up price and price display in case the tickersymbol is not available in the price feed.

chimp1984 (changes requested):

core/src/main/java/bisq/core/provider/price/PriceProvider.java
pingiun: On first glance this seems kind of hard because a lot of times currency codes are just passed around as strings instead of objects. But I'll investigate how I can make a nice solution

chimp1984: I will have a look into it.

chimp1984:
I looked into it and it is quite  a larger effort not worth it as long we dont know if it will be really used anyway.
I think we should accept the compromise that we will have 3 same prices for the 3 coins. They might disappear anyway if not traded.

chimp1984:
Sorry could not make a PR to your branch somehow ;-(
Here my PR with small changes: https://github.com/bisq-network/bisq/pull/4457
Can you close that one then?

pingiun:
Closed in favour of #4457

PR #4435, opened 2020-08-26
Add more detailed network statistics

chimp1984:
- Show in network settings UI data sent/received, total messages
sent/received, msg/sec and messages of last second
- Log every minute statistics


chimp1984:
Screen Shot 2020-08-25 at 21 44 26

chimp1984:
Here are statistics after a few hours. Left side is the PR version, right side are 2 apps running old verison.

The difference in number of messages is even larger. About factor 10! Data size of sent data is not much different but also a bit lower (about 5%). Data received is more or less same.

Screen Shot 2020-08-27 at 15 15 47

sqrrm (approved):
utACK

PR #4438, opened 2020-08-26
Reload dispute layout on reopen

sqrrm:
When a dispute is reopened it now shows the chat text entry field. Might help with https://github.com/bisq-network/roles/issues/102#issuecomment-680753125


chimp1984:
ACK

I think one problem is that users dont know the shortcut to re-open the dispute. I will look into that and might add a button to my other PR. For mediators a button instead of the shortcut is for sure a good idea. For users I am not that sure as they might overuse that. But maybe a warning popup might help.... 

It would be good as well that it auto-reopens if a message arrives as well that the number icon is updated, otherwise messages in closed disputes are easily overlooked.

sqrrm:
The mediator can tell the user to use ctrl-u to reopen. Another option is a hover help to explain that, let's not add another popup.

PR #4442, opened 2020-08-27
Exclude spot price provider tests by default

cd2357:
Exclude these tests from the default gradle test task. However, allow the option of having them included by specifying an optional gradle parameter (-Dtest.pricenode.includeSpotProviderTests=true).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -497,6 +497,14 @@ configure(project(':pricenode')) {
 
     test {
         useJUnitPlatform()
+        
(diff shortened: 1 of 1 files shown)


ghost:
It did fix the failing gradlew build.  Git whined a bit:

    git apply ../../fixtestissues.patch
    ../../fixtestissues.patch:19: trailing whitespace.
    warning: 1 line adds whitespace errors.

ripcurlx (approved):
ACK - works for me now - Thanks!

PR #4444, opened 2020-08-28
Use a prog arg for the hour of the restart to avoid the risk that

chimp1984:
multiple seeds restart around the same time which can cause loss of
data.

The option --seedNodeRestartTime=[0-23] is used to determine the hour
when the node restarts. GMT-0 is used. All nodes need to have a synced
clock (timezone does not matter).

A coordinator need to define which seed uses which hour for the restart.


chimp1984:
Superseeded by https://github.com/bisq-network/bisq/pull/4445

PR #4447, opened 2020-08-28
Tweak seednode bitcoin.conf, remove timeout, separate testnet block

wiz:
The `timeout` line of our bitcoin.conf occasionally causes it to get stuck, should remove it from all nodes.
Also separate testnet into its own block

### DIFF
--- a/seednode/bitcoin.conf
+++ b/seednode/bitcoin.conf
@@ -1,13 +1,21 @@
 server=1
+daemon=1
+listen=1
+discover=1
 txindex=1
 dbcache=1337
 maxconnections=1337
-timeout=30000
-listen=1
-discover=1
 peerbloomfilters=1
 onion=127.0.0.1:9050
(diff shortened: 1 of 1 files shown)


wiz:
@emzy this should address the issue you had recently

Emzy:
> @Emzy this should address the issue you had recently
This was the error in the log:
```
ProcessNewBlock: AcceptBlock FAILED (prev-blk-not-found (code 0))
ERROR: AcceptBlockHeader: prev block not found
```

wiz:
@emzy there is some more relevant log message around that, but basically one of your peers failed to send you a block in time, and because of the long `timeout=30000` setting it got stuck and held up your node for a long time. I never meant to commit that line into our configuration, it was a mistake so this PR just removes the timeout, and next time you have a peer with poor connection it will use the default timeout value

ripcurlx (approved):
utACK

PR #4451, opened 2020-08-30
Add check for option trade

chimp1984:
Show message as below to mediators.

Reufnd agent receives option trade info in case the mediator has closed it with reason option trade as part of the result summary (also visible to traders).

Add reason for dispute to summary.

If price is not available (e.g. BSQ) no message gets added. If price feed is not ready we try 3 times with 10 sec. delays before we give up.

Option trade: 
Screen Shot 2020-08-29 at 18 01 13

No option trade:
Screen Shot 2020-08-29 at 18 01 59


sqrrm (approved):
ACK

PR #4453, opened 2020-08-30
Add support for user name for Revolut accounts

chimp1984:
If a user has an existing account with phone number or email as
account ID we show a popup at startup where we require that he sets the
user name. This popup has no close button so he is forced to enter a
value. If there are multiple account multiple popups will be shown.

To not break signed accounts we keep accountId as internal id used for signing.
Old accounts get a popup to add the new required field userName but accountId is
left unchanged. Newly created accounts fill accountId with the value of userName.
In the UI we only use userName.

Input validation does only check for length (5-100 chars). Not sure what
are the requirements at Revolut. Can be changes easily if anyone gets
the specs.

### DIFF
--- a/core/src/main/java/bisq/core/payment/payload/RevolutAccountPayload.java
+++ b/core/src/main/java/bisq/core/payment/payload/RevolutAccountPayload.java
@@ -19,26 +19,36 @@ package bisq.core.payment.payload;
 
 import bisq.core.locale.Res;
 
+import bisq.common.proto.ProtoUtil;
+
 import com.google.protobuf.Message;
 
 import java.nio.charset.StandardCharsets;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 
 import lombok.EqualsAndHashCode;
-import lombok.Getter;
-import lombok.Setter;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 
+import javax.annotation.Nullable;
+
 @EqualsAndHashCode(callSuper = true)
 @ToString
-@Setter
-@Getter
 @Slf4j
 public final class RevolutAccountPayload extends PaymentAccountPayload {
-    private String accountId = "";
+    // Not used anymore from outside. Only used as internal Id to not break existing account witness objects
+    private String accountId = null;
+
+    // Was added in 1.3.8
+    // To not break signed accounts we keep accountId as internal id used for signing.
+    // Old accounts get a popup to add the new required field userName but accountId is
+    // left unchanged. Newly created accounts fill accountId with the value of userName.
+    // In the UI we only use userName.
+    @Nullable
+    private String userName = null;
 
     public RevolutAccountPayload(String paymentMethod, String id) {
         super(paymentMethod, id);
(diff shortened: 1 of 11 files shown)


desktop/src/main/java/bisq/desktop/util/validation/RevolutValidator.java
```diff
@@ -17,10 +17,13 @@
 
 package bisq.desktop.util.validation;
 
-public final class RevolutValidator extends PhoneNumberValidator {
+public final class RevolutValidator extends LengthValidator {
```
sqrrm, line 20: I would like someone with knowledge of this requirement to weigh in before merging.

desktop/src/main/java/bisq/desktop/main/overlays/windows/UpdateRevolutAccountWindow.java
```diff
@@ -0,0 +1,96 @@
...
+        createGridPane();
+        addHeadLine();
+        addContent();
+        addButtons();
+        applyStyles();
+        display();
+    }
+
+    private void addContent() {
+        addLabel(gridPane, ++rowIndex, Res.get("payment.account.revolut.addUserNameInfo", Res.get("payment.revolut.info"), revolutAccount.getAccountName()));
```
sqrrm, line 70: Keeping lines shorter helps the reviewer. and it would also be in accordance with our style guide.

core/src/main/java/bisq/core/payment/payload/RevolutAccountPayload.java
```diff
@@ -95,6 +109,26 @@ public String getPaymentDetailsForTradePopup() {
...
+        this.userName = userName;
+        // We only set accountId to userName for new accounts. Existing accounts have accountId set with email
+        // or phone nr. and we keep that to not break account signing.
+        if (accountId == null) {
+            accountId = userName;
+        }
+    }
+
+    public String getUserName() {
+        return userName != null ? userName : accountId;
```
sqrrm, line 127: Is this safe now that accountId can be null as well?

sqrrm (commented):
I have tested this and it keeps the sign status.

Would be good if someone with revolut knowledge acked before merge.

core/src/main/java/bisq/core/payment/payload/RevolutAccountPayload.java
chimp1984: I reverted accountId to empty string. Is prob. more safe. getUserName returns then also empty string by default at account creation.

desktop/src/main/java/bisq/desktop/util/validation/RevolutValidator.java
chimp1984: We have to consider that even the app forces users to set userName now at startup we should not cause bigger problems if the phone number or email is used (via getuserName). Not sure if validator is used in such a potential use case, but not worth to make usage of the account impossible just because of a more strict validator. Also requirements at Revolut might change and our validation would fail then. I think we should not have a close dependency on that.

sqrrm (approved):
ACK

If new information regarding username requirements emerge a new PR will have to be made.

PR #4454, opened 2020-08-31
Add code coverage reports to Codacy

freimair:
In the course of boosting Bisq's code quality and resilience, here is an attempt to get code coverage reports.

### Use Case "Keep track of test coverage"
We can configure travis to build `master` once a day. By doing this, codacy will display code coverage numbers and we can keep track of how we are doing.

### Use Case "Increase coverage"

Unfortunately, it seems that codacy only shows us numbers without the option to filter for packages. As a workaround, I have added a parameter  `only` to the coverage report task. With that, one can view the coverage for a specific module which holds a "total" of the module instead of the total of the whole project.

Running `./gradlew clean check` followed by `./gradlew codeCoverageReport -Ponly=p2p` will output an html-formatted and easy to read report in `build/reports/jacoco/codeCoverageReport/html/index.html` like so:

Screenshot from 2020-09-02 17-24-34

Given a future project sets its goal to "reach 30% test coverage in p2p", we can easily check if the goal has been reached.

### Todo for @ripcurlx 

- Schedule a daily master build in travis
- fix the env-variable


ripcurlx:
I just restarted the Travis job to see if it is picking up the environment variable correctly.

ghubstan:
Tested ACK

A markdown file in the docs folder showing usage would be nice. 

I used these gradle cmds to see test harness coverage -- it worked.

    ./gradlew  clean  check 
    ./gradlew  build  :apitest:installDaoSetup -x test
    ./gradlew -DrunApiTests=true :apitest:test codeCoverageReport -Ponly=apitest

ripcurlx:
There is still the error in the Travis log that it is not configured properly, although I have done it as documented in Codacy. I'll have a look at it soon again.

ripcurlx:
I did some testing in the past and the config argument was printed correctly, but still it didn't work in the setup. I'll have a more in-depth look soon.

PR #4455, opened 2020-08-31
Simplify witness signing

sqrrm:
This helps developers sign accounts as needed.
1. Do a trade between two accounts that need signing
1. Take trade to mediation
1. Resolve dispute with payout to buyer
1. Open signing tab under Accounts view with ctrl+i
1. Open dispute accounts signing with ctrl+s
1. Choose accounts
1. Add dev privilege key to sign the selected accounts

For mainnet this helps legacy arbitrator to sign without having to register as a legacy arbitrator, just add the private key when doing the signing.


sqrrm:
@ripcurlx not sure what codacy is on about with this PR, looks odd.

ripcurlx:
@sqrrm I just looked at the Codacy complaints in AccountView and SigningView and they are legit IMO.

sqrrm:
Yeah, the complaints were correct, although I don't like the forced use of equals(). Still good that codacy is actually doing some good work.

ripcurlx (approved):
utACK

Will be tested by myself during release testing for v1.3.8

PR #4457, opened 2020-08-31
Add Tether (ERC20, Omni and Liquid)

chimp1984:
Based on https://github.com/bisq-network/bisq/pull/4434 from @pingiun


pingiun (dismissed):

assets/src/main/java/bisq/asset/LiquidBitcoinAddressValidator.java
```diff
@@ -0,0 +1,12 @@
+package bisq.asset;
+
+public class LiquidBitcoinAddressValidator extends RegexAddressValidator {
+    static private final String REGEX = "^([a-km-zA-HJ-NP-Z1-9]{26,35}|[a-km-zA-HJ-NP-Z1-9]{80}|[a-z]{2,5}1[ac-hj-np-z02-9]{8,87}|[A-Z]{2,5}1[AC-HJ-NP-Z02-9]{8,87})$";
+    public LiquidBitcoinAddressValidator() {
+        super(REGEX);
+    }
+
+    public LiquidBitcoinAddressValidator(String regex, String errorMessageI18nKey) {
```
pingiun: this constructor should be removed, and the i18n key should be added to the super call in the default constructor

ripcurlx (approved):
utACK

I'll test this in more detail during release testing.

wiz:
As per my last comment in PR #4434 for Tether USD, **any new assets now need to implement their own price feeds**. This PR should not have been merged without implementing a price feed to go with it, and now ops team has to scramble at the last minute to implement a pricefeed for USDT... in the future @bisq-network/bisq-maintainers please do not merge PRs for new assets unless they implement their own price feed in the pricenode code, and please also coordinate with the @bisq-network/pricenode-operators so they can upgrade and support the new feed before release time 😅

pingiun:
@wiz the next release could still omit USDT right? I don't think we're in a rush to have it in the next client, so if the pricefeed code isn't ready these commits should just not be included in the next release...

flix1:
Hi, this has now been stalled for a while. I believe it is important enough to try to push for it in the next release.

This issue has been in the works since January 2020. See previous proposals here https://github.com/bisq-network/proposals/issues/170 and here https://github.com/bisq-network/bisq/pull/4434

wiz:
The reason that USDT was not added is because there are many USDT tokens on various blockchains / networks so it's not clear which one, and Bisq trade protocol has no way to support tokens that can be settled on different networks, also there was a concern raised that the price of USDT might differ across various USDT tokens if one network has issues etc.
https://wallet.tether.to/transparency

PR #4459, opened 2020-09-01
Remove Slack notification in travis build config

freimair:
While I played around with the travis config in order to get a test coverage report, I noticed that there still is some configuration for Slack. We have not used Slack for quite some time now so I figured it is time to remove that remnant of old times as well.


chimp1984 (commented):
utACK

ripcurlx (approved):
utACK

ripcurlx:
@freimair You need to update your key which signed the commit as it is expired.

freimair:
> @freimair You need to update your key which signed the commit as it is expired.

well that is a bummer. reuploaded the same exact key and all seems well again.

PR #4461, opened 2020-09-01
Make CoreApi a singleton

ghubstan:
This PR ensures a daemon never creates more than one `CoreApi` instance, but two additional, minor changes are included:

 * Add `@VisibleForTesting` annotation to `CompositeOptionSet` to explain why it is no longer package protected.

 *  Remove commented code from `BisqAppMain`.


sqrrm (approved):
utACK

PR #4462, opened 2020-09-01
Register dispute agents in apitest arbitrator daemon

ghubstan:
This PR adds a new `CoreDisputeAgentService` for handling dispute agent registration in arbitrator daemons.  API methods for dispute agent registration in `mainnet` mode are out of scope of this PR;  these changes are for supporting the auto-registration of a mediator and refund agent in `regtest/dao` mode for `apitest` cases.  

Some minor, unrelated changes are included, and listed below.

 * Inject new `CoreDisputeAgentService` into `CoreApi`.

 * Change a `DisputeAgentManager` 'dispute agent saved' log statement to indicate type of agent being saved.

 * Add `import static bisq.common.app.Version.VERSION` to `CoreApi`, and add `@SuppressWarnings("SameReturnValue")` annotation to the `getVersion` method.

 * Fix comment in `Scaffold` class.

[PR 4461](https://github.com/bisq-network/bisq/pull/4461) should be reviewed/merged before this one.


ghubstan:
The codacy issue `Perhaps 'coreDisputeAgentService' could be replaced by a local variable.` won't go away until API methods for registering dispute agents are implemented.   

The field declaration needs to stay as it is for auto-registering dispute agents in test arbitration daemons.

sqrrm:
@ghubstan @ripcurlx This PR can't be merged since codacy complains, and the complaint is legit if this were to be the end state of this work, but since it's a half way state, what do we do about it?

core/src/main/java/bisq/core/api/CoreDisputeAgentService.java
```diff
@@ -0,0 +1,187 @@
...
+
+    public Optional<Mediator> getMediator(NodeAddress nodeAddress) {
+        return mediatorManager.getDisputeAgentByNodeAddress(nodeAddress);
+    }
+
+    public Optional<RefundAgent> getRefundAgent(NodeAddress nodeAddress) {
+        return refundAgentManager.getDisputeAgentByNodeAddress(nodeAddress);
+    }
+
+    private boolean shouldRegisterTestArbitrationDisputeAgents() {
```
sqrrm: This seems odd, having all this test code in a non test class. Seems like part of this is useful for mainnet as well.

sqrrm (commented):
I might not get what's going on, but why are there test methods in CoreDisputeAgentService.java?

core/src/main/java/bisq/core/api/CoreDisputeAgentService.java
ghubstan: Until I write the API methods for registering dispute agents, it can only be done in the UI.  The test harness starts up supporting apps (bitcoind,seednode,arbnode,etc) in separate linux processes, so test cases can only change state in those supporting Bisq apps via the gRPC API.  I'm doing a little triage here, and writing gRPC API support for registering dispute agents from a CLI was moved to the back of the line.  Maybe that's a mistake, but this is my reasoning...

This odd class was a short cut for automatically registering dispute agents for api test cases only.  I did not want to automatically register dispute agents in any daemon but a test arb node.  On mainnet, a user should use the CLI to do this; during testing, a dev (or Travis CI) should not have to run a UI to register dispute agents during the dev/test cycle.

This is a short cut to getting a "simplest trading script" ready for demonstration -- but well tested, and without having to do anything in a UI.  I did not think that an arbitrator CLI + daemon would be in the scope of the "simplest trading script" work (maybe that was a mistake), and I intended to remove this test related code when I do build out the API to support arbitrator CLI + daemons.

It took much more time figuring out when dispute agents can be registered than coding this class.  If I had known already, I could have thought this through better than I did, and built the gRPC API support for registering dispute agents from a CLI instead of this temporary short cut.  If it's not acceptable, I can build the API support next, and remove this test code.

sqrrm: I don't think we should allow mainnet registration of dispute agent via cli for now. There are a lot of very specific dispute handling that I think is handled on the UI side. A UI of some sort would also be needed since it's a human facing role that's done manually.

For testing it's fine to have an automatically registered agent of course. Still would prefer if the test specifics were moved to the tests package.

ghubstan: How about a new API method `registerdisputeagent [dispute-agent-type]` that does not work on mainnet?  It would throw an exception to a mainnet CLI, and the user would see an `Error: dispute agents must be registered in the Bisq UI` msg.

Headless regtest api test cases could use this method.

Sound OK?

sqrrm: Yeah, sounds better. Let's go with that.

ghubstan:
The combined 9 commits below is a fix for the weird short cut I took for automatically registering dispute agents in api test cases.  Now it's done over gRPC with a `registerdisputeagent` method that will not work on mainnet.

It required some refactoring in the test harness so api calls could be made to an arbitration daemon, and there are a few other minor changes:  renaming a class, line formatting in the cli, and removing hard coded "localhost" strings.

core/src/main/java/bisq/core/api/CoreDisputeAgentService.java
ghubstan: The commits are a fix for the weird short cut I took for automatically registering dispute agents in api test cases.  Now it's done over gRPC with a `registerdisputeagent` method that will not work on mainnet.

It required some refactoring in the test harness so api calls could be made to an arbitration daemon, and there are a few other minor changes:  renaming a class, line formatting in the cli, and removing hard coded "localhost" strings.

ghubstan: I have an unresolved file conflict, might need some help tomorrow.

ghubstan:
Closing, a cleaner PR will be created from a branch based off master, with file conflict resolved.

PR #4466, opened 2020-09-03
Use InetAddress.getLoopbackAddress instead of InetAddress.getLocalHost()

chimp1984:
On OSX InetAddress.getLocalHost() can cause problems. On my machine I
get a 5 sec. delay at each start up in localhost mode. The
InetAddress.getLoopbackAddress call is super fast.

Both versions connect to my local btc node, but not sure what can be
the differences if local environment is configured differently as
default. But as those issues with very slow getLocalHost lookup seems
to be more risky I would recommend that we change it.

It only affects localhost regtest mode and users who have a local btc
node running. The detection code for that is using getLoopbackAddress
as well.
See: https://github.com/bisq-network/bisq/pull/4373/commits/a5cca0ee1e63089b9a5a0d8b2793e31006499288

### DIFF
--- a/core/src/main/java/bisq/core/btc/setup/WalletConfig.java
+++ b/core/src/main/java/bisq/core/btc/setup/WalletConfig.java
@@ -61,7 +61,6 @@ import com.google.common.util.concurrent.ListenableFuture;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.Proxy;
-import java.net.UnknownHostException;
 
 import java.nio.channels.FileLock;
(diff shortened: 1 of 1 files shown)


Emzy:
I would advice to always use 127.0.0.1 to connect to local running services.
No need to use a physical interface IP address.

chimp1984:
Thanks @Emzy . Then the getLoopbackAddress is the right approach. Was also changed to that in the local node detection code.

ripcurlx (approved):
utACK

PR #4469, opened 2020-09-03
Add getShowAccountUpdatesNotification which got deleted unintentionally

chimp1984:
at merge.

We keep daoPresentation and accountPresentation support even it is
not used atm. But if we add a new feature and add a badge again it
will be needed.


ripcurlx (approved):
utACK

PR #4471, opened 2020-09-03
Add flags to header to signal which dev setup is used

chimp1984:
Add to app header a post fix with info if Regtest or testnet,
LOCALHOST or DEV MODE is used.
At Mainnet without devmode nothing is added

Screen Shot 2020-09-03 at 12 59 21


ripcurlx (approved):
utACK

PR #4473, opened 2020-09-03
Remove USE_DEV_MODE_HEADER option

chimp1984:
@sqrrm I leave it to you if we remove it or not. But if we leave it please fix that the isDevMode gets overwritten if the new flag is not set.
IMO we dont need an additional flag. See https://github.com/bisq-network/bisq/pull/4471 as well which should help


sqrrm (commented):
NACK

My use case includes running 
```
--useDevMode=false
--useDevModeHeader=true
```

The bug:
`         DevEnv.setDevMode(config.useDevModeHeader);`
was fixed elsewhere.

Closing

PR #4474, opened 2020-09-03
Stop xmr proof service at dispute

chimp1984:
- Stop xmr proof service at dispute
- Check for network and wallet ready
- Add info icon
- Add tx confidence indicator
- UI and text improvements
- Change min required tx from 0 to 1


ripcurlx (approved):
utACK

PR #4475, opened 2020-09-04
Fix bug with all mobile notifications for disputes are sent at startup.

chimp1984:
Before: 
If mobile notifications have been set up at startup for each dispute there was sent a notification.

Now:
Only new mailbox messages trigger a notification at startup.

Any new dispute message triggers a notification as well.


ripcurlx (approved):
utACK

PR #4479, opened 2020-09-04
Improve handling of p2pNetworkAndWalletReady

chimp1984:
The p2pNetworkAndWalletReady MonadicBinding might be removed from GC
if its a local variable. I observed that in BisqSetup with a similar
setup. It might be an implementation weakness in MonadicBinding
(usage of weak references?). A tester reported that he does not see any
result, which might be cause that the service never gets the
onP2pNetworkAndWalletReady triggered if the MonadicBinding is not there
anymore.
By removing the listener we need at shutdown we need it anyway as class
field (so codacy does not complain anymore). As well added a check if
all is already complete to skip the MonadicBinding at all
(not expected case in onAllServicesInitialized).

### DIFF
--- a/core/src/main/java/bisq/core/trade/txproof/xmr/XmrTxProofService.java
+++ b/core/src/main/java/bisq/core/trade/txproof/xmr/XmrTxProofService.java
@@ -148,10 +148,6 @@ public class XmrTxProofService implements AssetTxProofService {
 
     @Override
     public void shutDown() {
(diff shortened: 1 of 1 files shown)


sqrrm (approved):
utACK

Definitely uglier, but let's hope it actually works rather than having intermittent failures.

chimp1984:
@ripcurlx Should be 1.3.8 not 1.3.9

PR #4480, opened 2020-09-04
Remove 'monero:' prefix when adding a XMR account

### LINKED ISSUE #4477: Monero address field validation doesn't support URI prefix "monero:"
From subreddit /r/Monero:
https://www.reddit.com/r/Monero/comments/ilzut0/monero_on_bisq/

```
I want to get some monero from bisq but when i copy/paste the address into the new account section (from GUI) bisq does not accept that format

if I remove "monero" from the beginning it does recognize it though.
```

chimp1984:
Fixes https://github.com/bisq-network/bisq/issues/4477

Some wallets copy the address with a 'monero:' prefix. If user pastes
that directly into the form he gets a validation error. We remove now
that prefix so the input is automatically adjusted to the address only.

Not sure if that can make it into 1.3.8 as the PR is not a bugfix, bu tI guess its such a small issues that we might allow to break the rule ;-). But anyway also not that important to delay to 1.3.9....


sqrrm (approved):
utACK

PR #4481, opened 2020-09-04
Add backward compatibility for revolut change

### LINKED ISSUE #4476: Trading with Revolut between v1.3.7 and v1.3.8 clients will fail
### Description

As we are requiring now a username to trade Revolut based on feedback from Mediators it won't be possible to trade Revolut between v1.3.7 and v1.3.8 clients. This is caused by a breach of contract.

#### Version

v1.3.8

### Steps to reproduce
Use old client (v1.3.7) with Revolut account using phone number and new client (v1.3.8) using a username.

### Screenshots
Bildschirmfoto 2020-09-04 um 11 09 12

This might cause confusion during the transition phase.

We could either force everyone to update to v1.3.8 or mention it in the update message so people are aware of this problem.
The problem with just mentioning is, that it will affect clients updating to v1.3.8 and clients who stay on v1.3.8 causing this error when trying to take an incompatible offer.

If we don't want to do a force update maybe we should filter out at least v1.3.7 Revolut offers for v1.3.8 clients.

chimp1984:
Fixes https://github.com/bisq-network/bisq/issues/4476

New appraoch:

We apply userName to accountId if it is not set (e.g. new account created with new version). We do not use that for display or for account signing in case both fields are the same but we need to use accountId in case the user trades with a not updated user who expects accountId as only field.
I improved a bit the display of account data in the trade screens.
In case accountId was set with the phone number (updated account with phone nr used for account signing) we show both userName and phone nr.

- Show phone number if accountId was set by old account. Otherwise show only userName
- For old users they will see the user name as phone number displayed if they trade with new users if the new user has created a new account. If he has updated an existing account the accountId (phone number) is used, so it displays the phone number.
- At step 2 changed display of own account data to show account name
- Add 'Recipients' prefix to account data of peer at step 2

Step 3: Buyers account data can be
- Phone number if peer is using old version
- User name if peer is updated user with new account (we apply userName to accountId)
- Phone number if user is on old version and peer is updated user with updated account (we keep accountId as phone number)
- User name/Phone number if peer is updated user with updated account

Users/testers who have used master with the new username set, may face unexpected issues. Best is to remove such accounts (if testers). If it happens to traders they can copy salt and create a new account with same phone number with old release version. So the same witness is created for account signing. And then update to master again to set the new user name.


desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/steps/seller/SellerStep3View.java
```diff
@@ -157,13 +154,19 @@ public void activate() {
...
-            trade.getAssetTxProofResultUpdateProperty().addListener(proofResultListener);
+        if (useXmrTxProof) {
+            proofResultListener = (observable, oldValue, newValue) -> {
+                applyAssetTxProofResult(trade.getAssetTxProofResult());
+            };
+
+            // we listen for updates on the trade autoConfirmResult field
+            if (assetTxProofResultField != null) {
+                trade.getAssetTxProofResultUpdateProperty().addListener(proofResultListener);
+                applyAssetTxProofResult(trade.getAssetTxProofResult());
```
sqrrm: This will be done in the next statement anyway

sqrrm:
```suggestion
```

core/src/main/java/bisq/core/payment/payload/RevolutAccountPayload.java
```diff
@@ -105,7 +110,34 @@ public static RevolutAccountPayload fromProto(protobuf.PaymentAccountPayload pro
 
     @Override
     public String getPaymentDetails() {
-        return Res.get(paymentMethodId) + " - " + Res.getWithCol("payment.account.userName") + " " + getUserName();
+        Tuple2<String, String> tuple = getLabelValueTuple();
+        return Res.get(paymentMethodId) + " - " + tuple.first + ": " + tuple.second;
+    }
+
+    private Tuple2<String, String> getLabelValueTuple() {
```
sqrrm, line 117: This is kind of a mess, but seems correct.

core/src/main/java/bisq/core/payment/payload/RevolutAccountPayload.java
```diff
@@ -105,7 +110,34 @@ public static RevolutAccountPayload fromProto(protobuf.PaymentAccountPayload pro
...
+            }
+        } else {
+            label = Res.get("payment.account.phoneNr");
+            value = accountId;
+        }
+        return new Tuple2<>(label, value);
+    }
+
+    public Tuple2<String, String> getRecipientsAccountData() {
+        Tuple2<String, String> tuple = getLabelValueTuple();
```
sqrrm, line 138: Using `var tuple` really helps make the code cleaner and easier to read in cases like this. Although that's personal preference.

sqrrm (changes requested):
The Revolut stuff looks good. I haven't tested it, so utACK on those parts.

desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/steps/seller/SellerStep3View.java
chimp1984: Yes, true. I will add a commit with a more clean handling.

chimp1984: There is more to clean up. See new commit.

core/src/main/java/bisq/core/payment/payload/RevolutAccountPayload.java
chimp1984: I am not a fan of 'var', prefer to have the type information more explicit. I know in such a small context it does not make much difference, but in larger context it does.

core/src/main/java/bisq/core/payment/payload/RevolutAccountPayload.java
chimp1984: Yes, not great all those special cases. Could be done mor elegantly for sure but this feature ate already much more time as expected, so tried to get it fixes as quickly I could ;-)

sqrrm (approved):
utACK

core/src/main/java/bisq/core/payment/payload/RevolutAccountPayload.java
```diff
@@ -116,23 +148,30 @@ public String getPaymentDetailsForTradePopup() {
...
     }
 
-    public boolean userNameNotSet() {
-        return userName == null;
+    public void setUserName(String userName) {
+        this.userName = userName;
+        // We need to set accountId as pre v1.3.8 clients expect the accountId field
+        if (accountId.isEmpty()) {
+            accountId = userName;
+        }
```
ripcurlx, line 175: I just tried to create a new Revolut account and saw that it doesn't behave as expected.

Bildschirmfoto 2020-09-09 um 16 12 09
Bildschirmfoto 2020-09-09 um 16 13 35

The problem is that it is checked at this place if the accountId is empty which will only work for the first character entered into the field e.g. `B` of `Bobby` this ends up with `accountId` set to `B` and userName set to `Bobby`

PR #4484, opened 2020-09-05
Scan disputes for accounts where same user used diff. real names.

chimp1984:


core/src/main/java/bisq/core/support/dispute/agent/FraudDetection.java
```diff
@@ -0,0 +1,215 @@
...
+    private void buildRealNameAccountMaps() {
+        buyerRealNameAccountByAddressMap.clear();
+        sellerRealNameAccountByAddressMap.clear();
+        disputeManager.getDisputesAsObservableList()
+                .forEach(dispute -> {
+                    Contract contract = dispute.getContract();
+                    PubKeyRing traderPubKeyRing = dispute.getTraderPubKeyRing();
+                    String traderPubKeyHash = getTraderPuKeyHash(traderPubKeyRing);
+                    String buyerPubKeyHash = getTraderPuKeyHash(contract.getBuyerPubKeyRing());
+                    boolean isBuyer = contract.isMyRoleBuyer(traderPubKeyRing);
```
sqrrm: Isn't this also known in the next if statement, could just set `true` and `false` instead of `isBuyer` when adding to map.

core/src/main/java/bisq/core/support/dispute/agent/FraudDetection.java
```diff
@@ -0,0 +1,215 @@
...
+    private String getTraderPuKeyHash(PubKeyRing pubKeyRing) {
+        return Utilities.encodeToHex(Hash.getRipemd160hash(pubKeyRing.toProtoMessage().toByteArray()));
+    }
+
+    private void addToMap(String pubKeyHash, Map<String, List<RealNameAccountInfo>> map,
+                          PaymentAccountPayload paymentAccountPayload,
+                          String address,
+                          Dispute dispute,
+                          boolean isBuyer) {
+        if (paymentAccountPayload instanceof PayloadWithHolderName) {
```
sqrrm: Better to return early instead of big if blocks.

sqrrm (approved):
utACK

core/src/main/java/bisq/core/support/dispute/agent/FraudDetection.java
chimp1984: I think you reviewed an older version, not in current version anymore.

core/src/main/java/bisq/core/support/dispute/agent/FraudDetection.java
chimp1984: I will use a stream filter in a new PR...

core/src/main/java/bisq/core/support/dispute/agent/FraudDetection.java
sqrrm: Yeah, started with first commit, then everything changed and I didn't remove this comment.

PR #4486, opened 2020-09-06
Deactivate confirm buttons once mediation started

### LINKED ISSUE #4162: Bisq allows seller to omit mediation
### Description

1. Buyer pays seller XMR. 
2. Seller doesn't confirm or respond chat.
3. Buyer waits one hour and opens mediation. 
4. Mediator demands proof of payment, which is provided by buyer.
5. Seller still doesn't answer to buyer or mediator.
6. Seller has 10 days to release locked funds, omitting mediator's suggestion.

### Expected behaviour

Trade protocol should allow to keep multisig deposit locked when mediation has been requested.
I don't know how to technically solve this flaw, but once trade period expires and mediation is open, both traders should agree on payout before multisig funds are released.

### Actual behaviour

As long as he acts before arbitration, seller doesn't have any incentive to accept mediator suggestion once they have received buyer's payment.

#### Related

https://github.com/bisq-network/bisq/pull/4099
https://github.com/bisq-network/bisq/issues/4064

chimp1984:
Fixes https://github.com/bisq-network/bisq/issues/4162

### DIFF
--- a/core/src/main/java/bisq/core/trade/protocol/TradeProtocol.java
+++ b/core/src/main/java/bisq/core/trade/protocol/TradeProtocol.java
@@ -420,4 +420,14 @@ public abstract class TradeProtocol {
             }
         }
     }
+
+    protected boolean wasDisputed(ErrorMessageHandler errorMessageHandler) {
+        if (trade.getDisputeState() != Trade.DisputeState.NO_DISPUTE) {
+            String msg = "Dispute have been opened once. We do not allow anymore to confirm payment by button click.";
+            log.error(msg);
+            errorMessageHandler.handleErrorMessage(msg);
+            return true;
+        }
+        return false;
+    }
 }
--- a/core/src/main/java/bisq/core/trade/protocol/BuyerAsMakerProtocol.java
+++ b/core/src/main/java/bisq/core/trade/protocol/BuyerAsMakerProtocol.java
@@ -212,6 +212,10 @@ public class BuyerAsMakerProtocol extends TradeProtocol implements BuyerProtocol
     // User clicked the "bank transfer started" button
     @Override
     public void onFiatPaymentStarted(ResultHandler resultHandler, ErrorMessageHandler errorMessageHandler) {
+        if (wasDisputed(errorMessageHandler)) {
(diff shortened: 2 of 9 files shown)


core/src/main/java/bisq/core/trade/protocol/TradeProtocol.java
```diff
@@ -420,4 +420,14 @@ private void cleanupTradableOnFault() {
             }
         }
     }
+
+    protected boolean wasDisputed(ErrorMessageHandler errorMessageHandler) {
+        if (trade.getDisputeState() != Trade.DisputeState.NO_DISPUTE) {
+            String msg = "Dispute have been opened once. We do not allow anymore to confirm payment by button click.";
+            log.error(msg);
```
sqrrm, line 427: Error log is too harsh here, info seems more reasonable.

sqrrm (changes requested):
Just need to fix log level

core/src/main/java/bisq/core/trade/protocol/TradeProtocol.java
chimp1984: I think that is not too harsh as it is not possible to get there as long a user does not write custom code. Will be also more visible in case the API starts to use those code branches and does not do the checks we do in the UI. Maybe even an exception is justified here?

chimp1984: But maybe the method does too much. Should only return a boolean and not handle the true case. I will have a look and maybe add a commit to an open PR.

chimp1984:
@sqrrm Upps, sorry I closed unintentionall and now it does not allow me to reopen. Could you reopen the PR?

chimp1984:
Ah damn I deleted my remote branch ;-(. Will have to fix that somehow and prob need to make a new PR ;-(

chimp1984:
Added again the PR to my repo but Github still does not let me reopen it. I will make a new PR.

chimp1984:
Replaced by https://github.com/bisq-network/bisq/pull/4493

PR #4487, opened 2020-09-06
Remove offer from takers offerbook after take offer is completed.

chimp1984:
We do not wait until the offer got removed by a network remove message but remove it
directly from the offer book. The broadcast gets now bundled and has 2 sec. delay so the
removal from the network is a bit slower as it has been before. To avoid that the taker gets
confused to see the same offer still in the offerbook we remove it manually. This removal has
only local effect. Other trader might see the offer for a few seconds
still (but cannot take it).


sqrrm (approved):
utACK

PR #4490, opened 2020-09-07
Minor bug fixes for XMR tx proof feature

ghost:
- SellerStep3View was retrieving the currency name rather than currency code.  This caused failure to ever display the Auto-confirm status field.
- `useXmrTxProof` flag was used before initialization in cases where `addContent()` is called before `activate()`.
- Fix missing CSS color code `xmr-orange`, was missing from dark mode.
- Fix log message spelling/typo errors.
- Remove address validator from XMR service address settings because it does not support https prefix.

[edit] @chimp1984 wants to work on SellerStep3View.java so I have removed it from this PR.


chimp1984 (dismissed):
utACK

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/4490#pullrequestreview-483209280

PR #4491, opened 2020-09-07
Bug fixes

chimp1984:
- useXmrTxProof was set at active but used in addContent which was
called before activate.
- Fix bug with using currencyName instead of currencyCode
- Refactor: Use methods in base class instead of repeated code


ripcurlx (approved):
utACK

PR #4497, opened 2020-09-07
Add note to SetXmrTxKeyWindow

m52go:
This PR adds a small note on the SetXmrTxKeyWindow about auto-confirm, and it makes clear that the inputs for tx info are optional.

**Before:**

Screenshot from 2020-09-07 11-43-56

**After:**

Screenshot from 2020-09-07 11-33-13


chimp1984 (approved):
utACK

ripcurlx (approved):
ACK

PR #4503, opened 2020-09-08
Xmr tool bugfix: Ignore future results after terminal result

chimp1984:
Ignore future results in case we had a terminal result already.
Avoid that a success result overwrites an earlier failed/error result.


ripcurlx (approved):
utACK

PR #4504, opened 2020-09-09
Use bisq's bitcoinj 0.15.8 - september 2020

oscarguindzberg:
Adapt Bisq to bitcoinj 0.15.8  (see https://github.com/bisq-network/bitcoinj/commits/bisq_0.15.8)
That is a pre-requisite for segwit support.
This PR replaces https://github.com/bisq-network/bisq/pull/4270


oscarguindzberg:
I see codacy suggested a couple of changes, I will fix them ASAP

core/src/main/java/bisq/core/btc/setup/BisqKeyChainGroupStructure.java
```diff
@@ -0,0 +1,80 @@
...
+            ChildNumber.ONE_HARDENED);
+
+    private boolean isBsqWallet;
+
+    public BisqKeyChainGroupStructure (boolean isBsqWallet) {
+        this.isBsqWallet = isBsqWallet;
+    }
+
+    @Override
+    public ImmutableList<ChildNumber> accountPathFor(Script.ScriptType outputScriptType) {
```
chimp1984, line 62: Aren't there more script types (MultiSig) to check?

core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
```diff
@@ -172,7 +172,6 @@ public WalletsSetup(RegTestHost regTestHost,
 
         btcWalletFileName = "bisq_" + config.baseCurrencyNetwork.getCurrencyCode() + ".wallet";
         params = Config.baseCurrencyNetworkParameters();
-        PeerGroup.setIgnoreHttpSeeds(true);
```
chimp1984: Are httpSeed still used by default in BitcoinJ? If so I would prefer to still make sure to not use http seeds as they have some privacy issues.

core/src/main/java/bisq/core/btc/setup/WalletConfig.java
```diff
@@ -495,6 +485,8 @@ private Wallet loadWallet(boolean shouldReplayWallet, File walletFile, boolean i
     }
 
     protected Wallet createWallet(boolean isBsqWallet) {
+        // Change preferredOutputScriptType of btc wallet to P2WPKH to start using segwit
+        Script.ScriptType preferredOutputScriptType = isBsqWallet ? Script.ScriptType.P2PKH : Script.ScriptType.P2PKH;
```
chimp1984: I guess the else case should be P2WPKH

chimp1984 (changes requested):

chimp1984:
@ripcurlx why it got flagged incorrectly (altcoin)?

ripcurlx:
> @ripcurlx why it got flagged incorrectly (altcoin)?

It touches the assets module.

core/src/main/java/bisq/core/btc/setup/BisqKeyChainGroupStructure.java
oscarguindzberg: Even though Bisq uses multisig (eg for payout tx signing), bisq does not use a multisig keychain.
A multisig keychain (MarriedKeyChain in bitcoinj) makes sense when you want to repeatedly generate new multisig addresses for the same group of people.

core/src/main/java/bisq/core/btc/setup/WalletConfig.java
oscarguindzberg: That will be true when we implement segwit.
For now, this is just migration to bitcoinj 0.15
The code is a placeholder for the code that is coming soon (see the comment in the previous line).
I will update the code to:
Script.ScriptType preferredOutputScriptType = Script.ScriptType.P2PKH;
To avoid confusions until segwit is implemented.

core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
oscarguindzberg: http seeds are not used by default in bitcoinj and are not used in bisq as is.
Having said that, I had another look at this and found they might be used if a bisq dev activates a certain feature of bitcoinj that is not being used today. So, I will add PeerGroup.setIgnoreHttpSeeds(true) again.

core/src/main/java/bisq/core/btc/setup/BisqKeyChainGroupStructure.java
chimp1984: Ok, thanks. Did not see the full context.

core/src/main/java/bisq/core/btc/setup/WalletConfig.java
chimp1984: Ah missed the comment.

oscarguindzberg:
I fixed codacy suggestions, but 2 of them. They say I should avoid calling toString() on a String object, but actually Sha256Hash.toString() is being called.

chimp1984:
> I fixed codacy suggestions, but 2 of them. They say I should avoid calling toString() on a String object, but actually Sha256Hash.toString() is being called.

Its in TransactionsListItem.java trade.getDepositTx().getTxId().toString()

chimp1984 (dismissed):
preliminary utACK
But I would like to do a more in depth review in a few days. So better wait for after release with merge and another ACK as its high risk area with wallet code changes.

ripcurlx:
@oscarguindzberg Could you please have a look at the two unnecessary `toString` calls Codacy is complaining about, so this PR is mergeable? Thanks!

chimp1984 (dismissed):
utACK

core/src/main/java/bisq/core/btc/model/AddressEntryList.java
```diff
@@ -105,9 +107,9 @@ public void onWalletReady(Wallet wallet) {
         if (!entrySet.isEmpty()) {
             Set<AddressEntry> toBeRemoved = new HashSet<>();
             entrySet.forEach(addressEntry -> {
-                DeterministicKey keyFromPubHash = (DeterministicKey) wallet.findKeyFromPubHash(addressEntry.getPubKeyHash());
+                DeterministicKey keyFromPubHash = (DeterministicKey) wallet.findKeyFromPubKeyHash(addressEntry.getPubKeyHash(), Script.ScriptType.P2PKH);
```
sqrrm: Prefer to break lines at 120 characters.

gradle/witness/gradle-witness.gradle
```diff
@@ -68,22 +68,23 @@ dependencyVerification {
         'net.glxn:qrgen:c85d9d8512d91e8ad11fe56259a7825bd50ce0245447e236cf168d1b17591882',
         'net.jcip:jcip-annotations:be5805392060c71474bf6c9a67a099471274d30b83eef84bfc4e0889a4f1dcc0',
         'net.sf.jopt-simple:jopt-simple:df26cc58f235f477db07f753ba5a3ab243ebe5789d9f89ecf68dd62ea9a66c28',
-        'network.bisq.btcd-cli4j:btcd-cli4j-core:203156fc63dc1202774de9818e4f21149549f79b25d356b08bb0c784be40c0e8',
-        'network.bisq.btcd-cli4j:btcd-cli4j-daemon:0a2783a851add6e3d8ae899ade48c041b250bfac64b6a4c5f6380ebcdbbe6848',
+        'network.bisq.btcd-cli4j:btcd-cli4j-core:4634b39de93764c4609e295e254e8c3b1427ba24febf43352f4f315029c5b1b3',
```
sqrrm, line 71: Upgraded btcd-cli4j, why was this necessary?

sqrrm (commented):
Tested in regtest and wallet conversion seems to have worked fine. Sending coins work fine, BSQ parsing looks correct.

What I haven't done is completely understood the changes to WalletConfig.java. The rest looks ok to me.

chimp1984 (dismissed):
utACK

oscarguindzberg:
@ripcurlx the two "unnecessary toString() calls Codacy is complaining about" seem like a Codacy bug to me. toString() is being called on Sha256Hash instances, not on String instances.

gradle/witness/gradle-witness.gradle
oscarguindzberg: Kind of long story...
A year ago I upgraded some Bisq dependencies to match bitcoinj 0.15.1 dependencies. eg upgraded guava from 20 to 27.
Then, when bitcoinj 0.15.8 came out with guava 28.2 , I upgraded Bisq to use that guava version.
None of those changes were merged yet, we are merging it as part of this PR.

Some time ago this code was merged into the btcd-cli4j repo https://github.com/bisq-network/btcd-cli4j/pull/4 . And then you reverted part of the change https://github.com/bisq-network/btcd-cli4j/pull/8.

So, I am not actually upgrading btcd-cli4j but downgrading it to the version that declares a dependency on guava 27.

In any case, bisq's gradle resolves not to include the guava version declared in btcd-cli4j but the version declared on its own build.gradle file.

To make things clear a new btcd-cli4j version declaring a dependency to guava 28.2-jre could be created, then we could upgrade to that btcd-cli4j version in bisq.

oscarguindzberg:
@sqrrm @ripcurlx I think this PR is ready for merging. Please, let me know if there is something I did not answer/fix.

core/src/main/java/bisq/core/btc/model/AddressEntryList.java
oscarguindzberg: done

oscarguindzberg:
@sqrrm  I re-requested review from you, I guess that is necessary to merge the PR since I pushed 3 new commits with a bugfix and code styling changes.

sqrrm (dismissed):
ACK

Waiting for @ripcurlx to test this before merging as well. Also need a resolution to the codacy complaints.

chimp1984 (dismissed):
utACK

chimp1984:
@sqrrm @ripcurlx The codady complains are false positives. getTxId() returns a ShaHash and toString() is correct on that. Seems a bug in codacy. Please ignore that and merge or deactivate that rule.

sqrrm:
Impossible to ignore, it blocks merging. So either we're changing codacy rules or we write bad code to get around codacy. It's a pity that's it's so annoying.

oscarguindzberg:
@sqrrm I added a commit https://github.com/bisq-network/bisq/pull/4504/commits/83a1e2f044901e773cd848660f4a08da2ce09e39 to make codacy happy. Check passes now.

sqrrm (dismissed):

oscarguindzberg:
I just signed my commits. I had to push again. I guess the PR requires approval again.

sqrrm (approved):

chimp1984:
> Impossible to ignore, it blocks merging. So either we're changing codacy rules or we write bad code to get around codacy. It's a pity that's it's so annoying.

Thats the real dnager with that too strict and inflexible robot. We bow down and do things we would not do otherwise (for instance i use less exception handling now as it enforces me to write custom exceptions which is often boilerplate, leading at the end to less safe code). We are the masters not codacy.

chimp1984 (changes requested):
NACK (to last commit bowing down to stupid codacy bot)

I think this is really wrong that we start to write less readible code because of that stupid bot. @ripcurlx Please deactivate codacy until we find a more flexible solution. I find this inacceptable that we cannot override clear mistakes or bad settings in codacy.

oscarguindzberg:
@chimp1984 The last commit (https://github.com/bisq-network/bisq/commit/83a1e2f044901e773cd848660f4a08da2ce09e39) does not seem wrong or less readable to me, it is just another way to do the same thing. Please, please, don't NACK this PR... it was so hard to be ready to merge. Let's fix codacy in another issue/PR.

chimp1984 (approved):
utACK

chimp1984:
Ok, I just got too  many times annoyed by a tool which seems to makes us slaves.

cd2357:
> Impossible to ignore, it blocks merging. So either we're changing codacy rules or we write bad code to get around codacy. It's a pity that's it's so annoying.

@sqrrm are you sure?

I played around with Codacy and it can be ignored when merging:

github-repo-merge-despite-codacy

sqrrm:
I guess admins can override codacy, but I'm not an admin so I can't.

cd2357:
Hmm.

Alternatively, Codacy can be removed from the "branch protection" settings.

github-codacy-branch-protection

In that case, the Codacy checks are only informative and merging is possible without admin permissions.

github-codacy-branch-protection-result

Which seems like a sane setup.

cd2357:
Other checks can still be enforced though, like successful travis builds.

sqrrm:
@cd2357 making them recommendations rather than enforced sounds like the more flexible and sane thing to do, all things considered. Thanks for looking that up, @ripcurlx could you make codacy optional? I think it could save some headaches.

PR #4505, opened 2020-09-09
Bugfix: Set username when saving the account not on every text change

ripcurlx:
There is no elegant solution to do this, as we didn't have a input normalization for e-mail, phone numbers in Revolut before and I don't know what is possible to enter as a username as well.
So the workaround is to set the user name only when the account is saved and not on every text change which cause that only the first character will be added as account id in the payload.

See https://github.com/bisq-network/bisq/pull/4481#pullrequestreview-485066342

### DIFF
--- a/desktop/src/main/java/bisq/desktop/components/paymentmethods/RevolutForm.java
+++ b/desktop/src/main/java/bisq/desktop/components/paymentmethods/RevolutForm.java
@@ -40,6 +40,8 @@ import javafx.scene.layout.GridPane;
 
 import lombok.extern.slf4j.Slf4j;
 
+import org.jetbrains.annotations.NotNull;
+
 import static bisq.desktop.util.FormBuilder.addCompactTopLabelTextField;
 import static bisq.desktop.util.FormBuilder.addCompactTopLabelTextFieldWithCopyIcon;
 import static bisq.desktop.util.FormBuilder.addTopLabelFlowPane;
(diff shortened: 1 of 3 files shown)


chimp1984:
I will have a look as well...

desktop/src/main/java/bisq/desktop/components/paymentmethods/RevolutForm.java
```diff
@@ -125,8 +126,23 @@ public void addFormForDisplayAccount() {
...
                 && account.getTradeCurrencies().size() > 0);
     }
+
+    @Override
+    public PaymentAccount getPaymentAccountForAccountCreation() {
+        String userName = getUserName();
+        if (!userName.isEmpty()) {
+            account.setUserName(userName);
+        }
+        return super.getPaymentAccount();
```
chimp1984, line 141: Better use account instead of super.getPaymentAccount(); as that would suggest that this is different than the visible account we used for setting the username.

desktop/src/main/java/bisq/desktop/components/paymentmethods/RevolutForm.java
```diff
@@ -73,7 +75,6 @@ public void addFormForAddAccount() {
         userNameInputTextField = FormBuilder.addInputTextField(gridPane, ++gridRow, Res.get("payment.account.userName"));
         userNameInputTextField.setValidator(validator);
         userNameInputTextField.textProperty().addListener((ov, oldValue, newValue) -> {
-            account.setUserName(newValue.trim());
             updateFromInputs();
```
chimp1984, line 78: I think the problem is that the setUserName method has side effects. I will make an alternative PR with solving the problem at the domain layer. Will be also more useful once used from API.

chimp1984 (commented):
NACK
I will make a PR with an alternative approach

PR #4506, opened 2020-09-09
Fix revolut accountID bug 

chimp1984:
Remove side effect in setUserName method and add extra handling for the moment we save the account. Only at that moment we check if we need to
set the accountId with the value of the userName.
We do that in the domain layer to avoid more domain logic code in the UI
layer.

Fixes bug found at:
https://github.com/bisq-network/bisq/pull/4481#pullrequestreview-485066342


ripcurlx (approved):
utACK

PR #4513, opened 2020-09-10
Revert Tether USD for now due to various issues

wiz:
Tether supports 7 different crypto networks, but Bisq currenly can't properly implement support for a multi-network crypto asset. After some discussion we decided to bump it to the next release until we can implement it properly.

https://wallet.tether.to/transparency
Screen Shot 2020-09-11 at 0 36 08


chimp1984 (approved):
utACK

chimp1984:
I agree we should postpone that until we have a more clear concept.
A way how to support it is to add a "parent currency" field to the Asset framework but that goes further to the TradeCurreny domain and its quite a large code change with high risk.
And then are the conceptual uncertainties:
- Do all USDT flavors have the same price? It is reasonable to assume some are more liquid than others and that might be reflected in price
- Do users mind which flavor they receive? I doubt that this is the case then we need to partition the market anyway or have more complicated supported flavor lists in the accounts and markets. We do similar for accounts with multiple currencies, but it was never used for altcoins and its not trivial both impl wise as from UX.

And most important question: Do we manage to get volume? We had it in the past and there was zero trades. Bisq users tend to dislike the stablecoin idea. We would need the more speculative trader types and not sure if Bisq can compete with centralized exchanges for those type of traders (high volume, lots of trades). 

So lets postpone that until those questions are answered and maybe make a test with the USDT ERC20 flavor only which would be then less complex but would carry the risk to cause backward compatibility issues once we want to extend the support.

ripcurlx (approved):
utACK

PR #4514, opened 2020-09-10
[WIP] Add cancel trade feature

chimp1984:
Implements https://github.com/bisq-network/proposals/issues/251

This PR adds the feature to cancel a trade for teh BTC buyer at trade step 2 (payment started screen).
If he requests for cancellation he will get refunded the min sec. deposit which would be use in a mediation case in case he loses (0.003 BTC). The rest of his sec. deposit goes to the seller. The seller will receive the message and can accept, reject or decide later. If accepted the trade get closed with the payout distribution that the seller receives his trade amount, his deposit + the deposit of the buyer minus the 0.003 BTC. If he rejects the trade continues as normal. The buyer could still click the payment started button at any time and that would trigger the next step in the protocol and the cancelation request is ignored. 
If a dispute gets opened the cancellation is disabled.

The feature needs good testing with the different scenarios.

It comes also with a few refactorings in the trade process code base. Most are not functional changes but some are. So we should get the whole trade process a good test.


core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
```diff
@@ -833,6 +832,67 @@ public Transaction sellerSignsAndFinalizesPayoutTx(Transaction depositTx,
...
+    public byte[] signCanceledTradePayoutTx(Transaction depositTx,
+                                            Coin buyerPayoutAmount,
+                                            Coin sellerPayoutAmount,
+                                            String buyerPayoutAddressString,
+                                            String sellerPayoutAddressString,
+                                            DeterministicKey myMultiSigKeyPair,
+                                            byte[] buyerPubKey,
+                                            byte[] sellerPubKey)
+            throws AddressFormatException, TransactionVerificationException {
+        Transaction preparedPayoutTx = createPayoutTx(depositTx, buyerPayoutAmount, sellerPayoutAmount, buyerPayoutAddressString, sellerPayoutAddressString);
```
sqrrm: Long line is long

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
```diff
@@ -833,6 +832,67 @@ public Transaction sellerSignsAndFinalizesPayoutTx(Transaction depositTx,
...
+        if (myMultiSigKeyPair.isEncrypted()) {
+            checkNotNull(aesKey);
+        }
+        ECKey.ECDSASignature mySignature = myMultiSigKeyPair.sign(sigHash, aesKey).toCanonicalised();
+        WalletService.printTx("prepared canceled trade payoutTx for sig creation", preparedPayoutTx);
+        WalletService.verifyTransaction(preparedPayoutTx);
+        return mySignature.encodeToDER();
+    }
+
+    public Transaction finalizeCanceledTradePayoutTx(Transaction depositTx,
```
sqrrm, line 866: I feel it would be easier and clearer to make more generic sign and finalize methods for the multisig payout. This copy paste of the signing code makes it look like there's something special about this case, but it's just a normal payout and the cause for how the distribution was decided isn't really important.

Just one `finalizeTradePayoutTx` and one `signTradePayoutTx` should be quite enough, reduce code duplication and reduce confusion.

core/src/main/java/bisq/core/trade/protocol/tasks/cancel/FinalizeCanceledTradePayoutTx.java
```diff
@@ -0,0 +1,119 @@
...
+            Contract contract = checkNotNull(trade.getContract(), "contract must not be null");
+
+            checkNotNull(trade.getTradeAmount(), "trade.getTradeAmount() must not be null");
+
+            byte[] mySignature = checkNotNull(processModel.getCanceledTradePayoutTxSignature(),
+                    "processModel.getTxSignatureFromCanceledTrade must not be null");
+            byte[] peersSignature = checkNotNull(tradingPeer.getCanceledTradePayoutTxSignature(),
+                    "tradingPeer.getTxSignatureFromCanceledTrade must not be null");
+
+            boolean isMyRoleBuyer = contract.isMyRoleBuyer(processModel.getPubKeyRing());
```
sqrrm: This is only used by sellercanceltradeprotocol, seems excessive to do these checks.

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -909,6 +910,30 @@ portfolio.failed.cantUnfail=This trade cannot be moved back to open trades at th
...
 portfolio.failed.delayedPayoutTxNull=The trade cannot be completed. Delayed payout transaction is null
 
+portfolio.pending.cancelTrade=Cancel trade
+portfolio.pending.requestSent=Cancel trade request message sent...
+portfolio.pending.requestArrived=Cancel trade request message arrived
+portfolio.pending.requestInMailbox=Cancel trade request message stored in mailbox
+portfolio.pending.requestFailed=Sending cancel trade request message failed
+portfolio.pending.requestGotRejected=Your cancel trade request was rejected by the peer
+portfolio.pending.doNotDecideYet=Don't decide yet
+portfolio.pending.requestCancelTradePopup=You can send a request to your trading peer for canceling the trade.\n\n\
```
sqrrm: Would be good to get @m52go to review this text

desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/steps/HandleCancelTradeRequestPresentation.java
```diff
@@ -0,0 +1,159 @@
...
+                });
+    }
+
+
+    private void onCanceledTradeStateChanged(CanceledTradeState newValue) {
+        log.error("onCanceledTradeStateChanged {} {}", newValue, trade.getId());
+        if (newValue == null) {
+            return;
+        }
+        switch (newValue) {
```
sqrrm: Why such a switch when only one case is used. Could just check that one case with an `if` and avoid 30 lines of code.

proto/src/main/proto/pb.proto
```diff
@@ -325,6 +329,25 @@ message MediatedPayoutTxSignatureMessage {
     NodeAddress sender_node_address = 4;
 }
 
+message RequestCancelTradeMessage {
+    string uid = 1;
+    string trade_id = 3;
+    bytes tx_signature = 2;
```
sqrrm:
```suggestion
    string trade_id = 2;
    bytes tx_signature = 3;
```

core/src/main/java/bisq/core/trade/BuyerTrade.java
```diff
@@ -107,4 +107,15 @@ public Coin getPayoutAmount() {
 
         return getOffer().getBuyerSecurityDeposit().add(getTradeAmount());
     }
+
+    @Override
+    public boolean wasCanceledTrade() {
+        switch (buyersCancelTradeState) {
```
sqrrm: This is how I like switch cases, default for anything not specifically tested

core/src/main/java/bisq/core/trade/Trade.java
```diff
@@ -821,6 +821,8 @@ public void onComplete() {
 
     public abstract Coin getPayoutAmount();
 
+    public abstract boolean wasCanceledTrade();
```
sqrrm: I would prefer `isCanceled()`. It's known that it's a `Trade` already and we use prefix `is` mostly everywhere

core/src/main/java/bisq/core/trade/protocol/tasks/seller/SellerMaybeSignWitnessAndSendWitnessMessage.java
```diff
@@ -90,7 +90,13 @@ protected void run() {
 
             signedWitness = signedWitnessOptional.get();
 
-            // Message sending is handled in base class.
+            // The signer broadcasts as well the signedWitness. In case the receiver will not get the message we have
+            // better resilience that the data is in the network. We set reBroadcast to true but do not expect that the
+            // signedWitness exists already as we just created it.
+            // TODO check with @sqrrm
+            processModel.getP2PService().addPersistableNetworkPayload(signedWitness, true);
```
sqrrm: This seems excessive as this is done during signing already.

sqrrm (changes requested):
During testing I get the accept/reject/decide later popup twice if the trade is not open when receiving the request. First I Accept/Reject and the same popup shows again and I must choose again. Perhaps a listener is attached when switching to that view or something like that.

The buyer should not be able to continue with the trade after the rejection with the current protocol since he's signing the payout with all going to seller. Better might be:
Accept path
1. Buyer proposes cancellation
1. Seller accepts, signs payout and sends signature back
1. Buyer finalizes cancellation

Reject path
1. Buyer proposes cancellation
1. Seller rejects
1. Buyer can continue with trade. No risk to buyer since the seller doesn't have the signed payouttx. No risk to seller since the potential cancellation tx would send all funds to seller.

Adding signedwitness to last message in protocol makes a lot of sense.

The handling of mailboxmessages wasn't obivous to me. Hence the failure to remove them from the network. Would it make sense to make a separate service to pop mailboxmessages from the p2p network store them locally to be processed by the trade protocol later. That way there is no need to worry about the post trade tast mailbox handling.

Added some minor comments inline.

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
chimp1984: Will make a line break...

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
chimp1984: Yes, agree. Will look into it.

core/src/main/java/bisq/core/trade/protocol/tasks/cancel/FinalizeCanceledTradePayoutTx.java
chimp1984: Which check do you mean? isMyRoleBuyer is used to set buyer/seller sig correctly.

desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/steps/HandleCancelTradeRequestPresentation.java
chimp1984: I think you reviewed an old version.

core/src/main/java/bisq/core/trade/Trade.java
chimp1984: Agree. Changed.

core/src/main/java/bisq/core/trade/protocol/tasks/seller/SellerMaybeSignWitnessAndSendWitnessMessage.java
chimp1984: Its an outdated version again. Not sure what you mean. 
```
 if (accountAgeWitnessService.isSignWitnessTrade(trade)) {
            // Broadcast is done in accountAgeWitness domain.
            accountAgeWitnessService.traderSignAndPublishPeersAccountAgeWitness(trade).ifPresent(witness -> signedWitness = witness);
        }
```

Is in the new code, part of the SellerSendPayoutTxPublishedMessage

chimp1984:
> During testing I get the accept/reject/decide later popup twice if the trade is not open when receiving the request. First I Accept/Reject and the same popup shows again and I must choose again. Perhaps a listener is attached when switching to that view or something like that.
> 
> The buyer should not be able to continue with the trade after the rejection with the current protocol since he's signing the payout with all going to seller. Better might be:
> Accept path
> 
>     1. Buyer proposes cancellation
> 
>     2. Seller accepts, signs payout and sends signature back
> 
>     3. Buyer finalizes cancellation
> 
> 
> Reject path
> 
>     1. Buyer proposes cancellation
> 
>     2. Seller rejects
> 
>     3. Buyer can continue with trade. No risk to buyer since the seller doesn't have the signed payouttx. No risk to seller since the potential cancellation tx would send all funds to seller.
> 
> 
> Adding signedwitness to last message in protocol makes a lot of sense.
> 
> The handling of mailboxmessages wasn't obivous to me. Hence the failure to remove them from the network. Would it make sense to make a separate service to pop mailboxmessages from the p2p network store them locally to be processed by the trade protocol later. That way there is no need to worry about the post trade tast mailbox handling.
> 
> Added some minor comments inline.

The popup comes everytime when you activate the pending trades UI and select the trade. Once you accept/reject it does not come again. 
Did you get 2 popups?

Re buyers signature:
Good catch! Yes that was a real security flaw. Thanks for catching it! Of course the buyer must not send the signature first but the other way round as the seller gets the funds back. I will change that and yes we need a 3rd step so that the buyers signs as the end and publishes the tx.

Re mailbox msg:
Yes that might be a good idea, then we have less risk that there are dangling msgs. We just need to add a new store for that and manage it locally. I will have a look into that once the persistence PR is merged.

Currently we allow the buyer to continue with the trade once he opened a cancel request. I think that is dangerous as it might cause conflicts in the trade protocol if 2 branches can be executed in parallel. As well it might lead to more complex situations and carries higher risk that it could be abused. So I think better to deactivate that the buyer can continue normally until the seller has replied.

chimp1984:
Better wait with further review until I have fixes the protocol. Will put it WIP until its ready again for review.

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
sqrrm: This was just a general comment that long lines are hard to read.

core/src/main/java/bisq/core/trade/protocol/tasks/cancel/FinalizeCanceledTradePayoutTx.java
sqrrm: Yeah, why is there a check for what role is active when we know it's the seller?

core/src/main/java/bisq/core/trade/protocol/tasks/seller/SellerMaybeSignWitnessAndSendWitnessMessage.java
sqrrm: Yeah, comment was made on old commit, not on final version. I comment as I go through the code...

sqrrm:
Regarding the popup: yes, it appears twice, ie, it reappears when accept or reject is pressed. This happens only when the trade view is not active when the cancel request is received.

core/src/main/java/bisq/core/trade/protocol/tasks/cancel/FinalizeCanceledTradePayoutTx.java
chimp1984: Ah ok.

chimp1984:
Will close for now as the risks and effort have been much larger as initially thought for a rather small and maybe even controversial feature. Might open a new PR again once I dedicate time to continue on that.

PR #4516, opened 2020-09-11
[WIP] Add check for refund agent if donation address is valid

chimp1984:
Replaces #4136

Here is a list of what got implemented:

Dispute:
- Add donationAddressOfDelayedPayoutTx

DisputeManager:
- Add ObservableList disputesWithInvalidDonationAddress
- At onOpenNewDisputeMessage which is called only at disputeAgents we check for invalid donation addresses and add it to the disputesWithInvalidDonationAddress ObservableList
- At onPeerOpenedDisputeMessage which is called only for the non-opening peer we do a validation to see if the donation address is valid. If not we log an error. No action done here as it is not the mandate of the peer to react to that.

DelayedPayoutTxValidation:
- We add check if dispute donation address matches the one from the trade tx. As well as added a isValidDonationAddress method for checking if the address matches any past donation addresses. 

ContractWindow:
- Add a text field to show donation address

DisputeSummaryWindow:
- We show a popup in case there is an invlaid dontion address
- At payout we return in case the donation address is invalid and it is the refund agent. For mediators we allow payout to be more flexible. But they should consultate devs is such happens.

PendingTradesDataModel:
- At doOpenDispute we validate the delayed payout tx and if invalid and if refund agent dispute we do not allow to continue (no tx broadcast, no dispute opened). For mediation we allow to be more flexible in case its a bug, but we log the issue. The mediator will get the popup and can react accordingly.
- We set additionally the delayedPayoutTx to the dispute for mediation disputes (was not set before) as that will be useful for validation later.

DisputeAgentView:
- We listen on disputesWithInvalidDonationAddress list changes and show a popup. This is global for all trades but trade ID is shown in popup, so agent can assign the problem to a dispute.


chimp1984:
I tested the feature quite intense. It is a bit tricky to test as it requires to manipulate the code so that the address can be changed and that validation on the peers side is not failing. As I don't expect that tester or other devs are going to do that effort I think a smoke test that the standard use cases are ok should be enough.
If anyone wants to test more in detail ping me and I can give some hints.

chimp1984:

PR #4517, opened 2020-09-12
[WIP] Dispute agents sign summary

chimp1984:
Depends on #4518

Adds signature and node address of dispute agent to summary message. With the node address we can look up the pubKey. The summary msg can be pasted into a tool to check if the signature is valid. 

This will help once refund agent revokes and reimbursement request need to be done. To make the statement of the requester verifiable they need to paste the summary message of the mediation and arbitration into the Github request. With that anyone can verify in the app if it is valid. 

Screen Shot 2020-09-11 at 23 55 42

Screen Shot 2020-09-12 at 00 16 44


wiz:
Don't you need to add the full Trade ID into the message?

wiz:
Don't you need to add the full Trade ID into the message? What about the payout addresses and other important info? Just signing "payout is X" is not very useful by itself.

chimp1984:
@wiz Agree. Will add more details to summary text. For traders and dispute agents it is a but redundant as they have the dispute details available, but now its for sharing with otheres. Maybe I use the report instead of the summary text.

chimp1984:
Screen Shot 2020-09-12 at 20 09 45

chimp1984:

PR #4518, opened 2020-09-12
[WIP] Allow refund agent close without payout

chimp1984:
Depends on #4516

Allows now that the refund agent can set both payouts to 0. In that case no payout tx will be created and the dispute is closed. The traders can do a reimbursement request from the DAO in such cases.


chimp1984:

Issue #4386, opened 2020-07-30
Add download option + instructions for QubesOS


cd2357:
Looking at the [GitHub issues](https://github.com/bisq-network/bisq/issues?q=is%3Aissue+qubes), it seems there are quite a few QubesOS users of Bisq.

Since the Qubes setup is not very straightforward, it would help to add a section on the Downloads page with instructions specific to QubesOS.

Here are the instructions I came up with sofar.

---
### 1. Create a new Qube `bisq-vm` based on the `debian-10` template

* Q Menu > Create Qubes VM > Qube based on a template
* Template: choose `debian-10`
* Check "Launch settings after creation"
* Press OK

Within a few seconds, the Qube is created and the Qube Settings popup opens. Adjust it as follows:

* Basic > Disk Storage > Private Storage max size
  * Set this to at least `5000MiB`
  * Assuming you only use this Qube for Bisq, this amount of storage will be used for all Bisq files (downloaded installer, installation folder, your local Bisq settings and profile, your Bisq wallet, Bisq logs, etc). Therefore it is important to have enough disk space available, because if Bisq runs out of storage space, this may lead to data corruption and loss of funds.
  * For reference: A fresh Qube containing only a fully sync-ed new installation of Bisq 1.3.6, with no trades / accounts / balances, uses about 1GB of space.
* Advanced > Memory / CPU
  * Set "Initial memory" `6000 MB`
    * Bisq needs 4 GB, the rest is for other OS processes
  * Set "vCPUs no." to `4`
  * Uncheck "Include in memory balancing"
    * This will ensure that the RAM allocated to this Qube is not shared with other Qubes
* Press OK

### 2. Setup binding dirs mapping for the system folders needed for the Bisq installation

```
sudo mkdir -p /rw/config/qubes-bind-dirs.d
```

### 3. Specify which system folders are mapped to binding dirs

Create the configuration file:

```
sudo nano /rw/config/qubes-bind-dirs.d/50_user.conf
```

Add the following lines in the file:

```
binds+=( '/opt/Bisq' )
binds+=( '/usr/share/desktop-directories' )
```

Save and close the file.

### 4. Create the binding dirs needed for the Bisq installation

```
sudo mkdir -p /rw/bind-dirs/opt/Bisq
sudo mkdir -p /rw/bind-dirs/usr/share/desktop-directories
```

### 6. Download the Debian package

Go to the Bisq Downloads page https://bisq.network/downloads and download the Debian package.

[...]

PR #4520, opened 2020-09-13
Show info popup if Bisq started under Qubes OS

cd2357:
If Bisq is started under Qubes, show popup informing the user about the OS-specific setup guide.

Implements suggestion from https://github.com/bisq-network/bisq/issues/4386#issuecomment-691240207

---

Screenshot of the popup shown on startup:

Screenshot from 2020-09-13 13-30-54


chimp1984 (approved):
utACK

ripcurlx (approved):
utACK

PR #4521, opened 2020-09-13
Increase timeouts

chimp1984:


ghost:
Could we increase `DEFAULT_BROADCAST_TIMEOUT`?  It seems many transactions currently result in an error `Broadcast of tx [...] not completed after 5 sec.`

sqrrm:
@jmacxx Looking at the `DEFAULT_BROADCAST_TIMEOUT` we probably have to do some more work to make it safe to increase
```java
    // Currently there is a bug in BitcoinJ causing the timeout at all BSQ transactions.
    // It is because BitcoinJ does not handle confidence object correctly in case as tx got altered after the
    // Wallet.complete() method is called which is the case for all BSQ txs. We will work on a fix for that but that
    // will take more time. In the meantime we reduce the timeout to 5 seconds to avoid that the trade protocol runs
    // into a timeout when using BSQ for trade fee.
    private static final int DEFAULT_BROADCAST_TIMEOUT = 5;
```

sqrrm (approved):
utACK

chimp1984:
According the logs there are expected timeouts for BSQ txs. We should look into the issue inside BitcoinJ to fix that or even reduce the timeout here, or handle it differently if BSQ or BTC only....

PR #4522, opened 2020-09-14
[WIP] Redesign persistence save to file task

chimp1984:
The model we borrowed from BitcoinJ wallet persistence comes with some complexity and performance costs if the queueUpForSave was called frequently (which is the case). We started for each queueUpForSave call a new thread and most returned without doing work if the persitable data was written by a concurrent task already.

A periodic timer task is a much more simple model. The costs for creating a new task is higher than a runnable execute without doing work in case that there was no new queueUpForSave call.
It also simplifies the usage for clients as the period for persiting is set at intitialisation and cannot be changed anymore (was a bit messy with many different delay values). The period for persistence should reflect the frequency of data changes and the importance that the data is the latest version if the user shuts down quickly. 

There is now also a saveNow method in case immediate write is needed. This comes as sync or async with completeHandler so we dont need to add a safety delay before continuing (e.g. shutdown).

### DIFF
--- a/common/src/main/java/bisq/common/storage/FileManager.java
+++ b/common/src/main/java/bisq/common/storage/FileManager.java
@@ -31,25 +31,23 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.PrintWriter;
 
-import java.util.concurrent.Callable;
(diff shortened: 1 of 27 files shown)


chimp1984:
Replaced by https://github.com/bisq-network/bisq/pull/4527

PR #4524, opened 2020-09-14
Add core support for registration of regtest dispute agents via API

ghubstan:
This PR does not support registering dispute agents on daemons running on the mainnet chain.

This PR replaces [4462](https://github.com/bisq-network/bisq/pull/4462).


core/src/main/java/bisq/core/api/CoreDisputeAgentsService.java
```diff
@@ -0,0 +1,145 @@
...
+        switch (disputeAgentType) {
+            case "arbitrator":
+                throw new IllegalArgumentException("arbitrators must be registered in a Bisq UI");
+            case "mediator":
+            case "refundagent":
+                NodeAddress nodeAddress = new NodeAddress(
+                        InetAddress.getLoopbackAddress().getHostAddress(), config.nodePort);
+                List<String> languageCodes = Arrays.asList("de", "en", "es", "fr");
+                ECKey ecKey = mediatorManager.getRegistrationKey(registrationKey);
+                String signature = mediatorManager.signStorageSignaturePubKey(Objects.requireNonNull(ecKey));
```
sqrrm: Both of these work the same at the moment, but it would be correct to use the specific manager depending on case. Better make a disputeAgentManager and set to whichever type is being registered, or just move the specific registration code to their switch cases.

core/src/main/java/bisq/core/api/CoreDisputeAgentsService.java
```diff
@@ -0,0 +1,145 @@
...
+                languageCodes,
+                new Date().getTime(),
+                ecKey.getPubKey(),
+                signature,
+                null,
+                null,
+                null
+        );
+        mediatorManager.addDisputeAgent(mediator, () -> {
+        }, errorMessage -> {
```
sqrrm, line 118: This is some ugly indentation, seems it's what we have specified as our standard though, and codacy agrees.

sqrrm (changes requested):

core/src/main/java/bisq/core/api/CoreDisputeAgentsService.java
ghubstan: A fix is in commit 57423f4.

core/src/main/java/bisq/core/api/CoreDisputeAgentsService.java
ghubstan: I thought the same about the indentation in lines 117-119.

sqrrm (approved):
utACK

PR #4525, opened 2020-09-14
Refactor apitest for api calls to any daemon type

ghubstan:
These changes fix the test harness so test api calls can be made to any type of daemon (arbitrator, bob, alice), not just alice.  A new `DisputeAgentRegistrationTest` case is included in these changes -- it needs to send requests to an arbitration daemon.

This PR should be reviewed & merged after [4524](https://github.com/bisq-network/bisq/pull/4524).


apitest/src/test/java/bisq/apitest/method/RegisterDisputeAgentsTest.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Test
+    @Order(1)
+    public void testRegisterArbitratorShouldThrowException() {
+        var req =
+                createRegisterDisputeAgentRequest("arbitrator");
+        Throwable exception = assertThrows(StatusRuntimeException.class, () ->
+                grpcStubs(arbdaemon).disputeAgentsService.registerDisputeAgent(req));
+        assertEquals("INVALID_ARGUMENT: arbitrators must be registered in a Bisq UI",
```
sqrrm, line 61: Relying on string matching for tests makes them fragile. At a minimum we should probably add a note for strings that are used in this way.

sqrrm (approved):
utACK

Looks ok, but relying on strings for the tests seems fragile as noticed.

apitest/src/test/java/bisq/apitest/method/RegisterDisputeAgentsTest.java
ghubstan: I will fix this in another PR.

sqrrm: Yeah, I think this happens quite a bit in other tests as well iirc. Not urgent, but good to have a better solution for the future.

ghubstan: This string matching problem has bitten me a few times, when I misspelled "bitciond" in some test case setups.  
PR [4540](https://github.com/bisq-network/bisq/pull/4540) fixes it.

PR #4527, opened 2020-09-15
[WIP] New persistence model

chimp1984:
While reseaching for the app-freeze problem a user reported I looked closer to the changes in our persistence framework. There have been added thread-safety to PersistableEnvelope classes. While trying to figure out if those changes might cause a deadlock in some rare edge cases I came to the conclusion that the overall model is not reflecting well our requirements. Worse, it had a lot of resource waste by creating new threads very frequently which got terminated mostly without doing actual work.
Beside that I questioned the need to persist the data that frequently. We read data only at startup (with a few irrelevant exceptions). Writing the data at shutdown would be sufficient for most type of data. The resource heavy data are the DaoState and the TradeStatistics as well as SequenceNumberMap which gets updated very frequently. All those data have very low priority and can be recovered in case of a sever crash. For more critical data like disputeList or trades we use a timer which writes to disk a minute after a persist request got called. We could improve that by calling a persistNow just at the moments when critical state changes happen (e.g. in a trade). Atm I left all the write requests as it was before and they basically trigger a request at any data change of an object.
To run those operations on the user thread avoids all the issues with threading and risk of deadlocks. The total performance is for sure better as before as it removed a lot of wasted efforts. If really needed we could introduce a threaded version as well with callback handling, but I guess that is not needed.

The changes are quite heavy over many areas. Some are pure refactoring but many are also design changes, like that the data objects are not responsible themselves anymore for persistence but their owner (e.g. trademanager not trade). So that made the overall architecture more clear.
 
I leave it WIP for now as it needs much more testing and I might add persistNow calls for the critical state changes and might eliminate request calls for  irrelevant data changes. There is also a test broken which is commented out atm and should be re-enabled. Not my forte dealing with unit test framework magic...


chimp1984:
Replaced by https://github.com/bisq-network/bisq/pull/4589

PR #4529, opened 2020-09-15
Make moving average code more safe

chimp1984:


ripcurlx (approved):
ACK

Tested it for a couple of currencies on Mainnet and didn't experience any problems.

PR #4534, opened 2020-09-17
Give core api a simple way to verify init status

ghubstan:
This change adds a new `StatusCheck` class, for use by any `Core*Service` instance needing to verify the server is ready to perform trading related functions.

Currently implements one method, `verifyCanTrade`, that throws an IllegalStateException with a CLI friendly error message if any of the following checks fail:

- the p2p network is bootstrapped
- the block chain sync is complete
- the minimum # of peers are connected
- the wallet is available
- the wallet balance is available (not null)


chimp1984 (approved):
utACK

sqrrm (approved):

PR #4540, opened 2020-09-19
Reduce apitest harness' dependency on string matching

ghubstan:
- Pass list of supporting app names as enum vararg to scaffold setup.
- Use dispute agent type constants in RegisterDisputeAgentsTest case.


apitest/src/test/java/bisq/apitest/method/RegisterDisputeAgentsTest.java
```diff
@@ -42,10 +44,14 @@
 @TestMethodOrder(OrderAnnotation.class)
 public class RegisterDisputeAgentsTest extends MethodTest {
 
+    private static final String ARBITRATOR = "arbitrator";
```
chimp1984, line 47: I guess those constants would be good to have in the domain classes and maybe enums are even better.

chimp1984 (approved):
utACK

sqrrm (approved):

PR #4542, opened 2020-09-21
Refactor regexvalidator

chimp1984:
Moves regex validator to core so it can be used from core classes


ghubstan:
ACK

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/4542#issuecomment-696162771

PR #4543, opened 2020-09-21
Add signature to dispute result and various other improvements

chimp1984:
Depends on #4542

Add various new features to dispute domain.

- Add signature to summary message and validation popup so the signature and the msg can be validated
- Allows refund agent to close cases without a payout
- Adds validation for node addresses, deposit and payout txs and donation address
- Add copy as csv button at report screen for dispute agents
- Improves search field and shows tooltip if match is found from which source the match was made (e.g. deposit Tx, json,..)


desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/PendingTradesDataModel.java
```diff
@@ -536,6 +538,25 @@ private void doOpenDispute(boolean isSupportTicket, Transaction depositTx) {
...
+                    donationAddressString::set);
+        } catch (DelayedPayoutTxValidation.DonationAddressException |
+                DelayedPayoutTxValidation.InvalidTxException |
+                DelayedPayoutTxValidation.InvalidLockTimeException |
+                DelayedPayoutTxValidation.MissingDelayedPayoutTxException |
+                DelayedPayoutTxValidation.AmountMismatchException e) {
+            // The peer sent us an invalid donation address. We do not return here as we don't want to break
+            // mediation/arbitration and log only the issue. The dispute agent will run validation as well and will get
+            // a popup displayed to react.
+            log.error("Donation address invalid. {}", e.toString());
```
sqrrm: This should be warning level log, error better used for unexpected code paths.

core/src/main/java/bisq/core/trade/DelayedPayoutTxValidation.java
```diff
@@ -83,26 +80,33 @@
...
-        if (allPastParamValues.contains(addressAsString)) {
-            return true;
-        }
+        // If Dao is deactivated we need to add the default address as getAllPastParamValues will not return us any.
+        allPastParamValues.add(Param.RECIPIENT_BTC_ADDRESS.getDefaultValue());
+
+        // If Dao is deactivated we need to add the past addresses used as well.
+        // This list need to be updated once a new address gets defined.
+        allPastParamValues.add("3EtUWqsGThPtjwUczw27YCo6EWvQdaPUyp"); // burning man 2019
+        allPastParamValues.add("3A8Zc1XioE2HRzYfbb5P8iemCS72M6vRJV"); // burningman2
```
sqrrm: I'm thinking we should phase out the older addresses and reject them if used for new trades.

desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeView.java
```diff
@@ -222,6 +263,12 @@ public void initialize() {
             showFullReport();
         });
 
+        sigCheckButton = new AutoTooltipButton(Res.get("support.sigCheck.button"));
+        HBox.setHgrow(sigCheckButton, Priority.NEVER);
+        sigCheckButton.setOnAction(e -> {
```
sqrrm, line 268: Wouldn't it be better to automatically check the signature and add the result to the dispute graphics somewhere? Also, the refundagent gets this message when a user opens a refund case, this message doesn't verify since there is a header from the user included
```
System message: Mediator's dispute summary:
Mediator's node address: localhost:3600
```
Without that header the it would verify. Perhaps we should also make that message autoverify by adding a `----HEADER----` and `----END HEADER----` separator?

Another, probably separate issue, the popup from double clicking a chat message cannot be closed nicely (this might be an old thing though), never tried double clicking the messages in the chat.

sqrrm (changes requested):
Overall this is a good change, but since the verification is added I think it should be made automatic to be actually useful and not just add confusion.

desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/PendingTradesDataModel.java
chimp1984: Agree.

chimp1984:
> Overall this is a good change, but since the verification is added I think it should be made automatic to be actually useful and not just add confusion.

The verification is not for the users themself but for DAO voter to verify a reimbursement request. So they have an easy way to verify. The trader need to copy paste the summary msg and voters verify it. They dont have the data so they could not do it automatically.

core/src/main/java/bisq/core/trade/DelayedPayoutTxValidation.java
chimp1984: There are still too many trades with those. I don't know exactly why that happens but could partially reproduce it with quick test with some out of sync DAO state. I think letting them fail is a bit harsh.

sqrrm:
Ah, that makes more sense. Could still add auto check for refund agent but that's less important I suppose.

desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeView.java
chimp1984: The signature is not intended as check for the traders/arbitrator as that is anyway part of the p2p messaging (each msg gets checked if it is signed correctly by the sender). It is only intended for DAO voters.

chimp1984: "Another, probably separate issue, the popup from double clicking a chat message cannot be closed nicely (this might be an old thing though), never tried double clicking the messages in the chat." -> can you make an issue for that. I want not to extend that PR further, already too much ;-)

chimp1984:
> Ah, that makes more sense. Could still add auto check for refund agent but that's less important I suppose.

Yes sould be probably added. I will make a new PR for that once that is merged.

core/src/main/java/bisq/core/trade/DelayedPayoutTxValidation.java
sqrrm: I think it's fair. An easy fix is to download a new version and disable the dao.

We don't have a good way to phase out those old addresses, but they are essentially phased out on the DAO side, but traders still use them. It's almost a year later so I think phasing it out is not that bad.

desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeView.java
sqrrm: Indeed, will look at that separately.

sqrrm:
Only thing left to fix is the log level, then merge.

sqrrm:
@chimp1984 looks like there is a conflict now.

chimp1984:
Somehow travis does not like to work anymore ;-(. Tried to trigger it by another commit but stuck in "Expected — Waiting for status to be reported "
@ripcurlx Any idea?

sqrrm (approved):
ACK

PR #4544, opened 2020-09-21
Remove unnecessary listeners in TransactionsView

oscarguindzberg:
WalletChangeEventListener is enough to listen for events that affect the list of btc txs.


chimp1984 (dismissed):
utACK

chimp1984 (approved):
utACK

ripcurlx (approved):
utACK

PR #4548, opened 2020-09-22
Remove unused, broken accessor

ghubstan:
The accessor added in commit e2bd89f broke the build, and was not used as originally intended -- by CoreWalletsService -- because the the api's wallet methods need addresses from the wallet, not just a valid address.


ghubstan:
Import statements were also ordered in this change.

ripcurlx (approved):
utACK

PR #4555, opened 2020-09-23
Display the calculated security deposit at order entry

### LINKED ISSUE #4426: Recommended security deposit display
### Description

Not exactly a bug, but UX improvement. 
I don't know how to see the security deposit suggestion, my security deposit is always the same. Removing the security deposit and going back and forth when creating an order doesn't make the suggested security deposit appear.

#### Version

2.3.7

### Steps to reproduce

Make an offer when you already used your own security deposit for another offer in this market and it will use this security deposit, without even being able to know what the suggestion is at this time.

### Expected behaviour

I suggest to display suggested security deposit when clicking the "i" info next to "set both traders' security deposit" field. Any other way to be able to consult suggested security deposit easily is welcome.

### Actual behaviour

I'm not able to find the suggested security deposit. If there's a way and there's no need to make further changes, I'll document how on the wiki.

### Screenshots

issue recommended deposit

### Additional info

Related PR: https://github.com/bisq-network/bisq/pull/4347

ghost:
Always display at order entry screen the calculated security deposit and leave the decision to make it higher or lower to users for every offer.  Do not persist the value to user settings, since Bisq now calculates the suggested value each time.  (Prior to this a user setting `minimum security deposit` value was being displayed)

Fixes #4426

Description of why calculated security deposit did not always display before: https://github.com/bisq-network/bisq/issues/4426#issuecomment-696496134

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/offer/MutableOfferDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/MutableOfferDataModel.java
@@ -174,7 +174,7 @@ public abstract class MutableOfferDataModel extends OfferDataModel implements Bs
         addressEntry = btcWalletService.getOrCreateAddressEntry(offerId, AddressEntry.Context.OFFER_FUNDING);
 
         useMarketBasedPrice.set(preferences.isUsePercentageBasedPrice());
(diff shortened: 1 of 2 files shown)


chimp1984 (approved):
utACK

sqrrm (approved):
utACK

PR #4554, opened 2020-09-23
Change 'createoffer' api method return type

ghubstan:
This updates the existing gRPC `createoffer` method to return the new Offer instead of a boolean.  Although `createoffer` still places the offer, it might be changed to give the user a chance to examine a newly created (and cached) offer before it is placed in a separate `confirmoffer offer-id`  or  `placeoffer offer-id` method.  

An api test case was added and the CLI's `createoffer` method was stubbed out -- to be implemented in another PR.

Several unrelated, minor changes were pushed before the most relevant ones.

(This PR should be reviewed/merged before [4558](https://github.com/bisq-network/bisq/pull/4558) and [4559](https://github.com/bisq-network/bisq/pull/4559).)

PR #4558, opened 2020-09-23
Use existing timer in unlockwallet(pwd, timeout)

ghubstan:
This is one of the changes @chimp1984 requested in https://github.com/chimp1984/bisq/commit/961703ecea62df62946ab001ec28205662688ccd.

(This PR depends on [4554.)](https://github.com/bisq-network/bisq/pull/4554)

PR #4559, opened 2020-09-23
Build out 'createoffer' API method

ghubstan:
A 1st cut of the API CLI's `createoffer` method implementation, pushed too early to get feedback early.

(This PR depends on [4554.)](https://github.com/bisq-network/bisq/pull/4554)

To see it work, start supporting apps in regtest/dao mode:

    $./bisq-apitest --apiPassword=xyz --supportingApps=bitcoind,seednode,arbdaemon,alicedaemon  --shutdownAfterTests=false

Tail the alicedaemon log:

    $tail -n 2000 -f apitest/build/resources/main/bisq-BTC_REGTEST_Alice_dao/bisq.log

Use the CLI to see the default payment accts (and get the payment-acct-id for `createoffer`):

    $./bisq-cli --password=xyz getpaymentaccts

    Name                Currency  Payment Method  ID                                                    
    PerfectMoney dummy       EUR  PERFECT_MONEY   c1ca87a6-918f-4a57-960a-979596dee7f6                
    ETH dummy                ETH  BLOCK_CHAINS    d55040c4-aff4-4ed4-a900-73df578f9d2a 

To see `createoffer` argument list:

    $./bisq-cli --help

Create a BUY / USD offer:

    $./bisq-cli --password=xyz createoffer c1ca87a6-918f-4a57-960a-979596dee7f6 buy usd 0.0625 0.05 true 0.00 0.15

    Buy/Sell  Price in usd for 1 BTC  BTC(min - max)            usd(min - max)  Payment Method  Creation Date (UTC)   ID
    BUY                  10,264.6136  0.05000000 - 0.06250000        513 - 642  Perfect Money   2020-09-23T23:21:29Z  KnuOzBKi-4b9cd6b7-7c2c-493e-8744-3083e0fafb28-139

View offers:

    $./bisq-cli --password=xyz getoffers buy usd


ghubstan:
The `createoffer` and `createpaymentacct` methods have complex argument lists, and different devs may want them defined in in different ways.   We should discuss?

Right now, `createoffer` also places the offer, and problems might be created by placing it before giving users a chance to review details first. I thought the default behavior should be to cache the new offer (with ID) in the daemon, show it to the user, and let them place the new offer with a different `confirmoffer offer-id` command.    But I need some consensus on that too.

ghubstan (commented):
@chimp1984 

RE:  [Why are there are two 'Offer createOffer()' createOffer methods?](https://github.com/bisq-network/bisq/pull/4558#discussion_r495619071)

(1) is for creating a new buy-or-sell order with a random offerId
(2) is for creating a new buy-or-sell order with a given offerId
	
I think 2nd could be altered and/or renamed to use as a new 'placeoffer' method, where a cached Offer (in the single user daemon instance) created in the 1st is passed to the 2nd, after the user enters a 'placeoffer offer-id' command.

chimp1984:
Ah that was for edit offer I guess. We remove and create a new one same offer ID... I guess the domain would need some love to make it more clear...

cli/src/main/java/bisq/cli/NegativeNumberOptions.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private final Map<Integer, String> negativeNumberParams = new HashMap<>();
+
+    String[] removeNegativeNumberOptions(String[] args) {
+        // Cache any negative number params that will be rejected by the parser.
+        // This should be called before command line parsing.
+        for (int i = 0; i < args.length; i++) {
+            if (isNegativeNumber.test(args[i])) {
+                String param = args[i];
+                negativeNumberParams.put(i - 1, new BigDecimal(param).toString());
```
sqrrm, line 42: Why put at `i-1`? Wouldn't it be more natural to use `i`?

cli/src/main/java/bisq/cli/NegativeNumberOptions.java
```diff
@@ -0,0 +1,80 @@
...
+
+    List<String> restoreNegativeNumberOptions(List<String> nonOptionArgs) {
+        // Put cached negative number params into a clone of the nonOptionArgs list.
+        // This should be called after command line parsing.
+        if (!negativeNumberParams.isEmpty()) {
+            List<String> nonOptionArgsClone = new ArrayList<>(nonOptionArgs);
+            negativeNumberParams.forEach((k, v) -> {
+                int idx = k;
+                nonOptionArgsClone.remove(idx);
+                nonOptionArgsClone.add(idx, v);
```
sqrrm: How about `nonOptionArgsClone.set(idx, v);`?

cli/src/main/java/bisq/cli/CliMain.java
```diff
@@ -169,8 +189,39 @@ public static void run(String[] args) {
                     return;
                 }
                 case createoffer: {
-                    // TODO
-                    out.println("offer created");
+                    if (nonOptionArgs.size() < 9)
+                        throw new IllegalArgumentException("incorrect parameter count,"
+                                + " expecting buy | sell, payment acct id, currency code, amount, min amount,"
+                                + " use-market-based-price, fixed-price | mkt-price-margin, security-deposit");
```
sqrrm: Expected more than 8 options but only 8 expected options listed

sqrrm (commented):
> The `createoffer` and `createpaymentacct` methods have complex argument lists, and different devs may want them defined in in different ways. We should discuss?
> 
> Right now, `createoffer` also places the offer, and problems might be created by placing it before giving users a chance to review details first. I thought the default behavior should be to cache the new offer (with ID) in the daemon, show it to the user, and let them place the new offer with a different `confirmoffer offer-id` command. But I need some consensus on that too.

For an API I'm not sure there is a use case for the `confirm offer` extra step. Automated code wouldn't gain anything and UIs should probably implement that locally.

core/src/main/java/bisq/core/api/CoreDisputeAgentsService.java
```diff
@@ -100,6 +100,10 @@ void registerDisputeAgent(String disputeAgentTypeString, String registrationKey)
                     signature = refundAgentManager.signStorageSignaturePubKey(Objects.requireNonNull(ecKey));
                     registerRefundAgent(nodeAddress, languageCodes, ecKey, signature);
                     return;
+                default:
+                    // Codacy requires default label in switch, Intellij does not.
+                    //noinspection UnnecessaryReturnStatement
+                    return;
```
sqrrm: I think we can ignore codacy on this one going forward as it's possible to merge even when it's complaining. Perhaps we could even configure codacy to ignore it.

common/src/main/java/bisq/common/taskrunner/Task.java
```diff
@@ -65,7 +67,7 @@ protected void failed(String message) {
 
     protected void failed(Throwable t) {
         log.error(errorMessage, t);
-        taskHandler.handleErrorMessage(errorMessage);
+        taskHandler.handleErrorMessage(format("%s: %s", errorMessage, t.getMessage()));
```
sqrrm: The UI might get a bit verbose on errors with this change.

cli/src/main/java/bisq/cli/CliMain.java
sqrrm: I think this was fixed in a later commit

sqrrm (commented):
Seems overall ok, just the UI error popup that could get a bit verbose.

cli/src/main/java/bisq/cli/NegativeNumberOptions.java
ghubstan: I use `negativeNumberParams.put(i - 1, ...)` because `i-1` is the index of the api method param, but `args` includes the method name too.
   ```
    args[0] = method name
    negativeNumberParams.get(0) = 1st parameter of method name ( i.e., args[1] )
```
I added an explanation in commit e09b821

File: cli/src/main/java/bisq/cli/NegativeNumberOptions.java (hunk 2 above)
Comment (ghubstan) [reply cli/src/main/java/bisq/cli/NegativeNumberOptions.java]: Ugh.  Fixed in commit  23a677d

File: common/src/main/java/bisq/common/taskrunner/Task.java (hunk 5 above)
Comment (ghubstan) [reply common/src/main/java/bisq/common/taskrunner/Task.java]: I reverted it, then pushed a change (631c3f4) -> `taskHandler.handleErrorMessage(t.getMessage());`, which is what I meant to do in the first place.  Now, the verbose msg in logged on the server, but just the exception msg is passed up to clients.

File: core/src/main/java/bisq/core/api/CoreDisputeAgentsService.java (hunk 4 above)
Comment (ghubstan) [reply core/src/main/java/bisq/core/api/CoreDisputeAgentsService.java]: Unused default switch label deleted in 0332711.

**Comment by ghubstan** [comment]
> > The `createoffer` and `createpaymentacct` methods have complex argument lists, and different devs may want them defined in in different ways. We should discuss?
> > Right now, `createoffer` also places the offer, and problems might be created by placing it before giving users a chance to review details first. I thought the default behavior should be to cache the new offer (with ID) in the daemon, show it to the user, and let them place the new offer with a different `confirmoffer offer-id` command. But I need some consensus on that too.
> 
> For an API I'm not sure there is a use case for the `confirm offer` extra step. Automated code wouldn't gain anything and UIs should probably implement that locally.

I was sure `placeoffer` was need, but if you and @chimp1984  think not, OK.   I closed the `placeoffer` PR https://github.com/bisq-network/bisq/pull/4574.

**Comment by ghubstan** [comment]
Closing and re-opening to force CI build, which failed to download jfx dependency in last build.

File: core/src/main/java/bisq/core/api/CoreDisputeAgentsService.java [hunk 6]
```diff
@@ -99,6 +99,7 @@ void registerDisputeAgent(String disputeAgentTypeString, String registrationKey)
                     ecKey = refundAgentManager.getRegistrationKey(registrationKey);
                     signature = refundAgentManager.signStorageSignaturePubKey(Objects.requireNonNull(ecKey));
                     registerRefundAgent(nodeAddress, languageCodes, ecKey, signature);
+                    break;
```
sqrrm: For symmetry I would probably have a return here, or break in the MEDIATOR case. Not serious, but it makes more sense when similar cases are handled the same way. In this case one has to wonder why there is a return in one case and break in the next.

cli/src/main/java/bisq/cli/NegativeNumberOptions.java
sqrrm: Wouldn't it be more natural to start the for loop at 1 then? Like `for (int i = 1; i < args.length; i++) {` since it's known that argument 0 is irrelevant and it would also make the code self explanatory.

core/src/main/java/bisq/core/api/CoreDisputeAgentsService.java
ghubstan: I did not make sense to me, but codacy wanted a break (statement).   I'll change it to a return.

ghubstan: My two ducat's worth... I think Intellij's code inspector is more intelligent than codacy's.

cli/src/main/java/bisq/cli/NegativeNumberOptions.java
ghubstan: Yes -- commit  723fc8f.

common/src/main/java/bisq/common/taskrunner/Task.java
```diff
@@ -65,7 +65,7 @@ protected void failed(String message) {
 
     protected void failed(Throwable t) {
         log.error(errorMessage, t);
-        taskHandler.handleErrorMessage(errorMessage);
+        taskHandler.handleErrorMessage(t.getMessage());
```
sqrrm: I would like to exclude this change from this PR. It would affect the UI notifications that depend on the errormessage being reasonable for display.

Generally refactoring is fine (although more care is needed now with the many big PRs coming in, to avoid conflicts) and behavioral changes are fine as long as the ramifications are taken into account, I feel that's lacking here.

That's the only issue I have left. The rest looks good.

ghubstan: I put `Task.java` back to the way it was before I made any changes to it. It now sends `An error occurred at task: ValidateOffer` up to the CLI, instead of the details of the error, i.e., `Amount is larger than 1.00 BTC`, but solving this is for another PR, with better analysis of how a solution affects the UI.

sqrrm (approved):
utACK

PR #4560, opened 2020-09-24
XMR autoconf should get amount from trade not offer

### LINKED ISSUE #4557: Monero auto-confirm fails with AMOUNT_NOT_MATCHING or NO_MATCH_FOUND in certain cases
### Description

The new Monero trade auto-confirm feature may fail if the XMR transaction is not yet confirmed at the time the BTC buyer presses "payment started" and submits the XMR proof of transaction data to BTC seller

#### Version

v1.3.9

### Steps to reproduce

1) Begin an XMR/BTC trade
2) After the buyer sends the XMR payment, submit the XMR proof of transaction before the XMR payment is confirmed by the Monero blockchain

### Expected behaviour

Bisq should retry after a while if the XMR transaction is not yet confirmed

### Actual behaviour

Bisq instantly fails the auto-confirm functionality, despite the XMR transaction getting confirmed shortly after

#### Additional info

```
Result{
     detail=Detail{
     numConfirmations=0,
     errorMsg='null'
} NO_MATCH_FOUND
} FAILED
```

```
This might not mean that the XMR transfer was invalid but you have to check yourself if the XMR transfer was correct. Result{
     detail=Detail{
     numConfirmations=0,
     errorMsg='null'
} AMOUNT_NOT_MATCHING
} FAILED
```

ghost:
* Get the amount from the trade rather than the offer.
* Log the expected amount when not matching the explorer result.  It would have made debugging this issue much more straightforward.

Fixes #4557

### DIFF
--- a/core/src/main/java/bisq/core/trade/txproof/xmr/XmrTxProofModel.java
+++ b/core/src/main/java/bisq/core/trade/txproof/xmr/XmrTxProofModel.java
@@ -60,8 +60,7 @@ public class XmrTxProofModel implements AssetTxProofModel {
         this.serviceAddress = serviceAddress;
         this.autoConfirmSettings = autoConfirmSettings;
 
-        Coin tradeAmount = trade.getTradeAmount();
-        Volume volume = checkNotNull(trade.getOffer()).getVolumeByAmount(tradeAmount);
+        Volume volume = trade.getTradeVolume();
         amount = DevEnv.isDevMode() ?
                 XmrTxProofModel.DEV_AMOUNT : // For dev testing we need to add the matching address to the dev tx key and dev view key
                 volume != null ? volume.getValue() * 10000L : 0L; // XMR satoshis have 12 decimal places vs. bitcoin's 8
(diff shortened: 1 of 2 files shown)


core/src/main/java/bisq/core/trade/txproof/xmr/XmrTxProofModel.java
```diff
@@ -60,8 +60,7 @@
         this.serviceAddress = serviceAddress;
         this.autoConfirmSettings = autoConfirmSettings;
 
-        Coin tradeAmount = trade.getTradeAmount();
-        Volume volume = checkNotNull(trade.getOffer()).getVolumeByAmount(tradeAmount);
+        Volume volume = trade.getTradeVolume();
```
chimp1984, line 63: Good catch. I think the trade price is the reason for the issue.

chimp1984 (approved):
utACK

Could you reproduce the issue that it was really caused by that?

ghost:
Yes I reproduced the problem by forcing the taker's price to differ from maker's price same way as it was reported in the logs.  

I did not catch this problem during testing because all were self trades / same machine and so there was no price discrepancy between the trading parties.  Also I tended to use a fixed XMR rate which in hindsight was a bit silly as most traders use floating rates.

sqrrm (approved):

PR #4493, opened 2020-09-07
Deactive confirm buttons once mediation started

chimp1984:
Fixes #4162

PR #4562, opened 2020-09-25
Revert "Deactive confirm buttons once mediation started"

wiz:
Reverts bisq-network/bisq#4493

This change should have been discussed more before merging. Obviously disabling a UI button does not prevent a malicious person from abusing the trade protocol, and it only serves to degrade the UX of Bisq for honest traders who open a support ticket.

In the past week, about 50% of my mediation cases have traders who are complaining they can no longer resolve trades on their own because the mediation case is ongoing. As a mediator, it is my goal to mediate the trade so that the traders can complete the trade on their own, but now traders can no longer do so :sweat:

### DIFF
--- a/core/src/main/java/bisq/core/trade/protocol/BuyerAsMakerProtocol.java
+++ b/core/src/main/java/bisq/core/trade/protocol/BuyerAsMakerProtocol.java
@@ -55,8 +55,6 @@ import bisq.common.handlers.ResultHandler;
 
 import lombok.extern.slf4j.Slf4j;
 
-import static com.google.common.base.Preconditions.checkArgument;
-
 @Slf4j
 public class BuyerAsMakerProtocol extends TradeProtocol implements BuyerProtocol, MakerProtocol {
     private final BuyerAsMakerTrade buyerAsMakerTrade;
@@ -214,9 +212,6 @@ public class BuyerAsMakerProtocol extends TradeProtocol implements BuyerProtocol
     // User clicked the "bank transfer started" button
     @Override
     public void onFiatPaymentStarted(ResultHandler resultHandler, ErrorMessageHandler errorMessageHandler) {
-        checkArgument(!wasDisputed(), "A call to onFiatPaymentStarted is not permitted once a " +
-                "dispute has been opened.");
-
         if (trade.isDepositConfirmed() && !trade.isFiatSent()) {
             buyerAsMakerTrade.setState(Trade.State.BUYER_CONFIRMED_IN_UI_FIAT_PAYMENT_INITIATED);
             TradeTaskRunner taskRunner = new TradeTaskRunner(buyerAsMakerTrade,
--- a/core/src/main/java/bisq/core/trade/protocol/BuyerAsTakerProtocol.java
+++ b/core/src/main/java/bisq/core/trade/protocol/BuyerAsTakerProtocol.java
@@ -59,7 +59,6 @@ import bisq.common.handlers.ResultHandler;
 
 import lombok.extern.slf4j.Slf4j;
 
-import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 @Slf4j
@@ -238,9 +237,6 @@ public class BuyerAsTakerProtocol extends TradeProtocol implements BuyerProtocol
     // User clicked the "bank transfer started" button
     @Override
     public void onFiatPaymentStarted(ResultHandler resultHandler, ErrorMessageHandler errorMessageHandler) {
-        checkArgument(!wasDisputed(), "A call to onFiatPaymentStarted is not permitted once a " +
-                "dispute has been opened.");
-
         if (!trade.isFiatSent()) {
             buyerAsTakerTrade.setState(Trade.State.BUYER_CONFIRMED_IN_UI_FIAT_PAYMENT_INITIATED);
 
--- a/core/src/main/java/bisq/core/trade/protocol/SellerAsMakerProtocol.java
+++ b/core/src/main/java/bisq/core/trade/protocol/SellerAsMakerProtocol.java
@@ -57,8 +57,6 @@ import bisq.common.handlers.ResultHandler;
 
 import lombok.extern.slf4j.Slf4j;
 
-import static com.google.common.base.Preconditions.checkArgument;
-
 @Slf4j
 public class SellerAsMakerProtocol extends TradeProtocol implements SellerProtocol, MakerProtocol {
     private final SellerAsMakerTrade sellerAsMakerTrade;
@@ -206,9 +204,6 @@ public class SellerAsMakerProtocol extends TradeProtocol implements SellerProtoc
     // User clicked the "bank transfer received" button, so we release the funds for payout
     @Override
     public void onFiatPaymentReceived(ResultHandler resultHandler, ErrorMessageHandler errorMessageHandler) {
-        checkArgument(!wasDisputed(), "A call to onFiatPaymentReceived is not permitted once a " +
-                "dispute has been opened.");
-
         if (trade.getPayoutTx() == null) {
             sellerAsMakerTrade.setState(Trade.State.SELLER_CONFIRMED_IN_UI_FIAT_PAYMENT_RECEIPT);
             TradeTaskRunner taskRunner = new TradeTaskRunner(sellerAsMakerTrade,
--- a/core/src/main/java/bisq/core/trade/protocol/SellerAsTakerProtocol.java
+++ b/core/src/main/java/bisq/core/trade/protocol/SellerAsTakerProtocol.java
@@ -56,7 +56,6 @@ import bisq.common.handlers.ResultHandler;
 
 import lombok.extern.slf4j.Slf4j;
 
-import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 @Slf4j
@@ -197,9 +196,6 @@ public class SellerAsTakerProtocol extends TradeProtocol implements SellerProtoc
     // User clicked the "bank transfer received" button, so we release the funds for payout
     @Override
     public void onFiatPaymentReceived(ResultHandler resultHandler, ErrorMessageHandler errorMessageHandler) {
-        checkArgument(!wasDisputed(), "A call to onFiatPaymentReceived is not permitted once a " +
-                "dispute has been opened.");
-
         if (trade.getPayoutTx() == null) {
             sellerAsTakerTrade.setState(Trade.State.SELLER_CONFIRMED_IN_UI_FIAT_PAYMENT_RECEIPT);
             TradeTaskRunner taskRunner = new TradeTaskRunner(sellerAsTakerTrade,
--- a/core/src/main/java/bisq/core/trade/protocol/TradeProtocol.java
+++ b/core/src/main/java/bisq/core/trade/protocol/TradeProtocol.java
@@ -344,10 +344,6 @@ public abstract class TradeProtocol {
         cleanup();
     }
 
-    protected boolean wasDisputed() {
-        return trade.getDisputeState() != Trade.DisputeState.NO_DISPUTE;
-    }
-
     private void sendAckMessage(@Nullable TradeMessage tradeMessage, boolean result, @Nullable String errorMessage) {
         // We complete at initial protocol setup with the setup listener tasks.
         // Other cases are if we start from an UI event the task runner (payment started, confirmed).
--- a/desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/PendingTradesDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/PendingTradesDataModel.java
@@ -185,7 +185,7 @@ public class PendingTradesDataModel extends ActivatableDataModel {
     }
 
     public void onPaymentStarted(ResultHandler resultHandler, ErrorMessageHandler errorMessageHandler) {
-        Trade trade = getTrade();
+        final Trade trade = getTrade();
         checkNotNull(trade, "trade must not be null");
         checkArgument(trade instanceof BuyerTrade, "Check failed: trade instanceof BuyerTrade");
         ((BuyerTrade) trade).onFiatPaymentStarted(resultHandler, errorMessageHandler);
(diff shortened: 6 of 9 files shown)


chimp1984:
I think we should not full revert the old PR but improve it so it still fullfills the intention to avoid sellers who wait 19 days before releasing the BTC but do not disrupt honest traders.

But I do not fully understand the statement that you want the traders to let complete themselfes. I consider mediation as a conflict resolution, not a pure help service. So if there is a problem you should figure out who deserved the payout. If buyer has not sent yet due some problems, you can wait until he has and then once seller confirms make the payout to seller.
They then can accept both the result and it is the same like if they would have completed by themself. 

What are the probmeatic cases which makes it worse now?
I think it could be that the seller had issues and then confirms receipt in mediation. So you close that he gets the deposit back but the buyer might have become not responsive. Before the seller still could confirm and the trade completed. Now it goes to arbitration, which is bad of course. So I understand thats an issue we need to solve.

For the buyer the deactivation was never of importance, I just did it to make it same for both to not need to explain the otherwise more complex reasoning. So we could leave it to the buyer to click the button even if mediation is opened.

For the seller we would like that honest sellers can confirm without need that the buyer accept the mediators payout suffesting. But we also want to prevent that the seller just waits 19 days and release the BTC just in the last moment to reduce from that extra time his chargeback risk. We could add a max. delay after the mediation case has been closed in which the seller can still confirm by button click. After that it will be disabled and mediation result (or arbitration) is enforced. 
If we dont do that we lose any way to punish a misbehaving seller as he can always ignore the mediated suggestion by confirming manually.
The delay for the tolerance time could be 2 days after the ticket is closed. While the ticket is open he can confirm manually any time as well. 

I think that is a reasonable approach to gain some extra protection but do not hurt UX or cause more arbitration cases.

@wiz What do you think? If you disagree please explain more in details the use cases where this causes problems.

chimp1984 (changes requested):
Concept NACK

wiz:
> I do not fully understand the statement that you want the traders to let complete themselfes. I consider mediation as a conflict resolution, not a pure help service.

But this is exactly my point: **Arbitration is a conflict resolution service; Mediation is just a help service.**

As a Mediator, I am merely a neutral third party who helps traders work out their issues so they can complete the trade between themselves. It is only after 10 or 20 days when my efforts to mediate are unsuccessful that Arbitration is required. Before this change, for about 80% of my mediation cases I never had to make a suggested payout, and could close with TRADE_ALREADY_SETTLED, because the mediation was either successful or unnecessary, and the traders were able to complete the trades themselves.

Just because a newbie user requests support does not mean a trade should require conflict resolution. Now that users can no longer complete trades without having a suggested payout from me, and since I cannot make a suggested payout without knowing all the information, this will incur significant delays for all trades that go into mediation, since most traders are slow to respond in chat, if they ever respond at all. It will also increase the number of Arbitration cases, which is against our stated goals.

But most of all, this UI tweak actually changes the way trading on Bisq works, and users are reporting it as a bug, which I agree with - there is no reason this button should be disabled. If your goal is to fix an actual security flaw in the trade protocol, please do it in trade protocol, and not in the GUI for security theater.

cd2357:
> for about 80% of my mediation cases I never had to make a suggested payout, and could close with TRADE_ALREADY_SETTLED, because the mediation was either successful or unnecessary, and the traders were able to complete the trades themselves.

Is that because these trades just went over the trade time limit, and the seller took longer to confirm the received payment?

If yes, then how about enforcing this trade period? As they say, rules that have no consequences are not rules.

This is how it can be enforced:
- as soon as trade begins, a countdown timer starts for the buyer: please make the payment within this time (A), else every 1d of delay will cost you 20% of your deposit in case of mediation
- once the payment is made, a countdown timer starts for the seller: please confirm the payment within this time (B), else every 1d of delay will cost you 20% of your deposit in case of mediation
- if trade enters mediation, but turns out the buyer lied (clicked "I sent payment" on time, but actually sent the payment much later), then buyer loses security deposit

It could be set up in such a way that the total trade period T is big enough to cover A + B + some buffer.

`T = A + B + buffer`

For example, in case of a SEPA trade of duration 6 days:
- A = 1 days (once the trade starts, buyer should send SEPA payment within 1 day)
- B = 3 days (once payment is marked as started, seller has 3 days to confirm = weekend, where banks are closed + 1 day)

Both parties are incentivized to stay within their timeframe. Both parties could technically be a bit late, but as long as the trade completes within its duration limit, then there is no mediation, so there is no penalty for that.

This means most SEPA trades would finish within at most 4-5 days. The trade limit of 6 days offers 1-2d of extra buffer for eventual delays. But key point here is, once the 6 days are passed, there is obviously a problem and a mediator has to be involved. In this case, it means one of the parties was (very) late, so mediation will involve a penalty for that party, but will be free of charge for the party who was on time.

Honest traders are not penalized, lazy or unattentive ones (who ignore their countdown timer) are, potential profiteurs who want to postpone payment as much as possible, are also penalized.

This should incentivize traders to make the payment + confirm it as soon as they can. As a consequence, that would avoid trades going over the trade period limit, which is the territory of potential profiteurs anyway (waiting for 19 days to make payment, etc).

Would this sort of incentive structure make sense?

chimp1984:
> But this is exactly my point: **Arbitration is a conflict resolution service; Mediation is just a help service.**
> 
> As a Mediator, I am merely a neutral third party who helps traders work out their issues so they can complete the trade between themselves. It is only after 10 or 20 days when my efforts to mediate are unsuccessful that Arbitration is required. Before this change, for about 80% of my mediation cases I never had to make a suggested payout, and could close with TRADE_ALREADY_SETTLED, because the mediation was either successful or unnecessary, and the traders were able to complete the trades themselves.

Thats great if they can settle themselfes, but of course mediation is also conflict resolution, otherwise there would not be the feature of the mediated payout suggestion. If one side breaks clearly the trade obligations the mediator has to apply some penalization, otherwise it's an invitation for traders to ignore the rules. I fear this aspect is not taken enough seriously by mediators as far I have seen some data (not in depth, but was my impression when responding to some reimbursement cases).
 
> But most of all, this UI tweak actually changes the way trading on Bisq works, and users are reporting it as a bug, which I agree with - there is no reason this button should be disabled. If your goal is to fix an actual security flaw in the trade protocol, please do it in trade protocol, and not in the GUI for security theater.

Lets not get into such a polemic discussion mode... I explained the details already why a trade protocol change would not be justified (would also come with backward compatibility challenges). 

As stated above I agree that we should improve it, but try to find an improvement which avoids the abuse from sellers, even if it is not a "security" feature but just makes it hard enough for 99.9% of traders to be able to keep abusing it, the goal for reducing cases where seller confirms in last moment is met.

chimp1984:
@cd2357 
Yes I agree to your view. It is basically designed like that. The buyer always have to send fast. In sepa case if a buyer send on day 5 and then it does not arrive in the trade period and mediation gets opened it should be clear to mediators that the buyer deserved some penalty. It is not written down clearly as that so probably we should define that more clearly. I think 1 day is reasonable for all payment methods. The 1 day duration payment methods could get half a day but less will problematic because of timezone differences...
We have the popups as half of the trade period with a warning that trade has to be completed in time. But that could be improved.

With stricter rules of course it increases the incentive to lie to the mediator. One cannot really proof if a "bank problem" was real or fake. So it still requires some flexibility and tolerance as we don't want to punish honest traders. The frustration of getting a penalty without doing anything wrong should not be under-estimated, so in doubt I would suggest to be tolerant.

To add some "soft" reputation for the speed of trading is a long time idea as well but as it comes with privacy issues and need to be implemented in a way that it cannot be gamed by self-trades it was never started. The negative score concept would avoid the "self-trade" problem thought have its own problems as well...

chimp1984:
We want to achieve 2 goals. 
1. We want to avoid that a fiat seller delays payout just before arbitration can be opened without any penalty.
2. We want to avoid that penalty is rendered irrelevant as the trader can continue the trade and ignoring the mediated payout suggestion.

For point 2:
For buyers it has less relevance as it requires that the seller agrees as well by confirming, so in conflict the seller has no incentive to do that (he would also gain from penalty as he receives that part) and if he agrees he basically "forgive" the buyer for some protocol violation. This might still be an issue as those protcol violations might not only hurt the other peer but could cause a more general problem to Bisq (e.g. attracting scammers if they get away easily), but I think that can be ignored. For the sellers side we have covered the fiat seller with point 1 and the altcoin seller does not gain anything from delaying the confirmation, in the opposite he is also interested to get his deposit unlocked, so I think we can also ignore that case.

So point 1 is the remaining problem we want to solve.
If we allow for fiat seller confirming while mediation is open he might try to delay the mediation result as long as possible. Same if we allow after mediation is closed we would need some tolerance time, say 1 day until he still can confirm but that would also render any penalty applied by mediators to have no consequences as well creates incentive to delay mediation result. So this does not solve our problem.

My suggesting is: 
- Enable for buyer and altcoin seller at all times (only once arbitration is opened it is disabled as there is no way anyway to do a normal payout anymore.
- Disable for fiat seller while mediation is open. Enable it at dispute result if no penalty was applied by mediators (derived from payout suggestion). So the mediator is the enabler for the confirm-button by his judgement expressed in the payout suggestion. This solves automatically point 2 as if the mediated payout suggestion is the same as the happy-path there is no avoid-penalty issue as no penalty was applied. 
If seller confirms receipt in mediation the mediator can close anyway so there is only a bit of extra friction and delay due the communication but the dispute has to be closed anyway from the mediator. So I don't see that this creates any significant difference. If mediator has applied a penalty to the buyer the seller still gets his button activated (no penaly for seller - this is the only condition we check). So the seller has 2 options: accept mediated result and hope that buyer accept as well, or click confirm button and "forgive" buyer. The confirm button would stay active also after accept button has been clicked. So he can try mediated suggestion and before it goes to arbitration click the confirm button so we avoid arbitration cases.

wiz:
Do you guys think Bisq is perfect and has no bugs? What if I told you the majority of mediation cases are users who need help working around bugs in Bisq and not looking for any conflict resolution service? The previous system worked very well, and the current system does not. I hope other Mediators will give their thoughts as well, but here are a few real-world examples of the cases I am dealing with, these are all real tickets from my inbox today:

1) Dispute was opened for a fiat trade, because the deposit TX failed to broadcast for some reason, and did not appear on the blockchain. However, the buyer says they already made payment for some reason anyway. So I asked the seller to restart their Bisq node and fortunately the deposit TX got broadcast upon restarting and confirmed. Now the parties can complete their trade normally, except oh wait they can't because the seller's UI button is disabled. As the Mediator, I don't want to make a suggested payout in favor of Buyer because he has not provided proof of payment and is no longer responding, but if he did make the payment on time then it's not his fault. This is a typical case where nobody did anything malicious, and parties should be able to complete the trade on their own after my mediation was successful, and no "conflict resolution" is necessary, because a "conflict" doesn't exist. It's simply one of many bugs in Bisq that caused a support ticket to be opened for.

2) Dispute is opened for a fiat trade, but neither party is responding or saying anything. This is also a typical case. Normally the parties would be able to complete the trade on their own, but now they cannot do so until I make a suggested payout, and since I don't have any information from either party, I can't make a suggested payout. Again, nobody being malicious, just people busy with their lives or something. A fiat trade for $100 isn't a top priority for most people, and one of them is probably a new user to Bisq... I fear this horrible UX will probably scare them away from using it.

3) Dispute is opened for a fiat trade, because Seller says never received payment from Buyer. The Buyer might have made payment, but I have to wait for him to respond and provide me with proof of payment, and until he does I don't have enough information to make a suggested payout right now. A few days later the funds show up after some bank delay, and the traders would normally be able to complete the trade on their own, but now they cannot because the UI button is disabled. Both parties must wait for either the Seller to acknowledge payment to me in chat or the Buyer to provide a proof of payment to me until I can make a suggested payout, whereas they could have just otherwise completed it normally. This is another example where nobody did anything malicious, and they were right to open a support ticket, but now they are punished for it, and basically inconvenienced for no reason.

Mediation is Mediation and a help service, please don't confuse it with enforcing part of the trade protocol. That's what Arbitration is for, which only begins 10 or 20 days after the trade start, if Mediation is unsuccessful. If you want to enforce the trade protocol, have the Arbitrator do it. I am just a higher ranking Support Agent and I don't have any power or desire to enforce the trade protocol.

RefundAgent:
I think the latest suggestion by chimp1984 should work. Then things will work more or less as now, but sellers cannot wait tactically for 19 days and then close the deal. 
I am still surprised that this is a problem, since I don't understand the incentives of waiting to the last minute to confirm receipt of payment. In particular - forced chargebacks cannot work in any economic system and cases of stolen bank accounts (where chargebacks can be ASKED) are virtually non-existent in the present Bisq-system. I would like to know how many cases of delayed confirmations there are.

chimp1984:
@wiz 
Case 1 and 3 can be resolved as it was earlier with my suggested solution. You do not apply any penalty on the suggested payout the the seller just either confirms or accept...
Case 2 is weird and rare I assume and we should try to make it more clear in the UI that users need to be online again and where to find the dispute screen. Seems some of the not responding cases is just from users not knowing much about Bisq (to be online to check state) or dont see the red button for support.... Maybe @pedromvpg or @ripcurlx have some ideas how we can improve that area.

sqrrm:
The mediation process is part of the rules enforcement process. In benign cases with traders that just need help, no enforcement is needed, just guidance and support. Cases where rules have been broken should result in a suggested penalty from mediators and if that's not accepted, nothing would be returned to the the aggrieving party in the refund process.

Reverting the previous PR doesn't seem right, making the pure support path smother sounds better.

wiz:
Yeah, thanks everyone, the purpose of this PR was to start a discussion and since that's been accomplished and will close this PR. I look forward to chimp's proposed improvements getting added to the next release.

chimp1984:
@wiz #4602 is open for review and testing...

MwithM:
Sorry for joining this discussion so late, but I have a doubt:

> We want to avoid that a fiat seller delays payout just before arbitration can be opened without any penalty.

Why is only fiat mentioned? Isn't it indifferent?

As buyer for an instant alt trade if seller clicks "payment received" 3 days late I'm being aggravated.
That's what should be prevented, and I'm not sure if the changes after this PR are looking for it.

MwithM:
Some sellers have been releasing the funds after mediation has been submitted because sellers are not replying or releasing funds. From #support keybase channel:

> "there is something really bad about the protocol of mediation.  
> 
> i had one trade that should last 24 hours. But the seller just dissapear never confirming my payment.  so i opened a mediation which take about 2-3 days more, so he had a lot of time to release until that time. mediation case was close with a deal proposed and a penalty for the almost 5 days delay of the trade. 
> 1 hour after the deal was offered by mediation, the seller release normally ignoring the deal and getaway without any penalty for all this delay. 
> 
> not fair at all."

It was knew it was possible to sign the normal payout even with the button deactivated, but I still don't know why they do it since they just keep their deposit locked for a longer time and gain nothing but pissing buyers.

PR #4568, opened 2020-09-29
Add segwit support to the BTC wallet

oscarguindzberg:
This is just the BTC wallet.
Trade protocol keeps using legacy addresses. 
This PR does not require a hardfork.

### DIFF
--- a/core/src/main/java/bisq/core/btc/setup/WalletConfig.java
+++ b/core/src/main/java/bisq/core/btc/setup/WalletConfig.java
@@ -341,8 +341,7 @@ public class WalletConfig extends AbstractIdleService {
 
     private Wallet loadWallet(boolean shouldReplayWallet, File walletFile, boolean isBsqWallet) throws Exception {
         Wallet wallet;
-        FileInputStream walletStream = new FileInputStream(walletFile);
-        try {
+        try (FileInputStream walletStream = new FileInputStream(walletFile)) {
             WalletExtension[] extArray = new WalletExtension[]{};
             Protos.Wallet proto = WalletProtobufSerializer.parseToProto(walletStream);
             final WalletProtobufSerializer serializer;
@@ -360,8 +359,6 @@ public class WalletConfig extends AbstractIdleService {
                         .accountPath(new BisqKeyChainGroupStructure(isBsqWallet).accountPathFor(Script.ScriptType.P2WPKH)).build();
                 wallet.addAndActivateHDChain(nativeSegwitKeyChain);
             }
-        } finally {
-            walletStream.close();
         }
         return wallet;
     }
--- a/core/src/main/java/bisq/core/btc/setup/WalletConfig.java
+++ b/core/src/main/java/bisq/core/btc/setup/WalletConfig.java
@@ -22,6 +22,7 @@ import bisq.core.btc.nodes.ProxySocketFactory;
 import bisq.core.btc.wallet.BisqRiskAnalysis;
 
 import bisq.common.config.Config;
+import bisq.common.file.FileUtil;
 
 import com.google.common.io.Closeables;
 import com.google.common.util.concurrent.*;
@@ -510,6 +511,14 @@ public class WalletConfig extends AbstractIdleService {
                 // wait for the aesKey to be set and this method to be invoked again.
                 return;
             }
+            // Do a backup of the wallet
+            File backup = new File(directory, WalletsSetup.PRE_SEGWIT_WALLET_BACKUP);
+            try {
+                FileUtil.copyFile(new File(directory, "bisq_BTC.wallet"), backup);
+            } catch (IOException e) {
+                log.error(e.toString(), e);
+            }
+
             // Btc wallet does not have a native segwit keychain, we should add one.
             DeterministicSeed seed = wallet.getKeyChainSeed();
             if (aesKey != null) {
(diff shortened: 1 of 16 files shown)


ripcurlx:
@oscarguindzberg If you run into Codacy issues please let me know upfront, so you don't need to create unnecessary workarounds. I just checked the current complaints and removed the pattern for comment intention as IntelliJ is doing it differently by default. The other remaining two are legit (formatting issue and unused import)
Bildschirmfoto 2020-09-29 um 09 46 10

chimp1984 (commented):
utACK

oscarguindzberg:
- Finished reviewing the source code to make sure this brings no bugs to the dao.
- Tested DAO txs (compensation requests, voting, etc) and trading. I tested both pre-segwit and post-segwit nodes. 
- Removed the WIP tag from this PR title.

This PR is ready to be merged IMHO.

oscarguindzberg:
@ripcurlx @sqrrm @chimp1984 Just bringing you attention in case you are in charge of approving/merging this PR.

core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
```diff
@@ -584,23 +584,13 @@ public AddressEntry getOrCreateAddressEntry(String offerId, AddressEntry.Context
             if (emptyAvailableAddressEntry.isPresent()) {
                 return addressEntryList.swapAvailableToAddressEntryWithOfferId(emptyAvailableAddressEntry.get(), context, offerId);
             } else {
-                AddressEntry entry = new AddressEntry(wallet.freshReceiveKey(), context, offerId);
+                AddressEntry entry = new AddressEntry(wallet.freshReceiveKey(), context, offerId, true);
```
chimp1984: Is that never called from a trade protocol entry? e.g. the one for the multisig.

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
```diff
@@ -601,6 +603,9 @@ public Transaction takerSignsDepositTx(boolean takerIsSeller,
             for (int i = 0; i < buyerInputs.size(); i++) {
                 TransactionInput transactionInput = makersDepositTx.getInputs().get(i);
                 depositTx.addInput(getTransactionInput(depositTx, getMakersScriptSigProgram(transactionInput), buyerInputs.get(i)));
+                if (!TransactionWitness.EMPTY.equals(transactionInput.getWitness())) {
+                    depositTx.getInput(depositTx.getInputs().size()-1).setWitness(transactionInput.getWitness());
```
chimp1984: Instead of `depositTx.getInputs().size()-1)` I would suggest to make a local var from the `getTransactionInput(depositTx, getMakersScriptSigProgram(transactionInput), buyerInputs.get(i))` we add earlier so its more clear. Maybe even do the check before adding, technically no difference but somehow more clean if the input gets completely defined before added.

chimp1984: As we do not support now segwit in the trade process, is that change needed at all? I dont see a risk to have it there but maybe better to leave all that out until we start adding segwit support to the trade protocol? Not sure if you started already in more pleaces to add support which do not interfere (as we discussed it is our strategy to add as much as possible without activating segwit in trade protocol...), so if you have started that process already all find, otherwise if its a single change it would be maybe more clear to start after the release with that process...

chimp1984 (changes requested):

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
oscarguindzberg: You seem to be right. In any case, it causes no harm.

core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
oscarguindzberg: It seems you commented on outdated code.

oscarguindzberg:
@ripcurlx Found a bug migrating encrypted wallets, added WIP tag. Will remove WIP once the bug is fixed.

core/src/main/java/bisq/core/btc/setup/WalletConfig.java
```diff
@@ -293,25 +301,34 @@ protected void startUp() throws Exception {
...
-                    vPeerGroup.startBlockChainDownload(tracker);
-                }
-
-                @Override
-                public void onFailure(Throwable t) {
-                    throw new RuntimeException(t);
+            if (migratedWalletToSegwit.get()) {
+                startPeerGroup();
+            } else {
+                migratedWalletToSegwit.addListener((observable, oldValue, newValue) -> startPeerGroup());
```
chimp1984: I think we should add a `if(newValue)` to make it more clear/safe. As we change only once from false to true it would not matter but feels more correct.

chimp1984 (commented):
utACK

See small suggestion in comment...

chimp1984 (approved):

core/src/main/java/bisq/core/btc/setup/WalletConfig.java
oscarguindzberg: Fixed https://github.com/bisq-network/bisq/pull/4568/commits/261e0ec714027be657ba7d1125ac65bd81b0135f

chimp1984 (commented):
utACK

sqrrm (approved):

wiz:
post merge ACK - tested OK on mainnet using encrypted wallet for send and receive, and also using mixed segwit/legacy inputs/outputs all worked fine

PR #4242, opened 2020-05-07
Upgrade JavaFX to v15

cd2357:
Update the gradle dependency to JavaFX 14.

This brings to Bisq the latest JavaFX fixes and improvements, especially
 in the areas of UI performance and memory management.

Initial tests on Qubes already showed several improvements:
* faster, more responsive UI
* popups display correctly (unlike the current setup, where 95% of the time modal dialogs had width and height 0; see #873)
* system tray icon is visible (before, it never showed for me on linux)

Tests on Mac also show better performance + better / more "native" UI integration.

To use the new JavaFX version, just `./gradlew clean build` Bisq before running it.

Note: since [JavaFX also uses VRAM](https://wiki.openjdk.java.net/pages/viewpage.action?pageId=20415996) (in addition to the normal system RAM
as used by the JDK), it could be that previous attempts to optimize the JDK's
memory usage (like #3677 and #4048) might have had no impact on JavaFX-
specific inefficiencies.

Related / likely related to: #350 #2135 #2509 #3128 #3307 #3308 #3343
#3430 #3657 #3683 #3686 #3786 #3787 #3892 #3917 #3918 #3936

PR #4572, opened 2020-09-29
Change Venezuelan currency from VEF to VES

### LINKED ISSUE #1920: Currency code for Venezuelan bolivar has changed
The app shows "VEF", but this year that code changed to "VES"

pingiun:
Fixes #1920

This adds a properties file to supply to java.util.Currency so that it knows the Venezuelan currency is VES. Adding additional data to this class is quite cumbersome. You need to supply a properties file, but it cannot be a class resource, it has to live in an actual file. I placed this logic in BisqExecutable.java but I don't think it's the best location. 

The currency currently has no display name, so the Bisq app defaults to "VES (VES)". I don't know how locale information within the Java jdk can be translated, so please help me with that. Alternatively I could do some checks in various places to check if the display name for VES is requested and we could use our own translation strings.
Update: I added translations from the bisq string resources

This code should be removed if we upgrade to Java 11 or newer, because the currency data is updated in those versions.


pingiun:
image
image

wiz:
Aren't we upgrading Java soon? If so this patch likely isn't needed.

pingiun:
I think @mwithm wanted to have this quite soon. If the Java version is upgraded in the next release, this should not be needed. Otherwise I think this is useful as a stop-gap

ripcurlx:
> I think @MwithM wanted to have this quite soon. If the Java version is upgraded in the next release, this should not be needed. Otherwise I think this is useful as a stop-gap

Hi @pingiun! I think it is not worth adding this patch as it looks like we'll update to JDK 11 already in 1.4.1 or 1.4.2.

MwithM:
@pingiun Thank you anyway for this doing this pull request so quickly.

ripcurlx:
Closing as it will be superseded by #4242.

PR #4573, opened 2020-09-29
Cleanup debug logs

chimp1984:
Remove toProto call just for debug logs. That was quite expensive at we called tha for each message sent.

Cleanup some logs and apply code inspection suggestion for common module.

Regarding logging:
- We should always use params,, then the method calls at the params are not executed if log level does not match, thus adding very tiny overhead to have a log.

isDebugEnabled() is not recommended if params are used. It caused more performance costs and adds boilerplate code.

See:
http://logging.apache.org/log4j/1.2/manual.html
"This will not incur the cost of parameter construction if debugging is disabled. On the other hand, if the logger is debug-enabled, it will incur twice the cost of evaluating whether the logger is enabled or not: once in debugEnabled and once in debug. This is an insignificant overhead because evaluating a logger takes about 1% of the time it takes to actually log."


chimp1984:
No idea why travis failed. My local build and tests are ok. Try again with another commit.

ripcurlx (approved):
utACK - log and other code changes are looking fine

PR #4576, opened 2020-09-30
Fix bug with missing remove of old filter when new filter is added.

chimp1984:
Fix bug with missing remove of old filter when new filter is added.

Add handling for removing invalid filters
When adding a new filter invalid filters get removed if they have been created with the same priv key.


ripcurlx (approved):
ACK

Code is looking fine and testing it on Mainnet where I experienced this issue solved my problem.

PR #4519, opened 2020-09-13
Support resource files tagged by version to reduce initial data requests

chimp1984:
This is based on #4233 from @freimair 

It applies the feature only to tradeStatistics to reduce risks in case there are issues.

PR #4577, opened 2020-09-30
[WIP] New trade stats

chimp1984:
Replace TradeStatistics2 with a new much smaller TradeStatistics3 object.

It converts existing TradeStatistics2 to the new data as well as data received from the network, which get rebroadcast as new data. 

The data store got reduced from 16.5 MB to 3.6 MB. If is also much better for privacy as it only stores the bare minimum of data and does not link to the offers.

It is still early WIP, missing tests and updates in other modules. Just wanted to share early as it should be deployed with the intitial data request reduction and the persistence redesign.  

Depends/based on https://github.com/bisq-network/bisq/pull/4519
Starts at commit 5e09eebe18ed3b1ba2ed934ac2d7f370506c725b earlier commits are from #4519


chimp1984:
Replaced by 4599

Issue #4569, opened 2020-09-29
Add info if peer has enabled XMR auto-confirmation


chimp1984:
An sell offer of a maker who has enabled the XMR auto-confirmation is more attractive to takers as it can be assumed that the payout will be much faster then in manual mode. We should make this information visible.

We could add to the extraMap an entry if the maker has the autoConf enabled. In the offer book we can then show a new icon to signal to takers that this offer will likely be confirmed fast. 
If the maker changes the setting after offer-creation we would need to update the offer (remove old once and add new one with updated extraMap). Still there might be a gap that the maker deactivates the auto-conf after a taker has taken the offer but before the XMR got confirmed, thus leading to a wrong expectation for the taker. I think that cannot be avoided and enforcing to use auto-conf once set in the offer might be too harsh/confusing.

If the maker is the buyer and would only like to accept takers who have auto-conf enabled we could add another entry in the extraMap to signal that. But it might be a bit too much UX hassle to explain this to makers at offer creation (e.g. "Do you want to only accept takers who have auto-conf enabled?").

Another approach might be to add a new payment method ALTCOIN_AUTO_CONF similar to SEPA_INSTANT. But I guess that would cause more effort and has the risk to partition the market. Also it will require more changes of the handling of the setting, as for normal XMR it would be taken from preferences, for the ALTCOIN_AUTO_CONF trades it would be enabled by default and cannot be changed.

So there are several paths to get there:

1. Add flag to offer and show it in offerbook
2. Update offer once maker changes autoconf
3. Add support that maker only accepts takers which have autoconf enabled (only if roles of taker is seller)

Or
Add new payment method type and enforce that any trade of that type require the auto-conf to be set.

I think to keep effort in limits we should start with option 1 and impl. it step by step and even roll it out if only step 1 is implemented.

Any comments?

chimp1984:
I implemented the first part: 

https://github.com/chimp1984/bisq/tree/show-if-maker-has-xmr-autoconf-enabled

Auto-conf offers get now a "rocket" ;-)
<img width="1037" alt="Screen Shot 2020-09-29 at 00 24 39" src="https://user-images.githubusercontent.com/54558767/94517298-4582a980-01ed-11eb-9b0f-dd652f86a177.png">


<img width="1126" alt="Screen Shot 2020-09-29 at 00 32 09" src="https://user-images.githubusercontent.com/54558767/94517292-43204f80-01ed-11eb-9c17-b7444f55241f.png">

[...]

PR #4578, opened 2020-09-30
Show if maker has xmr autoconf enabled

chimp1984:
Implements https://github.com/bisq-network/bisq/issues/4569

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -338,7 +338,7 @@ offerbook.offerersAcceptedBankSeats=Accepted seat of bank countries (taker):\n {
 offerbook.availableOffers=Available offers
 offerbook.filterByCurrency=Filter by currency
 offerbook.filterByPaymentMethod=Filter by payment method
-offerbook.timeSinceSigning=Time since signing
+offerbook.timeSinceSigning=Signed since
 offerbook.timeSinceSigning.info=This account was verified and {0}
 offerbook.timeSinceSigning.info.arbitrator=signed by an arbitrator and can sign peer accounts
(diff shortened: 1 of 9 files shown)


core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -338,7 +338,7 @@ offerbook.offerersAcceptedBankSeats=Accepted seat of bank countries (taker):\n {
 offerbook.availableOffers=Available offers
 offerbook.filterByCurrency=Filter by currency
 offerbook.filterByPaymentMethod=Filter by payment method
-offerbook.timeSinceSigning=Time since signing
+offerbook.timeSinceSigning=Signed since
```
ripcurlx, line 341: Did you change this to save space?

ripcurlx:
@chimp1984 How about using the fast clock instead of the rocket?
Bildschirmfoto 2020-09-30 um 17 58 33

chimp1984:
Yes ok as well, though i felt the rocket might add more incentive is looks a bit more "special"... but feel free to use the clock, no strong opinion here...

chimp1984:
Maybe the clock could be used for all instant trade options?

ghost:
Expressing my preference for the rocket!  I think it looks cool and may generate interest in the feature...

image

chimp1984:
Yes I think it sticks out a bit more than a more traditional icon. The "to the moon" crew will like it as well ;-)

ripcurlx:
> Yes I think it sticks out a bit more than a more traditional icon. The "to the moon" crew will like it as well ;-)

ok - voted down and I accept my defeat 😜

ripcurlx (approved):
ACK

As described in https://github.com/bisq-network/bisq/issues/4569 this sets for xmr trades the autoconf flag on offer creation. It won't be updated if the config is changed after the offer has been created.

PR #4583, opened 2020-10-01
Improve startup routines

chimp1984:
Refactorings and cleanups of Bisq startup code.

Removed the crypto policy checks as it is not needed anymore as far I am aware of (not 100% sure though).


sqrrm:
@chimp1984 Seems like some good refactoring. Would merge when bugs causing build to fail are fixed.

sqrrm:
@chimp1984 A conflict remains

sqrrm (commented):
utACK

sqrrm (approved):

PR #4574, opened 2020-09-29
Add 'placeoffer' method

ghubstan:
New offer placement is moved out of `createoffer` into a separate  method, to give users a chance to review new offer details before broadcasting it.  

When `createoffer` is called, the new offer is cached in `CoreOffersService` for 5 minutes, and details are displayed in the CLI console.  The user can broadcast the cached offer by calling `placeoffer offer-id security-deposit`. The security-deposit parameter is a % as double, e.g., 0.15.

PRs https://github.com/bisq-network/bisq/pull/4554 and https://github.com/bisq-network/bisq/pull/4559 need to be review/merged before this one.

PR #4584, opened 2020-10-01
Add api convenience 'getoffer offer-id' method

ghubstan:
An API convenience for getting a current offer with an offer id.  Also stubbed out a `takeoffer` proto def.

PR https://github.com/bisq-network/bisq/pull/4574 needs to be reviewed & merged before this one.


ghubstan:
Closing because this depends on `placeoffer`, which will not be needed (and closed too).  We re-do in another PR.

Issue #4575, opened 2020-09-29
Big mining fees


ghost:
Actually, the trading fees (mining fees especially are quite high). It makes trading of smaller amounts very problematic. (Like 4% on mining fee is quite a lot).

![image](https://user-images.githubusercontent.com/383295/94627262-4c78ed00-02bd-11eb-8652-d29a0898654b.png)

This must be pain for new users who has new account and low limits. And I am little bit afraid it may discourage them from using Bisq.

----

I understand that implementing SegWit will help. But also, is would be nice to have possibility to lower fees per byte (I don't mind longer waiting times).

ghost:
Here it is even 8%, it is extremely unreasonable to trade at such high transaction cost.

![image](https://user-images.githubusercontent.com/383295/94659390-815c6280-0304-11eb-9ddf-d2dbda9b42df.png)

cd2357:
@petrhejna unfortunately, there are other considerations at play that make this not a very straightforward choice.

The most reasonable workaround for this is to avoid trading small amounts when the [mempool](https://mempool.space) is full.

It is usually almost empty on weekends, so that's when you'll generally get the lowest fees.

ghost:
Thank @cd2357 for clearing it out. 

But sadly :( even when mempool is quite empty (for few hours now) the estimation is `BTC tx fee: txFeePerByte=102` which is enormous. 

![image](https://user-images.githubusercontent.com/383295/94989644-a952fc80-0576-11eb-8ee0-252671c2a2db.png)

ghost:
Another big problem is on-boarding of new users. This creates **big entry barrier**.

Even for me. I have a signed Revolut account but I also want to trade with wire transfers. 

So I am forced to take small trade (the seller charges 8% plus) and additional 4-7% of the trade in mining fees. And then wait 30 days.

wiz:
> But sadly :( even when mempool is quite empty (for few hours now) the estimation is BTC tx fee: txFeePerByte=102 which is enormous.

Yes, we've just increased yesterday to always have Bisq use "high" priority fee estimate. The problem is that it's impossible to do fee estimation accurately because of mining variance and random people suddenly broadcasting huge amounts of transactions, so if you use "medium" priority fee your TX can easily get buried and trade can get delayed for hours or days. This is simply the consequence of Bisq trades being on-chain Bitcoin transactions and the only solution is to move to a Layer 2 Network like Liquid or Lightning.

softsimon:
[...]

PR #4585, opened 2020-10-01
Increase our mempool fee estimation from "medium" to "high" priority

wiz:
Screen Shot 2020-10-02 at 5 15 00

Bisq currently uses the mempool.space API for fee estimation, selecting
the "medium" priority fee, which generally works very well. However, if
your timing is bad, sometimes the mempool can quickly fill up and your
TX can get stuck for several hours or even days, which degrades the UX
of Bisq.

This PR changes the Pricenode code to always use "high" priority fees,
which would at the time of writing use 99 sat/vB instead of 85 sat/vB,
and hopefully prevent the above-mentioned issue from occuring. Of
course if the mempool is empty the minimum fees would still be used, as
this only changes the "priority" from medium to high.

Fee estimates: https://mempool.space/
API endpoint: https://mempool.space/api/v1/fees/recommended

### DIFF
--- a/pricenode/src/main/java/bisq/price/mining/providers/MempoolFeeRateProvider.java
+++ b/pricenode/src/main/java/bisq/price/mining/providers/MempoolFeeRateProvider.java
@@ -91,7 +91,7 @@ abstract class MempoolFeeRateProvider extends FeeRateProvider {
 
     private long getEstimatedFeeRate() {
         return getFeeRatePredictions()
-            .filter(p -> p.getKey().equalsIgnoreCase("halfHourFee"))
+            .filter(p -> p.getKey().equalsIgnoreCase("fastestFee"))
             .map(Map.Entry::getValue)
             .findFirst()
             .map(r -> {


ghubstan:
utACK

Bayernatoor:
utACK .. good call!

cbeams:
utACK, seems reasonable, though I'm not sure how much of a problem this has been in practice. @m52go and I ran into this the other day while were recording a demo (had to wait many hours for a confirmation), but have @bisq-network/support-agents been hearing many complaints about this? It's probably worth doing because the difference in sats / overall dollar value isn't huge, but if it does become significant, people might start to balk.

cd2357:
Might turn away new users who are trading 0.01 BTC or less, see #4575

The difference in fees between medium and high-prio can be small or large, depending on how the mempool distribution looks like at that moment. For example now, the high-prio fees are 5x the medium ones:

Screenshot 2020-10-02 at 08 21 13

I agree, `fastestFee` (high prio) is technically better because it reduces wait times, but it will likely have hidden counterproductive consequences.

Maybe wait with this PR until segwit is live? Combined with lower tx size, the increased fee rate might have a minimal effect on the total mining fee paid per trade.

sqrrm:
utACK

It might be a while until we have segwit for trades. I would think it better to increase the fees sooner as we're having issues right now.

wiz:
@cd2357 I'm sure everyone knows Bisq will eventually get priced out of the base layer Bitcoin blockchain, and obviously the $100 trades are already "priced out" in the sense they are not profitable if they cost $10 in fees.

While we can optimize the amount of on-chain space used for a Bisq trade with Segwit or atomic trades etc., the only way to optimize the fees paid is to trade on the weekend, when fees are expensive we need to pay them or trades get stuck which causes a lot of real problems for users.

Obviously the only real solution to this "high fee problem" is getting Bisq working on Layer 2 such as a sidechain like Liquid or a network like Lightning where fees can be kept low, in exchange for various security tradeoffs, but that is another discussion....

My motivation for this PR is because today I spent $10 in fees for my DAO voting and it was still not confirmed in time due to a sudden surge in mempool activity, presumably everyone withdrawing from BitMEX at the same time due to the news today. It can also happen due to mining variance when there are just no blocks for an hour or longer and the mempool gets backed up.

sqrrm (approved):

MwithM:
Instant trades should pay high priority fees. Normal trades, I'm not sure what's best.

PR #4586, opened 2020-10-01
Reduce initial data request

chimp1984:
This is a clean version of https://github.com/bisq-network/bisq/pull/4519 which was based on #4233 from @freimair.
I decided to remove the git commit history and make new clean commits as it was too messy before and not easy to review.

Now the commits are clearly separated between pure refactoring, small not complex changes and the feature implementation.
It is basically only 2 commits which carry the main feature implementation. I hope that makes it easier to review and to get it merged.

In the last commit I deactivated the usage if the feature as we should deply it together with https://github.com/bisq-network/bisq/pull/4577 which is not ready yet and which will require some deployment planning and coordination. By reverting that last commit the feature gets activated.

By having the code base merged but not activated we avoid that the code base gets into problems with other large code change PRs which are pending but avoid any serious risk as it is a not executed code path.
Also we gain the benefit to have the version support deployed in case we decide to release the actiavted version in a later release.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/peers/getdata/messages/PreliminaryGetDataRequest.java
+++ b/p2p/src/main/java/bisq/network/p2p/peers/getdata/messages/PreliminaryGetDataRequest.java
@@ -28,6 +28,7 @@ import protobuf.NetworkEnvelope;
 
 import com.google.protobuf.ByteString;
 
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -35,7 +36,7 @@ import lombok.EqualsAndHashCode;
 import lombok.Value;
 import lombok.extern.slf4j.Slf4j;
 
-import org.jetbrains.annotations.NotNull;
+import javax.annotation.Nullable;
 
 @Slf4j
 @EqualsAndHashCode(callSuper = true)
@@ -43,9 +44,12 @@ import org.jetbrains.annotations.NotNull;
 public final class PreliminaryGetDataRequest extends GetDataRequest implements AnonymousMessage, SupportedCapabilitiesMessage {
     private final Capabilities supportedCapabilities;
 
-    public PreliminaryGetDataRequest(int nonce,
-                                     @NotNull Set<byte[]> excludedKeys) {
-        this(nonce, excludedKeys, Capabilities.app, Version.getP2PMessageVersion());
+    public PreliminaryGetDataRequest(int nonce, Set<byte[]> excludedKeys) {
+        this(nonce,
+                excludedKeys,
+                Version.VERSION,
+                Capabilities.app,
+                Version.getP2PMessageVersion());
     }
 
 
@@ -54,34 +58,40 @@ public final class PreliminaryGetDataRequest extends GetDataRequest implements A
     ///////////////////////////////////////////////////////////////////////////////////////////
 
     private PreliminaryGetDataRequest(int nonce,
-                                      @NotNull Set<byte[]> excludedKeys,
-                                      @NotNull Capabilities supportedCapabilities,
+                                      Set<byte[]> excludedKeys,
+                                      @Nullable String version,
+                                      Capabilities supportedCapabilities,
                                       int messageVersion) {
-        super(messageVersion, nonce, excludedKeys);
+        super(messageVersion, nonce, excludedKeys, version);
 
         this.supportedCapabilities = supportedCapabilities;
     }
 
     @Override
     public protobuf.NetworkEnvelope toProtoNetworkEnvelope() {
-        final protobuf.PreliminaryGetDataRequest.Builder builder = protobuf.PreliminaryGetDataRequest.newBuilder()
+        protobuf.PreliminaryGetDataRequest.Builder builder = protobuf.PreliminaryGetDataRequest.newBuilder()
                 .addAllSupportedCapabilities(Capabilities.toIntList(supportedCapabilities))
                 .setNonce(nonce)
                 .addAllExcludedKeys(excludedKeys.stream()
                         .map(ByteString::copyFrom)
                         .collect(Collectors.toList()));
-
+        Optional.ofNullable(version).ifPresent(builder::setVersion);
         NetworkEnvelope proto = getNetworkEnvelopeBuilder()
                 .setPreliminaryGetDataRequest(builder)
                 .build();
-        log.info("Sending a PreliminaryGetDataRequest with {} kB", proto.getSerializedSize() / 1000d);
+        log.info("Sending a PreliminaryGetDataRequest with {} kB and {} excluded key entries. Requesters version={}",
+                proto.getSerializedSize() / 1000d, excludedKeys.size(), version);
         return proto;
     }
 
     public static PreliminaryGetDataRequest fromProto(protobuf.PreliminaryGetDataRequest proto, int messageVersion) {
-        log.info("Received a PreliminaryGetDataRequest with {} kB", proto.getSerializedSize() / 1000d);
+        Set<byte[]> excludedKeys = ProtoUtil.byteSetFromProtoByteStringList(proto.getExcludedKeysList());
+        String requestersVersion = ProtoUtil.stringOrNullFromProto(proto.getVersion());
+        log.info("Received a PreliminaryGetDataRequest with {} kB and {} excluded key entries. Requesters version={}",
+                proto.getSerializedSize() / 1000d, excludedKeys.size(), requestersVersion);
         return new PreliminaryGetDataRequest(proto.getNonce(),
-                ProtoUtil.byteSetFromProtoByteStringList(proto.getExcludedKeysList()),
+                excludedKeys,
+                requestersVersion,
                 Capabilities.fromIntList(proto.getSupportedCapabilitiesList()),
                 messageVersion);
     }
--- a/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
+++ b/p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
@@ -47,6 +47,8 @@ import bisq.network.p2p.storage.payload.ProtectedStoragePayload;
 import bisq.network.p2p.storage.payload.RequiresOwnerIsOnlinePayload;
 import bisq.network.p2p.storage.persistence.AppendOnlyDataStoreListener;
 import bisq.network.p2p.storage.persistence.AppendOnlyDataStoreService;
+import bisq.network.p2p.storage.persistence.HistoricalDataStoreService;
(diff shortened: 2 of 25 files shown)


p2p/src/main/java/bisq/network/p2p/storage/persistence/StoreService.java
```diff
@@ -116,21 +116,25 @@ protected void makeFileFromResourceFile(String postFix) {
                 e.printStackTrace();
             }
         } else {
-            log.debug(fileName + " file exists already.");
+            log.info("No resource file have been copied. {} exists already.", fileName);
```
ghubstan: log.info("No resource file has been copied. {} exists already.", fileName);
OR
log.info("No resource file was copied. {} exists already.", fileName);

(grammar)

p2p/src/main/java/bisq/network/p2p/peers/getdata/messages/PreliminaryGetDataRequest.java
```diff
@@ -28,24 +28,28 @@
...
-    public PreliminaryGetDataRequest(int nonce,
-                                     @NotNull Set<byte[]> excludedKeys) {
-        this(nonce, excludedKeys, Capabilities.app, Version.getP2PMessageVersion());
+    public PreliminaryGetDataRequest(int nonce, Set<byte[]> excludedKeys) {
+        this(nonce,
+                excludedKeys,
+                Version.VERSION,
+                Capabilities.app,
+                Version.getP2PMessageVersion());
     }
```
ghubstan, line 54: extra white space

proto/src/main/proto/pb.proto
```diff
@@ -95,6 +95,7 @@ message PreliminaryGetDataRequest {
     int32 nonce = 21;
```
ghubstan: Shouldn't the proto msg field number be 1, not 21?
    `int32 nonce = 1;`

p2p/src/main/java/bisq/network/p2p/storage/persistence/HistoricalDataStoreService.java
```diff
@@ -0,0 +1,187 @@
+package bisq.network.p2p.storage.persistence;
```
ghubstan, line 18: Needs license info comment?

p2p/src/main/java/bisq/network/p2p/storage/persistence/PersistableNetworkPayloadStore.java
```diff
@@ -0,0 +1,20 @@
+package bisq.network.p2p.storage.persistence;
```
ghubstan, line 18: Needs license info comment?

p2p/src/main/java/bisq/network/p2p/storage/persistence/StoreService.java
chimp1984: Will fix.

proto/src/main/proto/pb.proto
chimp1984: Upps, will fix.

p2p/src/main/java/bisq/network/p2p/peers/getdata/messages/PreliminaryGetDataRequest.java
chimp1984: Where? the 2 linebreaks between constr. and comment lines are intended.

p2p/src/main/java/bisq/network/p2p/storage/persistence/HistoricalDataStoreService.java
chimp1984: will add

p2p/src/main/java/bisq/network/p2p/storage/persistence/PersistableNetworkPayloadStore.java
chimp1984: will add

ghubstan:
I take back the comment about extra white space -- I see it is intended and why.

utACK

proto/src/main/proto/pb.proto
```diff
@@ -92,7 +92,7 @@ message BundleOfEnvelopes {
 // get data
 
 message PreliminaryGetDataRequest {
-    int32 nonce = 21;
+    int32 nonce = 1;
```
sqrrm: This is old indexing so the index has to be kept, we cannot just change the indexes since it's already in production.

p2p/src/main/java/bisq/network/p2p/storage/persistence/HistoricalDataStoreService.java
```diff
@@ -0,0 +1,187 @@
...
+        // If we have a store with a newer version than the requesters version we will add those as well.
+        storesByVersion.entrySet().stream()
+                .filter(entry -> {
+                    // Old nodes not sending the version will get delivered all data
+                    if (requestersVersion == null) {
+                        log.info("The requester did not send a version. This is expected for not updated nodes.");
+                        return true;
+                    }
+
+                    // Otherwise we only add data if the requesters version is older then
```
sqrrm, line 76:
```suggestion
                    // Otherwise we only add data if the requesters version is older than
```

p2p/src/main/java/bisq/network/p2p/storage/persistence/HistoricalDataStoreService.java
```diff
@@ -0,0 +1,187 @@
...
+                    // the version of the particular store.
+                    String storeVersion = entry.getKey();
+                    boolean newVersion = Version.isNewVersion(storeVersion, requestersVersion);
+                    String details = newVersion ?
+                            "As our historical store is a newer version we add the data to our result map." :
+                            "As the requester version is not older as our historical store we do not " +
```
sqrrm, line 82:
```suggestion
                            "As the requester version is not older than our historical store we do not " +
```

p2p/src/main/java/bisq/network/p2p/storage/persistence/HistoricalDataStoreService.java
```diff
@@ -0,0 +1,187 @@
...
+        persist();
+    }
+
+    @Override
+    protected PersistableNetworkPayload putIfAbsent(P2PDataStorage.ByteArray hash, PersistableNetworkPayload payload) {
+        if (anyMapContainsKey(hash)) {
+            return null;
+        }
+
+        PersistableNetworkPayload previous = getMapOfLiveData().put(hash, payload);
```
sqrrm: I don't understand why this is done. First there is a check if the key exists in any map, then if if doesn't we add it and return what was there. Nothing should be there since we checked nothing was there.

Only thing I can think of is some handling some concurrency issue. I think a comment would be helpful here.

p2p/src/main/java/bisq/network/p2p/storage/persistence/HistoricalDataStoreService.java
```diff
@@ -0,0 +1,187 @@
...
+        storesByVersionBuilder.put(version, historicalStore);
+        allHistoricalDataBuilder.putAll(historicalStore.getMap());
+
+        if (wasCreatedFromResources) {
+            pruneStore(historicalStore, version);
+        }
+    }
+
+    private void pruneStore(PersistableNetworkPayloadStore historicalStore, String version) {
+        int preLive = getMapOfLiveData().keySet().size();
```
sqrrm:
```suggestion
        int preLive = getMapOfLiveData().size();
```

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -293,6 +300,57 @@ public GetDataResponse buildGetDataResponse(
...
+                    if (service instanceof HistoricalDataStoreService) {
+                        var historicalDataStoreService = (HistoricalDataStoreService<? extends PersistableNetworkPayloadStore>) service;
+                        // As we add the version to our request we only use the live data. Eventually missing data will be
+                        // derived from the version.
+                        serviceMap = historicalDataStoreService.getMapOfLiveData();
+                        map.putAll(serviceMap);
+                    } else {
+                        serviceMap = service.getMap();
+                        map.putAll(serviceMap);
+                    }
```
sqrrm:
```suggestion
                    } else {
                        serviceMap = service.getMap();
                    }
                    map.putAll(serviceMap);
```

p2p/src/main/java/bisq/network/p2p/storage/P2PDataStorage.java
```diff
@@ -293,6 +300,57 @@ public GetDataResponse buildGetDataResponse(
...
+                .forEach(service -> {
+                    Map<ByteArray, PersistableNetworkPayload> serviceMap;
+                    if (service instanceof HistoricalDataStoreService) {
+                        var historicalDataStoreService = (HistoricalDataStoreService<? extends PersistableNetworkPayloadStore>) service;
+                        serviceMap = historicalDataStoreService.getMapSinceVersion(requestersVersion);
+                        map.putAll(serviceMap);
+                    } else {
+                        serviceMap = service.getMap();
+                        map.putAll(serviceMap);
+                    }
```
sqrrm:
```suggestion
                    } else {
                        serviceMap = service.getMap();
                    }
                    map.putAll(serviceMap);
```

sqrrm (changes requested):
6f5bfde seems not necessary. If this code is activated but there is no historical data resource file it should work like does now.

Some changes suggested, but the protobuf index thing should be fixed.

proto/src/main/proto/pb.proto
chimp1984: Ah true. I thought I have introduced that, but checked master and history and is indeed a old issue. Will revert it so it is 21 again. Will add a comment as well.

chimp1984:
> [6f5bfde](https://github.com/bisq-network/bisq/commit/6f5bfde92fd2904dda9566bb2f73a97263c10327) seems not necessary. If this code is activated but there is no historical data resource file it should work like does now.
> 
> Some changes suggested, but the protobuf index thing should be fixed.

We do not expect a resource file without version number, so we could not deploy the updated trade stats resources if we use the new store.

p2p/src/main/java/bisq/network/p2p/storage/persistence/HistoricalDataStoreService.java
chimp1984: Yes it will be always null, but we override the method signature from MapStoreService and follow the putIfAbsent signature from Map. In our special case it wil be always null. I can change it to make it more clear and add a comment.

chimp1984:
@sqrrm thanks for the review. Added some minor commit and applied your suggestions.

sqrrm (approved):
ACK

PR #4589, opened 2020-10-02
Persistence redesign

chimp1984:
Based on https://github.com/bisq-network/bisq/pull/4586
Replaces https://github.com/bisq-network/bisq/pull/4527

Attempt for a more clean commit history (still a lot). As merge was getting pretty difficult I decided to re-apply all changes again and compare with old PR code and with base of #4586. Hope no bugs entered by the challenging merge process.

Update: As #4586 is now merged to master this PR got rebased on master.


common/src/main/java/bisq/common/persistence/PersistenceManager.java
```diff
@@ -0,0 +1,378 @@
...
+            }
+        });
+    }
+
+    private static void onWriteCompleted(ResultHandler completeHandler,
+                                         AtomicInteger openInstances,
+                                         PersistenceManager<?> persistenceManager) {
+        persistenceManager.shutdown();
+        openInstances.decrementAndGet();
+        if (openInstances.get() == 0) {
```
sqrrm: To reduce confusion, how about
```suggestion
        if (openInstances.decrementAndGet() == 0) {
```

common/src/main/java/bisq/common/persistence/PersistenceManager.java
```diff
@@ -0,0 +1,378 @@
...
+        this.dir = checkDir(dir);
+        this.persistenceProtoResolver = persistenceProtoResolver;
+        this.corruptedStorageFileHandler = corruptedStorageFileHandler;
+    }
+
+    ///////////////////////////////////////////////////////////////////////////////////////////
+    // API
+    ///////////////////////////////////////////////////////////////////////////////////////////
+
+    public void initialize(T persistable, Source sourcce) {
```
sqrrm: Misspelled `sourcce`

chimp1984:
Rebased again latest master merge.

common/src/main/java/bisq/common/persistence/PersistenceManager.java
```diff
@@ -0,0 +1,378 @@
...
+    // Write file to disk
+    ///////////////////////////////////////////////////////////////////////////////////////////
+
+    public void requestPersistence() {
+        persistenceRequested = true;
+
+        // We write to disk with a delay to avoid frequent write operations. Depending on the priority those delays
+        // can be rather long.
+        if (timer == null) {
+            timer = UserThread.runPeriodically(() -> {
```
sqrrm: Why use `runPeriodically()` instead of `runAfter()`? The effect would be the same but now it seems the write is intended to happen periodically, but two lines later that we see that it's not.

common/src/main/java/bisq/common/proto/persistable/PersistableList.java
```diff
@@ -17,43 +17,85 @@
...
-        list = new ArrayList<>();
     }
 
     public PersistableList(List<T> list) {
-        this.list = list;
+        setAll(list);
+    }
+
+    public void addListener(ListChangeListener<T> listener) {
+        ((ObservableList<T>) getList()).addListener(listener);
```
sqrrm: This does not look safe. I think `list` should be an `ObservableList` since it's required by this interface. It's first initiated to an `ArrayList` (line 39) but I assume all implementations of `PersistableList` passes an `ObservableList` as the constructor argument.

common/src/main/java/bisq/common/proto/persistable/PersistableList.java
```diff
@@ -17,43 +17,85 @@
...
+
+    public boolean add(T item) {
+        if (!list.contains(item)) {
+            list.add(item);
+            return true;
+        }
+        return false;
+    }
+
+    public boolean remove(T tradable) {
```
sqrrm: Naming the generic `tradable` is confusing, `item` is more neutral

common/src/main/java/bisq/common/proto/persistable/PersistableList.java
```diff
@@ -17,43 +17,85 @@
...
         return list.stream();
     }
 
-    private interface ExcludesDelegateMethods<T> {
-        Stream<T> stream();
+    public int size() {
+        return list.size();
+    }
+
+    public boolean contains(T thing) {
```
sqrrm: Rather use a uniform naming for the generic, like `item`

core/src/main/java/bisq/core/app/BisqExecutable.java
```diff
@@ -148,21 +153,30 @@ protected Injector getInjector() {
     }
 
     protected void applyInjector() {
-        setupPersistedDataHosts(injector);
+        // Subclassed might configure classes with the injector here
```
sqrrm:
```suggestion
        // Subclasses might configure classes with the injector here
```

sqrrm (changes requested):

sqrrm:
Review needed from @ManfredKarrer to be able to merge this

common/src/main/java/bisq/common/persistence/PersistenceManager.java
chimp1984: fixed

common/src/main/java/bisq/common/persistence/PersistenceManager.java
chimp1984: Yes was from earlier version, should be runAfter.

common/src/main/java/bisq/common/proto/persistable/PersistableList.java
chimp1984: Yes, was moving the listeners later to the base class. But I think best here is to enforce observeable list as it is used anyway by all subclasses.

chimp1984: After looking at subclasses its not true. many dont use ObservableList. I will move back the ObservableList methods to the classes which actually use them and remove all ObservableList casts from the base class

common/src/main/java/bisq/common/proto/persistable/PersistableList.java
chimp1984: Yes, will fix

common/src/main/java/bisq/common/proto/persistable/PersistableList.java
chimp1984: Agree.

core/src/main/java/bisq/core/app/BisqExecutable.java
chimp1984: Will fix in a new commit with other changes...

ManfredKarrer (dismissed):
utACK

common/src/main/java/bisq/common/persistence/PersistenceManager.java
```diff
@@ -184,8 +184,8 @@ public PersistenceManager(@Named(Config.STORAGE_DIR) File dir,
     // API
     ///////////////////////////////////////////////////////////////////////////////////////////
 
-    public void initialize(T persistable, Source sourcce) {
-        this.initialize(persistable, persistable.getDefaultStorageFileName(), sourcce);
+    public void initialize(T persistable, Source source) {
+        this.initialize(persistable, persistable.getDefaultStorageFileName(), source);
     }
 
     public void initialize(T persistable, String fileName, Source sourcce) {
```
sqrrm: The typo continues...

common/src/main/java/bisq/common/proto/persistable/PersistableListAsObservable.java
```diff
@@ -0,0 +1,50 @@
...
+import javafx.collections.ObservableList;
+
+import java.util.List;
+
+public abstract class PersistableListAsObservable<T extends PersistablePayload> extends PersistableList<T> {
+
+    public PersistableListAsObservable() {
+    }
+
+    protected PersistableListAsObservable(List<T> list) {
```
sqrrm: Should take `ObservableList<T>`

common/src/main/java/bisq/common/persistence/PersistenceManager.java
chimp1984: ups...

common/src/main/java/bisq/common/proto/persistable/PersistableListAsObservable.java
chimp1984: No, that is called from protobug methods and there we map to a list, will be copied over to the observable in the base class. We could use a collection here as well to make it more clear that its not setting the list.

sqrrm: Right you are, since the list is created by `createList()`. No need to change, this is correct.

chimp1984: See last commit...

sqrrm (approved):
ACK

PR #4592, opened 2020-10-03
Improve bisq setup

chimp1984:
Provides a new AppStartupState class for listening for combined startup states. 
Extract domain startup code to a new class to avoid that BisqSetup gets overloaded.

We often need to wait until network and wallet is ready or other combination of startup states.
To avoid those repeated checks for the state or setting of listeners on different domains we provide here a
collection of useful states.

I did not use listeners as we want to query the state as well, so we needed fields keeping the state. Using a BooleanProperty provides both a field for keeping the state as well as a convenient way to add a listener.
E.g.:
```
appStartupState.walletAndNetworkReadyProperty().addListener((observable, oldValue, newValue) -> {
            if (newValue) {
                ....
            }
        });
```

### DIFF
--- a/core/src/main/java/bisq/core/app/AppStartupState.java
+++ b/core/src/main/java/bisq/core/app/AppStartupState.java
@@ -1,0 +1,149 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.core.app;
+
+import bisq.core.btc.setup.WalletsSetup;
+
+import bisq.network.p2p.BootstrapListener;
+import bisq.network.p2p.P2PService;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+import org.fxmisc.easybind.EasyBind;
+import org.fxmisc.easybind.monadic.MonadicBinding;
+
+import javafx.beans.property.BooleanProperty;
+import javafx.beans.property.ReadOnlyBooleanProperty;
+import javafx.beans.property.SimpleBooleanProperty;
+
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * We often need to wait until network and wallet is ready or other combination of startup states.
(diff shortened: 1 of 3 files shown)


chimp1984:
Codacy does not read comments :-). There is a reason to not make that field local. Please ignore stupid codacy bot.

ghubstan:
Note to @ghubstan : delete `bisq.core.api.StatusCheck` after this is merged, use `AppStartupState` instead.

chimp1984:
Rebased on master.

core/src/main/java/bisq/core/app/AppStartupState.java
```diff
@@ -0,0 +1,149 @@
...
+            if (walletsSetup.hasSufficientPeersForBroadcast())
+                hasSufficientPeersForBroadcast.set(true);
+        });
+
+        p2pNetworkAndWalletInitialized = EasyBind.combine(updatedDataReceived,
+                isBlockDownloadComplete,
+                hasSufficientPeersForBroadcast,
+                allDomainServicesInitialized,
+                (a, b, c, d) -> {
+                    if (a && b && c) {
```
sqrrm, line 80: Why is this check excluding `allDomainServicesInitialized`? I couldn't figure the use for this `walletAndNetworkReady` and it seems to be not used. Is there a plan for future usage or could it be removed?

core/src/main/java/bisq/core/app/AppStartupState.java
```diff
@@ -0,0 +1,149 @@
...
+        return isBlockDownloadComplete;
+    }
+
+    public boolean isHasSufficientPeersForBroadcast() {
+        return hasSufficientPeersForBroadcast.get();
+    }
+
+    public ReadOnlyBooleanProperty hasSufficientPeersForBroadcastProperty() {
+        return hasSufficientPeersForBroadcast;
+    }
```
sqrrm, line 147: These look unused. Am I missing something,

core/src/main/java/bisq/core/app/AppStartupState.java
chimp1984: So far not but in future PRs it will be used. The allDomainServicesInitialized might disappear over time or only be used were we need a ordered initialisation, but I think that need to be delegated to the domain where it is used as we did with DaoStartup class. The listeners dont know about the order, but most domains should be independent anyway, and if not it is a sign that there is a parent domain which need to do the initialisation of its children. Of if not there that we miss such a parent daomain.

core/src/main/java/bisq/core/app/AppStartupState.java
chimp1984: Yes, they are there for future usage. Some might not be needed, once we have refactored it completely we might remove some, but better to have it in place now so we can start migrating. 
At the checks for when to start a domain specific init is duplicated in many places and often not correct. This PR should start to refactor those cases and avoid those issues.

sqrrm:
I'm ready to merge this but will wait for https://github.com/bisq-network/bisq/pull/4568 to get merged first.

ripcurlx:
I removed the Codacy pattern based on the comment next to the variable. @chimp1984 Please force push to re-trigger Codacy checks. Thanks!

sqrrm (approved):
ACK

After decision to merge segwit changes directly into release branch, this can now be merged.

PR #4594, opened 2020-10-03
Tidy up CorePriceService (refactor)

ghubstan:
Remove unnecessary String.format call and do a null check.


chimp1984 (approved):
utACK

ripcurlx (approved):
utACK

PR #4596, opened 2020-10-05
Simplify GrpcOffersService#createOffer (refactor)

ghubstan:
Push the placeOffer call down into `CoreOffersService#createOffer`, and use a countdown latch to wait for async placeOffer to complete.

### DIFF
--- a/core/src/main/java/bisq/core/api/CoreOffersService.java
+++ b/core/src/main/java/bisq/core/api/CoreOffersService.java
@@ -27,6 +27,7 @@ import bisq.core.payment.PaymentAccount;
 import bisq.core.user.User;
 
 import org.bitcoinj.core.Coin;
+import org.bitcoinj.core.Transaction;
 import org.bitcoinj.utils.Fiat;
 
 import javax.inject.Inject;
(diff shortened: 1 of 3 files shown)


core/src/main/java/bisq/core/api/CoreOffersService.java
```diff
@@ -87,15 +88,16 @@ public CoreOffersService(CreateOfferService createOfferService,
...
-    Offer createOffer(String currencyCode,
-                      String directionAsString,
-                      String priceAsString,
-                      boolean useMarketBasedPrice,
-                      double marketPriceMargin,
-                      long amountAsLong,
-                      long minAmountAsLong,
-                      double buyerSecurityDeposit,
-                      String paymentAccountId) {
+    Offer createAndPlaceOffer(String currencyCode,
```
chimp1984: I think it would be more clear to not return the offer, as it is passed via the `Consumer<Offer> resultHandler`.

chimp1984 (changes requested):

core/src/main/java/bisq/core/api/CoreOffersService.java
ghubstan: Clever.   Fix is in commit 30e6dae.

chimp1984 (approved):
utACK

ripcurlx (approved):
utACK

PR #4597, opened 2020-10-05
Reduce mempool fee estimation back to "medium" priority

wiz:
Now that we've recently optimized the "medium" fee estimation algorithm
in mempool/mempool@15bb5a9, we can move Bisq back to "medium", since
always using "high" will be a bit wasteful on the weekends when the
mempool is empty.

Simulation:
new-mempool-fees

Currently:
Screen Shot 2020-10-06 at 1 37 20


ifarnung:
I certainly appreciate this as I was noticing 99 sat fees when the next block was going to include 5 sat txs.  Glad that the estimate is getting more accurate.  Great work on this, and mempool is much much better than the Earn.com. Glad that the switch was made possible.

ripcurlx (approved):
utACK

PR #4599, opened 2020-10-06
Migrate to new trade statistics

chimp1984:
Replaces #4577

This must not be merged to master before the seed nodes are deployed with that version. We should merge it to a releasebranch and then start setting up the seeds. I put it as draft to avoid accidental merge. But it is open for review.

We replace the old trade statistics which a minimal version and reduce data size from 16.5 MB to 3.9 MB. The old data gets converted to the new format.

Not updated users will not see new trade statistics and old ones only if they have peers with the old version as well. As all seed nodes will run the updated version they will not provide the old statistics. So for those users it will look weird that trade volume goes to zero over time. But I think that is an acceptable price and the alternative would be higher complexity by supporting boths trade statistics still. As 1.4.0 has so many important features I guess most active users will update quickly anyway.

If 2 traders with an old version trade they will publish both the statistics and new nodes will convert them to new trade stats using the hash of the old as hash for the new and so avoiding duplicates.
If 1 of the peers has update and the other not it will be detected via capabilities. In the new version onlt the seller publishes. So if the new version is the buyer he will not publish anyway, so we have only the old version publsihed and get converted to the new version by receiving peers. If the updated user is the seller he will check the capability of the peer and if its an old verison he will no publish so that the buyer with the old version publishes. If both users have updated then only the seller publishes.
So we have only one scenario where both publish (both have not updated) and we solve that with using the hash.

At startup we convert the existing trades stats to the new version (running in a thread, takes about 4 sec). After that we delete the old trade stats. The old trade stats will be excluded from initial data requests and response.


ripcurlx:
@chimp1984 Do we want allow empty method bodies without a comment why we do this? https://app.codacy.com/gh/bisq-network/bisq/pullRequest?prid=6288805

chimp1984:
> @chimp1984 Do we want allow empty method bodies without a comment why we do this? https://app.codacy.com/gh/bisq-network/bisq/pullRequest?prid=6288805

Yes I would prefer that. Often u use only one method of a listener interface with several methods. Those comments like // do nothing are point less anyway and further more complex explaination is mostly also not adequate.

ripcurlx:
> > @chimp1984 Do we want allow empty method bodies without a comment why we do this? https://app.codacy.com/gh/bisq-network/bisq/pullRequest?prid=6288805
> 
> Yes I would prefer that. Often u use only one method of a listener interface with several methods. Those comments like // do nothing are point less anyway and further more complex explaination is mostly also not adequate.

Ok - removed the pattern. Please force push to re-trigger Codacy validation. Thanks!

chimp1984:
Replaced by @4610

PR #4600, opened 2020-10-06
Setting jvm option  `-Djdk.module.illegalAccess=deny`

chimp1984:
Setting jvm option  `-Djdk.module.illegalAccess=deny` avoids the warning logs at startup

Not sure how to add it to gradle so its set at normal compilation as well...

User get confused by the warning as it sounds somehow alerting.

`WARNING: An illegal reflective access operation has occurred
 WARNING: Illegal reflective access by com.google.inject.internal.cglib.core.$ReflectUtils$1 (file:/Users/dev/.gradle/caches/modules-2/files-2.1/com.google.inject/guice/4.2.2/6dacbe18e5eaa7f6c9c36db33b42e7985e94ce77/guice-4.2.2.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
 WARNING: Please consider reporting this to the maintainers of com.google.inject.internal.cglib.core.$ReflectUtils$1
 WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
 WARNING: All illegal access operations will be denied in a future release`

### DIFF
--- a/desktop/package/linux/package.sh
+++ b/desktop/package/linux/package.sh
@@ -111,6 +111,7 @@ $JAVA_HOME/bin/javapackager \
     -BjvmOptions=-Xss1280k \
     -BjvmOptions=-XX:MaxRAM=4g \
     -BjvmOptions=-Djava.net.preferIPv4Stack=true \
+    -BjvmOptions=-Djdk.module.illegalAccess=deny \
     -outfile Bisq-$version \
     -v
 
--- a/desktop/package/macosx/create_app.sh
+++ b/desktop/package/macosx/create_app.sh
@@ -91,6 +91,7 @@ $JAVA_HOME/bin/javapackager \
     -srcfiles "Bisq-$version.jar" \
     -appclass bisq.desktop.app.BisqAppMain \
     -outfile Bisq \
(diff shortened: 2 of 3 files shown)


ripcurlx:
> Setting jvm option `-Djdk.module.illegalAccess=deny` avoids the warning logs at startup

I guess it is fine as we are not using Java 9 modules so far. It seems to be fixed in https://github.com/google/guice/commit/85e30beafe55551a649025d3c12e831214057412, but not released yet.

ripcurlx (approved):
utACK

cd2357:
FYI: this may work well with Java 10 and the current JavaFX, but it caused issues with guice in #4242 when the packaged app contains more recent JREs (tested with Java 11, 14 and 15).

More specifically, when opening the packaged app:
- when it was packaged with this flag: starting the app just opened an empty window (with no contents whatsoever) + logs contained guice and illegal reflection exceptions
- when it was packaged without this flag: UI worked as expected

So for now I disabled it in the new jpackage script used in #4242 with https://github.com/bisq-network/bisq/pull/4242/commits/696a788e5edcbca883c3546c36c997e6459b482c.

PR #4602, opened 2020-10-06
Re activate confirm buttons again

chimp1984:
Implements what we discussed in https://github.com/bisq-network/bisq/pull/4562

>     * Enable for buyer and altcoin seller at all times (only once arbitration is opened it is disabled as there is no way anyway to do a normal payout anymore.
> 
>     * Disable for fiat seller while mediation is open. Enable it at dispute result if no penalty was applied by mediators (derived from payout suggestion). So the mediator is the enabler for the confirm-button by his judgement expressed in the payout suggestion. This solves automatically point 2 as if the mediated payout suggestion is the same as the happy-path there is no avoid-penalty issue as no penalty was applied.
>       If seller confirms receipt in mediation the mediator can close anyway so there is only a bit of extra friction and delay due the communication but the dispute has to be closed anyway from the mediator. So I don't see that this creates any significant difference. If mediator has applied a penalty to the buyer the seller still gets his button activated (no penaly for seller - this is the only condition we check). So the seller has 2 options: accept mediated result and hope that buyer accept as well, or click confirm button and "forgive" buyer. The confirm button would stay active also after accept button has been clicked. So he can try mediated suggestion and before it goes to arbitration click the confirm button so we avoid arbitration cases.

### DIFF
--- a/core/src/main/java/bisq/core/dao/monitoring/DaoStateMonitoringService.java
+++ b/core/src/main/java/bisq/core/dao/monitoring/DaoStateMonitoringService.java
@@ -109,13 +109,13 @@ public class DaoStateMonitoringService implements DaoSetupService, DaoStateListe
     @Getter
     private boolean isInConflictWithSeedNode;
     @Getter
-    private ObservableList<UtxoMismatch> utxoMismatches = FXCollections.observableArrayList();
+    private final ObservableList<UtxoMismatch> utxoMismatches = FXCollections.observableArrayList();
 
-    private List<Checkpoint> checkpoints = Arrays.asList(
+    private final List<Checkpoint> checkpoints = Arrays.asList(
             new Checkpoint(586920, Utilities.decodeFromHex("523aaad4e760f6ac6196fec1b3ec9a2f42e5b272"))
     );
     private boolean checkpointFailed;
-    private boolean ignoreDevMsg;
+    private final boolean ignoreDevMsg;
     private int numCalls;
     private long accumulatedDuration;
 
--- a/core/src/main/java/bisq/core/support/dispute/mediation/MediationManager.java
+++ b/core/src/main/java/bisq/core/support/dispute/mediation/MediationManager.java
@@ -202,6 +202,10 @@ public final class MediationManager extends DisputeManager<MediationDisputeList>
             Trade trade = tradeOptional.get();
             if (trade.getDisputeState() == Trade.DisputeState.MEDIATION_REQUESTED ||
                     trade.getDisputeState() == Trade.DisputeState.MEDIATION_STARTED_BY_PEER) {
+                trade.getProcessModel().setBuyerPayoutAmountFromMediation(disputeResult.getBuyerPayoutAmount().value);
+                trade.getProcessModel().setSellerPayoutAmountFromMediation(disputeResult.getSellerPayoutAmount().value);
+                tradeManager.requestPersistence();
+
                 trade.setDisputeState(Trade.DisputeState.MEDIATION_CLOSED);
             }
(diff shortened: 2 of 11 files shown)


wiz (approved):
utACK - thank you very much

ripcurlx (approved):
utACK

Code looks fine. Maybe could be improved by adding a label next to the deactivated confirm button or at least a tooltip what needs to be done to proceed with the trade.

MwithM:
> We want to avoid that a fiat seller delays payout just before arbitration can be opened without any penalty.

Why is only fiat mentioned and not alts? Isn't it indifferent?

As buyer for an instant alt trade if seller clicks "payment received" 3 days late I'm being aggravated.

chimp1984:
@MwithM 
For Fiat there is a rational incentive to delay to gain more chargeback protection. For alts that does not exist. For both there is a incentive to release early to unlock the deposit. I guess the problem with lazy traders not releasing has never materialized as a bigger problem and the risk that the buyer does not accept and the trade ends up in arbitration seems to not justify to apply it to that case as well.

MwithM:
If there's supposed to be a reason to delay btc release for fiat (chargeback risk) but not for alts, it makes more sense to be able to punish altcoin sellers, because they can't even have an excuse to release the btc payment late.

Even if there's no economic incentive to release btc late, sellers have less value locked and can break trading rules without consequences if they don't need their btc soon. To me, this is something that needs to be fixed.

I know that the first implementation to solve it has been a pain for moderators and traders who open mediation too early or without a valid reason (I suffered that myself) but still there's got to be a way to make sellers more responsible.

chimp1984:
Do you know if there are such cases and if so how many? As said the option has the cost/risk that it ends up in arbitration and thats much worse then lack of punishment.

MwithM:
From my experience, this didn't happen very much, only a couple times. But it makes any trading period window just a promise that can be easily not kept, generating two issues that potentially affect all trades:
1. Trading periods aren't relevant, only timelock for delayed payout tx. Fast settlement times are a very important metric to decide to trade somewhere, that's why Bisq implemented instant trading or monero's autoconfirmation. If this is not fixed, settlements can be delayed for almost 10 or 20 days.
2. Buyers could delay the alt or fiat payment to prevent this. If sellers can't be punished for delaying the release of BTC, it would not be fair to penalize buyers and not sellers for delaying their payment. It doesn't matter the reason why they would do that.

This might not be the most important issue Bisq needs to fix, but it needs to be fixed and it's been known for a long time.

I don't know if it's feasible, but after traders solve the issues when they're not related with an unresponsive seller, could it be possible to make mediators re-activate the "payment sent" button when they get notification from seller that payment has been received? That way only seller needs to take action.

chimp1984:
I think we should not over-interpret that. It does not seem to be a big problem and confirming late is a "soft" violation and they always can lie in dispute so it cannot be fixed anyway reliably.

I think a better approach is to look into options to expand the reputation feature. Currently one can only locally mark good and bad traders or block them even. It would be more useful if we find ways how to bring that to a more network wide scale. The negative score idea was one approach. The account age witness signing is an implementation for a special case of a distributed reputation system. Maybe we can find other solutions for the satisfaction level with the peer. Not easy but likely possible.

MwithM:
Confirming late should not be a soft violation, specially with 10 and 20 days to start arbitration. Just because they can lie, it doesn't matter what explanation they give, unless it's a clear bug and they collaborate with the mediator to solve it.

Reputation just doesn't work with altcoins, and can be easily faked. That's why Bisq uses security deposits instead of reputation systems like localbitcoins.

PR #4595, opened 2020-10-03
Remove redundant DisputeAgentType enum

ghubstan:
Use existing `SupportType` enum instead, and map the valid CLI `registerdisputeagent` method's dispute-agent-type parameter names to SupportType.

Adjusted affected tests.

PR #4605, opened 2020-10-07
Fix register dipute agent test assert

ghubstan:
This should have been included in commit 5f9b1e6, PR [4595](https://github.com/bisq-network/bisq/pull/4595).


ripcurlx (approved):
utACK

PR #4607, opened 2020-10-07
Do not log full filter object in case of an invalid Filter

chimp1984:
Do not log full filter object in case of an invalid Filter

Instead log signerPubKeyAsHex and creation date. We have currently the old filter still around (intended) so we get that log called and it would confuse ppl reading the log without knowing more about the context.


ripcurlx (approved):
utACK

PR #4608, opened 2020-10-07
Add new 'getoffer offer-id' api method

ghubstan:
There are a number of use cases where a user may want to see a single offer instead of every offer for a currency pair on the buy or sell side.  The changes are:

- Add getoffer to grpc.proto
- Add new method to GrpcOffersService, CoreApi, CoreOffersService
- Add new method to CLI
- Adjust create offer tests to use this new convenience

### DIFF
--- a/core/src/main/java/bisq/core/api/CoreOffersService.java
+++ b/core/src/main/java/bisq/core/api/CoreOffersService.java
@@ -47,6 +47,7 @@ import static bisq.common.util.MathUtils.scaleUpByPowerOf10;
 import static bisq.core.locale.CurrencyUtil.isCryptoCurrency;
 import static bisq.core.offer.OfferPayload.Direction;
 import static bisq.core.offer.OfferPayload.Direction.BUY;
+import static java.lang.String.format;
 
 @Slf4j
 class CoreOffersService {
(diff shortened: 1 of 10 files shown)


core/src/main/java/bisq/core/api/CoreOffersService.java
```diff
@@ -67,6 +68,16 @@ public CoreOffersService(CreateOfferService createOfferService,
         this.user = user;
     }
 
+    Offer getOffer(String id) {
+        List<Offer> offers = offerBookService.getOffers().stream()
+                .filter(o -> o.getId().equals(id))
+                .collect(Collectors.toList());
```
sqrrm: Not wrong, but it's a bit odd to make list when we're looking for one element. `.findAny()` returns an `Optional<>` which you could then do `offer.isPresent()` and `return offer.get()` making it clear we only expect to find one offer.

ghubstan: Resolved by commit 8394bd1

chimp1984: As we expect only one I would use a findAny and create an optional instead of a list.

chimp1984: Ah see you changed that in 8394bd1f6401a0b6f6843bd3b31e3c3b391d7f3c ;-)

chimp1984 (approved):
utACK

sqrrm (approved):
utACK

PR #4609, opened 2020-10-07
Remove address prefix for mailbox messages

chimp1984:
Remove address prefix in case the peer has updated vesion as well.
Processes all mailbox messages in a thread. (1000 msgs takes about 1 sec).

EDIT: Merged master and rebased
Capability handling is still WIP

### DIFF
--- a/common/src/main/java/bisq/common/app/Capability.java
+++ b/common/src/main/java/bisq/common/app/Capability.java
@@ -41,5 +41,5 @@ public enum Capability {
     MEDIATION,                          // Supports mediation feature
     REFUND_AGENT,                       // Supports refund agents
     TRADE_STATISTICS_HASH_UPDATE,       // We changed the hash method in 1.2.0 and that requires update to 1.2.2 for handling it correctly, otherwise the seed nodes have to process too much data.
-    NO_ADDRESS_PRE_FIX                   // At 1.4.0 we removed the prefix filter for mailbox messages. If a peer has that capability we do not sent the prefix.
+    NO_ADDRESS_PRE_FIX                  // At 1.4.0 we removed the prefix filter for mailbox messages. If a peer has that capability we do not sent the prefix.
 }
--- a/common/src/main/java/bisq/common/app/Capability.java
+++ b/common/src/main/java/bisq/common/app/Capability.java
@@ -40,5 +40,6 @@ public enum Capability {
     SIGNED_ACCOUNT_AGE_WITNESS,         // Supports the signed account age witness feature
     MEDIATION,                          // Supports mediation feature
     REFUND_AGENT,                       // Supports refund agents
-    TRADE_STATISTICS_HASH_UPDATE        // We changed the hash method in 1.2.0 and that requires update to 1.2.2 for handling it correctly, otherwise the seed nodes have to process too much data.
+    TRADE_STATISTICS_HASH_UPDATE,       // We changed the hash method in 1.2.0 and that requires update to 1.2.2 for handling it correctly, otherwise the seed nodes have to process too much data.
+    NO_ADDRESS_PRE_FIX                   // At 1.4.0 we removed the prefix filter for mailbox messages. If a peer has that capability we do not sent the prefix.
 }
--- a/p2p/src/main/java/bisq/network/p2p/P2PService.java
+++ b/p2p/src/main/java/bisq/network/p2p/P2PService.java
@@ -323,8 +323,6 @@ public class P2PService implements SetupListener, MessageListener, ConnectionLis
                 "seedNodeOfPreliminaryDataRequest must be present");
 
         requestDataManager.requestUpdateData();
-        /*if (Capabilities.app.containsAll(Capability.SEED_NODE))
-            UserThread.runPeriodically(() -> requestDataManager.requestUpdateData(), 1, TimeUnit.HOURS);*/
 
         // If we start up first time we don't have any peers so we need to request from seed node.
         // As well it can be that the persisted peer list is outdated with dead peers.
--- a/p2p/src/main/java/bisq/network/p2p/P2PService.java
+++ b/p2p/src/main/java/bisq/network/p2p/P2PService.java
@@ -452,7 +452,7 @@ public class P2PService implements SetupListener, MessageListener, ConnectionLis
     private void processSingleMailboxEntry(Collection<ProtectedMailboxStorageEntry> protectedMailboxStorageEntries) {
         checkArgument(protectedMailboxStorageEntries.size() == 1);
         var decryptedEntries = new ArrayList<>(getDecryptedEntries(protectedMailboxStorageEntries));
-        if (protectedMailboxStorageEntries.size() == 1) {
+        if (decryptedEntries.size() == 1) {
             storeMailboxDataAndNotifyListeners(decryptedEntries.get(0));
         }
     }
--- a/p2p/src/main/java/bisq/network/p2p/P2PService.java
+++ b/p2p/src/main/java/bisq/network/p2p/P2PService.java
@@ -53,7 +53,6 @@ import bisq.common.crypto.PubKeyRing;
 import bisq.common.proto.ProtobufferException;
 import bisq.common.proto.network.NetworkEnvelope;
 import bisq.common.proto.persistable.PersistedDataHost;
-import bisq.common.util.Tuple2;
 import bisq.common.util.Utilities;
 
 import com.google.inject.Inject;
@@ -97,6 +96,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import lombok.Getter;
+import lombok.Value;
 
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -126,7 +126,7 @@ public class P2PService implements SetupListener, MessageListener, ConnectionLis
     private final Set<DecryptedMailboxListener> decryptedMailboxListeners = new CopyOnWriteArraySet<>();
     private final Set<P2PServiceListener> p2pServiceListeners = new CopyOnWriteArraySet<>();
     @Getter
-    private final Map<String, Tuple2<ProtectedMailboxStorageEntry, DecryptedMessageWithPubKey>> mailboxMap = new HashMap<>();
+    private final Map<String, MailboxItem> mailboxItemsByUid = new HashMap<>();
     private final Set<Runnable> shutDownResultHandlers = new CopyOnWriteArraySet<>();
     private final BooleanProperty hiddenServicePublished = new SimpleBooleanProperty();
     private final BooleanProperty preliminaryDataReceived = new SimpleBooleanProperty();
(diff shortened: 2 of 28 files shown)


common/src/main/java/bisq/common/app/Capability.java
```diff
@@ -40,5 +40,6 @@
     SIGNED_ACCOUNT_AGE_WITNESS,         // Supports the signed account age witness feature
     MEDIATION,                          // Supports mediation feature
     REFUND_AGENT,                       // Supports refund agents
-    TRADE_STATISTICS_HASH_UPDATE        // We changed the hash method in 1.2.0 and that requires update to 1.2.2 for handling it correctly, otherwise the seed nodes have to process too much data.
+    TRADE_STATISTICS_HASH_UPDATE,       // We changed the hash method in 1.2.0 and that requires update to 1.2.2 for handling it correctly, otherwise the seed nodes have to process too much data.
+    NO_ADDRESS_PRE_FIX                   // At 1.4.0 we removed the prefix filter for mailbox messages. If a peer has that capability we do not sent the prefix.
```
sqrrm:
```suggestion
    NO_ADDRESS_PRE_FIX                  // At 1.4.0 we removed the prefix filter for mailbox messages. If a peer has that capability we do not sent the prefix.
```

p2p/src/main/java/bisq/network/p2p/P2PService.java
```diff
@@ -398,58 +404,120 @@ public void onError(Throwable throwable) {
...
+        if (entries.size() > 1) {
+            threadedBatchProcessMailboxEntries(entries);
+        } else if (entries.size() == 1) {
+            processSingleMailboxEntry(entries);
+        }
+    }
+
+    private void processSingleMailboxEntry(Collection<ProtectedMailboxStorageEntry> protectedMailboxStorageEntries) {
+        var decryptedEntries = new ArrayList<>(getDecryptedEntries(protectedMailboxStorageEntries));
+        checkArgument(decryptedEntries.size() == 1);
```
sqrrm: There doesn't seem to be a guarantee that this should be `1` here. If the message in`protectedMailboxStorageEntries` is not for us, this would fail.

Not sure what the intention is here. Perhaps do
```
checkArgument(protectedMailboxStorageEntries.size() == 1);
decryptedEntries.foreach(this::storeMailboxDataAndNotifyListeners);
```

p2p/src/main/java/bisq/network/p2p/P2PService.java
```diff
@@ -398,58 +404,120 @@ public void onError(Throwable throwable) {
...
+        return decryptedEntries;
     }
 
+    @Nullable
+    private Tuple2<ProtectedMailboxStorageEntry, DecryptedMessageWithPubKey> decryptProtectedMailboxStorageEntry(
+            ProtectedMailboxStorageEntry protectedMailboxStorageEntry) {
+        try {
+            DecryptedMessageWithPubKey decryptedMessageWithPubKey = encryptionService.decryptAndVerify(protectedMailboxStorageEntry
+                    .getMailboxStoragePayload()
+                    .getPrefixedSealedAndSignedMessage()
```
sqrrm, line 498: Is your plan to keep the prefixed messages but just use a random prefix so that messages are not attributable? That would be a little bit of overhead but make for an easier transition.

Can we then get rid of the prefix later?

p2p/src/main/java/bisq/network/p2p/P2PService.java
```diff
@@ -513,61 +583,41 @@ public void onFailure(@NotNull Throwable throwable) {
         }
     }
 
+    private PrefixedSealedAndSignedMessage getPrefixedSealedAndSignedMessage(NodeAddress peersNodeAddress,
+                                                                             PubKeyRing pubKeyRing,
+                                                                             NetworkEnvelope message) throws CryptoException {
+        byte[] addressPrefixHash;
+        if (peerManager.peerHasCapability(peersNodeAddress, Capability.NO_ADDRESS_PRE_FIX)) {
```
sqrrm, line 591: Oh, I see, keep prefix structure but don't set it, makes sense.

p2p/src/main/java/bisq/network/p2p/network/Connection.java
```diff
@@ -882,4 +864,51 @@ public void run() {
...
+        return false;
+    }
+
+    @Nullable
+    private NodeAddress getSenderNodeAddress(NetworkEnvelope networkEnvelope) {
+        return getPeersNodeAddressOptional().isPresent() ?
+                getPeersNodeAddressOptional().get() :
+                networkEnvelope instanceof SendersNodeAddressMessage ?
+                        ((SendersNodeAddressMessage) networkEnvelope).getSenderNodeAddress() :
+                        null;
```
sqrrm: There is `orElse` for Optional
```suggestion
        return getPeersNodeAddressOptional().orElse(
                networkEnvelope instanceof SendersNodeAddressMessage ?
                        ((SendersNodeAddressMessage) networkEnvelope).getSenderNodeAddress() :
                        null);
```

p2p/src/main/java/bisq/network/p2p/peers/PeerManager.java
```diff
@@ -470,6 +523,17 @@ public void addToReportedPeers(Set<Peer> reportedPeersToAdd, Connection connecti
         }
     }
 
+    private void applyCapabilities(Connection connection, Capabilities capabilities) {
+        if (capabilities != null && !capabilities.isEmpty()) {
```
sqrrm: Better return early

p2p/src/main/java/bisq/network/p2p/P2PService.java
chimp1984: Yes good catch. Should be:

```
 private void processSingleMailboxEntry(Collection<ProtectedMailboxStorageEntry> protectedMailboxStorageEntries) {
        checkArgument(protectedMailboxStorageEntries.size() == 1);
        var decryptedEntries = new ArrayList<>(getDecryptedEntries(protectedMailboxStorageEntries));
        if (protectedMailboxStorageEntries.size() == 1) {
            storeMailboxDataAndNotifyListeners(decryptedEntries.get(0));
        }
    }
```

chimp1984: We get called only in case of 1 entry, also the name reflects that. But getDecryptedEntries result could be empty so we need another check.

p2p/src/main/java/bisq/network/p2p/P2PService.java
chimp1984: No, we use an empty bye array if we know that the peer has updated.

chimp1984: We cannot use null or remove it as the serialisation would fail then for not updated users (even not intended for them)

chimp1984:
Complete with dev testing...

p2p/src/main/java/bisq/network/p2p/peers/PeerManager.java
```diff
@@ -410,29 +570,22 @@ private void removeSuperfluousSeedNodes() {
...
         List<Peer> reportedPeersClone = new ArrayList<>(reportedPeers);
         Optional<Peer> reportedPeerOptional = reportedPeersClone.stream()
                 .filter(e -> e.getNodeAddress().equals(nodeAddress)).findAny();
         if (reportedPeerOptional.isPresent()) {
             Peer reportedPeer = reportedPeerOptional.get();
             removeReportedPeer(reportedPeer);
-            return reportedPeer;
-        } else {
-            return null;
         }
```
sqrrm: Easier way
```
        reportedPeersClone.stream()
                .filter(e -> e.getNodeAddress().equals(nodeAddress))
                .findAny()
                .ifPresent(this::removeReportedPeer);
```

p2p/src/main/java/bisq/network/p2p/peers/PeerManager.java
```diff
@@ -155,121 +162,274 @@ public void onMinuteTick() {
...
         if (lostAllConnections) {
             lostAllConnections = false;
             stopped = false;
-            listeners.stream().forEach(Listener::onNewConnectionAfterAllConnectionsLost);
+            listeners.forEach(Listener::onNewConnectionAfterAllConnectionsLost);
+        }
+
+        if (connection.getPeersNodeAddressOptional().isPresent()) {
+            findPeer(connection.getPeersNodeAddressOptional().get())
+                    .ifPresent(Peer::onConnection);
```
sqrrm:
```suggestion
        connection.getPeersNodeAddressOptional().flatMap(this::findPeer).ifPresent(Peer::onConnection);
```

p2p/src/main/java/bisq/network/p2p/peers/peerexchange/PeerExchangeManager.java
```diff
@@ -147,7 +147,8 @@ public void onDisconnect(CloseConnectionReason closeConnectionReason, Connection
             }, RETRY_DELAY_SEC);
         }
 
-        if (peerManager.isNodeBanned(closeConnectionReason, connection))
+        if (peerManager.isPeerBanned(closeConnectionReason, connection) &&
+                connection.getPeersNodeAddressOptional().isPresent())
             seedNodeAddresses.remove(connection.getPeersNodeAddressOptional().get());
```
sqrrm:
```suggestion
        if (peerManager.isPeerBanned(closeConnectionReason, connection))
            connection.getPeersNodeAddressOptional().ifPresent(seedNodeAddresses::remove);
```

p2p/src/main/java/bisq/network/p2p/peers/PeerManager.java
```diff
@@ -505,59 +632,62 @@ private void printReportedPeers() {
...
+    //  Persisted peers
     ///////////////////////////////////////////////////////////////////////////////////////////
 
     private boolean removePersistedPeer(Peer persistedPeer) {
-        if (persistedPeers.contains(persistedPeer)) {
-            persistedPeers.remove(persistedPeer);
-            peerList.setAll(persistedPeers);
-            persistenceManager.requestPersistence();
+        if (getPersistedPeers().contains(persistedPeer)) {
+            //getPersistedPeers().remove(persistedPeer);
```
sqrrm: I don't understand why this line is commented out. Seems the peer is not removed as it should now.

sqrrm (commented):
I think we should think about `Optional<>` as a very short stream and use in the same way we use streams. It flows nicely using `ifPresent`, `orElse` and `flatMap` rather than checking `isPresent` and then treating it as a value.

p2p/src/main/java/bisq/network/p2p/peers/PeerManager.java
chimp1984: Ups that was for dev testing as in regtest environment the failed connection attempts quickly rise and then the peers get removed... will enable it again.

p2p/src/main/java/bisq/network/p2p/peers/PeerManager.java
chimp1984: Will apply. Just did not intend to refactor too much but I agree....

p2p/src/main/java/bisq/network/p2p/peers/PeerManager.java
chimp1984: Will apply.

p2p/src/main/java/bisq/network/p2p/peers/peerexchange/PeerExchangeManager.java
chimp1984: Will apply.

chimp1984:
Replaced by #4610

Issue #3893, opened 2020-01-12
Bisq nodes leak TXID of every trade when TradeStatistics are generated


wiz:
### Background
When a Bisq trade offer is accepted, each Bisq node participating in the trade creates a TradeStatistics data object and broadcasts it to the P2P network. This trade statistics data is used by every Bisq node to generate trading volume graphs, price charts, and is also available on the Bisq Markets API service.

<img width="2557" alt="Screen Shot 2020-01-12 at 20 23 11" src="https://user-images.githubusercontent.com/232186/72218023-6a5cc100-3579-11ea-8599-9274b6eb2fb2.png">

### Issue
The TradeStatistics2 object contains excessive metadata about the trade, specifically the on-chain TXID of the maker's deposit. Unfortunately, because the offerId of every Bisq trade is mapped to the on-chain Bitcoin depositTxID, this allows malicious blockchain analysis of all Bisq trades.

Example data object:
```
  {
    "currency": "JPY", 
    "direction": "SELL",
    "tradePrice": 8791986900,
    "tradeAmount": 10000, 
    "tradeDate": 1578784489588,
    "paymentMethod": "F2F", 
    "offerDate": 1578784398352,
    "useMarketBasedPrice": true,
    "marketPriceMargin": 0.0,
    "offerAmount": 10000, 
    "offerMinAmount": 10000, 
    "offerId": "12635-224f7143-3366-46e7-9e14-7fa6f39fcb2b-125",
    "depositTxId": "9c67453e57cfc80e2c121caf54f8f739cef6c5d7e9afdceec7843436a920f9d8",
    "currencyPair": "BTC/JPY",
    "primaryMarketDirection": "SELL",
    "primaryMarketTradePrice": 87919869000000,
    "primaryMarketTradeAmount": 10000, 
    "primaryMarketTradeVolume": 8791980000
  }
```

Example blockchain analysis of this trade:
https://blockstream.info/tx/9c67453e57cfc80e2c121caf54f8f739cef6c5d7e9afdceec7843436a920f9d8?expand

### How to Reproduce
1. Start Bisq with `--dumpStatistics=true` option enabled
2. After a few minutes, a `trade_statistics.db` file will be generated in your `$HOME/.bisq/btc_mainnet/db/` datadir.
3. Extract the mapping of offer ID and deposit TXID by `grep Id trade_statistics.json`
4. Paste any Bitcoin TXID into any Bitcoin Block Explorer

### Expected Result
[...]

PR #4611, opened 2020-10-08
New trade statistics

chimp1984:
Replaces #4599

Based on #4610 . Initial commits are from #4610. Starts at commit 00bed02839a0be741dad3eef578c7ad530cda093

We replace the old trade statistics which a minimal version and reduce data size from 16.5 MB to 3.9 MB. The old data gets converted to the new format.

Not updated users will not see new trade statistics and old ones only if they have peers with the old version as well. As all seed nodes will run the updated version they will not provide the old statistics. So for those users it will look weird that trade volume goes to zero over time. But I think that is an acceptable price and the alternative would be higher complexity by supporting boths trade statistics still. As 1.4.0 has so many important features I guess most active users will update quickly anyway.

If 2 traders with an old version trade they will publish both the statistics and new nodes will convert them to new trade stats using the hash of the old as hash for the new and so avoiding duplicates.
If 1 of the peers has update and the other not it will be detected via capabilities. In the new version onlt the seller publishes. So if the new version is the buyer he will not publish anyway, so we have only the old version publsihed and get converted to the new version by receiving peers. If the updated user is the seller he will check the capability of the peer and if its an old verison he will no publish so that the buyer with the old version publishes. If both users have updated then only the seller publishes.
So we have only one scenario where both publish (both have not updated) and we solve that with using the hash.

At startup we convert the existing trades stats to the new version (running in a thread, takes about 4 sec). After that we delete the old trade stats. The old trade stats will be excluded from initial data requests and response.


wiz:
Fixes #3893

core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java
```diff
@@ -0,0 +1,355 @@
...
+        this.mediator = mediator;
+        this.refundAgent = refundAgent;
+        this.extraDataMap = ExtraDataMapValidator.getValidatedExtraDataMap(extraDataMap);
+
+        this.hash = hash == null ? createHash() : hash;
+    }
+
+    public byte[] createHash() {
+        // We create hash from all fields excluding hash itself. We use json as simple data serialisation.
+        // TradeDate is different for both peers so we ignore it for hash. ExtraDataMap is ignored as well as at
```
sqrrm, line 208: Date is not excluded from Json serialization. I think that's correct though since only seller is publishing.

I think this comment is wrong.

core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java
```diff
@@ -0,0 +1,355 @@
...
+        BLOCK_CHAINS,
+        PROMPT_PAY,
+        ADVANCED_CASH,
+        BLOCK_CHAINS_INSTANT
+    }
+
+    private final String currency;
+    private final long price;
+    private final long amount;
+    private final String paymentMethod;
```
sqrrm, line 111: Why not save this as a long? Could set first element as `UNKNOWN` for clients that haven't upgraded after a new method is added.

core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java
chimp1984: Yes will change the comment.

core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java
chimp1984: Because its more flexible for future changes. If we add a new method and a not updated user gets that data it will now show the string even the payment method is not included yet, otherwise it would show Unknown.

chimp1984: As we store only the ordinal as string, its from data size same...only when paymentMethod is stored as full string we get a bit more data.

core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java
chimp1984: Will add if there is any new bugfix PR...

PR #4612, opened 2020-10-09
Segwit for the trade protocol

oscarguindzberg:
Use segwit for trade protocol transactions.

### DIFF
--- a/common/src/main/java/bisq/common/app/Version.java
+++ b/common/src/main/java/bisq/common/app/Version.java
@@ -92,10 +92,13 @@ public class Version {
 
     // The version no. of the current protocol. The offer holds that version.
     // A taker will check the version of the offers to see if his version is compatible.
-    // Offers created with the old version will become invalid and have to be canceled.
+    // For the switch to version 2, offers created with the old version will become invalid and have to be canceled.
+    // For the switch to version 3, offers created with the old version can be migrated to version 3 just by opening
+    // the Bisq app.
     // VERSION = 0.5.0 -> TRADE_PROTOCOL_VERSION = 1
     // Version 1.2.2 -> TRADE_PROTOCOL_VERSION = 2
-    public static final int TRADE_PROTOCOL_VERSION = 2;
+    // Version 1.5.0 -> TRADE_PROTOCOL_VERSION = 3
+    public static final int TRADE_PROTOCOL_VERSION = 3;
     private static int p2pMessageVersion;
 
     public static final String BSQ_TX_VERSION = "1";
--- a/core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
@@ -36,7 +36,6 @@ import org.bitcoinj.core.AddressFormatException;
 import org.bitcoinj.core.Coin;
 import org.bitcoinj.core.ECKey;
 import org.bitcoinj.core.InsufficientMoneyException;
-import org.bitcoinj.core.LegacyAddress;
 import org.bitcoinj.core.NetworkParameters;
 import org.bitcoinj.core.SegwitAddress;
 import org.bitcoinj.core.Sha256Hash;
--- a/core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
@@ -1275,11 +1275,8 @@ public class TradeWalletService {
                 input.setScriptSig(ScriptBuilder.createInputScript(txSig, sigKey));
             }
         } else if (ScriptPattern.isP2WPKH(scriptPubKey)) {
-            // TODO: Consider using this alternative way to build the scriptCode (taken from bitcoinj master)
-            // Script scriptCode = ScriptBuilder.createP2PKHOutputScript(sigKey)
-            Script scriptCode = new ScriptBuilder().data(
-                    ScriptBuilder.createOutputScript(LegacyAddress.fromKey(transaction.getParams(), sigKey)).getProgram())
-                    .build();
+            // scriptCode is expected to have the format of a legacy P2PKH output script
+            Script scriptCode = ScriptBuilder.createP2PKHOutputScript(sigKey);
             Coin value = input.getValue();
             TransactionSignature txSig = transaction.calculateWitnessSignature(inputIndex, sigKey, scriptCode, value,
                     Transaction.SigHash.ALL, false);
--- a/core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
@@ -602,7 +602,10 @@ public class TradeWalletService {
             // We grab the signature from the makersDepositTx and apply it to the new tx input
             for (int i = 0; i < buyerInputs.size(); i++) {
                 TransactionInput makersInput = makersDepositTx.getInputs().get(i);
-                byte[] makersScriptSigProgram = getMakersScriptSigProgram(makersInput);
+                byte[] makersScriptSigProgram = makersInput.getScriptSig().getProgram();
+                if (makersScriptSigProgram.length == 0 && TransactionWitness.EMPTY.equals(makersInput.getWitness())) {
+                    throw new TransactionVerificationException("Inputs from maker not signed.");
+                }
                 TransactionInput input = getTransactionInput(depositTx, makersScriptSigProgram, buyerInputs.get(i));
                 if (!TransactionWitness.EMPTY.equals(makersInput.getWitness())) {
                     input.setWitness(makersInput.getWitness());
@@ -1178,15 +1181,6 @@ public class TradeWalletService {
         return new RawTransactionInput(input.getOutpoint().getIndex(),
                 input.getConnectedOutput().getParentTransaction().bitcoinSerialize(false),
                 input.getValue().value);
-    }
-
-    private byte[] getMakersScriptSigProgram(TransactionInput transactionInput) throws TransactionVerificationException {
-        byte[] scriptProgram = transactionInput.getScriptSig().getProgram();
-        if (scriptProgram.length == 0) {
-            throw new TransactionVerificationException("Inputs from maker not signed.");
-        }
-
-        return scriptProgram;
     }
 
     private TransactionInput getTransactionInput(Transaction depositTx,
--- a/core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
@@ -38,6 +38,7 @@ import org.bitcoinj.core.ECKey;
 import org.bitcoinj.core.InsufficientMoneyException;
 import org.bitcoinj.core.LegacyAddress;
 import org.bitcoinj.core.NetworkParameters;
+import org.bitcoinj.core.SegwitAddress;
 import org.bitcoinj.core.Sha256Hash;
 import org.bitcoinj.core.SignatureDecodeException;
 import org.bitcoinj.core.Transaction;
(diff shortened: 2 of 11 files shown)


core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
```diff
@@ -957,7 +1013,15 @@ public Transaction traderSignAndFinalizeDisputedPayoutTx(byte[] depositTxSeriali
 
         // take care of sorting!
         Script redeemScript = get2of3MultiSigRedeemScript(buyerPubKey, sellerPubKey, arbitratorPubKey);
-        Sha256Hash sigHash = payoutTx.hashForSignature(0, redeemScript, Transaction.SigHash.ALL, false);
+        Sha256Hash sigHash;
+        boolean hashedMultiSigOutputIsSegwit = !ScriptPattern.isP2SH(hashedMultiSigOutput.getScriptPubKey());
```
chimp1984: Maybe better to use "hashedMultiSigOutputIsLegacy" to avoid the 2x negation. Makes it easier to read...

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
```diff
@@ -1181,6 +1181,11 @@ private RawTransactionInput getRawInputFromTransactionInput(@NotNull Transaction
                 "input.getConnectedOutput().getParentTransaction() must not be null");
         checkNotNull(input.getValue(), "input.getValue() must not be null");
 
+        // If we would be sure all bisq nodes migrated to bitcoinj 0.15, bitcoinSerialize(true) can be used.
+        // Since there is no way to check migration to bitcoinj 0.15,
+        // Config.SEGWIT_TRADE_PROTOCOL_ACTIVE could be checked instead.
+        // Actually, given the serialized tx is used just as the connected tx of an input,
+        // we don't need the the connected tx's witness data.
```
chimp1984: duplicate "the the"

chimp1984: Isn't the input always from the users own wallet (not the peers input)? If so I don't understand the above comment. Then this code is only execute anyway if the user runs the new version and segwit could be used.

common/src/main/java/bisq/common/app/Version.java
```diff
@@ -92,10 +92,13 @@ private static int getSubVersion(String version, int index) {
...
     // The version no. of the current protocol. The offer holds that version.
     // A taker will check the version of the offers to see if his version is compatible.
-    // Offers created with the old version will become invalid and have to be canceled.
+    // For the switch to version 2, offers created with the old version will become invalid and have to be canceled.
+    // For the switch to version 3, offers created with the old version can be migrated to version 3 just by opening
+    // the Bisq app.
     // VERSION = 0.5.0 -> TRADE_PROTOCOL_VERSION = 1
     // Version 1.2.2 -> TRADE_PROTOCOL_VERSION = 2
-    public static final int TRADE_PROTOCOL_VERSION = 2;
+    // Version 1.4.1 -> TRADE_PROTOCOL_VERSION = 3
```
chimp1984: We will be on 1.4.2 or more likely 1.5.0

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
```diff
@@ -1193,10 +1196,6 @@ private RawTransactionInput getRawInputFromTransactionInput(@NotNull Transaction
 
     private byte[] getMakersScriptSigProgram(TransactionInput transactionInput) throws TransactionVerificationException {
         byte[] scriptProgram = transactionInput.getScriptSig().getProgram();
-        if (scriptProgram.length == 0) {
-            throw new TransactionVerificationException("Inputs from maker not signed.");
-        }
-
         return scriptProgram;
```
chimp1984: Variable is redundant now. Also does not throw exception anymore

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
```diff
@@ -1286,11 +1286,7 @@ private void signInput(Transaction transaction, TransactionInput input, int inpu
                 input.setScriptSig(ScriptBuilder.createInputScript(txSig, sigKey));
             }
         } else if (ScriptPattern.isP2WPKH(scriptPubKey)) {
-            // TODO: Consider using this alternative way to build the scriptCode (taken from bitcoinj master)
-            // Script scriptCode = ScriptBuilder.createP2PKHOutputScript(sigKey)
-            Script scriptCode = new ScriptBuilder().data(
-                    ScriptBuilder.createOutputScript(LegacyAddress.fromKey(transaction.getParams(), sigKey)).getProgram())
-                    .build();
+            Script scriptCode = ScriptBuilder.createP2PKHOutputScript(sigKey);
```
chimp1984, line 1272: Is createP2PKHOutputScript correct? Would expect a segwit method...

core/src/main/java/bisq/core/trade/protocol/tasks/buyer/BuyerSignsDelayedPayoutTx.java
```diff
@@ -46,7 +48,12 @@ protected void run() {
             runInterceptHook();
 
             Transaction preparedDelayedPayoutTx = checkNotNull(processModel.getPreparedDelayedPayoutTx());
+
             BtcWalletService btcWalletService = processModel.getBtcWalletService();
+            NetworkParameters params = btcWalletService.getParams();
+            Transaction preparedDepositTx = new Transaction(params, processModel.getPreparedDepositTx());
+            Coin delayedPayoutTxInputValue = preparedDepositTx.getOutput(0).getValue();
+
```
chimp1984, line 54: Why not take the input from preparedDelayedPayoutTx? If that is not possible, I would prefer to pass over the deposit tx to the signDelayedPayoutTx method and do the deail work in the TradeWalletService to keep the transaction domain more isolated there.

chimp1984 (changes requested):

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
oscarguindzberg: fixed.

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
oscarguindzberg: > duplicate "the the"

fixed

oscarguindzberg: > Isn't the input always from the users own wallet (not the peers input)? If so I don't understand the above comment. Then this code is only execute anyway if the user runs the new version and segwit could be used.

When creating the deposit tx, traders will have a tx with an input from her wallet and an input from the counterparty.  Traders collaborate creating the deposit tx. They send RawTransactionInput over p2p. So, if you are trading with a node that was not upgraded to bitcoinj 0.15 and RawTransactionInput.parentTransaction is serialized with segwit data, the other node won't be able to parse RawTransactionInput.parentTransaction correctly. 
Extra comments:
- RawTransactionInput.parentTransaction could be replaced with RawTransactionInput.parentTransactionHash since the tx is only used to obtain its hash.
- Since we upgraded the trade protocol version, it is not possible to take a trade from a not-upgraded node, but better be on the safe side. This comment made probably more sense for the previous release.

oscarguindzberg: I will update the comment to make it more clear.

common/src/main/java/bisq/common/app/Version.java
oscarguindzberg: fixed.

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
oscarguindzberg: This all started with scriptProgram.length == 0 being valid for a segwit input.
I just removed the method and moved its content to where it was being used, i.e. takerSignsDepositTx().
See takerSignsDepositTx():
```
                byte[] makersScriptSigProgram = makersInput.getScriptSig().getProgram();
                if (makersScriptSigProgram.length == 0 && TransactionWitness.EMPTY.equals(makersInput.getWitness())) {
                    throw new TransactionVerificationException("Inputs from maker not signed.");
                }
```
See https://github.com/bisq-network/bisq/pull/4612/commits/7eb808d2bf0af0b4b292ff44289459eaabcbcde4

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
oscarguindzberg: Yes, it is correct.
I know it sounds weird.
To sign a P2WPKH input you have to serialize the tx with a specific format, hash it and then sign the hash.
That specific format includes something called the scriptCode which should be: OP_DUP OP_HASH160 pubKeyHash OP_EQUALVERIFY OP_CHECKSIG. That is the same script format used in P2PKH outputs.
Detailed info: https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki

core/src/main/java/bisq/core/trade/protocol/tasks/buyer/BuyerSignsDelayedPayoutTx.java
oscarguindzberg: In bitcoin (the protocol) inputs do not include the amount.
In bitcoinj (the library) you can connect a TransactionInput to its related TransactionOutput and then ask the TransactionInput its value. But it is only possible if you connected both of them.
I moved the code to obtain the input value to TradeWalletService.

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
chimp1984: Yes, I saw. But now the method does not throw TransactionVerificationException anymore.

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
chimp1984: Ok. Maybe a comment would be good with backgorund info.

oscarguindzberg:
I deleted Config.SEGWIT_TRADE_PROTOCOL_ACTIVE

oscarguindzberg:
@chimp1984 I addressed all your comments.

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
```diff
@@ -1183,11 +1183,8 @@ private RawTransactionInput getRawInputFromTransactionInput(@NotNull Transaction
...
         checkNotNull(input.getValue(), "input.getValue() must not be null");
 
-        // If we would be sure all bisq nodes migrated to bitcoinj 0.15, bitcoinSerialize(true) can be used.
-        // Since there is no way to check migration to bitcoinj 0.15,
-        // Config.SEGWIT_TRADE_PROTOCOL_ACTIVE could be checked instead.
-        // Actually, given the serialized tx is used just as the connected tx of an input,
-        // we don't need the connected tx's witness data.
+        // bitcoinSerialize(false) is used just in case the serialized tx is parsed by a bisq node still using
+        // bitcoinj 0.14. The serialized tx is just used to obtain its hash, so the witness data is not relevant.
         return new RawTransactionInput(input.getOutpoint().getIndex(),
```
chimp1984, line 1180: getRawInputFromTransactionInput is used only by methods for creating the deposit tx. With the trade protocol update we do not have mixed cases of segwit/legacy here. Or do I miss something?

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
chimp1984: I saw you removed the method. Good so...

wiz:
is this ready to begin testing?

chimp1984:
@wiz Not really. You could test it locally but not with mainnet with real users.... 
We merge first all pending PRs to master, then create a release branch and merge segwit into it. Its a trade protocol update so master must not use that.

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
oscarguindzberg: done.

core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java
oscarguindzberg: yes, you are right, just fixed the comment to explain that.

oscarguindzberg:
@wiz I suggest you wait until I remove the WIP tag.

chimp1984 (dismissed):
utACK

oscarguindzberg:
I just squashed commits and polished commit messages, but I had not updated the codebase.

oscarguindzberg:
- WIP tag removed
- Pointed PR to release/v1.5.0

chimp1984 (approved):
utACK

PR #4615, opened 2020-10-10
Fix trade protocol check for refund agent ticket close

chimp1984:
When a refund agent closes a case we have checked that the trade protocol is the same between offer and refund agent.

This would fail once we update trade protocol version when we release segwit. So this fixes that as it only checks if the version is not the old version (1) where refund agents have not been part of the system.
Also checks for legacy arbitrator that it only can be used for version 1 cases. Not tested and not assumed that this code will be executed as old arbitrators would use the old app in case they need to still work on a old case.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/support/dispute/agent/arbitration/ArbitratorView.java
+++ b/desktop/src/main/java/bisq/desktop/main/support/dispute/agent/arbitration/ArbitratorView.java
@@ -18,6 +18,7 @@
 package bisq.desktop.main.support.dispute.agent.arbitration;
 
 import bisq.desktop.common.view.FxmlView;
+import bisq.desktop.main.overlays.popups.Popup;
 import bisq.desktop.main.overlays.windows.ContractWindow;
 import bisq.desktop.main.overlays.windows.DisputeSummaryWindow;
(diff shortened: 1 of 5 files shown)


desktop/src/main/java/bisq/desktop/main/support/dispute/agent/arbitration/ArbitratorView.java
```diff
@@ -84,4 +86,17 @@ protected SupportType getType() {
     protected DisputeSession getConcreteDisputeChatSession(Dispute dispute) {
         return new ArbitrationSession(dispute, disputeManager.isTrader(dispute));
     }
+
+    @Override
+    protected void onCloseDispute(Dispute dispute) {
+        long protocolVersion = dispute.getContract().getOfferPayload().getProtocolVersion();
+        // Only cases with protocolVersion 1 are candidates for legacy arbitration.
+        // This code path is not tested and it is not assumed that it is still be used as old arbitrators would use
+        // their old Bisq version if sill cases are pending.
```
ripcurlx: NIT: `sill` instead of `still`

ripcurlx (dismissed):
utACK - Besides a little typo: https://github.com/bisq-network/bisq/pull/4615/files#r507562986

ripcurlx (approved):
utACK

PR #4619, opened 2020-10-11
Set memo to tx after tx creation and not on broadcast success

### LINKED ISSUE #4616: Memo field is empty after memo was set at withdrawal
### Description

Memo field is empty after memo was set at withdrawal

#### Version

release testing 1.4.0

### Steps to reproduce

create a withdrawal tx (segwit) in send funds with some text set in memo field

### Expected behaviour

memo should be visible in tx list

### Actual behaviour

memo is not visible in tx list

chimp1984:
Fixes #4616

Set memo to tx after tx creation and not on broadcast success, as broadcast success might not get called (even it is broadcast -> pending btcj bug)


ripcurlx (approved):
utACK

PR #4628, opened 2020-10-12
Use a thread in PersistenceManager for reading persisted files instead of using threads in client code

chimp1984:
Bases on 1.4.0 release branch but intended to get into master after release. Set it as draft to make more clear its not ready for merge yet (though ready for review). Code changes are at commit fa2163fb0f25dd870cdec81c2b3a50229d8350af.

Use a thread in PersistenceManager for reading persisted files instead of using threads in client code

Before we use a thread in readFromResources and readAllPersisted. To avoid that client code need to deal with
threading we moved that to the PersistenceManager and changed the API accordingly so it will not return the persisted object but calls a consumer once it is completed with reading.


ManfredKarrer (approved):
utACK

sqrrm (approved):
utACK

PR #4629, opened 2020-10-12
Rename getsAll payout option in Mediation to getsHalf

huey735:
Renamed it so to better represent the outcome of this option
since the winning party only gets half of the losing party's
security deposit.


chimp1984 (changes requested):
NACK It is not the half it is the min. amount of 0.003 BTC any trader gets back to leave some incentive to accept mediated payout. 
Just when the sec deposit is the min. (0.006 BTC) it looks like its 50%.

Also that is used for refund agent as well, who do not have that limitation to payout at least that min. amount.

huey735:
@chimp1984 just to be sure, this is a static value and when you say
>it is the min. amount

It's not to be inferred that there's a max. value, right? Or any other value possible for that matter? I'm trying to find the right wording to put in place.

chimp1984:
Yes it is a hard coded value. It is there so that in any case the losing party still has an incentive to accept the suggested payout. Otherwise we would get more refund agent cases.

Not sure what you mean with max value? 
The trade amount + 2x deposit - the min refund (0.003 btc) is the max amount the winner can get.

Wording is not that easy here... and it would be nice to find a wording which also fits for refund agent where he can do a 0 btc payout to the loser. Otherwise we need to have some extra code handling it there differently (not a big deal but a bit extra effort).

huey735:
I think this clarifies it for me
>The trade amount + 2x deposit - the min refund (0.003 btc) is the max amount the winner can get.

I was wondering why the word *min* was being used. I thought, if it was a static value, the winning party would get trade amount + 1 security deposit + 0.003 btc. But following your formula, the amount they receive from the other trader's security deposit varies.

What do you think of:
`disputeSummaryWindow.payout.getsCompensated=BTC {0} also gets part of the security deposit`
This makes it clear enough that this option doesn't award the winning party all of the funds in the multisig. And as soon as the mediator/arbitrator selects it they get to see the distribution.

chimp1984 (changes requested):
NACK

It was before:
"BTC seller gets all" (or BTC buyer...)
Now: 
"BTC seller also gets compensated" (or BTC buyer...)

This radio button is meant to give to the buyer or sell the max amount (all - min deposit refund if mediator, if arbitrator all). So the meaning of "also gets compensated" in incorrect.

- Only change `displayStrings.properties` file (not other property files as those will be overwrtitten by translation tool)
Suggested string: `disputeSummaryWindow.payout.getsAll=Max. payout to BTC {0}`

huey735:
The "gets all" verbiage doesn't it's terribly counter intuitive to me since the winning party does not in fact "get all". When a mediator chooses that option is because the winning party is getting compensated for the improper behaviour on the losing party's side. But your suggestion seems like a good compromise.

As to the files that I should change. I'll restore the other languages to normal but I should also change the variables on `DisputeSummaryWindow.java`, otherwise the text will break, right?

chimp1984:
Easiest is to change only the text in the property file not the key and nothing in the java code. It is common that strings change in meaning and we do not always update the keys because of that.

wiz:
huey, this PR is changing too many unrelated things to just fix one string. let me help you re-do it if you really just want to PR that one string.

huey735:
Ok @chimp1984 @wiz I reverted the changes in this PR and changed only the string. It made sense to me to change all things related to the string and not just the text but Wiz explained to me that that's not the way to go.

ripcurlx:
@chimp1984 @wiz Could you please respond to @huey735 's comment? Thanks!

wiz (approved):
ACK

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/4629#pullrequestreview-524144909

ripcurlx:
@huey735 I just saw that you haven't signed your commits. Could you please do so and maybe squashing all commits into one?

huey735:
Replaced by https://github.com/bisq-network/bisq/pull/4825

PR #4632, opened 2020-10-12
Fix send non bsq funds

oscarguindzberg:
Fixes a bug sending non-BSQ funds from the BSQ wallet: If the destination address is segwit, an exception is thrown:

org.bitcoinj.core.AddressFormatException$InvalidCharacter: Invalid character '0' at position 17
     at org.bitcoinj.core.Base58.decode(Base58.java:132)
     at org.bitcoinj.core.Base58.decodeChecked(Base58.java:171)
     at org.bitcoinj.core.LegacyAddress.fromBase58(LegacyAddress.java:146)
     at bisq.core.btc.wallet.BsqWalletService.getPreparedSendTx(BsqWalletService.java:534)
     at bisq.core.btc.wallet.BsqWalletService.getPreparedSendBtcTx(BsqWalletService.java:525)
     at bisq.desktop.main.dao.wallet.send.BsqSendView.lambda$addSendBtcGroup$6(BsqSendView.java:300)

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
@@ -513,7 +513,7 @@ public class BsqWalletService extends WalletService implements DaoStateListener 
 
     public Transaction getPreparedSendBsqTx(String receiverAddress, Coin receiverAmount)
             throws AddressFormatException, InsufficientBsqException, WalletException, TransactionVerificationException, BsqChangeBelowDustException {
-        return getPreparedSendTx(receiverAddress, receiverAmount, bsqCoinSelector);
+        return getPreparedSendTx(receiverAddress, receiverAmount, bsqCoinSelector, false);
     }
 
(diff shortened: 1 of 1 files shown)


chimp1984 (approved):
utACK

chimp1984:
Actually I think it would  be more safe if we only do that for the getPreparedSendBtcTx method.
Now we call getPreparedSendTx from there but we could add a param to signal if segwit is permitted for outputs or not. that way we would ensure that for sending bsq it would fail when receiver address is segwit.

ripcurlx:
> Actually I think it would be more safe if we only do that for the getPreparedSendBtcTx method.
> Now we call getPreparedSendTx from there but we could add a param to signal if segwit is permitted for outputs or not. that way we would ensure that for sending bsq it would fail when receiver address is segwit.

@oscarguindzberg I'll merge it as soon as chimp's comment is addressed.

oscarguindzberg:
Just fixed the code to address @chimp1984 's comment.

chimp1984 (approved):
utACK

PR #4633, opened 2020-10-13
Apply limit after filtering for shouldTransmitPayloadToPeer

chimp1984:
We want to limit the output not in between the various filter operations.
shouldTransmitPayloadToPeer is a rather cheap operation so no need to protect that by applying the limit before that.

If maintainers prefer to not merge it into release let me know, then I will do the PR to master.


ripcurlx:
@chimp1984 If you don't think it is crucial for this release let's not merge anything additional to the release branch at this point.

Issue #4371, opened 2020-07-23
Support setting one or more user-defined block explorers


cbeams:
Currently Bisq supports a fixed set of options for which block explorer to use when following links in the application to Bitcoin transactions and addresses. These options are hard-coded at https://github.com/bisq-network/bisq/blob/release/v1.3.6/core/src/main/java/bisq/core/user/Preferences.java#L80-L99 and show up in the application here:

<img width="1312" alt="image" src="https://user-images.githubusercontent.com/301810/88283182-39f59c80-cceb-11ea-8aa1-1bc6c3154570.png">

Use of any centralized block explorer is a known privacy risk, because the service may be logging queries, including the user's IP address and other unique information and linking them to the Bitcoin addresses and/or transactions being queried. This risk can be mitigated by using a Tor-based block explorer service, but such use is still not ideal. See [these](https://en.bitcoin.it/Privacy#Blockchain_explorer_websites) [two](https://en.bitcoin.it/Privacy#Bad_privacy_example_-_Using_a_blockchain_explorer) sections of @chris-belcher's Privacy article in the Bitcoin Wiki's for more details.

The ideal approach is to use a private block explorer, i.e. one running on one's own local network that queries one's own Bitcoin Core node. Doing so eliminates the risk of being tracked by other parties as no other parties are involved.

It is easy enough to add one's own block explorer by modifying `Preferences.java` (see link above), but ideally one would be able to specify one or more custom block explorer URLs in the `Settings->Preferences` screen above. A precedent for this kind of configuration functionality can be found in the `Settings->Network Info` screen:

<img width="1312" alt="image" src="https://user-images.githubusercontent.com/301810/88284301-79bd8380-cced-11ea-9590-ef82a3c1fa81.png">

Note how, if the user selects the 'Use custom Bitcoin Core nodes' radio button, they can then provide one or more custom node addresses in the input field below. The interaction would need to be different in the case of specifying custom block explorers, because each block explorer requires two URL templates, one for transaction lookups and one for address lookups, but this existing, somewhat similar configuration functionality may be useful as a starting point when considering how to implement this feature.

While it might be easiest to support adding just one custom block explorer, it would be ideal to allow specifying multiple of them, because in the case of an explorer running on one's local network, one might wish to access it via clearnet http when at home and via Tor when remote. This would require specifying two sets of URLs. And of course the user might want to specify any number of additional custom explorer services.

cbeams:
[...]

PR #4634, opened 2020-10-13
Add http://mynode.local:3002 as a bitcoin block explorer

huey735:

### DIFF
--- a/core/src/main/java/bisq/core/user/Preferences.java
+++ b/core/src/main/java/bisq/core/user/Preferences.java
@@ -98,7 +98,8 @@ public final class Preferences implements PersistedDataHost, BridgeAddressProvid
             new BlockChainExplorer("SoChain. Wow.", "https://chain.so/tx/BTC/", "https://chain.so/address/BTC/"),
             new BlockChainExplorer("Blockchain.info", "https://blockchain.info/tx/", "https://blockchain.info/address/"),
             new BlockChainExplorer("Insight", "https://insight.bitpay.com/tx/", "https://insight.bitpay.com/address/"),
-            new BlockChainExplorer("Blockchair", "https://blockchair.com/bitcoin/transaction/", "https://blockchair.com/bitcoin/address/")
+            new BlockChainExplorer("Blockchair", "https://blockchair.com/bitcoin/transaction/", "https://blockchair.com/bitcoin/address/"),
+            new BlockChainExplorer("Mynode.local", "http://mynode.local:3002/tx/", "http://mynode.local:3002/address/")
     ));
     private static final ArrayList<BlockChainExplorer> BTC_TEST_NET_EXPLORERS = new ArrayList<>(Arrays.asList(
             new BlockChainExplorer("Blockstream.info", "https://blockstream.info/testnet/tx/", "https://blockstream.info/testnet/address/"),


chimp1984:
I get a page not found

huey735:
It is an implementation of [btc-rpc-explorer](https://github.com/janoside/btc-rpc-explorer) that can be run with the [Mynode project](http://mynodebtc.com/) locally.

chimp1984:
Ah ok. I guess that might confuse other uses as well... Better might be to add a UI to enter a custom explorer or by prog arg.

huey735:
I find it unlikely. My guess is that most people don't change the default and those that do, choose one that they're familiar with. Anyways, maybe I can add little disclaimer to the name field
`Mynode (requires dedicated node)`

cd2357:
> Add http://mynode.local:3002 as a bitcoin block explorer

Or even https://custom-ip-or-hostname:3002

Some people run their own explorer on a raspi or some other host.

But that could be a future improvement.

core/src/main/java/bisq/core/user/Preferences.java
```diff
@@ -98,7 +98,8 @@
             new BlockChainExplorer("SoChain. Wow.", "https://chain.so/tx/BTC/", "https://chain.so/address/BTC/"),
             new BlockChainExplorer("Blockchain.info", "https://blockchain.info/tx/", "https://blockchain.info/address/"),
             new BlockChainExplorer("Insight", "https://insight.bitpay.com/tx/", "https://insight.bitpay.com/address/"),
-            new BlockChainExplorer("Blockchair", "https://blockchair.com/bitcoin/transaction/", "https://blockchair.com/bitcoin/address/")
+            new BlockChainExplorer("Blockchair", "https://blockchair.com/bitcoin/transaction/", "https://blockchair.com/bitcoin/address/"),
+            new BlockChainExplorer("Mynode.local", "http://mynode.local:3002/tx/", "http://mynode.local:3002/address/")
```
cd2357, line 102: How about calling it "Local BTC RPC Explorer"?

"BTC RPC Explorer" is the official name of the explorer project: https://github.com/janoside/btc-rpc-explorer

cd2357: But no, its fine, please disregard that, its only a cosmetic thing.

cd2357:
utACK

huey735:
@cd2357 
>Some people run their own explorer on a raspi or some other host.

@chimp1984 
>Ah ok. I guess that might confuse other uses as well... Better might be to add a UI to enter a custom explorer or by prog arg.

These two suggestions make sense together. Maybe it's worth to add custom field where the user would paste in the url because as @cd2357 said many people run their own block explorer. Be it btc-rpc-explorer, mempool.space, or other. 

But implementing that in the UI is beyond my skills.

dcod3d:
There's already an open issue for using custom block explorer #4371 

But IMHO, a non public explorer should not be an option that's built into the app.

cd2357:
> But IMHO, a non public explorer should not be an option that's built into the app.

Why not?

Edit: I mean, Bisq allows people to use their own (non-public) bitcoin node. I don't see why letting them use their own (non-public) blockchain explorer would be a problem.

So from my side: ACK 👍

dcod3d:
I agree that bisq should allow (even encourage) use of an user's own blockchain explorer. However, hard coding in one company's implementation (explorer/url) isn't the right way to do it.

sqrrm:
NACK

It doesn't make sense to hard code a link that doesn't work for most people. This should be either a command line argument and/or a text field where the specific explorer could be added.

huey735:
Better implemented on https://github.com/bisq-network/bisq/pull/4733

PR #4635, opened 2020-10-13
 Change write DAO json files to disk strategy

chimp1984:
Change write DAO json files to disk strategy

We have written all files each time at each new block which accumulated to about
1 GB of data, this led to stability issues and very high disk IO for the explorer nodes.

Now we write only the data of the new block. We also remove the deletion of all
files at startup.
The dao state is still written in a monolithic file so that cannot be optimized
but we added a new directly where we add each block named by block height.
Looking up the latest written file tells us our last persisted block.
So all that data is equivalent to the monolithic daoState data which should
be removed once the webapp devs have implemented the change to use the blocks directory.

As this code is affecting only explorer nodes I think it can be merged to 1.4.0 without big risk. It would make it easier for operators to use that version instead of waiting to get it once merged into master. Alternatively they could run from my branch but I think it is more clean that way.

### DIFF
--- a/common/src/main/java/bisq/common/file/JsonFileManager.java
+++ b/common/src/main/java/bisq/common/file/JsonFileManager.java
@@ -24,13 +24,26 @@ import java.nio.file.Paths;
 import java.io.File;
 import java.io.PrintWriter;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.ThreadPoolExecutor;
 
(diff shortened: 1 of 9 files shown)


ripcurlx (approved):
utACK

sqrrm:
@ManfredKarrer This PR needs an ACK from you

ManfredKarrer (commented):
utACK

PR #4636, opened 2020-10-13
Check for limit after filtering

chimp1984:
We want to limit the output not in between the various filter operations.
shouldTransmitPayloadToPeer is a rather cheap operation so no need to protect that by applying the limit before that.

Based on 1.4.0 release branch.


chimp1984:
@sqrrm Could you review that? Would be good to get the pending PRs in asap....

sqrrm:
@chimp1984 Could you rebase on master and resolve conflicts? Will make it easier to review.

chimp1984:
This has been already solved by other PRs in the meantime.

PR #4638, opened 2020-10-13
Fix bug at mediation with old client

chimp1984:
If dispute opener is old client the delayed payout tx is not sent in mediation case (only in refund agent cases).
At 1.4.0 we send it as well in mediation case and the mediator get a warning shown in case its missing.
To avoid that warning we check if dispute is of refund agent type and only check in that case.
This can be removed once we have enforced update to 1.4.0 (segwit will require that).
Also added checks to not add null entries in the duplicates checks.


sqrrm (approved):
utACK

PR #4642, opened 2020-10-14
Change data response behaviour

chimp1984:
Add DateSortedTruncatablePayload interface for TradeStatistics2

We check first if we need to truncate dateSortedTruncatablePayloads, if so we have sorted by date and truncate in the way that we receive the most recent data. We define the maxItems in the class implementing the interface (3000 for trade stats).
Later we apply the maxEntries check the combined list and if we need to truncate here as well (10 000) we have added the dateSortedTruncatablePayloads at the end so those will get truncated with higher prio.

This code change is only relevant for seed nodes (data response). It needs to get into 1.4.0.
Old nodes send hashes for trade stats which are different to the new ones, so the seed would send all 10k trade stat objects and other important data like account age witness might get dropped as we hit the 10 000 item limit.


sqrrm (approved):
utACK

PR #4643, opened 2020-10-14
Add GetInventory messages

chimp1984:
This will be used for monitoring seed nodes.
Instead of requesting all data (we cannot request all in fact as it is too large)
we request the number of items the node has.

This code will not have any impact atm. It will be triggered once a new monitor module gets added which
will send the GetInventoryRequest to the seeds.

It would be nice to get it into 1.4.0 as then we have it already deployed for the seed nodes. Otherwise they will need to update in a few days again once the new monitor is running. The new monitor project will be a PR to master and coming soon as well :-)

Update: Here is the light weight monitor: https://github.com/chimp1984/bisq/tree/add-InventoryMonitor-module 
Will wait with making a PR until I can test it with seeds once that PR is merged.


sqrrm (approved):
utACK

PR #4645, opened 2020-10-14
Adding two torv3 seednodes, removing torv2 nodes

mrosseel:
Adding two torv3 seednodes
removing two torv2 nodes


ripcurlx (approved):
utACK

PR #4647, opened 2020-10-14
Persist seen mediator warnings

chimp1984:


ripcurlx (approved):
ACK

Fixes the issue of empty warning windows and allows mediators to decide if they want to see the specific warning window again (for the exact same mediation case).

PR #4648, opened 2020-10-14
Dont include dead transactions in check for unconfirmed txs chain

chimp1984:


ripcurlx (approved):
ACK

Fixes my issue of loitering past delayed deposit transactions blocking me to take an offer. Could be the case that this is caused of my "broken" Regtest account, but better to include this for heavy traders, who might run into similar issues.

PR #4651, opened 2020-10-15
Only remove offer locally when necessary

ripcurlx:
Based on @stejbac findings an offer shouldn't be removed locally when the trade is cancelled before any fees were paid.
Currently this causes following warning messages when cancelling the taking of an offer.
Screenshot from 2020-10-14 18-49-59


sqrrm (approved):
utACK

stejbac:
I just did a quick test on regtest. The error has now gone upon cancelling on the first step and cancelling immediately on the second step. However, if the taker goes on to fund the trade from their Bisq wallet and then cancels, the error appears again after dismissing the two warning dialog boxes about having already funded the offer.

I think one or more of the remaining `offer()` calls should be `offer(false)` calls.

Issue #4657, opened 2020-10-16
New 'Native segwit format' checkbox does not function


cbeams:
v1.4.0 (pre-release) adds a 'Native segwit format (Bech32)' checkbox to the `Funds > Receive Funds` screen, as seen below:

![paste](https://user-images.githubusercontent.com/301810/96266356-d9fb5300-0fc6-11eb-977e-145751c98261.png)

When checking the box, I would assume that the value in the 'Address' would change to reflect an available Bech32 address, but it does not. Notice how the `18ey...` P2PKH address remains the same with box checked:

![paste (1)](https://user-images.githubusercontent.com/301810/96266625-247ccf80-0fc7-11eb-9265-83e77bc1462c.png)

I do see a single unused Bech32 address available in the table of addresses at the top of the `Funds > Receive Funds` screen, and I would assume that when checking the box, this address would be automatically selected, but it is not. Likewise, I would expect that when I click to select various addresses in that table, that the Bech32 checkbox would automatically check and uncheck itself depending on whether the selected address is P2PKH or Bech32, but it does not.

It appears that the checkbox is simply a no-op right now. I have not yet tried to actually receive funds in v1.4.0, as I'm a bit concerned whether this disconnected functionality might cause a problem.

oscarguindzberg:
This is a graphic design problem.
The checkbox is associated with the "Generate new address" button, not the displayed address.
Being far away from that button may get people confused.
I will fix it.

cbeams:
Ah! Makes sense now. Thanks.

ripcurlx:
Fix merged into v1.4.1 release branch. Closing as complete.

PR #4658, opened 2020-10-16
Move segwit checkbox

oscarguindzberg:
Fixes https://github.com/bisq-network/bisq/issues/4657


oscarguindzberg:
Not sure whether this fix should get into 1.4.0.
It is a graphic design fix.

ripcurlx (approved):
utACK - I think it is fine in that case to add this to the hotfix release, as it caused confusion in our v1.4.0 pre-release.

PR #4662, opened 2020-10-16
Fix republish trade stats

chimp1984:
At init trades we iterate all trades (closed and pending) and check if the trade statistics contain a trade stat from that trade. First we check for the new TradeStatistics3, if not found we lookup for the hash of TradeStatistics2 created from that trade. If also not found and if it is valid data we republish the TradeStatistics3.

Remove republishing at SellerProtocol as we don't know the capability of the peer at that moment and as we also want to republish for completed trades.


sqrrm (approved):
utACK

PR #4670, opened 2020-10-19
[WIP] Redesign how we persist json files for explorer nodes 

chimp1984:
Currently we write at each new block all data to disk. This is about 1 GB and causes disk IO issues. We use mutable json data which also have different structure than Bisq DAO model data. This was to fulfill the needs of the original explorer which did not contain any business logic and internal model.

Now we write block chain data as immutable json data and add mutable data structures for the mutable data. Thus disk IO will decrease dramatically. We removed the json wrapper classes and use the Bisq model classes as input for the automatic json conversion. The explorer web app need to maintain its own internal model (it does already according to @softsimon) and create the missing information by lookups of the different models.
The `bisq.core.dao.state.DaoStateService` is the main class where that business logic is implemented.
The main model in Bisq is `bisq.core.dao.state.model.DaoState` and the immutable blockchain models are in `bisq.core.dao.state.model.blockchain`.

Based on release 1.4.0 branch, merged current master into it. Rebased....


wiz:
Thanks I will build this PR branch and set it up on mempool.ninja later today or tomorrow

chimp1984:
@wiz Please coordinate with @softsimon before using it. It will require quite some effort in his side to make that all work. I stripped out all the mutable data from the tx, txo, blocks.

wiz:
Yes, this is only the first phase to get new JSON sample data for simon, the production explorer won't be updated until after we modify the mempool backend to match it

ManfredKarrer (dismissed):
utACK

ManfredKarrer (dismissed):
utACK

chimp1984:
@wiz @softsimon 
Do you prefer to keep the old code base in master/release branch until the backend is ready for the changes in that PR? If so I would make that a WIP PR and make a new one with reverting the earlier changes. 
If you do not need to update to 1.4.2 or 1.5.0 (segwit) we can merge that in here but it will of course not work with old backend code...

softsimon:
Yes @chimp1984, keep the old code base until the explorer backend is ready for these new changes which will take more time

wiz:
yeah, we really appreciate you fixing this in the code, we're just a bit backed up at the moment to implement it right now so please wait a while before merging this PR

chimp1984:
Ok I will make a new PR with the revert to the old original (slow) version so that the backend works without problems.

chimp1984:
Here it is: https://github.com/bisq-network/bisq/pull/4689

chimp1984:
Still relevant

chimp1984:
@wiz @softsimon  Any progress in the web app?

wiz:
@chimp1984 Apologies for the late reply on this. Basically we still want to implement this but there were just many higher priority issues so we haven't gotten around to it. We still have the JSON dump issue for BSQ data so we will want to get this merged at some point, but not right now. Please give us some more time.

chimp1984:
Yes I understand. Its now so outdated that the PR will not serve much anymore... but let me know once you have resources for it and we coordinate it better.

PR #4672, opened 2020-10-20
Refactor offer/trade related classes in core and desktop

ghubstan:
These refactoring changes are for reducing existing and potential duplication coming with the addition of new trading protocol support in the gRPC API.  Some minor styling and logic simplification changes are also included.

- Convert OfferUtil to injected singleton, and move various offer related utility methods into it.

- Delete both MakerFeeProvider classes, which were wrappers around the same static old OfferUtil method.

- Inject OfferUtil into CreateOfferDataModel, CreateOfferViewModel,  TakeOfferDataModel, TakeOfferViewModel, MutableOfferDataModel,  MutableOfferViewModel, OfferDataModel, EditOfferDataModel,  EditOfferViewModel

- Refactor TakeOfferViewModel

	Use OfferUtil, remove unused fields & methods.
	Made minor logic simplification, style and formatting changes.

- MutableOfferDataModel

	Made minor logic simplification, style and formatting changes.

- MutableOfferView uses new paymentAccount.isHalCashAccount().

- MutableOfferViewModel

	Refactored to use new VolumeUtil, CoinUtil, OfferUtil.
	Removed unused fields & accessors.
	Made minor style change.

- Refactored OfferDataModel to use new OfferUtil

- Refactor CreateOfferService

	Inject and use OfferUtil
	Move some utility methods to OfferUtil
	Remove unused fields

- Offer

	Refactored to use new VolumeUtil for volume calculations.
	Made stateProperty and errorMessageProperty fields private.

- PaymentAccount

	Moved isHalCashAccount type check to this class.
	Moved getTradeCurrency logic to this class.

- Contract, TradeStatistics2, TradeStatistics3

	Refactored to use new VolumeUtil for volume calculations.

- Trade

	Refactored to use new VolumeUtil for volume calculations.
	Made minor logic simplification, style and formatting changes.

- CoinUtil

	Moved some coin utility methods into this class

- CoinUtilTest

	Moved (coin related) tests from CoinCryptoUtilsTest and OfferUtilTest
	into CoinUtilTest, and deleted OfferUtilTest, CoinCryptoUtilsTest.

- Adjust create and edit offer tests to model refactoring

PR #4673, opened 2020-10-20
Add 'takeoffer' API method

ghubstan:
PR https://github.com/bisq-network/bisq/pull/4672 has to be reviewed & merged before this PR.

- Add new `core.offer.takeoffer.TakeOfferModel`

	Would have been nice to move more logic from 
	`bisq.desktop.main.offer.takeoffer.TakeOfferDataModel`,
	but it has JFX dependencies that cannot be use in :core.

- Add grpc protos to support `takeoffer`, `confirmpaymentsent`, `confirmpaymentreceived`

	Only `takeoffer` is implemented in this commit.

- Refactor `OfferInfo` grpc proto wrapper, and add offer state field

- Add new `TradeInfo` grpc proto wrapper

- Implement `takeoffer` on server and cli side

- Refactor offer/trade tests, add test cases


apitest/src/test/java/bisq/apitest/method/trade/TakeBuyBTCOfferTest.java
```diff
@@ -0,0 +1,132 @@
...
+    @Test
+    @Order(1)
+    public void testTakeAlicesBuyOffer() {
+        try {
+            var alicesOffer = createAliceOffer(alicesAccount, "buy", "usd", 12500000);
+            var offerId = alicesOffer.getId();
+
+            // Wait for Alice's AddToOfferBook task.
+            // Wait times vary;  my logs show >= 2 second delay.
+            sleep(3000);
```
sqrrm, line 75: All these wait times in the test will make the test very slow eventually. I don't really have a solution to it, more of a general observation.

Maybe grouping all the tasks that need waiting on and then run the tests, but that would also group the tests, making them less independent and thus less useful.

apitest/src/test/java/bisq/apitest/method/trade/TakeBuyBTCOfferTest.java
```diff
@@ -0,0 +1,132 @@
...
+    public void testBobsConfirmPaymentReceived() {
+        var trade = getTrade(bobdaemon, tradeId);
+        assertNotNull(trade);
+
+        confirmPaymentReceived(bobdaemon, trade.getTradeId());
+        sleep(3000);
+
+        trade = getTrade(bobdaemon, tradeId);
+        // TODO is this a bug?  Why is offer.state == available?
+        assertEquals(AVAILABLE.name(), trade.getOffer().getState());
```
sqrrm, line 129: I suspect it's keeping the last state after the offer was taken, although I haven't looked into it before. There is no `TAKEN` state either. I think it's used to handle the state of the offer during its life cycle, which ends once the offer is taken and everything is then handled by the `Trade`.

core/src/main/java/bisq/core/api/CoreTradesService.java
```diff
@@ -0,0 +1,91 @@
...
+                takeOfferModel.getTakerFee(),
+                takeOfferModel.isCurrencyForTakerFeeBtc(),
+                offer.getPrice().getValue(),
+                takeOfferModel.getFundsNeededForTrade(),
+                offer,
+                paymentAccountId,
+                useSavingsWallet,
+                trade -> {
+                    resultHandler.accept(trade);
+                },
```
sqrrm, line 74:
```suggestion
                resultHandler::accept,
```

core/src/main/java/bisq/core/api/CoreTradesService.java
```diff
@@ -0,0 +1,91 @@
...
+
+
+    Trade getTrade(String tradeId) {
+        return getTradeWithId(tradeId);
+    }
+
+    private Trade getTradeWithId(String tradeId) {
+        return tradeManager.getTradeById(tradeId).orElseThrow(() ->
+                new IllegalArgumentException(format("trade with id '%s' not found", tradeId)));
+    }
```
sqrrm, line 90: I don't understand why this has to be split in two methods, plans for future PRs?

core/src/main/java/bisq/core/offer/takeoffer/TakeOfferModel.java
```diff
@@ -0,0 +1,293 @@
...
+
+        // trade fee tx: 260 bytes (1 input)
+        // deposit tx: 336 bytes (1 MS output+ OP_RETURN) - 414 bytes (1 MS output + OP_RETURN + change in case of smaller trade amount)
+        // payout tx: 371 bytes
+        // disputed payout tx: 408 bytes
+
+        // Set the default values (in rare cases if the fee request was not done yet we get the hard coded default values)
+        // But the "take offer" happens usually after that so we should have already the value from the estimation service.
+        txFeePerByteFromFeeService = feeService.getTxFeePerByte();
+        txFeeFromFeeService = offerUtil.getTxFeeBySize(txFeePerByteFromFeeService, feeTxSize);
```
sqrrm, line 143: Is this really necessary for the api? We probably want to wait for the fee to be calculated properly as an atomic part of initModel.

core/src/main/java/bisq/core/offer/takeoffer/TakeOfferModel.java
```diff
@@ -0,0 +1,293 @@
...
+
+        // We request to get the actual estimated fee
+        log.info("Start requestTxFee: txFeeFromFeeService={}", txFeeFromFeeService);
+        feeService.requestFees(() -> {
+            txFeePerByteFromFeeService = feeService.getTxFeePerByte();
+            txFeeFromFeeService = offerUtil.getTxFeeBySize(txFeePerByteFromFeeService, feeTxSize);
+            calculateTotalToPay();
```
sqrrm, line 150: Fees are calculated here, after getting the reply from feeService.

core/src/main/java/bisq/core/offer/takeoffer/TakeOfferModel.java
```diff
@@ -0,0 +1,293 @@
...
+        this.amount = valueOf(Math.min(offer.getAmount().value, getMaxTradeLimit()));
+        this.securityDeposit = offer.getDirection() == SELL
+                ? offer.getBuyerSecurityDeposit()
+                : offer.getSellerSecurityDeposit();
+        this.isCurrencyForTakerFeeBtc = offerUtil.isCurrencyForTakerFeeBtc(amount);
+        this.takerFee = offerUtil.getTakerFee(isCurrencyForTakerFeeBtc, amount);
+
+        calculateTxFees();
+        calculateVolume();
+        calculateTotalToPay();
```
sqrrm, line 114: This is already done in calculateTxFees

core/src/main/java/bisq/core/offer/takeoffer/TakeOfferModel.java
```diff
@@ -0,0 +1,293 @@
...
+    private Coin totalAvailableBalance;
+    @Getter
+    private Coin balance;
+    @Getter
+    private boolean isBtcWalletFunded;
+    @Getter
+    private Volume volume;
+
+    @Inject
+    public TakeOfferModel(AccountAgeWitnessService accountAgeWitnessService,
```
sqrrm, line 83: There's a lot of code that would be nice to reuse from `TakeOfferDataModel` here. Can be done later though in some pure refactor commit.

sqrrm (changes requested):
Some smaller comments, but over all I think it looks good again.

apitest/src/test/java/bisq/apitest/method/trade/TakeBuyBTCOfferTest.java
ghubstan: I think it's reach 6m already...
I don't see a way around it now, but over time, old, stable 'method' (or 'unit') test cases can be consolidated in 'scenario', then 'e2e' test cases, which would run them all in one test case.  That will help.   I was thinking of a dev/test cycle like this: a new feature/bug is tested in a method/unit test, then gradually moved to older scenario/e2e test cases.

apitest/src/test/java/bisq/apitest/method/trade/TakeBuyBTCOfferTest.java
ghubstan: I looked odd, but I don't see any bad side effects yet.  Should I leave those comments?  Reword them?

core/src/main/java/bisq/core/api/CoreTradesService.java
ghubstan: I will add this change to https://github.com/bisq-network/bisq/pull/4679 (the tip of the PR/branch chain).  OK?

core/src/main/java/bisq/core/api/CoreTradesService.java
ghubstan: There's no need to split it, not sure why I did that.  I will change in https://github.com/bisq-network/bisq/pull/4679 (the tip of the PR/branch chain).

apitest/src/test/java/bisq/apitest/method/trade/TakeBuyBTCOfferTest.java
sqrrm: Sounds like a plan. If the tests never run they're useless, so it's important that they can be run within a reasonable time frame. Might also be good with a consolidated scenario to find interdependency issues.

apitest/src/test/java/bisq/apitest/method/trade/TakeBuyBTCOfferTest.java
sqrrm: Best remove the `TODO` but leave a comment if you feel it's warranted.

core/src/main/java/bisq/core/api/CoreTradesService.java
ghubstan: Resolved in commit fa0e05a, PR https://github.com/bisq-network/bisq/pull/4679.

core/src/main/java/bisq/core/api/CoreTradesService.java
ghubstan: Resolved in commit 31a3119, for PR https://github.com/bisq-network/bisq/pull/4679

core/src/main/java/bisq/core/offer/takeoffer/TakeOfferModel.java
ghubstan: I mentioned that I wanted to move more from `TakeOfferDataModel into core `TakeOfferModel`, but I could not pass JFX property values into the new core model.   If you or @chimp1984 know how, do tell.

sqrrm: Don't think we want to have any jfx dependency in core, rather some stuff could probably be rewritten in `TakeOfferDataModel` to utilize the new class.

core/src/main/java/bisq/core/offer/takeoffer/TakeOfferModel.java
ghubstan: Resolved by commit d463dd1, in PR https://github.com/bisq-network/bisq/pull/4679.

core/src/main/java/bisq/core/offer/takeoffer/TakeOfferModel.java
ghubstan: Resolved by commit d463dd1, in PR #4679.

core/src/main/java/bisq/core/offer/takeoffer/TakeOfferModel.java
ghubstan: Resolved by commit 1f3554e in PR https://github.com/bisq-network/bisq/pull/4679.

PR #4676, opened 2020-10-21
Show current tx fee rate on main screen

### LINKED ISSUE #4412: Show current tx fee rate on main screen
### Description

It would be very useful for the current fee estimation to be shown on the main screen - at the moment the only way to see it is to watch the Bisq log file for a message like this:

    INFO  b.c.p.f.FeeService: BTC tx fee: txFeePerByte=11 

If `txFeePerByte=11` was shown in the main screen it would be much easier to determine if it is a good time to make a transaction.
It is more useful than the `0.00 BTC Reserved` and `0.00 BTC Locked` currently shown because they can also be found on the Funds screen.

#### Version

all Bisq, up to currently 1.3.7

### Expected behaviour

Show the tx fee on the main screen, to the right of Available balance, replacing Reserved/Locked balances.

ghost:
Tx fee rate is shown in the footer alongside `Synchronized with Bitcoin Mainnet (using Tor) (x sat/vB)`.

Fixes #4412

### Screenshot:

image

### DIFF
--- a/core/src/main/java/bisq/core/app/P2PNetworkSetup.java
+++ b/core/src/main/java/bisq/core/app/P2PNetworkSetup.java
@@ -17,6 +17,7 @@
 
 package bisq.core.app;
 
+import bisq.core.btc.setup.WalletsSetup;
 import bisq.core.locale.Res;
 import bisq.core.provider.price.PriceFeedService;
 import bisq.core.user.Preferences;
@@ -50,6 +51,7 @@ import javax.annotation.Nullable;
 public class P2PNetworkSetup {
     private final PriceFeedService priceFeedService;
     private final P2PService p2PService;
+    private final WalletsSetup walletsSetup;
     private final Preferences preferences;
 
     @SuppressWarnings("FieldCanBeLocal")
@@ -73,10 +75,12 @@ public class P2PNetworkSetup {
     @Inject
     public P2PNetworkSetup(PriceFeedService priceFeedService,
                            P2PService p2PService,
+                           WalletsSetup walletsSetup,
                            Preferences preferences) {
 
         this.priceFeedService = priceFeedService;
         this.p2PService = p2PService;
+        this.walletsSetup = walletsSetup;
         this.preferences = preferences;
     }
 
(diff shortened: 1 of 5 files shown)


ripcurlx:
I think we might need more context for everyone to understand this number. What do you think about changing it to: `Synchronized with Bitcoin Mainnet (using Tor) / Bitcoin fee: x sat/vB`? Or something like that.

cd2357:
> I think we might need more context for everyone to understand this number. What do you think about changing it to: `Synchronized with Bitcoin Mainnet (using Tor) / Bitcoin fee: x sat/vB`? Or something like that.

Maybe instead of `Bitcoin fee: x sat/vB` => `Feerate x sat/vB`?

Fewer characters, less space taken, with same meaning.

ghost:
image

image

IMO both suggestions make it worse!  [Just an opinion]. :trollface: ...but I'm OK to be overruled. :stuck_out_tongue: :v:

cd2357:
The screenshots are very persuasive :)  Your initial version looks the best.

chimp1984:
@jmacxx Consider to only show it at final sync state, so the space issues are not a problem as only at sync it gets long. Another option is to hide the "synchroonized with btc network" after a while or shorten it. The "using Tor" is not needed IMO.
I agree with @ripcurlx that we should show some context, as short as possible. @m52go Might have some good idea?
Maybe the whole footer text can be reviewed and shortened?  Without context its a bit weird as the other data is about network. Fee rate is another domain.

m52go:
- Shorten "synchronized" to "synced"
- Move Bitcoin network peer count to right with Bisq network peer count

Suggestion:

Left:
`Synced with Bitcoin Mainnet (using Tor) / Current fee rate: 86 sat/vB`

Right:
`Bitcoin network peers: 9 / Bisq network peers: 9`

ghost:
@wiz Could you weigh in with an opinion?


image

ghost:
@chimp1984 There is text added to the footer in the desktop [MainViewModel.java](https://github.com/bisq-network/bisq/blob/b9e1398e295d9ae287875f7a22b3d5cef6be88ef/desktop/src/main/java/bisq/desktop/main/MainViewModel.java#L599) which is unknown at the point where the bulk of the synced message is built in WalletAppSetup.java.  So not sure how to avoid that.  The `Synchronizing with DAO` pops in to the message briefly at each block received.

chimp1984:
Not sure as well without trying it out myself, but I think we do not need to show all info all the time. So if there is a dao block sync maybe remove some of the other info text. Or shorten it as well. Update messages are displayed there as a well...

ghost:
Ready!

image

image

image

wiz:
adding the rate as a number is good, but also visualizing it with colors or a progress bar meter thing would be better

chimp1984:
> adding the rate as a number is good, but also visualizing it with colors or a progress bar meter thing would be better

I fear that would overload it. UI is already heavy with information...

wiz:
good point, but just something simple like ```<span color=red>100 sat/vB</span>``` would go a long way, would just need to think about what colors for what ranges

chimp1984:
Maybe @pedromvpg could add his opinion or make a suggestion? But probably not important enough to waste his precious time...

ghost:
I'd prefer to keep it simple and be done with it.  Only really wanted (x sat/vB) on display.

pedromvpg:
I think it looks great an it would be very helpful.

Thinking that adding color would potentially call too much attention to itself.
Would this be primarily to inform users fee's are particularly high? In relation to what? Some previous span of time? versus a 1 sat/byte fee?

Also, do we need to clarify the fee rate targets 1 block confirmation?

chimp1984 (approved):
utACK

ripcurlx (approved):
utACK

PR #4677, opened 2020-10-21
Fix wrong HISTORY version entry.

chimp1984:
This fields contains the versions where we have tagged a resource file.
It does not need to be the latest version in case we have not added a new resource file.


ripcurlx:
Superseded by https://github.com/bisq-network/bisq/pull/4678

PR #4678, opened 2020-10-21
Rename history field in version

chimp1984:
This field contains the versions where we have tagged a resource file.
It does not need to be the latest version in case we have not added a new resource file.

Additional to the fix in #4677 this comes with a renaming to make usage more clear.

I leave it to maintainers which PR to pick for the release fix...


ripcurlx (approved):
utACK

PR #4680, opened 2020-10-21
Add inventory monitor module

chimp1984:
This adds a new module containing the new inventory monitor project. 
It is running since a few weeks under http://46.101.179.224/ and some of the seeds have updated to that branch already.
It is the implementation of https://github.com/bisq-network/projects/issues/45 
@jmacxx is working on a webapp which should finally replace the simple http server used atm.
It will still need fine tuning of the deviation alert/warning thresholds but it is already pretty useful and helps to detect seed node issues.
Most of the code is only relevant to the monitor project and seed nodes, so it carries very low risk for normal users.

PR #4685, opened 2020-10-22
Update monitor with last version from freimairs repo

chimp1984:
Update monitor code base with latest version form @freimair [repo](https://github.com/freimair/bisq/commits/monitor). 
Current version does not work due a nullpointer in connection. This was solved in @freimair's repo, but we removed it in that PR as it will have a better solution in https://github.com/bisq-network/bisq/pull/4680. We will make another new PR where this PR will be based on https://github.com/bisq-network/bisq/pull/4680 so the monitor become valid again.

### DIFF
--- a/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessStore.java
+++ b/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessStore.java
@@ -35,7 +35,7 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 public class AccountAgeWitnessStore extends PersistableNetworkPayloadStore<AccountAgeWitness> {
 
-    AccountAgeWitnessStore() {
+    public AccountAgeWitnessStore() {
     }
 
 
--- a/core/src/main/java/bisq/core/trade/statistics/TradeStatistics3Store.java
+++ b/core/src/main/java/bisq/core/trade/statistics/TradeStatistics3Store.java
@@ -35,7 +35,7 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 public class TradeStatistics3Store extends PersistableNetworkPayloadStore<TradeStatistics3> {
 
-    TradeStatistics3Store() {
+    public TradeStatistics3Store() {
     }
 
 
--- a/monitor/src/main/java/bisq/monitor/AvailableTor.java
+++ b/monitor/src/main/java/bisq/monitor/AvailableTor.java
@@ -17,13 +17,15 @@
 
 package bisq.monitor;
 
-import java.io.File;
-import org.berndpruenster.netlayer.tor.Tor;
 import bisq.network.p2p.network.TorMode;
+
+import org.berndpruenster.netlayer.tor.Tor;
+
+import java.io.File;
 
 /**
  * This class uses an already defined Tor via <code>Tor.getDefault()</code>
- * 
+ *
  * @author Florian Reimair
  *
  */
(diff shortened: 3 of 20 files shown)


sqrrm (dismissed):
utACK

sqrrm (approved):
utACK

freimair:
ahm, guys, there is a reason why I did not create a PR for my local changes to the monitor:
- the stuff would affect every users, but it is only needed for the monitor (although you reverted the change anyways)
- the new metrics are not production ready because the accompanying project https://github.com/bisq-network/projects/issues/22 has not been granted (I only started work because in the very early days of projects, I was told that the project is of interest and I can start work on it)

so @chimp1984 you pretty much pulled in stuff created by other people without even contacting the people you pulled the stuff from and the reviewer @sqrrm  did not even fire the thing up. That is not good practice!

chimp1984:
The only non-monitor change was in Connection which I reverted and is replaced by another better solution from another PR.
The monitor as it checked in causes immediately a nullpointer. Not keeping the project up to data since April is no good practice. I don't know which code base you use for the monitor webpage, there is also nothing checked in from the grafana side and I assume there is quite a bit to customize/config to be able to run the server to deliver the same result as the monitor. Do you have any repo for that? If no can you create one? You got paid quite a lot of BSQ for a monitor which never fulfilled its goals - to be an alert system for seed nodes. I expect that at least we can re-use the product we have paid for without re-building all the non-visible stuff (grafana config).

freimair:
> which never fulfilled its goals - to be an alert system for seed nodes

that is not entirely true. You before all should know that.

Back in the days we started monitoring stuff because we wanted to find out how healthy the Bisq network is. First task, for example, was to see if the Tor network is ok, so we can establish a baseline for Bisq's p2p network performance. Since then, we added metrics of areas which seemed interesting and we learned a whole lot about our p2p network. Being an alert system for the seed nodes has only been another use case that got added later. Sadly, we never managed to pull it off, simply because it turned out to be very hard to rule out false positives. We even tried binding compensation rates to seednode uptimes (https://monitor.bisq.network/d/G0Mk_uIWk/seednode-uptime-playground?orgId=1), however, these efforts stalled as well because nobody wanted to make a decision. The most recent addition has been the network size estimation which clearly showed us that although we expected thousands of users, bisq only has <400 at any time (meaning we have ~33 clients per seednode, and a seednode can host 50 - so Bisq's p2p network is actually hardly decentralized). Plus, in the last months, the initial data request sizes outgrew the capabilities of the monitors Tor uplink (because first, data sizes grew significantly, and second, we once had 6 seednodes, now we have 12) and hence, some metrics started glitching out more than I liked it (please be aware that I formulated and implemented a fix months ago which you recently took over).

All in all, the monitor helped finding issues and design counter strategies. And that is what the monitor was created for - that is what I and others got paid for. The bad part is that the issues we found turned out to be very substantial and improving the situation requires deep changes in the p2p part of the code. Thus the counter strategies either never got merged or have been reverted shortly after being applied (the daily seednode restart prevails and seednodes are not synced up properly still), other changes where tagged a "hosting issue" while totally ignoring the data.

chimp1984:
> Being an alert system for the seed nodes has only been another use case that got added later.

Maybe you saw that this way, for me it was always the main goal.

freimair:
well, the monitoring project has been designed that way and accepted that way. So I did exactly that. Here is the accompanying proposal: https://github.com/bisq-network/proposals/issues/62

PR #4692, opened 2020-10-23
Block on tx-fee-request in core TakeOfferModel init

ghubstan:
The core api's take offer model needs to wait on async operations during initialization.

Added license comment too.

### DIFF
--- a/core/src/main/java/bisq/core/offer/takeoffer/TakeOfferModel.java
+++ b/core/src/main/java/bisq/core/offer/takeoffer/TakeOfferModel.java
@@ -1,3 +1,20 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
 package bisq.core.offer.takeoffer;
 
 import bisq.core.account.witness.AccountAgeWitnessService;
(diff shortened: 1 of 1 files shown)


core/src/main/java/bisq/core/offer/takeoffer/TakeOfferModel.java
```diff
@@ -141,12 +160,19 @@ private void calculateTxFees() {
...
-        });
+        txFeePerByteFromFeeService = getTxFeePerByte();
+        txFeeFromFeeService = offerUtil.getTxFeeBySize(txFeePerByteFromFeeService, feeTxSize);
+        log.info("{} txFeePerByte = {}", feeService.getClass().getSimpleName(), txFeePerByteFromFeeService);
+    }
+
+    private Coin getTxFeePerByte() {
+        try {
+            CompletableFuture<Void> feeRequestFuture = CompletableFuture.runAsync(feeService::requestFees);
+            feeRequestFuture.get();  // Block until async fee request is complete.
```
chimp1984, line 171: I assume that method is only used by API so far. But to make it usable for GUI app as well a blocking request is not possible as it would block the UI thread. Is there a strong reason why you need to do it blocking and not via a callback/result handler design?

sqrrm (approved):
utACK

@chimp1984 what was the reason for not waiting for the fee here? Seems like it should.

chimp1984 (commented):
> utACK
> 
> @chimp1984 what was the reason for not waiting for the fee here? Seems like it should.

I don't have the full context here, but I prefer to not add blocking code, but deal with asycn calls via callbacks. But if its API only and hard otherwise I am fine...

core/src/main/java/bisq/core/offer/takeoffer/TakeOfferModel.java
sqrrm: My bad, thought this was already used by UI. Would be better to handle this with callbacks. That way it should be usable by both API and UI eventually.

ghubstan: If I understand correctly (@sqrrm), you want this `TakeOfferModel` snippet changed to using a callback instead of a `CompletableFuture`.

Do you mind if I make this change in a later PR?  I am working on a chain of 15 pr/branches, and this is branch #1 in the chain.

sqrrm: Yeah, can be done in a later PR. Having it block is still more correct than before and since this is only used by API for now, it's not a problem to merge.

PR #4696, opened 2020-10-24
Add fields to grpc TradeInfo proto & wrapper

ghubstan:
Checking trade status via the api requires additional proto msg fields.

https://github.com/bisq-network/bisq/pull/4692 should be reviewed and merged before this one.

### DIFF
--- a/core/src/main/java/bisq/core/api/model/OfferInfo.java
+++ b/core/src/main/java/bisq/core/api/model/OfferInfo.java
@@ -125,13 +125,9 @@ public class OfferInfo implements Payload {
                 .build();
     }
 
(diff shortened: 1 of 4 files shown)


sqrrm (approved):
utACK

Is it really worth having this TradeInfo structure rather than using Trade itself?

ghubstan:
> utACK
> 
> Is it really worth having this TradeInfo structure rather than using Trade itself?

It seems so because the client cannot depend on core, where Trade lives.  But I will investigate.

sqrrm:
That's probably right actually, there's a lot of stuff in Trade that you wouldn't want to expose. This way you get more control of what's exposed.

PR #4697, opened 2020-10-24
Replace TradeUtils with singleton TradeUtil

ghubstan:
The API is going to need some desktop trade utilities, which should be shared between `:desktop` and `:core.api`.   

The next PR in this chain will move trade related utilities in desktop model classes into this new singleton.

https://github.com/bisq-network/bisq/pull/4696 should be reviewed and merged before this one.


sqrrm (approved):
utACK

PR #4699, opened 2020-10-25
Refactor PendingTradesViewModel methods -> TradeUtil & OfferUtil

ghubstan:
PR https://github.com/bisq-network/bisq/pull/4697 should be reviewed & merged before this PR.

PR #4700, opened 2020-10-25
Refactor PendingTradesDataModel methods -> TradeUtil

ghubstan:
Scope of this refactoring is small;  more can be done, but the short term goal of the trade utils refactoring is to share trade util logic with core api.
    
- Removed unused method getCurrencyCode()
- Made minor style changes
- Removed duplicated code block

PR https://github.com/bisq-network/bisq/pull/4699 should be reviewed & merged before this PR.


sqrrm (approved):
utACK

PR #4701, opened 2020-10-25
Add getRole(tradeId) to core api

ghubstan:
API users will need to see their role as maker/taker when looking at trade details.

- Add getRole(trade) to TradeUtil
- Add getTradeRole(tradeId) to CoreApi, CoreTradesService
- Add role field to TradeInfo proto its wrapper class

PR https://github.com/bisq-network/bisq/pull/4700 should be reviewed & merged before this PR.


ghubstan:
This does not add a `getrole` api method to the CLI, it just makes the user's trade role available within the output from a `gettrade trade-id` call.   Future trade utils refactoring may make this cleaner, and push the coreApi.getRole call down & out of GrpcTradesService.

core/src/main/java/bisq/core/trade/TradeUtil.java
```diff
@@ -173,6 +174,27 @@ public String getPaymentMethodNameWithCountryCode(Trade trade) {
         return paymentMethodDescription;
     }
 
+    /**
+     * Returns a string describing a trader's role for a given trade.
+     * @param trade Trade
+     * @return String describing a trader's role for a given trade
+     */
+    public String getRole(Trade trade) {
```
sqrrm, line 182: Could consider renaming to `getMyRole` but perhaps that's rather implicit.

core/src/main/java/bisq/core/trade/TradeUtil.java
```diff
@@ -173,6 +174,27 @@ public String getPaymentMethodNameWithCountryCode(Trade trade) {
...
+                    trade.getShortId()));
+
+        Offer offer = trade.getOffer();
+        if (offer == null)
+            throw new IllegalStateException(format("could not get role because no offer was found for trade '%s'",
+                    trade.getShortId()));
+
+        return getRole(contract.isBuyerMakerAndSellerTaker(),
+                offer.isMyOffer(keyRing),
+                offer.getCurrencyCode());
```
sqrrm, line 195: I don't get why currenycode should be needed to decide role

proto/src/main/proto/grpc.proto
```diff
@@ -218,25 +218,26 @@ message TradeInfo {
...
+    string state = 15;
+    string phase = 16;
+    string tradePeriodState = 17;
+    bool isDepositPublished = 18;
+    bool isDepositConfirmed = 19;
+    bool isFiatSent = 20;
+    bool isFiatReceived = 21;
+    bool isPayoutPublished = 22;
+    bool isWithdrawn = 23;
+    string contractAsJson = 24;
```
sqrrm, line 240: Would be easier to just add new fields at the end, but I guess there is some system to it.

sqrrm (approved):
utACK

Some comments but nothing blocking merging, just stuff to consider.

core/src/main/java/bisq/core/trade/TradeUtil.java
ghubstan: Below is an example of the console output for a CLI's `gettrade id` command, where the role = `BTC buyer as taker`, showing what is being bought or sold.

```
ID      My Role             Price in USD for 1 BTC  Amount(BTC)  Tx Fee(BTC)  Taker Fee(BTC)  Deposit Published  Deposit Confirmed  Fiat Sent  Fiat Received  Payout Published  Withdrawn  
AGnSOw  BTC buyer as taker             13,263.6578   0.12500000   0.00068800      0.00037500  YES                NO                 NO         NO             NO                NO        
```

proto/src/main/proto/grpc.proto
ghubstan: There is a system, but don't know what it's called yet.  I'm just trying to keep field 'types' grouped together as I add more -- protocol flags together, tx-ids together, price & amt together, json-contract always the last field,  ...

I thought about backward compatibility problems being caused by doing this, but I think it's safe to assume the CLI and server will always be in sync until a single server can support multiple clients on remote hosts.

core/src/main/java/bisq/core/trade/TradeUtil.java
sqrrm: Ok, I get it, thanks.

proto/src/main/proto/grpc.proto
sqrrm: I think you should not assume that CLI and server is in sync. Basically, once we announce the API publicly the protocol is fixed and should only be added to and old fields deprecated but not changed.

ghubstan: OK, I will stop re-ordering fields like this before the public announcement.

PR #4704, opened 2020-10-25
Add 'gettrade' to api method CLI

ghubstan:
PR https://github.com/bisq-network/bisq/pull/4703 should be reviewed & merged before this PR.

PR #4705, opened 2020-10-25
Optionally print the json contract for a given trade id

ghubstan:
Add boolean 'showcontract' argument to api's 'gettrade'.

PR https://github.com/bisq-network/bisq/pull/4704 should be reviewed & merged before this PR.


sqrrm (approved):
utACK

PR #4710, opened 2020-10-26
Segwit fee estimation

oscarguindzberg:
Users reduce btc miner fees for segwit txs.

Pre-segwit miners used to evaluate tx fee and tx size. Txs were selected by their fee/byte.
Segwit brought the concept of "virtual size".
virtual size of a tx = tx serialized without witness data size + witness size (i.e. pub keys and signatures) / 4.
The virtual size of a legacy tx is the same as its size.
Segwit txs have similar sizes compared to legacy txs, but their virtual size is smaller.
New terms are now used such as vsize, vbyte, fee/vbyte, etc.
Miners now compare txs by its fee/vbyte (This is not entirely true, since they use tx weight, but we can do our math using fee/vbyte which is roughly equivalent)
See https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#additional-definitions for more info.

This PR does not include renaming size to vsize, feePerByte to feePerVbyte, etc.
I expect this PR won't be merged to master for quite some time.
The renames will introduce hundreds of changes all over the code, leading to merging hell.
I plan to do the renames either when merging date is closer or as an independent PR.

I suggest to use this link https://github.com/oscarguindzberg/bisq/compare/segwit...oscarguindzberg:fee-estimation to understand the content of this PR until https://github.com/bisq-network/bisq/pull/4612 is merged. Otherwise you will find here the content for both PRs making understanding this PR more difficult

### DIFF
--- a/core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
+++ b/core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
@@ -41,9 +41,24 @@ import static com.google.common.base.Preconditions.checkArgument;
  */
 @Slf4j
 public class TxFeeEstimationService {
-    public static int TYPICAL_TX_WITH_1_INPUT_SIZE = 260;
-    private static int DEPOSIT_TX_SIZE = 320;
-    private static int PAYOUT_TX_SIZE = 380;
+
+//  Size/vsize of typical trade txs
+//  Real txs size/vsize may vary in 1 or 2 bytes from the estimated values.
+//  Values calculated with https://gist.github.com/oscarguindzberg/3d1349cb65d9fd9af9de0feaa3fd27ac
+//  legacy fee tx with 1 input, maker/taker fee paid in btc size/vsize = 258
+//  legacy deposit tx without change size/vsize = 381
+//  legacy deposit tx with change size/vsize = 414
+//  legacy payout tx size/vsize = 337
+//  legacy delayed payout tx size/vsize = 302
+//  segwit fee tx with 1 input, maker/taker fee paid in btc vsize = 173
+//  segwit deposit tx without change vsize = 232
+//  segwit deposit tx with change vsize = 263
+//  segwit payout tx vsize = 169
+//  segwit delayed payout tx vsize = 139
+    public static int TYPICAL_TX_WITH_1_INPUT_SIZE = 175;
+    private static int DEPOSIT_TX_SIZE = 233;
+    private static int PAYOUT_TX_SIZE = 169;
+
     private static int BSQ_INPUT_INCREASE = 150;
     private static int MAX_ITERATIONS = 10;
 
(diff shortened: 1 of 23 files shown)


oscarguindzberg:
btw, the intention of this PR is not to refactor/improve the way fee calculation is done in bisq. It just pretends to take advantage of segwit fee reduction.

core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
```diff
@@ -548,6 +577,23 @@ public Transaction completePreparedBsqTx(Transaction preparedBsqTx,
         return resultTx;
     }
 
+    private Tuple2<Integer, Integer> getNumInputs(Transaction tx) {
+        int numLegacyInputs = 0;
+        int numSegwitInputs = 0;
+        for (TransactionInput input : tx.getInputs()) {
+            TransactionOutput connectedOutput = input.getConnectedOutput();
+            if (connectedOutput == null || ScriptPattern.isP2PKH(connectedOutput.getScriptPubKey()) ||
```
chimp1984, line 585: I assume the motivation for treating the connectedOutput == null as legacyInput is that if we dont know the inputs we use the potentially higher fee estimation. If correct, maybe a comment could help to make it more clear.

core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
```diff
@@ -41,9 +41,24 @@
...
+//  legacy payout tx size/vsize = 337
+//  legacy delayed payout tx size/vsize = 302
+//  segwit fee tx with 1 input, maker/taker fee paid in btc vsize = 173
+//  segwit deposit tx without change vsize = 232
+//  segwit deposit tx with change vsize = 263
+//  segwit payout tx vsize = 169
+//  segwit delayed payout tx vsize = 139
+    public static int TYPICAL_TX_WITH_1_INPUT_SIZE = 175;
+    private static int DEPOSIT_TX_SIZE = 233;
+    private static int PAYOUT_TX_SIZE = 169;
```
chimp1984: PAYOUT_TX_SIZE is not used anymore and can be deleted.

core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
```diff
@@ -87,14 +102,14 @@ public TxFeeEstimationService(FeeService feeService,
...
-        // We start with min taker fee size of 260
+        // We start with min taker fee size of 175
         int estimatedTxSize = TYPICAL_TX_WITH_1_INPUT_SIZE;
         try {
             estimatedTxSize = getEstimatedTxSize(List.of(tradeFee, amount), estimatedTxSize, txFeePerByte, btcWalletService);
         } catch (InsufficientMoneyException e) {
             if (isTaker) {
-                // if we cannot do the estimation we use the payout tx size
-                estimatedTxSize = PAYOUT_TX_SIZE;
+                // if we cannot do the estimation we use the deposit tx size
```
chimp1984: Maybe change comment to: "if we cannot do the estimation we use the largest of our txs which is the deposit tx

chimp1984 (dismissed):
utACK

core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
oscarguindzberg: Done

core/src/main/java/bisq/core/btc/TxFeeEstimationService.java
oscarguindzberg: Done

chimp1984:
> btw, the intention of this PR is not to refactor/improve the way fee calculation is done in bisq. It just pretends to take advantage of segwit fee reduction.

Is it planned to move to a bitcoinj native estimation model in a future PR?

oscarguindzberg:
> > btw, the intention of this PR is not to refactor/improve the way fee calculation is done in bisq. It just pretends to take advantage of segwit fee reduction.
> 
> Is it planned to move to a bitcoinj native estimation model in a future PR?

I have no plans to add any further code regarding fee estimation. What do you mean by "bitcoinj native estimation model"? Bisq uses bitcoinj fee estimation code and builds on top of it. Maybe you mean to use bicoinj 0.15's SendRequest.recipientsPayFees feature?. I think it would be great to rewrite bisq's fee estimation code, but I think it is out of scope for the segwit integration project.

chimp1984:
I was referring to the fee estimation you implemented for BitcoinJ. I don't know any details about it, but I thought it was planned to use that in future.

oscarguindzberg:
- WIP tag removed
- Pointed PR to release/v1.5.0

chimp1984 (approved):
utACK

PR #4709, opened 2020-10-26
Stub out api methods 'keepfunds', 'withdrawfunds'

ghubstan:
This PR adds trade closing methods to the api:
 
1.  `keepfunds trade-id`  --  keep trade funds in the Bisq wallet
2.  `withdrawfunds trade-id, address`  -- send trade funds to an external BTC wallet

Changes:
- Add grpc protos
- Add new methods to GrpcTradesService, CoreApi
- Stub out implementations in CoreTradesService
- Add methods to CLI

PR https://github.com/bisq-network/bisq/pull/4705 should be reviewed & merged before this PR.

PR #4711, opened 2020-10-26
Implement api methods 'keepfunds', 'withdrawfunds'

ghubstan:
The `CoreTradesService` was also refactored to work with the api methods implemented in this PR:

- `keepfunds` -- Close trade, keep funds in Bisq wallet

- `withdrawfunds` -- Close trade, withdraw funds to external BTC wallet

A `getKey` accessor was added to `CoreWalletsService` (needed by withdrawfunds impl).

PR https://github.com/bisq-network/bisq/pull/4709 should be reviewed & merged before this PR.


core/src/main/java/bisq/core/api/CoreTradesService.java
```diff
@@ -116,26 +137,109 @@ void confirmPaymentReceived(String tradeId) {
...
+                        new IllegalArgumentException(format("trade with id '%s' not found", tradeId))
+                ));
+    }
+
+    private Optional<Trade> getOpenTrade(String tradeId) {
+        return tradeManager.getTradeById(tradeId);
+    }
+
+    private Optional<Trade> getClosedTrade(String tradeId) {
+        return closedTradableManager.getTradableById(tradeId).map(value -> (Trade) value);
```
sqrrm, line 202: There is no guarantee that the `Tradable` is a `Trade` here, probably good to add a check for `instanceof`.

sqrrm (approved):
utACK

A comment to tend to in a later PR

The usage of the wallet and the key might have to be refactored for safer key handling, but I think that should be done as a focused refactoring.

core/src/main/java/bisq/core/api/CoreTradesService.java
ghubstan: Fixed in PR https://github.com/bisq-network/bisq/pull/4731, commit https://github.com/bisq-network/bisq/pull/4731/commits/fcdfc687e45d1d613838d7189aa4e8edd41228e2.

PR #4715, opened 2020-10-27
Cleaner formatting for trading charts date axes

deusmax:
Clean up date axes style for easy reading. Made date labels shorter,
with clear visual correspondence to relevant axes tick mark.
Tick marks with labels are now larger than those without.

Currently, the X-axis for the Price and Volume charts is not well designed. The labels are large, all the tick marks are the same. It is difficult to estimate to which tick mark the date label corresponds.  Here is a typical example for the DAY interval:
Bisq-chartaxes-day_orig-2020-10-27

The fix proposed here draws the labels more compact and the respective ticks stand-out. The chart view becomes cleaner and more visually intuitive. Above example, becomes:
Bisq-chartaxes-day_new-2020-10-27

The labels take less horizontal space by splitting into 2-levels, using 24-hour time instead of AM/PM. The examples here show English locale, though localization for the month names is maintained, as before.

Of course, further customization is possible, but the above, I feel, provides an acceptable improvement to chart display, without deviating from the current UI or introducing new features.

Similar results are obtained with the other chart interval  values. Examples (10min, HOUR, WEEK, MOTH):
Bisq-chartaxes-10min_new-2020-10-27
Bisq-chartaxes-hour_new-2020-10-27
Bisq-chartaxes-week_new-2020-10-27
Bisq-chartaxes-month_new-2020-10-27

Bisq is a unique tool. It deserves better charts.
Hope you find this a small improvement in the right direction.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/market/trades/TradesChartsView.java
+++ b/desktop/src/main/java/bisq/desktop/main/market/trades/TradesChartsView.java
@@ -479,11 +479,27 @@ public class TradesChartsView extends ActivatableViewAndModel<VBox, TradesCharts
             @Override
             public String toString(Number object) {
                 long index = MathUtils.doubleToLong((double) object);
+                // The last tick is on the chart edge, it is not well spaced with
+                // the previous tick and interferes with its label.
+                if (model.maxTicks + 1 == index) return "";
+
                 long time = model.getTimeFromTickIndex(index);
-                if (model.tickUnit.ordinal() <= TradesChartsViewModel.TickUnit.DAY.ordinal())
-                    return index % 4 == 0 ? DisplayUtils.formatDateAxis(new Date(time)) : "";
-                else
-                    return index % 3 == 0 ? DisplayUtils.formatTimeAxis(new Date(time)) : "";
+                String fmt = "";
+                switch (model.tickUnit) {
+                case YEAR  : fmt = "yyyy";
+                    break;
+                case MONTH : fmt = "MMMyy";
+                    break;
+                case WEEK  :
+                case DAY   : fmt = "dd/MMM\nyyyy";
+                    break;
+                case HOUR  :
+                case MINUTE_10: fmt = "HH:mm\ndd/MMM";
+                    break;
+                default:        // nothing here
+                }
+
+                return DisplayUtils.formatDateAxis(new Date(time), fmt);
             }
 
             @Override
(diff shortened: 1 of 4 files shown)


ripcurlx:
@deusmax I think this is a good idea! Maybe we can get it even cleaner by getting rid of the day & month when year is selected.
Bildschirmfoto 2020-10-29 um 10 41 27
And get rid of the day when month is selected.
Bildschirmfoto 2020-10-29 um 10 41 12
What do you think?

deusmax:
@ripcurlx  Thanks. Agree these are good points and indeed the charts would be even better.
These changes you propose, may require to do some minor work on the chart logic. 
The code here applies changes only the chart's element configuration/formatting. It does not make any logical changes, even minor ones. In fact I explicitly avoided doing so.

I'll start working on the labels as you point out.... and get back.
Also, the Year-interval  charts don't really need to go all the way back to 1930 !!, do they ?

deusmax:
Implemented the suggestions by @ripcurlx .

**Yearly charts**, show only year.
Bisq-chartaxes-year_new-2020-10-29

**Monthly charts,** got rid of the date and chopped the year.
Bisq-chartaxes-month_new-2020-10-29

desktop/src/main/java/bisq/desktop/main/market/trades/TradesChartsView.java
```diff
@@ -479,11 +479,27 @@ private void layoutChart() {
...
+                    break;
+                case MONTH : fmt = "MMMyy";
+                    break;
+                case WEEK  :
+                case DAY   : fmt = "dd/MMM\nyyyy";
+                    break;
+                case HOUR  :
+                case MINUTE_10: fmt = "HH:mm\ndd/MMM";
+                    break;
+                default:        // nothing here
```
ghubstan, line 503: You might want to reformat the style of the switch statement so `fmt = "..."` are on separate lines.  Not sure codacy will complain about this, but other devs might want the switch style to be consistent with the rest of the code base.
(This is the 1st time I've seen this style in Bisq;  I'm not saying there may be other examples in the code base.)

```
switch (model.tickUnit) {
                    case YEAR:
                        fmt = "yyyy";
                        break;
                    case MONTH:
                        fmt = "MMMyy";
                        break;
                    case WEEK:
                    case DAY:
                        fmt = "dd/MMM\nyyyy";
                        break;
                    case HOUR:
                    case MINUTE_10:
                        fmt = "HH:mm\ndd/MMM";
                        break;
                    default:        // nothing here
                }

```

ghubstan (commented):
I tested it, looks nice.  If I could (I'm not a JFX expert), I would center-align the labels where possible, e.g.,
```
29/Oct
 2020
```

But I would not block approval and merging without that.  
I will add a `Tested ACK` if the switch statement styling is changed.

deusmax:
@ghubstan : pushed the formatting block changes.

Still looking into centering the labels.

ghubstan:
> @deusmax: 
>
> I will add a Tested ACK if the switch statement styling is changed.

> @ghubstan : pushed the formatting block changes.
> 
> Still looking into centering the labels.


Tested ACK.

ghubstan (approved):

sqrrm (approved):

PR #4718, opened 2020-10-27
Cleanup user interface elements in Send Funds screen

### LINKED ISSUE #4656: Elongate 'withdraw to address' field to accommodate long bech32 addresses
v1.4.0 enables withdrawing to bech32 addresses, but the 'Withdraw to Address' field in the `Funds > Send Funds` screen is not wide enough to accommodate the longest address values. For example, the address shown below is 63 characters long, but only 54 characters are shown (notice how the trailing 'h' is truncated). This requires the user to horizontally scroll in order to see the entire address and visually validate its correctness:

image

Contrast this with the 'Address' field in `Funds > Receive Funds`, which is more than wide enough to accommodate any address length:

image

Recommended fix is to simply make the width of the former field the same as the latter field.

### LINKED ISSUE #4708: Send funds should default to "Amount includes mining fees"
### Description

Not a bug per se, but an improvement proposal. The "Send Funds" menu defaults to "Use all available inputs", but then to "Amount excludes mining fee", which then leads to "insufficient funds". It should default to "Amount includes mining fee"

#### Version

1.4.2

### Steps to reproduce

Open Bisq v1.4.2, go to "Funds", then "Send Funds", paste in "Withdraw to address" and press "Withdraw Selected". Withdrawal will fail due to insufficient funds.

### Expected behaviour

"Send Funds" menu defaults to  "Use all available inputs" & "Amount includes mining fee". Withdrawal goes through by just pasting destination address.

### Actual behaviour

Withdrawal fails due to "insufficient funds" error.

#### Device or machine

*nix amd64

ghost:
Selects "Amount includes mining fee" by default.  Fixes #4708
Unrestrict the length of "Withdraw to address" & "Withdrawal memo" fields.  Fixes #4656

image

Fixes #4708
Fixes #4656

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/funds/withdrawal/WithdrawalView.java
+++ b/desktop/src/main/java/bisq/desktop/main/funds/withdrawal/WithdrawalView.java
@@ -121,7 +121,7 @@ public class WithdrawalView extends ActivatableView<VBox, Void> {
     @FXML
     TableColumn<WithdrawalListItem, WithdrawalListItem> addressColumn, balanceColumn, selectColumn;
 
-    private RadioButton useAllInputsRadioButton, useCustomInputsRadioButton, feeExcludedRadioButton;
+    private RadioButton useAllInputsRadioButton, useCustomInputsRadioButton, feeExcludedRadioButton, feeIncludedRadioButton;
     private Label amountLabel;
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
utACK -  code is looking fine :+1:

PR #4716, opened 2020-10-27
Test api method 'canceloffer'

ghubstan:
Add method test.

PR https://github.com/bisq-network/bisq/pull/4714 should be reviewed & merged before this PR.

PR #4722, opened 2020-10-28
Upgrade jupiterVersion = '5.7.0'

ghubstan:
This upgrades jupiter for all subprojects using this library, but  it does not affect released code.

PR https://github.com/bisq-network/bisq/pull/4716 should be reviewed & merged before this PR.


sqrrm (approved):
utACK

This is for testing which makes it a less risky upgrade. I saw you got an ack from Chimp as well, good practice to ask around before upgrading libraries, thanks.

PR #4723, opened 2020-10-28
Improve apitest cmd line console logging

ghubstan:
- Show full stack traces in console
- Do use previously cached test outputs
- Do not show skipped test name in console (too noisy)
- Show test run summary, including number of skipped tests
- Show note about skipped test info in the html report
- Show link to html report on test SUCCESS

PR https://github.com/bisq-network/bisq/pull/4722 should be reviewed & merged before this PR.

PR #4724, opened 2020-10-28
Refactor api test fixture setup

ghubstan:
- Remove dead code from AbstractLinuxProcess.

- Make default dummy accts static in MethodTest.

- Simplify gRPC stub creation, btc block generation, dispute agent
  registration, and dummy acct initialization in test case startup.
  
- Make ExpectedProtocolStatus visible to scenario test pkg.

- Remove redundant ScenarioTest superclass.

PR https://github.com/bisq-network/bisq/pull/4723 should be reviewed & merged before this PR.


sqrrm (approved):
utACK

PR #4733, opened 2020-10-31
Allow user-defined block explorers

### LINKED ISSUE #4371: Support setting one or more user-defined block explorers

ghost:
In Settings/General Preferences the Bitcoin Explorer and Bisq Explorer may now be user defined in addition to the hard coded block explorer choices.

This implementation takes advantage of the existing Protobuf implementation that stores a BlockExplorer Name/UrlTx/UrlAddress for BTC and for BSQ.

A new popup dialog is implemented which allows editing of the explorer details (name, txUrl, and addressUrl).  Any of the system-defined explorer details may be selected into the edit fields thus allowing them to be used as templates, customized or an entirely new explorer to be used.

If the user switches choice from a customized back to a pre-defined explorer then the customized settings are gone; this is because the Protobuf storage only allows saving one selection.

Fixes #4371
Closes #4634

image

image

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/overlays/windows/EditCustomExplorerWindow.java
+++ b/desktop/src/main/java/bisq/desktop/main/overlays/windows/EditCustomExplorerWindow.java
@@ -1,0 +1,199 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.desktop.main.overlays.windows;
+
+import bisq.desktop.components.AutoTooltipButton;
+import bisq.desktop.components.AutoTooltipLabel;
+import bisq.desktop.components.InputTextField;
+import bisq.desktop.main.overlays.Overlay;
+import bisq.desktop.util.Layout;
+import bisq.desktop.util.validation.LengthValidator;
+
+import bisq.core.locale.Res;
+import bisq.core.util.validation.UrlInputValidator;
+
+import bisq.common.util.Tuple2;
+
+import bisq.core.user.BlockChainExplorer;
+
+import javafx.scene.control.Button;
+import javafx.scene.control.Label;
+import javafx.scene.control.ListCell;
+import javafx.scene.control.ListView;
+import javafx.scene.input.MouseEvent;
+import javafx.scene.layout.AnchorPane;
+import javafx.scene.layout.ColumnConstraints;
+import javafx.scene.layout.GridPane;
+import javafx.scene.layout.HBox;
+import javafx.scene.layout.VBox;
+
+import javafx.geometry.Insets;
+import javafx.geometry.Pos;
+import javafx.geometry.VPos;
+
+import javafx.event.EventHandler;
+
+import javafx.collections.FXCollections;
+
+import javafx.util.Callback;
+
+import java.util.ArrayList;
+
+import static bisq.desktop.util.FormBuilder.*;
+import static javafx.beans.binding.Bindings.createBooleanBinding;
+
+public class EditCustomExplorerWindow extends Overlay<EditCustomExplorerWindow> {
+
+    private InputTextField nameInputTextField, txUrlInputTextField, addressUrlInputTextField;
+    private UrlInputValidator urlInputValidator;
+    private BlockChainExplorer currentExplorer;
+    private ListView<BlockChainExplorer> listView;
+
+    public EditCustomExplorerWindow(String coin,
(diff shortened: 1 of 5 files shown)


ghost:
Fixes #4634

cd2357:
ACK, tested and works as expected.

desktop/src/main/java/bisq/desktop/main/settings/preferences/PreferencesView.java
```diff
@@ -937,6 +906,34 @@ private void activateDisplayPreferences() {
 
         resetDontShowAgainButton.setOnAction(e -> preferences.resetDontShowAgain());
 
+        editCustomBtcExplorer.setOnAction(e -> {
+            EditCustomExplorerWindow urlWindow = new EditCustomExplorerWindow("BTC",
+                    preferences.getBlockChainExplorer(), preferences.getBlockChainExplorers());
+            urlWindow
+                    .actionButtonText("Save")
```
chimp1984: I think you forgot here to use a resource string. I guess its: `Res.get("shared.save")`

desktop/src/main/java/bisq/desktop/main/settings/preferences/PreferencesView.java
```diff
@@ -937,6 +906,34 @@ private void activateDisplayPreferences() {
...
+                    .closeButtonText(Res.get("shared.cancel"))
+                    .onClose(urlWindow::hide)
+                    .show();
+        });
+
+        editCustomBsqExplorer.setOnAction(e -> {
+            EditCustomExplorerWindow urlWindow = new EditCustomExplorerWindow("BSQ",
+                    preferences.getBsqBlockChainExplorer(), preferences.getBsqBlockChainExplorers());
+            urlWindow
+                    .actionButtonText("Save")
```
chimp1984: I think you forgot here to use a resource string. I guess its: Res.get("shared.save")

chimp1984 (dismissed):
utACK

Beside 2 nits... See comments

chimp1984 (dismissed):
utACK

desktop/src/main/java/bisq/desktop/main/overlays/windows/EditCustomExplorerWindow.java
```diff
@@ -0,0 +1,198 @@
...
+
+        Button button = new Button(">>");
+        button.setOnAction(e -> {
+            BlockChainExplorer blockChainExplorer = listView.getSelectionModel().getSelectedItem();
+            if (blockChainExplorer != null) {
+                nameInputTextField.setText(blockChainExplorer.name);
+                txUrlInputTextField.setText(blockChainExplorer.txUrl);
+                addressUrlInputTextField.setText(blockChainExplorer.addressUrl);
+            }
+        });
```
ripcurlx: Please use our AutoTooltipButton or JFXButton to match the design of all other buttons.

ripcurlx (changes requested):
NACK

Code looks fine, but please change the style of the `>>` button to match our general button style.

desktop/src/main/java/bisq/desktop/main/overlays/windows/EditCustomExplorerWindow.java
ghost: :+1: Done!

sqrrm:
@ripcurlx Are you ok with this PR now? Will leave merge for you

ripcurlx (approved):

cbeams:
Great to see this functionality, thanks @jmacxx!

One note for potential future contributions: it would be great to be able to have more than one custom explorer. This is from my original description of the feature at https://github.com/bisq-network/bisq/issues/4371#issue-664425011:

> While it might be easiest to support adding just one custom block explorer, it would be ideal to allow specifying multiple of them, because in the case of an explorer running on one's local network, one might wish to access it via clearnet http when at home and via Tor when remote. This would require specifying two sets of URLs. And of course the user might want to specify any number of additional custom explorer services.

PR #4736, opened 2020-11-01
Admin filter for auto-confirmation service addresses

### LINKED ISSUE #4683: Add field for Monero Explorer nodes to the Bisq filter

ghost:
A service address can be filtered out using the admin key.  This would be useful if a federated server failed, it could be filtered out so that the auto-confirmation operates with the remaining services.  If all federated service addresses were filtered out the user would have to either confirm manually or use their own explorer service.

example:
filter

Fixes #4683

### DIFF
--- a/core/src/main/java/bisq/core/filter/FilterManager.java
+++ b/core/src/main/java/bisq/core/filter/FilterManager.java
@@ -398,6 +398,12 @@ public class FilterManager {
                         .anyMatch(e -> e.equals(nodeAddress.getFullAddress()));
     }
 
+    public boolean isAutoConfExplorerBanned(String address) {
+        return getFilter() != null &&
+                getFilter().getBannedAutoConfExplorers().stream()
+                        .anyMatch(e -> e.equals(address));
+    }
+
     public boolean requireUpdateToNewVersionForTrading() {
         if (getFilter() == null) {
             return false;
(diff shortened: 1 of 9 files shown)


core/src/main/java/bisq/core/filter/FilterManager.java
```diff
@@ -398,6 +398,12 @@ public boolean isNodeAddressBanned(NodeAddress nodeAddress) {
                         .anyMatch(e -> e.equals(nodeAddress.getFullAddress()));
     }
 
+    public boolean isAutoConfExplorerBanned(String address) {
+        return getDevFilter() != null &&
```
chimp1984: Here we need to use `getFilter` not `getDevFilter`. DevFilter is the locally stored of the filter admin. Not great naming... should be changed to a more clear name...

chimp1984 (changes requested):

core/src/main/java/bisq/core/filter/FilterManager.java
ghost: Argh... thanks for finding that, switched to devFilter for testing: looks like I forgot to revert.

chimp1984 (approved):
utACK

ripcurlx (approved):
utACK - Code looks fine.

PR #4739, opened 2020-11-02
Hide proposal delete button

chimp1984:
We have a bug in removing a proposal which is not trivial to fix
(p2p network layer). Until that bug is fixed
it is better to not show the remove button as it
confused users and a removed proposal will reappear with a
high probability at the vote phase.


ripcurlx (approved):
utACK

PR #4078, opened 2020-03-19
Add new payment method: Australian PayID

mercurytoxic:
<h1>Australia PayID 🇦🇺</h1>

<h2>Summary</h2>
This PR adds support for Bisq trades using the Australian New Payments Platform (NPP)[1], known casually by everyone in Australia as "PayID".
<p>

- Country Restrictions: Australia only
- Currency Restrictions: AUD only
- Chargeback Risks: None, cannot reverse or chargeback
- Per-Trade Risk Limit: LOW (0.5 BTC * 2.0 multiplier = 1 BTC)
- Per-Trade Time Limit: DAY (transfers are near-instant, and 24/7/365 for major banks)

### Overview

- _A PayID is something easy to remember like a phone number, email address or an Australian Business Number (ABN), that you can securely link to your bank, credit union or building society account. Then when you want somebody to transfer funds into your account, you can give them your PayID instead of your BSB and account number. To pay another person’s PayID you just need to log into your participating financial institutions internet or mobile banking._[2]
- Transfers using PayID are usually near instant, compared to National Bank Transfers that in Australia usually take 1-2 business days for the deposits to clear. (Depends on individual bank policies, frequently first time transfers above a certain amount might have a 24 hour hold)

- A great number of [Australian banks support PayID](https://nppa.com.au/find-an-institution/)

### Security

- Payments with PayID using NPP are made by logging into your usual online banking and authorizing a payment.
- When sending a payment using PayID you are able to check the name of the recipient before confirming the payment.
- Bank transfers are not reversible or chargeback-able

### Privacy

- Only PayID and the recepient's name are shared.

<h2>How it works</h2>

- PayID is free to use.
- Transfers usually instant, but depending on the financial institutions that are used the first time a payment is made to a new PayID, the payment might be held for 24 hours.
- Available 24/7/365 since payments are electronic.

### Account Information
To send a PayID transfer, the sender needs the following information:

- PayID
- Optional is the recepient's name to verify is the correct PayID account

### Sending a Transfer
After the sender inputs the PayID in their online banking, the receiving account holder's name is displayed to the sender so they can verify they are sending funds to the correct account before they finalize the transaction.

The transfer is usually completed instantly, or on the next business day if it's the first time sending to a new PayID  or one of the participating banks does not fully support PayID and depends on fallback BSB (routing) and account numbers.

<h2>Fraud Risks</h2>

[...]

PR #4743, opened 2020-11-03
Add new payment method: Australian PayID

ripcurlx:
See #4078 for details.

I had to create my own branch as I wasn't able to push to the PR and I had to cherry-pick the original commit as it wasn't signed by @mercurytoxic.

Tested it on Regtest and works still with the old test trade on Mainnet.


sqrrm (approved):
utACK

core/src/main/java/bisq/core/payment/AustraliaPayid.java
```diff
@@ -17,53 +17,37 @@
...
-    public String getPayID()
-    {
-        return ((AustraliaPayIDPayload) paymentAccountPayload).getPayid();
+    public String getPayid() {
+        return ((AustraliaPayidPayload) paymentAccountPayload).getPayid();
     }
-    public void setPayID(String payid)
-    {
+
+    public void setPayid(String payid) {
```
chimp1984, line 40: Our naming convention would be: `setPayId`  and `payId`
Would require also changes in the protobuf

chimp1984:
Please add the an "AUSTRALIA_PAYID" entry in TradeStatistics3.PaymentMethodMapper enum

chimp1984 (changes requested):

core/src/main/java/bisq/core/payment/AustraliaPayid.java
sqrrm: I think payid is one word in this case. It's not an id as part of bisq, just the name of the payment method.

chimp1984: True. They use PayID... anyway not sure whats best here, but also not really important...

ripcurlx:
> Please add the an "AUSTRALIA_PAYID" entry in TradeStatistics3.PaymentMethodMapper enum

It was already added by you as there was a live trade on Mainnet by the dev 😉

core/src/main/java/bisq/core/payment/AustraliaPayid.java
ripcurlx: I had it like that before, but I agree with @sqrrm that it is actually a brand name and should be without camelcase in that case.

chimp1984:
> > Please add the an "AUSTRALIA_PAYID" entry in TradeStatistics3.PaymentMethodMapper enum
> 
> It was already added by you as there was a live trade on Mainnet by the dev 😉

Ah, true ;-)

PR #4744, opened 2020-11-03
Cleanup signed witness

sqrrm:
Implements https://github.com/bisq-network/proposals/issues/264

Payment methods using only accountagewitness can trade the full amount immediately, both when buying and when selling.

Payments methods with signing can trade the full amount immediately when selling, but has limits for buying according to the schedule described:
- 0.01 BTC for unsigned and up to 30 days after signing
- 0.125 BTC 30 days after signing
- 0.25 BTC 60 days after signing

### DIFF
--- a/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
+++ b/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
@@ -379,7 +379,7 @@ public class AccountAgeWitnessService {
 
         long limit = OfferRestrictions.TOLERATED_SMALL_TRADE_AMOUNT.value;
         var factor = PaymentMethod.hasChargebackRisk(paymentMethod, currencyCode) ?
-                signedTypeFactor(accountAgeCategory, direction) : normalFactor(accountAgeCategory);
+                signedTypeFactor(accountAgeCategory, direction) : normalFactor();
         if (factor > 0) {
             limit = MathUtils.roundDoubleToLong((double) maxTradeLimit.value * factor);
         }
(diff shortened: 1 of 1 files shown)


core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
```diff
@@ -365,58 +365,70 @@ private long getTradeLimit(Coin maxTradeLimit,
...
-            return limit;
+        double factor;
+        boolean isRisky = PaymentMethod.hasChargebackRisk(paymentMethod, currencyCode);
+        if (!isRisky || direction == OfferPayload.Direction.SELL) {
+            // Get age of witness rather than time since signing for non risky payment methods and for selling
+            accountAgeCategory = getAccountAgeCategory(getAccountAge(accountAgeWitness, new Date()));
+        }
+        long limit = OfferRestrictions.TOLERATED_SMALL_TRADE_AMOUNT.value;
+        if (direction == OfferPayload.Direction.BUY && isRisky) {
+            // Used only for bying of BTC with risky payment methods
```
ripcurlx: Nit: Just a little typo in the old code `buying` vs `bying`

ripcurlx (approved):
utACK

Code changes are looking fine and reflect the consensus found in https://github.com/bisq-network/proposals/issues/264.

@m52go This will need a couple of changes (will make it easier for users) in https://bisq.wiki/Account_limits when v1.5.0 gets released.

ripcurlx:
We definitely have to keep an eye on all popups during trade and account creation to see if the new limits are communicated correctly everywhere. I guess a couple of changes will be required here and there.

PR #4706, opened 2020-10-25
Remove outliers when calculating BSQ rate

chimp1984:
Remove outliers when calculating BSQ rate.
Add percentage for upper and lower threshold for outlier detection.
It filters outliers in the BSQ and in the USD market.

Settings:
settings

As one can see below even a small % threshold removes the few outliers we had recently: 5 % is default value.

With 0%:
0perc

With 1%:
1perc

With 5%:
5perc

With 10%:
10perc

With 20%:
20perc

With 49% (max):
Screen Shot 2020-10-25 at 18 39 56

PR #4745, opened 2020-11-03
Fix bug with reading historical data

chimp1984:
Fix bugs with reading historical data:
- Use fileName not getFileName() in readHistoricalStoreFromResources
- use complete handler once reading of all historical data is completed where we
   build the ImmutableMaps and complete the readFromResources method

Fixes https://github.com/bisq-network/bisq/pull/4706#issuecomment-721156401


ripcurlx (approved):
utACK

sqrrm (commented):
utACK

cbeams:
ACK (after the fact of merging), as I've tested this locally, and the problem I reported in https://github.com/bisq-network/bisq/pull/4706#issuecomment-721156401 is indeed resolved. There is however another problem still persisting, and I'll comment about that back on the #4706 issue.

PR #4746, opened 2020-11-04
Clean up mailbox messages for closed trades

chimp1984:
We got growing pending mailbox messages since 1.4.2. (about 2000 now, normal is about 500). With the trade protocol refactoring a cleanup for mailbox messages has been missed.

Steps to reproduce issue:
Start a trade and let peer go offline. When Seller confirms receipt he sends a msg to buyer, as he is offline its a mailbox msg. Seller moves funds to his wallet and his trade is closed. Buyer goes online and closes his trade as well. He sends back a AckMessage that he received the trade message but as sellers trade is closed that ackMsg is not being processed (done in tradeProtocol class which is not initialized for closed trades).

Solution:
We iterate closed and failed trades and look up if there are some pending mailbox messages and if we find any we remove them from the network. 
It can be that there are failure scenarios where mailbox messages did not got removed as well. This should fix those cases as well.


sqrrm (approved):
utACK

PR #4751, opened 2020-11-05
Reduce seednode default maxconnections to 20

wiz:
@chimp1984


chimp1984 (approved):
utACK

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/4751#pullrequestreview-524462626

PR #4754, opened 2020-11-05
Complete Segwit fee estimation

oscarguindzberg:
This is something I forgot to add in https://github.com/bisq-network/bisq/pull/4710
This is the commit that was supposed to include the changes https://github.com/bisq-network/bisq/commit/29f23fe50c6727f8784f3af7c8de3d988ba517ac


chimp1984 (approved):
utACK

PR #4756, opened 2020-11-06
Remove HitBTC as a pricenode data provider

wiz:
Recently, some Monero traders were complaining the XMR/BTC price on the Bisq Price Index was off from most exchanges. Indeed, it seems HitBTC is trading at a -10% divergence for some reason, I guess they have low liquidity or some other reason, but after taking a look we are currently using HitBTC for the following assets:

* AEON - only 2 trades ever, last traded March 2019
* EMC - never traded
* GRIN - looks like only fake trades, last traded Jan 2019
* PART - only 7 trades ever, last traded August 2020
* XRC - last traded December 2019
* XMR - obviously very important for Bisq

So I feel it's worth it to drop HitBTC as a data provider from Bisq to
make the Monero traders happy and make the Monero price more accurate.


ripcurlx:
@wiz You forgot to remove the test cases as well. See Travis results.

wiz:
I guess that's what I get for building with `./gradlew build -x test` all the time :joy:

ghubstan (commented):
utACK

ghubstan (commented):
utACK

These two utACKs are for each commit, as the review button was on those commits, not this PR.

cd2357:
Good idea to remove HitBTC

https://twitter.com/Raindropactual/status/1208241130342629376

sqrrm (approved):

PR #4740, opened 2020-11-02
Center align tick marks text for trading charts

deusmax:
New method addTickMarkLabelCssClass to grab the axis tick label and add a CSS class to it.

Signed-off-by: Deus Max <deusmax@gmx.com>
Acked-by: Deus Max <deusmax@gmx.com>

Your PR description here.
@ghubstan  as discussed in PR#4175. to center-align tick mark text.
Add a CSS class to the axis tick marks text. The CSS class can be used to add any css command, such as `-fx-text-align: center` to the labels.
Here is a screenshot of the results:
Bisq-chartaxes-day_centeralign-2020-11-02

PR #4761, opened 2020-11-07
Apply format to trading charts axis Y values

deusmax:
This PR completes the improvements to trading charts introduced in PRs #4715 and #4740, by formatting the Y-axis values of the **Price** chart as:
 - Fiat: Apply separator format to numbers
 - Crypto: remove some (not all) trailing zeroes  from fractional part

Here is the result for the EUR chart, similar for other fiat:
Bisq-charts-Yaxis_EUR-day-2020-11-04

Below are before and after for various cryptos:
XMR before:
Bisq-charts-Yaxis_XMR0-day-2020-11-05
XMR after:
Bisq-charts-Yaxis_XMR-day-2020-11-05
BSQ before:
Bisq-charts-Yaxis_BSQ0-day-2020-11-05
BSQ after
Bisq-charts-Yaxis_BSQ-day-2020-11-05
ETH before:
Bisq-charts-Yaxis_ETH0-week-2020-11-05
ETH after:
Bisq-charts-Yaxis_ETH-week-2020-11-05

What do you think ?

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/market/trades/TradesChartsView.java
+++ b/desktop/src/main/java/bisq/desktop/main/market/trades/TradesChartsView.java
@@ -89,6 +89,8 @@ import javafx.collections.transformation.SortedList;
 import javafx.util.Callback;
 import javafx.util.StringConverter;
 
+import java.text.DecimalFormat;
+
 import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


ghubstan (commented):
utACK

ghubstan (commented):
Are you sure the first three zeros being replaced will always be the three trailing zeros?

deusmax:
@ghubstan 
Yes. A RegEx is used that anchors at the trailing end of the string, so only the last zeroes are matched. It can not have multiple matches.
The use of either function replaceFirst or replaceAll does not matter.

ghubstan (approved):
> Yes. A RegEx is used that anchors at the trailing end of the string, so only the last zeroes are matched. It can no have multiple mathches.
> The use of either function replaceFirst or replaceAll does not matter.

Missed that... thanks.

utACK

ripcurlx (approved):
utACK

PR #4749, opened 2020-11-05
Various improvements in GetBlocks handling

chimp1984:

PR #4764, opened 2020-11-08
Don't allow trade start if BitcoinJ is not fully synced

### LINKED ISSUE #4707: Take Offer Error: LockTime of Maker is more than 3 blocks different to the LockTime I have
#### Version
1.3.9 and 1.4.2

### Actual behaviour

When taking an offer for 0.01 BTC with EUR/Revolut with trade ID izxdhj-877b1ad8-1aa3-4fa1-b471-30758e64ba71-139 I get:
Take Offer Error: LockTime of Maker is more than 3 blocks different to the LockTime I calculated.
Screenshot below. 
Tried several times. The same happens when I perform a local take offer in national currency THB and national bank transfer.

### Screenshots

image

image

image

### LINKED ISSUE #4727: Don't allow trade start if BitcoinJ is not fully synced
### Description

While BitcoinJ is syncing the lock time would be incorrectly calculated leading to a failed trade. Currently users could take and offer or offers of a maker can be taken during sync.  

We should add for both maker and taker side the check that BitcoinJ need to be synced. The sync of a new block (veyry short time) can be ignored as the locktime has some tolerance built it to not fail in case both traders have a small difference of the block height.

Case of a user who had that problem:
https://github.com/bisq-network/bisq/issues/4707#issuecomment-718417697

ghost:
Adds a check that if chain height is more than 3 blocks of the reported height from bitcoin peers -> it is considered a serious error and the user is prevented from taking or creating an offer, also any open orders are prevented from being taken as it would cause a protocol error.  

It displays a popup message informing the user that Bisq is not currently synced, advising them to wait for one Bitcoin block and if necessary to do an SPV resync.

Additionally under Settings/Network, a field has been added to show the chain height of Bisq vs the Peer group.

Fixes: #4727 
Fixes: #4707

### Screenshots

image

Network info showing chain height:
image

Logged upon attempting to take or create offer when not synced:
image


ghost:
@chimp1984 I wanted to get your ideas on this.  Is it too restrictive?  

Here are the places where the new restriction would have an impact:
image

chimp1984:
@jmacxx Thanks for picking that up!

Have you seen the recent changes I did for the footer to show the latest BTC block? Not sure if that can help for the popup msg... I think its merged in master...

The Network info label for chain height might be too technical (Peer group). 

I think it should be enough to apply it only to the trade use case. Until now we never had it and only with the introduction of the locktime we got an issue with not being on the latest block.

ghost:
Ahh, I had not seen your PR for putting chainHeight in the footer, it is not yet merged.  What I've noticed is that chainHeight property is 0 until the first new block comes in after startup which can be a 10, 20, or 30 minutes wait.  Perhaps it should be initialized; `getChain().getBestChainHeight()` returns the current block number.

I'll remove the label in network settings and rework this PR as you suggested, thanks.

chimp1984:
> What I've noticed is that chainHeight property is 0 until the first new block comes in after startup which can be a 10, 20, or 30 minutes wait. Perhaps it should be initialized; `getChain().getBestChainHeight()` returns the current block number.

Oh, yes that should be fixed. Feel free to add a fix for that.

chimp1984:
> I'll remove the label in network settings

Maybe its good to have it there, just a bit less technical label... something like "Latest BTC block height"

chimp1984:
>  I think its merged in master...

No, it is part of #4749
 https://github.com/bisq-network/bisq/pull/4749/commits/9bfdea123db12891125cb9f3cb9be56a6a57c8ee

ghost:
I've made the changes above - in addition any open orders will also be prevented from being taken in the error case where chain is out of sync.

core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
```diff
@@ -522,6 +523,16 @@ public boolean isDownloadComplete() {
         return downloadPercentageProperty().get() == 1d;
     }
 
+    public boolean isChainHeightSyncedToPeers() {
```
chimp1984: Maybe better rename it so it reflects that we have a tolernace for the check... 
Maybe: `inSyncWithMostCommonChainHeight` ? (still not great but the "MostCommon" suggests we are not operating on exact data here.

chimp1984 (dismissed):
utACK

core/src/main/java/bisq/core/btc/setup/WalletsSetup.java
ghost: Done!  renamed to `isChainHeightSyncedWithinTolerance`

chimp1984 (approved):
utACK

ripcurlx (approved):
utACK

PR #4767, opened 2020-11-08
Remove default fee rate from Bisq startup screen

### LINKED ISSUE #4765: Bisq startup screen shows incorrect mempool fee rate
Since it hasn't updated any data yet, it's just displaying old data @jmacxx 

スクリーンショット 2020-11-08 21 03 37
スクリーンショット 2020-11-08 21 02 25

ghost:
The display of fee rate on the startup screen is misleading because at startup a fee has not been received from the mempool fee service and instead the default value of 50 sats is shown.

This change causes the fee rate not to be displayed until a response has been received from the service.

Fixes #4765

 
Bisq loading screen, fee no longer displayed:
image

 
Footer of Bisq main screen:
image


ripcurlx (approved):
utACK

PR #4768, opened 2020-11-08
Add new burningman address for delayed payout txs

chimp1984:
See: https://github.com/bisq-network/roles/issues/80#issuecomment-723577776

@ripcurlx Would be good to get that into the release branch.


ripcurlx:
@burningman3 Is there a reason why there are already transactions going to these addresses? 2 txs to the new payouts address and 26 to the new fee address?

ripcurlx:
> @burningman3 Is there a reason why there are already transactions going to these addresses? 2 txs to the new payouts address and 26 to the new fee address?

Or are they configured in the filter already?

ripcurlx (approved):
ACK

Verified addresses and fee address was successfully added to the filter.

PR #4769, opened 2020-11-09
Update release process to match new trade statistics delivery

ripcurlx:
Also did some minor release process cleanup that was overdue

### DIFF
--- a/desktop/package/macosx/copy_dbs.sh
+++ b/desktop/package/macosx/copy_dbs.sh
@@ -2,12 +2,16 @@
 
 cd $(dirname $0)/../../../
 
+version="1.5.0"
+
 # Set BISQ_DIR as environment var to the path of your locally synced Bisq data directory e.g. BISQ_DIR=~/Library/Application\ Support/Bisq
 
 dbDir=$BISQ_DIR/btc_mainnet/db
 resDir=p2p/src/main/resources
 
-cp "$dbDir/TradeStatistics2Store" "$resDir/TradeStatistics2Store_BTC_MAINNET"
+# Only commit new TradeStatistics3Store if you plan to add it to
+# https://github.com/bisq-network/bisq/blob/0345c795e2c227d827a1f239a323dda1250f4e69/common/src/main/java/bisq/common/app/Version.java#L40 as well.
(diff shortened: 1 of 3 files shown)


ghubstan (changes requested):
It's not due to this commit, but a link in release-process.md is broken.
Under [Build release](https://github.com/bisq-network/bisq/blob/a95630754bb0b4aaaf3390994883ca1bfef0a308/docs/release-process.md#build-release) -> 
`Make sure all version numbers are updated (update version variables and run replace_version_number.sh).`, the link to `replace_version_number.sh` is broken.

ghubstan (changes requested):
Under [Sign release on macOS](https://github.com/bisq-network/bisq/blob/a95630754bb0b4aaaf3390994883ca1bfef0a308/docs/release-process.md#sign-release-on-macos), the link to `finalize.sh` is also broken.

desktop/package/macosx/copy_dbs.sh
```diff
@@ -2,12 +2,14 @@
...
 
+version="1.5.0"
+
 # Set BISQ_DIR as environment var to the path of your locally synced Bisq data directory e.g. BISQ_DIR=~/Library/Application\ Support/Bisq
 
 dbDir=$BISQ_DIR/btc_mainnet/db
 resDir=p2p/src/main/resources
 
-cp "$dbDir/TradeStatistics2Store" "$resDir/TradeStatistics2Store_BTC_MAINNET"
+cp "$dbDir/TradeStatistics3Store" "$resDir/TradeStatistics3Store_${version}_BTC_MAINNET"
```
chimp1984, line 14: We do not update with each version (pre releases/hotfixes) the TradeStatistics3Store....

chimp1984 (dismissed):
utACK

desktop/package/macosx/copy_dbs.sh
ripcurlx: I'm aware of it. I just wanted to add it to the script, so I don't have to copy it manually. I'll just not commit it if there is no intention to add a new trade stats store version to the release.

ripcurlx: I've added an additional comment to the script to make it more clear.

sqrrm (approved):

PR #4771, opened 2020-11-09
Fix BTC address tbl output formatting

ghubstan:
- Adjusts the CLI's output formatting for segwit address lengths.

- Anticipates possibility this formatting method will be used  to display other types of crypto addresses.


ghubstan:
Will be included in later PR.

PR #4772, opened 2020-11-09
Stub out new api method 'getunusedbsqaddress'

ghubstan:
This api method will return one unused BSQ address.  It will take no arguments, and return a String.


ghubstan:
The codacy complaint about unused field `private final BsqWalletService bsqWalletService` is fixed in PR https://github.com/bisq-network/bisq/pull/4773, where the method stubbed out here is implemented, and field is used.  Hopefully this codacy complaint will not block merge.

ghubstan:
Will be included in later PR.

PR #4773, opened 2020-11-09
Implement and test api method 'getunusedbsqaddress'

ghubstan:
This method returns one unused BSQ address.  It takes no arguments and returns a String.

PR https://github.com/bisq-network/bisq/pull/4772 should be reviewd & merged before this PR.


ghubstan:
Will be included in later PR.

PR #4774, opened 2020-11-09
Market Offers chart modifications

deusmax:
## Intro
This is an attempt to simplify the Market Offers chart, to provide a better  (I think) visual "feeling". 
In particular, the scientific precision of the X-axis values is confusing to the user: 

For example, for EUR this:
Bisq-MarketOfferBook-EURold-2020-11-09
becomes:
Bisq-MarketOfferBook-EURnew-2020-11-09

For XMR:
Bisq-MarketOfferBook-default-XMR-2020-11-08
becomes:
Bisq-MarketOfferBook-new-XMR-2020-11-08

For GBP:
Bisq-MarketOfferBook-GBPold-2020-11-09
becomes:
Bisq-MarketOfferBook-GBPnew-2020-11-09

## The BSQ problem
The above works well (my opinion of course) for the remaining currencies (including USD, which I forgot to capture a pic - sorry), except for BSQ.
BSQ offers have some crazy values. The current charts implementation, filters-out values that have a factor 3 difference, the outliers are removed.
In the current PR, the filtering-out has been removed. I believe if there is a problem, the data should show it, not hide it. Hiding bad data creates a false sense of OKness. Better to see it:
Current BSQ chart:
Bisq-MarketOfferBook-BSQold-2020-11-09
becomes:
Bisq-MarketOfferBook-BSQnew-2020-11-09

Looking at the Sell BTC (red) table, we see the last few values represent orders of magnitude jumps from the main core.
Putting back the filtering is easy, as is taking it off. I hope a discussion on how to best handle this issue follows. Is there one ongoing, already ?

Perhaps an outliers-toggle switch like in that other DAO chart on BSQs.

## The translations
The Y-axis label is "BTC Offer Volume". As a multi-lingual configuration is used on this chart, a translation web service was used to provide stock translations, which probably sometimes fail miserably. I don't know the Bisq policy on this, please advice.

### DIFF
--- a/core/src/main/java/bisq/core/locale/CurrencyUtil.java
+++ b/core/src/main/java/bisq/core/locale/CurrencyUtil.java
@@ -543,4 +543,8 @@ public class CurrencyUtil {
         else
             return Res.get(translationKey, currencyCode, Res.getBaseCurrencyCode());
     }
+
+    public static String getOfferVolumeCode(String currencyCode) {
+        return Res.get("shared.offerVolumeCode", currencyCode);
+    }
 }
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -52,6 +52,7 @@ shared.P2P=P2P
 shared.oneOffer=offer
 shared.multipleOffers=offers
 shared.Offer=Offer
+shared.offerVolumeCode={0} Offer Volume
 shared.openOffers=open offers
 shared.trade=trade
 shared.trades=trades
--- a/desktop/src/main/java/bisq/desktop/bisq.css
+++ b/desktop/src/main/java/bisq/desktop/bisq.css
@@ -1751,6 +1751,10 @@ textfield */
     -fx-alignment: center-left;
 }
 
+#charts .axisy .axis-label {
+    -fx-alignment: center;
+}
+
 /********************************************************************************************************************
  *                                                                                                                  *
  * Highlight buttons                                                                                                  *
--- a/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java
+++ b/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java
@@ -88,6 +88,8 @@ import javafx.collections.FXCollections;
 import javafx.collections.ListChangeListener;
 import javafx.collections.ObservableList;
 
+import java.text.DecimalFormat;
+
 import javafx.util.Callback;
 import javafx.util.StringConverter;
 
@@ -95,6 +97,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import static bisq.desktop.util.FormBuilder.addTopLabelAutocompleteComboBox;
 import static bisq.desktop.util.Layout.INITIAL_WINDOW_HEIGHT;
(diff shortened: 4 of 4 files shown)


chimp1984:
Thanks for those improvements!

> In the current PR, the filtering-out has been removed. I believe if there is a problem, the data should show it, not hide it. Hiding bad data creates a false sense of OKness.

It is not bad data it is just people putting up offer far away from market price. In other currencies where we get an external price feed from other exchanges we have a limit of the max. deviation (I guess 50%) but for those currencies where no external price is available (as BSQ) that is not the case. I think thats the reason why we see in BSQ such outliers and in other markets not (would be probably the case as well if we would not have that limit). I guess ppl set up such offer in the hope that someone make a mistake at some point and take it.... 

So to make the chart usable, I think we have to add the filter again to hide the outliers. I think a toggle is not needed as the display of the outliers included chart is pretty useless. But no strong opinion on that.

core/src/main/resources/i18n/displayStrings_de.properties
```diff
@@ -52,6 +52,7 @@ shared.P2P=P2P
 shared.oneOffer=Angebot
 shared.multipleOffers=Angebote
 shared.Offer=Angebot
+shared.offerVolumeCode={0} Angebotsvolumen
```
chimp1984: Any edit of the other language files will be reverted once updated from translators. For devs displayStrings.properties is the oinly file to edit, the rest is done via Transifex.

chimp1984:
> The Y-axis label is "BTC Offer Volume". As a multi-lingual configuration is used on this chart, a translation web service was used to provide stock translations, which probably sometimes fail miserably. I don't know the Bisq policy on this, please advice.

We have translators who work with Transifex. Devs only edit the displayStrings.properties, changes in the other files will have not effect as they get reverted. If a string is not traanslated the english version is the fallback.

core/src/main/resources/i18n/displayStrings_de.properties
deusmax: Hi @chimp1984, 
Great! I will revert  the changes to the language files. Makes it easier !

deusmax:
On the BSQ **bad data** issue.
Exactly as you say, it is not **bad data**. The chart is doing its job correctly. The problem should be solved at the data level, if possible. Doing so will also appeal to new users, who get discouraged or scared from crazy numbers. The numbers are still in the table.
One way to control it is to place a cap on the daily % price change up or down. Like 20 or 50 percent.
This is interesting, but perhaps off-topic here.

I'll add the filtering back-in.

chimp1984:
> One way to control it is to place a cap on the daily % price change up or down. Like 20 or 50 percent.

Yes to apply a 50% diff from market price would be probably good. Though those would be only applied to new offers, and I assume those offers are out since long will stay there for longer...
For BSQ it could be applied as we have sufficient data to calculate the average, but we could not apply that generally to a currency where not external price feed is there, as it might be that there are just very few trades or none at all... But a problem to irrelevant to spend time on ;-)

chimp1984 (changes requested):
Beside the removed outlieer filter for BSQ it looks good to me.

deusmax:
Removed the stock translations and added back the filtering as suggested.
The filtering is generic not specific to BSQ, as before.
Here is the new look of BSQ market offer chart:
Bisq-MarketOfferBook-BSQnew-2020-11-10

chimp1984 (dismissed):
utACK

deusmax:
Needs another review, as I pushed a minor edit - replacing the hard-coded data cutoff value with a constant variable.
I was unaware of the previous review at the time.

chimp1984 (approved):
utACK

chimp1984:
> Needs another review, as I pushed a minor edit - replacing the hard-coded data cutoff value with a constant variable.
> I was unaware of the previous review at the time.

No problem... Thanks for the PR!

deusmax:
It seems the change from openjdk10 to openjdk12, made Travis complain. 
Travis fixed !!

deusmax:
@chimp1984 
Wouldn't it be nice to have this PR part of v1.5 ?
It would contribute another visual change to the version bump.

ripcurlx (approved):
utACK

@deusmax I'll cherry-pick this commit into the v1.5.0 release branch as it doesn't include any critical change and can be verified/tested quite quickly.

deusmax:
Great !

PR #4777, opened 2020-11-10
Remove unused translations

ripcurlx:
While looking for translations that need to be changed because of our new simplified account restrictions I removed all unused translations and added suppress inspection annotations. This helps during development with Intellij to keep our translations file more clean by looking at the warnings within the file.


ripcurlx:
Added [WIP] to the title as I will add the required changes for the simplified account limits and other translation fixes to this branch.

chimp1984 (commented):
utACK

Issue #4775, opened 2020-11-10
[REGTEST] Exception in BSQ wallet transactions view


ghost:
I'm getting an exception in regtest when opening the BSQ transactions view.  Not sure if it is a legitimate problem or if I have a screwed up wallet.  Last time I did any BSQ transactions in regtest was Oct 12 during testing of v1.4.0.  I was sending BSQ from Alice to Bob and vice-versa.  Have not noticed any issues until now.

The exception: `org.bitcoinj.core.SegwitAddress cannot be cast to org.bitcoinj.core.LegacyAddress `

At this line of code: https://github.com/bisq-network/bisq/blob/0345c795e2c227d827a1f239a323dda1250f4e69/desktop/src/main/java/bisq/desktop/main/dao/wallet/tx/BsqTxListItem.java#L106

The transaction that is causing the exception:
<details><summary>details</summary>

    -0.0001 BTC total value (sends 2.49997904 BTC and receives 2.49987904 BTC)
      confidence: Appeared in best chain at height 2671, depth 660. Source: NETWORK
      f54ac3658bcf92df563299007a7f326b096d9e5c3dcf2460aef251a768c40fde, wtxid ca2c67935fb8a549a5f54c312ea480f167c6f07e95f1f9a5040752cb513f0042
      updated: 2020-10-12T14:44:38Z
      included in best chain at: 2020-10-12T14:44:50Z
      purpose: USER_PAYMENT
         in   PUSHDATA(72)[304502210080767bdcb33a70be5a87ebbe5fab47084dd3c734e3380c43a103702afca2b7f6022061f836f490527103b2965d0cb85311e2f4af7c149ab575453c9ddd2c8fc28e8101] PUSHDATA(33)[02218b2da40151be2d3b3d9dd91e8a566f125b485386c4f747305ad899c1e0eb68]  2.49997904 BTC (249997904)
              P2PKH addr:miWJKUMVvcxe9a5ihQitwqkUwspYpEx9ix  outpoint:fb84433d6952db5e3ee650fae6c6eefcdf91e0c433229acc20fd527f1f58ce60:1
         in   <empty>  0.05760397 BTC (5760397)
[...]

PR #4779, opened 2020-11-10
Fix incorrect bsq address handling

chimp1984:
Fixes https://github.com/bisq-network/bisq/issues/4775


ghost (approved):
ACK - this resolves the issue I observed in #4775

ripcurlx (approved):
utACK

PR #4781, opened 2020-11-11
Use 30 days average USD/BSQ price for market cap

chimp1984:
We used latest trade price to calculate the market cap which is confusing and volatile. Better to use the 30 day average.


ripcurlx (approved):
utACK

PR #4782, opened 2020-11-11
Refactor trade limit check

sqrrm:
Return early for anything that doesn't have age dependent trade limits
Remove unnecessary method
Fix switch case

Added another suggested refactor before closing the matter for now.


chimp1984 (commented):
utACK

ripcurlx (approved):
utACK

PR #4784, opened 2020-11-12
Redesign mailbox msg

chimp1984:
This PR fixes an issue with not removed mailbox messages from the trade process.
I redesigned the way how we remove the messages. Instead of removing it from the network after successful processing we remove it immediately when we received it after the listeners are called. We keep it locally stored so in case the client was not ready to process it at that moment it  requests the data later. We also persist the messages for the edge case that the user shuts down the app before the client was ready for processing. In that case the message wuld be lost if not persisted as it was removed from the network but only kept in RAM for later access which is lost at restart. By persisting the data we get them again at restart.
After processing we also try to remove the data if still in the P2PDataStorage to be sure that the message got removed from the network if the initial removal failed because we have not been bootstrapped.

By removing the messages at receipt we can expect that the number of messages will go down quite a lot. Currently its about 1000, normally it was about 400-600. I guess after that release and when old messages have been cleaned from TTL we might have only about 200-300 messages in the network.

This PR did only the minimum of changes to limit risks for this release. We should clean up later a bit more and continue refactoring the mailbox message handling so it gets more robust and uniform.

### DIFF
--- a/core/src/main/java/bisq/core/alert/PrivateNotificationManager.java
+++ b/core/src/main/java/bisq/core/alert/PrivateNotificationManager.java
@@ -133,7 +133,7 @@ public class PrivateNotificationManager {
     }
 
     public void removePrivateNotification() {
-        p2PService.removeEntryFromMailbox(decryptedMessageWithPubKey);
+        p2PService.removeMailboxMsg(decryptedMessageWithPubKey);
     }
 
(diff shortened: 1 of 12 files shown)


ripcurlx (approved):
utACK

sqrrm (commented):
utACK

PR #4786, opened 2020-11-12
Fix historical data store issue

chimp1984:
As we might have same keys in multiple maps and merge those to 1 map we
cannot use an immutable map when merging the maps. Instead we copy our merged map
at the end into a immutable map.

Also prune live data always instead of just when we apply historical data from resource files.


ripcurlx (approved):
ACK

Solved my problem on Regtest

PR #4791, opened 2020-11-13
Remove legacy BSQ explorers

wiz:
@mrosseel @sqrrm @m52go since the legacy BSQ explorers aren't stable, we should remove them for now per @chimp1984


ripcurlx (approved):
utACK

PR #4793, opened 2020-11-13
Define new grpc bsq/btc balances protos

ghubstan:
This change adds proto serivces and messages to support future api implementations for serving bsq, btc or all wallet balances.

- RPC `GetBsqBalances` will return complete BSQ balance info.
- Message `BsqBalanceInfo` is proto returned by rpc `GetBsqBalances`.

- RPC `GetBtcBalances` wil return complete BTC balance info.
- Message `BtcBalanceInfo` is proto returned by rpc `GetBtcBalances`.

- RPC `GetBalances` returns complete BTC and BSQ balance info.
- Message `BalancesInfo` is proto returned by rpc `GetBalances`.

RPC `GetBalance` remains unchnaged;  it still returns only the available
BTC balance.  It may be deprecated and removed in a future PR.

PR #4794, opened 2020-11-13
Add proto wrappers for sending bsq, btc or all balances

ghubstan:
This change adds proto wrappers for sending bsq, btc, or all balances to the CLI.  They will be used in future api method implementations: `getbsqbalance`, `getbtcbalance` and `getbalances`.

This is the second in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4793.  

PR https://github.com/bisq-network/bisq/pull/4793 should be reviewed before this one.


sqrrm (commented):
utACK

PR #4795, opened 2020-11-13
Define proto for api method 'getunusedbsqaddress'

ghubstan:
This change adds a proto to support a future api implementation for  getting an unused, bsq funding address.

This is the third in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4793.  

PR https://github.com/bisq-network/bisq/pull/4794 should be reviewed before this one.


sqrrm (commented):
utACK

PR #4796, opened 2020-11-13
Define proto for api method 'sendbsq'

ghubstan:
This change adds the proto to support a future api implementation for transfering bsq.

This is the 4th in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4793.  

PR https://github.com/bisq-network/bisq/pull/4795 should be reviewed before this one.


ghubstan:
Force travis ci rebuild after it could not download jetty jar.

sqrrm (commented):
utACK

It would make sense to make somewhat bigger PRs, like 4793-4796 as one since they touch the same stuff that's very tightly coupled.

PR #4800, opened 2020-11-13
Stub out api method 'sendbsq' in core

ghubstan:
All the boilerplate for this method is included in this change, but the implementation in `CoreWalletsService#sendBsq` is commented out until the needed logic to send bsq is refactored out of desktop's `BsqSendView` class -- to be included in the next PR.
    
- Added new method to CLI.
    
- Added the `gRPC` server boilerplate.
    
- Added the core implementation, commented out.
    
- Left `sendbsq` tests commented out.

This is the 8th in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4793.  

PR https://github.com/bisq-network/bisq/pull/4799 should be reviewed before this one.

PR #4801, opened 2020-11-13
Refactor desktop's BsqSendView 'sendbsq', share with api

ghubstan:
Moved just enough code out of `BsqSendView` to avoid desktop/api `sendbsq` duplication.

- Created new `BsqTransferModel` to hold tx details shared by desktop and api.

- Created new `BsqTransferService` to send bsq using a `BsqTransferModel` shared  by desktop and api.
  
- Uncommented `CoreWalletsService#sendBsq` implementation.

- Uncommented `sendbsq` tests.  

This is the 9th in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4793.  

PR https://github.com/bisq-network/bisq/pull/4800 should be reviewed before this one.


ghubstan:
Backing out the changes made to `BsqSendView` in commit 446bd32 does not affect the `sendbsq` functionality in the api or desktop, but the small bit of duplication resulting from this single file revert should be removed in a later PR.

core/src/main/java/bisq/core/api/CoreWalletsService.java
```diff
@@ -331,6 +342,25 @@ private void verifyEncryptedWalletIsUnlocked() {
...
+        try {
+            return bsqFormatter.getAddressFromBsqAddress(address);
+        } catch (Throwable t) {
+            log.error("", t);
+            throw new IllegalStateException(format("%s is not a valid bsq address", address));
+        }
+    }
+
+    // Returns a Coin for the double amount, or a RuntimeException if invalid.
+    private Coin getValidBsqTransferAmount(double amount) {
```
sqrrm, line 354: Depending on what you consider a valid amount in this context, you might want to consider any dust amount as invalid since it's too small to send. The naming of this method indicates dust amounts should be invalid.

It's possible to burn less than dust since the burnt satoshis are added to the mining fee and the BSQ change lowered by the burnt amount. That wouldn't be considered a transfer though.

core/src/main/java/bisq/core/btc/model/BsqTransferModel.java
```diff
@@ -0,0 +1,77 @@
...
+        this.miningFee = signedTx.getFee();
+        this.txSize = signedTx.bitcoinSerialize().length;
+        this.txType = TxType.TRANSFER_BSQ;
+    }
+
+    public String getReceiverAddressAsString() {
+        return receiverAddress.toString();
+    }
+
+    public double getMiningFeeInSatoshisPerByte() {
```
sqrrm, line 43: Why use double for satoshis? Long seems better.

core/src/main/java/bisq/core/api/CoreWalletsService.java
```diff
@@ -354,7 +352,7 @@ private LegacyAddress getValidBsqLegacyAddress(String address) {
 
     // Returns a Coin for the double amount, or a RuntimeException if invalid.
     private Coin getValidBsqTransferAmount(double amount) {
-        Coin amountAsCoin = parseToCoin(new BigDecimal(amount).toString(), bsqFormatter);
+        Coin amountAsCoin = parseToCoin(Double.toString(amount), bsqFormatter);
```
sqrrm, line 355: What is the reason to use `double` for any monetary values? In general I have never seen it done and for anything that's dealing with bitcoins I would use `long` and count satoshis.

proto/src/main/proto/grpc.proto
```diff
@@ -340,6 +342,14 @@ message GetUnusedBsqAddressReply {
      string address = 1;
 }
 
+message SendBsqRequest {
+    string address = 1;
+    double amount = 2;
```
sqrrm, line 350: This doesn't make sense to me. BSQ amounts are always going to be an integer amount of satoshis and all other coin amounts in the pb spec are `uint64`.

core/src/main/java/bisq/core/api/CoreWalletsService.java
ghubstan: After much discussion on keybase, changed amount type to String in commit 65df9e1503a9a38518ac29544e8a62cd4ff1045a.

A user needs to be able to enter a representation of a decimal value like "0.25" instead of a satoshi value like 25000000.
The server will convert the string to a Coin for use in a transaction.

proto/src/main/proto/grpc.proto
ghubstan: After much discussion on keybase, changed amount type to String in commit 65df9e1503a9a38518ac29544e8a62cd4ff1045a.

A user needs to be able to enter a representation of a decimal value like "0.25" instead of a satoshi value like 25000000.
The server will convert the string to a Coin for use in a transaction.

core/src/main/java/bisq/core/btc/model/BsqTransferModel.java
ghubstan: After much discussion on keybase, changed amount type to String in commit 65df9e1503a9a38518ac29544e8a62cd4ff1045a.

A user needs to be able to enter a representation of a decimal value like "0.25" instead of a satoshi value like 25000000.
The server will convert the string to a Coin for use in a transaction.

ghubstan: The CoinUtil method is uses returns a double, but deleted this almost unused method anyway in commit 2efd094f68f2b867624a7c27191b42a08bb15f8f.

core/src/main/java/bisq/core/api/CoreWalletsService.java
ghubstan: Commit 7f636e45f995db168501b0772ed13d4a1eb231f1 ensures that any BSQ amt < min dust is invalid.

ghubstan:
This PR is being closed after requested changes were committed in a the final, downstream PR (https://github.com/bisq-network/bisq/pull/4858) in this chain.

Merging of the final PR in the chain implies approval and merging of this PR.

PR #4804, opened 2020-11-13
Improve persistence manager

chimp1984:
We observed some issues with persistence (corrupted files). This PR intends to fix that by adding more checks so that edge cases where it was possible that methods got called repeatadly or in a state where they should not be called anymore are avoided.

### DIFF
--- a/common/src/main/java/bisq/common/persistence/PersistenceManager.java
+++ b/common/src/main/java/bisq/common/persistence/PersistenceManager.java
@@ -44,6 +44,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


common/src/main/java/bisq/common/persistence/PersistenceManager.java
```diff
@@ -166,6 +183,7 @@ private static void onWriteCompleted(ResultHandler completeHandler,
     @Nullable
     private Timer timer;
     private ExecutorService writeToDiskExecutor;
+    public final AtomicBoolean initCalled = new AtomicBoolean(false);
```
sqrrm, line 186: `initCalled` is always false as it's only set here

sqrrm (changes requested):

common/src/main/java/bisq/common/persistence/PersistenceManager.java
chimp1984: Ups, thanks for spotting that...

sqrrm (approved):
utACK

PR #4810, opened 2020-11-14
[WIP] Learning to config travic-ci to print gradle test reports

ghubstan:
This PR should never be merged.  Don't even look.


ghubstan:
All done.

Appending the following to the yaml config file prints the gradle test report, including stdout, for a single test case:

```
after_script:
    - pwd
    - ls -l core/build/reports/tests/test/classes/bisq.core.util.ProtoUtilTest.html
    - cat core/build/reports/tests/test/classes/bisq.core.util.ProtoUtilTest.html

```

I needed to do this because a test case that was failing only on Travis CI, with no way to see the test case's stdout/stderr.  This will help me debug when I check it in again.

Issue #4806, opened 2020-11-14
Transaction returned to "start payment2 status after PC restarted


Stam57:
SUPPORT REQUESTS: This is for reporting bugs in the Bisq app.
   If you have a support request, please join #support on Bisq's
   Keybase team at https://keybase.io/team/Bisq

### Description

  Last night I completed the transaction by paying the seller. The seller verified the payment and I received in my phone a 'transaction complete' message through the bisq app. I never visited the bisq app in my PC after that until this morning when I realised that the PC had done an automatic restart because of some new windows updates. When I opened bisq in the PC I received a warning message about the time left to complete the transaction. I pressed the 'payment started' button and then I received the warning about the bug.

#### Version

 1.4.2

### Steps to reproduce

as in description above

### Expected behaviour

the transaction should have been ended and completed

### Actual behaviour

the transaction returned to a previous step (step 2: start payment)

### Screenshots

Screenshots if gui related, drag and drop to add to the issue

#### Device or machine

windows 10, intel CPU

#### Additional info

 Additional information useful for debugging (e.g. logs)
[bisq2.log](https://github.com/bisq-network/bisq/files/5540454/bisq2.log)

ghost:
From looking at the log, you had two trades (NPNOY and 89599304) which successfully completed & paid to your wallet; then at midnight your machine did an automatic reboot.  When Bisq was started again the trades erroneously reverted to an earlier state.  I am investigating this as it looks like a bug.  I'll attempt to reproduce the error and update here with any progress.  Thanks.

ghost:
This is reproducible by killing the Bisq process without allowing it to do a graceful shutdown, which is what the Windows midnight reboot did.  The trade state was not able to be saved to disk, so upon startup the last-saved state was back at when the trade was created.  However it does fully recover after pressing start payment again, it goes straight to the final stage which is send funds to external address / keep funds in Bisq wallet.

Given that the scenario is quite rare, and that it does recover quite gracefully I would lean towards not spending resources on developing and testing a fix.

Alternatively, a simple fix may be to call `requestPersistence` after each state change.

@chimp1984 opinions?

chimp1984:
@jmacxx Ah thanks for finding out the issue! I suspected that there are hard shutdown scenarios where the peristence at shutdown does not get triggered. 
[...]

Issue #4762, opened 2020-11-07
Delayed payout transaction is missing


huey735:
### Description
Bisq failed to create the delayed payout transaction needed to send funds to Arbitration.

#### Version
1.4.2

### Steps to reproduce
Unable to reproduce

### Expected behaviour
Bisq should create a delayed payout transaction sending funds to the Arbitrator address along with the deposit transaction.

### Actual behaviour
Bisq failed to create the delayed payout transaction needed to send funds to Arbitration.

### Screenshots

<img width="957" alt="btcBuyerTaker-errorMessage" src="https://user-images.githubusercontent.com/44140017/98450377-43890000-2134-11eb-9c97-9fdc2d29f249.PNG">

<img width="960" alt="btcBuyerTaker-tradeDetails" src="https://user-images.githubusercontent.com/44140017/98450359-15a3bb80-2134-11eb-9553-719271f2c3aa.PNG">

#### Device or machine
Windows 10

chimp1984:
As stated in the popup, instruct both traders to open a mediation and accept the suggested payout so that the trade amount goes back to seller (assume buyer has not send funds as stated in the popup). That way the trade can be closed without further risk. Please request the log files from both traders and PM me.

huey735:
The Deposit transaction also wasn't successfully created despite the message. So the trade failed in the end. I'll send you the logs.

chimp1984:
Was the trader the seller or buyer? It would be helpful if users dont blur so much data. Most data is not relevant regarding privacy...

PR #4789, opened 2020-11-13
Fix remaining blackmail vulnerabilities

stejbac:
This PR attempts to take advantage of the SegWit trade protocol upgrade (hard fork), to fix all the remaining cases where a buyer or seller can publish a deposit tx without their peer having a valid delayed payout tx (at least in the fully SegWit case, when taking an offer created in v1.5.0+). When taking pre-1.5.0 offers, the deposit tx inputs will be of mixed type, which makes it malleable and thus impossible to completely prevent either trader from altering its ID and thus rendering their peer's delayed payout tx invalid. In that case, the buyer should still fully validate their delayed payout tx anyway, against the agreed deposit tx, so that if the latter changes before confirming, the trade hopefully fails before the buyer makes payment. The PR additionally fixes a bug where (it appears) the buyer fails to check the signature of the final delayed payout tx.

These changes are intended to prevent blackmail attacks (or reduce the risk of them in not-fully-SegWit cases where the deposit tx is malleable).

PR #4816, opened 2020-11-18
Fix issues with missing persistence for trade state

chimp1984:
There are some bug reports which suggest that the persistence at shutdown is not working reliable (hard kill, crash,...).
We reduce the persistence interval radically to ensure important data is written immediately (200 ms delay).
Further the `requestPersistence` has been missing in most trade tasks. I guess that happened when merging the large refactoring projects (trade protocol, persistence). If shutdown routine gets executed it did not had any effect as we wrote then the trade state, but if that was missing we might miss important trade state/data changes.
 
Fixes https://github.com/bisq-network/bisq/issues/4806
Fixes https://github.com/bisq-network/bisq/issues/4762

### DIFF
--- a/core/src/main/java/bisq/core/trade/protocol/BuyerAsTakerProtocol.java
+++ b/core/src/main/java/bisq/core/trade/protocol/BuyerAsTakerProtocol.java
@@ -65,7 +65,6 @@ public class BuyerAsTakerProtocol extends BuyerProtocol implements TakerProtocol
 
         Offer offer = checkNotNull(trade.getOffer());
         processModel.getTradingPeer().setPubKeyRing(offer.getPubKeyRing());
-        processModel.getTradeManager().requestPersistence();
     }
 
 
--- a/core/src/main/java/bisq/core/trade/protocol/BuyerAsTakerProtocol.java
+++ b/core/src/main/java/bisq/core/trade/protocol/BuyerAsTakerProtocol.java
@@ -65,6 +65,7 @@ public class BuyerAsTakerProtocol extends BuyerProtocol implements TakerProtocol
 
         Offer offer = checkNotNull(trade.getOffer());
         processModel.getTradingPeer().setPubKeyRing(offer.getPubKeyRing());
+        processModel.getTradeManager().requestPersistence();
     }
 
 
@@ -83,7 +84,10 @@ public class BuyerAsTakerProtocol extends BuyerProtocol implements TakerProtocol
                         BuyerAsTakerCreatesDepositTxInputs.class,
                         TakerSendInputsForDepositTxRequest.class)
                         .withTimeout(30))
-                .run(() -> processModel.setTempTradingPeerNodeAddress(trade.getTradingPeerNodeAddress()))
+                .run(() -> {
+                    processModel.setTempTradingPeerNodeAddress(trade.getTradingPeerNodeAddress());
+                    processModel.getTradeManager().requestPersistence();
+                })
                 .executeTasks();
     }
 
--- a/core/src/main/java/bisq/core/trade/protocol/SellerAsTakerProtocol.java
+++ b/core/src/main/java/bisq/core/trade/protocol/SellerAsTakerProtocol.java
@@ -59,7 +59,6 @@ public class SellerAsTakerProtocol extends SellerProtocol implements TakerProtoc
         super(trade);
         Offer offer = checkNotNull(trade.getOffer());
         processModel.getTradingPeer().setPubKeyRing(offer.getPubKeyRing());
-        processModel.getTradeManager().requestPersistence();
     }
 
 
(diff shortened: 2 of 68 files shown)


ghost:
This would not resolve #4806 since `requestPersistence` for PendingTrades is only called twice: when the trade is new, and when the user closes the trade.  There was > 12 hours in between where the trade progressed through all its states with no call to `requestPersistence`.  

**[EDIT]**  Ok, I see you're trying to persist in SIGTERM.  Maybe it would be ok.  If the delay is only 200ms maybe that is enough time before the process is shut down.  It wouldn't help in the case of power cut though.

chimp1984:
Oh, yes you are right. I though at trade task complete we call it but I fear that got lost in some merge. It was a bit complicate with several larger PRs (refactoring of trade protocol + refactoring of persistence).
I will check how to fix it best...

chimp1984:
I added the `requestPersistence` at each data change in trade. Hope I did not miss any. We could add some redundancy if we call it at the `complete()` method of each tradeTask but not sure if thats needed.

chimp1984:
Failing test is resolved by @stejbac (https://github.com/bisq-network/bisq/pull/4789#issuecomment-729586729). Will merge in once hie PR is merged.

ghost:
ACK, tested, fixes #4806.

chimp1984:
I double checked all trade (and sub models) related fields as well as all trade protocol tasks and protocols. Hope I did not miss a call but as there is quite a lot of redundancy I think its safe now.

sqrrm:
I feel it might be cleaner to do the persist in the `complete()` method at the end of each trade task. It feels like that's what's being done with this PR anyway, but separately in each task with the chance of missing it somewhere.

Apart from that I think this looks ok and could be merged.

core/src/main/java/bisq/core/trade/protocol/BuyerAsTakerProtocol.java
```diff
@@ -65,6 +65,7 @@ public BuyerAsTakerProtocol(BuyerAsTakerTrade trade) {
 
         Offer offer = checkNotNull(trade.getOffer());
         processModel.getTradingPeer().setPubKeyRing(offer.getPubKeyRing());
+        processModel.getTradeManager().requestPersistence();
```
ghost: I'm getting a NPE here, processModel.tradeManager is null

core/src/main/java/bisq/core/trade/protocol/SellerAsTakerProtocol.java
```diff
@@ -59,6 +59,7 @@ public SellerAsTakerProtocol(SellerAsTakerTrade trade) {
         super(trade);
         Offer offer = checkNotNull(trade.getOffer());
         processModel.getTradingPeer().setPubKeyRing(offer.getPubKeyRing());
+        processModel.getTradeManager().requestPersistence();
```
ghost: I get a NPE here, processModel.tradeManager is null.

ghost (changes requested):

core/src/main/java/bisq/core/trade/protocol/BuyerAsTakerProtocol.java
chimp1984: Ah in the constructor it must not be called as trademanager is not set yet. I will remove that.

core/src/main/java/bisq/core/trade/protocol/SellerAsTakerProtocol.java
chimp1984: Ah in the constructor it must not be called as trademanager is not set yet. I will remove that.

chimp1984:
> I feel it might be cleaner to do the persist in the `complete()` method at the end of each trade task. It feels like that's what's being done with this PR anyway, but separately in each task with the chance of missing it somewhere.
> 
> Apart from that I think this looks ok and could be merged.

I was wondering myself if we should add at the tradeTask complete handler a perist call. It would add extra redundancy in case I missed one as well in case we add some tasks or have some changes and the call is forgotten. 
I have no strong opinion for or against it. Signal with up or downvote if you want to have it in.

sqrrm:
With the persistence call in complete() it's not necessary to persist in each trade task.

chimp1984:
> With the persistence call in complete() it's not necessary to persist in each trade task.

A reason why I did not do want to rely only on complete is as there are async tasks and i want to be sure that if the task never completes (shutdown) that the intermediate state change is persisted. But also not happy that its that verbose now ;-(. But its more similar as it was before, just that some calls have been done inside the trade setter methods.

sqrrm:
Agreed that some persist calls should still be done in the tasks, but all of those done synchronously just before complete() are no longer needed.

PR #4817, opened 2020-11-18
Improve filter

chimp1984:
Add delay between remove and add new filter

We saw that some seed nodes have 2 filters after filter update.
This should not be the case as the remove is broadcast before the
add but seems there is some issue in the P2P storage which does
not cover that correctly.
By adding a 5 sec delay between the remove and add we mitigate
that issue, though should be fixes in the P2P layer but that
will be a more complex and larger effort.


ripcurlx (approved):
utACK

Is there a reason why you chose exactly a 5 sec delay?

PR #4818, opened 2020-11-18
Add api method 'getpaymentmethods'

ghubstan:
Returns a list of supported payment method IDs.  This gives CLI users the correct payment method id for creating a new payment account.

- Added gRPC service `GetPaymentMethods`.

- Added gRPC boilerplate method to `GrpcPaymentAccountsService`.

- Added implimentation to `CoreApi` -> `CorePaymentAccountsService`.

- Refactored PaymentAccount test hierarchy.

- Add api method to CLI.

This is the 13th in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4793.  

PR https://github.com/bisq-network/bisq/pull/4808 should be reviewed before this one.

PR #4819, opened 2020-11-18
Add api method 'getpaymentacctform'

ghubstan:
Takes a paymentMethodId, and returns a new payment account form as a json String to the CLI, which persists the json string to file in the CLI user's current working directory.  The user will edit the json file and pass it as an argument to api method `createpaymentacct`.

- Added new `common.util.ReflectionUtils`, to help with json acct form serialization/de-serialization.  

- Added convenience `isCountryBasedPaymentAccount` method to `abstract PaymentAccount` to help with json acct form serialization/de-serialization.

- Added new (gson) `core.api.model.PaymentAccountTypeAdapter`.

- Added new `core.api.model.PaymentAccountForm`.  This class uses the gson `PaymentAccountTypeAdapter` to generate editable json forms for `PaymentAccount` sub-types, and instantiate `PaymentAccount` instances from edited json forms.

- Added gRPC service `GetPaymentAccountForm`.

- Added gRPC boilerplate method to `GrpcPaymentAccountsService`.

- Added implimentation to `CoreApi` -> `CorePaymentAccountsService`.

- Added api method to CLI.  

- Added api tests.

This is the 14th in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4793.  

PR https://github.com/bisq-network/bisq/pull/4818 should be reviewed before this one.


ghubstan:
Backing out the changes made to `BsqSendView` in commit 446bd32 does not affect the `sendbsq` functionality in the api or desktop, but the small bit of duplication resulting from this single file revert should be removed in a later PR.

cli/src/main/java/bisq/cli/CliMain.java
```diff
@@ -529,6 +555,23 @@ private static Method getMethodFromCmd(String methodName) {
         return Method.valueOf(methodName.toLowerCase());
     }
 
+    private static void saveFileToDisk(String prefix, String suffix, String text) {
+        String timestamp = Long.toUnsignedString(new Date().getTime());
+        String relativeFileName = prefix + "_" + timestamp + suffix;
+        try {
+            Path path = Paths.get(relativeFileName);
```
sqrrm, line 562: Where will this be written? Active directory from where cli is run?

apitest/src/test/java/bisq/apitest/method/payment/AbstractPaymentAccountTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+
+
+import bisq.apitest.method.MethodTest;
+
+@Slf4j
+public class AbstractPaymentAccountTest extends MethodTest {
+
+    static final String PROPERTY_NAME_COMMENT = "_COMMENT_";
+    static final String PROPERTY_VALUE_COMMENT = "Please do not edit the paymentMethodId field.";
```
sqrrm, line 41: I don't think there is a need to be polite. Better with something like `"Do not manually edit the paymentMethodId field"`

core/src/main/java/bisq/core/api/model/PaymentAccountForm.java
```diff
@@ -0,0 +1,216 @@
...
+@Singleton
+@Slf4j
+public class PaymentAccountForm {
+
+    private final GsonBuilder gsonBuilder = new GsonBuilder()
+            .setPrettyPrinting()
+            .serializeNulls();
+
+    // Names of PaymentAccount fields to exclude from json forms.
+    private final String[] excludedFields = new String[]{
```
sqrrm, line 116: Why are all these fields excluded? Some probably should be but at least SALT would be necessary to set manually to be able to recover an account.

core/src/main/java/bisq/core/api/model/PaymentAccountTypeAdapter.java
```diff
@@ -0,0 +1,329 @@
...
+        this.isExcludedField = (f) -> Arrays.stream(excludedFields).anyMatch(e -> e.equals(f.getName()));
+        this.fieldSettersMap = getFieldSetterMap();
+    }
+
+    @Override
+    public void write(JsonWriter out, PaymentAccount account) throws IOException {
+        // We write a blank payment acct form for a payment method id.
+        // We're not serializing a real payment account instance here.
+        if (log.isDebugEnabled())
+            log.debug("Writing PaymentAccount json form for fields with accessors...");
```
sqrrm, line 98: Why check `if isDebugEnabled()` instead of just `log.debug(...)` without the `if`?

sqrrm (commented):
I had hoped it would be easier to add the generic form for accounts, like using the protobuf from accountPayloads, but I can't figure how it could be done. I like the slim api, just a bit cumbersome to understand the serialization and reflection. I think it looks mostly good though. Just a few comments that need some more thought.

cli/src/main/java/bisq/cli/CliMain.java
ghubstan: Yes, the file with a timestamp in its name is written to the active directory from where cli is run, or what I would call the current working directory.  The filename can be changed to anything, anywhere, then  be used as the single argument to `createpaymentacct`.  If the file is in the active directory, the path does not need to be specified, e.g.,
`$ createpaymentacct zelle_form.json`

apitest/src/test/java/bisq/apitest/method/payment/AbstractPaymentAccountTest.java
ghubstan: OK.  Politeness will be removed.

core/src/main/java/bisq/core/api/model/PaymentAccountForm.java
ghubstan: These are fields the user would not see or fill while creating a payment account in the UI.

core/src/main/java/bisq/core/api/model/PaymentAccountTypeAdapter.java
ghubstan: It is intended to save unnecessary time in the logging layer, that's why I do this.  In this case, the performance savings is trivially small because the logged message is just a string literal.

For things like 
```
    log.debug("{} blah {} blah blah {} ... {} .. {}", 
       someCheapCalculation1, 
       someCheapCalculation2, 
       someExpensiveCalculation3, 
       literal4, 
       someReallyExpensiveCalculation5);
```

the cheap and expensive calculations are performed even if the log level is ERROR.

If wrapped by `if (log.isDebugEnabled())`, the calculations are skipped.  But I don't like the additional lines either.  And in this case it buys nothing, I'll fix it.

I think there are situations where this is a good thing to do, but this isn't one of them.   There are other places I'll fix for the same reason.

sqrrm: Agreed that it's worth doing in certain cases where the log is incurring a significant load and/or is very frequent.

ghubstan: Unnecessary `log.isDebugEnabled()` were removed in commit 45cfb952450ddfa50458aca259bf90447c52dab9.

apitest/src/test/java/bisq/apitest/method/payment/AbstractPaymentAccountTest.java
ghubstan: Fixed in commit 62a2aa85434213493c6d5225de9b694123ad67c4.

core/src/main/java/bisq/core/api/model/PaymentAccountForm.java
ghubstan: Wouldn't it be better to have a `recoverpaymentacct salt` api method?

Do you or can you `createpaymentacct` with a predefined salt?

sqrrm: You can set the salt, it's used as part of the hash to create the accountAgeWitness so it needs to be set to the same salt as previously if an account is to be created on a new client with the same details and account age preserved. Making it optional is reasonable I think. Default to generate a random salt, but if set then the user set salt will be used.

ghubstan: Fixed in commit 9b4bdfc5ad8ff3e76296cb134b55ceba4397d158

ghubstan:
This PR is being closed after requested changes were committed in a the final, downstream PR (https://github.com/bisq-network/bisq/pull/4858) in this chain.

Merging of the final PR in the chain implies approval and merging of this PR.

PR #4820, opened 2020-11-18
Use AES key for signing when wallet is encrypted

oscarguindzberg:
I forgot to pass the AES key as a parameter. Without this change, encrypted wallets can not sign transactions


ripcurlx (approved):
utACK

ripcurlx:
I'll merge it although Travis is complaining as the fix for the failing tests is already included in commit https://github.com/bisq-network/bisq/pull/4789/commits/fd798952e8655ce07fcd07e80c2cf7fbb03dee45

PR #4825, opened 2020-11-19
Rename getsAll payout option string

huey735:
Renamed it to better represent the outcome of this option since
the winning party doesn't in fact get all the funds in escrow.

To replace https://github.com/bisq-network/bisq/pull/4629 as I was unable to properly squash all commits.


chimp1984 (approved):
utACK

ripcurlx (approved):
utACK

PR #4841, opened 2020-11-23
Prompt for confirmation before initiating SPV resync

### LINKED ISSUE #4833: Deleting SPV files should ask for confirmation
### Description

When clicking that button in the Network settings the application simply shows a pop-up mentioning the files will be deleted on the next start up. The only option at that point is to shut down. Considering that it can take a long time to resync and the user might not be aware of that before clicking, it would be nice to have an option to cancel the action there.  

#### Version

Bisq 1.4.2

ghost:
Allows user to back off if invoked by accident.

Fixes #4833 

### Screenshots:

After pressing the resynch button: (user can still cancel)
image

After reboot:
image

After chain has synched:
image


ripcurlx (approved):
utACK

PR #4843, opened 2020-11-23
Add 1 sec delay before calling exit

chimp1984:
To persist in the very last moment before exit might cause problems on some OS.
We do not have confirmed that this might be an issue but to be on the safe side
we add a 1 sec. delay between persistence completed and exit.


ripcurlx (approved):
utACK

ripcurlx:
Resolved conflict and will merge it as soon as Travis is finished.

PR #4848, opened 2020-11-25
In seednode installer, don't run random shell script to install git lfs

wiz:
This was only needed on very old versions of ubuntu, and in any case we shouldn't call some random shell script from the internet hosted by a TTP


ripcurlx (approved):
utACK

PR #4851, opened 2020-11-26
Improve getBlocks request handling

chimp1984:


ghubstan (commented):
utACK

core/src/main/java/bisq/core/dao/node/lite/LiteNode.java
```diff
@@ -87,7 +95,24 @@ public void start() {
...
+            if (blockDownloadListener == null) {
+                blockDownloadListener = (observable, oldValue, newValue) -> {
+                    if ((double) newValue == 1) {
+                        setupWalletBestBlockListener();
+                        UserThread.execute(() -> walletsSetup.downloadPercentageProperty().removeListener(blockDownloadListener));
+                    }
+                };
+                walletsSetup.downloadPercentageProperty().addListener(blockDownloadListener);
+            }
+        }
```
sqrrm:
```suggestion
        } else if (blockDownloadListener == null) {
            blockDownloadListener = (observable, oldValue, newValue) -> {
                if ((double) newValue == 1) {
                    setupWalletBestBlockListener();
                    UserThread.execute(() -> walletsSetup.downloadPercentageProperty().removeListener(blockDownloadListener));
                }
            };
            walletsSetup.downloadPercentageProperty().addListener(blockDownloadListener);
        }
```

core/src/main/java/bisq/core/dao/node/lite/LiteNode.java
```diff
@@ -97,20 +122,20 @@ public void start() {
                 checkForBlockReceivedTimer.stop();
             }
 
-            int height = block.getHeight();
-            log.info("New block at height {} from bsqWalletService", height);
+            int btcWalletHeight = btcBlock.getHeight();
+            log.error("New block at height {} from bsqWalletService", btcWalletHeight);
```
sqrrm: Why error log level?

core/src/main/java/bisq/core/dao/node/lite/LiteNode.java
```diff
@@ -97,20 +122,20 @@ public void start() {
...
+                if (daoChainHeight < btcWalletHeight) {
                     log.warn("We did not receive a block from the network {} seconds after we saw the new block in BicoinJ. " +
                                     "We request from our seed nodes missing blocks from block height {}.",
-                            CHECK_FOR_BLOCK_RECEIVED_DELAY_SEC, chainHeight + 1);
-                    liteNodeNetworkService.requestBlocks(chainHeight + 1);
+                            CHECK_FOR_BLOCK_RECEIVED_DELAY_SEC, daoChainHeight + 1);
+                    liteNodeNetworkService.requestBlocks(daoChainHeight + 1);
                 }
-            }, CHECK_FOR_BLOCK_RECEIVED_DELAY_SEC);
+            }, CHECK_FOR_BLOCK_RECEIVED_DELAY_SEC, TimeUnit.MILLISECONDS);
```
sqrrm: Timeout unit should be sec, is this a test artefact together with the error log level?

core/src/main/java/bisq/core/dao/node/lite/LiteNode.java
```diff
@@ -229,11 +257,12 @@ private void runDelayedBatchProcessing(List<RawBlock> blocks, Runnable resultHan
     // We received a new block
     private void onNewBlockReceived(RawBlock block) {
         int blockHeight = block.getHeight();
-        log.debug("onNewBlockReceived: block at height {}, hash={}", blockHeight, block.getHash());
+        log.info("onNewBlockReceived: block at height {}, hash={}. Our DAO chainHeight={}", blockHeight, block.getHash(), chainTipHeight);
```
sqrrm: That's quite a bit of logging but perhaps ok since it's once per received block

core/src/main/java/bisq/core/dao/node/lite/network/RequestBlocksHandler.java
```diff
@@ -108,56 +108,55 @@ public void cancel() {
...
-                        handleFault(errorMessage, nodeAddress, CloseConnectionReason.SEND_MSG_FAILURE);
+                        String errorMessage = "A timeout occurred when sending getBlocksRequest:" + getBlocksRequest +
+                                " on peersNodeAddress:" + nodeAddress;
+                        log.debug("{} / RequestDataHandler={}", errorMessage, RequestBlocksHandler.this);
+                        handleFault(errorMessage, nodeAddress, CloseConnectionReason.SEND_MSG_TIMEOUT);
                     } else {
-                        log.trace("We have stopped already. We ignore that networkNode.sendMessage.onFailure call. " +
-                                "Might be caused by a previous timeout.");
+                        log.warn("We have stopped already. We ignore that timeoutTimer.run call. " +
+                                "Might be caused by an previous networkNode.sendMessage.onFailure.");
```
sqrrm:
```suggestion
                                "Might be caused by a previous networkNode.sendMessage.onFailure.");
```

core/src/main/java/bisq/core/dao/node/lite/network/RequestBlocksHandler.java
```diff
@@ -168,31 +167,36 @@ public void onFailure(@NotNull Throwable throwable) {
     @Override
     public void onMessage(NetworkEnvelope networkEnvelope, Connection connection) {
         if (networkEnvelope instanceof GetBlocksResponse) {
```
sqrrm, line 165:
```suggestion
        if (!(networkEnvelope instanceof GetBlocksResponse)) return; // Also need to remove other brace
```

core/src/main/java/bisq/core/dao/node/lite/network/RequestBlocksHandler.java
```diff
@@ -168,31 +167,36 @@ public void onFailure(@NotNull Throwable throwable) {
...
+            if (getBlocksResponse.getRequestNonce() != nonce) {
+                log.warn("Nonce not matching. That can happen rarely if we get a response after a canceled " +
+                                "handshake (timeout causes connection close but peer might have sent a msg before " +
+                                "connection was closed).\n\t" +
+                                "We drop that message. nonce={} / requestNonce={}",
+                        nonce, getBlocksResponse.getRequestNonce());
+                return;
+            }
+
+            cleanup();
```
sqrrm: Doesn't timeoutTimer need to be stopped here like before?
```
        stopTimeoutTimer();
```

core/src/main/java/bisq/core/dao/node/lite/LiteNode.java
chimp1984: I will make a commit with the suggestion. Have better control if I do it in the IDE...

core/src/main/java/bisq/core/dao/node/lite/LiteNode.java
chimp1984: Ups was from dev testing

core/src/main/java/bisq/core/dao/node/lite/LiteNode.java
chimp1984: Ups, another dev setup leftover ;-(

core/src/main/java/bisq/core/dao/node/lite/LiteNode.java
chimp1984: We can lower loglevel once we see that the feature works more reliable. Prefer to keep it now.

core/src/main/java/bisq/core/dao/node/lite/network/RequestBlocksHandler.java
chimp1984: Found more like that. Will fix in new commit.

core/src/main/java/bisq/core/dao/node/lite/LiteNode.java
sqrrm: Yes, I only make the code suggestion here as a way to explain more precisely what I mean. Best do it on your own to be sure it's correct.

core/src/main/java/bisq/core/dao/node/lite/network/RequestBlocksHandler.java
chimp1984: cleanup calls stopTimeoutTimer. But will consolidate cancel/stop/cleanup

core/src/main/java/bisq/core/dao/node/lite/network/RequestBlocksHandler.java
chimp1984: I prefer to not return early here as this style is commonly used in many places and often there are several if else branches. It is more like a dispatcher or switch case style (if switch would be more flexible in java). I prefer to use the early return for validation use cases but here its more to select the msg we are interested in.

sqrrm: That's a good point. Might be cleaner to make a `handleGetBlocksResponse(GetBlocksResponse r)` and other respective dispatch handlers like is done in the trade protocol for these cases.

ripcurlx:
@sqrrm Is there anything else you want @chimp1984 to address?

sqrrm:
@ripcurlx I have nothing more but in private @chimp1984 mentioned he might want to add some more so I'll wait for that to merge.

chimp1984:
@ripcurlx I would like to do myself another review before merging it.

ManfredKarrer (approved):
utACK

sqrrm (approved):
utACK

PR #4855, opened 2020-11-26
Make requested PR changes and minor bug fixes

ghubstan:
This PR contains change requested to prior PRs in this chain:

- Resolve https://github.com/bisq-network/bisq/pull/4799#discussion_r530311866

- Resolve https://github.com/bisq-network/bisq/pull/4799#discussion_r530324726

- Add create `AmazonGiftCardAccount` api test.

- Implement `OfferInfo.fromProto`

- Add missing newlines to CLI's console output statements.

- Remove deprecated api's `CoreWalletsService#getAvailableBalance`.

This is the 20th in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4793.  

PR https://github.com/bisq-network/bisq/pull/4836 should be reviewed before this one.

PR #4858, opened 2020-11-26
Add tx fee rate api methods:  gettxfeerate, settxfeerate, unsettxfeerate

ghubstan:
These new api methods are for viewing the `FeeService's` current tx fee rate, and setting/unsetting the custom tx fee rate preference.

This is the 21st in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4793.  

PR https://github.com/bisq-network/bisq/pull/4855 should be reviewed before this one.


cli/src/main/java/bisq/cli/CurrencyFormat.java
```diff
@@ -50,6 +53,23 @@ public static String formatBsq(long sats) {
         return BSQ_FORMAT.format(BigDecimal.valueOf(sats).divide(BSQ_SATOSHI_DIVISOR));
     }
 
+    public static String formatTxFeeRate(TxFeeRateInfo txFeeRateInfo) {
+        String stdTxFeeRate = formatTxFeeRate(txFeeRateInfo.getStdTxFeeRate());
+        String customTxFeeRate = txFeeRateInfo.getCustomTxFeeRate() < 0
+                ? formatTxFeeRate(txFeeRateInfo.getCustomTxFeeRate())
+                : null;
```
sqrrm: I don't understand what's going on here.

core/src/main/java/bisq/core/api/CoreWalletsService.java
```diff
@@ -185,6 +196,37 @@ void sendBsq(String address,
         }
     }
 
+    TxFeeRateInfo getTxFeeRate() {
+        try {
+            CompletableFuture<Void> feeRequestFuture = CompletableFuture.runAsync(feeService::requestFees);
+            feeRequestFuture.get();  // Block until async fee request is complete.
```
sqrrm: Is it really a good idea to block here? As discussed before, could we use a callback to avoid blocking?

core/src/main/java/bisq/core/api/CoreWalletsService.java
```diff
@@ -185,6 +196,37 @@ void sendBsq(String address,
...
+        Coin stdTxFeeRate = feeService.getTxFeePerVbyte();
+        Coin customTxFeeRate = preferences.isUseCustomWithdrawalTxFee()
+                ? Coin.valueOf(preferences.getWithdrawalTxFeeInVbytes())
+                : Coin.NEGATIVE_SATOSHI;
+
+        return new TxFeeRateInfo(stdTxFeeRate.value, customTxFeeRate.value);
+    }
+
+    TxFeeRateInfo setTxFeeRatePreference(long txFeeRate) {
+        if (txFeeRate <= 0)
```
sqrrm, line 233: We also don't want to create txs with fees < 1 sat/b, but perhaps that doesn't have to be checked here.

cli/src/main/java/bisq/cli/CurrencyFormat.java
ghubstan: To clear this up, I should probably add a `boolean useCustomWithdrawalTxFee` field to `TxFeeRateInfo`.  This would be set from the `Preference#useCustomWithdrawalTxFee` value.     (Remember that `Preferences` is not visible to the CLI, and these `*Info` proto wrappers are for passing the minimum data needed from server to CLI.)

This confusion is caused by not having that field.  When `txFeeRateInfo.getCustomTxFeeRate() == -1`, the custom tx fee rate preference is ignored in the CLI.  In core, it means the preference.useCustomWithdrawalTxFee == false.

In pseudo English, it means

```
String customTxFeeRate = (if are we using the fee service's tx fee rate)
                ? (format the custom tx fee rate)
                : null;

```

Adding a `boolean useCustomWithdrawalTxFee` field to `TxFeeRateInfo` would fix this.

core/src/main/java/bisq/core/api/CoreWalletsService.java
ghubstan: I'm not sure what the difference would be between future.get() and using a callback.   The wait for a response has to happen somewhere before `GrpcWalletsService` returns the gRPC reply.

Do you think `FeeService#requestFees` should be refactored to take a callback parameter?  Were you and chimp wanting to  refactor FeeService#requestFees?  Or should that be left alone, and make adjustments only in the api?

Another way would be to deal with the result handling is do it in the `GrpcWalletsService`.  I try to keep those boilerplate `Grpc*Service` classes as small as possible, but sometimes dealing with async calls in that layer is unavoidable.

If we are concerned about blocking forever, we could block with a timeout:  
`feeRequestFuture.get(5, SECONDS);  // Block until async fee request is complete.`

core/src/main/java/bisq/core/api/CoreWalletsService.java
ghubstan: Doing this check here makes the CLI's error msg stylistically consistent with others.  

This brings up a general issue that is going to need quite a bit of work in the API:  displaying validation error messages coming from the server's core, in the CLI.    But for now, getting all the v1 methods working for correct parameters is the short term priority.

cli/src/main/java/bisq/cli/CurrencyFormat.java
ghubstan: Fixed in commit faf030fbc5810e46a0d6fdab154b154ef39f72db.

core/src/main/java/bisq/core/api/CoreWalletsService.java
sqrrm: I was thinking something like, in CoreWalletService
```
    TxFeeRateInfo getTxFeeRate(ResultHandler resultHandler, ErrorMessageHandler errorMessageHandler) {
            CompletableFuture<Void> feeRequestFuture = CompletableFuture.runAsync(feeService::requestFees);
            Futures.addCallback(feeRequestFuture, new FutureCallback<>() {

                @Override
                public void onSuccess(@Nullable Object result) {
                    resultHandler.handleResult();
                }

                @Override
                public void onFailure(Throwable t) {
                    errorMessageHandler.handleErrorMessage(t.getMessage());
                }
            }, MoreExecutors.directExecutor());
    }
```
where the resultHandler is doing the gRPC reply, something like
```
    @Override
    public void getTxFeeRate(GetTxFeeRateRequest req,
                             StreamObserver<GetTxFeeRateReply> responseObserver) {
        ResultHandler resultHandler = () -> {
            var reply = GetTxFeeRateReply.newBuilder()
                    .setTxFeeRateInfo(txFeeRateInfo.toProtoMessage())
                    .build();
            responseObserver.onNext(reply);
            responseObserver.onCompleted();
        };
        coreApi.getTxFeeRate(resultHandler, errorHandler);
    }
```
Bad code, just to illustrate the idea.

core/src/main/java/bisq/core/api/CoreWalletsService.java
sqrrm: Agreed, not an issue for now, good to think of how to handle it though as you say.

core/src/main/java/bisq/core/api/CoreWalletsService.java
ghubstan: This is exactly [erhmmm... similar] what I'm doing, will test soon...

ghubstan: Commit 987d89319e8a824033879846ac43003ea219cfd0 is a solution to the blocking problem.  I don't pass an error handler to keep it simple, and not do that before I understand how it would help here.

sqrrm (commented):
utACK

I think this looks much better and this is probably ready for merging now. We'll have to put some more thought into the general thread handling from the api side but that's not for this PR.

sqrrm (approved):
utACK

ghubstan:
Force another travis build because a dependency download failed that last time.

PR #4860, opened 2020-11-27
Add taker check for deposit amount

stejbac:
Make sure the taker checks the value of the 2-of-2 multisig output of the deposit tx created by the maker, before signing it. This avoids a potential security hole, where the maker attempts to steal most of the deposit by using the wrong output value and adding an extra 'change' output to himself.

Note that there's no actual vulnerability at present, as the output value is indirectly checked via the validation of the delayed payout tx. In particular, the extra checks added in 345426f as part of #4789 (Fix remaining blackmail vulnerabilities) place a lower bound on the delayed payout tx input value and with it the deposit tx output value. However, explicitly checking the amount is more robust.

--

This PR is mainly to provide a more robust solution than that already included in the 1.5.0 release.


chimp1984 (approved):
utACK

sqrrm (approved):
utACK

PR #4861, opened 2020-11-27
install_java.sh: halt on errors

nylen:
This PR adds `set -e` to the script to install the correct Java environment.

Without this, when the call to the system package manager (to install `curl`) fails for whatever reason, the installation process blindly continues. It probably will just fail at each further step but this could also have unintended effects. The correct behavior is to abort the script when the first error occurs.


ripcurlx:
@devinbileck As it touches your code, could you please ACK this change?

devinbileck (approved):
utACK

sqrrm (approved):

sqrrm:
@nylen All commits need to be signed. Please make a new PR, referring this one with your commit signed, otherwise I cannot merge.

ripcurlx:
Closing as superseded by https://github.com/bisq-network/bisq/pull/4893.

Issue #4837, opened 2020-11-23
Fix various 'account management' usability issues in the Bisq client to optimize for traders with multiple fiat and altcoin accounts


pazza83:
### Description

When a user creates multiple fiat or altcoin accounts it creates usability problems. 

Bisq could make a the proposed changes below to allow traders to optimize their fiat or altcoin accounts for increased usability. 

This would increase usability, facilitating traders to add more accounts to the platform and and in turn has a positive impact on liquidity. The more multiple fiat or altcoin accounts someone has the more options they have to trade.

I am hoping these suggestions are all low hanging apples :apple::apple::apple:

#### Version

1.4.2

### Examples and suggestions

***

#### National currency account screen

![National currency account page](https://user-images.githubusercontent.com/66367547/99932444-fde14f80-2d4f-11eb-8de1-6ed40f9cfe00.png)

Switching being national currency accounts becomes unusable/difficult when the account selected has multiple lines of information.

`Suggestion - There is lots of space on the right. I would suggesting using a wo column layout (account info on the right).
`
***

#### Select trading account - in Buy / Sell A

![buy and sell](https://user-images.githubusercontent.com/66367547/99932567-5add0580-2d50-11eb-9e97-a3c07148ea23.png)

The trading account field is too short for payment account names.

`Suggestion - Treble the size of the field (you can see from the image below, even doubling it would be too short for Revolut)
`

***

#### Select trading account - in Buy / Sell B

![select trading account](https://user-images.githubusercontent.com/66367547/99932698-c1622380-2d50-11eb-90bb-91dcd20fa3b2.png)

If you have more than a scrollable window of account types this field becomes messy quickly.

`Suggestion - alphabetize list. It would also be good if the client discounted payment options based on selected currency. For example in the list above only two of the account types shown is able to make / take offers of the currency selected 
`

***

#### Manage accounts

![manage accounts](https://user-images.githubusercontent.com/66367547/99932891-5533ef80-2d51-11eb-8e11-5bbd85a0dcaf.png)

Fiat and altcoin accounts accounts get messy when more than a scrollable windows worth of accounts are added. 

`Suggestion - alphabetize list. I would also add 25% to the field width. You can see Revolut being cut off again.
`

***

#### Allow renaming of 'account name' field

![selected account](https://user-images.githubusercontent.com/66367547/99932979-95936d80-2d51-11eb-9c8c-61bb4ad85a2b.png)

[...]

PR #4863, opened 2020-11-27
Account management usability improvements

ghost:
Implements some requests from #4837 :

- Widen payment account dropdown in create & edit offer screens
- Sort payment account dropdown by name in create & edit offer screens
- Sort account listview by name in manage fiat accounts screen
- Sort account listview by name in manage altcoin accounts screen
- Bug fix: do not adjust BTC amount in edit offer screen

There was a bug where editing an existing offer could have its payment account changed to one with more restrictive limits and the offer would still be allowed (although not take-able). It should not be allowed to be saved in this case.  This change takes proper heed of the `allowAmountUpdate` flag which is false for offer edits.  
e.g. edit an F2F buy for 0.25, change to Zelle unsigned, the offer will prevent saving.  
Ref: https://github.com/bisq-network/bisq/issues/4837#issuecomment-734647423

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/account/content/altcoinaccounts/AltCoinAccountsDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/account/content/altcoinaccounts/AltCoinAccountsDataModel.java
@@ -43,6 +43,7 @@ import javafx.collections.ObservableList;
 import javafx.collections.SetChangeListener;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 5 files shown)


pazza83:
Thanks @jmacxx that was super quick  👍

ripcurlx:
@jmacxx It looks like existing tests are failing. Could you please have a look why and adapt it if it is a false positive? Thanks!

ghost:
Sorry I had got into a bad habit of building with `-x test`.  Fixed the issue in https://github.com/bisq-network/bisq/pull/4863/commits/0af139c666d5d243f12f27cd8ba58c735354b1c8

ripcurlx (approved):
utACK

PR #4875, opened 2020-12-02
Replace use of Collections to FXCollections in desktop.

### LINKED ISSUE #4388: UI: Replace java.util.Collections with FXCollections
As per a [JavaFX performance tuning thread](https://wiki.openjdk.java.net/display/OpenJFX/Performance+Tips+and+Tricks), the use of the `FXCollections` API is encouraged when optimizing for performance.

As per the [`FXCollections javadoc`](https://docs.oracle.com/javase/8/javafx/api/javafx/collections/FXCollections.html):

> Utility class that consists of static methods that are 1:1 copies of `java.util.Collections` methods.
>
> The utility methods are here mainly for performance reasons. All methods are optimized in a way that they yield only limited number of notifications. On the other hand, `java.util.Collections` methods might call "modification methods" on an `ObservableList` multiple times, resulting in a number of notifications.

Replace all usages of `java.util.Collections` with `javafx.collections.FXCollections` in the `bisq-desktop` project (the only one where JavaFX is used).

deusmax:
#\Replace use of java library java.util.Collections with
javafx.collections.FXCollections instead.
Fixes #4388 

Replace use of java.util.Collections, except for a few places that use Collections.singleton and Collections.nCopies which do not have direct replacements in FXCollections.
Run code and did not notice any issues.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/components/indicator/skin/StaticProgressIndicatorSkin.java
+++ b/desktop/src/main/java/bisq/desktop/components/indicator/skin/StaticProgressIndicatorSkin.java
@@ -75,8 +75,10 @@ import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.IntegerProperty;
 import javafx.beans.property.ObjectProperty;
 
+import javafx.collections.ObservableList;
+import javafx.collections.FXCollections;
+
 import java.util.ArrayList;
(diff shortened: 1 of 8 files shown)


ripcurlx (approved):
utACK

Code changes are looking fine :+1:

PR #4877, opened 2020-12-02
Fix warning about rejected transaction

### LINKED ISSUE #4876: Badly formatted warning message about rejected transaction
### Description

Transifex user Trigger67 has identified a small error in one English string for a warning message.
The string `popup.warning.trade.txRejected` contains an extra curly brace: `... Transaction ID={2}} ...`.
This has been propagated to almost all languages, only a few were fixed by translators.

#### Version

latest master: 09677151

### Steps to reproduce

Not reproduced with the actual application.

### Expected behaviour

The message produced should contain `... Transaction ID=0123456789abcdef\n ...`

### Actual behaviour

I assume the message would now look something like `... Transaction ID=0123456789abcdef}\n ...`

### Screenshots

N/A

#### Device or machine

Universal

#### Additional info

N/A

Jakub-CZ:
Fixes #4876 

My first PR here, so I'm starting small. I'm just fixing the English string `popup.warning.trade.txRejected` which contains an extra curly brace: `... Transaction ID={2}} ...`.

Most translations have inherited this error.


ripcurlx (approved):
utACK

Could catch and welcome to Bisq :+1:

PR #4883, opened 2020-12-03
Prevent that a wrong tx is set as deposit tx

### LINKED ISSUE #4873: Past payout tx used as deposit tx
### Description

We saw quite a few cases where the trade fails because a past payout tx is used as deposit tx in the trade.
We (credit goes to @jmacxx) could reproduce it now (still not easy as it only happens rarely). The reason is that we reset the address entry context for the Multisig address/key, and because those addresses are not detected in the check if an address is used they can be reused in a later trade. If both traders have repeated trades and by chance reuse the same address/key to create the MS address, the input address of a past payout tx is then the same address as the new output address of the deposit tx.
We use the address for listening for tx confidence and this can trigger then in the trade protocol that the wrong tx is interpreted and set a deposit tx.

To fix that we need to avoid resetting the address entry for Multisig. And to avoid that already past txs where the address entry was already reset and which could be reused any time again we also need to add some additional checks when we setup the confidence listener. We need to check if the input addresses are the 2 trade fee addresses and if the address we are looking up is used as output and not as input.

A PR is in preparation and I recommend that we release a hotfix based on 1.5.0 including only that PR and enforce an update after a few days when most users have updated to avoid furture issues from that bug. 

@oscarguindzberg  I do not think that issue is related to https://github.com/bisq-network/bisq/issues/4870 but also not 100% sure.

chimp1984:
Fixes #4873

Beside what is mentioned in #4873 I found another (minor) issue. We use an immutable list for address entries but the coinLockedInMultiSig value was set in the object via a setter. This change would be lost when accessed as when one open the wallet tool with cmd+j (after take offe the value was 0 there instead of the locked up funds). As the field is only used for lookup for locked funds from failed trades the bug was not very visible.
coinLockedInMultiSig is now a final field set by the constructor.

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqSetup.java
+++ b/core/src/main/java/bisq/core/app/BisqSetup.java
@@ -467,7 +467,7 @@ public class BisqSetup {
                     .filter(e -> setOfAllTradeIds.contains(e.getOfferId()) &&
                             e.getContext() == AddressEntry.Context.MULTI_SIG)
                     .forEach(e -> {
-                        Coin balance = e.getCoinLockedInMultiSig();
+                        Coin balance = e.getCoinLockedInMultiSigAsCoin();
                         if (balance.isPositive()) {
                             String message = Res.get("popup.warning.lockedUpFunds",
                                     formatter.formatCoinWithCode(balance), e.getAddressString(), e.getOfferId());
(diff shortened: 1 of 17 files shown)


ghost:
ACK (re-tested). :ballot_box_with_check:

ghost (approved):

ripcurlx (approved):
utACK

PR #4884, opened 2020-12-03
Refactor: move getDust(Transaction transaction)

ghubstan:
The api is going to need this method, so it was moved from desktop to core.

- Moved the method from `WithdrawalView` to `WalletService`.

- Removed an unused initializer in `WithdrawalView`.


sqrrm (approved):
utACK

Issue #4864, opened 2020-11-28
Unable to proceed in trade  BUG


f321x:
### Description
 I cant proceed in my trade because it doesnt accept when i press "Payment made". It suddenly switches between trading step 2 and 3 and doesnt stay. I think the counterparty (seller) doesnt see the trade as payment done and we arent able to process in the trade. Did SPV resync but it didnt help.

#### Version
1.5.0

### Steps to reproduce
Do sepa trade, press "Payment made" and it switches randomly between step 2 and 3 and doesnt stay. Also when i restart bisq it is always back at step 2 until i press "payment made" several times then it switches between 2 and 3 again.

### Expected behaviour
I expected it to go to step 3 so the seller can send me the funds i bought of him.

### Actual behaviour
Switches back and forth between step 3 and 2 and we cant complete the trade

### Screenshots
![Screenshot1](https://user-images.githubusercontent.com/51097237/100517081-89ac0f00-3188-11eb-8a1a-99495f2e819d.png)
![Screenshot2](https://user-images.githubusercontent.com/51097237/100517085-8ca6ff80-3188-11eb-82d2-8c464c70d53f.png)

#### Device or machine
Lenovo Laptop with Ubuntu 20.04.1 LTS, Core i7, 8gb ram, 1gbit/s internet connection

#### Additional info
Can provide log file if it helps

ghost:
Thanks for reporting this.  Logfiles are always helpful and appreciated.  Alternatively you could open terminal and use the following command to watch the log file, then press Payment made and see what errors are reported.

    tail -n 100 -f ~/.local/share/Bisq/bisq.log

f321x:
[bisq.log](https://github.com/bisq-network/bisq/files/5610941/bisq.log)
[bisq_1.log](https://github.com/bisq-network/bisq/files/5610942/bisq_1.log)
[bisq_2.log](https://github.com/bisq-network/bisq/files/5610944/bisq_2.log)
[bisq_3.log](https://github.com/bisq-network/bisq/files/5610945/bisq_3.log)

Hey, here are the logfiles out of an new bisq backup, hope they help in finding the issue

ghost:
[...]

PR #4885, opened 2020-12-03
Avoid that buyer switches back to trade step 2 at startup after payment confirmed

chimp1984:
Fixes https://github.com/bisq-network/bisq/issues/4864

When seller if offline we resend the CounterCurrencyTransferStartedMessage at startup.
That caused the trade state set to BUYER_SENT_FIAT_PAYMENT_INITIATED_MSG and then after
the msg was stored in mailbox to BUYER_STORED_IN_MAILBOX_FIAT_PAYMENT_INITIATED_MSG.
Those 2 msg trigger diff. UI states which led to the UI glitch that the UI moved to
step 2 and then to step 3 which was correct but confusing to the user.

Now we only apply BUYER_SENT_FIAT_PAYMENT_INITIATED_MSG is trade state ordinal is
smaller avoiding that UI glitch.

This is not a critical bug fix so open to maintainers to add it to 1.5.1. But I think its also very low risk and easy to review and test.

### DIFF
--- a/core/src/main/java/bisq/core/trade/protocol/tasks/buyer/BuyerSendCounterCurrencyTransferStartedMessage.java
+++ b/core/src/main/java/bisq/core/trade/protocol/tasks/buyer/BuyerSendCounterCurrencyTransferStartedMessage.java
@@ -82,7 +82,9 @@ public class BuyerSendCounterCurrencyTransferStartedMessage extends SendMailboxM
 
     @Override
     protected void setStateSent() {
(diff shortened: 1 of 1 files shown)


chimp1984:
How to test it?

- Create a trade and close seller after take offer. 
- Create a new block, buyer gets to step 2
- Buyer confirms payment sent. Msg is stored in mailbox, buyer is at step 3
- Restart buyer and leave screen at open trades. You will see at start that it starts with step2, sends the msg again and then goes to step 3. This automatic switch from step 2 to step 3 is confusing to users and this PR fixes that.

ripcurlx (approved):
ACK

Fixes the problem for me on Regtest

Issue #4879, opened 2020-12-02
Missing Taker fee tx and ghost deposit tx


Mycelial1:
### Description
I am reporting my specific issue with this new bug that other users have been experiencing. 

Upon startup, I get the java.lang.NullPointerException. 

Trade id lwujzg was taken once, and resulted in a "missing taker fee transaction". I thought it was my spotty internet connection not sending messages because upon further inspection it said "The trade protocol did not finish in 30 seconds"

![first fail](https://user-images.githubusercontent.com/34950746/100937297-d2353680-34c0-11eb-8a99-aac8799b0144.png)
![trade protocl did not finish in 30 secs](https://user-images.githubusercontent.com/34950746/100939113-c4cd7b80-34c3-11eb-9c68-0f4bcbebb12b.png)

The same offer was kept on the order book and taken by someone else and the deposit transaction was going in and out of the mempool. I checked today and turns out that it's confirmed with about 42 confirmations, but the trade prompt still says "waiting on blockchain confirmation". I do not have icons  for chat, or to move to failed trade but the UI says that there are coins "locked". Those locked coins are for this trade because I don't have any other open trades.
![noicons](https://user-images.githubusercontent.com/34950746/100938262-6358dd00-34c2-11eb-87ae-09da2690ea7a.png)

#### Version
1.5.0

### Steps to reproduce

I was BSQ maker as seller. 

#### Additional info

here is a snippet from my logs

################################################################# 
Dec-02 06:00:23.352 [JavaFX Application Thread] INFO  b.n.p.s.P2PDataStorage: Processing 1262 protectedStorageEntries took 147 ms. 
Dec-02 06:00:24.639 [JavaFX Application Thread] INFO  b.n.p.s.P2PDataStorage: Processing 79 persistableNetworkPayloads took 1286 ms. 
Dec-02 06:00:24.871 [JavaFX Application Thread] INFO  b.c.t.p.FluentProtocol$Condition: We received a STARTUP event at phase TAKER_FEE_PUBLISHED and state MAKER_SAW_ARRIVED_PUBLISH_DEPOSIT_TX_REQUEST, tradeId=lwujzg-558246d8-5c6c-4358-8ea4-be39e6db2081-139 
Dec-02 06:00:24.885 [JavaFX Application Thread] INFO  b.c.t.TaskRunner: Run task: BuyerSetupDepositTxListener 
[...]

PR #4889, opened 2020-12-03
Avoid case of multi trades with same offer

chimp1984:
Fixes https://github.com/bisq-network/bisq/issues/4879

We had removed the offer after the first trade task. If it failed there the offer stayed online and could be taken by another user. This caused the issue that one failed trade was in the trade list and then later a valid trade entered as well. As we used the offerId as key for a map which was used for the processModel lookup, that caused a bug with a Nullpointer exception.

This PR fixes that problem by using a uid instead of the offerId to make the map lookup more solid.
It also moves the task to remove the offer as the very first task so it cannot happen anymore that a offer stays online and leads later to 2 trades with the same offer id.

It also increased the realtively short timeout of 30 seconds in the some trade protocol task runners. We saw some cases where those timeouts triggered errorMessages but the trade could complete without problem, so that should help to reduce such cases. We do not fail on those timeouts, they only trigger an error message. The general trade protocol timeout would cause a failure of the trade.

### DIFF
--- a/core/src/main/java/bisq/core/trade/TradeManager.java
+++ b/core/src/main/java/bisq/core/trade/TradeManager.java
@@ -89,6 +89,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -259,7 +260,8 @@ public class TradeManager implements PersistedDataHost, DecryptedDirectMessageLi
                     openOffer.getMediatorNodeAddress(),
                     openOffer.getRefundAgentNodeAddress(),
                     btcWalletService,
-                    getNewProcessModel(offer));
+                    getNewProcessModel(offer),
+                    UUID.randomUUID().toString());
         } else {
             trade = new SellerAsMakerTrade(offer,
                     Coin.valueOf(inputsForDepositTxRequest.getTxFee()),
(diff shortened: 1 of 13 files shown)


core/src/main/java/bisq/core/trade/protocol/BuyerAsMakerProtocol.java
```diff
@@ -100,8 +100,8 @@ protected void handle(DelayedPayoutTxSignatureRequest message, NodeAddress peer)
                 .with(message)
                 .from(peer))
                 .setup(tasks(
-                        BuyerProcessDelayedPayoutTxSignatureRequest.class,
                         MakerRemovesOpenOffer.class,
+                        BuyerProcessDelayedPayoutTxSignatureRequest.class,
```
sqrrm, line 104: Could this be abused by sending a bad request, causing offers to be removed but guaranteed failed trade?

core/src/main/java/bisq/core/trade/TradeManager.java
```diff
@@ -269,10 +271,15 @@ private void handleTakeOfferRequest(NodeAddress peer, InputsForDepositTxRequest
...
                     btcWalletService,
-                    getNewProcessModel(offer));
+                    getNewProcessModel(offer),
+                    UUID.randomUUID().toString());
         }
         TradeProtocol tradeProtocol = TradeProtocolFactory.getNewTradeProtocol(trade);
-        tradeProtocolByTradeId.put(trade.getId(), tradeProtocol);
+        TradeProtocol prev = tradeProtocolByTradeId.put(trade.getUid(), tradeProtocol);
+        if (prev != null) {
+            log.error("We had already an entry with uid {}", trade.getUid());
```
sqrrm, line 280: How would this happen? Collision of randomUUID?

sqrrm (approved):
utACK

core/src/main/java/bisq/core/trade/TradeManager.java
chimp1984: If for some reason the same trade is entering those methods. Should not be possible but prefer to keep the log to detect it if there is still some bug.

core/src/main/java/bisq/core/trade/protocol/BuyerAsMakerProtocol.java
chimp1984: Was thinking the same, but a taker could do it anyway at any point later as well. We could add some feature to restore offers which got closed from a trade failure by cloning  the offer and change the id. So that would render such an attack as ineffective.

sqrrm: Yeah, it's an area that needs some more thought and this is not the real issue.

Issue #4856, opened 2020-11-26
add max log size field to Account >> Backup


Bisq-knight:
### Description

Not a bug, but a suggestion.

We often run into problems that occurred multiple days back and the bisq logs seem to be insufficient to investigate the moment of some reported problem. I understand that many people would not be fond of keeping +500mb of logs but I do think there are some people who would be fine with it.

With that in mind I'd like to propose that we put a field in Account >> Backup with the max size of the logs you would allow.

Currently it is about 100mb  (~10  files)

chimp1984:
See https://github.com/bisq-network/bisq/pull/4891

chimp1984:
It is set to 20 now. I am not sure if adding it as user preference is good as users might set it very low and then in case of bugs we don't have enough data to figure out the issue. Also as the logs are set up very early in setup its not as trivial as other preferences, but require prob. to do it via properties file.

cd2357:
Looks like the increase to 20 log files is a good compromise.

PR #4891, opened 2020-12-03
Increase max number of log files to 20.

chimp1984:
Fixes https://github.com/bisq-network/bisq/issues/4856

This is a quick fix for the request at https://github.com/bisq-network/bisq/issues/4856.
I only increases statically the numbe rof log files from 10 to 20. As we have too many bugs atm I think that helps to find the issues and can be reverted later once not needed anymore, and/or implemented with a setting in the preferences. As log is setup very early in the setup I am not sure that will be that easy... the propery file need to be used probably. 
We should use a high value as default, as otherwise once a bug happen it might not help if it is older and already removed.

No hard need to get into 1.5.1 but as its that low risk and have be helpful for future debugging I recommend to get it in as well.


ripcurlx (approved):
utACK

PR #4892, opened 2020-12-03
Offer Book chart outlier filter improved.

### LINKED ISSUE #4871: Outlier filter for Market Offer Chart  removes body of offers.
### Description

Offer Book filtering is used to cut display of excessive value offers. Chart display is affected only, the offers continue to exist in the offer table below. Excessive values are often seen, especially for BSQ trades.

The case is observed, for this EUR offers instant, the filter is cutting out the main body of offers and displays the outlier.
i.e. the opposite of the desired behavior.
See screenshot.

Other markets displaying OK.

#### Version

v1.5.0

### Steps to reproduce

Depends on market offers outlier values.

### Expected behavior

Should remove outlier

### Actual behavior

Removes main body of Sell BTC chart display

### Screenshots

Bisq-MarketOfferBook-bug-EUR-lowValue-2020-12-02

#### Additional info

@chimp1984 
Related to PR#4774

deusmax:
Outlier data filter used for the Market Offer Book chart, now accounts
for the switching display position of Buy, Sell data depending on
currency type (Crypto or Fiat).

Fixes #4871  

Some of the Market Offers have values far from the main trading body (outliers). Including these points in a chart destroys any significant and useful information presented in the graph. An outlier filter, applied during the production of these charts, is used to remove such points. 

The implementation of this filter did not take into account the chart differences between fiat and crypto currencies. The one in this PR does.
The recently increasing Fiat value of BTC, coupled with extreme outlier values, has exposed the limitations. This PR provides a filter that gives better results and should be much more stable. 

See issue #4871 for before & after screenshots.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java
+++ b/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java
@@ -127,7 +127,7 @@ public class OfferBookChartView extends ActivatableViewAndModel<VBox, OfferBookC
     private HBox bottomHBox;
     private ListChangeListener<OfferBookListItem> changeListener;
     private ListChangeListener<CurrencyListItem> currencyListItemsListener;
-    private final double chartDataFactor = 3;
+    private final double dataLimitFactor = 3;
     private final double initialOfferTableViewHeight = 121;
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
ACK

Fixes the issue as expected e.g. for the CAD market right now

**Before:**
Bildschirmfoto 2020-12-04 um 16 53 03

**After:**
Bildschirmfoto 2020-12-04 um 16 53 11

PR #4896, opened 2020-12-04
Refactor BtcWalletService to let api override fee rates

ghubstan:
`BtcWalletService` was changed to allow the api to override tx fee rates from the `sendbsq` and `sendbtc` methods.  The api methods will still be able to use the network fee service and custom tx fee rate preference, and set / unset the custom tx fee rate preference, but the change will permit the addition of an optional `txFeeRate` parameter to the `sendbsq` and `sendbtc` methods (todo).  A few other minor changes (style and removal of never thrown ex spec) were also made to this class.

Two `BtcWalletService` methods were refactored.

- The redundant (was always true) `boolean isSendTx` argument was removed from the `completePreparedVoteRevealTx` method signature.
  
- The redundant (was always true) `boolean useCustomTxFee` was removed from the `completePreparedBsqTx` method signature.

- The `completePreparedSendBsqTx` method was overloaded with a 2nd parameter (`Coin txFeePerVbyte`) to allow api to override fee service and custom tx fee rate when sending BSQ or BTC.

- The `completePreparedBsqTx` method was overloaded with a 3rd parameter (`Coin txFeePerVbyte`) to allow api to override fee service and custom tx fee rate when sending BSQ or BTC.

The following line was deleted from the `completePreparedBsqTx` method because `txFeePerVbyte` is now an argument:

	`Coin txFeePerVbyte = useCustomTxFee ? getTxFeeForWithdrawalPerVbyte() : feeService.getTxFeePerVbyte();`
	
This `useCustomTxFee` value was always true, and redundant here because  `getTxFeeForWithdrawalPerVbyte()` returns `feeService.getTxFeePerVbyte()` or the custom fee rate preference. i.e.,

`Coin txFeePerVbyte = useCustomTxFee ? getTxFeeForWithdrawalPerVbyte() : feeService.getTxFeePerVbyte();`
is equivalent to	
`Coin txFeePerVbyte = getTxFeeForWithdrawalPerVbyte();`

`LockupTxService`, `UnlockTxService`, `BsqSendView`, and `BsqTransferService` were adjusted to the `BtcWalletService` refactoring.

This is the 2nd in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4884.
PR https://github.com/bisq-network/bisq/pull/4884 should be reviewed before this one.


ghubstan:
force travis build after jar download failed

core/src/main/java/bisq/core/dao/governance/bond/lockup/LockupTxService.java
```diff
@@ -103,7 +103,7 @@ private Transaction getLockupTx(Coin lockupAmount, int lockTime, LockupReason lo
             throws InsufficientMoneyException, WalletException, TransactionVerificationException, IOException {
         byte[] opReturnData = BondConsensus.getLockupOpReturnData(lockTime, lockupReason, hash);
         Transaction preparedTx = bsqWalletService.getPreparedLockupTx(lockupAmount);
-        Transaction txWithBtcFee = btcWalletService.completePreparedBsqTx(preparedTx, true, opReturnData);
+        Transaction txWithBtcFee = btcWalletService.completePreparedBsqTx(preparedTx, opReturnData);
```
sqrrm, line 106: Unrelated to the changes in this PR, but why are we using custom fee for a DAO operation? This is the lockup tx, should probably use the fee from feeService.

core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
```diff
@@ -487,8 +499,6 @@ public Transaction completePreparedBsqTx(Transaction preparedBsqTx,
         int sigSizePerInput = 106;
         // typical size for a tx with 2 inputs
         int txVsizeWithUnsignedInputs = 203;
-        // If useCustomTxFee we allow overriding the estimated fee from preferences
-        Coin txFeePerVbyte = useCustomTxFee ? getTxFeeForWithdrawalPerVbyte() : feeService.getTxFeePerVbyte();
```
sqrrm, line 491: Looks like custom fee was always used when calling this method anyway.

sqrrm (commented):
I think the changes look ok.

Needs review from @ManfredKarrer

core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
ghubstan: [edited some mistakes in the comment]

When `useCustomTxFee == true` `&&` `preferences.isUseCustomWithdrawalTxFee() == false`,  the fee service rate was used.   (That's a bug.)

`useCustomTxFee == false`  was not used

chimp1984: The name of useCustomTxFee was badly choosen. The intention was that for certain important transactions we want to use always the fee service (like in trade txs). I think in some BSQ txs that was the case earlier as well but got changed later as BSQ txs do not cause harm to other peers in case the fee was too low set.

So intention was to have an option to enforce usage of the fee service independent of what user has set in preferences. The false case would have done that. The true case will use whatever the user has set in preferences.

core/src/main/java/bisq/core/dao/governance/bond/lockup/LockupTxService.java
chimp1984: It is the users responsibility and does not cause harm to others if he set a too low fee and need to resynce/redo the tx. I guess it was set to fee service rate initially but later changed... But yes, there might be arguments pro and con...

chimp1984 (approved):
utACK

ManfredKarrer (approved):
utACK

ghubstan:

PR #4897, opened 2020-12-04
Allow take offer in case account age witness is not found and trade amount is small

chimp1984:
In case the peers witness is not found and the trade amount 
is not exceeding TOLERATED_SMALL_TRADE_AMOUNT (0.01 BTC) we
return true for verifyPeersTradeAmount.
Before such offers could not be taken even the witness
check would be irrelevant as the trade amount is below the
threshold where we require account age witness.


sqrrm (approved):
utACK

PR #4898, opened 2020-12-04
Dont shut down seed in regtest mode if all connections lost

chimp1984:
For dev testing (usually on BTC_REGTEST) we don't want to get the seed shut down as it is normal that the seed is the only actively running node.


sqrrm (approved):
utACK

PR #4899, opened 2020-12-04
Privacy improvements for manual payout

### LINKED ISSUE #4061: Stop requesting traders private keys for Manual payout
### Description

At present, in the rare cases Mediators need to create manual payout transactions to release the funds from the multisig, they need to request the private keys of both traders and sign the transaction for them. This is a terrible violation of one of the core Bitcoin principles.

### Expected behaviour

Once a trader opens mediation the mediator gets access to the trade's contract with all the necessary information to create the transaction. They should do so and share the unsigned transaction to the traders so they can sign and broadcast it.

### Actual behaviour

The mediator creates. signs the transaction for both traders and broadcasts it.

### Screenshots

emergencyMultisigPayoutTool

#### Additional info

I've created a guide for an [alternative method using tools outside of Bisq](https://bisq.wiki/Manual_payout). It'd be great if we could add the following functions to the Bisq software:
1. verify and sign message with a specific address
2. verify and sign a transaction
3. encrypt and decrypt message with a specific address

It's imperative that the sign, encrypt and decrypt functions are done inside the Bisq software due to the need of private key. Electrum and Coinbin's UI may be worth copying.

Related
[Legacy Manual payout process](https://docs.bisq.network/manual-dispute-payout.html)
Update to new protocol - https://github.com/bisq-network/bisq/pull/3694

ping @bisq-network/mediators @wiz @Bisq-knight

ghost:
The goal of this PR is to improve privacy, by not requiring mediators to ask for users private keys.

- Redesign the UI
- Add import/export of payout settings
- Mediator does not need private key
- User can sign using own wallet or private key
- Validation of input fields
- Calculate the tx fee based on inputs
- Display of the generated txid & hex so it can be checked

The emergency multisig payout tool has been redesigned so that the unsigned payout transaction can be built, exported to the users for them to sign, and then the two signatures applied by the mediator before broadcasting the payout.

The screen is split into tabs (or a menu) allowing choices for Inputs, Import/Export, Sign and Build.  See screenshots below.

Inputs would be filled out by the mediator.  Most of the info can be obtained from the trade's contract.  The amountInMultisig can be obtained from checking an explorer - amountInMultisig is necessary because this value is part of the hashed signature (in segwit), so the TX will not be valid unless this amount is specified correctly.

**update 2020-12-14 : added the option to import data from a mediation ticket (see last screenshot)**

The mediator would then export the settings and paste the exported string to both users in mediation chat.  e.g.

    c2Vnd2l0OmE5MGJlY2Q3NGFlY2Y3ZGNhYzcyYzYyY2U3ZjRlZGMzNmNhYjQ0MTEwNmMxZWNhNDgzMDFlZjNiNGU1MmRhN2Y6MC4wMzIzMzM6MC4wMjY6MC4wMDY6YmNydDFxbXR0bjVwMnMwdnVjdnRwenMwd2oydXJtbW16bjhrMGRmMzBuZjU6YmNydDFxNTcwa2wzNGNzdndrbGt6MDQwNTk1ejg2MHRjd3lwemprdnB2czA6MDJiYTI5YmE0MWY2MDhmZWE4YWI3ZTQ2NzRkNDRjYWYxY2NiNjg1NDVjNzlmNjUxNGZiOTc4ZjNlZGVlZGZhZDQyOjAzMmU3M2RmOGQyYTEzNGYwZGY0MTcxZjU5ZjNkNDMxNjExNmZmMDAwYzViN2IyYzdiOGNiYWMyYWUzYjliODNjMw==

Users would open the tool (ctrl+g) click import and paste in the supplied string.  Then they would click sign and press "LOCATE KEY IN WALLET" followed by "GENERATE SIGNATURE".  Then they would paste the signature string back to the mediator.

(The mediator would give the user some guidance on how to open the tool and sign).

The mediator, upon receiving the two signature strings would click "BUILD" and paste the buyer and seller signatures into the fields, followed by "BUILD" and/or "BROADCAST".  The txId and txHex are displayed in a text box (similar to the UI of coinbin), so the txId could by copied into an explorer to check the TX status.

Entering the parameters:
image

Exporting:
image

Signing (by users):
image

Building:
image

Update 2020-12-14 added ability to import data from mediation ticket:
image

Fixes #4061 

@huey735, mediators are invited to review/comment on this proposed solution.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/overlays/windows/ManualPayoutTxWindow.java
+++ b/desktop/src/main/java/bisq/desktop/main/overlays/windows/ManualPayoutTxWindow.java
@@ -528,13 +528,21 @@ public class ManualPayoutTxWindow extends Overlay<ManualPayoutTxWindow> {
                 sellerSignatureAsHex.getText().length() > 0);
     }
 
+    private Coin getInputFieldAsCoin(InputTextField inputTextField) {
+        try {
+            return Coin.parseCoin(inputTextField.getText());
+        } catch (RuntimeException ignore) {
+        }
+        return Coin.ZERO;
+    }
+
     private void calculateTxFee() {
         if (buyerPayoutAmount.getText().length() > 0 &&
                 sellerPayoutAmount.getText().length() > 0 &&
                 amountInMultisig.getText().length() > 0) {
-            Coin txFeeValue = Coin.parseCoin(amountInMultisig.getText())
-                    .subtract(Coin.parseCoin(buyerPayoutAmount.getText()))
-                    .subtract(Coin.parseCoin(sellerPayoutAmount.getText()));
+            Coin txFeeValue = getInputFieldAsCoin(amountInMultisig)
+                    .subtract(getInputFieldAsCoin(buyerPayoutAmount))
+                    .subtract(getInputFieldAsCoin(sellerPayoutAmount));
             txFee.setText(txFeeValue.toPlainString());
         }
     }
@@ -654,9 +662,9 @@ public class ManualPayoutTxWindow extends Overlay<ManualPayoutTxWindow> {
         String retVal = "";
         try {
             Tuple2<String, String> combined = tradeWalletService.emergencyBuildPayoutTxFrom2of2MultiSig(depositTxHex.getText(),
-                    Coin.parseCoin(buyerPayoutAmount.getText()),
-                    Coin.parseCoin(sellerPayoutAmount.getText()),
-                    Coin.parseCoin(txFee.getText()),
+                    getInputFieldAsCoin(buyerPayoutAmount),
+                    getInputFieldAsCoin(sellerPayoutAmount),
+                    getInputFieldAsCoin(txFee),
                     buyerAddressString.getText(),
                     sellerAddressString.getText(),
                     buyerPubKeyAsHex.getText(),
@@ -667,7 +675,7 @@ public class ManualPayoutTxWindow extends Overlay<ManualPayoutTxWindow> {
             retVal = tradeWalletService.emergencyGenerateSignature(
                     unsignedTxHex,
                     redeemScriptHex,
-                    Coin.parseCoin(amountInMultisig.getText()),
+                    getInputFieldAsCoin(amountInMultisig),
                     privateKeyHex.getText());
         } catch (IllegalArgumentException ee) {
             log.error(ee.toString());
@@ -687,9 +695,9 @@ public class ManualPayoutTxWindow extends Overlay<ManualPayoutTxWindow> {
             try {
                 // grab data from the inputs pane, build an unsigned tx and write it to the TextArea
                 Tuple2<String, String> combined = tradeWalletService.emergencyBuildPayoutTxFrom2of2MultiSig(depositTxHex.getText(),
-                        Coin.parseCoin(buyerPayoutAmount.getText()),
-                        Coin.parseCoin(sellerPayoutAmount.getText()),
-                        Coin.parseCoin(txFee.getText()),
+                        getInputFieldAsCoin(buyerPayoutAmount),
+                        getInputFieldAsCoin(sellerPayoutAmount),
+                        getInputFieldAsCoin(txFee),
                         buyerAddressString.getText(),
                         sellerAddressString.getText(),
                         buyerPubKeyAsHex.getText(),
--- a/desktop/src/main/java/bisq/desktop/main/overlays/windows/ManualPayoutTxWindow.java
+++ b/desktop/src/main/java/bisq/desktop/main/overlays/windows/ManualPayoutTxWindow.java
@@ -17,10 +17,13 @@
 
 package bisq.desktop.main.overlays.windows;
 
+import bisq.desktop.components.AutoTooltipButton;
+import bisq.desktop.components.BisqTextArea;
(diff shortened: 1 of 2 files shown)


Bisq-knight:
Sounds good! I'm gonna try it out with a case I currently have that requires that. So the steps are:
1. Mediator inputs the payout details
2. Mediator exports the unsigned tx
3. Mediator sends both users the transaction to sign, which
4. user opens the tool (also with ctrl+g) 
5. user signs 
6. user sends mediator the signature
7. Mediaor combines the two signatures
8. Mediator broadcasts the tx

ghost:
Yes, those are the steps.  I'm here & on keybase if you have any issues.

Bisq-knight:
Like I said in our discussion in Keybase, this does look very promising and certainly would make users' lives easier as they do feel a like intimidated when they read the tutorial on how to get the private keys. 

Plus it removes our reliance on Coinb.in as part of the current ["private" process](https://bisq.wiki/Manual_payout), a service that could be taken down at any time

chimp1984:
Thanks @jmacxx for your initiative!

Maybe we can combine that for https://github.com/bisq-network/proposals/issues/287 ?

E.g. If we make it a bit more automated then the users could use that tool to make an alternative payout with arbitrary outputs and fees so it can serve both the idea to boost stuck tx chains by CPFP and to be used as emergency tool if normal payout options do not work. Of course adding too much complexity to an emergency tool is not a good idea...

ghost:
> E.g. If we make it a bit more automated then the users could use that tool to make an alternative payout with arbitrary outputs and fees so it can serve both the idea to boost stuck tx chains by CPFP and to be used as emergency tool if normal payout options do not work. Of course adding too much complexity to an emergency tool is not a good idea...

@chimp1984 Its a good idea and I'd like to take some time to think about a UI design for it.

chimp1984:
> > E.g. If we make it a bit more automated then the users could use that tool to make an alternative payout with arbitrary outputs and fees so it can serve both the idea to boost stuck tx chains by CPFP and to be used as emergency tool if normal payout options do not work. Of course adding too much complexity to an emergency tool is not a good idea...
> 
> @chimp1984 Its a good idea and I'd like to take some time to think about a UI design for it.

Actually after more thought, I found that it might be ok to do it manually. It is exceptional anyway and the manual process via mediator brings some safety. Otherwise there might be risky situations (e.g. buyer sending signature but then later it gets confirmed and he starts the payment, but seller takes signature and gets back his trade amount with it...). I started a while back a "cancel trade" sub-protocol and it turned out that its more tricky and risky as it seemed first. Mainly because it introduced a parallel execution stream and that can cause problems.

So if your tool supports to change the miner fee, it would fulfill the basic requirements for a CPFP. The calculation how much fee is required can also be done manually by a explorer look up. I think its not worth the effort for that edge case.... Just would be good that we have a tool at hand to un-block stuck txs in times when the fees spike and sustain high.

chimp1984 (approved):
utACK

Will test it tomorrow... Would be good to make the fee editable as well so it can be used for CPFP use case as well.

ghost:
To adjust the fee, adjust one or both payout amounts.

The fee adjusts to the leftover.
No matter what is entered, the fee is `input - output1 - output2`.
e.g.

    INPUT = 0.02202  OUTPUTS 0.00600 0.01600 FEE 0.000020
    INPUT = 0.02202  OUTPUTS 0.00599 0.01590 FEE 0.000040

desktop/src/main/java/bisq/desktop/main/overlays/windows/ManualPayoutTxWindow.java
```diff
@@ -100,94 +198,549 @@ protected void setupKeyHandler(Scene scene) {
...
-                UserThread.execute(() -> {
-                    String txId = result != null ? result.getTxId().toString() : "null";
-                    new Popup().information("Transaction successful published. Transaction ID: " + txId).show();
-                });
-            }
+    private void calculateTxFee() {
+        if (buyerPayoutAmount.getText().length() > 0 &&
+                sellerPayoutAmount.getText().length() > 0 &&
+                amountInMultisig.getText().length() > 0) {
+            Coin txFeeValue = Coin.parseCoin(amountInMultisig.getText())
```
chimp1984: Coin.parseCoin can throw exceptions if input format is not as expected. Best to add a try catch...

chimp1984 (dismissed):
ACK
Tested now and worked all! Great tool!
Small nit with catching invalid inputs at parseCoin...

ripcurlx:
Just re-triggered Travis, seems to be quite unreliable recently. I'll cherry-pick this into v1.5.2.

desktop/src/main/java/bisq/desktop/main/overlays/windows/ManualPayoutTxWindow.java
ripcurlx: I'll wait for this before I cherry-pick.

ghost: Resolved by https://github.com/bisq-network/bisq/pull/4899/commits/e888be319256ef1fd5a429a367274df35fed21b2

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/4899#pullrequestreview-553930349

PR #4902, opened 2020-12-05
Refactor outlier filtering for Market Offer chart.

deusmax:
Moved the outlier filtering outside of the updateChartData function.

This PR refactors code from PR#4892 for filtering outlier values in Market Offer chart.
The outlier data filtering is moved outside of the updateChartData function.
Each step is simplified and cleaner.


ripcurlx (approved):
utACK

PR #4907, opened 2020-12-07
Prevent that we write data at shutdown before we have read the data.

### LINKED ISSUE #4844: My offers (open offers, trade, history) are not displayed.
### Description

After Bisq start, there are no offers at open offers, open trades or history. I did not force close Bisq, but performed regular and small update to OS before shutting down laptop. Available funds are displayed, but there's no locked or reserved funds.

#### Version

v1.4.2
Tested after upgrade to v1.5 without results.

### Expected behaviour

Display open offers, and two ongoing trades I have.

#### Device or machine

OpenSUSE Tumbleweed 20201121

#### Additional info

Logs are shared at Keybase, this issue is opened for register.

chimp1984:
Fixes https://github.com/bisq-network/bisq/issues/4844


ripcurlx (approved):
utACK

PR #4909, opened 2020-12-07
Fix bsq to fiat calculation

### LINKED ISSUE #4865: BSQ trading fee in fiat not calculated properly
### Description

When placing or taking an offer, BSQ paid in EUR displayed is overestimated. It should use the [30 day average](https://github.com/bisq-network/bisq/pull/4781) but seems like something is failing.

#### Version

v1.5

### Steps to reproduce

Try taking or making an offer.

### Expected behaviour

Use 30 day average to give a value approximate to half of BTC trading fee in fiat.

### Actual behaviour

It gives a EUR value 5 times higher than paying in BTC.

### Screenshots

imagen

### Additional info

I checked 30 day average and it's normal (5122 sats/BSQ, 0.86 USD/BSQ) so that is not missing. Maybe it still taking the info from only the last trade?

chimp1984:
Fixes https://github.com/bisq-network/bisq/issues/4865

We used last trade price of BSQ for calculating the BSQ fee as fiat currency. This could distort the real amount in case the last trade had a price very different to average price.
We use now the 30 day average BSQ price.

### DIFF
--- a/core/src/main/java/bisq/core/offer/OfferUtil.java
+++ b/core/src/main/java/bisq/core/offer/OfferUtil.java
@@ -292,10 +292,14 @@ public class OfferUtil {
     public Optional<Volume> getFeeInUserFiatCurrency(Coin makerFee,
                                                      boolean isCurrencyForMakerFeeBtc,
                                                      CoinFormatter bsqFormatter) {
-        // We use the users currency derived from his selected country.  We don't use the
-        // preferredTradeCurrency from preferences as that can be also set to an altcoin.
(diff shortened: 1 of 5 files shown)


chimp1984:
As there was a recent BSQ trade with quite a higher price than the 30 days average one can see the difference from the release version to that PR version in the screenshots below.

Screen Shot 2020-12-07 at 13 56 37
Screen Shot 2020-12-07 at 13 56 25

ripcurlx (approved):
utACK

PR #4912, opened 2020-12-08
Add "Raw deposit transaction as hex" to details window

chimp1984:
Add "Raw deposit transaction as hex" to details window

Rename Contract as json button to detail data


sqrrm (approved):
utACK

PR #4914, opened 2020-12-08
UI improvements

chimp1984:
Small UI improvements
- Make altcoin account screen 2 colums wide (same as fiat accounts)
- Increase text area height
- increase number of visible rows in accounts for better scolling

Cleanups
- Remove old `buyerSecurityDepositNews0.9.5` flag
- Remove unused method
- Make field local


ripcurlx (approved):
utACK

PR #4915, opened 2020-12-08
Merge 1.5.1 to master

chimp1984:
@ripcurlx This merges 1.5.1. to master but not updating the SNAPSHOT postfix of the version. I leave that for you as I am not sure where it needs to be updated.
Is that really required anyway?


ripcurlx:
Superseded by https://github.com/bisq-network/bisq/pull/4918.

ripcurlx:
> @ripcurlx This merges 1.5.1. to master but not updating the SNAPSHOT postfix of the version. I leave that for you as I am not sure where it needs to be updated.
> Is that really required anyway?

Atm it is not so critical as I'm the only one creating the binaries anyways. As soon as we have the new build process in place we could start building nightly builds with Travis which would result with clearly named files that it is not a release, but rather a snapshot.

PR #4917, opened 2020-12-08
Detect and handle invalid maker fee tx

chimp1984:
Check if maker fee tx is already spent or has invalid structure.
If so show popup to remove that offer.
Add maker fee tx id to offer details window.
If invalid, add error msg to offer details window.

### DIFF
--- a/core/src/main/java/bisq/core/offer/OfferUtil.java
+++ b/core/src/main/java/bisq/core/offer/OfferUtil.java
@@ -19,6 +19,7 @@ package bisq.core.offer;
 
 import bisq.core.account.witness.AccountAgeWitnessService;
 import bisq.core.btc.wallet.BsqWalletService;
+import bisq.core.btc.wallet.BtcWalletService;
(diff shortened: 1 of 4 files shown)


core/src/main/java/bisq/core/offer/OfferUtil.java
```diff
@@ -383,4 +387,53 @@ public void validateOfferData(double buyerSecurityDeposit,
...
+                log.error("spentByTransactionInput {}", spentByTransactionInput);
+            }
+        } else {
+            errorMsg = "The maker fee tx is invalid as it does not has at least 2 outputs." + extraString +
+                    "\nMakerFeeTx=" + makerFeeTx.toString();
+        }
+
+        if (errorMsg == null) {
+            return Optional.empty();
+        }
```
ripcurlx, line 429: Maybe it would be easier to read if that is used as the else branch for line 421, as this is the only case where errorMsg can be null.

ripcurlx (approved):
utACK

PR #4921, opened 2020-12-08
Detect missing deposit tx to allow moving to failed trades

### LINKED ISSUE #4905: Unable to move missing deposit tx trade to failed trades
### Description

After a couple days without the deposit tx (which was displayed at Bisq UI) being broadcasted, mediator tells me to make a SPV resync and send trade to failed.
After SPV resync, this "null deposit tx, send to failed" warning is displayed.
imagen
But it's not possible to send it to failed. I still have locked funds.

#### Version

1.5

### Expected behaviour

Red symbol at the top right of the trade should appear, to move trade to failed.

### Actual behaviour

No red symbol. Funds locked.

### Screenshots

imagen

#### Device or machine

Linux Mint.

chimp1984:
We got some reports where users have no deposit tx displayed in the
trade after spv resync but cannot move the trade to failed trades.
It seems the invalid txId is still stored in the trade but the tx itself
got removed from the wallet after reysnc. We check not that both the tx
and the txId need to be present.

Fixes https://github.com/bisq-network/bisq/issues/4905

### DIFF
--- a/core/src/main/java/bisq/core/trade/Trade.java
+++ b/core/src/main/java/bisq/core/trade/Trade.java
@@ -1030,6 +1030,7 @@ public abstract class Trade implements Tradable, Model {
         return offer.getOfferFeePaymentTxId() == null ||
                 getTakerFeeTxId() == null ||
                 getDepositTxId() == null ||
+                getDepositTx() == null ||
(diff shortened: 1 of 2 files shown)


ripcurlx (approved):
utACK

PR #4900, opened 2020-12-04
Add optional txFeeRate parameter to api sendbsq

ghubstan:
If present in the api's `sendbsq` command, the parameter will override the fee service and custom fee rate setting for the BSQ transaction.

Also changed the `sendbsq` grpc return type to a lightweight TX proto wrapper.

Besides some small refactoring in the CLI, all the changes are adjustments for this new `sendbsq` parameter and its new grpc return value.

This is the 3rd in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4884.
PR https://github.com/bisq-network/bisq/pull/4896 should be reviewed before this one.

PR #4922, opened 2020-12-09
Add new api method 'sendbtc' and test

ghubstan:
Takes an address, amount, and optional txfeerate param, returns a lightweight TxInfo proto.

Also overloaded two `BtcWalletService` methods to allow `sendbtc` to pass in the tx fee rate -- overriding the fee service and custom fee rate setting.

This is the 4th in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4884.
PR https://github.com/bisq-network/bisq/pull/4900 should be reviewed before this one.


ghubstan:
Force travis jar download & build

apitest/src/test/java/bisq/apitest/method/wallet/BtcWalletTest.java
```diff
@@ -92,6 +93,38 @@ public void testFundAlicesBtcWallet(final TestInfo testInfo) {
...
+                        0,
+                        700000000,
+                        0);
+        verifyBtcBalances(alicesExpectedBalances, alicesBalances);
+
+        BtcBalanceInfo bobsBalances = getBtcBalances(bobdaemon);
+        log.debug("{} Bob's BTC Balances:\n{}",
+                testName(testInfo),
+                formatBtcBalanceInfoTbl(bobsBalances));
+        // We cannot (?) predict the exact tx size and calculate how much in tx fees were
```
sqrrm, line 121: Should be deterministic since you're controlling all the inputs and the fee.

sqrrm (commented):
utACK

Waiting for previous PRs in the chain

apitest/src/test/java/bisq/apitest/method/wallet/BtcWalletTest.java
ghubstan: I think so too, but the `sendbtc` tx size randomly varies between two distinct values (219 bytes or 220 bytes) from test run to test run, hence the available balance range [1549978000, 1549978100].
Is that a bug?  
Should I be more specific about this in that test case's comment?

sqrrm: If the same address is used every time it should be deterministic, but if it's sending to different addresses due to some non deterministic events within the test it could result in different size txs.

ghubstan:

apitest/src/test/java/bisq/apitest/method/wallet/BtcWalletTest.java
ghubstan: Commit 9250845a66bc7cda2c369806bf0949537bb5ac18 changes the comment:
```
// The sendbtc tx weight and size randomly varies between two distinct values
// (876 wu, 219 bytes, OR 880 wu, 220 bytes) from test run to test run, hence
// the assertion of an available balance range [1549978000, 1549978100].
```

PR #4923, opened 2020-12-09
Fix sorting of table columns

chimp1984:
Fix incorrect or not optimized sorting of table columns


ripcurlx (approved):
utACK

pazza83:
This is working really well. Many thanks.

PR #4925, opened 2020-12-09
Support tx memo field for btc withdrawals from api

ghubstan:
- Added optional memo parameter to the api's `sendbtc` and `withdrawfunds` commands.

- Removed the `@Nullable` annotation was removed because protobuf does not support null.  The implementation will pass null if parameter string is empty.

- Visibility in two `CoreWalletService` wallet check methods were changed from private to pkg protected so the `CoreTradeService` could use them.
  
- Adjusted affected tests.  (Asserting the memo field was set on a transaction cannot be checked from `apitest` yet.)

This is the 5th in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4884.
PR https://github.com/bisq-network/bisq/pull/4922 should be reviewed before this one.


sqrrm (commented):
utACK

ghubstan:

PR #4929, opened 2020-12-10
Make sign age column sortable

chimp1984:
Add support for sorting by signed account age

We separate in 4 groups and sort each by days.
1. signed accounts
2. not yet signed account, we use account age
3. not required signing, we use account age
4. no account age and no signing required (altcoins)


ripcurlx (approved):
ACK

Tested it on Regtest

pazza83:
This is great, works really well.

PR #4930, opened 2020-12-10
Use HistoricalDataStoreService for AccountAgeWitnessStorageService 

chimp1984:
Use HistoricalDataStoreService for AccountAgeWitnessStorageService 

@ripcurlx We need to deploy the resource file with the version number and add the version to Version.HISTORICAL_RESOURCE_FILE_VERSION_TAGS. It is not required that we update all HistoricalDataStores (e.g. we can deploay the AccountAgeWitnessStore_1.5.2 but no new TradeStatistics2Store...


ripcurlx (approved):
utACK

ripcurlx:
I've added it to the script as well.

PR #4936, opened 2020-12-11
Use Bisq's UserThread.executor in gRPC server

ghubstan:
This is the 6th in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4884.
PR https://github.com/bisq-network/bisq/pull/4925 should be reviewed before this one.

PR #4937, opened 2020-12-12
Append nullable withdrawalTxId field to Trade proto message

ghubstan:
The `withdrawalTxId` field will be set in `TradeManager#onWithdrawRequest` after successful trade completion and withdrawal of funds.

Persisting `withdrawalTxId` will allow the api and ui to find the `withdrawalTxId` for a completed trade after the seller withdraws funds to an external wallet.  In turn, the withdrawal tx's memo field will be accessible in a new api `getTransaction(id)` api method (todo) .

Changed:

- Appended field `string withdrawal_tx_id = 40` to pb.proto's Trade message.

- Added nullable `String withdrawalTxId` to Trade entity class.

- Added `trade.setWithdrawalTxId(transaction.getTxId().toString())` in `TradeManager#onWithdrawRequest`'s callback.

This is the 7th in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4884.
PR https://github.com/bisq-network/bisq/pull/4936 should be reviewed before this one.


ghubstan:
Force travis jar download

core/src/main/java/bisq/core/trade/Trade.java
```diff
@@ -317,6 +317,10 @@ public boolean isArbitrated() {
     @Getter
     @Setter
     private String payoutTxId;
+    @Nullable
+    @Getter
+    @Setter
+    private String withdrawalTxId;
```
sqrrm, line 323: I'm not convinced about adding this tx to the trade. The Trade object is rather bloated anyway and I don't feel this tx is really associated.

ghubstan: Your call.  I'll [adjust the next PR](https://github.com/bisq-network/bisq/pull/4948) if you don't want it.

I thought it was an inexpensive way to associate the withdrawal tx to the trade.  To me, setting a memo on a trade withdrawal tx isn't very useful if you don't have an easy way to find that tx.

By the way, this tx is associated (in my head) when I optionally withdraw (seller) trade funds to an external wallet, as the final step of the trade protocol.

sqrrm: Perhaps you're right. I'm just cautious of adding more stuff to Trade. Will have a sleep on it.

ghubstan: I did not think this would make much sense to anyone until they've seen the next [PR](https://github.com/bisq-network/bisq/pull/4948), and I explain my reason a bit more/better.


Here is Bob's `gettrade` CLI output after withdrawing funds to an external wallet:

```
ID       My Role             Price in USD for 1 BTC  Amount(BTC)  Tx Fee(BTC)  Taker Fee(BTC)  Deposit Published  Deposit Confirmed  Fiat Sent  Fiat Received  Payout Published  Withdrawn  Withdrawal TX ID  
5006844  BTC buyer as taker             19,410.6419   0.12500000   0.00012932      0.00037500  YES                YES                YES        YES            YES               YES        0128e08406542a15ecc27234afa9ca9f4112f74d105459d18b4cfde93028291b
``` 

If `Withdrawn` = NO, the `Withdrawal TX ID` column would not appear in the console output.

Here is Bob's `gettransaction` CLI output:

```
Tx ID                                                             Is Confirmed  Tx Inputs (BTC)  Tx Outputs (BTC)  Tx Fee (BTC)  Tx Size (Bytes)  Memo  
0128e08406542a15ecc27234afa9ca9f4112f74d105459d18b4cfde93028291b           YES       0.14375000        0.14368229    0.00006771              193  Bob's trade withdrawal
```

Likewise, if there is no memo, the `Memo` column is not displayed.

chimp1984: The payout address might be useful for looking up the spent tx of that utxo which is the withdrawal tx.
I am not so sure if we should add that to the trade as it is not really part of the trade. The trade ends when you have received the funds. Most users I assume keep the funds anyway in the wallet so there is no withdrawal tx but the funds are used as input for another offer or trade.

sqrrm: Thought so to begin with as well. Now I'm thinking it makes more sense to associate the withdrawal with the trade as it's a way to know where the bitcoin went. Considering we have the feature to fund trades externally and withdraw directly it seems this withdrawal is directly related to this trade.

I now have an utACK for this PR.

chimp1984 (commented):
NACK

As discussed in keybase I think the withdrawal is not part of the trade conceptually. If funds are kept in Bisq the funds are likely used to fund the next trade fee tx and that is tagged with the new offer/trade, so then the same tx would appear in 2 different trades.

ghubstan:
Closing, and reverting this PR's commit in https://github.com/bisq-network/bisq/pull/4948.

PR #4942, opened 2020-12-14
Refactor http client

chimp1984:
Refactors and fixes HttpClient

Add support for POST (Needed for follow up PR)
Do not allow multiple requests before one request is completed
Use different httpClients for getFee and getPrice


ripcurlx (approved):
ACK

Code looks fine and I'm still receiving correct prices from pricenodes when running on Mainnet.

sqrrm:
utACK

I thumbs up for the well scoped commits.

PR #4943, opened 2020-12-14
Add tx broadcast to mempool explorer api

chimp1984:
Send raw tx to mempool space explorer nodes.
Use https://mempool.space/api

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/http/MemPoolSpaceTxBroadcaster.java
+++ b/core/src/main/java/bisq/core/btc/wallet/http/MemPoolSpaceTxBroadcaster.java
@@ -121,7 +121,7 @@ public class MemPoolSpaceTxBroadcaster {
                 if (cause instanceof HttpException) {
                     int responseCode = ((HttpException) cause).getResponseCode();
                     String message = cause.getMessage();
-                    if (responseCode == 400 && message.equals("sendrawtransaction RPC error: {\"code\":-27,\"message\":\"Transaction already in block chain\"}")) {
+                    if (responseCode == 400 && message.contains("code\":-27")) {
                         log.info("Broadcast of raw tx to {} failed as transaction {} is already confirmed",
                                 serviceAddress, txIdToSend);
                     } else {
--- a/core/src/main/java/bisq/core/btc/wallet/http/MemPoolSpaceTxBroadcaster.java
+++ b/core/src/main/java/bisq/core/btc/wallet/http/MemPoolSpaceTxBroadcaster.java
@@ -122,6 +122,7 @@ public class MemPoolSpaceTxBroadcaster {
                 if (cause instanceof HttpException) {
                     int responseCode = ((HttpException) cause).getResponseCode();
                     String message = cause.getMessage();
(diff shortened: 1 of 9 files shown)


core/src/main/java/bisq/core/btc/wallet/http/MemPoolSpaceTxBroadcaster.java
```diff
@@ -115,8 +114,21 @@ public void onSuccess(String txId) {
             }
 
             public void onFailure(@NotNull Throwable throwable) {
-                log.warn("Broadcast of raw tx with txId {} to {} failed. Error={}",
-                        txIdToSend, url, throwable.toString());
+                Throwable cause = throwable.getCause();
+                if (cause instanceof HttpException) {
+                    int responseCode = ((HttpException) cause).getResponseCode();
+                    String message = cause.getMessage();
+                    if (responseCode == 400 && message.equals("sendrawtransaction RPC error: {\"code\":-27,\"message\":\"Transaction already in block chain\"}")) {
```
wiz: Instead of checking for that exact message, maybe just checking for a substring of `\"code\":-27` is better, since the exact string might change in future releases of Bitcoin Core RPC

chimp1984: ok. do you have a list of other return codes/messages?

wiz: https://github.com/bitcoin/bitcoin/blob/master/src/rpc/protocol.h

sqrrm (approved):
utACK

My main concern with this PR is that it reduces privacy in that the mempool operators get more information on bisq trades. I don't think this matters much as long as we have the very typical signature for bisq trades though.

dcod3d:
Is it possible to not use this api if we are connected to our own nodes?

wiz:
Good point, there should be an option to disable this for privacy concerns, since it's just a backup broadcast feature anyway.

chimp1984:
Sure can be all added, just today EOD is code freeze for 1.5.2....

But to bring that into the right context: 
The privacy issue here is that node operators could track which transactions are from Bisq. For trade txs that is pretty irrelevant as the trade txs have a pretty significant fingerprint from their structure. For withdrawal txs that more of a concern.

But the Bitcoin node operators are in the same position anyway (much worse if we would use the public network as bloomfilters are terribly broken privacy wise). So for those who do not run their own node there is not much of a change. 

But adding a simple option in the settings might be problematic as well as it requires some conext about the problem to make any informed decision. If users deactivate it and end up with failed trades it also impacts the trade peer (and cause lot of cost for Bisq from supports side).
So to find the right solution here is not as trivial...

dcod3d:
I agree if using the provided nodes, this really isn't a privacy issue. But can this be disabled if using my own nodes.

chimp1984:
PR is made: https://github.com/bisq-network/bisq/pull/4958

If local node is used the tx broadcast will not be done.

I am not sure if that is perfect as it still carries the risk that a tx does not get broadcast (might be even a higher risk as there is only 1 node). To override that in the settings might be a solution but as said without context this might confuse more as it helps....

PR #4944, opened 2020-12-14
Improve TransferWise account

chimp1984:
Rename 'Supported currencies' to 'Currencies for receiving funds'
Do not autofill all currencies by default but keep all unselected.


ripcurlx (approved):
utACK

PR #4947, opened 2020-12-14
Fix bug with maker fee tx at create offer

chimp1984:
Fix null pointer at offer details window at create offer. As that window is only shown if devMode=false it was missed in previous PR.


sqrrm (approved):
utACK

PR #4949, opened 2020-12-14
Show price deviation in % for fixed price offers and BSQ

chimp1984:
Add price deviation in % to fix price offers and BSQ (using 30 day average)
Allow sorting by % price in "show all" mode.

There are performance issues in the offer book as we use the view model for the item value calculations and do not cache the values. With those changes here it got worse. So we need to refactor the handling of the ListItem so that the ListItem caches the data and only updates when needed (e.g. when a market price feed update arrives). There will be follow up commits for that.

UPDATE:
The issue was the 30 day average BSQ price which was calculated too frequently. This is fixed now and the table behaves similar like the master version. Still need for performance improvement in that area, but seems that PR does not make things worse.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBookView.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBookView.java
@@ -1010,12 +1010,12 @@ public class OfferBookView extends ActivatableViewAndModel<GridPane, OfferBookVi
                             }
 
                             @Override
-                            public void updateItem(final OfferBookListItem newItem, boolean empty) {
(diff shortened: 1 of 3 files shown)


chimp1984:
Screen Shot 2020-12-14 at 17 45 42

chimp1984:
Travis has problems again...

sqrrm (approved):
ACK

pazza83:
This is such a good addition 👍 

Please can it also be added to the 'Trades' section.

Being able to now export the trades section is great, but it would be even better is the deviation could also be exported.

An option to filter this section by 'payment method' as well as 'currency' would also be really useful.

Add payment method

PR #4950, opened 2020-12-15
Improve offer publishing

chimp1984:
Change handling of delay at republishing

The delay was too long. Some users have 100 - 200
offers and with the 700 ms delay it takes 70-140 sec.
This causes more stress for the network and UI due
permanent adding of offers. We decreased delay to 30 ms per offer.
So with 200 offers it would be about 6 sec. Maybe we could reduce
it even further but as it is hard to test in the live network with
so many offers it is better to not be too radical with the change.

### DIFF
--- a/core/src/main/java/bisq/core/offer/OpenOfferManager.java
+++ b/core/src/main/java/bisq/core/offer/OpenOfferManager.java
@@ -900,12 +900,10 @@ public class OpenOfferManager implements PeerManager.Listener, DecryptedDirectMe
         offerBookService.addOffer(openOffer.getOffer(),
                 () -> {
                     if (!stopped) {
-                        log.debug("Successfully added offer to P2P network.");
                         // Refresh means we send only the data needed to refresh the TTL (hash, signature and sequence no.)
-                        if (periodicRefreshOffersTimer == null)
+                        if (periodicRefreshOffersTimer == null) {
                             startPeriodicRefreshOffersTimer();
-                    } else {
-                        log.debug("We have stopped already. We ignore that offerBookService.republishOffers.onSuccess call.");
+                        }
                     }
                 },
                 errorMessage -> {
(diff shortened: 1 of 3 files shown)


core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -871,39 +871,46 @@ private void maybeUpdatePersistedOffers() {
...
+        stopPeriodicRefreshOffersTimer();
+    }
+
+    private void republishOffers(List<OpenOffer> list) {
+        if (list.isEmpty()) {
+            return;
         }
+
+        OpenOffer openOffer = list.remove(0);
+        if (!openOffers.contains(openOffer) || openOffer.isDeactivated()) {
```
sqrrm: Shouldn't this be `!openOffer.isDeactivated()` since it's being republished. That was the check before this change.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -871,39 +871,46 @@ private void maybeUpdatePersistedOffers() {
...
+            republishOffers(list);
+        }
+
+        republishOffer(openOffer,
+                () -> UserThread.runAfter(() -> republishOffers(list),
+                        30, TimeUnit.MILLISECONDS));
```
sqrrm: I don't understand the reason for the delay, all offers will be prepared to republish recursively, then during the recursion unwind they will all get this call so they delay is likely less than 1ms between each offer republish, but all the calls to republish will happen 30ms later. I think this is not much different to republishing without delay.

The previous version had a random delay, but it was a multiple of the number of orders and 700, so 200 orders ran into 140 seconds republish time and each offer would be spaced out in that time.

sqrrm (changes requested):

core/src/main/java/bisq/core/offer/OpenOfferManager.java
chimp1984: No it is correct, we call our recursive method with the list where we have removed that offer. But the following lines neede to be in a else branch.
I made a PR where I invert the if check and add a comment.

sqrrm: Yes, understood after the fix. Also the fix handles the recursion issue so I think it's good to merge.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
chimp1984: Ah damn, yes you are right...Wanted to avoid the way it was done before, but the delay from the callback is too short to be used as delay. I am acutally not very sure what happens if one publishes 200 offers without delay. It might be that connections to other peers get dropped due dos protection, but not sure if we hit those thresholds. To make it more smooth for the node and the network the delay was used, but could be that it makes things actually worse as it causes more messages (we use bundledEnvelope now to group msg in 1 container).

Maybe best to leave the PR for now open and we do more tests first, just hard to test 200 offers on mainnet ;-(.

sqrrm: Yeah, could be it's better to have no delay and get all these messages into one bundle (or just a few) rather than spreading over time

ripcurlx:
Travis is failing because of broken Tests.

ripcurlx:
@sqrrm Could you please review the changes so it can be added to the release branch asap?

sqrrm (commented):
utACK

Best wait to merge to master until it's been tested

sqrrm (approved):
utACK

Good find

PR #4951, opened 2020-12-15
[WIP] Add a lookup cache for txOutputs

chimp1984:
TxOutputs are requested very often and the previous iterations
accumulate lot of cpu time as profiler shows.

### DIFF
--- a/core/src/main/java/bisq/core/dao/state/DaoStateService.java
+++ b/core/src/main/java/bisq/core/dao/state/DaoStateService.java
@@ -47,6 +47,7 @@ import java.util.Comparator;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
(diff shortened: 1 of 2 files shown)


chimp1984:
Here are performance test results.

PR version:
100 calls of daoStateService.getTxOutputsByTxOutputType(TxOutputType.BSQ_OUTPUT) took 2508 ms 
100000 calls of daoStateService.existsTxOutput(TxOutputKey) took 2 ms 
100000 calls of daoStateService.getTxOutput(TxOutputKey) took 3 ms 

Master:
100 calls of daoStateService.getTxOutputsByTxOutputType(TxOutputType.BSQ_OUTPUT) took 4206 ms 
100 calls of daoStateService.existsTxOutput(TxOutputKey) took 2907 ms 
100 calls of daoStateService.existsTxOutput(TxOutputKey) took 2946 ms 

The daoStateService.getTxOutputsByTxOutputType call takes about 30 ms which is very expensive and 
requires more improvement but as it iterates the cache map it is expected to has little impact.

The other 2 methods havea huge difference.
Note that iterations have been increased from 100 to 100000 to get some meaningful time log.

I think we should measure all methods and find more hotspots...

As its a risky domain, I leave it to maintainers to merge the PR or delay it for later.

chimp1984:
I checked how often the methods are called:
getTxOutputsByTxOutputType is called about 500 times after a few minutes and clicking 
though most DAO screens resulting in 15 sec accululated CPU time.

getTxOutput and existsTxOutput is only used from the BsqCoinSelector and NonBsqCoinSelector 
in edge cases, so only relevant for users who have BSQ in their wallet. I got existsTxOutput called 
3 times and getTxOutput called 0 times with a live app with BSQ.

So unfortunately the big improvements will not have much impact as not called often 
and the one which is called more often brings not so much difference. 

So not sure if the added risk and the added memory footprint by adding a cache as 
additional data structure (150k objects) are worth it. 

I would tend to postpone that improvement to find the more severe bottlenecks. I am still a bit surprised as profiler 
reported the getTxOutputsByTxOutputType method as hotspot.

PR #4952, opened 2020-12-15
Allow sending private notifications from avatar icon at trade

chimp1984:
Allow sending private notifications from avatar icon at trade.
Before it was using the makers pubKeyRing from the offer and that fails if the user was the maker as he cannot send to himself a private notofication.


sqrrm (approved):
utACK

PR #4956, opened 2020-12-15
Sort payment methods by display string not by id.

chimp1984:
Apply sorting in drop down boxes.


sqrrm (commented):
Looks ok, but why do we want to sort by display string rather than id?

pazza83:
I am assuming strings are the user generated names for their payment accounts?
This allows for payment methods to be sorted by alphabetical order. This would be a great change.

sqrrm (approved):
utACK

Thanks @pazza83 for explaining why this is good

PR #4957, opened 2020-12-15
Sign single account

sqrrm:
This a proposed solution to https://github.com/bisq-network/growth/issues/212

The user to get signed copies the account data and sends it over whichever channel, keybase for example. With this change the arbitrator can sign this account.

Step 1, user copies account data (ctrl+c to copy AccountAgeWitness and PubKey to clipboard)
image

Step 2, arbitrator imports data (under accounts view, open signing tab with ctrl+i, in signing tab, open specific account signing window with ctrl+p)
image

Step 3, arbitrator signs with their private key (this could be helpful for devs that want to test signed accounts as well)
image

Step 4, done
image

I know this is very pretty, but please don't ask me to implement all the UI stuff due to this flash feature.

### DIFF
--- a/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
+++ b/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
@@ -880,4 +880,10 @@ public class AccountAgeWitnessService {
                 !peerHasSignedWitness(trade) &&
                 tradeAmountIsSufficient(trade.getTradeAmount());
     }
+
+    public String signInfoFromAccount(PaymentAccount paymentAccount) {
+        var pubKey = keyRing.getSignatureKeyPair().getPublic();
+        var witness = getMyWitness(paymentAccount.getPaymentAccountPayload());
+        return Utilities.bytesAsHexString(witness.getHash()) + "," + Utilities.bytesAsHexString(pubKey.getEncoded());
+    }
 }
--- a/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
+++ b/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
@@ -51,6 +51,7 @@ import bisq.common.crypto.PubKeyRing;
 import bisq.common.crypto.Sig;
 import bisq.common.handlers.ErrorMessageHandler;
 import bisq.common.util.MathUtils;
+import bisq.common.util.Tuple2;
 import bisq.common.util.Utilities;
 
 import org.bitcoinj.core.Coin;
@@ -885,5 +886,25 @@ public class AccountAgeWitnessService {
         var pubKey = keyRing.getSignatureKeyPair().getPublic();
         var witness = getMyWitness(paymentAccount.getPaymentAccountPayload());
         return Utilities.bytesAsHexString(witness.getHash()) + "," + Utilities.bytesAsHexString(pubKey.getEncoded());
+    }
+
+    public Tuple2<AccountAgeWitness, byte[]> signInfoFromString(String signInfo) {
+        var parts = signInfo.split(",");
+        if (parts.length != 2){
+            return null;
+        }
+        byte[] pubKeyHash;
+        Optional<AccountAgeWitness> accountAgeWitness;
+        try {
+            var accountAgeWitnessHash = Utilities.decodeFromHex(parts[0]);
+            pubKeyHash = Utilities.decodeFromHex(parts[1]);
+            accountAgeWitness = getWitnessByHash(accountAgeWitnessHash);
+        } catch (Exception e) {
+            return null;
+        }
+
+        return accountAgeWitness
+                .map(ageWitness -> new Tuple2<>(ageWitness, pubKeyHash))
+                .orElse(null);
     }
 }
--- a/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
+++ b/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
@@ -882,13 +882,13 @@ public class AccountAgeWitnessService {
                 tradeAmountIsSufficient(trade.getTradeAmount());
     }
 
-    public String signInfoFromAccount(PaymentAccount paymentAccount) {
+    public String getSignInfoFromAccount(PaymentAccount paymentAccount) {
         var pubKey = keyRing.getSignatureKeyPair().getPublic();
         var witness = getMyWitness(paymentAccount.getPaymentAccountPayload());
         return Utilities.bytesAsHexString(witness.getHash()) + "," + Utilities.bytesAsHexString(pubKey.getEncoded());
     }
 
-    public Tuple2<AccountAgeWitness, byte[]> signInfoFromString(String signInfo) {
+    public Tuple2<AccountAgeWitness, byte[]> getSignInfoFromString(String signInfo) {
         var parts = signInfo.split(",");
         if (parts.length != 2) {
             return null;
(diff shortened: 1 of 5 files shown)


core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
```diff
@@ -880,4 +880,10 @@ public boolean isSignWitnessTrade(Trade trade) {
                 !peerHasSignedWitness(trade) &&
                 tradeAmountIsSufficient(trade.getTradeAmount());
     }
+
+    public String signInfoFromAccount(PaymentAccount paymentAccount) {
+        var pubKey = keyRing.getSignatureKeyPair().getPublic();
+        var witness = getMyWitness(paymentAccount.getPaymentAccountPayload());
+        return Utilities.bytesAsHexString(witness.getHash()) + "," + Utilities.bytesAsHexString(pubKey.getEncoded());
```
chimp1984, line 888: Is that a debug util or is the "," concatenation used in normal cases?

chimp1984: If debug tool better to rename method to make it more clear and add a comment...

desktop/src/main/java/bisq/desktop/main/account/content/PaymentAccountsView.java
```diff
@@ -62,6 +62,8 @@ public void initialize() {
                 accountAgeWitnessService.getAccountAgeWitnessUtils().logSigners();
             } else if (Utilities.isCtrlShiftPressed(KeyCode.U, event)) {
                 accountAgeWitnessService.getAccountAgeWitnessUtils().logUnsignedSignerPubKeys();
+            } else if (Utilities.isAltOrCtrlPressed(KeyCode.C, event)) {
```
chimp1984, line 65: Cmd+c is a very common key command, not sure if its better to use a more exotic one so that user do not trigger that accidentally. Prob. no harm by doing that, but confusing...

chimp1984: As its only used for fiat, better move it to fiat class instead of base class

core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
```diff
@@ -886,4 +887,24 @@ public String signInfoFromAccount(PaymentAccount paymentAccount) {
...
+        Optional<AccountAgeWitness> accountAgeWitness;
+        try {
+            var accountAgeWitnessHash = Utilities.decodeFromHex(parts[0]);
+            pubKeyHash = Utilities.decodeFromHex(parts[1]);
+            accountAgeWitness = getWitnessByHash(accountAgeWitnessHash);
+        } catch (Exception e) {
+            return null;
+        }
+
+        return accountAgeWitness
```
chimp1984: Why not move return into the try block?

```
return getWitnessByHash(accountAgeWitnessHash)
                .map(ageWitness -> new Tuple2<>(ageWitness, pubKeyHash))
                .orElse(null);
```

desktop/src/main/java/bisq/desktop/main/overlays/windows/SignSpecificWitnessWindow.java
```diff
@@ -194,15 +161,24 @@ private void addSuccessContent() {
...
         actionButton.setDisable(true);
-        actionButton.setOnAction(e -> selectAccountAgeWitness());
+        actionButton.setOnAction(e -> importAccountAgeWitness());
 
         closeButton = (AutoTooltipButton) buttonTuple.second;
         closeButton.setOnAction(e -> hide());
+    }
 
+    private ECKey checkedArbitratorKey() {
+        var arbitratorKey = arbitratorManager.getRegistrationKey(privateKey.getText());
```
chimp1984, line 176: So that use old arbitrators, right? there are none registered anymore. Wouldn't it be better to use mediators or refund agents?

core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
```diff
@@ -886,4 +887,24 @@ public String signInfoFromAccount(PaymentAccount paymentAccount) {
         var witness = getMyWitness(paymentAccount.getPaymentAccountPayload());
         return Utilities.bytesAsHexString(witness.getHash()) + "," + Utilities.bytesAsHexString(pubKey.getEncoded());
     }
+
+    public Tuple2<AccountAgeWitness, byte[]> signInfoFromString(String signInfo) {
+        var parts = signInfo.split(",");
```
chimp1984, line 892: Ah now I see why u use "," for concatination....
A pipe or a static field for it would make it a bit more intuitive... but no worry... being picky ;-)

core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
chimp1984: After reviewing the rest I understand now better... I stil would recommend to rename the method so it is more clear its not a normal user method but a special tool... also i am old-school and prefer get prefix... here it could read as intent as well: sign infoFromAccount.... instead of: get signInfo from account

chimp1984 (dismissed):
utACK, some nits and suggestions but looks ok... not reviewed UI part much

desktop/src/main/java/bisq/desktop/main/account/content/PaymentAccountsView.java
sqrrm: I like adding it here, and I like ctrl+c since it's the common copy key and that's what we're doing here, copying the account info to clipboard.

core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
sqrrm: Yes, the naming was not good, changed to `getSignInfo...`

core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
sqrrm: Yes, done.

desktop/src/main/java/bisq/desktop/main/account/content/PaymentAccountsView.java
chimp1984: it opens a popup window right?

core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
sqrrm: Won't change, just easier to read with `,` as separator.

desktop/src/main/java/bisq/desktop/main/account/content/PaymentAccountsView.java
chimp1984: or just copy to clipboard?

chimp1984: sorry have not looked too close to the code base....

chimp1984 (approved):
utACK

desktop/src/main/java/bisq/desktop/main/overlays/windows/SignSpecificWitnessWindow.java
sqrrm: It might be but that's a bigger change and I'm not convinced. I think that would add another complication to adding more mediators.

desktop/src/main/java/bisq/desktop/main/account/content/PaymentAccountsView.java
sqrrm: Just copy to clipboard, no need for popup window.

ripcurlx (approved):
utACK

Will test it in detail during release testing.

PR #4958, opened 2020-12-15
Do not use tx broadcast to mempool nodes if a local btc nodes is used

chimp1984:
See discussion at: https://github.com/bisq-network/bisq/pull/4943#issuecomment-745615764


sqrrm (approved):
utACK

I agree that it's not clear that this is the right solution, but I think this is at least a proper balance to allow those that run their own node for privacy reasons to not share data they didn't intend to share.

PR #4960, opened 2020-12-16
Refactor gRPC service error handling

ghubstan:
This change reduces gRPC service error handling duplication by moving it to a `@Singleton CoreApiExceptionHandler `encapsulating everything needed to map an expected or unexpected core api exception to a `gRPC StatusRuntimeException` before sending it to the client.  It also fixes some cases were gRPC error handling was missing.

This is the 9th in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4884.
PR https://github.com/bisq-network/bisq/pull/4948 should be reviewed before this one.


ghubstan:
Force travis jar download.

PR #4961, opened 2020-12-16
Remove m52go btc node ip

m52go:
I need to move my bitcoin node to a less expensive server, but my hosting provider is giving me issues. This PR removes the IP for now while those issues are sorted.


wiz (approved):
ACK

ripcurlx (approved):
utACK

PR #4963, opened 2020-12-16
Call completeHandler if services in store services is empty.

chimp1984:
If one starts with --daoActivated=false there is no service
set up in one of the data store services so it never calls
the result handler and the app never starts up.


ripcurlx (approved):
utACK

sqrrm (approved):
utACK

PR #4966, opened 2020-12-17
Prevent excessive api calls

ghubstan:
This change provides a gRPC `CallRateMeteringInterceptor` to help protect the server and network against being overloaded by CLI scripting mistakes.

An interceptor instance can be configured on a gRPC service to set per method call rate limits on one or more of a service's methods. For example, the `GrpcOffersService` could be configured with this interceptor to set the `createoffer` rate limit to 5/hour, and the `takeoffer` call rate limit could be set to 20/day.  Whenever a call rate limit is exceeded, the gRPC call is aborted and the client receives a "rate limit exceeded" error.

Below is a simple example showing how to set rate limits for only the `getVersion` method in `GrpcVersionService`.

    final ServerInterceptor[] interceptors() {
        return new ServerInterceptor[]{
                new CallRateMeteringInterceptor(new HashMap<>() {{
                    put("getVersion", new GrpcCallRateMeter(2, SECONDS));
                }})
        };
    }

It specifies a CLI can execute `getversion` 2 times / second.

This is not a throttling mechanism, there is no blocking nor locking to slow call rates.  When call rates are exceeded, calls are simply aborted.

This is the 10th in a chain of PRs beginning with https://github.com/bisq-network/bisq/pull/4884.
PR https://github.com/bisq-network/bisq/pull/4960 should be reviewed before this one.


daemon/src/main/java/bisq/daemon/grpc/interceptor/CallRateMeteringInterceptor.java
```diff
@@ -0,0 +1,107 @@
...
+        log.error(StringUtils.capitalize(msg) + ".");
+        serverCall.close(FAILED_PRECONDITION.withDescription(msg), new Metadata());
+    }
+
+    private void handlePermissionDeniedErrorAndCloseCall(String methodName,
+                                                         GrpcCallRateMeter rateMeter,
+                                                         ServerCall<?, ?> serverCall)
+            throws StatusRuntimeException {
+        String msg = getDefaultRateExceededError(methodName, rateMeter);
+        log.error(StringUtils.capitalize(msg) + ".");
```
sqrrm: This should probably be a `log.warn` as it's not an error but rather expected when the rate limit is exceeded. Might consider changing the method name as well for the same reason.

daemon/src/main/java/bisq/daemon/grpc/interceptor/GrpcCallRateMeter.java
```diff
@@ -0,0 +1,65 @@
...
+
+import static java.lang.String.format;
+
+@Slf4j
+public final class GrpcCallRateMeter {
+
+    @Getter
+    private final long allowedCallsPerTimeUnit;
+    @Getter
+    private final TimeUnit timeUnit;
```
sqrrm, line 21: I find this a bit odd. I think it makes more sense to set the time window in seconds and just have a `int` with number of seconds in the time window.

daemon/src/main/java/bisq/daemon/grpc/interceptor/GrpcCallRateMeter.java
```diff
@@ -0,0 +1,65 @@
...
+    public GrpcCallRateMeter(long allowedCallsPerTimeUnit, TimeUnit timeUnit) {
+        this.allowedCallsPerTimeUnit = allowedCallsPerTimeUnit;
+        this.timeUnit = timeUnit;
+    }
+
+    public void start() {
+        if (timer != null)
+            timer.stop();
+
+        timer = UserThread.runPeriodically(() -> callsCount = 0, 1, timeUnit);
```
sqrrm: My expectation is to have a moving time window with the rate limit. The way it's done now I can send double the max amount if I send all at the end of the first period and then continues at the beginning of the second period.

daemon/src/main/java/bisq/daemon/grpc/interceptor/GrpcServiceRateMeteringConfig.java
```diff
@@ -0,0 +1,268 @@
...
+            rateMeterConfigs.stream().filter(c -> c.isConfigForGrpcService(grpcServiceClassName))
+                    .findFirst().ifPresentOrElse(
+                    (config) -> config.addMethodCallRateMeter(methodName, maxCalls, timeUnit),
+                    () -> rateMeterConfigs.add(new GrpcServiceRateMeteringConfig(grpcServiceClassName)
+                            .addMethodCallRateMeter(methodName, maxCalls, timeUnit)));
+        }
+
+        public File build() {
+            File tmpFile = serializeRateMeterDefinitions();
+            File configFile = new File("/tmp/ratemeters.json");
```
sqrrm: Not sure this is right unless it's only supposed to work on *nix.

sqrrm (changes requested):
Mainly I would like to see a moving time window for the rate limit.

Also, I might have missed but is the api only supposed to work on *nix systems? Seems very specific in many places.

daemon/src/main/java/bisq/daemon/grpc/interceptor/GrpcServiceRateMeteringConfig.java
ghubstan: It is only supposed to work on *nix but I changed line 233 to
`File configFile = Paths.get(getProperty("java.io.tmpdir"), "ratemeters.json").toFile();`
(Commit 01546ad)

daemon/src/main/java/bisq/daemon/grpc/interceptor/GrpcCallRateMeter.java
ghubstan: I rewrote the rate meter class in commit 01546ad.  There is a running call count over the current time window, instead of using this timer (oops).

daemon/src/main/java/bisq/daemon/grpc/interceptor/GrpcCallRateMeter.java
ghubstan: Fixed in commit 01546ad.  The rate meter can still use a DAYS and HOURS TimeUnit, though I supposed most method rate meters will restrict calls on a per second basis.

daemon/src/main/java/bisq/daemon/grpc/interceptor/CallRateMeteringInterceptor.java
ghubstan: Fixed in commit 01546ad.

ghubstan:
Force travis jar download.

daemon/src/main/java/bisq/daemon/grpc/interceptor/GrpcCallRateMeter.java
```diff
@@ -1,95 +1,92 @@
...
+
     @Getter
-    private transient boolean isRunning;
+    private transient final long timeUnitIntervalInMilliseconds;
 
-    @Nullable
-    private Timer timer;
+    private transient final ArrayDeque<Long> callTimestamps;
 
     public GrpcCallRateMeter(int allowedCallsPerTimeUnit, TimeUnit timeUnit) {
```
sqrrm: `allowedCallsPerTimeUnit` is a bit misleading as it's really allowedCallsPerTimeWindow.

daemon/src/main/java/bisq/daemon/grpc/interceptor/GrpcCallRateMeter.java
```diff
@@ -1,95 +1,92 @@
...
+        this.allowedCallsPerTimeUnit = allowedCallsPerTimeUnit;
+        this.timeUnit = timeUnit;
+        this.numTimeUnits = numTimeUnits;
+        this.timeUnitIntervalInMilliseconds = timeUnit.toMillis(1) * numTimeUnits;
+        this.callTimestamps = new ArrayDeque<>();
     }
 
-    public void incrementCallsCount() {
-        callsCount++;
+    public boolean isAllowed() {
```
sqrrm: `isAllowed()` indicates it's just checking if something is allowed, this method i changing the state of the class. I would prefer something `call(successCallback, failCallback)` or `checkAndIncrement()`

sqrrm (changes requested):
Functionality now looks right. Just some comments on naming.

daemon/src/main/java/bisq/daemon/grpc/interceptor/GrpcCallRateMeter.java
ghubstan: Fixed in commit 10727fc0830330e004351cd0567b994173a5d5c5.

daemon/src/main/java/bisq/daemon/grpc/interceptor/GrpcCallRateMeter.java
ghubstan: Fixed in commit 10727fc0830330e004351cd0567b994173a5d5c5.

sqrrm (approved):
utACK

PR #4969, opened 2020-12-18
Make it easier to translate trade roles

ripcurlx:
As the German translation was quite embarrassing I think it might be similar in other languages (see `Angebotstyp`). 
**Before**
Bildschirmfoto 2020-12-18 um 16 24 23
**After**
Bildschirmfoto 2020-12-18 um 17 08 35

I changed the source to make it easier to translate (at least I hope so)

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -3039,8 +3039,8 @@ navigation.dao.wallet.receive=\"DAO/BSQ Wallet/Receive\"
 
 formatter.formatVolumeLabel={0} amount{1}
 formatter.makerTaker=Maker as {0} {1} / Taker as {2} {3}
-formatter.youAreAsMaker=You are {0} {1} as maker / Taker is {2} {3}
-formatter.youAreAsTaker=You are {0} {1} as taker / Maker is {2} {3}
+formatter.youAreAsMaker=You are: {1} {0} (maker) / Taker is: {3} {2}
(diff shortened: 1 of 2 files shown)


sqrrm (approved):
utACK

Jakub-CZ:
I agree that some strings are impossible to translate in a way so that they would read in a natural way, or be grammatically correct at least.

Going forward, I'd suggest to steer away from composing non-trivial strings (sentences or longer statements) from shorter substrings like this.

Alternatively, these composite strings shouldn't rely on shared tokens (e.g. `shared.selling`) that may be used in all kinds of grammatical forms and contexts. For example: had there been something like `formatter.youAreAsMaker.selling=selling` used only in this context and nowhere else, this PR could've been avoided, as the translators would've had much more freedom to translate the various combinations in any way they see fit.

From what I saw when reviewing the Czech translation, I think it would be perfectly manageable to have individual strings for each combination of buying/selling etc. Even translators would have an easier time as they wouldn't have to be thinking about how to rephrase the whole string just because they must use "buying" instead of "buys", "buy", "purchase", or whatever else would fit in better in that language and context.

**Example:**
instead of
```formatter.youAreAsMaker=You are: {1} {0} (maker) / Taker is: {3} {2}```
there are really only two options:
```formatter.youAreBuyerAsMaker=You are: {0} buyer (maker) / Taker is: {1} seller```
```formatter.youAreSellerAsMaker=You are: {0} seller (maker) / Taker is: {1} buyer```

ripcurlx:
Yes, I think so too. As a developer you just intuitively try to generalize stuff if it looks like close repetition. But in the case of translations it makes quite harder instead.

PR #4971, opened 2020-12-19
Avoid resync from genesis in case of dao state issues

chimp1984:
In case a reset of the dao state was triggered we delete now all dao store files and request a shut down of the app.

After a restart the resource files are used. This avoids cases where a resync from
genesis got triggered (observed on seed nodes, not on desktop apps).

Seed nodes and headless apps get shut down automatically.
In case of the desktop app we show a warn popup with shutdown
button and no close button, so we enforce a shutdown to avoid
complications in case the user would continue.


ripcurlx:
A test is broken. Please see Travis log.

sqrrm (approved):
utACK

@chimp1984 You need to get the ack from @ManfredKarrer as well

ManfredKarrer (approved):
utACK

PR #4973, opened 2020-12-19
[1.5.2] Improve logs

chimp1984:
Reduce verbosity of logs.

I think all those commites are very low risk, but up to maintainers of they prefer to postpone it to master instead...


sqrrm (approved):
utACK

PR #4975, opened 2020-12-19
Improve CleanupMailboxMessages

chimp1984:
Check if message matches trade before checking pubKey

Use stream filters instead of checks in method.
Change log level from error to warn.


sqrrm (approved):
utACK

PR #4976, opened 2020-12-19
[1.5.2] Remove NGN (Nigerian Naira) as their central bank blocked Transferwise

chimp1984:
Remove NGN (Nigerian Naira) as their central bank blocked Transferwise

See: https://twitter.com/cenbank/status/1339196770774093825

I tested it with setting that currency, doing offers and trades,
then remove it and continue the trades.
Offers with NGN cannot be taken anymore.


sqrrm (approved):
utACK

PR #4977, opened 2020-12-19
[1.5.2] Add extra note for amazon gift cards

chimp1984:
Add extra note for amazon gift cards

There have been many reports with account block with AGC payments. It seems they block amounts > 100 USD. The cryptic trade ID as message might be another factor to get flagged. The change in the text should help to avoid those issues.

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -3315,10 +3315,14 @@ payment.payid.info=A PayID like a phone number, email address or an Australian B
   bank, credit union or building society account. You need to have already created a PayID with your Australian financial institution. \
   Both sending and receiving financial institutions must support PayID. For more information please check [HYPERLINK:https://payid.com.au/faqs/]
 payment.amazonGiftCard.info=To pay with Amazon eGift Card you need to purchase an Amazon eGift Card at your Amazon account and \
-  use the BTC seller''s email or mobile nr. as receiver. Amazon sends then an email or text message to the receiver. \
-  Use the trade ID for the message field.\n\n\
-  Amazon eGift Cards can only be redeemed by Amazon accounts with the same currency.\n\n\
-  For more information visit the Amazon eGift Card webpage. [HYPERLINK:https://www.amazon.com/Amazon-1_US_Email-eGift-Card/dp/B004LLIKVU]
+  use the BTC seller''s email or mobile nr. as receiver. Amazon sends then an email or text message to the receiver.\n\
+  Amazon eGift Cards can only be redeemed by Amazon accounts with the same currency.\n\
+  For more information visit the Amazon eGift Card webpage. [HYPERLINK:https://www.amazon.com/Amazon-1_US_Email-eGift-Card/dp/B004LLIKVU]\n\n\
+  It is recommended to not send more than 100 USD per gift card as it seems that Amazon is considering higher amounts as \
+  risky and often flags such transactions as fraudulent.\n\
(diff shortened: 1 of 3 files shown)


chimp1984:
Screen Shot 2020-12-19 at 17 13 53

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -3315,10 +3315,14 @@ payment.payid.info=A PayID like a phone number, email address or an Australian B
...
-  Amazon eGift Cards can only be redeemed by Amazon accounts with the same currency.\n\n\
-  For more information visit the Amazon eGift Card webpage. [HYPERLINK:https://www.amazon.com/Amazon-1_US_Email-eGift-Card/dp/B004LLIKVU]
+  use the BTC seller''s email or mobile nr. as receiver. Amazon sends then an email or text message to the receiver.\n\
+  Amazon eGift Cards can only be redeemed by Amazon accounts with the same currency.\n\
+  For more information visit the Amazon eGift Card webpage. [HYPERLINK:https://www.amazon.com/Amazon-1_US_Email-eGift-Card/dp/B004LLIKVU]\n\n\
+  It is recommended to not send more than 100 USD per gift card as it seems that Amazon is considering higher amounts as \
+  risky and often flags such transactions as fraudulent.\n\
+  Split the amount up to several smaller amounts and use some creative reference text instead of the trade ID (e.g. 'Happy birthday Susan').\n\
+  Exchange in the trade chat with the peer which reference text you have used so they are able to \
+  assign the payment to the trade.
```
sqrrm: I think it would be good if @m52go could have a look at this text to make it easier to understand. Might help avoid some mediation cases.

sqrrm (changes requested):

core/src/main/resources/i18n/displayStrings.properties
chimp1984: Agree!

pazza83:
Can we have @bisq-network/mediators input on this to see if there are any issues they are experiencing. 

Multiple transactions can get messy, and might also cause additional problems to the user if flagged by Amazon.

I am not sure if the issues experienced are due to a user making a first transaction with Amazon eGift cards. Amazon might put in place additional security measures that are removed once payment has been verified by the user. I see this as a good thing. It is Amazon acknowledging that payment is irreversible through additional safety measures.

I know fist transactions with Zelle also face these concerns.

m52go:
If this 100 USD soft/recommended cap is kept, my suggested text is below.
- link to wiki instead of Amazon (more thorough and informative for Bisq traders)
- specify that gift cards can only be used on the Amazon site they were purchased from (to my knowledge, this is the real constraint, not currency)

```
payment.amazonGiftCard.info=To pay with Amazon eGift Card, you will need to send an Amazon eGift Card to the BTC seller via your Amazon account. \n\n\
  Bisq will show the BTC seller''s email address or phone number where the gift card should be sent, and you must include the trade ID in the gift \
  card''s message field. Please see the wiki [HYPERLINK:https://bisq.wiki/Amazon_eGift_card] for further details and best practices. \n\n\
  Three important notes:\n\
  - try to send gift cards with amounts of 100 USD or smaller, as Amazon is known to flag larger gift cards as fraudulent\n\
  - try to use creative, believable text for the gift card''s message (e.g., "Happy birthday Susan!") along with the trade ID (and use trader chat \
  to tell your trading peer the reference text you picked so they can verify your payment)\n\
  - Amazon eGift Cards can only be redeemed on the Amazon website they were purchased on (e.g., a gift card purchased on amazon.it can only be redeemed on amazon.it)
```

Screenshot from 2020-12-20 04-13-24

ripcurlx:
Merging based on @m52go review which resolves @sqrrm's change request.

PR #4978, opened 2020-12-19
Translate checkArgument messages

### LINKED ISSUE #2619: Translate checkArgument messages
There are several checkArgument messages that are hard coded in English. Since these messages are presented to the user, they need to be translated. Will need to find all usages of checkArgument and add translations.
For example, one such instance https://github.com/bisq-network/bisq/blob/54f1a9ff5f07a3965fd6faa29ba60bb646736a0a/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java#L558-L561

Jakub-CZ:
Fixes #2619

Messages in some checkArgument calls are replaced with translatable strings.

There are many more instances, incl. ones in calls of similar functions (notEmpty, notBlank, notNull). However, most seem to be unreachable by the users; the GUI prevents the conditions from ever happening. Many messages are not even intended to be seen by users; they include things like private method names and local variable names.

In fact, I wasn't able to reproduce even some of those few messages that are updated by this PR.

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
@@ -25,6 +25,7 @@ import bisq.core.btc.model.AddressEntry;
 import bisq.core.btc.model.AddressEntryList;
 import bisq.core.btc.setup.WalletsSetup;
 import bisq.core.btc.wallet.http.MemPoolSpaceTxBroadcaster;
+import bisq.core.locale.Res;
 import bisq.core.provider.fee.FeeService;
 import bisq.core.user.Preferences;
 
@@ -1187,7 +1188,7 @@ public class BtcWalletService extends WalletService {
         Transaction tx = new Transaction(params);
         final Coin receiverAmount = amount.subtract(fee);
         Preconditions.checkArgument(Restrictions.isAboveDust(receiverAmount),
-                "The amount is too low (dust limit).");
+                Res.get("validation.amountBelowDust.short"));
         tx.addOutput(receiverAmount, Address.fromString(params, toAddress));
 
         SendRequest sendRequest = SendRequest.forTx(tx);
@@ -1217,7 +1218,7 @@ public class BtcWalletService extends WalletService {
         Transaction tx = new Transaction(params);
         final Coin netValue = amount.subtract(fee);
         checkArgument(Restrictions.isAboveDust(netValue),
-                "The amount is too low (dust limit).");
+                Res.get("validation.amountBelowDust.short"));
 
         tx.addOutput(netValue, Address.fromString(params, toAddress));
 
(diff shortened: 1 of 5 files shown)


Jakub-CZ:
Hi, this is my 2nd PR, just a little less trivial than my first one. I picked it up mainly to learn a bit more about the internals.

As I was making changes I realized that this rabbit hole goes deeper than I originally anticipated. On one hand, translating **all** `checkArgument` messages could introduce dozens and dozens of new strings, lots of extra work for the translators. On the other hand, most messages seem to be practically impossible to reach by regular users.

core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
```diff
@@ -1278,18 +1279,18 @@ public Transaction createRefundPayoutTx(Coin buyerAmount,
             throws AddressFormatException, InsufficientMoneyException, WalletException, TransactionVerificationException {
         Transaction tx = new Transaction(params);
         Preconditions.checkArgument(buyerAmount.add(sellerAmount).isPositive(),
-                "The sellerAmount + buyerAmount must be positive.");
+                Res.get("validation.refundPayoutTx.buyerSellerNotPositive"));
         // buyerAmount can be 0
         if (buyerAmount.isPositive()) {
             Preconditions.checkArgument(Restrictions.isAboveDust(buyerAmount),
-                    "The buyerAmount is too low (dust limit).");
+                    Res.get("validation.amountBelowDust.shortVar", "buyerAmount"));
```
sqrrm, line 1286: This kind of argument defeats the aim of making the strings translatable. In this case it's better to just make two strings, or reuse `validation.amountBelowDust.short`

sqrrm (changes requested):
In general I don't think it's worth moving these strings to resources. Some do have value as they're useful for users, but it's mostly meant as a way to trouble shoot issues together with users, or as a safety check that should never happen but just a good to have.

Adding more strings to resources adds more work for translators for very little benefit. I can still see the benefit of some of these strings so would ack if @ManfredKarrer agrees, as this is a change to core DAO code which can't be merged without his ack as per project guidelines.

ManfredKarrer (commented):
NACK

I don't think we should start translating those type of strings. They are mostly in place for catching edge case bugs and for debugging. In case such an error occurs the user can send a screenshot of the error popup to developers. It is not intended as error messages the user can deal with, so translation does not add value here.
It would also add more effort in dev process and for translators.

I appreciate your intentions and effort but I fear that PR is going in the wrong direction. Best to search the GH issue list for "best first issues" or the like. There are plenty of issues and TODOs open specially in the UI side which would add lot of value to Bisq.

Jakub-CZ:
Understood. I did pick an issue tagged with "good first issue", sadly it seems that the issue wasn't very well defined.

I agree that most instances of `checkArgument` are used for those "should never happen" kind of checks. The rest should rely on custom Exceptions rather than these generic `IllegalArgumentException`s.

ManfredKarrer:
@Jakub-CZ Yes, not your fault but Bisq's lack of maintaining the GH issues better.... too much work for too few devs ;-)

ripcurlx:
@Jakub-CZ Sorry about that! This was marked incorrectly as a `good first issue` by a GitHub issue maintainer.

PR #4980, opened 2020-12-19
Improve paragraph under "Wallet balances"

### LINKED ISSUE #4979: Improve wording in Wallet Info view
### Description

Very small correction in one message introduced in #4935.

(Sorry if it's overtly nitpicky, it bothered me a bit when translating the message.)

#### Version

149b52859ad539c066148ff6591a6d1229794608

### Steps to reproduce

This relates to the sentence:

> For Bitcoin the sum of the 'available balance' and the 'reserved for offers balance' must match the internal wallet balance \
  displayed here.

### Expected/Actual behaviour

- "_the sum ... must match the internal wallet balance displayed here_"
  - When Bisq says that something **must** be true then as a user I'd expect that I may have the means to make that happen. Is that really the case?
  - If not, then just a simple statement "_the sum ... matches ..._" may be more fitting here.
- The second balance referred in the sentence is in fact labeled as 'Reserved **in** offers'

### Screenshots

in #4935

#### Device or machine

N/A

Jakub-CZ:
Fixes #4979

Just a simple message improvement; see the issue.

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -1398,8 +1398,7 @@ account.menu.notifications=Notifications
 
 account.menu.walletInfo.balance.headLine=Wallet balances
 account.menu.walletInfo.balance.info=This shows the internal wallet balance including unconfirmed transactions.\n\
-  For Bitcoin the sum of the 'available balance' and the 'reserved in offers balance' matches the internal wallet balance \
-  displayed here.
+  For BTC, the internal wallet balance shown below should match the sum of the 'Available' and 'Reserved' balances shown in the top right of this window. \
 account.menu.walletInfo.xpub.headLine=Watch keys (xpub keys)
 account.menu.walletInfo.walletSelector={0} {1} wallet
 account.menu.walletInfo.path.headLine=HD keychain paths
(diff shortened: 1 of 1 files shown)


ripcurlx:
@m52go Could you please review the text changes? Thanks!

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1398,7 +1398,7 @@ account.menu.notifications=Notifications
 
 account.menu.walletInfo.balance.headLine=Wallet balances
 account.menu.walletInfo.balance.info=This shows the internal wallet balance including unconfirmed transactions.\n\
-  For Bitcoin the sum of the 'available balance' and the 'reserved for offers balance' must match the internal wallet balance \
+  For Bitcoin the sum of the 'available balance' and the 'reserved in offers balance' matches the internal wallet balance \
```
m52go: More clear to reference "BTC" instead of "Bitcoin" since balance below is labeled as "BTC".

Also, the label in top right is just labeled "Reserved" so it's probably better to use "Reserved" instead of "Reserved in offers" (I know the longer string shows on hover, but it's not immediately obvious, and most people probably don't hover).

```suggestion
  For BTC, the internal wallet balance shown below should match the sum of the 'Available' and 'Reserved' balances shown in the top right of this window. \
```

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1398,7 +1398,7 @@ account.menu.notifications=Notifications
...
   displayed here.
```
m52go: If the suggestion in https://github.com/bisq-network/bisq/pull/4980/files#r546901592 is accepted, then this line should be removed.

```suggestion

```

ripcurlx (approved):
utACK

PR #4981, opened 2020-12-20
[1.5.2] Fix wrong account age data access

chimp1984:
We need to user accountAgeWitnessStorageService as this supports the historical data store.
Using p2PService.getP2PDataStorage().getAppendOnlyDataStoreMap() would not deliver all the data.


sqrrm (approved):
utACK

PR #4982, opened 2020-12-20
Avoid standard shortcuts

sqrrm:
Change shortcut to copy account info for witness signing
to ctrl+shift+c to avoid conflicts with standard OS shortcuts.


chimp1984 (commented):
utACK

PR #4987, opened 2020-12-21
Remove redundant popup

chimp1984:
We had 2 times nearly the same popup after canceling a create or take offer when offer was funded.
I reused old show again key so ppl who had already used that dont see again the popup.


ripcurlx (approved):
utACK

ripcurlx (approved):
utACK

PR #4989, opened 2020-12-21
Adjust create transferwise acct test

ghubstan:
A new acct now has no default trade currencies.  Also fixed imports to not mix `jupiter` and `junit` asserts.


ripcurlx (approved):
utACK

PR #4991, opened 2020-12-21
Improve path text on wallet info screen

m52go:
Main goal was to make BSQ spending warning more clear. Also made minor wording improvement to other parts.


ripcurlx (approved):
utACK

PR #4999, opened 2020-12-24
Add generic map (cookie) to UserPayload

chimp1984:
Add generic map (cookie) to UserPayload for persisting UI states like app layout

This opens the app window in the same position and size as it has been before closing.
This new cookie field can be used as generic map container for UI state or not domain specific non critical data. It should NOT be seen as easy way to throw any kind of data in! Intention is to make it easier for persisting app state which is not important enough that developers do all the work with adding a proto field but still would be an improvement of UX.


ripcurlx (approved):
utACK

PR #5000, opened 2020-12-24
Implement Segwit for BSQ

stejbac:
This PR attempts to add support for native Segwit (p2wpkh) BSQ addresses.

The [btcd-cli4j](https://github.com/priiduneemre/btcd-cli4j) Json-RPC interface for bitcoind (which is used by DAO full nodes) is replaced, as that code appears to be unmaintained for about 5 years and only maintained compatibility up to bitcoind-0.10.3. In particular, it does not support the extra Segwit fields in the `getblock` Json responses returned to the Java client, including the `txinwitness` field of the raw tx inputs. These are needed to fill in the `TxInput.pubKey` fields of the inputs of each BSQ tx that makes up the DAO state, used in turn to validate staked merit (from issuance txs) and messages signed for a proof-of-burn tx. Thus the RPC client needs upgrading for us to support compensation and proof-of-burn txs with native Segwit BSQ inputs.

A new, home-grown RPC client is provided using the [jsonrpc4j](https://github.com/briandilley/jsonrpc4j) library. Additionally, a simple home-grown socket daemon is provided to receive block hashes sent by the _blocknotify_ script run by bitcoind, replacing the btcd-cli4j version.

- TODO: Improve the RPC client startup logic - should send a ping or `getnetworkinfo` message to bitcoind & check the version. **[edit: now done]**
- TODO: Fix a consensus issue caused by failure of _RpcService.java_ to account for at least one spurious fully native Segwit BSQ tx (d1f45e55be6101b1b75e6bf9fc5e5341c6ab420647be7555863bbbddd84e92f3) which appeared in block 660384, dated 2020-12-07 (thus missing a `TxInput.pubKey` field in the DAO state). I'm not sure yet if there are any other consensus issues. **[edit: now done]**
- TODO: Actually implement and enable the native segwit BSQ keychain and make sure there are no issues.
  **[edit: should be finished - created a separate PR #5109 for the changes]**


chimp1984:
Ah great to see work on that! I will have a look soon into it...

>     * TODO: Fix a consensus issue caused by failure of _RpcService.java_ to account for at least one spurious fully native Segwit BSQ tx (d1f45e55be6101b1b75e6bf9fc5e5341c6ab420647be7555863bbbddd84e92f3) which appeared in block 660384, dated 2020-12-07 (thus missing a `TxInput.pubKey` field in the DAO state). I'm not sure yet if there are any other consensus issues.

I assume some dev has created that tx. With the release code base its should not be possible to create a BSQ tx with segwit input.
The amount got burned:
https://mempool.emzy.de/de/bisq/tx/d1f45e55be6101b1b75e6bf9fc5e5341c6ab420647be7555863bbbddd84e92f3

chimp1984:
I reviewed the current state. Looks all good to me. Will do a more in depth review once its out of draft state.

core/src/main/java/bisq/core/dao/node/full/RpcService.java
```diff
@@ -246,14 +254,18 @@ private static RawTx getTxFromRawTransaction(RawTransaction rawDtoTx,
...
+                    String pubKeyAsHex = extractPubKeyAsHex(rawInput, rawInput == rawDtoTx.getVIn().get(0) &&
+                            !BSQ_TXS_DISALLOWING_SEGWIT_PUB_KEYS.contains(txId));
                     if (pubKeyAsHex == null) {
                         log.debug("pubKeyAsHex is not set as we received a not supported sigScript. " +
                                         "txId={}, asm={}, txInWitness={}",
                                 rawDtoTx.getTxId(), rawInput.getScriptSig().getAsm(), rawInput.getTxInWitness());
                     }
-                    return new TxInput(rawInput.getTxId(), rawInput.getVOut(), pubKeyAsHex);
+                    var txInput = new TxInput(rawInput.getTxId(), rawInput.getVOut(), pubKeyAsHex);
+                    txInput.isSegwit = rawInput.getTxInWitness() != null;
```
chimp1984: Wouldn't it be better to add that to the constructor so the TxInput is immutable?

stejbac:
Rebasing onto a more recent version of master (00c82bb7 on 2020/12/28, 1.5.2 merge, #5012), to avoid conflicts in build.gradle and gradle-witness.gradle caused by a bitcoinj upgrade...

stejbac:
I'm thinking it may be better to put the actual BSQ wallet changes in a separate PR, so that anyone running a full DAO node has a chance to upgrade to a Bisq version including these RPC changes first, before segwit BSQ txs start appearing which would otherwise break their consensus with the rest of the network. (So the above changes should now be ready for review.)

chimp1984:
Have you done a full DAO rebuild from genesis with that branch and check if the hashes in the monitor are the same as those from current seeds? It takes a while, probably several hours or so....

Have you seen my ealrier commnet at a review? ("Wouldn't it be better to add that to the constructor so the TxInput is immutable?")

stejbac:
I did a full DAO rebuild with an earlier version of the current branch. I should probably do another rebuild - I'll start that now.

TxInput should be immutable in the current version. There was an experimental `isSegwit` mutable flag that I added to it for testing and accidentally pushed that about a week ago, then amended and force pushed the intended commit (fdc739d) a little while later, so I think that dismissed your review comment (which I missed).

stejbac:
I've finished a DAO rebuild from genesis with the current branch and the hashes are still in agreement with the seed nodes.

chimp1984:
@stejbac I think it would be more safe if we fork the jsonrpc4j libray to our own repo and use that. What do you think?

core/src/main/java/bisq/core/dao/node/full/RpcService.java
```diff
@@ -246,8 +254,10 @@ private static RawTx getTxFromRawTransaction(RawTransaction rawDtoTx,
                     // To maintain backwards compatibility when serializing and hashing the DAO state,
                     // segwit pubKeys are only extracted for the first input, as this will always be a
                     // BSQ input. Later inputs might be segwit BTC, which would have had a null pubKey
-                    // recorded in the DAO state prior to the segwit upgrade of the RPC client.
-                    String pubKeyAsHex = extractPubKeyAsHex(rawInput, rawInput == rawDtoTx.getVIn().get(0));
+                    // recorded in the DAO state prior to the segwit upgrade of the RPC client. Spurious
+                    // segwit BSQ inputs in txs mined prior to the upgrade also require exclusion.
+                    String pubKeyAsHex = extractPubKeyAsHex(rawInput, rawInput == rawDtoTx.getVIn().get(0) &&
+                            !BSQ_TXS_DISALLOWING_SEGWIT_PUB_KEYS.contains(txId));
```
chimp1984: Wouldn't it be better to use a block height at which we activate the support for segwit inputs. It is technically a hard fork which would become activated at that block height.
That should also cover the past tx(s) with a segwit input.

chimp1984 (changes requested):
Beside the suggestion to use a block height for activating the change all looks good to me.

core/src/main/java/bisq/core/dao/node/full/RpcService.java
stejbac: (No longer applicable)

core/src/main/java/bisq/core/dao/node/full/RpcService.java
stejbac: Yes, perhaps that would be better. I'll push a change to do that shortly, as well as no longer excluding all but the first segwit input pubkey (to be on the safe side and consistent with the original pubkey extraction for P2PKH inputs).

chimp1984 (commented):
utACK

chimp1984 (commented):
utACK

ripcurlx:
@stejbac Could you please resolve the conflict? I hope it will trigger Travis as well (seems to be stuck somehow)

stejbac:
I resolved the conflict from the web interface this time. (Perhaps it would have been better to do a rebase - I'm not sure.)

core/src/main/java/bisq/core/dao/node/full/rpc/dto/RawBlock.java
```diff
@@ -0,0 +1,84 @@
...
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown = true, value = "ntx")
+@JsonPropertyOrder({"hash", "confirmations", "strippedsize", "size", "weight", "height", "version", "versionHex",
+        "merkleroot", "tx", "time", "mediantime", "nonce", "bits", "difficulty", "chainwork", "nTx",
+        "previousblockhash", "nextblockhash"})
+public class RawBlock {
```
sqrrm: The name overload of `full/RawBlock` and `full/rpc/dto/RawBlock` is a bit confusing.

ripcurlx:
@stejbac It would be great for this PR to get some information on how to test this best during release testing if it is getting merged for v1.5.5. Thanks!

chimp1984:
@ripcurlx I think it was not intended to get it into 1.5.5 but next release.

core/src/main/java/bisq/core/dao/node/full/rpc/dto/RawBlock.java
stejbac: I just took the names from the very similar DTO classes defined in the original btcd-cli4j library. Perhaps I could rename them from `Raw...` to `DtoBlock`, `DtoTransaction`, etc.

sqrrm: Yeah, I think that would make it clearer, adding a dto prefix or infix for each of the dto classes would make it clear what's going on. It is mostly clear as the variable names you've used include dto, but even better if the class names themselves also adhered to that pattern.

stejbac:
Rebasing onto a more recent version of master (d3971ef on 2021/01/16, #5072 merge), to avoid the earlier conflict in `RpcService` (caused by a package rename), instead of using a merge. Also adding a commit to address the `RawX` name clash mentioned above.

chimp1984 (dismissed):
utACK

sqrrm:
@stejbac Once you remove the draft tag I can merge this PR

chimp1984:
@sqrrm Lets coordinate before merge, we need to test that very well before it goes into master, also considering potential fork risks. Probably best to test it in a dedicated dev branch and only merge quickly before release to reduce time span where master and release have a difference.

sqrrm:
@chimp1984 that sounds reasonable. I'm still not able to review this PR but I will do when I'm back in business.

chimp1984:
Activation date is `ACTIVATE_HARD_FORK_2_HEIGHT_MAINNET = 672646;` which would be in about 12 days. I think we need to move that a bit more into the future, depending on release plan. I think most people have updated usually after 2 weeks of release. Though to get extra safety in case there are issues with the release (might be issues not related to that PR) I think a more conservative 4 weeks after release might be better.

@stejbac What would you think?

chimp1984:
@stejbac @ripcurlx @sqrrm @jmacxx 
We need to ensure that all seed nodes as well as explorer nodes have updated ahead in time before release. 
The `ACTIVATE_HARD_FORK_2_HEIGHT_MAINNET` need to be adjusted once release date is more clear.
I tested quite a lot but more testing would be good of course. Not updated nodes partizipating in DAO governance will get invalid issuance proposals in case those have been created with segwit addresses as we have a validation where a LegacyAddress is expected. We should communicate clearly to all DAO stakeholders/contributors that they must update before making proposals or doing voting. I don't see any risk for consensus issues but they would get locally out of sync and would need to resync the DAO state. To reduce risks I think we should try to set the activation date to a non critical period like the proposal period but avoid the more critical ones at voting/blind vote/result.

Thanks @stejbac for the great work!

@ripcurlx Can you arrange the compensation from the remaining funds from the Segwit sponsors? It was agreed that @stejbac will get what remains for that Segwit BSQ implementaion.

stejbac:
I think 4 weeks after the 1.5.7 release would be reasonable. Assuming the release is around a week from now, that would place the fork activation in the middle of the Cycle 23 proposal phase, away from the critical periods.

What about the BSQ wallet changes in #5109? Perhaps it would be OK if that PR is merged only into master after the release but before the fork activation, but if #5109 is merged into the release before the fork activation then users will start generating segwit-only BSQ UXTOs, which could get selected as proposal tx inputs. So some DAO participants may encounter problems if they try to make compensation requests early in Cycle 23, as any segwit BSQ tx inputs will be missing the necessary `pubKey` field - I believe this would cause the issued BSQ not to count towards that user's merit as it should (but haven't tested this). Similarly, problems would occur if a user (say the Burning Man) tried to burn any segwit BSQ in a proof-of-burn tx before the fork activation date.

chimp1984:
Ah yes, good point regarding the wallet PR. To me it seems it has to be held back as long the activation time is not triggered. Actually merging this PR asap with short term activation time should be ok as it does not represent much risk. All BSQ txs are still legacy we only support the new RPC and detect pubkeys from segwit txs. After a few weeks when most users have updated we can release #5109 so that segwit addresses will be used.

stejbac:
One possible issue with using a short activation time is that anyone running a DAO full node will have monitor hashes out of consensus with the rest of the network if they don't upgrade before the activation date, even though there would be no segwit BSQ txs yet. That's because the segwit BTC (non-BSQ) inputs will also start having nonnull `pubKey` fields after the activation date (though I don't believe that really affects anything other than the hashes).

chimp1984:
Ah ok. So lets play safe with sufficient time for activation and then as second step the wallet release.

Conza88:
General comms being made to prepare users?

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/5000#pullrequestreview-583820438

ripcurlx:
@stejbac As I'm preparing everything for the v1.6.0 pre-release right now. Is it really necessary that all seednodes are already updated before the pre-release as well? As the activation date is in the future and we don't ship the other Wallet PR yet, it looks to me as if this comment by @chimp1984 is related to the state where this PR contained also the Wallet code, am I wrong?

stejbac:
I don't think it's strictly necessary for the seed nodes to be updated at this point (but they should obviously be updated before the hard fork activation date). There shouldn't be any change in the network monitor hashes or any other loss of consensus before that date, unless there is a bug in the code - I didn't spot any such issues during testing on mainnet a month or two ago. I'll run another full DAO rebuild shortly to double check.

stejbac:
My full DAO node (running against Bitcoin Core 0.21.0) is still in consensus with the seed nodes after doing a rebuild from genesis. (I'm not sure if there has been an issue with any of the seed nodes recently, though, as my lite node fell out of consensus with them a few days ago for some reason - it wasn't running any of the 1.6.0 changes at the time.)

The current DAO state hash (from my full node) is 44d66d64... @ block height 676797. The proposal and blind vote hashes are also in sync with the seed nodes.

chimp1984:
I can confirm the DAO state hash on my local lite node.

wiz:
@chimp1984 @stejbac how should mempool.space display bech32 addresses? currently we add `B` in front: https://mempool.space/bisq/tx/d1f45e55be6101b1b75e6bf9fc5e5341c6ab420647be7555863bbbddd84e92f3

chimp1984:
I think the Bisq UI will handle it the same (B prefix)

Issue #4998, opened 2020-12-24
Deactivate offer if market price is above or below a user defined trigger price


chimp1984:
### Description

When using % based price there is a risk that if market price moves too drastically that the offer gets taken to a price which the maker would not like to accept. By adding a trigger price which deactivates the offer automatically if the market price hits that price we can protect makers against such price swings.

If its a sell offer the maker is interested to protect themself against too low prices. If its a buy offer the maker is interested to protect themself against too high prices. So only one field is needed.

Here is a WIP version of the implementation in the create offer view:
<img width="1197" alt="Screen Shot 2020-12-23 at 20 29 36" src="https://user-images.githubusercontent.com/54558767/103049030-88d78480-455e-11eb-8392-950eb858f02e.png">

In the open offer view at portfolio we will add a new column for the value and in edit offer it can be changed.

Re-activating the offer once the price exits the undesired range might be an option as well, but is not planned for the initial version.

Please leave comments if you have any inputs/feedback to that feature.

ripcurlx:
Closing as complete.

PR #4993, opened 2020-12-23
Add percentage price to open offer view

chimp1984:
Add percentage price to open offer view.
Extract code from offer book to PriceUtil class for reuse

PR #5001, opened 2020-12-25
Deactivate open offer if trigger price is reached

chimp1984:
Implements https://github.com/bisq-network/bisq/issues/4998

Based on https://github.com/bisq-network/bisq/pull/4993

### DIFF
--- a/core/src/main/java/bisq/core/offer/PriceEventHandler.java
+++ b/core/src/main/java/bisq/core/offer/PriceEventHandler.java
@@ -1,0 +1,148 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.core.offer;
+
+import bisq.core.locale.CurrencyUtil;
+import bisq.core.monetary.Altcoin;
+import bisq.core.monetary.Price;
+import bisq.core.provider.price.MarketPrice;
+import bisq.core.provider.price.PriceFeedService;
+
+import bisq.common.util.MathUtils;
+
+import org.bitcoinj.utils.Fiat;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+import javafx.collections.ListChangeListener;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+import lombok.extern.slf4j.Slf4j;
+
+import static bisq.common.util.MathUtils.roundDoubleToLong;
+import static bisq.common.util.MathUtils.scaleUpByPowerOf10;
+
+@Slf4j
+@Singleton
+public class PriceEventHandler {
+    private final OpenOfferManager openOfferManager;
+    private final PriceFeedService priceFeedService;
+    private final Map<String, Set<OpenOffer>> openOffersByCurrency = new HashMap<>();
+
+    @Inject
+    public PriceEventHandler(OpenOfferManager openOfferManager, PriceFeedService priceFeedService) {
+        this.openOfferManager = openOfferManager;
+        this.priceFeedService = priceFeedService;
+    }
+
+    public void onAllServicesInitialized() {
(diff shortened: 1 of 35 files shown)


desktop/src/main/java/bisq/desktop/main/PriceUtil.java
```diff
@@ -0,0 +1,151 @@
+/*
```
ghubstan, line 1: I instinctively need to ask if  `PriceUtil` could be moved to core.  I see dependencies on ChatView.log and ui validators, but much of the util logic could be made available to the api.

ghubstan (commented):
I do see the offer disabling trigger price in the edit offer view, [EDIT] and **can** edit it.

But I can't see the disabling trigger price in the published offer views (below).  Squeezing it into the list view might not be feasible, but adding it to the other two might be trivial.

Confirm Offer View..
where-is-trigger-price

Published Offer List View...
where-is-trigger-price-in-offer-list

Published Offer Detail View...
where-is-trigger-price-in-published-offer-detail

desktop/src/main/java/bisq/desktop/main/PriceUtil.java
chimp1984: Yes I was thinking the same... was just to lazy to separate pure domain code with formatting stuff which is using desktop scope classes... But yes we should refactor that.

chimp1984:
I was in the process to add a field there but as it only gets the offer passed but not the openOffer I was not sure... But probably its better to add it as well.
The triggerPrice is a field in openOffer not in Offer. Offer will be a part of trade and I think trigger price is a local setting of the user not a part of the mutual data which will be shared with the peer.

There is a trigger price column in openOffersView but it is hidden at default stage width as there is not enough space. If you scale larger it will get added. Similar we use in closed trades, there are a few detail columns only shown when width is larger. I am aware that many users might miss those hidden features, but have not found a good solution yet for it...
Maybe some icon on the table indicating that the table changes on resize, but as that is not a common UI feature I fear the icon will be just confusing...

chimp1984:
Also note that I changed/fixed the above/below term in the price field description. Before it was only changing for BUY/SELL but not for fiat/altcoin. But as fiat price and altcoin price have inverted logic I applied that there as well.
E.g. For a BTC seller a higher fiat price is good but a lower altcoin price.

ghubstan:
It's good that I can edit the trigger price... 
I am (reg) testing, so far so good.   
I wonder how difficult it would be to automatically re-enable an offer after the price goes back into the "enable range".
By the way, we shouldn't be doing this today ;-/

chimp1984:
Screen Shot 2020-12-25 at 00 33 23
Screen Shot 2020-12-25 at 00 09 10
Screen Shot 2020-12-25 at 00 34 08

chimp1984:
> I wonder how difficult it would be to automatically re-enable an offer after the price goes back into the "enable range".

Yes was also wondering if that might be good... Not sure also if we should show a popup once a offer gets deactivated. In th list view the shild icon gets yellow in that case. If price is still outside of permitted range the offer also cannot be re-activated directly (would trigger deactivation at next price update) but the user need to edit the offer... Not 100% sure if that is clear to users...

Also we do support that feature only for % based price, if you use fixed price you cannot set it as well if there is no price feed like BSQ. Wondering if it would make sense to allow it for those as well?

> By the way, we shouldn't be doing this today ;-/

Haha... I don't care much about xmas and holidays...

chimp1984:
If offer got deactivated user need to edit offer and see the validation error:
Screen Shot 2020-12-25 at 12 36 40

ghubstan:
> Haha... I don't care much about xmas and holidays...

We need time off just like religious people ;-)

> Not sure also if we should show a popup once a offer gets deactivated.

Not having one is OK for me.

> Wondering if it would make sense to allow it for those as well?

Yes, but maybe put that in another PR?

ghubstan:
> the shield icon gets yellow in that case..

Ah, I see.  Didn't know what that icon meant but the tooltip explains.   Good, better that taking up more real-estate with a 'Trigger Price' column header.  I think it's fine the way it is.

core/src/main/java/bisq/core/offer/PriceEventHandler.java
```diff
@@ -0,0 +1,148 @@
...
+        if (triggerPrice > 0) {
+            OfferPayload.Direction direction = openOffer.getOffer().getDirection();
+            boolean triggered = direction == OfferPayload.Direction.BUY ?
+                    marketPriceAsLong > triggerPrice :
+                    marketPriceAsLong < triggerPrice;
+            if (triggered) {
+                log.error("Market price exceeded the trigger price of the open offer. " +
+                                "We deactivate the open offer with ID {}. Currency: {}; offer direction: {}; " +
+                                "Market price: {}; Upper price threshold : {}",
+                        openOffer.getOffer().getShortId(),
```
ghubstan: Minor detail... placement of colons are not consistent (the last one).  
And you sure you wanted to use semi-colons to delimit key value pairs?  Is that consistent with other error log statements?

chimp1984: I think we have no style rule for that...

ghubstan (approved):
Tested ACK

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/5001#pullrequestreview-559171289

PR #5006, opened 2020-12-25
Display up-to-date error messages under text fields

### LINKED ISSUE #5005: Error messages under input fields are not updated
### Description

As user types into a validated text field, if the new string is still invalid but the error message (reason of invalidity) changed,
the message is not updated to reflect the new reason.

#### Version

c52d6691

### Steps to reproduce

1. DAO -> Governance->Make Proposal->Proposal for changing a parameter
2. Select BTC maker fee
3. While the current value is 0.001, gradually type `0.0011`. Notice the error messages displayed for each partially entered value.

### Expected behaviour

The error message reflects the reason why the value in the text field is invalid.

E.g.: going from **0.00** to **0.001** should switch from "_dust limit not allowed_" to "_must be different from current value_".

### Actual behaviour

The first error message that "snaps in" and is not updated when the reason for invalidity changes. One of the ways to get a correct message is to focus out of the field and then refocus the field again.

### Screenshots

Below are the states of a field as the value `0.0011` is being entered, and then cleared again.
instant-error-message-update BEFORE

#### Device or machine

Windows 10, OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.9.1+1)

Jakub-CZ:
As user types into a validated text field,
if the new string is still invalid but the error message changed,
ensure the new message will be displayed.

Fixes #5005

### Before
Below are the states of a field as the value `0.0011` is being entered, and then cleared again.

instant-error-message-update BEFORE

### After
Now, the error message indicates the error condition every step of the way.

instant-error-message-update AFTER

### DIFF
--- a/core/src/main/java/bisq/core/util/validation/InputValidator.java
+++ b/core/src/main/java/bisq/core/util/validation/InputValidator.java
@@ -21,6 +21,7 @@ import bisq.core.locale.Res;
 
 import java.math.BigInteger;
 
+import java.util.Objects;
 import java.util.function.Function;
 
 public class InputValidator {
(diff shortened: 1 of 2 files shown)


sqrrm (approved):
ACK

PR #5010, opened 2020-12-28
Dao performance improvements

chimp1984:


ripcurlx (approved):
utACK

But as agreed it needs the ACK from @ManfredKarrer

ManfredKarrer (approved):
utACK

PR #5014, opened 2020-12-28
Add indication of last GH CR issue

sqrrm:
Adds a marker for end of compensation request creation on github (currently set to 7 days before the end of the proposal phase).

See brighter gray in the proposal phase bar
image

See darker gray in the proposal phase bar
image


chimp1984 (approved):
utACK

ripcurlx (approved):
utACK

PR #5015, opened 2020-12-28
Fix missing apostrophe in Withdrawal view

Jakub-CZ:
In **FUNDS > SEND FUNDS**, labels "_Sender's amount_"/"_Receiver's amount_" are missing an apostrophe.

The single quote characters are present in the property file but they get eaten up even though there is no variable substitution taking place (at least it shouldn't be). In the code, however, variable values are provided for no reason.

### Before:
Send funds BEFORE

### After:
Send funds AFTER


ripcurlx (approved):
utACK - good Find!

PR #5016, opened 2020-12-28
Change trade period for transferwise from 1 day to 4 days

chimp1984:
If a bank transfer it takes longer than 1 day. 1 day works only for local TW to local TW account if the sender has already funds on their account.
The trade period is part of the offer, so all offers created with 1.5.2 or earlier will have still 1 day. Users don't need to change anything, the old account will use 4 days for any new offer.


ripcurlx:
@pazza83 Could you please ACK this as payment method maintainer? Thanks!

Conza88:
Does make sense. Can be up to 4.

pazza83:
ACK

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/5016#issuecomment-754324081

PR #5021, opened 2020-12-29
Adjust lombok annotations to reduce build warnings

ghubstan:
This change fixes the underlying problems behind five compiler warnings:  `Generating equals/hashCode implementation but without a call to superclass`.

Some core classes use `@Value` and `@Data` annotations where not necessary, for example, where `equals`, `hashCode`, and `toString` methods, and `final` class and field modifiers are defined by a developer.

From lombok classes and documents for the shortcut annotations [`@Value`](https://projectlombok.org/features/Value)  and [`@Data`]( https://projectlombok.org/features/Data), we find some implicit annotations applied by these shortcuts are not needed.  Adjustments are described below.

- `Tx` Replace `@Value` with `@Getter`.

- `TxOutput`  Replace `@Data` with `@Getter` and `@Setter`.

- `ChangeParamProposal`  Replace `@Value` with `@Getter`.

- `GetInventoryRequest`, `GetInventoryResponse`  An `@EqualsAndHashCode(callSuper = false) `is added to supress build warnings.


sqrrm (commented):
Makes me nervous changing this magic annotations in the DAO module. It compiles nicely and seems to work fine.

This is the typical case where @ManfredKarrer has to review to make sure this is not breaking some hashes we rely on in the DAO.

core/src/main/java/bisq/core/dao/state/model/blockchain/TxOutput.java
```diff
@@ -36,7 +37,8 @@
  * Gets persisted.
  */
 @Immutable
-@Data
+@Getter
+@Setter
```
chimp1984: There is no write access and its Immutable. The @Data was wrong and should have been a @Value annotation.
Can you remove the @Setter?

core/src/main/java/bisq/core/network/p2p/inventory/messages/GetInventoryRequest.java
```diff
@@ -21,9 +21,11 @@
 import bisq.common.app.Version;
 import bisq.common.proto.network.NetworkEnvelope;
 
+import lombok.EqualsAndHashCode;
 import lombok.Value;
 
 @Value
```
chimp1984: That should also be replaced by a @Getter and @ToString.

core/src/main/java/bisq/core/network/p2p/inventory/messages/GetInventoryResponse.java
```diff
@@ -28,9 +28,11 @@
 import java.util.HashMap;
 import java.util.Map;
 
+import lombok.EqualsAndHashCode;
 import lombok.Value;
 
 @Value
```
chimp1984: That should also be replaced by a @Getter and @ToString.

chimp1984 (changes requested):

core/src/main/java/bisq/core/dao/state/model/blockchain/TxOutput.java
ghubstan: Setter removed in commit c2c1ac2087a8862dbf02007adf299aba460f92f5.

core/src/main/java/bisq/core/network/p2p/inventory/messages/GetInventoryRequest.java
ghubstan: Commit  c2c1ac2 
Removes `@Value`, `@EqualsAndHashCode(callSuper = false)`
Adds `@Getter`,`@ToString`

I'm not absolutely certain this is what you asked for (by removing `@Value`), but keeping `@Value` , `@Getter`, `@ToString` is redundant because `@Getter`, `@ToString` are implicitly defined by `@Value`.

core/src/main/java/bisq/core/network/p2p/inventory/messages/GetInventoryResponse.java
ghubstan: Same as above...  commit  c2c1ac2 
Removes `@Value`, `@EqualsAndHashCode(callSuper = false)`
Adds `@Getter`,`@ToString`

I'm not absolutely certain this is what you asked for (by removing `@Value`), but keeping `@Value` , `@Getter`, `@ToString` is redundant because `@Getter`, `@ToString` are implicitly defined by `@Value`.

core/src/main/java/bisq/core/network/p2p/inventory/messages/GetInventoryRequest.java
```diff
@@ -21,11 +21,11 @@
 import bisq.common.app.Version;
 import bisq.common.proto.network.NetworkEnvelope;
 
-import lombok.EqualsAndHashCode;
-import lombok.Value;
+import lombok.Getter;
+import lombok.ToString;
 
-@Value
-@EqualsAndHashCode(callSuper = false)
```
chimp1984: EqualsAndHashCode should be left

core/src/main/java/bisq/core/network/p2p/inventory/messages/GetInventoryResponse.java
```diff
@@ -28,11 +28,11 @@
 import java.util.HashMap;
 import java.util.Map;
 
-import lombok.EqualsAndHashCode;
-import lombok.Value;
+import lombok.Getter;
+import lombok.ToString;
 
-@Value
-@EqualsAndHashCode(callSuper = false)
```
chimp1984: EqualsAndHashCode should be left

core/src/main/java/bisq/core/network/p2p/inventory/messages/GetInventoryRequest.java
chimp1984: I meant that @Value produces Lombok warnings as well so replace it with EqualsAndHashCode, Getter and ToString

core/src/main/java/bisq/core/network/p2p/inventory/messages/GetInventoryRequest.java
ghubstan: commit 0c9c96165b2df15d9bd4c8a671d3083192f11403

core/src/main/java/bisq/core/network/p2p/inventory/messages/GetInventoryResponse.java
ghubstan: commit 0c9c96165b2df15d9bd4c8a671d3083192f11403

core/src/main/java/bisq/core/network/p2p/inventory/messages/GetInventoryResponse.java
chimp1984: Ah sorry I see now the confusion as the marked line was EqualsAndHashCode instead of Value. My fault... sorry...

ManfredKarrer (approved):
utACK

sqrrm (approved):

PR #5026, opened 2020-12-30
Add dos protection

chimp1984:


wiz (approved):
ACK - tested okay on a few lite nodes, they started up very quickly

sqrrm (approved):
utACK

PR #5030, opened 2020-12-31
[1.5.4] We need to use the getMapOfAllData method for HistoricalDataStoreServices

chimp1984:
The map access to HistoricalDataStoreServices from the p2pDataStorage was incorrect. I has accessed the live map instead of all data. That has led that at addPersistableNetworkPayload the data got rebroadcast as it was assumed the data has not been already received.

How to test it?
- Add some logs in P2PDataStorage.addPersistableNetworkPayload to see if data gets added and broadcast
- Change the  delay at republishAllFiatAccounts in AccountAgeWitnessService otherwise it takes 20-60 sec. for republishing
- Create a new fresh regtest seed and new fresh app node. 
- Start the app. Create one fiat payment account. Shut down and copy the AccountAgeWitnessStore which must have 1 entry now and copy it to the resources and rename it to AccountAgeWitnessStore_1.5.0_BTC_REGTEST
- Restart seed node and app
- Now you should see repeated logs that the data gets added and broadcast

It is still not clear to me why we did not see much more AccountAgeWitness data but only one in massive numbers...

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/storage/persistence/AppendOnlyDataStoreService.java
+++ b/p2p/src/main/java/bisq/network/p2p/storage/persistence/AppendOnlyDataStoreService.java
@@ -78,7 +78,12 @@ public class AppendOnlyDataStoreService {
 
     public Map<P2PDataStorage.ByteArray, PersistableNetworkPayload> getMap() {
         return services.stream()
-                .flatMap(service -> service.getMap().entrySet().stream())
+                .flatMap(service -> {
+                    Map<P2PDataStorage.ByteArray, PersistableNetworkPayload> map = service instanceof HistoricalDataStoreService ?
+                            ((HistoricalDataStoreService) service).getMapOfAllData() :
+                            service.getMap();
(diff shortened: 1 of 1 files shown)


chimp1984:
>It is still not clear to me why we did not see much more AccountAgeWitness data but only one in massive numbers...

I found out the reason:
AccountAgeWitness has 1 day date tolerance, if its older the addPersistableNetworkPayload returns with false and do not relay the data. This helps us as the data which flooded the network is now > 1 day so it stopped flooding. This is also the reason why not all AccountAgeWitness data from hsitorical data store did flood the network. They got though republished but they got ignored by all peers as the date was too old.

The hash of that data of the problematic item was in the historical data store with an older date, but the user had for some reason not that historical data store or lost the live data so he could not find his AccountAgeWitness with the original date and created a new one with the date of 29.12. The other network peers had that hash in historical data but due the bug fixed in that PR it was not found so it got processed further and relayed again.

sqrrm (approved):
utACK

PR #5033, opened 2020-12-31
[1.5.4] Only publish witness if we are in date tolerance

chimp1984:


chimp1984:
I think that one would be a nice to have for 1.5.4. as well. Trivial change... But up to maintainers...

sqrrm (approved):
utACK

PR #5037, opened 2021-01-01
Update translations and fix a broken German translation

ripcurlx:
Problematic key in German translation: dao.wallet.send.sendFunds.details

This was broken since version v1.5.2 and prevents users to send BSQ when German is selected.
Current workaround is to switch to English for this task.


sqrrm (approved):
utACK

PR #5038, opened 2021-01-01
Add filter support on network level

chimp1984:
Adds a new field to the filter window for banning nodes on the network level. The existing one is only for banning nodes from trading. 

Based on #5031
Starts at commit 15b8f64

### DIFF
--- a/core/src/main/java/bisq/core/filter/Filter.java
+++ b/core/src/main/java/bisq/core/filter/Filter.java
@@ -35,6 +35,7 @@ import java.security.PublicKey;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
@@ -47,7 +48,10 @@ import javax.annotation.Nullable;
 @Value
 public final class Filter implements ProtectedStoragePayload, ExpirablePayload {
     private final List<String> bannedOfferIds;
+
+    // Those banned nodes cannot trade but still can access the network
     private final List<String> bannedNodeAddress;
+
     private final List<String> bannedAutoConfExplorers;
     private final List<PaymentAccountFilter> bannedPaymentAccounts;
     private final List<String> bannedCurrencies;
@@ -91,6 +95,11 @@ public final class Filter implements ProtectedStoragePayload, ExpirablePayload {
     // added at v1.3.8
     private final boolean disableAutoConf;
 
+    // added at v1.5.5
+    // In contrast to bannedNodeAddress those addresses cannot access the network but are rejected immediately.
+    // Should be only used in case of dos attacks
+    private final Set<String> nodeAddressesBannedFromNetwork;
+
     // After we have created the signature from the filter data we clone it and apply the signature
     static Filter cloneWithSig(Filter filter, String signatureAsBase64) {
         return new Filter(filter.getBannedOfferIds(),
@@ -117,7 +126,8 @@ public final class Filter implements ProtectedStoragePayload, ExpirablePayload {
                 filter.getSignerPubKeyAsHex(),
                 filter.getBannedPrivilegedDevPubKeys(),
                 filter.isDisableAutoConf(),
-                filter.getBannedAutoConfExplorers());
+                filter.getBannedAutoConfExplorers(),
+                filter.getNodeAddressesBannedFromNetwork());
     }
 
     // Used for signature verification as we created the sig without the signatureAsBase64 field we set it to null again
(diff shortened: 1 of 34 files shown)


p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
```diff
@@ -256,13 +250,21 @@ private void logContents(GetDataResponse getDataResponse) {
...
         sb.append("#################################################################");
         log.info(sb.toString());
     }
 
+    private void addDetails(Map<String, Tuple2<AtomicInteger, AtomicInteger>> numPayloadsByClassName,
+                            NetworkPayload networkPayload, String className) {
+        numPayloadsByClassName.putIfAbsent(className, new Tuple2<>(new AtomicInteger(0),
+                new AtomicInteger(0)));
+        numPayloadsByClassName.get(className).first.getAndIncrement();
+        numPayloadsByClassName.get(className).second.getAndAdd(networkPayload.toProtoMessage().getSerializedSize());
```
sqrrm, line 268: Might be worth adding a comment that this log info comes at a cost for in case it turns out to be a problem in the future

p2p/src/test/java/bisq/network/p2p/DummySeedNode.java
```diff
@@ -137,10 +135,11 @@ public void processArgs(String[] args) {
                     checkArgument(arg.contains(":") && arg.split(":").length > 1 && arg.split(":")[1].length() > 3,
                             "Wrong program argument " + arg);
                     List<String> list = Arrays.asList(arg.split(","));
+                    Set<NodeAddress> bannedPeers = new HashSet<>();
```
sqrrm: This set isn't used so why even gather the banned peers?

core/src/main/java/bisq/core/filter/Filter.java
```diff
@@ -91,6 +95,11 @@
     // added at v1.3.8
     private final boolean disableAutoConf;
 
+    // added at v1.5.5
+    // In contrast to bannedNodeAddress those addresses cannot access the network but are rejected immediately.
+    // Should be only used in case of dos attacks
+    private final Set<String> nodeAddressesBannedFromNetwork;
```
sqrrm, line 96: Naming is confusing since there is a `bannedNodeAddress` and a `nodeAddressesBannedFromNetwork`, better to have more descriptive names.

proto/src/main/proto/pb.proto
```diff
@@ -677,6 +677,7 @@ message Filter {
     repeated string bannedPrivilegedDevPubKeys = 23;
     bool disable_auto_conf = 24;
     repeated string banned_auto_conf_explorers = 25;
+    repeated string node_addresses_banned_from_network = 26;
```
sqrrm, line 680: Filter hash won't be backwards compatible so care is needed when this is rolled out

sqrrm (commented):
Looks good in general but esp naming comment would be good to fix to avoid future confusion

p2p/src/test/java/bisq/network/p2p/DummySeedNode.java
chimp1984: Agree. Did just quick adjustment to changes...

p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataHandler.java
chimp1984: Yes can add a comment

core/src/main/java/bisq/core/filter/Filter.java
chimp1984: I think `nodeAddressesBannedFromNetwork` is pretty clear (and long ;-)) no? The `bannedNodeAddress` is not great but renaming it would require a rename if the protobuf field as well to avoid that those fields get out of sync and was not sure if thats worth it. Also not clear for a good alternative name.

proto/src/main/proto/pb.proto
chimp1984: Ah true. Damn... We can keep old filter and create a new one. Or use extraMap (not great)....
There will be more addition to filter, so I guess to handle multiple filters is best. Most users update quickly so we do not need to maintain the old filters forever...

wiz:
This is dangerous because if you accidentally filter out all nodes the entire P2P network would go down immediately and no way to fix it - maybe include some protection for that

chimp1984:
> This is dangerous because if you accidentally filter out all nodes the entire P2P network would go down immediately and no way to fix it - maybe include some protection for that

You have to add a list of onions, we cannot even know all nodes... And the filter can be ignored by a prog arg (--ignoreDevMsg).

core/src/main/java/bisq/core/filter/Filter.java
chimp1984: Renamed it to nodeAddressesBannedFromTrading

sqrrm: Much better this way, no need for comments since code is self describing. Agreed that it was the `bannedNodeAddress` that needed a name change.

sqrrm (approved):
utACK

PR #5041, opened 2021-01-02
Request only non-seed nodes if no seed nodes are available

chimp1984:
In case requests to all seed nodes fail we take persisted and reported peers for requests.
Before we added also the seed nodes which would delay connection to other nodes a lot.


sqrrm (approved):
utACK

PR #5042, opened 2021-01-02
Select show-all currencies if TransferWise is selected

chimp1984:
If we select TransferWise we switch to show all currencies as TransferWise supports sending to most currencies.

We could make a different handling for BUY and SELL side as sending to nearly all currencies is possible but not receiving (only the currencies which have been enabled in the account). But I think that would confuse the users and would appear like a bug as the reason is not trivial to see.


sqrrm (approved):
utACK

Conza88:
Possible to do this for Revolut as well?

PR #5045, opened 2021-01-02
Add option to hide non supported payment methods

chimp1984:
Add option in preferences to hide payment methods from combobox in the Offer views which are not part of the users accounts.

Default value is false, so same behaviour as before that PR.
If no payment account is setup then we also show all payment methods.
In that case (no payment account) we disable the toggle as well as show it in false state.

Based on #5042 
Starts at commit eddb7cb

### DIFF
--- a/core/src/main/java/bisq/core/user/Preferences.java
+++ b/core/src/main/java/bisq/core/user/Preferences.java
@@ -487,6 +487,11 @@ public final class Preferences implements PersistedDataHost, BridgeAddressProvid
         });
     }
 
+    public void setHideNonAccountPaymentMethods(boolean hideNonAccountPaymentMethods) {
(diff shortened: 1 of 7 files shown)


sqrrm (approved):
ACK

Took me a while to figure it was just the combo box this filtered. Should've read the PR description more thoroughly.

Conza88:
Kind of confused what this changes exactly? Photo may help? 

The toggle already removes (hides) non supported payment methods? 

image

ripcurlx:
> Kind of confused what this changes exactly? Photo may help?
> 
> The toggle already removes (hides) non supported payment methods?
> 
> image

Using that toggle hides all `greyed out` offers you can't take as you don't have a corresponding payment account to do so.

PR #5047, opened 2021-01-02
Add cache for signature verification results and a lookup map by ownerPubKey

chimp1984:
When adding a new filter to the offer book it became has shown that there are some performance bottlenecks in the signedWitness domain. With Profiling 2 methods have been found causing the main costs.

Before that optimisation the new filter took about 265 ms (200 ms for the signature checks and 65 ms for the iteration testing for same ownerPubKeys. After the optimisation it is about 15 ms. 

This code was also called already by the item renderer for the take offer button to show if it is enabled or not. So the offer book becomes faster as well.

### DIFF
--- a/core/src/main/java/bisq/core/account/sign/SignedWitnessService.java
+++ b/core/src/main/java/bisq/core/account/sign/SignedWitnessService.java
@@ -82,6 +82,17 @@ public class SignedWitnessService {
     // This avoids iterations over the signedWitnessMap for getting the set of such SignedWitnesses.
     private final Map<P2PDataStorage.ByteArray, Set<SignedWitness>> signedWitnessSetByAccountAgeWitnessHash = new HashMap<>();
 
+    // Iterating over all SignedWitnesses and do a byte array comparison is a bit expensive and
+    // it is called at filtering the offer book many times, so we use a lookup map for fast
+    // access to the set of SignedWitness which match the ownerPubKey.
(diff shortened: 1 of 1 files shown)


Conza88:
Epic 👌🥇

core/src/main/java/bisq/core/account/sign/SignedWitnessService.java
```diff
@@ -393,10 +417,15 @@ private boolean verifySignatureWithDSAKey(SignedWitness signedWitness) {
...
-        return getSignedWitnessMapValues().stream()
-                .filter(e -> Arrays.equals(e.getWitnessOwnerPubKey(), ownerPubKey))
-                .filter(e -> !excluded.contains(new P2PDataStorage.ByteArray(e.getSignerPubKey())))
-                .collect(Collectors.toSet());
+        P2PDataStorage.ByteArray key = new P2PDataStorage.ByteArray(ownerPubKey);
+        if (signedWitnessSetByOwnerPubKey.containsKey(key)) {
+            return signedWitnessSetByOwnerPubKey.get(key).stream()
+                    .filter(e -> !excluded.contains(new P2PDataStorage.ByteArray(e.getSignerPubKey())))
+                    .collect(Collectors.toSet());
+
```
sqrrm, line 425:
```suggestion
```

sqrrm (approved):
ACK

PR #5048, opened 2021-01-02
Add toggle for hiding not takable offers

chimp1984:
Adds a toggle to the offer book views to filtering offers which can be taken from users accounts. It uses the same code which is used for the disable state for take offer buttons, so all the greyed out offers are hidden if toggle is selected. It can be combined with the other 2 dropbox filters.

We cache some method results for performance reasons. This also improves the performance without selecting the toggle.
https://github.com/bisq-network/bisq/pull/5047 will add more performance improvements which will speed up the first calls to the expensive methods further.

<strike>Based on #5045. Commits start at 26c1c0e </strike>
Rebased on master.

### DIFF
--- a/core/src/main/java/bisq/core/user/Preferences.java
+++ b/core/src/main/java/bisq/core/user/Preferences.java
@@ -772,6 +772,11 @@ public final class Preferences implements PersistedDataHost, BridgeAddressProvid
         requestPersistence();
     }
 
+    public void setShowOffersMatchingMyAccounts(boolean value) {
+        prefPayload.setShowOffersMatchingMyAccounts(value);
+        requestPersistence();
+    }
+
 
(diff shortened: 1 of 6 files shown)


chimp1984:
Screen Shot 2021-01-02 at 18 38 22
Screen Shot 2021-01-02 at 18 38 30

Conza88:
Ha, excellent. Possible to include own offers and show in 'matching my accounts'? 
I suspect - yes - that would be wanted? Although not technically 'takeable' (but can action i.e. 'remove') and is kind of their account? 

Also - can you filter between what is currently available in your accounts? e.g. toggle is on for 'takeable' / 'matching' ... and you want to only see "Transferwise"? That works?

PR #5052, opened 2021-01-04
Add a 'payment method details' screen

### LINKED ISSUE #5049: Add a 'payment method details' screen in Bisq client
### Description

Bisq has a section that shows active offers per currency. It is in a screen called 'details'

It would be great if Bisq had a section that shows active offers per payment  method. 

I proposed the current screen called 'details' to be renamed to ' currency details' and a new screen to be added called 'payment method details'

### Benefits

It would be great for Bisq users to see how many deals are available for what particular types of payment methods eg; Zelle, Revolut, TransferWise.

This would help traders see what is currently popular / available. It might prompt traders to create / add payment methods that are popular to their Bisq client. 

It would also be really useful to my role of Payment Maintainer when generating reports of payment method activity. 

#### Version

1.5.4

### Actual behaviour

On Bisq client

Market > Details

### Expected behaviour

On Bisq client there to be two options

Market > Payment Method Details

Market > Currency Details

### Screenshots

Add payment method details

ghost:
Fixes #5049

image

---

image

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -292,8 +292,8 @@ mainView.version.update=(Update available)
 ####################################################################
 
 market.tabs.offerBook=Offer book
-market.tabs.spreadCurrency=Currency Details
-market.tabs.spreadPayment=Payment Method Details
+market.tabs.spreadCurrency=Offers by Currency
+market.tabs.spreadPayment=Offers by Payment Method
 market.tabs.trades=Trades
 
 # OfferBookChartView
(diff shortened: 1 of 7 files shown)


core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -292,7 +292,8 @@ mainView.version.update=(Update available)
 ####################################################################
 
 market.tabs.offerBook=Offer book
-market.tabs.spread=Details
+market.tabs.spreadCurrency=Currency Details
```
chimp1984: Maybe "Offers by currency" and "Offers by Payment Method" would be better?

chimp1984 (dismissed):
utACK

chimp1984:
That was fast! ;-)

sqrrm (approved):
utACK

PR #5056, opened 2021-01-05
Add new api methods 'getmyoffers' and 'getmyoffer'

ghubstan:
Similar to `getoffers` and `getoffer`, but filters out offers not created by the user.  The new methods are so similar some offer list filtering and sorting was refactored in `CoreOffersService`.

Also fixed some createoffer apitest cases in anticipation of a new [`OfferFilter`](https://github.com/bisq-network/bisq/pull/5053), which will filter out offers not matching any user payment account.

### DIFF
--- a/core/src/main/java/bisq/core/api/CoreOffersService.java
+++ b/core/src/main/java/bisq/core/api/CoreOffersService.java
@@ -40,6 +40,7 @@ import java.math.BigDecimal;
 import java.util.Comparator;
 import java.util.List;
 import java.util.function.Consumer;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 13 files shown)


core/src/main/java/bisq/core/api/CoreOffersService.java
```diff
@@ -79,23 +84,28 @@ Offer getOffer(String id) {
...
-            offers.sort(Comparator.comparing(Offer::getPrice).reversed());
-        else
-            offers.sort(Comparator.comparing(Offer::getPrice));
-
+    List<Offer> getMyOffers(String direction, String currencyCode) {
+        List<Offer> offers = offerBookService.getOffers().stream()
+                .filter(o -> o.isMyOffer(keyRing))
+                .filter(o -> offerMatchesDirectionAndCurrency(o, direction, currencyCode))
+                .collect(Collectors.toList());
+        sortOffers(offers, direction);
```
sqrrm: Why not sort the offers in the stream, wouldn't that be more efficient? To make it more readable you could make a method `offerComparator`.

sqrrm (dismissed):
utACK

@ghubstan There's a comment I'll give you a chance to reply to before merging.

core/src/main/java/bisq/core/api/CoreOffersService.java
ghubstan: Like this 9689c3edfcb04a20b97718c64a71d0aed56b9a4e ?

ghubstan:
force travis jar download

core/src/main/java/bisq/core/api/CoreOffersService.java
sqrrm: Yeah. Since the stream is there anyway I'm thinking it would be possible to do some optimization by the stream interface, but I'm not sure. It's also easier to follow all the operations on Collection when lined up in stream operations. The Collection.sort() is probably more efficient as a stand alone sort method though.

sqrrm (approved):
utACK

PR #5059, opened 2021-01-06
Fix log statement format

ghubstan:
Fix log statement format

A log statement was mismatching argument placeholders with argument values because it was formatted as 
    `log.info("msg={}, \n" + args...)`	
instead of
    `log.info("msg={}", args...)`

Some code formatting was also done to this block, and a closer approximation of a CLI `createoffer` param list replaces the next  log statement.


ripcurlx (approved):
utACK

PR #5061, opened 2021-01-07
Persist failed attempts of decrypting mailbox messages

chimp1984:
We persist failed attempts to decrypt mailbox messages (expected if mailbox message was not addressed to us).
This improves performance at processing mailbox messages.
On a fast 4 core machine 1000 mailbox messages take about 1.5 second. At second start-up using the persisted data
it only takes about 30 ms.
We clean up old entries at startup using the TTL field.


ripcurlx (approved):
utACK

PR #5063, opened 2021-01-08
Fix two CLI option parsing bugs

ghubstan:
- Fix bug locating index of method name in String[] args when handling any negative number opt.

- Make `createoffer` method's security-deposit param format consistent with UI's.
When creating an offer, the CLI should take "15.0", not "0.15" as a 15% security deposit.  This is consistent with the UI, and the CLI's mkt-price-margin input format.


sqrrm (approved):
utACK

PR #5065, opened 2021-01-08
Integrate new protection tools into api's offer & trade services

ghubstan:
- Injected `OfferFilter` into `CoreOffersService` and `CoreTradesService`.
The filter constrains `getoffer(s)` results to offers an api user can take, as the first part of ongoing protection tools / api integration.

- Created a new `CoreContext` singleton.
Initially, lets `Core*Services` know if the user is using the api -- `isApiUser=true` if the current thread's name is "BisqDaemonMain", or the name contains "grpc".  We cannot check `BisqDaemonMain.class.getSimpleName()` because `:core` cannot have a circular dependency on `:daemon`, but there is probably a better way to do this than depending on the thread name set in `BisqDaemonMain#configUserThread()`.  We do this anticipating future `:desktop` dependencies on the core api, and we don't pass hardcoded `isApiUser=true` params to lower level domain objects.  

- Added `@Singleton` annotation to all Core*Service classes.
      
Related to https://github.com/bisq-network/bisq/pull/5053.

PR https://github.com/bisq-network/bisq/pull/5063 should be reviewed and merged before this one.

### DIFF
--- a/core/src/main/java/bisq/core/api/CoreContext.java
+++ b/core/src/main/java/bisq/core/api/CoreContext.java
@@ -1,0 +1,39 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.core.api;
+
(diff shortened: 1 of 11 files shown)


ghubstan:
I made a mistake by making the CLI method name a posix-style opt.  I will change it to be the only positional parameter in a command.

core/src/main/java/bisq/core/api/CoreContext.java
```diff
@@ -0,0 +1,39 @@
...
+@Slf4j
+class CoreContext {
+
+    @Inject
+    public CoreContext() {
+    }
+
+    public boolean isApiUser() {
+        String threadName = currentThread().getName();
+        return threadName.equals("BisqDaemonMain") || threadName.contains("grpc");
```
sqrrm: I find this ugly. There must be some easier and not quite as ugly way to do this.

sqrrm (commented):
It hurts my soul somewhat with that thread parsing to figure out context. I think we should try to think a bit more before merging something like that.

core/src/main/java/bisq/core/api/CoreContext.java
ghubstan: Resolved by commit 0e779a4

sqrrm (approved):
utACK

I much prefer this solution

PR #5066, opened 2021-01-08
Add toggle for displaying volume in tradestatistics chart in USD

chimp1984:
Adds a toggle to show trade statistics volume in USD instead of BTC

Screen Shot 2021-01-08 at 16 43 00
Screen Shot 2021-01-08 at 16 43 06

### DIFF
--- a/core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java
+++ b/core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java
@@ -50,6 +50,9 @@ import org.bitcoinj.utils.Fiat;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Charsets;
 
+import java.time.LocalDateTime;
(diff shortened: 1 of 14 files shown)


Conza88:
Does it have to be USD? 😅
Anyway to provide fiat choice without being too hard?

chimp1984:
It need to be fiat currency sufficiently trades, so EUR or USD are the only one which always got some volume. We calculate the historical price based on those trades. The calculation is a bit expensive so I dont want to do it for multiple currencies. Also for most USD price is kind of the main price metric. Does not mean any endoresement of USD of course - the biggest shitcoin in fiat land. ;-)

sqrrm (approved):
utACK

cbeams:
Glad to see this feature. I just tweeted about it here: https://twitter.com/cbeams/status/1357298837052866560, and added a a follow-on feature request at #5154.

PR #5068, opened 2021-01-09
Seed node branch

chimp1984:
Based on #5066


chimp1984:
Will rebase on master once #5066 is merged

PR #5070, opened 2021-01-10
Improve cleanup tor dir at seeds

chimp1984:
If all connections lost at seed nodes we clean tor cache files and trigger a restart.
We set a flag to clear tor cache files at shutdown and at startup we check for that flag and clean the tor cache before tor is started. Before we did that at shutdown but that can cause problems as tor is running at that moment.

Based on #5066 (to avoid merge conflicts, as we added an entry to CookieKey there)
PR starts at 8ecfa8bcbad204c54f413c4693c6059e8ded403a

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/market/trades/TradesChartsViewModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/market/trades/TradesChartsViewModel.java
@@ -61,7 +61,6 @@ import javafx.util.Pair;
 
 import java.time.LocalDateTime;
 import java.time.ZoneId;
-import java.time.ZonedDateTime;
 import java.time.temporal.ChronoUnit;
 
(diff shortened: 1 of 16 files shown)


desktop/src/main/java/bisq/desktop/main/market/trades/TradesChartsViewModel.java
```diff
@@ -116,7 +117,7 @@
     final int maxTicks = 90;
     private int selectedTabIndex;
     final Map<TickUnit, Map<Long, Long>> usdPriceMapsPerTickUnit = new HashMap<>();
-
+    private boolean fillTradeCurrenciesOnActiavetCalled;
```
sqrrm, line 120: spelling

sqrrm (approved):
utACK

PR #5071, opened 2021-01-10
Show a confirmation of successfully sending BTC or BSQ from wallet

### LINKED ISSUE #5067: Disorienting warnings during BSQ sending of funds
### Description

After I took an offer to sell BSQ and asked to start the transfer of funds, I entered the BSQ amount and the buyers address in the appropriate fields in the SEND page of the DAO tab and I pressed "SEND BSQ FUNDS.
A pop up window appeared asking me to enter my password to verify the sending of funds.
After I entered correct the pass, the fields "Receiver's BSQ address" and "BSQ amount to sent" became red informing me that "empty input is not allowed".
That made me think that for some reason after I entered my pass the fields were cleared and I had to re-enter the values.
I did it and I sent the same amount to the same address twice!
Can you please correct that and instead of making red the fields, show a message of successful sending?

#### Version

1.5.4

### Steps to reproduce

- Take offer to sell BSQ
- Go to DAO tab
- Enter amount and receiver's address in appropriate fields and press "Send BSQ Funds"
- Pop up window appears asking for password to verify sending
- Enter password correctly
- Password pop up window closes and fields "Receiver's Address" and "Amount to Withdraw" immediately become red with a notice underneath "empty input is not allowed".

### Expected behaviour

It would be more clear if instead of turning the fields red and showing the "empty input is not allowed" notice, for a pop up window to appear informing whether the password was correct and the withdrawal successful (or initiated). Maybe even a prompt to check the "Transactions" tab and see if the transaction has started.

### Actual behaviour

The red fields and the warning notice under the fields gives the impression that something went wrong. In my case I thought that I had to first enter my password and then enter the input in the fields, so I did the withdrawal twice.

### Screenshots

Red_Fields

#### Device or machine

Desktop PC, Windows 10

#### Additional info

I can always upload the bisq log files if it is necessary.

Thanks for your help!

ghost:
As requested, this shows a message of successful sending.
The confirmation details include amount, recipient address and txId.
Includes a link to open txId details in the user's external block explorer.
To be consistent it does the same for both BSQ and BTC.

Fixes #5067 

### Screenshots:

SendBsqTxId

Note the technicality with sending BSQ is you cannot look up the details in a BSQ explorer until +1 bitcoin confs have passed.

@m52go any thoughts on improving the note text shown in the above popup?

https://github.com/jmacxx/bisq/blob/8f24f0ab9c6d12143c09ecd1bb3bcbaa082bb3e4/core/src/main/resources/i18n/displayStrings.properties#L2699

--- 

SendBtcTx

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -2693,6 +2693,13 @@ tradeDetailsWindow.tradeState=Trade state
 tradeDetailsWindow.agentAddresses=Arbitrator/Mediator
 tradeDetailsWindow.detailData=Detail data
 
+txDetailsWindow.headline=Transaction Details
+txDetailsWindow.btc.note=You have sent BTC.
+txDetailsWindow.bsq.note=You have sent BSQ funds.  \
+  BSQ is colored bitcoin, so the transaction will not show in a BSQ explorer until it has been confirmed in a bitcoin block.
+txDetailsWindow.sentTo=Sent to
+txDetailsWindow.txId=TxId
+
 walletPasswordWindow.headline=Enter password to unlock
 
 torNetworkSettingWindow.header=Tor networks settings
--- a/desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
+++ b/desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
@@ -262,8 +262,12 @@ public class BsqSendView extends ActivatableView<GridPane, Void> implements BsqB
                             bsqFormatter,
                             btcFormatter,
                             () -> {
+                                receiversAddressInputTextField.setValidator(null);
                                 receiversAddressInputTextField.setText("");
+                                receiversAddressInputTextField.setValidator(bsqAddressValidator);
+                                amountInputTextField.setValidator(null);
                                 amountInputTextField.setText("");
+                                amountInputTextField.setValidator(bsqValidator);
                             });
                 } catch (BsqChangeBelowDustException e) {
                     String msg = Res.get("popup.warning.bsqChangeBelowDustException", bsqFormatter.formatCoinWithCode(e.getOutputValue()));
--- a/desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
+++ b/desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
@@ -27,6 +27,7 @@ import bisq.desktop.main.dao.wallet.BsqBalanceUtil;
 import bisq.desktop.main.funds.FundsView;
 import bisq.desktop.main.funds.deposit.DepositView;
 import bisq.desktop.main.overlays.popups.Popup;
+import bisq.desktop.main.overlays.windows.TxDetailsBsq;
 import bisq.desktop.main.overlays.windows.WalletPasswordWindow;
 import bisq.desktop.util.GUIUtil;
 import bisq.desktop.util.Layout;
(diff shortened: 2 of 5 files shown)


core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -2693,6 +2693,13 @@ tradeDetailsWindow.tradeState=Trade state
 tradeDetailsWindow.agentAddresses=Arbitrator/Mediator
 tradeDetailsWindow.detailData=Detail data
 
+txDetailsWindow.headline=Transaction Details
+txDetailsWindow.btc.note=You have sent BTC.
+txDetailsWindow.bsq.note=You have sent BSQ funds.  \
+  Please note that BSQ is colored Bitcoin, it will not show up in a BSQ explorer until at least 1 Bitcoin block has been confirmed.
```
m52go: I got the impression this string is saying the BSQ transaction will show in the BSQ explorer as soon as 1 more bitcoin block is confirmed, whether it includes the BSQ transaction or not. 

Suggestion to make more specific:

```suggestion
txDetailsWindow.bsq.note=You have sent BSQ funds.  \
  BSQ is colored bitcoin, so the transaction will not show in a BSQ explorer until it has been confirmed in a bitcoin block.
```

ghost: Thankyou @m52go, change applied.

wiz:
Nice!! Not related to this PR but it would be cool to have 2 links to the block explorer, one for clearnet and one for Tor

pazza83:
Thanks for doing this so quickly. 

Would also be good if mining fees where displayed.

ghost:
Mining fees are displayed on the "Confirm withdrawal request" popup immediately prior.  There's no need to display them after sending.
Screenshot from 2021-01-10 14-10-23

ghost:
Added "Don't show again" checkbox.

image

desktop/src/main/java/bisq/desktop/main/overlays/windows/TxDetails.java
```diff
@@ -0,0 +1,66 @@
...
+        addButtons();
+        addDontShowAgainCheckBox();
+        applyStyles();
+        display();
+    }
+
+    protected void addContent() {
+        GridPane.setColumnSpan(
+                addMultilineLabel(gridPane, ++rowIndex, note, 0), 2);
+        gridPane.add(new Label(""), 0, ++rowIndex);  // spacer
```
ripcurlx: I normally use for this cases `Region` which is more lightweight as `Label`

e.g. 
```suggestion
        Region spacer = new Region();
        spacer.setMinHeight(20);
        gridPane.add(spacer, 0, ++rowIndex);
```
Would have the same visual appearance.

desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
```diff
@@ -373,6 +375,12 @@ private void showPublishTxPopup(Coin receiverAmount,
                         @Override
                         public void onSuccess(Transaction transaction) {
                             log.debug("Successfully sent tx with id {}", txWithBtcFee.getTxId().toString());
+                            String key = "showTransactionSentBsq";
```
ripcurlx, line 382: It would be great to get rid of the error message on the empty fields after the successful transfer as well.
For the `BsqSendView` this would mean changing in the `showPublishTxPopup` resultHandler into
```
() -> {
       receiversAddressInputTextField.setValidator(null);
       receiversAddressInputTextField.setText("");
       receiversAddressInputTextField.setValidator(bsqAddressValidator);
       amountInputTextField.setValidator(null);
       amountInputTextField.setText("");
       amountInputTextField.setValidator(bsqValidator);
       });
```

ripcurlx (changes requested):
NACK - it does work as expected, but I would suggest two small changes to use a more lightweight container in one case and improve the UX after the success popup is closed. Thanks for picking up this issue!

desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
ripcurlx: Not beautiful, but it works :wink:

ghost: Thanks @ripcurlx, implemented your suggestions. :+1:

ripcurlx (approved):

PR #5072, opened 2021-01-11
Persist and republish mailbox messages

chimp1984:
For better resiliance in case all seed nodes would go offline, we persist the mailbox messages and republish them at startup.
We had already a persistence for own mailbox messages and reuse that but add all mailbox messages.
We check for the TTL to not publish expired messages as well we check if we have received a remove message for that mailbox message.

To not overload too much the P2pService class I refactored the mailbox related code out to a new class. The first commits are refactoring commits. The code for the new feature starts at b91a1f7ef7ce183aa2e0dfc3899a0a954d8a7eb3


Conza88:
Awesome. Yes please

p2p/src/main/java/bisq/network/p2p/mailbox/MailboxItem.java
```diff
@@ -24,31 +24,70 @@
...
+
+    public boolean isMine() {
+        return decryptedMessageWithPubKey != null;
+    }
+
+    public String getUid() {
+        if (decryptedMessageWithPubKey != null) {
+            // We use uid from mailboxMessage in case its ours as we have the at removeMailboxMsg only the
+            // decryptedMessageWithPubKey available which contains the mailboxMessage.
+            MailboxMessage mailboxMessage = (MailboxMessage) checkNotNull(decryptedMessageWithPubKey).getNetworkEnvelope();
```
sqrrm: Why the need for `checkNotNull` when that's checked in the if already

p2p/src/main/java/bisq/network/p2p/mailbox/MailboxMessageService.java
```diff
@@ -406,18 +404,25 @@ private MailboxItem decryptProtectedMailboxStorageEntry(ProtectedMailboxStorageE
             log.error(e.toString());
             e.getStackTrace();
         }
-        return null;
+        return new MailboxItem(protectedMailboxStorageEntry, null);
```
sqrrm, line 481: Unnecessary empty line

p2p/src/main/java/bisq/network/p2p/storage/RemovedPayloadsStorageService.java
```diff
@@ -0,0 +1,69 @@
...
+    ///////////////////////////////////////////////////////////////////////////////////////////
+    // PersistedDataHost
+    ///////////////////////////////////////////////////////////////////////////////////////////
+
+    @Override
+    public void readPersisted(Runnable completeHandler) {
+        long cutOffDate = System.currentTimeMillis() - MailboxStoragePayload.TTL;
+        persistenceManager.readPersisted(persisted -> {
+                    persisted.getDateByHashes().entrySet().stream()
+                            .filter(e -> e.getValue() < cutOffDate)
```
sqrrm: I don't think I understand the usage of this class. Here is looks like when the persisted removedPayloadsMap is read the hashes that are older than the cutoff date are stored, but that means some removed data will be lost.

sqrrm (commented):
Would it make sense to have seednodes request all mailbox messages on startup and then republish the discrepancy between the response and their locally persisted messages?

That would reduce the amount of republishing for the price of only seednodes asking for all messages from the other seednodes on startup.

p2p/src/main/java/bisq/network/p2p/mailbox/MailboxItem.java
chimp1984: will remove it

p2p/src/main/java/bisq/network/p2p/storage/RemovedPayloadsStorageService.java
chimp1984: Ups that is wrong. Should `>`. It cleans out old entries so it does not grow endlessly.
The class it to persist the remove call so we do not add again removed mailbox messages. I will add some comments as well.

chimp1984:
Force pushed with rebase on master and fixes from the code review.

It still needs more testing, so please do not merge yet.

p2p/src/main/java/bisq/network/p2p/storage/RemovedPayloadsStorageService.java
sqrrm: Ok, that makes sense, thought so at first, then second guessed myself and got confused.

chimp1984:
> Would it make sense to have seednodes request all mailbox messages on startup and then republish the discrepancy between the response and their locally persisted messages?
> 
> That would reduce the amount of republishing for the price of only seednodes asking for all messages from the other seednodes on startup.

As discussed in our call I think its better to keep it more simple and to use a prog argument to enable republishing for any node (not only seed node). By default we would enable it only on one v2 and one v3 node, that is enough to increasing reliability and in any emergency case we can activate it where needed.

Conza88:
Awesome

p2p/src/main/java/bisq/network/p2p/mailbox/MailboxMessageService.java
```diff
@@ -271,18 +271,19 @@ public void removeMailboxMsg(MailboxMessage mailboxMessage) {
...
-                // So to be sure it gets removed we try to remove it now again.
-                // In case it was removed earlier it will return early anyway inside the p2pDataStorage.
-                removeMailboxEntryFromNetwork(mailboxItemsByUid.get(uid).getProtectedMailboxStorageEntry());
-
-                // We will get called the onRemoved handler which triggers removeMailboxItemFromMap as well.
-                // But as we use the uid from the decrypted data which is not available at onRemoved we need to
-                // call removeMailboxItemFromMap here. The onRemoved only removes foreign mailBoxMessages.
-                log.trace("## removeMailboxMsg uid={}", uid);
-                removeMailboxItemFromLocalStore(uid);
+                if (mailboxItemsByUid.containsKey(uid)) {
```
sqrrm: Prefer early return

sqrrm (commented):
Seems mostly good. I wonder about the size check to push out old messages though.

chimp1984:
Pr is running on 75% of the seeds since a few days. I guess its safe now to merge into master...

sqrrm (approved):
utACK

PR #5076, opened 2021-01-11
Stub out support for OpenOffer's triggerPrice in api

ghubstan:
This is a feature that will not be included in api v1, but partial support is added with this change to the server.  CLI will pass a default `triggerPrice` of 0 (default, unused) with the `createoffer` command.

When fully implemented, an optional trigger-price param will be  added to the `createoffer` method, and its value will only be visible to the offer owner.  New `enableoffer` and `disableoffer` methods will also need to be added at that time.

This is the 3rd in a chain of PRs, starting with https://github.com/bisq-network/bisq/pull/5063.  
https://github.com/bisq-network/bisq/pull/5065 should be reviewed & merged before this one.


ghubstan:
force travis build (download jar)

sqrrm (commented):
Triggerprice changes look ok.

The reason I don't feel it's an important feature for the api is that the api user could listen to the price feed and disable offers programmatically rather than relying on the built in feature. At least once enable/disable offer is implemented.

sqrrm (approved):
utACK

PR #5077, opened 2021-01-11
Fix request amount bounds in ReimbursementValidator

stejbac:
Use `ReimbursementConsensus.get[Min|Max]ReimbursementRequestAmount` in place of `CompensationConsensus.get[Min|Max]CompensationRequestAmount`, which was erroneously copied (it appears) from the very similar `CompensationValidator` class.

This ensures the correct upper bound (currently 80,000 BSQ, starting at 10,000 BSQ and doubled in cycles 12, 13 & 14 as part of https://github.com/bisq-network/proposals/issues/203) is placed on reimbursement requests, instead of the erroneous 100,000 BSQ upper bound for compensation requests.

--

This touches the DAO packages and potentially affects consensus. I don't believe any reimbursement requests exceeding the intended limit for each given cycle were actually made, so this shouldn't invalidate past vote cycles or BSQ issuance.

Converting DaoStateStore to text, filtering and manually copying out the relevant values (so there could be mistakes), I found 30 reimbursement requests in total, as follows:

| Block height of request | BSQ Amount |
| --- | --- |
| 600609 | 434.60 |
| 601624 | 143.00 |
| 603238 | 30.00 |
| 623401 | 476.00 |
| 625775 | 892.00 |
| 628607 | 921.00 |
| 630997 | 1950.00 |
| 636176 | 428.00 |
| 639551 | 1411.99 |
| 640347 | 29200.00 |
| 640347 | 27933.67 |
| 640741 | 34141.70 |
| 643632 | 1379.97 |
| 645119 | 9850.60 |
| 645233 | 97.53 |
| 645272 | 10371.00 |
| 649445 | 50217.66 |
| 649680 | 16069.22 |
| 650086 | 21407.20 |
| 652676 | 21267.15 |
| 653608 | 206.00 |
| 654003 | 364.00 |
| 654850 | 1586.74 |
| 654859 | 27934.70 |
| 658129 | 1586.74 |
| 659360 | 19629.90 |
| 663715 | 61875.30 |
| 663903 | 852.80 |
| 664138 | 25580.70 |
| 664138 | 25580.70 |

The first eight requests are before the parameter changes in Cycle 14 activated (height 637267), so every reimbursement request after that should have had an 80,000 BSQ limit anyway.

### DIFF
--- a/core/src/main/java/bisq/core/dao/governance/proposal/reimbursement/ReimbursementValidator.java
+++ b/core/src/main/java/bisq/core/dao/governance/proposal/reimbursement/ReimbursementValidator.java
@@ -21,7 +21,6 @@ import bisq.core.dao.governance.ConsensusCritical;
 import bisq.core.dao.governance.period.PeriodService;
 import bisq.core.dao.governance.proposal.ProposalValidationException;
 import bisq.core.dao.governance.proposal.ProposalValidator;
-import bisq.core.dao.governance.proposal.compensation.CompensationConsensus;
 import bisq.core.dao.state.DaoStateService;
 import bisq.core.dao.state.model.governance.Proposal;
 import bisq.core.dao.state.model.governance.ReimbursementProposal;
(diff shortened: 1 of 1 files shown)


ripcurlx:
As by previous agreement this needs an ACK by @ManfredKarrer

ManfredKarrer (approved):
utACK

Yes seems that was a copy&paste bug. Thanks for spotting it!
To ensure it does not break anything, could you build from genesis and compare final Dao state hashes? A breakpoint in the catches should also spot any case when it would have been triggered (don't assume so).

stejbac:
I've just finished a rebuild from genesis (running as a DAO full node) and the final DAO state hashes still agree with those of the seed nodes. I put a breakpoint in each of the two catch blocks in `ReimbursementValidator.validateDataFields` and neither of them hit during the rebuild.

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/5077#pullrequestreview-568761902

PR #5080, opened 2021-01-14
Change rule for "reason for payment" field to not use trade ID but leave it empty

chimp1984:
Implements https://github.com/bisq-network/bisq/issues/2869

- Adds popup at startup with rule change for users who have already a fiat account.
- Change TAC text to leave field empty and not use Bisq, BTC or trade id
- Change buyer popup in trade process to tell to leave field empty
- Remove text fields with reason for payment in trade process
- Remove text from popup for seller to check for reason for payment to match trade id

Hope I did not missed some...

@m52go @pazza83 I am not 100% sure if the popup at startup is enough to announce the change. I think we should additionally communicate it clearly in all our channels and of course to our support guys.

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -2884,6 +2884,9 @@ popup.shutDownInProgress.headline=Shut down in progress
 popup.shutDownInProgress.msg=Shutting down application can take a few seconds.\nPlease don't interrupt this process.
 
 popup.attention.forTradeWithId=Attention required for trade with ID {0}
+popup.attention.reasonForPaymentRuleChange=Version 1.5.5. introduces a trade rule change regarding the usage of \
+  the \"reason for payment\" field in bank transfers. Please leave the field empty and DO NOT use the trade ID anymore \
+  as it was required in the past.
 
 popup.info.multiplePaymentAccounts.headline=Multiple payment accounts available
 popup.info.multiplePaymentAccounts.msg=You have multiple payment accounts available for this offer. Please make sure you've picked the right one.
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -2884,9 +2884,9 @@ popup.shutDownInProgress.headline=Shut down in progress
 popup.shutDownInProgress.msg=Shutting down application can take a few seconds.\nPlease don't interrupt this process.
 
 popup.attention.forTradeWithId=Attention required for trade with ID {0}
-popup.attention.reasonForPaymentRuleChange=Version 1.5.5. introduces a trade rule change regarding the usage of \
-  the \"reason for payment\" field in bank transfers. Please leave the field empty and DO NOT use the trade ID anymore \
-  as it was required in the past.
+popup.attention.reasonForPaymentRuleChange=Version 1.5.5 introduces a critical trade rule change regarding \
+  the \"reason for payment\" field in bank transfers. Please leave this field empty -- \
+  DO NOT use the trade ID as \"reason for payment\" anymore.
 
 popup.info.multiplePaymentAccounts.headline=Multiple payment accounts available
 popup.info.multiplePaymentAccounts.msg=You have multiple payment accounts available for this offer. Please make sure you've picked the right one.
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -636,7 +636,7 @@ portfolio.pending.step1.openForDispute=The deposit transaction is still not conf
 portfolio.pending.step2.confReached=Your trade has reached at least one blockchain confirmation.\n(You can wait for more confirmations if you want - 6 confirmations are considered as very secure.)\n\n
 
 portfolio.pending.step2_buyer.copyPaste=(You can copy & paste the values from the main screen after closing that popup.)
-portfolio.pending.step2_buyer.refTextWarn=DO NOT use any additional notice in the \"reason for payment\" text like bitcoin, BTC, Bisq or the trade ID. Leave the field empty.
+portfolio.pending.step2_buyer.refTextWarn=Leave the \"reason for payment\" field empty. DO NOT put the trade ID or any other text like 'bitcoin', 'BTC', or 'Bisq'.
 # suppress inspection "TrailingSpacesInProperty"
 portfolio.pending.step2_buyer.accountDetails=Here are the trade account details of the BTC seller:\n
 portfolio.pending.step2_buyer.fees=If your bank charges fees you have to cover those fees.
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -114,7 +114,6 @@ shared.waitingForFunds=Waiting for funds...
 shared.depositTransactionId=Deposit transaction ID
 shared.TheBTCBuyer=The BTC buyer
 shared.You=You
-shared.reasonForPayment=Reason for payment
 shared.sendingConfirmation=Sending confirmation...
 shared.sendingConfirmationAgain=Please send confirmation again
 shared.exportCSV=Export to CSV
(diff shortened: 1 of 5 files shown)


chimp1984:
Screen Shot 2021-01-13 at 23 31 48
Screen Shot 2021-01-13 at 23 33 09

Screen Shot 2021-01-13 at 23 44 31

chimp1984:
@m52go Could you have a look to the text changes?

Conza88:
REALLY think if this is going ahead to advise or push folks to continue discussing in chat with trading partner. That's the important part. Communication. 

There's no indication anywhere why this change is being made either. It will likely make life harder for extensive traders, no? That's not ideal. They'd want to know WHY this sudden change. I think stating clearly the rationale would help immensely. 

Obviously the high volume traders are more at risk getting banned as well, so should support a move that lessens that.

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -637,12 +636,9 @@ portfolio.pending.step1.openForDispute=The deposit transaction is still not conf
 portfolio.pending.step2.confReached=Your trade has reached at least one blockchain confirmation.\n(You can wait for more confirmations if you want - 6 confirmations are considered as very secure.)\n\n
 
 portfolio.pending.step2_buyer.copyPaste=(You can copy & paste the values from the main screen after closing that popup.)
-portfolio.pending.step2_buyer.refTextWarn=DO NOT use any additional notice in the \"reason for payment\" text like bitcoin, BTC or Bisq.
+portfolio.pending.step2_buyer.refTextWarn=DO NOT use any additional notice in the \"reason for payment\" text like bitcoin, BTC, Bisq or the trade ID. Leave the field empty.
```
```suggestion
portfolio.pending.step2_buyer.refTextWarn=Leave the \"reason for payment\" field empty. DO NOT put the trade ID or any other text like 'bitcoin', 'BTC', or 'Bisq'.
```

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -2892,6 +2884,9 @@ popup.shutDownInProgress.headline=Shut down in progress
 popup.shutDownInProgress.msg=Shutting down application can take a few seconds.\nPlease don't interrupt this process.
 
 popup.attention.forTradeWithId=Attention required for trade with ID {0}
+popup.attention.reasonForPaymentRuleChange=Version 1.5.5. introduces a trade rule change regarding the usage of \
+  the \"reason for payment\" field in bank transfers. Please leave the field empty and DO NOT use the trade ID anymore \
+  as it was required in the past.
```
```suggestion
popup.attention.reasonForPaymentRuleChange=Version 1.5.5 introduces a critical trade rule change regarding \
  the \"reason for payment\" field in bank transfers. Please leave this field empty -- \
  DO NOT use the trade ID as \"reason for payment\" anymore.
```

desktop/src/main/java/bisq/desktop/main/overlays/windows/TacWindow.java
```diff
@@ -83,16 +83,18 @@ public void show() {
 
                 "6. The user confirms that they have read and agreed to the rules regarding the dispute process:\n" +
                 "    - You must complete trades within the maximum duration specified for each payment method.\n" +
-                "    - You must enter the trade ID in the \"reason for payment\" text field (if possible) when doing the fiat payment transfer.\n" +
+                "    - DO NOT use any additional notice in the \"reason for payment\" text like bitcoin, BTC, Bisq or the trade ID. Leave the field empty.\n" +
```
m52go:
```suggestion
                "    - Leave the \"reason for payment\" field empty. DO NOT put the trade ID or any other text like 'bitcoin', 'BTC', or 'Bisq'.\n" +
```

m52go:
> Screen Shot 2021-01-13 at 23 33 09

^ there is a lot of text on this screen...this reduces the chance that the user reads any of it. 

I'm not sure the parts on multiple confirmations, bank details, copy/paste, etc. are that important.

Suggestion to simplify:

---

Your trade has reached at least 1 blockchain confirmation.

Please pay 303 EUR via the specified payment method to the BTC seller. You'll find the seller's account details on the next screen.

Important: when making the payment, leave the \"reason for payment\" field empty. DO NOT put the trade ID or any other text like 'bitcoin', 'BTC', or 'Bisq'.

If your bank charges you any fees to make the transfer, you are responsible for paying those fees.

---

Conza88:
Nice. Is there a potential wiki link if someone is new, or old school and wants to know more? Something that may cover _why_ we're no longer suggesting putting in trade ID? It's a massive step from MUST have trade ID, to **DO NOT**. 

> Your trade has reached at least 1 blockchain confirmation.
> 
> Please pay 303 EUR via the specified payment method to the BTC seller. You'll find the seller's account details on the next screen.
> 
> Important: when making the payment, leave the "reason for payment" field empty. **DO NOT** put the trade ID or any other text like 'bitcoin', 'BTC', or 'Bisq'. **_(i) This is to reduce the risk of fiat accounts being deactivated or banned_.** You are free to discuss via trader chat if an alternate 'trade ID' would be suitable to you both.
> 
> If your bank charges you any fees to make the transfer, you are responsible for paying those fees.

Added: "**_(i) This is to reduce the risk of fiat accounts being deactivated or banned_.** 

You are free to discuss via trader chat if an alternate 'trade ID' would be suitable to you both."

Ideally the (i) only appears when hovered over, not sure if can be done in popup.

chimp1984:
> ^ there is a lot of text on this screen...this reduces the chance that the user reads any of it.

> I'm not sure the parts on multiple confirmations, bank details, copy/paste, etc. are that important.

I agree, but I would prefer to keep that for another PR. There are more popup text to improve but as the text is composed of several blocks depending on payment method its also not that trivial and need then better testing with the diff. payment methods.

Any dev who wants to pick that up?

@m52go Can you check if there are reference on the webpage or other resources regarding that fields?

chimp1984:
> Nice. Is there a potential wiki link if someone is new, or old school and wants to know more? Something that may cover why > we're no longer suggesting putting in trade ID? It's a massive step from MUST have trade ID, to DO NOT.

Yes I agree, but I think in the app we should keep things minimal and that context is only relevant for existing users not for new users. Maybe @m52go has a suggestion how to communicate best the reason for the change. To add a sentence to the popup at startup might be ok as well, as its only shown to existing users.

m52go:
> Can you check if there are reference on the webpage or other resources regarding that fields?

Yes I'll check. There are probably some spots on the wiki, not sure about website.

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -637,12 +636,9 @@ portfolio.pending.step1.openForDispute=The deposit transaction is still not conf
 portfolio.pending.step2.confReached=Your trade has reached at least one blockchain confirmation.\n(You can wait for more confirmations if you want - 6 confirmations are considered as very secure.)\n\n
 
 portfolio.pending.step2_buyer.copyPaste=(You can copy & paste the values from the main screen after closing that popup.)
-portfolio.pending.step2_buyer.refTextWarn=DO NOT use any additional notice in the \"reason for payment\" text like bitcoin, BTC or Bisq.
+portfolio.pending.step2_buyer.refTextWarn=Leave the \"reason for payment\" field empty. DO NOT put the trade ID or any other text like 'bitcoin', 'BTC', or 'Bisq'.
```
cbeams, line 639: As I mentioned in Keybase, I'm concerned that it might be an issue that some banks / payment providers will require a non-empty 'reason for payment' field. We can wait and see if this is actually a problem, but just a heads-up that it might be.

cbeams:
utACK. What about i18n versions of these messages? Is the process tight enough there to ensure we get translations (for the most important languages) coordinated with this release?

core/src/main/resources/i18n/displayStrings.properties
pazza83: I can confirm that some banks require something to be entered into the reference field. Don't thinks it will be too much of a problem until a new user enters 'Bitcoin'

Maybe wiki etc can highlight a list of banned words and phrases which would result in entire loss of security deposit (BTC, Bisq, Bitcoin, crypto, the tx id etc), I would suggest user use their first name if required to enter something (if same as bank details), or something innocuous like 'funds'.

chimp1984: I guess most users will navigate around banks weird UX issues by entering a dash or whatever . Also the reason for payment field is not shown at all banks, or some characters are not permitted or they have limited it to weird short lenghts... banks UX sucks anyway so much that ppl are used to it I assume.

chimp1984 (commented):
> utACK. What about i18n versions of these messages? Is the process tight enough there to ensure we get translations (for the most important languages) coordinated with this release?

@ripcurlx @m52go Could you push translaters, or at least delete the old entries so english will be the fallback?

m52go:
When is the next release expected? cc @mwithm

chimp1984:
> When is the next release expected? cc @MwithM

I am not sure, but I guess code freeze will be soon. @ripcurlx Any concrete date?

Conza88:
"I'm concerned that it might be an issue that some banks / payment providers will require a non-empty 'reason for payment' field."

I think this is also very likely. There's value in recen advising traders discuss what could be used imo.

________________________________

ghost (approved):
ACK tested locally with Amazon, USPSMO and Transferwise payment methods.

ripcurlx:
> > When is the next release expected? cc @MwithM
> 
> I am not sure, but I guess code freeze will be soon. @ripcurlx Any concrete date?

Codefreeze will be on Wednesday, but I'll update the translation files the week after. So there should be enough time for translators to come up with proper translations.

ripcurlx (approved):
utACK

pazza83:
Sorry, I missed this. 

I had a look at the text and it all looks great to me. I am sure there will be some teething issues with traders using trade ID as payment references but these will quickly be sorted out and the odd trader using them until informed otherwise does not seem like much, if any, risk.

I agree the Wiki / website needs to ensure that it is edited to include the new trade protocol. I will contact @m52go on Keybase about what is needed and agree a plan.

PR #5081, opened 2021-01-14
Use posix-style CLI opts, provide method help

ghubstan:
Posix-style method opts replace the formerly ambiguous, positional method params in the CLI.

This change also adds new api `CoreHelpService`, for serving method  specific help from the server.  (Help text is defined on the server to avoid duplicating it when the api starts serving RESTful clients.)

- Add `CoreHelpService` to serve method help in man page format.

- Add some method help docs (man pages) to core's resources: `createoffer`, `takeoffer`, and several no-arg methods.
  
- Add `gRPC` `CoreHelpService` stubs to CLI.
  
- Move CLI method enum to its own class to help reduce file-size of growing `CliMain` class.

- Add CLI method specific posix-style option parsers.  Opts are parsed by the `joptsimple` lib, adding much needed client side validation to `bisq-cli` commands.

- Adjust CLI to new posix-style opts, and method help service.

- Delete uneeded `NegativeNumberOptions` class because `joptsimple` correctly handles negative posix style opt values (joptsimple was treating negative param values as opt labels).

- Adjust `bats` mainnet smoke test to new posix-style CLI opts.


ghubstan:
Both CI builds failed due to jar download problem.

core/src/main/java/bisq/core/api/CoreApi.java
```diff
@@ -72,14 +73,15 @@
     @Inject
     public CoreApi(Config config,
                    CoreDisputeAgentsService coreDisputeAgentsService,
-                   CoreOffersService coreOffersService,
+                   CoreHelpService coreHelpService, CoreOffersService coreOffersService,
```
sqrrm, line 76: Line break needed

sqrrm (approved):
utACK

core/src/main/java/bisq/core/api/CoreApi.java
ghubstan: This fix (5538914cf87f194874d5263a7af732d8a8e041ce) will go into the next PR.

PR #5083, opened 2021-01-15
Avoid reverse DNS lookup in BtcNodeConverterTest

stejbac:
Prevent failure of `testConvertClearNode()` on some machines, caused by use of `InetAddress.getHostName` on the mock peer address. This does a reverse DNS lookup and potentially returns something other than the expected "192.168.0.1" string. (For example, on my current local network I get "BrightBox.ee".)

Also avoid an unnecessary `getHostName()` call on the SOCKS5 Tor proxy InetAddress in `WalletConfig`, by using an alternative `InetSocketAddress` constructor.


stejbac:
Rebasing onto a more recent version of master to force a rebuild...

ripcurlx (approved):
utACK

PR #5084, opened 2021-01-16
Add core api method help docs

ghubstan:


ghubstan:
Force Travis build (jar download).

PR #5087, opened 2021-01-18
Keep main navigation items left aligned

### LINKED ISSUE #5086: Navigation Responsive Layout
### Description
I've been noticing a lot of users using the client in fullscreen or constantly resizing the width of the window. The layout of the navigation looks broken whenever the width goes beyond the default on load. It would look a lot more polished if we keep the navigation contained in its original width.

#### Version
1.5.1

### Expected behaviour
Navigation should have a maximum width so we maintain equal distances between elements.

### Actual behaviour
Navigation elements align to the edges of the container.

### Screenshots

00 Overview
01 Current Default Width
02 Current Increased Width
03 Proposed with max width

ripcurlx:
Fixes #5086.


ghost (approved):
ACK :+1:

sqrrm (approved):

PR #5094, opened 2021-01-19
Fix bug with ignoreLocalBtcNode

chimp1984:
Do not pass proxySocketFactory to BlockingClientManager if we use a local Bitcoin node.


sqrrm (approved):
utACK

PR #5099, opened 2021-01-21
Manual payout tool: prevent absurdly high fee payout

ghost:
This PR adds display of the Tx fee as a percentage, and includes validation to disallow the payout processing if tx fee is > 10% of the multisig amount.  The Tx cannot be signed or built until validation passes.

Fixes a bug whereby whitespace in an amount caused it to be parsed as 0.  It now trims whitespace from numeric input fields before parsing.

---

**Screenshot showing the new Tx fee % field:**

image

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/overlays/windows/ManualPayoutTxWindow.java
+++ b/desktop/src/main/java/bisq/desktop/main/overlays/windows/ManualPayoutTxWindow.java
@@ -24,6 +24,7 @@ import bisq.desktop.main.overlays.Overlay;
 import bisq.desktop.main.overlays.popups.Popup;
 import bisq.desktop.util.GUIUtil;
 import bisq.desktop.util.validation.LengthValidator;
+import bisq.desktop.util.validation.PercentageNumberValidator;
 
 import bisq.core.btc.exceptions.TransactionVerificationException;
 import bisq.core.btc.exceptions.TxBroadcastException;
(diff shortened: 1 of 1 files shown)


desktop/src/main/java/bisq/desktop/main/overlays/windows/ManualPayoutTxWindow.java
```diff
@@ -544,6 +553,8 @@ private void calculateTxFee() {
                     .subtract(getInputFieldAsCoin(buyerPayoutAmount))
                     .subtract(getInputFieldAsCoin(sellerPayoutAmount));
             txFee.setText(txFeeValue.toPlainString());
+            double feePercent = (double) txFeeValue.value / getInputFieldAsCoin(amountInMultisig).value;
```
chimp1984, line 556: Maybe safer to wrap it into a try catch? E.g. if getInputFieldAsCoin(amountInMultisig).value can be 0 it would throw an exception.

chimp1984 (changes requested):

desktop/src/main/java/bisq/desktop/main/overlays/windows/ManualPayoutTxWindow.java
ghost: No exception, it returns `-Infinity`.  So I don't see any issue.

image

chimp1984 (approved):
utACK

ripcurlx (approved):
utACK

ripcurlx:
I just cherry-picked it into the release branch

PR #5103, opened 2021-01-22
Define gRPC api call rate constraints

ghubstan:
The general rule is limit calls that change p2p data to 1/minute, others to 1/second.  An exception is made to set/remove wallet password methods (1/5s), due to the latency of writing wallet changes to disk.

This change may affect api testing in the future.  If that happens, further changes to the call rate metering interceptor may be made to loosen the constraints when running in regtest/dev mode.

### DIFF
--- a/daemon/src/main/java/bisq/daemon/grpc/GrpcDisputeAgentsService.java
+++ b/daemon/src/main/java/bisq/daemon/grpc/GrpcDisputeAgentsService.java
@@ -56,14 +56,13 @@ class GrpcDisputeAgentsService extends DisputeAgentsGrpc.DisputeAgentsImplBase {
     }
 
     final Optional<ServerInterceptor> rateMeteringInterceptor() {
-        CallRateMeteringInterceptor defaultCallRateMeteringInterceptor =
-                new CallRateMeteringInterceptor(new HashMap<>() {{
-                    // You can only register mainnet dispute agents in the UI.
-                    // Do not limit devs' ability to register test agents.
-                    put("registerDisputeAgent", new GrpcCallRateMeter(1, SECONDS));
-                }});
-
         return getCustomRateMeteringInterceptor(coreApi.getConfig().appDataDir, this.getClass())
-                .or(() -> Optional.of(defaultCallRateMeteringInterceptor));
+                .or(() -> Optional.of(CallRateMeteringInterceptor.valueOf(
+                        new HashMap<>() {{
+                            // You can only register mainnet dispute agents in the UI.
+                            // Do not limit devs' ability to register test agents.
+                            put("registerDisputeAgent", new GrpcCallRateMeter(1, SECONDS));
+                        }}
+                )));
     }
 }
--- a/daemon/src/main/java/bisq/daemon/grpc/GrpcDisputeAgentsService.java
+++ b/daemon/src/main/java/bisq/daemon/grpc/GrpcDisputeAgentsService.java
@@ -6,11 +6,23 @@ import bisq.proto.grpc.DisputeAgentsGrpc;
 import bisq.proto.grpc.RegisterDisputeAgentReply;
 import bisq.proto.grpc.RegisterDisputeAgentRequest;
 
+import io.grpc.ServerInterceptor;
 import io.grpc.stub.StreamObserver;
 
 import javax.inject.Inject;
 
+import java.util.HashMap;
+import java.util.Optional;
+
 import lombok.extern.slf4j.Slf4j;
+
+import static bisq.daemon.grpc.interceptor.GrpcServiceRateMeteringConfig.getCustomRateMeteringInterceptor;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+
+
+import bisq.daemon.grpc.interceptor.CallRateMeteringInterceptor;
+import bisq.daemon.grpc.interceptor.GrpcCallRateMeter;
 
 @Slf4j
 class GrpcDisputeAgentsService extends DisputeAgentsGrpc.DisputeAgentsImplBase {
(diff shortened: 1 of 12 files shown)


daemon/src/main/java/bisq/daemon/grpc/GrpcWalletsService.java
```diff
@@ -330,4 +341,37 @@ public void unlockWallet(UnlockWalletRequest req,
...
+                    put("sendBtc", new GrpcCallRateMeter(1, MINUTES));
+                    put("getTxFeeRate", new GrpcCallRateMeter(1, SECONDS));
+                    put("setTxFeeRatePreference", new GrpcCallRateMeter(1, SECONDS));
+                    put("unsetTxFeeRatePreference", new GrpcCallRateMeter(1, SECONDS));
+                    put("getTransaction", new GrpcCallRateMeter(1, SECONDS));
+
+                    // Trying to set or remove a wallet password several times before the 1st attempt has time to
+                    // persist the change to disk may corrupt the wallet, so allow only 1 attempt per 5 seconds.
+                    put("setWalletPassword", new GrpcCallRateMeter(1, SECONDS, 5));
+                    put("removeWalletPassword", new GrpcCallRateMeter(1, SECONDS, 5));
```
dmos62: This seems like a race condition that can have severe consequences (wallet corruption). Maybe further reduce the rate to be on the safe side?

daemon/src/main/java/bisq/daemon/grpc/GrpcWalletsService.java
```diff
@@ -330,4 +341,37 @@ public void unlockWallet(UnlockWalletRequest req,
...
+
+    final Optional<ServerInterceptor> rateMeteringInterceptor() {
+        CallRateMeteringInterceptor defaultCallRateMeteringInterceptor =
+                new CallRateMeteringInterceptor(new HashMap<>() {{
+                    put("getBalances", new GrpcCallRateMeter(1, SECONDS));
+                    put("getAddressBalance", new GrpcCallRateMeter(1, SECONDS));
+                    put("getFundingAddresses", new GrpcCallRateMeter(1, SECONDS));
+                    put("getUnusedBsqAddress", new GrpcCallRateMeter(1, SECONDS));
+                    put("sendBsq", new GrpcCallRateMeter(1, MINUTES));
+                    put("sendBtc", new GrpcCallRateMeter(1, MINUTES));
```
dmos62: This might be a silly question. What would be the effect of the extreme case of not rate limiting these mutating gRPC calls (the ones with once per minute rate)? Is this to protect the daemon or the Bisq network? The reason I'm asking is to make sure that this isn't an important DDOS protection mechanism.

daemon/src/main/java/bisq/daemon/grpc/GrpcDisputeAgentsService.java
```diff
@@ -36,4 +48,22 @@ public void registerDisputeAgent(RegisterDisputeAgentRequest req,
...
+        CallRateMeteringInterceptor defaultCallRateMeteringInterceptor =
+                new CallRateMeteringInterceptor(new HashMap<>() {{
+                    // You can only register mainnet dispute agents in the UI.
+                    // Do not limit devs' ability to register test agents.
+                    put("registerDisputeAgent", new GrpcCallRateMeter(1, SECONDS));
+                }});
+
+        return getCustomRateMeteringInterceptor(coreApi.getConfig().appDataDir, this.getClass())
+                .or(() -> Optional.of(defaultCallRateMeteringInterceptor));
+    }
```
dmos62, line 67: These `rateMeteringInterceptor` methods seem repetitive. Are they degradeable to constants? The unique items seem to be the arguments to `put()` and `this`; might be possible to factor out the rest of the logic into a common method. I'm not really opposed to leaving this unchanged, just something to consider.

daemon/src/main/java/bisq/daemon/grpc/GrpcWalletsService.java
ghubstan: This is not to protect against malicious attacks.  A lot more analysis is needed for that.

The rate metering is to help prevent scripting mistakes.  We don't want anyone accidentally spamming the network with a create offer loop, or anyone accidentally emptying her own wallet.

daemon/src/main/java/bisq/daemon/grpc/GrpcWalletsService.java
ghubstan: Maybe.  I was advised to set it to 1/sec, but bumped it to 5/sec for the reason you give here.  Let's see what others have to say after they take a look at this.

daemon/src/main/java/bisq/daemon/grpc/GrpcDisputeAgentsService.java
ghubstan: Not sure how to fully answer this, but we cannot create a super class for these Grpc*Service classes, if that's what you are suggesting.  A Grpc service class must extend a class generated from a proto def.

Something else you might not realize... but ought to know:  a unique interceptor is bound to each Grpc*Service, and each interceptor has it's own unique set of rate meters.

ghubstan: But I could create an interceptor factory method to factor out some of the repetition.

ghubstan: Commit c996240 only factors out repeated calls to 'new CallRateMeteringInterceptor'.

I want to leave the `rateMeteringInterceptor` methods there because I expect there will be a need for other types of interceptors in the future, and I want it to be clear that the calling `final ServerInterceptor[] interceptors()` method should get them from a `<type>Interceptor` method.

sqrrm (approved):
utACK

PR #5108, opened 2021-01-23
New Feature: Pre-release notifications

### LINKED ISSUE #4532: Add option to specify which type of release to upgrade to (i.e. release/pre-release)
We recently started providing pre-release versions that are available a few days prior to an official release in the hopes of preventing unnecessary hotfixes after a release by allowing those that wish to try it out and provide feedback.

However, these pre-release versions are not notified/downloaded within the application and must be manually downloaded from Github. This means the user must rely on being notified of the pre-release through Keybase or by being subscribed to the repository. In addition, they must (hopefully) manually verify the downloaded files.

If we were to add the option to allow the user to specify which type of releases they want to upgrade to, that would allow them to be notified within the application and make the upgrade process much easier as it would handle the download and verification steps automatically. The biggest benefit may be that it potentially make more users try out the pre-release versions and provide early feedback.

ghost:
Fixes #4532

Adds the ability for pre-release notifications to be sent by the admin key holder.
Users have the option in Settings to subscribe for pre-release notifications.  (Default=off).

---

### Screenshots

Generating a pre-release notification:

image

User setting for subscribing to pre-release notifications:

image

Notification of a pre-release:

image

Notification link is shown in the footer bar (if subscribed) :

image

---

### Testing matrix for alert messages:

The alert can be either a regular release notification, a pre-release notification, or simply an general-purpose message.  This is a list of tests performed:

* Verify that a general text alert message can be sent by old and new clients.
* Verify that a general text alert message can be received by old and new clients.
* Verify that a general text alert message can be removed by old and new clients.
* Verify that a software release message can be sent by old and new clients.
* Verify that a software release message can be received by old and new clients.
* Verify that a software release message can be removed by old and new clients.
* Verify that a software PRE-release message can be sent by new clients.
* Verify that a software PRE-release message can be received by new clients.
* Verify that a software PRE-release message can be removed by new clients.
* Verify that a software PRE-release message does not affect old clients (it is not seen).
* Verify that a software PRE-release message removal does not affect old clients.
* Verify that a software PRE-release message followed by a software release message for the same version is handled correctly.
* Verify that the PRE-release message only pops up on clients that have opted-in via the User Preferences setting.
* Verify that the footer PRE-release notification link only shows up on clients that have opted-in via the User Preferences setting.
* Verify that if user responds with "Ignore this version" option the software release message is not shown again upon startup.
* Verify that if user responds with "Ignore this version" option the software pre-release message is not shown again upon startup.


chimp1984 (approved):
utACK

I assume the alert will fail if the version of the alert sender and the receiver is not the same as a new field got added and the signature check will fail as the hash of the payload is different.
If that is the case (I have not tested it) I think it is also not a big problem, we just need to be aware and Alert sender should use only the old version until most ppl have updated. 
Alternatively we could pack the new data into the extraDataMap, but I think its not worth it for that case. Old users who have not updated for longer time will miss Alerts once the alert sender uses the new version.

ghost:
@chimp1984 That was one of the tests, and it does work.  I send a full-release alert `New version available, 1.9.9` from 1.5.5 to older versions (e.g. 1.5.4, etc); they handle it correctly, displaying the download screen.

Seems it is the textual alert message that is signed, not the proto object.  So extending the proto object is fine.


https://github.com/bisq-network/bisq/blob/3781b5f8b336fb26229c9d1b7f79cdc0a47f9331/core/src/main/java/bisq/core/alert/AlertManager.java#L161-L165

sqrrm (approved):

PR #5109, opened 2021-01-23
Wallet changes for Segwit BSQ implementation

stejbac:
This is a continuation from #5000, which replaces Bisq's Json RPC client to allow Segwit data (`"txinwitness"` fields) to be retrieved by DAO full nodes, in order to fill in the pubkeys of BSQ tx Segwit inputs (as necessary to support fully Segwit compensation and proof-of-burn txs).

This PR makes the necessary changes to the user's wallet and address formatting + validation to support Segwit BSQ, migrating the wallet upon startup in an analogous way to the Segwit BTC wallet upgrade (#4568). With these changes, newly generated BSQ addresses are native Segwit (p2wpkh) and take the form _'B' + \<bech32-address\>_, e.g.

> Bbc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq **(mainnet)**
> Bbcrt1qj4t04q4mpssfv9uzpm6w2n8rlhs7es47ktqx8x **(regtest)**

They should be valid everywhere the old base58 (p2pkh) BSQ addresses are, which continue to be supported (just never selected as new change or recipient addresses).

--

Note that this PR probably shouldn't be merged into master until the technical hard fork introduced by #5000 activates on mainnet, as otherwise upgraded users will have difficultly generating valid compensation requests or proof-of-burn txs. This is because native Segwit BSQ coins from their wallet may be arbitrarily selected as tx inputs.

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqSetup.java
+++ b/core/src/main/java/bisq/core/app/BisqSetup.java
@@ -424,7 +424,9 @@ public class BisqSetup {
                 requestWalletPasswordHandler.accept(aesKey -> {
                     walletsManager.setAesKey(aesKey);
                     walletsSetup.getWalletConfig().maybeAddSegwitKeychain(walletsSetup.getWalletConfig().btcWallet(),
-                            aesKey);
+                            aesKey, false);
+                    walletsSetup.getWalletConfig().maybeAddSegwitKeychain(walletsSetup.getWalletConfig().bsqWallet(),
+                            aesKey, true);
                     if (getResyncSpvSemaphore()) {
                         if (showFirstPopupIfResyncSPVRequestedHandler != null)
                             showFirstPopupIfResyncSPVRequestedHandler.run();
--- a/core/src/main/java/bisq/core/app/BisqSetup.java
+++ b/core/src/main/java/bisq/core/app/BisqSetup.java
@@ -423,10 +423,7 @@ public class BisqSetup {
             if (requestWalletPasswordHandler != null) {
                 requestWalletPasswordHandler.accept(aesKey -> {
                     walletsManager.setAesKey(aesKey);
-                    walletsSetup.getWalletConfig().maybeAddSegwitKeychain(walletsSetup.getWalletConfig().btcWallet(),
-                            aesKey, false);
-                    walletsSetup.getWalletConfig().maybeAddSegwitKeychain(walletsSetup.getWalletConfig().bsqWallet(),
-                            aesKey, true);
+                    walletsManager.maybeAddSegwitKeychains(aesKey);
                     if (getResyncSpvSemaphore()) {
                         if (showFirstPopupIfResyncSPVRequestedHandler != null)
                             showFirstPopupIfResyncSPVRequestedHandler.run();
--- a/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
@@ -527,7 +527,7 @@ public class BsqWalletService extends WalletService implements DaoStateListener 
     public Transaction getPreparedSendBsqTx(String receiverAddress, Coin receiverAmount)
             throws AddressFormatException, InsufficientBsqException, WalletException,
             TransactionVerificationException, BsqChangeBelowDustException {
-        return getPreparedSendTx(receiverAddress, receiverAmount, bsqCoinSelector, false);
+        return getPreparedSendTx(receiverAddress, receiverAmount, bsqCoinSelector);
     }
 
     public Transaction getPreparedSendBsqTx(String receiverAddress,
@@ -538,7 +538,7 @@ public class BsqWalletService extends WalletService implements DaoStateListener 
         if (utxoCandidates != null) {
             bsqCoinSelector.setUtxoCandidates(utxoCandidates);
         }
-        return getPreparedSendTx(receiverAddress, receiverAmount, bsqCoinSelector, false);
+        return getPreparedSendTx(receiverAddress, receiverAmount, bsqCoinSelector);
     }
 
     ///////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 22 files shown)


core/src/main/java/bisq/core/app/BisqSetup.java
```diff
@@ -416,7 +416,9 @@ private void initWallet() {
                 requestWalletPasswordHandler.accept(aesKey -> {
                     walletsManager.setAesKey(aesKey);
                     walletsSetup.getWalletConfig().maybeAddSegwitKeychain(walletsSetup.getWalletConfig().btcWallet(),
-                            aesKey);
+                            aesKey, false);
+                    walletsSetup.getWalletConfig().maybeAddSegwitKeychain(walletsSetup.getWalletConfig().bsqWallet(),
```
chimp1984: Maybe we can handle that in walletsManager, which is used for operations which should be applied to both wallets, so it is closer to the wallet domain.

core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
```diff
@@ -770,12 +770,13 @@ public Transaction getPreparedUnlockTx(TxOutput lockupTxOutput) throws AddressFo
...
-    private LegacyAddress getChangeAddress() {
+    private Address getChangeAddress() {
         return getUnusedAddress();
     }
 
-    public LegacyAddress getUnusedAddress() {
-        return (LegacyAddress) wallet.getIssuedReceiveAddresses().stream()
+    public Address getUnusedAddress() {
+        return wallet.getIssuedReceiveAddresses().stream()
+                .filter(address -> Script.ScriptType.P2WPKH.equals(address.getOutputScriptType()))
```
chimp1984, line 816: So we enforce a new segwit address in case a unused legacy address would be available. Just want to double check that this was your intention. I agree that this makes sense so no old dangling legacy addresses will be used.

chimp1984 (dismissed):
utACK

Looks all good to me. 

The Base58AddressValidator is still used in MockTestnetCoin. I think that can be replaced with BitcoinAddressValidator now.

core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
stejbac: Yes, that was my intention. The default active keychain (that is, the segwit one) comes last in the list of keychains queried by bitcoinj for issued addresses, so any unused legacy addresses would show up first in the filtered list. That was also what I observed on regtest, if I recall correctly.

core/src/main/java/bisq/core/app/BisqSetup.java
stejbac: Yes, it would probably be neater to put it there. I'll push a commit for that shortly.

chimp1984 (commented):
utACK

chimp1984:
Maybe it would be more safe to convert the 2 segwit PRs to draft until we are ready for merging it? (on the right side GH added a feature recently to convert a PR to draft)

chimp1984:
@stejbac Is that PR ready for the upcomig release?

chimp1984:
I just checked the ACTIVATE_HARD_FORK_2_HEIGHT_MAINNET is at block 680300 which was 3 days ago (2021-04-23 11:33).
https://mempool.space/block/00000000000000000009267fac0111c0e2feef71d001229e11e88c0c77e4d9a3

Proposal phase ends May 12th, so if we release soon we do not risk to fall into the voting period which carries a bit more risk.
DAO partizipants (proposal, voting) should update before posting proposal and doing voting, so reduce risks as well.

stejbac:
I just tried merging with master and there are some slightly nontrivial conflicts in `CoreWalletsService` and a few other places. As the draft changes haven't been touched for a while, I'll try rebasing onto master as it stood a few weeks ago, which should address all the conflicts I saw. (That way it will be easier to cherry pick into the 1.6.3 release if there's time.)

ripcurlx:
> I just tried merging with master and there are some slightly nontrivial conflicts in `CoreWalletsService` and a few other places. As the draft changes haven't been touched for a while, I'll try rebasing onto master as it stood a few weeks ago, which should address all the conflicts I saw. (That way it will be easier to cherry pick into the 1.6.3 release if there's time.)

Hi Steven! Just let me know when it is ready to be merged and tested. Please point the PR directly to the v1.6.3 release branch so we can test it there a bit before it is merged into master. Thanks!

stejbac:
I've rebased onto de8213a (2021/04/20 - shared by master and release/1.6.3).

stejbac:
(Amended the last commit of the rebase to remove a missed unused constructor parameter from `CoreWalletsService`.)

ghubstan:
Some important changes in `CoreWalletsService` will be lost if this PR is successfully merged.  

I cannot attend to the details until later today, but I the current master's `CoreWalletsService` should be included in this PR if possible (with the additional changes you require).

stejbac:
It appears the only difference between `CoreWalletService` on current master and the de8213a version I rebased off is a few lines calling `feeService.getMinFeePerVByte()`, in place of the hard coded fee `BTC_DAO_REGTEST.getDefaultMinFeePerVbyte()`. But I don't think those lines conflict (textually or otherwise) with the changes from this PR, so they shouldn't be inadvertently reverted as a result of merging this PR.

ghubstan:
> It appears the only difference between `CoreWalletService` on current master and the [de8213a](https://github.com/bisq-network/bisq/commit/de8213a95d3e9334a48e7c8f937f7b11b7796afa) version I rebased off is a few lines calling `feeService.getMinFeePerVByte()`, in place of the hard coded fee `BTC_DAO_REGTEST.getDefaultMinFeePerVbyte()`. But I don't think those lines conflict (textually or otherwise) with the changes from this PR, so they shouldn't be inadvertently reverted as a result of merging this PR.

That  `CoreWalletService` diff is related to other changes in `grpc.proto` , `TxFeeRateInfo` (a proto wrapper), etc.  If you can successfully merge, I think I can easily re-add those changes.   I am concerned about other recent api changes to master being overwritten;  I can review after a successful merging of this PR.  [EDIT  This is a false alarm, `grpc.proto` , `TxFeeRateInfo` 
 should be unchanged after the merge.]

chimp1984:
I had locally merged yesterday and my branch matches exactly the rebased one. But would be good if @ghubstan can double check the changes in `CoreWalletService`.

The `walletsManager.maybeAddSegwitKeychains(aesKey);`  is used instead only applying it on the btc keychain.

ghubstan:
> But would be good if @ghubstan can double check the changes in `CoreWalletService`.

Looks OK.  (I just checked the diff between this PR's `CoreWalletService` and current master's. )

ripcurlx (approved):
utACK based on reviews from @chimp1984. I'll test it as part of release testing for v1.6.3.

PR #5111, opened 2021-01-24
Fix log message when opening browser link

ghost:
A series of stack traces were being logged when opening a browser link on linux.  This is because we try three different variations of the command to open the browser: "kde-open", "gnome-open", "xdg-open" and ones not supported by the distro throw an error.

It was not necessary to log huge stack traces if/when the first two of these commands is not successful.
Instead just log one line via `e.toString()`.

Would be nice to get this in 1.5.5 :pray:  :cherries: 

Previously:
image

Now:
image


ripcurlx (approved):
utACK

ripcurlx:
As it is just showing up in the log file, I think we can wait until the next release. WDYT?

PR #5114, opened 2021-01-24
Cash By Mail currency selection to use drop down combo

ghost:
This changes CBM to use one currency per payment account.
Currency list now features all available currencies.

v1.5.5 :pray: :cherries:


ripcurlx (approved):
utACK

PR #5093, opened 2021-01-19
Add api trade simulation scripts

ghubstan:
Two regtest trading simulation scripts are contained in this change:

- `trade-simulation.sh` goes through the steps of creating country based F2F payment accounts for Bob & Alice, plus each step of a trade from `createoffer` to completion.
  
- `limit-order-simulation.sh` shows a way to trigger creation of an offer when a limit price is reached.
  
Each script prints CLI commands and output as they are run, providing a basic tutorial for some of the api's features.

Examples:

- `trade-simulation.sh`          

  Simulate the entire trade protocol between Bob (taker) & Alice (maker),
  where Alice buys 0.1 BTC from Bob, paying in Renminbi (CYN).

  Note the script takes a country code (CN) not a currency code, so the
  script can create the appropriate country based face to face payment accounts.

  `$ apitest/scripts/trade-simulation.sh -d buy -c cn -m 0.0 -a 0.1`

- `limit-order.sh`          

  Create a CAD/BUY 0.1 BTC order at mkt price margin of 0.0% if price falls to
  or below 47900 CAD.

  Note the script takes a country code (CA) not a currency code, so the script
 can create the appropriate country based face to face  payment accounts.

  `$ apitest/scripts/limit-order-simulation.sh -l 47900 -d buy -c CA -m 0.0 -a 0.1`

  Create a USD/SELL 0.1 BTC order at mkt price margin of 0.0% if price rises to
  or above 37200 USD.

  `$ apitest/scripts/limit-order-simulation.sh -l 37200 -d sell -c US -m 0.0 -a 0.1
`

Both scripts depend on functions contained in supporting `bash` and `python3` scripts.

https://github.com/bisq-network/bisq/pull/5084 should be merged before this one.

PR #5116, opened 2021-01-25
Fix bash syntax inconsistencies

ghubstan:
Many unnecessary braces around `${VARIABLE}` references were removed.

Credit to @dmos62 for pointing out syntax inconsistencies while reviewing https://github.com/bisq-network/bisq/pull/5093.


sqrrm (approved):
utACK

Mmm, braces...

PR #5117, opened 2021-01-26
Specify Amazon eGift Card country

### LINKED ISSUE #5098: Specify Amazon eGift Card country
### Description
I can't buy a gift card on amazon.de to send to someone whose email is for an account on amazon.fr. It's important that the sender buys the Amazon eGift Card on same country site as the receivers. This is easily circumvent because any user with an Amazon account can simply log onto different Amazon sites and buy the gift card there. So maybe we don't need to restrict this payment method by countries but it should be explicit to the BTC buyer what Amazon website the BTC seller is using.

@chimp1984

ghost:
Fixes #5098

- Adds countryCode to AmazonGiftCardAccountPayload
- Prompts the user to upgrade their old AmazonGiftCard account where necessary (due to new field).
- Trade buyer step 2 prompts use of the appropriate Amazon website for buying gift card.

[update] [Changed implementation](https://github.com/bisq-network/bisq/pull/5117#issuecomment-771025193) per suggestion from @chimp1984.  Now it upgrades the existing accounts at startup the same way that the upgrade to Revolut was done.  Accounts that are non-EUR are upgraded automatically, EUR accounts prompt the user to choose the Eurozone country:

image

 
---
Example: A French person buying BTC from German person.

---

image

---

image

image

---

image

image

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqSetup.java
+++ b/core/src/main/java/bisq/core/app/BisqSetup.java
@@ -33,6 +33,7 @@ import bisq.core.dao.governance.voteresult.VoteResultException;
 import bisq.core.dao.state.unconfirmed.UnconfirmedBsqChangeOutputListService;
 import bisq.core.locale.Res;
 import bisq.core.offer.OpenOfferManager;
+import bisq.core.payment.AmazonGiftCardAccount;
 import bisq.core.payment.PaymentAccount;
 import bisq.core.payment.RevolutAccount;
 import bisq.core.payment.payload.PaymentMethod;
@@ -177,6 +178,9 @@ public class BisqSetup {
     @Setter
     @Nullable
     private Consumer<List<RevolutAccount>> revolutAccountsUpdateHandler;
+    @Setter
+    @Nullable
+    private Consumer<List<AmazonGiftCardAccount>> amazonGiftCardAccountsUpdateHandler;
     @Setter
     @Nullable
     private Runnable osxKeyLoggerWarningHandler;
@@ -453,6 +457,7 @@ public class BisqSetup {
                 filterWarningHandler,
                 voteResultExceptionHandler,
                 revolutAccountsUpdateHandler,
+                amazonGiftCardAccountsUpdateHandler,
                 daoRequiresRestartHandler);
 
         if (walletsSetup.downloadPercentageProperty().get() == 1) {
--- a/core/src/main/java/bisq/core/app/DomainInitialisation.java
+++ b/core/src/main/java/bisq/core/app/DomainInitialisation.java
@@ -35,6 +35,7 @@ import bisq.core.notifications.alerts.market.MarketAlerts;
 import bisq.core.notifications.alerts.price.PriceAlert;
 import bisq.core.offer.OpenOfferManager;
 import bisq.core.offer.TriggerPriceService;
+import bisq.core.payment.AmazonGiftCardAccount;
 import bisq.core.payment.RevolutAccount;
 import bisq.core.payment.TradeLimits;
 import bisq.core.provider.fee.FeeService;
@@ -189,6 +190,7 @@ public class DomainInitialisation {
                                    Consumer<String> filterWarningHandler,
                                    Consumer<VoteResultException> voteResultExceptionHandler,
                                    Consumer<List<RevolutAccount>> revolutAccountsUpdateHandler,
+                                   Consumer<List<AmazonGiftCardAccount>> amazonGiftCardAccountsUpdateHandler,
                                    Runnable daoRequiresRestartHandler) {
         clockWatcher.start();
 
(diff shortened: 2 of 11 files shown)


pazza83:
Thanks for this, I think the payment info screen at the bottom of the trade window makes this really clear for users.

chimp1984:
Maybe it need also a comment that the trader can remove the payment account only if there is no open trade or offer with it.
Alternatively we could do it similar like in Revolut account update at startup: Detecting if there is an old account and if so require the user to add the missing data and adding that to the existing account instead of removing and adding a new one. That would avoid the issues with open trades/offers and has less hassle for the user. Only issue is that the country dropdown should be added to that popup, so a custom window will be needed.

ghost:
Moved to draft while I implement [suggestion](https://github.com/bisq-network/bisq/pull/5117#issuecomment-769513167) from @chimp1984.  Using https://github.com/bisq-network/bisq/pull/4481 as reference.

Also noticed that the new field has to be excluded from trade contract, e.g. https://github.com/bisq-network/bisq/blob/1a8aba8221f8344eb12cf1cdb5eb9bb751699595/core/src/main/java/bisq/core/payment/payload/RevolutAccountPayload.java#L56-L60

ghost:
@chimp1984 I don't know how to make this change compatible with prior versions.

When both clients are new:
- Alice creates an offer to sell BTC for Amazon.
- Bob takes the offer and is instructed to buy the Amazon GC from Alice's **country of residence**.

Here's the problem:
- Alice is running **old** Bisq (1.5.4) and creates an offer to sell BTC for Amazon.  Her account does not include country.
- Bob is running new Bisq version that includes this feature.
- Bob takes the offer, but gets a failed trade because Alice (old version) did not supply the country in the contract -> `An error occurred at task: TakerVerifyAndSignContract Contracts are not matching`

What's the best way to resolve this?

chimp1984:
Hm... that is always tricky. I think with Revolut it was similar but there was some solution. I would need to dig deeper to figure it out. I think excluding it from json was the solution...It is not a critivcal field so I don't see any problem to not have that in the contract.

From RevolutAccountPayload:
```
 // For backward compatibility we need to exclude the new field for the contract json.
    // We can remove that after a while when risk that users with pre 1.3.8 version trade with updated
    // users is very low.
    @JsonExclude
    @Getter
    private String userName = "";
```

chimp1984:
Account age is another area to take care of, but that data should not be part of the hash anyway.

ghost:
If the field is excluded from contract, the functionality does not work.  The buyer needs to be shown the seller's country so that they buy the amazon eGift from the appropriate country-based website.

chimp1984:
The payment account data are exchanged in the trade messages so the peer gets all data. We use a hash of the contract as json and that causes the problems when adding a new field, but with json exclude that is fixed.

ghost:
Thanks @chimp1984, excluding the field from JSON resolved it again.  For some reason during testing I thought it was missing on new-new clients but now its fine so I'll just chalk it off to me getting confused.  :confused:

ghost:
### Testing summary:


**Check that old Bisq can trade with new Bisq.**

- Run two instances, 1.5.4 (old) and this PR (new) Bisq.  Set up Amazon EUR accounts in each.
- In old Bisq, create one Buy and one Sell offer.
- In new Bisq, create one Buy and one Sell offer.
- Take all offers from each side, resulting in 4 trades.
- Generate/wait for a blockchain confirmation.
- Note that when trading with an old counterparty they do not provide the country information, so it will be blank.
- Initiate payment on each trade and verify that the counterparty screen looks good.
- Mark payment accepted on each trade and complete the trades.
 
---

**Check that account upgrade works for each possible currency:**

- Using old Bisq version (1.5.4), create 11 AmazonGiftCard accounts, one for each currency: 
- EUR, AUD, CAD, GBP, INR, JPY, SAR, SEK, SGD, TRY, USD.
- Close the app, upgrade to this PR (new) Bisq.
- Start up the app and verify that all the accounts are upgraded.  In the case of EUR, verify that you are prompted to choose the Eurozone country for that account.
 
---

**Check that new Bisq can trade with new Bisq.**

Same procedure as first test, but with both instances running this PR (new) Bisq.

ripcurlx (changes requested):
NACK - Trading does work across versions and also the upgrading does work as expected. The only thing I noticed is that if you are establishing a trade between an old an a new Amazon account the current client is seeing two empty fields which might leave the user confused what to do (see `buy giftcard at` and `country`). I think in that case we should advise the user to contact the peer in the trader chat to discuss on which country store the egift card has to be purchased at.

Bildschirmfoto 2021-02-12 um 10 44 32

ghost:
@ripcurlx requested change made.  When seller is old, buyer will be prompted to use Trader Chat:

image

ripcurlx (approved):

PR #5092, opened 2021-01-19
Add payment method "Cash by mail"

ghost:
Closes #1101 

This is built on top of #5080 and #5091.
The code for this PR is contained in https://github.com/bisq-network/bisq/commit/e11209d395bfcea22b7bfd07d38f6137c6144664

----

image

----

image

PR #5118, opened 2021-01-26
Wording change for CashByMail

ghost:
Suggested by @m52go.  https://github.com/bisq-network/bisq/pull/5092#issuecomment-767049339

v1.5.5. :pray: :cherries:


m52go:
Thanks -- subtle, but makes a big difference I think.

ripcurlx (approved):
utACK

PR #5120, opened 2021-01-27
Speed up deposit and transactions view loads

stejbac:
This PR speeds up the loading of the deposit ("Receive funds") and transactions views (under the "Funds" tab), by adding some caches to `WalletService` and `DisputeListService`, lazily loading tx confidence tooltips for the two respective lists of items (`DepositListItem` and `TransactionsListItem`) and applying some low level optimisations to `TransactionsAwareTrade.isRelatedToTransaction`.

This also fixes a fairly minor bug in the transactions view due to a broken `TranscationsAwareTrade.isRefundPayoutTx` method, as well as a somewhat serious memory leak in the deposit view due to missing `BalanceListener` removals from the `BtcWalletService` instance.

The performance issues in these two views had started to become severe with the number of transactions and addresses now in my Bisq wallet. The deposit view was the worst, taking around 25 seconds to load, with bottlenecks in `WalletService.getTransactionConfidence` & `WalletService.getNumTxOutputsForAddress` as seen with JProfiler:

Screenshot from 2021-01-27 02-41-06

The transactions view (with 1210 rows) was taking around 5 seconds to load, with bottlenecks in `TransactionsAwareTrade.isRelatedToTransaction` & `Tooltip.<init>`:

Screenshot from 2021-01-27 02-41-25

They both seem significantly faster (in my Bisq client) with the changes in this PR, with the biggest improvement to the deposit view via the two caches added to `WalletService`. The quadratic time issue (# transactions * # past trades) causing the slowness of the transactions view is alleviated but not completely fixed by these changes. It may be possible to do more extensive lazy loading of the relevant `TransactionsListItem` fields to get further speedups.

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/WalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/WalletService.java
@@ -73,7 +73,9 @@ import org.bitcoinj.wallet.listeners.WalletReorganizeEventListener;
 import javax.inject.Inject;
 
 import com.google.common.collect.ImmutableMultiset;
+import com.google.common.collect.ImmutableSetMultimap;
 import com.google.common.collect.Multiset;
+import com.google.common.collect.SetMultimap;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.MoreExecutors;
@@ -84,12 +86,15 @@ import javafx.beans.property.SimpleIntegerProperty;
 import org.bouncycastle.crypto.params.KeyParameter;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
@@ -116,6 +121,7 @@ public abstract class WalletService {
     private final CopyOnWriteArraySet<BalanceListener> balanceListeners = new CopyOnWriteArraySet<>();
     private final WalletChangeEventListener cacheInvalidationListener;
     private final AtomicReference<Multiset<Address>> txOutputAddressCache = new AtomicReference<>();
+    private final AtomicReference<SetMultimap<Address, Transaction>> addressToMatchingTxSetCache = new AtomicReference<>();
     @Getter
     protected Wallet wallet;
     @Getter
@@ -138,7 +144,10 @@ public abstract class WalletService {
 
         params = walletsSetup.getParams();
 
-        cacheInvalidationListener = wallet -> txOutputAddressCache.set(null);
+        cacheInvalidationListener = wallet -> {
+            txOutputAddressCache.set(null);
+            addressToMatchingTxSetCache.set(null);
+        };
     }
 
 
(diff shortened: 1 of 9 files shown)


core/src/main/java/bisq/core/btc/wallet/WalletService.java
```diff
@@ -408,18 +430,15 @@ public TransactionConfidence getConfidenceForTxId(String txId) {
...
 
-    protected TransactionConfidence getTransactionConfidence(Transaction tx, Address address) {
-        List<TransactionConfidence> transactionConfidenceList = getOutputsWithConnectedOutputs(tx)
-                .stream()
-                .filter(WalletService::isOutputScriptConvertibleToAddress)
-                .filter(output -> address != null && address.equals(getAddressFromOutput(output)))
-                .map(o -> tx.getConfidence())
-                .collect(Collectors.toList());
-        return getMostRecentConfidence(transactionConfidenceList);
+    private TransactionConfidence getTransactionConfidence(Transaction tx, Address address) {
```
chimp1984, line 433: Can we add a @Nullable annotation (was before as well nullable as getMostRecentConfidence is nullable).

core/src/main/java/bisq/core/btc/wallet/WalletService.java
```diff
@@ -408,18 +430,15 @@ public TransactionConfidence getConfidenceForTxId(String txId) {
...
+        boolean matchesAddress = getOutputsWithConnectedOutputs(tx).stream()
+                .anyMatch(output -> address != null && address.equals(getAddressFromOutput(output)));
+
+        return matchesAddress ? getMostRecentConfidence(List.of(tx.getConfidence())) : null;
```
chimp1984: I think we have an issue here:
Assuming there was the same address used as connected output of an input and as the output of the given tx.
We had then 2 different txs in the previous version and the getMostRecentConfidence took the most recent one (the outputs as the connected output of the input could not be newer). Now we drop any of the 2 txs by `anyMatch` and use only the one of the tx (tx output). This would be correct in case we had 2 addresses connected to that tx.
But if the address was only used in a connected output of the input we would take the wrong txConfidence as we would take the actual tx and not the tx associated with the connected output of the input.

chimp1984 (changes requested):
I think the getTransactionConfidence method is not correct. Otherwise all looks good. Was not aware of the guava Multimap framework...Could be used in many places...

core/src/main/java/bisq/core/btc/wallet/WalletService.java
stejbac: OK, I agree it probably should be `@Nullable`.

core/src/main/java/bisq/core/btc/wallet/WalletService.java
stejbac: Yes, I wasn't completely sure of the intent of the original version and tried to preserve its behaviour, but it looks like it may have been broken. The original stream pipeline had the line:

> .map(o -> tx.getConfidence())

But `tx` is the underlying method parameter and so this just fed a repeating list of the same `TransactionConfidence` object into `getMostRecentConfidence`.

I think it may be fixed by restoring the original pipeline, but using the lines:

> .flatMap(o -> Stream.ofNullable(o.getParentTransaction()))
> .map(Transaction::getConfidence)

in place of the `o -> tx.getConfidence` mapping.

This would make the method return null if (there was no matching output and) none of the tx inputs with the given address has a connecting parent tx (though I'm not sure if the latter ever happens). I guess that would be the right behaviour.

chimp1984: Oh, yes looks like that `.map(o -> tx.getConfidence())` was a bug. I think intention was to find all txs where the address was used for receiving or sending BTC. And take the most recent confidence. Unforunately the design flaw to use addresses so dominantely causes many issues in many area. Maybe we can slowely try to convert the domain model to use txos as the core entity and if the UI needs to show addresses we grab the collection of txos related to that address.

chimp1984 (approved):
utACK

sqrrm (approved):

PR #5121, opened 2021-01-27
[WIP] Refactor offerpayload

sqrrm:
This is preparatory work for no fee tx offers. There should be no
behavioral changes with this commit.

Split OfferPayload into abstract OfferPayload and FeeTxOfferPayload with
implementation. Instance variables need to reside in FeeTxOfferPayload
in the same order as previously to keep the generated json contract
format.

Contract is still using FeeTxOfferPayload as I don't think Contract
will be used with any other OfferPayload type.

Only FeeTxOfferPayload will have the ability to update capabilities.
It seems better to just delete the old offer and create a new one
if there is no loss in doing so, which there won't be without the
fee tx.


sqrrm:
Replaced by https://github.com/bisq-network/bisq/pull/5729

PR #5122, opened 2021-01-27
Fix stale trade statistics list view when new entries arrive

stejbac:
Ensure the trade statistics list in `TradesChartsView` doesn't go stale upon new statistics arrivals, by moving a `fillList()` call from the _onChangeConfirmed_ currency combobox event handler to the listener of the `selectedTradeStatistics` `TradesChartsViewModel` field.

(Also avoid unnecessary use of an `ObservableList` as a temporary variable in the `fillList()` method.)

--

I first spotted this bug a little while ago and it appears to have been present since early November 2020, at least looking through the commit history, so it most likely affects both master and the current v1.5.5 release candidate.


chimp1984 (commented):
ACK

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/5122#pullrequestreview-578613629

PR #5126, opened 2021-01-28
Check wallet is avaiable and/or unlocked on createoffer

ghubstan:
Do not attempt to create an offer if the server wallet is unavailable.  And if the wallet is encrypted, do not attempt to create an offer if the wallet is not unlocked.


sqrrm (approved):
utACK

PR #5127, opened 2021-01-28
[WIP] Add atomic offerpayload

sqrrm:
Builds on #5121

## Add AtomicOfferPayload

More fields might be needed for AtomicOfferPayload. Proof of work for
offer broadcast and refresh TTL should probably be handled in
StoragePayload or some other wrapper as it would be useful for most
messages.

Offer protobuf has oneof FeeTxOfferPayload and AtomicOfferPayload. Old
clients will be able to handle FeeTxOfferPayloads as they have index 1
but won't be able to parse Offers with any other payload. This is
acceptable as they can still trade with old style payloads but fail to
parse the new ones and log an error instead. Log is however a bit ugly.


sqrrm:
Replaced by https://github.com/bisq-network/bisq/pull/5729

PR #5128, opened 2021-01-28
Make some trade simulation functions re-usable

ghubstan:
Code duplication needs to be reduced as new scripts are added.

- Refactor the `createpaymentacct` functions.
  The steps required to create a payment account are still displayed, but only for Alice, not Bob.

- Treat the trade protocol simulation as an atomic function.
  This will reduce 'main' simulation script size as new ones are added.
  
- Move `createoffer` command generation to its own function. 

This PRs branch was created off `01-fix-bash-syntax`, and https://github.com/bisq-network/bisq/pull/5116 needs to be reviewed and merged before this one.


sqrrm (approved):
utACK

A bit odd indeed with the atomic function to perform the full trade, but it does what it should in this case.

ghubstan:
> utACK
> 
> A bit odd indeed with the atomic function to perform the full trade, but it does what it should in this case.

I wanted the calling scripts to be simpler.
I broke various protocol steps into their own functions -- `waitfortradedepositpublished`, `waitfortradedepositconfirmed`, `waitfortradepaymentsent`, etc. -- in the next PR.  That made `executetrade` a little smaller, but still atomic.

PR #5129, opened 2021-01-28
Add apitest rolling offer simulation script

ghubstan:
The script demonstrates a way to always keep one offer in the market.  As soon as the offer is taken and the trade is completed, another offer is placed.

PR https://github.com/bisq-network/bisq/pull/5128 should be reviewed and merge before this one.


apitest/scripts/trade-simulation-utils.sh
```diff
@@ -221,6 +221,271 @@ getfirstofferid() {
...
+        echo "$GETTRADE_CMD_OUTPUT"
+        printbreak
+
+        TRADE_DETAIL=$(gettradedetail "$GETTRADE_CMD_OUTPUT")
+        exitoncommandalert $?
+
+        IS_TRADE_PAYMENT_SENT=$(istradepaymentreceived "$TRADE_DETAIL" "$MAKER_OR_TAKER")
+        exitoncommandalert $?
+        printdate "$SELLER:  Has buyer's payment been transferred to seller's fiat account?  $IS_TRADE_PAYMENT_SENT"
+        if [ "$IS_TRADE_PAYMENT_SENT" = "YES" ]
```
sqrrm, line 443: Perhaps a comment to indicate that receipt is assumed as soon as payment sent message is received.

apitest/scripts/rolling-offer-simulation.sh
```diff
@@ -0,0 +1,119 @@
...
+#  - Linux or OSX with bash, Java 10, or Java 11-12 (JDK language compatibility 10), and bitcoin-core (v0.19, v0.20).
+#
+#  - Bisq must be fully built with apitest dao setup files installed.
+#    Build command:  `./gradlew clean build :apitest:installDaoSetup`
+#
+#  - All supporting nodes must be run locally, in dev/dao/regtest mode:
+#           bitcoind, seednode, arbdaemon, alicedaemon, bobdaemon
+#
+#    These should be run using the apitest harness.  From the root project dir, run:
+#    `$ ./bisq-apitest --apiPassword=xyz --supportingApps=bitcoind,seednode,arbdaemon,alicedaemon,bobdaemon --shutdownAfterTests=false`
```
sqrrm, line 22: This kind of simulation would be nice to have for normal regtest that I would run. Could you add instructions on how to run this for my normal alice client without starting the other as I already have them started? I just want a script that uses my normal regtest alice data dir which is already funded and has a history of trades, failed trades and other weird history.

apitest/scripts/rolling-offer-simulation.sh
```diff
@@ -0,0 +1,119 @@
...
+    printdate "ALICE CLI: $CMD"
+    OFFER=$($CMD)
+    exitoncommandalert $?
+    echo "$OFFER"
+    printbreak
+    sleeptraced 3
+
+    # Generate some btc blocks.
+    printdate "Generating btc blocks after publishing Alice's offer."
+    genbtcblocks 3 2
```
sqrrm, line 107: If this is to be used for dev testing I would need to be able to control the block generation completely rather than have them automatically generated.

ghubstan: These simulation scripts are intended to demonstration how the CLI is used to create offers and perform trades.  

A test case that automate one side of a trade, and require manual CLI use from the other side, has been written, but not checked in.  We should talk about the requirements of that test case before I return to work on it and submit it in a new PR.

apitest/scripts/rolling-offer-simulation.sh
ghubstan: This is related to https://github.com/bisq-network/bisq/pull/5129#discussion_r569491295.   We had a chat on keybase and agreed that dev-test script or java testcases will be added, and these scripts' scope will be limited to serving as non-interactive  demos / tutorials.

apitest/scripts/trade-simulation-utils.sh
ghubstan: Comment added in commit 072364b4902370fc1a3757640bec896917955f2d.

sqrrm (approved):
utACK

PR #5133, opened 2021-01-30
[v1.5.5] Prevent overlapping in offer book tools bar

ripcurlx:
I've adjusted the layout to prevent hidding of information in our minimum resolution. 

**Before**
Bildschirmfoto 2021-01-30 um 16 36 14

**After**
Bildschirmfoto 2021-01-30 um 16 49 01


ghost (dismissed):
Tested, ACK :+1: 

Minor nit: `create offer` button alignment looks a bit off, but not a big issue.

ripcurlx:
> Tested, ACK 👍
> 
> Minor nit: `create offer` button alignment looks a bit off, but not a big issue.

Ah - true this is as it is now center - aligned by default in this HBox. I'll push a fix for it.

ripcurlx:
@jmacxx The alignment should be now similar to what it was before.
Bildschirmfoto 2021-02-01 um 09 29 56

sqrrm (approved):
utACK

PR #5136, opened 2021-01-30
Show popup if no filter is delivered

chimp1984:
On mainNet we expect to have received a filter object, if not show a popup to the user to inform the Bisq devs.


sqrrm (approved):
utACK

PR #5137, opened 2021-01-31
Replace deprecated bisq explorer settings

ghost:
If user preferences were set to a federated explorer that has been retired, switch it to another explorer from the provided list.
Users can still choose their own explorer, or enter a custom explorer address (such as their local node).


ripcurlx (dismissed):
utACK

ripcurlx:
@jmacxx You missed to sign your commit. Everything else is looking fine.

ghost:
@ripcurlx  :+1: Signed it now.

ripcurlx (approved):
utACK

PR #5139, opened 2021-01-31
onUpdatedDataReceived must be called in the correct order

chimp1984:
We cannot use a listener at RequestDataManager as the order
is not defined if doing so.
So we use P2PService as our controlling entity to call
further clients in the correct order.


ripcurlx:
@chimp1984 Could you please point it to master, so we have the fix there as well and I'll cherry-pick it from there into the release branch? Thanks!

ripcurlx:
Superseded by #5141

PR #5141, opened 2021-01-31
onUpdatedDataReceived must be called in the correct order.

chimp1984:
We cannot use a listener at RequestDataManager as the order
is not defined if doing so.
So we use P2PService as our controlling entity to call
further clients in the correct order.


ghost (approved):
Tested, ACK :+1:

ripcurlx (approved):
utACK

PR #5142, opened 2021-02-01
Add coin input control

chimp1984:
Adds input selection for BSQ send transactions. We use a utxo based model in contrast to the address based model in the Funds screen. If any dev wants to apply that to the funds screen as well would be good. Also it is better IMO to hide it in a popup as it is a bit technical and might confuse some users.

Screen Shot 2021-01-31 at 21 30 01

Screen Shot 2021-01-31 at 21 29 41


core/src/main/java/bisq/core/btc/wallet/NonBsqCoinSelector.java
```diff
@@ -60,9 +64,9 @@ protected boolean isTxOutputSpendable(TransactionOutput output) {
         return !daoStateService.existsTxOutput(key) || daoStateService.isRejectedIssuanceOutput(key);
     }
 
-    // BTC utxo in the BSQ wallet are usually from rejected comp request so we don't expect dust attack utxos here.
+    // Prevent to use dust attack utxos
```
sqrrm:
```suggestion
    // Prevent usage of dust attack utxos
```

core/src/main/java/bisq/core/btc/wallet/BisqDefaultCoinSelector.java
```diff
@@ -65,7 +74,16 @@ public BisqDefaultCoinSelector() {
     public CoinSelection select(Coin target, List<TransactionOutput> candidates) {
         ArrayList<TransactionOutput> selected = new ArrayList<>();
         // Sort the inputs by age*value so we get the highest "coin days" spent.
-        ArrayList<TransactionOutput> sortedOutputs = new ArrayList<>(candidates);
+
+        ArrayList<TransactionOutput> sortedOutputs;
+        if (utxoCandidates != null) {
+            sortedOutputs = new ArrayList<>(utxoCandidates);
+            // We we reuse the selectors we reset the transactionOutputCandidates field
```
sqrrm:
```suggestion
            // We reuse the selectors. Reset the transactionOutputCandidates field
```

desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
```diff
@@ -171,11 +188,16 @@ protected void activate() {
 
         bsqWalletService.addBsqBalanceListener(this);
 
+        // We reset the input selection at active to have all inputs selected, otherwise the user
```
sqrrm:
```suggestion
        // We reset the input selection at activate to have all inputs selected, otherwise the user
```

desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
```diff
@@ -282,16 +322,49 @@ private void onSendBsq() {
...
 
+    private void onBsqInputControl() {
+        List<TransactionOutput> unspentTransactionOutputs = bsqWalletService.getSpendableBsqTransactionOutputs();
+        if (bsqUtxoCandidates == null) {
+            bsqUtxoCandidates = new HashSet<>(unspentTransactionOutputs);
+        } else {
+            // If we had some selection stored we need to update to already spent entries
+            bsqUtxoCandidates = bsqUtxoCandidates.stream().
+                    filter(e -> unspentTransactionOutputs.contains(e)).
+                    collect(Collectors.toSet());
```
sqrrm, line 339: This would remove transactions that are no longer spendable BSQ outputs from the selected outputs. I think that could be confusing. Better to check if the `bsqUtxoCandidates` are still all spendable, if they are then all is good, if not, then reset the selection to allow choosing between all `unspentTransactionOutputs` again.

desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
```diff
@@ -306,43 +379,80 @@ private void addSendBtcGroup() {
...
+
+    private void onBtcInputControl() {
+        List<TransactionOutput> unspentTransactionOutputs = bsqWalletService.getSpendableNonBsqTransactionOutputs();
+        if (btcUtxoCandidates == null) {
+            btcUtxoCandidates = new HashSet<>(unspentTransactionOutputs);
+        } else {
+            // If we had some selection stored we need to update to already spent entries
+            btcUtxoCandidates = btcUtxoCandidates.stream().
+                    filter(e -> unspentTransactionOutputs.contains(e)).
+                    collect(Collectors.toSet());
```
sqrrm, line 402: Same comment as for the BSQ, could be confusing if the selection changes without the users knowing about it

sqrrm (commented):
Testing issue:
1. Start with a BSQ wallet with no BTC balance
2. BSQ selection popup shows when button is pressed
3. Send BTC to BSQ receive address (remove leading "B")
4. BTC balance section shows in BSQ wallet but neither "select inputs" button work.

desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
chimp1984: To request the spendable via bsqWalletService.getSpendableBsqTransactionOutputs() and then check if the bsqUtxoCandidate is in that set does check if the bsqUtxoCandidate is spendable. We use the coinSelector for that and there is no API for checking one utxo if it is still spendable.

Or do I misunderstand what you mean?

desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
chimp1984: If you make a send tx and spend a previous selected utxo we need to remove that. We always show that spendable utxos and those are marked the user has selected. A previous selected which got spent need to be removed.

chimp1984:
> BTC balance section shows in BSQ wallet but neither "select inputs" button work.

There was a bug with the button handlers. Fixed in b37dc15

ripcurlx:
Would be great to reset the validation of the input fields similar as it was before, so it doesn't look like an error after coming back from the success popup.

Bildschirmfoto 2021-02-18 um 21 43 54

sqrrm (approved):
ACK

PR #5149, opened 2021-02-02
Slow down burnt BSQ moving average 4 fold

dmos62:
Reduces the BSQ burnt moving average period in the BSQ Supply chart from 15-day to 60-day. This makes the moving average more readable. 15-day MA was for the early days of tracking BSQ burn when the burn events were more frequent and smaller.

15-day burnt Bisq MA (current state):

15-day burnt Bisq MA

30-day burnt Bisq MA:

30-day burnt Bisq MA

60-day burnt Bisq MA (proposed state):

60-day burnt Bisq MA

Note: an exponential moving average (as opposed to a simple moving average, which is what this chart is using now) would further improve readability, but implementing these algorithms on Java Streams is not trivial, so unless there's demand for it, I'll leave that as is.

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -2445,7 +2445,7 @@ dao.factsAndFigures.supply.compRequestIssueAmount=BSQ issued for compensation re
 dao.factsAndFigures.supply.reimbursementAmount=BSQ issued for reimbursement requests
 
 dao.factsAndFigures.supply.burnt=BSQ burnt
-dao.factsAndFigures.supply.burntMovingAverage=15-day moving average
+dao.factsAndFigures.supply.burntMovingAverage=60-day moving average
 dao.factsAndFigures.supply.burntZoomToInliers=Zoom to inliers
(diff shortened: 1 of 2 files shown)


chimp1984:
I fear we have worked in parallel....
See: https://github.com/bisq-network/bisq/pull/5156#issuecomment-773822610
In my version I have removed the second chart and added toggles for series, interval selectors and a timeline navigation. I think then the problems with outliers is gone as you simly hide the series which blows up the y-axis. Zooming into the time will help as well.

dmos62:
@chimp1984 Your new charts are looking good. Please close this if your work will be merged for next release.

chimp1984:
I am still working on it. Once its finished we can see if a MA is still needed.

dmos62:
I think that being able to use a longer interval, e.g. monthly, makes an MA
unnecessary.

wallclockbuilder (approved):
UtAck

ripcurlx:
Closing as by https://github.com/bisq-network/bisq/pull/5149#issuecomment-774637583

PR #5155, opened 2021-02-04
Fix find pid on OSX

ghubstan:
The trade simulation scripts check that bitcoind, a seed node, an arbitration node, and Bob & Alice nodes are running at startup, but the bash `pgrep` command used to find the PIDs does not work on OSX (darwin20).  An adjustment was made to the env script to find java PIDs with `ps` and `awk`,  if running on OSX.

Two additonal startup checks were added: make sure the script is running on supported OS (Linux or OSX), and `bitcoin-cli` is in the `PATH`.

Also removed some usage error text about registering dispute agents;  the scripts do that now.


ripcurlx (approved):
utACK

Issue #5151, opened 2021-02-04
Small subset of Offer Maker Tx do not reach the mempool


ghost:
A number of offers have been showing up in the offer book with transaction ids that cannot be found in either the mempool or the blockchain.  When such an offer is taken it causes an immediate failed trade.  Not much is known about the cause of this but this issue is to track progress of the investigation.

- These phantom maker txIds are coming in at a rate of 1 or 2 per day, out of approx. 100 new offers per day.
- I am checking all Bisq offers using a javascript tool that looks up each txid on a block explorer.  The txIds are extracted in a log file from Bisq's offer book.
- We've in-app messaged the onion address of some offers in this state asking them to get in touch with Bisq developers.  So far no one has come forward.
- Some of them have, after a while (like 24 hours), managed to correct the issue and the txid will suddenly be found by the block explorer with a few confirmations.  Looking at those txids nothing looks out of the ordinary.  These "fixing itself" events seem to coincide with the offer maker going offline and coming back online again.  **Perhaps a reboot / SPV resync fixes it?**
- It is assumed that some of them end up causing failed-trade fee reimbursement by the taker. e.g. https://github.com/bisq-network/support/issues/809 looks like one.

---

**Example 1:** (offer has been open for 7 days at the time of writing this; this user has another similar offer open with the same problem.  Has not responded to a couple of in-app messages from devs yet.)

![phantomA](https://user-images.githubusercontent.com/47253594/106838303-a588d100-6661-11eb-8fd8-420d60413e18.png)
![phantomB](https://user-images.githubusercontent.com/47253594/106838308-a7529480-6661-11eb-99ef-2775da208805.png)

---

**Example 2:**

![phantom0](https://user-images.githubusercontent.com/47253594/106837471-28109100-6660-11eb-828a-1eda1381811a.png)
![phantom1](https://user-images.githubusercontent.com/47253594/106837474-2a72eb00-6660-11eb-8077-f69b69d32037.png)

Conza88:
> * We've in-app messaged the onion address of some offers in this state asking them to get in touch with Bisq developers.  So far no one has come forward.

Sorry off topic: interested if any documentation on how that can be done? ;o

ganaque:
This is the issue I'm facing

ghost:
[...]

PR #5160, opened 2021-02-06
Verify maker & taker fee transactions via Mempool lookup

### LINKED ISSUE #5119: Add mempool lookup for trade transactions
### Description

To avoid issues with invalid transactions in the trade process we should add a mempool lookup to detect early if a transaction did not get accepted in the mempool.

Beside the use case to verify if a tx is available and valid we can use that for verifying the trade fee payment of the peer.

#### Background context
BitocinJ only knows about own transactions not about the peers transaction. Also we do not get notified about an invalid transaction in case one of the inputs was already spent (TODO: check out if we receive an error back and if so if we can handle that). 

#### Limitations
- It takes a bit of time between the publishing of a transaction until is is found on an explorer. Depending on the use case that need to be considered. 
- There is no guarantee that every explorer node receives every transaction. 

Those edge cases for false positives should be rather rare and the benefit will ourweight the risks of false positives. 

#### Maker fee tx:
The maker can periodically check that their maker fee txs are found in the mempool. Once found they get removed from the list of transactions to check.

#### Taker verifies maker fee tx:
The maker fee tx ID is in the offer. At take offer the taker could request the mempool API to see if the maker fee transaction is found. If not, the taker receives a warning popup. They can manually lookup the tx as well to be sure it was not a false positive. For that case they can override the suggestion of the app to not take that offer.

#### Taker verifies maker fee payment:
When requesting the maker fee tx the taker can verify that the trade fee payment was as expected. There need to be used historical fee values (DAO parameters) in case the offer has been created in the past. There need to be a tolerance window as it might be that the offer was created at the last block of a cycle but the maker fee tx got into the block in the next cycle which could have changed the maker fee. One way would be to take the fee of the cycle of the maker fee tx and if that would fail to go one cycle back and try again with that previous value. 

#### Verify that deposit tx inputs are matching the trade fee txs outputs. 
This is not a mempool lookup issue but simply an additional check to detect bugs in case the deposit tx would be incorrect (there have been a past bug where the deposit tx was taken from another trade).

#### Maker verifies taker fee payment
We do not need to verify the taker fee transaction directly as the deposit transaction would never get confirmed if any of the inputs would be invalid. If the deposit transaction is not confirmed the trade protocol stays at step 1 and no fiat/altcoins will be transferred. To verify the fee payment of the taker we cannot do it reliably at the take offer phase in the trade protocol as the time is too short and the risks for false positives too high. But we can verify it later once the deposit transaction has been confirmed. In case the maker is the buyer they would never get to step 2 to start the payment in case the fee payment was incorrect. In case the maker is the seller, they would never get to step 3 for confirming receipt and releasing the BTC if the fee payment was not valid.

#### Arbitrator checks if delayed payout transaction is confirmed
When an arbitration ticket gets opened the arbitrator checks if the delayed payout transaction is confirmed. If not they need to wait until its confirmed. 

### Error handling
In case the lookup does not succeed, the app should use a second mempool API and try again. Only if that fails as well we consider the transaction as not found. The nodes are selected randomly.  

### Privacy
We use the mempool explorers operated by Bisq contributors. Those run as hidden service, so the only thing the explorer learns is that such a lookup request is likely a Bisq transaction, but as the trade transactions have some sort of "fingerprint" due their structure and connections that should not be a critical issue. 

### Preferences
We should allow users to setup their own mempool explorer (we only support one API type) in the preferences. 
As the feature is not only a protection for the user but also for the peer and for the Bisq DAO (that trade fees are paid correctly) we should not make that feature optional.

ghost:
Closes #5119

- Queries mempool at checkOfferAvailability stage (take offer).
- Validates maker transaction and trading fee.
- Indicates status at UI level (see screenshots below)
- Queries mempool at trade initiation (step 1&2) before fiat payment initiated.
- Maker validates taker transaction and trading fee.
- Informs user to open a dispute if the trading fees were not paid
- Arbitrator ticket informs about the confirmation status of delayed payout

---

**Screenshots**

https://imgur.com/a/tSaUgRL

---

**Safeguards**

Can be disabled universally via a filter, and individually via config setting.  

In regtest mode the API calls are bypassed, returning success status.

---

**Accuracy**

The fee schedule is set via DAO parameters.  Therefore checking that the correct fee was paid involves comparing the block height of an offer Tx to the DAO parameters which were relevant at that time.

I observed some variance in fee payments, perhaps because different users nodes may have been out of consensus with the DAO state.  

The validation code has a threshold at which it considers fees paid to be acceptable.  Currently this is set at 95%, i.e. if the fee paid is >= 95% of what was expected, the validation passes.

Very old offers (of which there are a handful still out there), use fee addresses which are not known but perhaps were used prior to the DAO.  So a threshold of blockheight 599999 is used such that the unknown fee addresses are allowed to pass when the offer Tx is older than the threshold.  Alternatively we could add [1PUXU1M...](https://mempool.space/address/1PUXU1MQ82JC3Hx1NN5tZs3BaTAJVg72MC) and [18GzH11...](https://mempool.space/address/18GzH11T5h2fpvUoBJDub7MgNJVw3FfqQ8) to the validator's list of accepted fee addresses.

    private final double feeTolerance = 0.95;     // we expect fees to be at least 95% of target
    private final long blockTolerance = 599999L;  // allow really old offers with weird fee addresses

Out of ~450 current offers ~9 are flagged as having underpaid the trading fee.

----

**Checks own open offers**

To implement the requirement that one's own offers be checked I integrated a validation in the TriggerPriceService so that if an offer is found to be bad it gets automatically disabled.  This can help prevent the issues we've been seeing of [offers that do not get published to the mempool for hours/days sometimes](https://github.com/bisq-network/bisq/issues/5151).  Now if an offer Tx is not found, it will get disabled and the user will be motivated to investigate.

----

**Testing**

Tests were written in `TxValidatorTest` to check the various error scenarios, be it underpaid BSQ, BTC, unknown fee address, or unknown tx.  Most of the tests use data from real-world scenarios like the dust-fee scam.

----

### PR Checklist:

- [X] Validate offer maker tx and fees.
- [X] Arbitrator checks that delayed payout tx is confirmed.
- [x] Verify deposit tx inputs match maker/taker outputs.
- [x] Maker verify taker fee payment.
- [x] Validates own open offers, disables ones that are bad.


Conza88:
I thought it did this already 😳

ripcurlx:
> I thought it did this already 😳

Only for transactions that are linked to your wallet (e.g. Bisq fee tx of the trading peer wasn't checked)

Conza88:
> * If there is a problem it warns user to back out (user can still continue).

Would there be a scenario when that was ever advisable? Or worth doing?

core/src/main/java/bisq/core/provider/mempool/TxValidator.java
```diff
@@ -0,0 +1,379 @@
...
+        if (jsonVIn0Value == null || jsonFeeValue == null) {
+            throw new JsonSyntaxException("vin/vout missing data");
+        }
+        Coin expectedFee = isMaker ?
+                getMakerFeeHistorical(false, tradeAmount, blockHeight) :
+                getTakerFeeHistorical(false, tradeAmount, blockHeight);
+        long feeValue = jsonVIn0Value.getAsLong() - jsonFeeValue.getAsLong();
+        // if the first output (BSQ) is greater than the first input (BSQ) include the second input (presumably BSQ)
+        if (jsonFeeValue.getAsLong() > jsonVIn0Value.getAsLong()) {
+            // in this case 2 or more UTXOs were spent to pay the fee:
```
chimp1984, line 216: The code handles only the case of 1 or 2 BSQ inputs. Maybe also add a comment that we building on the concention that BSQ inputs are the before BTC inputs, which is not a consensus rule (for outputs it is) but our code handles it like that, so for the fee check that is good enough as we do not expect that ppl use custom txs for those txs.
It could be also a rare case that there is no BSQ change output. That can be if the input is exactly the amount to be burned or if the change is < dust, then it gets added to miner fee.

Another issue is that we do not check if the burned satoshis are BSQ or BTC. That would require that the tx is confirmed as only at confirmation the BSQ explorers recognize a BSQ tx. As confirmation will happen later we need to do that check after the deposit tx is confirmed in the trade protocol. So before the buyer sends the fiat he will check the takers BSQ fee it the BSQ explorer returns it as a valid BSQ tx and then before the seller confirms receipt to check the makers fee tx.

core/src/main/java/bisq/core/provider/mempool/TxValidator.java
```diff
@@ -0,0 +1,379 @@
...
+
+    private boolean checkFeeAmountBSQ(String jsonTxt, Coin tradeAmount, boolean isMaker, long blockHeight) {
+        JsonArray jsonVin;
+        JsonArray jsonVout;
+        jsonVin = getVinAndVout(jsonTxt).first;
+        jsonVout = getVinAndVout(jsonTxt).second;
+        JsonObject jsonVin0 = jsonVin.get(0).getAsJsonObject();
+        JsonObject jsonVout0 = jsonVout.get(0).getAsJsonObject();
+        JsonElement jsonVIn0Value = jsonVin0.getAsJsonObject("prevout").get("value");
+        JsonElement jsonFeeValue = jsonVout0.get("value");
```
chimp1984, line 206: The BSQ fee check is tricky. Maybe a better approach is to postpone it directly to the trade protocol states when the deposit tx is confirmed and use the BSQ explorer for delivering the fee.
In BsqWalletService.completeBsqTradingFeeTx there have been added a while ago a change which made the input 0 a BTC output in case there is no BSQ output. I think that change is not needed (at least I tested it without and it worked, but don't want to touch that at the moment...). So to take that additional option into account all gets more complicated....

core/src/main/java/bisq/core/trade/TradeDataValidation.java
```diff
@@ -355,6 +355,27 @@ public static void validatePayoutTxInput(Transaction depositTx,
         }
     }
 
+    public static void validateDepositInputs(Trade trade) throws InvalidTxException {
+        // assumption: deposit tx always has 2 inputs, the maker and taker
```
chimp1984, line 359: Correct.

core/src/main/java/bisq/core/trade/TradeDataValidation.java
```diff
@@ -355,6 +355,27 @@ public static void validatePayoutTxInput(Transaction depositTx,
...
+            throw new InvalidTxException("Deposit transaction is null or has unexpected input count");
+        }
+        Transaction depositTx = trade.getDepositTx();
+        String txIdInput0 = depositTx.getInput(0).getOutpoint().getHash().toString();
+        String txIdInput1 = depositTx.getInput(1).getOutpoint().getHash().toString();
+        String contractMakerTxId = trade.getContract().getOfferPayload().getOfferFeePaymentTxId();
+        String contractTakerTxId = trade.getContract().getTakerFeeTxID();
+        boolean makerFirstMatch = contractMakerTxId.equalsIgnoreCase(txIdInput0) && contractTakerTxId.equalsIgnoreCase(txIdInput1);
+        boolean takerFirstMatch = contractMakerTxId.equalsIgnoreCase(txIdInput1) && contractTakerTxId.equalsIgnoreCase(txIdInput0);
+        if (!makerFirstMatch && !takerFirstMatch) {
```
chimp1984, line 370: The order is determined by buyer/seller role (not maker/taker) but your check here is sufficient...

core/src/main/java/bisq/core/provider/mempool/MempoolService.java
```diff
@@ -0,0 +1,250 @@
...
+import java.util.Optional;
+import java.util.function.Consumer;
+
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+import javax.annotation.Nullable;
+
+@Slf4j
+public class MempoolService {
```
chimp1984, line 55: I assume it should be a @Singleton . At the moment we get created 2 instances.

core/src/main/java/bisq/core/provider/mempool/MempoolService.java
```diff
@@ -0,0 +1,250 @@
...
+                          FilterManager filterManager,
+                          DaoFacade daoFacade,
+                          DaoStateService daoStateService) {
+        this.socks5ProxyProvider = socks5ProxyProvider;
+        this.config = config;
+        this.preferences = preferences;
+        this.filterManager = filterManager;
+        this.daoFacade = daoFacade;
+        this.daoStateService = daoStateService;
+        this.btcFeeReceivers = getAllBtcFeeReceivers();
```
chimp1984: As we rely here on the daostate we need to fill btcFeeReceivers via a onAllServicesInitialized method.

core/src/main/java/bisq/core/provider/mempool/MempoolService.java
```diff
@@ -0,0 +1,250 @@
...
+        mempoolRequest.getTxStatus(future, txValidator.getTxId(), null);
+    }
+
+    private FutureCallback<String> callbackForMakerTxValidation(MempoolRequest theRequest, TxValidator txValidator, Consumer<TxValidator> resultHandler) {
+        outstandingRequests++;
+        FutureCallback<String> myCallback = new FutureCallback<>() {
+            @Override
+            public void onSuccess(@Nullable String jsonTxt) {
+                outstandingRequests--;
+                UserThread.execute(() -> {
```
chimp1984, line 159: As we get called from another thread here its more safe to move the outstandingRequests update into the userthread mapped runnable as well.

core/src/main/java/bisq/core/provider/mempool/MempoolService.java
```diff
@@ -0,0 +1,250 @@
...
+                });
+            }
+            @Override
+            public void onFailure(Throwable throwable) {
+                outstandingRequests--;
+                log.warn("onFailure - {}", throwable.toString());
+                if (theRequest.switchToAnotherProvider()) {
+                    retryValidateOfferMakerTx(theRequest, txValidator, resultHandler);
+                } else {
+                    UserThread.execute(() -> {
```
chimp1984: I think here as well its more safe to move all code into the runnable to avoid potential threading issues.

desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferView.java
```diff
@@ -898,6 +899,21 @@ private void addButtons() {
         });
     }
 
+    private void nextStepCheckMakerTx() {
+        // the tx validation check has had plenty of time to complete, but if for some reason it has not returned
+        // we continue anyway since the check is not crucial.
+        // note, it would be great if there was a real tri-state boolean we could use here, instead of -1, 0, and 1
```
chimp1984, line 905: Using Optional<Boolean> would be an option to handle that. Another to use boxed Boolean type and check for null but Optional is more clear IMO.

desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferViewModel.java
```diff
@@ -462,6 +463,7 @@ private void updateButtonDisableState() {
         boolean inputDataValid = isBtcInputValid(amount.get()).isValid
                 && dataModel.isMinAmountLessOrEqualAmount()
                 && !dataModel.isAmountLargerThanOfferAmount()
+                && dataModel.mempoolStatus.get() >= 0
```
chimp1984: Do we want to block in case response is slow (tor can be slow)?

chimp1984 (changes requested):

desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferViewModel.java
ghost: Initially I had it not blocking the next button, but depending on the speed of the mempool response you'd be told about a bad offer sometimes but not others.  So it felt unpredictable, and since the `checkOfferAvailabilityRequest` already blocks the next button I thought it best to have the mempool check do the same.

ghost:
:+1:  Thanks @chimp1984 for the review & suggestions.  I applied your changes, and will think about the suggestion of checking BSQ fees using a BSQ explorer.

ripcurlx:
@jmacxx I'm reviewing this PR right now and think it would be great to get it into v1.6.0. Could you please resolve the conflicts mentioned and I'm happy to merge it after a positive review.

core/src/main/java/bisq/core/offer/TriggerPriceService.java
```diff
@@ -152,6 +156,20 @@ private void checkPriceThreshold(MarketPrice marketPrice, OpenOffer openOffer) {
             openOfferManager.deactivateOpenOffer(openOffer, () -> {
             }, errorMessage -> {
             });
+        } else if (openOffer.getState() == OpenOffer.State.AVAILABLE) {
+            // check the mempool if it has not been done before
+            if (openOffer.getMempoolStatus() < 0 && mempoolService.canRequestBeMade()) {
+                mempoolService.validateOfferMakerTx(openOffer.getOffer().getOfferPayload(), (txValidator -> {
+                    openOffer.setMempoolStatus(txValidator.isFail() ? 0 : 1);
```
ripcurlx, line 163: So if it fails to validate the tx the offer will be deactivated after the next price update (~60 sec later), right?

ghost: Yes exactly.  Cleaner than duplicating the deactivation code within the async response.

One thing I'd like to do though is delay 60 seconds before checking a new offer, because at the moment it checks immediately upon creation which can be too soon for the tx to have propagated.

ripcurlx:
I just got following error while trying to open a dispute as mediator:
```
März-17 15:33:03.377 [JavaFX Application Thread] ERROR bisq.common.setup.CommonSetup: Uncaught Exception from thread JavaFX Application Thread 
März-17 15:33:03.377 [JavaFX Application Thread] ERROR bisq.common.setup.CommonSetup: throwableMessage= null 
März-17 15:33:03.378 [JavaFX Application Thread] ERROR bisq.common.setup.CommonSetup: throwableClass= class java.lang.NullPointerException 
März-17 15:33:03.381 [JavaFX Application Thread] ERROR bisq.common.setup.CommonSetup: Stack trace:
java.lang.NullPointerException
	at bisq.core.provider.mempool.TxValidator.initialSanityChecks(TxValidator.java:260)
	at bisq.core.provider.mempool.TxValidator.parseJsonValidateTx(TxValidator.java:131)
```
This happens as `jsonTxt` is "" on Regtest.

desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferDataModel.java
```diff
@@ -241,6 +254,15 @@ void initWithData(Offer offer) {
             }
         });
 
+        mempoolStatus.setValue(-1);
+        mempoolService.validateOfferMakerTx(offer.getOfferPayload(), (txValidator -> {
+            mempoolStatus.setValue(txValidator.isFail() ? 0 : 1);
+            if (txValidator.isFail()) {
+                mempoolStatusText = txValidator.toString();
+                log.warn("Mempool check of OfferFeePaymentTxId returned errors: [{}]", mempoolStatusText);
```
ripcurlx: This should be 
```suggestion
                log.info("Mempool check of OfferFeePaymentTxId returned errors: [{}]", mempoolStatusText);
```

ghost: :+1: There are a lot of other similar messages that should probably be changed.  I'll go through them now and make the change to info.

ripcurlx: Recognized it as I'm always checking for warning in the log :wink:

ripcurlx:
It wasn't changed in this PR, but I missed it in the other one. 
Please don't use `version-new` as class to highlight this text, as there might be a problem in the future if someone changes this class. It also shows a hand icon which indicates that this underlined label is clickable.

`tradePeriodEnd.getStyleClass().add("version-new"); // highlight field when the trade period is still active`

Better would be e.g. to create a new class `.alert` without underline and hand, so it doesn't look like a link.
```
.alert {
    -fx-text-fill: -bs-rd-error-red;
}

```
Bildschirmfoto 2021-03-17 um 16 08 06

ripcurlx:
And one more thing I missed in the other PR and feels a little bit weird for the user. If a dispute is closed you get the notification popup. When you click on it, it opens the support section and the issue is selected. But it is greyed out and users might not know what to do. Maybe it is better to open in this case automatically the same action as a double click would trigger. WDYT?

ripcurlx:
Also it would be good to show a (1) badge over the chat icon for the system message, so people are looking into it. In the past the chat was open by default in-line. So at least a notification would be good.

desktop/src/main/java/bisq/desktop/main/offer/takeoffer/TakeOfferDataModel.java
```diff
@@ -173,6 +185,7 @@ protected void activate() {
             tradeManager.checkOfferAvailability(offer,
                     false,
                     () -> {
+                log.warn("OfferAvailability completed!");
```
ripcurlx: I think something went wrong with commit squashing as the `log.warn` is in again.

ghost: Thanks!  My bad for debugging using log.warn()

ripcurlx (approved):
ACK - tested it on Regtest and Mainnet. So far everything is looking fine besides the minor issues that are fixed now.

PR #5172, opened 2021-02-08
Fix NPE at offer entry when trading account switched

### LINKED ISSUE #5166: Null Pointer exception
### Description

I switched the payment account while preparing an offer to buy bitcoin

#### Version

1.5.5

### Steps to reproduce

can't

### Expected behaviour

no null pointer exception

### Actual behaviour

see logfile (null pointer exception dialog)

### Screenshots

too late, screen cleared after clicking on report issue

#### Device or machine

macOS Big Sur on MacMini M1

#### Additional info
[bisq.log](https://github.com/bisq-network/bisq/files/5938874/bisq.log)

ghost:
Fixes #5166

To reproduce the original error:

- You have two or more SEPA accounts.
- Fill in fields for a new offer using SEPA -> e.g. Buy 0.01 EUR/BTC at 1% above market price.
- Enter a value in the field "Deactivate offer if market price is above" value.
- Switch the trading account to the other SEPA account.
- NPE.


sqrrm (approved):
ACK

PR #5191, opened 2021-02-13
Fix startup error: "You must have bootstrapped before adding data to the P2P network"

### LINKED ISSUE #5190: You must have bootstrapped before adding data to the P2P network
### Steps to reproduce

Any time trying to start the application the error is thrown and that's it.

### Expected behaviour

Open even after the error message

### Actual behaviour

Stuck with one useless application and losing money

### Screenshots

image

#### Device or machine

Linux

#### Additional info

Restarted and several times re-trying.
[error_bisq.log](https://github.com/bisq-network/bisq/files/5973929/error_bisq.log)

ghost:
`TriggerPriceService` is changed to use a `BootstrapListener` on the P2PService so that it can start after the P2P network is ready.

Fixes #5190 : Prevents an Uncaught Exception error at startup.

----

### To test (pre):

1. Create offer Buy, choosing "Deactivate offer if over x" price with a trigger price just above current spot.
2. Close down Bisq and wait for the market price to move over the deactivation price.
3. Start Bisq, the error message above will pop up.

### To test (post):

After applying this patch, at step 3 the offer will be deactivated successfully on startup and not produce an error popup.

### DIFF
--- a/core/src/main/java/bisq/core/offer/TriggerPriceService.java
+++ b/core/src/main/java/bisq/core/offer/TriggerPriceService.java
@@ -23,6 +23,9 @@ import bisq.core.monetary.Price;
 import bisq.core.provider.price.MarketPrice;
 import bisq.core.provider.price.PriceFeedService;
 
+import bisq.network.p2p.BootstrapListener;
+import bisq.network.p2p.P2PService;
+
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
utACK

PR #5196, opened 2021-02-16
Prevent erroneous logging of 'SignaturePubKey in message does not match'

### LINKED ISSUE #5194: Error message repeated: `SignaturePubKey in message does not match`
Since v1.5.5 the following error is logged quite frequently:

`ERROR b.c.t.p.TradeProtocol: SignaturePubKey in message does not match the SignaturePubKey we have set for our trading peer. `

- Happens during trading, and when your offers are checked (OfferAvailabilityRequest/Response).
- It does not seem to affect functionality, but some users have contacted support worried by it.

[I am investigating / working on a fix.]

ghost:
The order of checks was wrong.

PRE: it was checking the SignaturePubKey for messages that were handled by other subsystems or other trades.
POST: make sure that this is a relevant message before checking that its signature matches the trade.

Fixes #5194 

---

### To reproduce the issue:

1. Have 3 Bisq instances: Alice, Bob, and Chris.
2. Alice creates an offer to buy.
3. Bob and Chris both consider taking the offer => Alice receives normal "OfferAvailabilityRequest" messages from them.
4. Bob takes the offer, initiating a trade with Alice.
5. Alice creates another offer.
6. Chris considers taking the offer => Alice logs 2 error messages `SignaturePubKey in message does not match`.
7. Anyone other than Bob considers taking the offer => Alice logs the same error messages.
8. Anyone other than Bob trades with Alice => Alice logs the same error messages.

### Verification:

After the fix, trades can be executed simultaneously between any number of peers without the error being logged.

### DIFF
--- a/core/src/main/java/bisq/core/trade/protocol/TradeProtocol.java
+++ b/core/src/main/java/bisq/core/trade/protocol/TradeProtocol.java
@@ -108,12 +108,12 @@ public abstract class TradeProtocol implements DecryptedDirectMessageListener, D
 
     @Override
     public void onDirectMessage(DecryptedMessageWithPubKey decryptedMessageWithPubKey, NodeAddress peer) {
-        if (!isPubKeyValid(decryptedMessageWithPubKey)) {
+        NetworkEnvelope networkEnvelope = decryptedMessageWithPubKey.getNetworkEnvelope();
+        if (!isMyMessage(networkEnvelope)) {
             return;
         }
 
-        NetworkEnvelope networkEnvelope = decryptedMessageWithPubKey.getNetworkEnvelope();
-        if (!isMyMessage(networkEnvelope)) {
+        if (!isPubKeyValid(decryptedMessageWithPubKey)) {
             return;
         }
 


ripcurlx (approved):
ACK - Tested it on Regtest. 👍

huey735:
There seems to be a bug in the following scenario:
1. The mediator suggests that BTC buyer get the trade amount.
2. The BTC buyer accepts it.
3. The BTC seller doesn't and re-opens the ticket
4. The mediator suggests a new resolution in which the BTC seller gets the trade amount.
5. The BTC seller accepts this one.
6. The BTC buyer doesn't get a chance to reject it as it's automatically accepted and the trade is closed.

ripcurlx:
> There seems to be a bug in the following scenario:
> 
> 1. The mediator suggests that BTC buyer get the trade amount.
> 2. The BTC buyer accepts it.
> 3. The BTC seller doesn't and re-opens the ticket
> 4. The mediator suggests a new resolution in which the BTC seller gets the trade amount.
> 5. The BTC seller accepts this one.
> 6. The BTC buyer doesn't get a chance to reject it as it's automatically accepted and the trade is closed.

@huey735 I don't think this is related to this PR. Would be great to open another issue for this so it can be tracked and fixed.

PR #5202, opened 2021-02-18
Add api method 'stop'

ghubstan:
This change gives CLI users a `stop` command to shutdown a daemon.

- Server side gRPC boilerplate was added to` :proto` and `:daemon`.

- When the new `GrpcShutdownService` accepts a `stop` request from the CLI,
  it calls `UserThread.runAfter(BisqHeadlessApp.getShutDownHandler(), 500, MILLISECONDS);
`
- Method help file `stop-help.txt` was added to `:core`.

- Client side gRPC boilerplate and `stop` implementation was added to `:cli`.


ghubstan:
Force Travis build after failed jar download.

sqrrm (approved):
utACK

PR #5203, opened 2021-02-18
Add CLI testing bot to :apitest

ghubstan:
RobotBob reads a file `bot-script.json`, instructing it to make and take offers as per an `actions` json array, e.g. `["make","take","take","make",...]`, and the tester will manually run CLI commands provided by the bot during each step in a trade.

The test case (`ScriptedBotTest`) can be run with the test harness, which will start and shutdown all the regtest/dao apps: bitcoind, seednode, arbnode, bob & alice nodes.   Or, the test case can be run without the test harness, and the user manages his own daemons.


sqrrm (approved):
utACK

PR #5205, opened 2021-02-18
Fix multiple request preliminary data calls

### LINKED ISSUE #5204: Multiple call to requestPreliminaryData can cause startup problems
### Description

We call requestPreliminaryData on requestDataManager also from MailboxMessageService beside the P2PService.
This does not lead to a duplicated request as we check that we have only one request per address open, but we trigger the cleanup timer (2 minutes) which removes the handlers, so if the response takes longer than 2 minutes it will not be recognized.

chimp1984:
Fixes https://github.com/bisq-network/bisq/issues/5204

The PR is done in small improvements steps. Feel free to use final version instead of reviewing all commits.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/P2PService.java
+++ b/p2p/src/main/java/bisq/network/p2p/P2PService.java
@@ -330,6 +330,7 @@ public class P2PService implements SetupListener, MessageListener, ConnectionLis
 
     @Override
     public void onNoSeedNodeAvailable() {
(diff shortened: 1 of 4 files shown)


ripcurlx (approved):
ACK - Code looks fine IMO and I tested it on Mainnet re-producing the error and afterwards successfully starting with this fix. Actually I had exactly this issue the other day while having tor connections issue which I mentioned in the #ops channel. Should have investigated into this immediately myself. Thanks for the fix!

PR #5215, opened 2021-02-21
Fix Multiple Warnings on Enable All Offers

### LINKED ISSUE #5214: Multiple Warnings on Enabling Offers
### Description

Multiple warnings on enabling offers

#### Version

v1.5.6

### Steps to reproduce

If the overall toggle switch for enabling all offers is toggled too early after startup, then the warning window about waiting for the application to be fully connected, is displayed for each offer (three times in this case).

### Expected behaviour

The dialog is only shown once.

### Actual behaviour

It is shown multiple times.

### Screenshots
enabling

warning

ghost:
Display 1 warning not multiple when not bootstrapped

Fixes #5214


ripcurlx (approved):
utACK

PR #5216, opened 2021-02-22
Rename summary notes "Seller not responding"  category

huey735:
I renamed the string for the
disputeSummaryWindow.reason.SELLER_NOT_RESPONDING key because
BTC buyers can also be unresponsive.


ripcurlx (approved):
utACK

PR #5222, opened 2021-02-23
Fix key shortcut documentation when using CMD+Y

ripcurlx:
This behavior changed as you cannot manually fail an open trade anymore. Only unfail it after selecting it in the failed trades section and pressing CMD+Y


sqrrm (approved):
utACK

PR #5223, opened 2021-02-23
Fix length check text on manual payout tool

BtcContributor:
Fix the test displayed on the manual payout check tool.

Before it used everywhere the text _The length must be between X and Y_.

bisq

P.S.
This is my first PR, so let me know if everything is ok.


ghost (changes requested):
I disagree that adding a third variable is appropriate.
Can do the same functionality by checking `minLength==maxLength`.

ghost (dismissed):
:+1:  ACK

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/5223#pullrequestreview-596489357

PR #5224, opened 2021-02-23
Increase min withdrawal tx fee to 15 sats/vB

ghost:
Due to mempool congestion, increased from 2 to 15

Requested by @wiz.

### DIFF
--- a/common/src/main/java/bisq/common/config/BaseCurrencyNetwork.java
+++ b/common/src/main/java/bisq/common/config/BaseCurrencyNetwork.java
@@ -73,6 +73,6 @@ public enum BaseCurrencyNetwork {
     }
 
     public long getDefaultMinFeePerVbyte() {
-        return 2;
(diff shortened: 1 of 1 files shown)


wiz:
Thanks for this emergency fix to prevent wallet corruption. To be clear, we need a proper solution that uses the dynamic `mempoolminfee`, but that will require the following steps:

* mempool.space/api needs to add the `mempoolminfee` value to the current fee estimation endpoint
* bisq pricenodes need to store and serve the `mempoolminfee` value on its `/getAllMarketInfo` endpoint
* bisq nodes need to dynamically adjust the value you're changing in this PR to `mempoolminfee`

ripcurlx (approved):
utACK - I'll cherry-pick it into v1.5.7

ghost:
Note for people in the future: this was subsequently replaced by https://github.com/bisq-network/bisq/pull/5235

PR #5232, opened 2021-02-24
Fix dark mode readibility of specific texts

### LINKED ISSUE #5230: Fix color of message status information in Dark mode
It is hardly readable in dark mode right now if message is saved in receivers's mailbox.
Bildschirmfoto 2021-02-24 um 10 21 11

Same issue exists for other info messages.

Bildschirmfoto 2021-02-24 um 10 24 18

BtcContributor:
Fixes #5230

Fixed the original issue case and an another similar one.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/shared/ChatView.java
+++ b/desktop/src/main/java/bisq/desktop/main/shared/ChatView.java
@@ -488,12 +488,10 @@ public class ChatView extends AnchorPane {
                             visible = true;
                             icon = AwesomeIcon.OK;
                             text = Res.get("support.arrived");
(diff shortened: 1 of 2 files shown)


ripcurlx (changes requested):
NACK - I think the color and opacity for light mode is fine. It just needs a different color in css for dark mode to make it more readable.

BtcContributor:
Sorry @ripcurlx, now I get it.
CSS changed, now it should be correct.

bisq

ripcurlx (approved):

ripcurlx:
I think it doesn't have any effect on other UI faint focus outlines as we use the jfoenix library. But let's keep an eye open during next testing release cycle.

PR #5234, opened 2021-02-24
Fix BankId leak

### LINKED ISSUE #5201: Bank id leaking
Hey guys 👋 
One of the strongest point of Bisq is to keep users’ info private and to show them just to the users who are involved in the trade.
However, if you pick randomly one buy btc SEPA offer and expand its details, you can see the bank id near the description of the payment method. Example: SEPA (XXXX)
This potentially lets everyone put the id online and find the bank where the creditor will get the payment.

Wouldn’t be better for privacy purposes to hide the bank id from the payment details and to show it just when you accept the offer?

BtcContributor:
Fixes #5201

Now BankId is no more displayed when a user requests details on the offer book.


wallclockbuilder (dismissed):
ACK

desktop/src/main/java/bisq/desktop/main/overlays/windows/OfferDetailsWindow.java
```diff
@@ -253,21 +253,20 @@ private void addContent() {
             addConfirmationLabelLabel(gridPane, ++rowIndex, Res.get("offerDetailsWindow.myTradingAccount"), myPaymentAccount.getAccountName());
         } else {
             final String method = Res.get(paymentMethod.getId());
-            String methodWithBankId = method + bankId;
             String paymentMethodLabel = Res.get("shared.paymentMethod");
             if (countryCode != null && (isNationalBanks || isSpecificBanks || isSepa)) {
```
ripcurlx: For specific banks payment method we do want to show the bank name and/or id as it would be otherwise for someone to match ones offer using this payment method.

desktop/src/main/java/bisq/desktop/main/overlays/windows/OfferDetailsWindow.java
```diff
@@ -253,21 +253,20 @@ private void addContent() {
...
                 if (BankUtil.isBankIdRequired(countryCode))
                     addConfirmationLabelLabel(gridPane, ++rowIndex,
-                            paymentMethodLabel + " " + Res.get("offerDetailsWindow.offererBankId"),
```
ripcurlx, line 261: If you remove a translations please also remove it from `displayStrings.properties`. Otherwise we'll have more and more unnecessary strings to be translated. But in this case you probably need them anyways for the specific banks payment method

ripcurlx (changes requested):
NACK - please see my comments.

ripcurlx:
@pazza83 @BtcContributor This will remove the bank id and name from the offer details window, but of course for a sophisticated adversary this still can be accessed by code.

ripcurlx (approved):
utACK

PR #5235, opened 2021-02-25
Obtain minimumFee from mempool api in place of hardcoded value

### LINKED ISSUE #5229: Bisq does not check mempoolminfee when creating new transactions
Following up from discussions on Keybase, Bisq needs to enforce `mempoolminfee` value, since it currently uses a hard-coded value and the network value is variable:

1) Pricenode needs to save the `minimumFee` value from https://mempool.space/api/v1/fees/recommended
2) Pricenode needs to serve the `minimumFee` value from (1) on https://price.bisq.wiz.biz/getFees
3) Bisq needs to save the `minimumFee` value it gets when updating from (2) once per minute
4) Bisq needs to return the `minimumFee` value it gets from (3) in `bisq.common.config.BaseCurrencyNetwork::getDefaultMinFeePerVbyte()`

cc @jmacxx

ghost:
Fixes #5229

**example data from mempool api:**

    {"fastestFee":35,"halfHourFee":28,"hourFee":25,"minimumFee":6}

**Bisq pricenode obtains fee info:**

    Feb-24 22:10:45.291 [Timer-25] INFO  b.p.m.p.MempoolFeeRateProvider$First: Retrieved estimated mining fee of 28 sat/vB and minimumFee of 12 sat/vB from mempool.space 
    Feb-24 22:10:45.500 [Timer-25] INFO  b.p.m.p.MempoolFeeRateProvider$Fourth: Retrieved estimated mining fee of 27 sat/vB and minimumFee of 12 sat/vB from mempool.bisq.services 
    Feb-24 22:10:46.767 [Timer-25] INFO  b.p.m.p.MempoolFeeRateProvider$Second: Retrieved estimated mining fee of 27 sat/vB and minimumFee of 12 sat/vB from mempool.emzy.de 
    Feb-24 22:10:47.307 [Timer-25] INFO  b.p.m.p.MempoolFeeRateProvider$Third: Retrieved estimated mining fee of 27 sat/vB and minimumFee of 12 sat/vB from mempool.ninja 

---

**averaged fee data published by PriceNode:**

    {
      "bitcoinFeesTs" : 1614226269,
      "dataMap" : {
        "btcTxFee" : 27,
        "btcMinTxFee" : 12
      }
    }

---

**Bisq app obtains fee from PriceNode:**

    INFO  b.core.provider.fee.FeeService: BTC tx fee: txFeePerVbyte=27 minFeePerVbyte=12 

---

**Bisq withdrawal fee rate in settings screen cannot be set below the minFeePerVbyte**

image

### DIFF
--- a/core/src/main/java/bisq/core/provider/fee/FeeService.java
+++ b/core/src/main/java/bisq/core/provider/fee/FeeService.java
@@ -98,6 +98,7 @@ public class FeeService {
     private Map<String, Long> timeStampMap;
     @Getter
     private long lastRequest;
+    @Getter
     private long minFeePerVByte;
     private long epochInSecondAtLastRequest;
 
@@ -157,14 +158,15 @@ public class FeeService {
                         epochInSecondAtLastRequest = timeStampMap.get("bitcoinFeesTs");
                         final Map<String, Long> map = result.second;
                         txFeePerVbyte = map.get("BTC");
+                        minFeePerVByte = map.get("btcMinTxFee");
 
                         if (txFeePerVbyte < minFeePerVByte) {
-                            log.warn("The delivered fee per vbyte is smaller than the min. default fee of 5 sat/vbyte");
+                            log.warn("The delivered fee of {} sat/vbyte is smaller than the min. default fee of {} sat/vbyte", txFeePerVbyte, minFeePerVByte);
                             txFeePerVbyte = minFeePerVByte;
                         }
 
                         feeUpdateCounter.set(feeUpdateCounter.get() + 1);
-                        log.info("BTC tx fee: txFeePerVbyte={}", txFeePerVbyte);
+                        log.info("BTC tx fee: txFeePerVbyte={} minFeePerVbyte={}", txFeePerVbyte, minFeePerVByte);
                         if (resultHandler != null)
                             resultHandler.run();
                     });
--- a/core/src/main/java/bisq/core/provider/fee/FeeService.java
+++ b/core/src/main/java/bisq/core/provider/fee/FeeService.java
@@ -155,10 +155,10 @@ public class FeeService {
                     UserThread.execute(() -> {
                         checkNotNull(result, "Result must not be null at getFees");
                         timeStampMap = result.first;
-                        epochInSecondAtLastRequest = timeStampMap.get("bitcoinFeesTs");
+                        epochInSecondAtLastRequest = timeStampMap.get(Config.BTC_FEES_TS);
                         final Map<String, Long> map = result.second;
-                        txFeePerVbyte = map.get("BTC");
-                        minFeePerVByte = map.get("btcMinTxFee");
+                        txFeePerVbyte = map.get(Config.BTC_TX_FEE);
+                        minFeePerVByte = map.get(Config.BTC_MIN_TX_FEE);
 
                         if (txFeePerVbyte < minFeePerVByte) {
                             log.warn("The delivered fee of {} sat/vbyte is smaller than the min. default fee of {} sat/vbyte", txFeePerVbyte, minFeePerVByte);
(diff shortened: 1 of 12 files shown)


wiz:
Beautiful

ripcurlx:
@jmacxx Tests are broken right now. You need to adapt the Preferences constructor in the tests.

ghost:
fixed, thanks @ripcurlx! :+1:

core/src/main/java/bisq/core/user/Preferences.java
```diff
@@ -907,7 +911,7 @@ public boolean getPayFeeInBtc() {
 
     public long getWithdrawalTxFeeInVbytes() {
         return Math.max(prefPayload.getWithdrawalTxFeeInVbytes(),
-                Config.baseCurrencyNetwork().getDefaultMinFeePerVbyte());
```
ghubstan, line 910: Will `BaseCurrencyNetwork.getDefaultMinFeePerVbyte()` be `@Deprecated` and/or removed?

core/src/main/java/bisq/core/provider/fee/FeeService.java
```diff
@@ -157,14 +158,15 @@ public void onSuccess(@Nullable Tuple2<Map<String, Long>, Map<String, Long>> res
                         epochInSecondAtLastRequest = timeStampMap.get("bitcoinFeesTs");
                         final Map<String, Long> map = result.second;
                         txFeePerVbyte = map.get("BTC");
+                        minFeePerVByte = map.get("btcMinTxFee");
```
ghubstan: Just because this map key `btcMinTxFee` is used in at least one other class, should it be defined as a static field in one class to avoid typos?   I'm not sure, but it's something to consider...

ghubstan (commented):
I'm done with the review, but hold back on an ACK because I want to know if the old `BaseCurrencyNetwork#getDefaultMinFeePerVbyte` method can be removed if never to be used again.  (I'm chasing these min fee rate changes in the api+tests).

core/src/main/java/bisq/core/user/Preferences.java
ghost: We still use `BaseCurrencyNetwork.getDefaultMinFeePerVbyte()` in the following cases: (a) when Bisq has started up before the first API call to PriceNode completes, (b) if the PriceNode is unreachable, and (c) if PriceNode does not provide the new field in the JSON.

core/src/main/java/bisq/core/provider/fee/FeeService.java
ghost: Yes great idea, in fact I did run into a scenario like that during testing.  So I will extract those hardcoded strings shared between Bisq and PriceNode into `Common/Config.java`.  Thanks.

core/src/main/java/bisq/core/user/Preferences.java
ghubstan: I was too lazy to check myself, so thanks for that. (resolved)

ghost:
> I'm done with the review, but hold back on an ACK because I want to know if the old

Review issues have been commented on and/or resolved.  :+1:

ripcurlx (approved):
ACK - tested it on Regtest. I only get the default value set as long as it is not delivered by https://price.bisq.wiz.biz/ on regtest. Still want to wait for feedback from @wiz on the exact deployment procedure and ops state before this gets merged.

wiz (approved):
ACK - running PR branch on https://price.bisq.wiz.biz/getFees and new `minimumFee` field correctly matches 2x the minimum fee displayed on mempool.space

ripcurlx:
Tested it and a couple of edge cases against https://price.bisq.wiz.biz/. Everything seems to behave fine. The only UX drawback is, that we are manipulating the used fee if the new min fee is above the custom set value.

wiz:
> we are manipulating the used fee if the new min fee is above the custom set value.

that's exactly what we want tho, for bisq to create valid transactions that will be accepted to the mempool, and not immediately corrupt the user's wallet heh - maybe later @jmacxx can improve the send page to have better fee display

ripcurlx:
It is used again as soon as the minimum fee is below the custom fee again. As it prevents the problem of corrupted wallets which we see every now and then I'll create a hotfix branch based on v1.5.7.

PR #5238, opened 2021-02-25
Adjust api to new minimum fee per vbyte

ghubstan:
Since PR https://github.com/bisq-network/bisq/pull/5224, the `BaseCurrencyNetwork#getDefaultMinFeePerVbyte`  returns 15 (sats/byte).  See commit b341bb6e891c9a6f8ebb9ac9a94c919e0ce18d74.

This change adjusts the api to the new min tx fee rate by validating the api's setTxFeeRatePreference param, and throwing an appropriate  exception if the param value is below the minimum.  Also adjusted a broken test, and added a new test to check the  exception.

**TODO** Prepare to adjust api again to a different minimum fee per vbyte.  Commit c33ac1b9834fb9f7f14e553d09776f94efc9d13d changed the source of the min tx fee rate (`FeeService`), and the api will adjust after the associated PR https://github.com/bisq-network/bisq/pull/5235 is merged.  For now, the required change is a comment.


sqrrm (approved):
utACK

PR #5239, opened 2021-02-25
Add step for signature creation for Arch Linux

ripcurlx:


docs/release-process.md
```diff
@@ -164,6 +164,15 @@ If all was successful:
  Check the checkbox for update, set the version number (e.g. 0.9.4) and add the short version of the release notes.
  * After sending the Update message leave it running for about 1 minute to give time for good propagation.
  * Make a backup of that alert sender app data directory
+ * To support source code signature verification for Arch Linux download `Source code (tar.gz)`, sign it and
+ upload signature.
+```
+    # sign source code bundle
+    gpg --digest-algo SHA256 --local-user $BISQ_GPG_USER --output bisq-${NEW-VERSION}.tar.gz.asc --detach-sig --armor bisq-${NEW-VERSION}.tar.gz
```
wallclockbuilder: Looks good. Only one thing.
The BASH specification does not allow the dash character in environment variable names. You probably want to replace that with an underscore like you already did in `$BISQ_GPG_USER`.

When I try it as it is using the dash in the name, setting the env var 
`$ NEW-VERSION="v9.9.9"` 
produces error
`-bash: NEW-VERSION=v9.9.9: command not found`
So when I ran the command you provided command 
`$ gpg --digest-algo SHA256 --local-user "${BISQ_GPG_USER}" --output bisq-${NEW-VERSION}.tar.gz.asc --detach-sig --armor bisq-${NEW-VERSION}.tar.gz`
 I get the error
`gpg: can't open 'bisq-VERSION.tar.gz': No such file or directory
gpg: signing failed: No such file or directory`
And when I check for the file output 
`$ ls bisq-v9.9.9.tar.gz.asc || echo $`
It does not exist 
`ls: bisq-v9.9.9.tar.gz.asc: No such file or directory` 
No surprise there.

When I set the env var with underscore replacing the dash 
`$ NEW_VERSION="v9.9.9"`
then run the command using the underscore env var version 
`$ gpg --digest-algo SHA256 --local-user "${BISQ_GPG_USER}" --output bisq-${NEW_VERSION}.tar.gz.asc --detach-sig --armor bisq-${NEW_VERSION}.tar.gz` 
and then check for the output file
`ls bisq-v9.9.9.tar.gz.asc || echo $`
I get the file without error 
`bisq-v9.9.9.tar.gz.asc`

docs/release-process.md
```diff
@@ -164,6 +164,15 @@ If all was successful:
...
+
+    # verify signature of source code bundle
+    gpg --digest-algo SHA256 --verify bisq-${NEW-VERSION}.tar.gz{.asc*,}
```
wallclockbuilder: Same applies here as in my previous comment. Its a good idea to replace the dash in the env var name with an underscore.

wallclockbuilder (changes requested):
LGTM.
Will approve after changes are effected.

wallclockbuilder (changes requested):
NACK

ripcurlx:
Yes, I had the ${NEW-VERSION} just as a placeholder in mind, but it is true that it is confusing as is. I'll adapt the naming so it works properly when using it as an env variable.

wallclockbuilder (approved):
ACK

sqrrm (approved):

PR #5240, opened 2021-02-25
Refactor grpc stub boilerplate from CliMain to GrpcClient

ghubstan:
`CliMain` no longer uses `gRPC` stubs directly, just parses & validates command options, and prints responses.

Also, anticipates reuse by java bots (avoiding grpc boilerplate duplication on the client side), reduces size of `CliMain` class, and will allow clean up of duplicated grpc boilerplate in apitest cases.

PR https://github.com/bisq-network/bisq/pull/5238 should be reviewed/merged before this one.


sqrrm (approved):
utACK

PR #5244, opened 2021-02-26
Refactor apitest cases to use GrpcClient

ghubstan:
This is a follow up to PR https://github.com/bisq-network/bisq/pull/5240, which moved client side `gRPC` boilerplate to a new, reusable GrpcClient class.

From this change forward, all client side `gRPC` request/response boilerplate (including any used by tests) should now live in `GrpcClient`.

PR https://github.com/bisq-network/bisq/pull/5240 should be reviewed/merged before this one.


sqrrm (approved):
utACK

PR #5249, opened 2021-02-27
Improve gRPC server error logging

ghubstan:
- Use the logger of the `gRPC` service throwing the exception, not the exception handler's logger.

- Log `trade not found` exceptions as warnings instead cluttering the log with full stack traces.


ghubstan:
Snuck in call rate metering test bug fixes with commits 98ff6cf, e8d1f03 and f90d2ce.

sqrrm (approved):
utACK

PR #5250, opened 2021-02-28
Fix call rate metering interceptor bug

ghubstan:
The gRPC interceptor was not using the correct method/ratemeter map key, and failing to find a rate meter for the server call.

- Fix the rate meter key lookup bug.

- Disable almost all strict, default call rate meters in tests.
  Made an adjustment to the test harness' scaffold setup so rate meter testing or disabling config file is always picked up by bob and alice daemons (not arb daemons because there is no dao-setup for that node).

- Set arbitration daemon's `registerDisputeAgent` rate = 10/second. 
  To not interfere with the test harness.  (There is no pre-existing arb node appDataDir before that daemon starts.)

PR https://github.com/bisq-network/bisq/pull/5249 has to be reviewed/merged before this one.


ghubstan:
Mainnet bats tests are working now.

wallclockbuilder (approved):
UtAck

sqrrm (approved):
utACK

PR #5256, opened 2021-02-28
Fix prompt text overflow in create order

### LINKED ISSUE #5189: Trigger price truncation of prompt text not working
I just recognized in the German translation that the truncation of the new trigger price feature is not configured properly. Also the provided text is not properly translated (I'll take care of this on Transifex myself).

Bildschirmfoto 2021-02-12 um 11 02 44

forcryptos:
Added StringUtils class with truncatePromptText method.

truncatePromptText is called in MutableOfferView(5) and
in MutableOfferViewModel(1) to cut too long text.

Fixes #5189

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/BisqDefaultCoinSelector.java
+++ b/core/src/main/java/bisq/core/btc/wallet/BisqDefaultCoinSelector.java
@@ -33,7 +33,10 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 
+import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
+
+import javax.annotation.Nullable;
 
 /**
  * Used from org.bitcoinj.wallet.DefaultCoinSelector but added selectOutput method and changed static methods to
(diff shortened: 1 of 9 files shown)


wallclockbuilder (approved):
UtAck

ripcurlx:
I think this should be handled differently.  I'm quite sure the auto-truncation of the prompt text is a supported feature, but it just doesn't work because `input.setMaxWidth(Double.MAX_VALUE);` is set in `getEditableValueBox`. I didn't have time to look into it in more detail, but I think this is the way to fix this issue.

forcryptos:
Thank you @ripcurlx !

Will work on it asap.

forcryptos:
@ripcurlx I've tried playing with setMaxWidth with not success, it seems to be only offsetting all the content (label, prompt and tooltip). There is not many methods related to prompts. Maybe it would be in the css. 

I do not know much about javaFx yet. I think the best way forward is to let someone more experienced look into it when he has the time, and if there is no easy fix, I can maybe simply set a hard limit in the getEditableValueBox.

PR #5272, opened 2021-03-03
Fix Set Password text button

### LINKED ISSUE #3379: password is not saving
Password is not saving!

is saving - only when you click "I have already done a buckup" not other buttons.

bisq_LI

BtcContributor:
Fixes #3379

Now it looks like this:

bisq10

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -3155,6 +3155,7 @@ password.forgotPassword=Forgot password?
 password.backupReminder=Please note that when setting a wallet password all automatically created backups from the unencrypted wallet will be deleted.\n\n\
   It is highly recommended that you make a backup of the application directory and write down your seed words before setting a password!
 password.backupWasDone=I have already made a backup
+password.setPassword=Set the password
 
 seed.seedWords=Wallet seed words
 seed.enterSeedWords=Enter wallet seed words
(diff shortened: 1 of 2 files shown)


wallclockbuilder (dismissed):
UtAck

ripcurlx (changes requested):
NACK - It is kind of critical that the user understands what the setting of the password will cause. We know that users hardly read text so we want to prevent the user making a mistake. Maybe the text for the call to action should be more something like `Set password (I have already made a backup)` 👈  @m52go any ideas on a perfect condensed wording?

BtcContributor:
I didn't want to fill a button with too much text.

Maybe some part of the explanation text could be in bold?

We could also change from "Go to Account/Wallet seed" to "Make a backup".

wallclockbuilder:
Other wallets e.g. BlueWallet, CoboVault deal with this by requiring the user to confirm the backup seed before setting the password. This way you know for sure they backed up the seed and also got the correct seed.

wallclockbuilder:
For backups bisq can also verify. Ask user for backup location. Confirm backup directory signature matches current backup directory signature. Both signatures should match as long as both directories have the same files and structure.

m52go:
> Maybe some part of the explanation text could be in bold?

I don't think that's possible, unfortunately.

> We could also change from "Go to Account/Wallet seed" to "Make a backup".

I like this suggestion. So maybe the green button can be "Set Password (I already made a backup)" and the other can be "Make Backup" which goes to `Account` > `Backup`. If it fits...

BtcContributor:
Pushed changes as stated by @m52go, thanks!

ripcurlx (approved):

PR #5292, opened 2021-03-10
Fix singleton CoreContext usage

ghubstan:
After `GrpcServer` builds a gRPC server instance, it calls injected `CoreContext#setApiUser(true)`.  The same, singleton `CoreContext` instance is injected into CoreService constructors, which use the `coreContext#isApiUser` value (set true by the server builder) to set their own class level boolean `isApiUser` fields.

When `CoreContext` was added, I was fooled by the correct behavior I saw while stepping through the server debugger.  But now I see that the correct boolean value in `CoreContext` is only available after CoreService constructors have finished instantiating themselves -- not in time.

In this change, the `isApiUser` field is removed from CoreServices, `CoreContext coreContext` becomes a class level final field in CoreServices, and the correct `coreContex.isApiUser` value is available in the CoreService methods.

This PR is the first in a new API PR chain.


sqrrm (approved):
utACK

PR #5293, opened 2021-03-10
Add NGN as a default currency

sqrrm:
This adds NGN as a default currency, as suggested in https://github.com/bisq-network/growth/issues/156#issuecomment-794871823


wallclockbuilder (approved):
LGTM.

utACK.

ripcurlx (approved):
utACK

PR #5294, opened 2021-03-10
Improve takeoffer error handing

ghubstan:
UI specific task runner validation errors cannot be passed to the client via a typical `ErrorMessageHandler`.  This change uses a new `GrpcErrorMessageHandler` implementation that places an optional `AvailabilityResult` in the `takeoffer` response if an offer is not available.  The client can derive a better failure reason message from the reply's AvailabilityResult.

- `GrpcErrorMessageHandler`  Add new `ErrorMessageHandler` implementation to get around the api specific problem of having to use an interface that is designed to build task error messages for the UI.
  
- `GrpcExceptionHandler`  Add a method for working with the  `ErrorMessageHandler` interface.

- `GrpcTradesService, CoreApi, CoreTradesService`:  Adjust the `takeoffer` error handling to give a failure reason provided by the new `GrpcErrorMessageHandler`. 

- `pb.proto`  Add `PRICE_CHECK_FAILED` to enum `AvailabilityResult`.

- `grpc.proto`  Add enum `AvailabilityResult` to `TakeOfferReply` message.  This field is populated by the `GrpcErrorMessageHandler` when `takeoffer` fails, making it easier to give CLI users a stylistically consistent failure reason.

- `OpenOfferManager`  Inject `CoreContex` to know isApiUser.  Set AvailabilityResult = `PRICE_CHECK_FAILED` if isApiUser=true, leaving it as `UNKNOWN_FAILURE` for isApiUser=false.

- OpenOfferManagerTest  Adjusted to new `CoreContext` constructor arg.  
  
- `GrpcClient`  Adjusted `takeOffer` to convert `AvailabilityResult` to user error message if the call fails -- does not return a trade.
  
- Added grpc exception catch blocks missing from some api test cases in `apitest/src/test/java/bisq/apitest/method/trade`.


ghubstan:
PR https://github.com/bisq-network/bisq/pull/5292 should be reviewed and merged before this one.

core/src/main/java/bisq/core/offer/AvailabilityResult.java
```diff
@@ -23,11 +23,12 @@
...
     NO_MEDIATORS,
     USER_IGNORED,
-    MISSING_MANDATORY_CAPABILITY,
-    NO_REFUND_AGENTS,
+    @SuppressWarnings("unused") MISSING_MANDATORY_CAPABILITY,
+    @SuppressWarnings("unused") NO_REFUND_AGENTS,
     UNCONF_TX_LIMIT_HIT,
-    MAKER_DENIED_API_USER
+    MAKER_DENIED_API_USER,
+    PRICE_CHECK_FAILED
```
sqrrm: Just to confirm, this would fail to parse by old clients and thus use the first element in the enum, which is `UNKNOWN_FAILURE`, which makes sense.

wallclockbuilder (dismissed):
Tests passing.
LGTM.

ACK

core/src/main/java/bisq/core/offer/AvailabilityResult.java
ghubstan: I confirmed this by sending `protobuf.AvailabilityResult.PRICE_CHECK_FAILED(13)` to a client that does not  have that proto enum value, and the client received `UNKNOWN_ENUM_VALUE_AvailabilityResult_13`.

sqrrm (approved):
utACK

PR #5302, opened 2021-03-14
Avoid NPE due to null fee while wrapping tx in TxInfo

ghubstan:
Add debugging convenience to core api model's `TxInfo`.

PR https://github.com/bisq-network/bisq/pull/5294 should be reviewed/merged before this one.


sqrrm (approved):
utACK

PR #5316, opened 2021-03-15
Add pay from BSQ wallet button

chimp1984:
Add button to buyers trade step 2 view in case of a BSQ trade to fill in the amount and address to the BSQ send view.

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -671,6 +671,7 @@ portfolio.pending.step2_buyer.amountToTransfer=Amount to transfer
 portfolio.pending.step2_buyer.sellersAddress=Seller''s {0} address
 portfolio.pending.step2_buyer.buyerAccount=Your payment account to be used
 portfolio.pending.step2_buyer.paymentStarted=Payment started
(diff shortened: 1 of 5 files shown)


chimp1984:
Screenshot 2021-03-15 at 17 45 49
Screenshot 2021-03-15 at 17 45 54

ripcurlx:
To me `Pay from BSQ wallet` sounds like it would automatically pay the BSQ from the BSQ wallet. Maybe it should be labeled `Go to BSQ wallet`?

wallclockbuilder:
Maybe a tooltip? Right next to the button.
[ PAY FROM BSQ WALLET ] ( i ) Fills out the address and amount. You have to press send to pay.

ripcurlx (approved):
utACK - I missed that it pre-populates the input fields to send the BSQ, so it is not a simple go to wallet button. After thinking about it I think it is fine doing it like this, as even if a user thinks this will automatically send the BSQ will see that she is re-directed to the wallet for confirmation.

ripcurlx:
> Maybe a tooltip? Right next to the button.
> [ PAY FROM BSQ WALLET ] ( i ) Fills out the address and amount. You have to press send to pay.

Yes, maybe an info could improve this, but I don't think it is a showstopper for this PR.

PR #5318, opened 2021-03-16
Lower tolerated small amount

sqrrm:
Split tolerated amount for placing offers with unsigned accounts
and tolerated amount for taking offers.

Implements suggestion in https://github.com/bisq-network/proposals/issues/322

Split the small amount check between own side and peer side. Lower size that can be created but allows for taking already created buy offers with the old higher amount. The peer side check can later be lowered once the outstanding higher amount offers have been taken.

### DIFF
--- a/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
+++ b/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
@@ -414,14 +414,16 @@ public class AccountAgeWitnessService {
                                AccountAgeWitness accountAgeWitness,
                                AccountAge accountAgeCategory,
                                OfferPayload.Direction direction,
-                               PaymentMethod paymentMethod) {
+                               PaymentMethod paymentMethod,
+                               boolean isMyLimit) {
         if (CurrencyUtil.isCryptoCurrency(currencyCode) ||
                 !PaymentMethod.hasChargebackRisk(paymentMethod, currencyCode) ||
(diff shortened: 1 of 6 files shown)


ripcurlx (approved):
utACK - Code looks fine. Will take a look in more detail during backwards comp. tests during release testing.

sqrrm:
@ripcurlx The backwards compatibility is the main issue and def good that you test it.

pazza83:
Hi @sqrrm 

I think 0.0025 BTC is too low: https://github.com/bisq-network/proposals/issues/322#issuecomment-800733552

Is this commit putting in place the code to choose a selected amount or putting in place the 0.0025 BTC limit for new offers?

sqrrm:
@pazza83 This changes the max buy amount for unsigned accounts to 0.0025 BTC.

pazza83:
@sqrrm is the signing requirement to get signed still 0.0025 BTC?

RefundAgent:
This change done in this haphazard and non-democratic way will most likely be the final nail in the coffin of Bisq. Fortunately there are privacy-protecting alternatives these days.

The Bisq "team" has zero understanding of economics and probabilities. So long and thanks for all the fish.

chimp1984:
@sqrrm @ripcurlx 
I agree that this PR seems to got merged too quickly before consensus have been found. There have been clear concerns from @pazza83 in the related proposal. I would suggest to revert that commit until consensus is found.

PR #5319, opened 2021-03-16
Change file mentioned to from .desktop to .service

wallclockbuilder:
Bisq Monitor does not include a `.desktop` in the monitor folder.
It does have a `.service` file which is spoken about in the lines that follow the mention.
I think this was a brain fart. Author meant to introduce `.service` file, mentioned `.desktop` instead.

### DIFF
--- a/monitor/README.md
+++ b/monitor/README.md
@@ -1,6 +1,5 @@
 # Bisq Network Monitor Node
 
-
 The Bisq monitor node collects a set of metrics which are of interest to developers and users alike. These metrics are then made available through reporters.
 
 The *Settled* release features these metrics:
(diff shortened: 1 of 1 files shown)


monitor/README.md
```diff
@@ -1,5 +1,6 @@
 # Bisq Network Monitor Node
 
+
```
ripcurlx: Could you please remove this empty line? Thanks!

ripcurlx (changes requested):
NACK - Please see my comment.

monitor/README.md
wallclockbuilder: Done.

ripcurlx (approved):
utACK

PR #5320, opened 2021-03-16
Allow resend of fiat payment sent message in exceptional circumstance

ghost:
This is a proposed temporary solution the number 1 mediation issue of "Unable to confirm payment received".
tldr; Allow the Buyer able to re-send the "Fiat payment initiated" message by the following method.

**Current problem:**

- Buyer sends "Fiat payment initiated" message.
- Seller either does not receive it or somehow has a problem processing it (as was the case in support case id=YCWGPDVA)
- Buyer is unable to re-send the "Fiat payment initiated" message.
 

**Proposed solution:**

- Buyer sends "Fiat payment initiated" message.
- Seller either does not receive it or somehow has a problem processing it.
- Either party opens a support ticket, explains the problem (Seller cannot confirm, buyer cannot re-send "Fiat payment initiated")
- Mediator closes the ticket as "Bug", advising both to reject mediation, and the buyer to double-click on Trade Step 2 line item in the GUI.
- Buyer is then prompted to re-send the "Fiat payment initiated" message.
- Seller sees the subsequent "Fiat payment initiated" message and the trade can continue.

 

Internally, the code will check that if user double clicks on buyer step 2 (Start Payment), and if a dispute has been closed, then it will move the trade state back to BUYER_CONFIRMED_IN_UI_FIAT_PAYMENT_INITIATED.  This will then cause the existing workflow to show a button requesting that the trader resend the message.

Protection from misuse is added by only allowing this workflow after a dispute ticket has been closed.  It also allows mediators to observe if this solution really does address the problem.

---

After opening a support ticket, mediator will instruct buyer to double-click on step 2:

image

Then buyer is able to re-send the payment started message:

image

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/BuyerSubView.java
+++ b/desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/BuyerSubView.java
@@ -24,6 +24,7 @@ import bisq.desktop.main.portfolio.pendingtrades.steps.buyer.BuyerStep3View;
 import bisq.desktop.main.portfolio.pendingtrades.steps.buyer.BuyerStep4View;
 
 import bisq.core.locale.Res;
+import bisq.core.trade.Trade;
 
 import org.fxmisc.easybind.EasyBind;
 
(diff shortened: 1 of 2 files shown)


ghost:
Closing because this does not really help fix the problem.  We still need to find the reason that payment started message does not arrive at the seller, or that the seller does not advance to the next step.

PR #5323, opened 2021-03-17
Update commands for mediator/refund agent

wallclockbuilder:
Commands are flipped in the documentation.
CMD+D for Mediator
CMD+N for Refund Agent


ripcurlx (approved):
ACK

Issue #3305, opened 2019-09-22
Deposit transaction ID - Transaction not found


bitsanity:
re: https://bisq.community/t/deposit-transaction-id-transaction-not-found/8424

Quoting huey:

> There’s an issue that has bee causing an error when broadcasting transactions. In this case, it seems that the make offer transaction wasn’t successfully broadcast but the Bisq Network still accepted it as a valid offer.

Details:

Trade is stuck on Step 1 “Wait for blockchain confirmation”. The Deposit and Maker fee transaction IDs shown are not found by block explorer(s).

Version: Bisq-64bit-1.1.6.deb
Trade ID: efeikv2-bfc01e8c-6541-49ed-8c7d-ff1f17c21877-115
Deposit txid: 4e96b45716582fc18b2ab8aabceaaa5c450412329644858625ed4966dcf84ab6
Maker txid: c10e3ecf7a516b687afefcbf86d00c26497f84dcc7640d7a7d5d998a5d40d179
Taker txid: 4df815d6a4b6250d965a40dfa7118e023b74e0f1dc1663401932c05ae2f5665e

Only the Taker txid is recognized by block explorers.

huey735:
@sqrrm @ripcurlx 
This seems like an exceptional error of the trade protocol. The offer shouldn't have made to the offer book without a confirmation of the makeOfferTx.

Related to https://github.com/bisq-network/bisq/issues/3306

bitsanity:
Something still broken - transactions are not making it to the blockchain.

I see the dispute payout transaction from earlier is still not found on blockchain:
https://www.blockchain.com/btc/tx/8391d4799bffdd9d80349e5d4324eac5b45fa77179ac340bcd8cdb36f2f34cfc

Without realizing, I tried to send bitcoin from my Bisq wallet and it failed to appear in the mempool in time.

https://www.blockchain.com/btc/tx/2642b5e520bd407d2385bf6a60f8d7492a0ad76d5333959a73ba215bde390b8a

My log entries include:
fee 0.0000815 BTC for 813 bytes (10 Satoshi/Byte)
which looks reasonable (about $0.81 USD)

I see the onion stuff and peer messaging going on. There is an entry:

Sep-22 14:02:25.383 [JavaFX Application Thread] INFO o.b.c.TransactionBroadcast: broadcastTransaction: We have 9 peers, adding 2642b5e520bd407d2385bf6a60f8d7492a0ad76d5333959a73ba215bde390b8a to the memory pool

[...]

PR #5325, opened 2021-03-17
Prevent trading when not connected to the Bitcoin network

ghost:
It is currently possible to create and take offers while not connected to the bitcoin network.

This should have been fixed by https://github.com/bisq-network/bisq/pull/4764 but that PR only considered the scenario where chainheight from the DAO was stale.  

The fix is to also handle the other part of the equation, when the **chainheight of bitcoin network peers is stale**.  

---

Instances of this problem have been experienced with offers that have transactions not found by blockexplorers, as described in https://github.com/bisq-network/bisq/issues/5151 and https://github.com/bisq-network/bisq/issues/3305.  

---

**To reproduce the issue:**

- Start Bisq, specifying a bitcoin node address that is not running a node.  (For regtest see `DEFAULT_REGTEST_HOST`)
- Bisq will start up and display "Connecting to Bitcoin network" in the footer bar.
- Create offers, take offers.  No error messages are displayed.

**To verify the fix:**

- Start Bisq, specifying a bitcoin node address that is not running a node.  (For regtest see `DEFAULT_REGTEST_HOST`)
- Bisq will start up and display "Connecting to Bitcoin network" in the footer bar.
- Attempt to create offers, take offers.  You will be informed as that Bisq is not in sync with the network.

image

Network Settings:

image


ripcurlx (approved):
utACK

PR #5326, opened 2021-03-18
Lower min deposit to 0.001 BTC

sqrrm:
Lower min refund to half of min deposit as it doesn't make sense to have
a higher min refund than the min deposit.

Lower min deposit to 0.001 BTC to be more reasonable in fiat terms. Would be good to change to use a fiat feed to determine this value in the future.

Implements suggestion in https://github.com/bisq-network/proposals/issues/323

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/Restrictions.java
+++ b/core/src/main/java/bisq/core/btc/wallet/Restrictions.java
@@ -69,7 +69,7 @@ public class Restrictions {
     // So 0.0005 BTC is the min. buyer security deposit even with amount of 0.0001 BTC and 0.05% percentage value.
     public static Coin getMinBuyerSecurityDepositAsCoin() {
         if (MIN_BUYER_SECURITY_DEPOSIT == null)
-            MIN_BUYER_SECURITY_DEPOSIT = Coin.parseCoin("0.006"); // 0.006 BTC about 42 USD @ 7000 USD/BTC
+            MIN_BUYER_SECURITY_DEPOSIT = Coin.parseCoin("0.001"); // 0.001 BTC is 60 USD @ 60000 USD/BTC
         return MIN_BUYER_SECURITY_DEPOSIT;
     }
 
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
utACK - I'll test it in-depth during backwards comp. release tests.

MwithM:
Mediators always have to leave 0.003BTC for one of the parties. Could you test that it won't be an issue?

HeadyWook:
Has a scale been considered before? For example, if a constant price in fiat of $31.25 is desired then:

USD-----------------------------%BTC
$100,000,000    *    0.00003125%
$50,000,000    *    0.0000625%
$10,000,000    *    0.0003125%
$5,000,000    *    0.000625%
$1,000,000    *    0.003125%
$500,000    *    0.00625%
$100,000    *    0.03125%
$50,000    *    0.0625%
$10,000    *    0.315%
$5,000    *    0.625%
$1,000    *    3.125%
$500    *    6.25%
$100    *    31.25%
$50    *    62.5%
$10    *    312.5%
$5    *   625%
$1    *    3125%

What if there were some sort of scale used to determine minimum security deposit based on bitcoin price in USD (for this example) rather than again make a fixed minimum security deposit like before (i.g., 0.006)? If price goes up or down significantly, then another change will be needed if a static minimum amount is set with no scale, as proposed here with the 0.001 BTC. If security deposits were based on a scale, it would adjust as price changes. If the value of the security deposit changes significantly then the trade can be placed on hold until more collateral is deposited (price goes up) or returned (price goes down).

HeadyWook:
Also, what is to happen to the smaller trades on Bisq with the current security deposit of 0.006 BTC?

For example, there is currently a trade up for 0.001 BTC but requires a security deposit of 0.006 BTC (600%!)

Perhaps the difference should be returned, allowed to edit, or moved to make another trade?

ripcurlx:
> Mediators always have to leave 0.003BTC for one of the parties. Could you test that it won't be an issue?

I'll test it and post my results here.

BtcContributor:
> 
> 
> Mediators always have to leave 0.003BTC for one of the parties. Could you test that it won't be an issue?

This has been adapted as well.
Now it is always 0.0005 btc.

https://github.com/bisq-network/bisq/pull/5326/files#diff-283bca2cb1b273b6c2f9a208191605540c51bfdca2d222dceacf82bc3b68f9aaR90

PR #5330, opened 2021-03-18
Disambiguate payment acct json form's "country" field

ghubstan:
Make it clear the user needs to enter a two letter country code in the api's payment account json form's `country` field, not a country name.  A blank payment account json form now has a  `country` default value = `your two letter country code`.

The json form's field name was not changed to countryCode for the following reason:

The API's payment account forms are dynamically generated using reflection, and `PaymentAccount` class hierarchy `@Setter` methods determine which form fields are included or excluded.  Reflection and `@Setter` methods are also used when reading completed json forms, to set the field values on a new PaymentAccount instance before it is persisted.

`CountryBasedPaymentAccount` subclasses like `F2FAccount` have a `country` field and a `@Setter`, not a countryCode field, hence this shortcut to informing the user the country field value is a country-code.


ripcurlx (approved):
utACK

PR #5332, opened 2021-03-19
Fix dark mode readibility of specific texts - Part 2

BtcContributor:
This PR complements https://github.com/bisq-network/bisq/pull/5232.

We also need to remove opacity for the remaining 2 cases so to not have different looks.
Indeed we have no opacity set for some messages and a setOpacity(0.5) for other ones.

CSS is the same for everything:
`.status-icon {
    -fx-text-fill: -fx-faint-focus-color;
}`

Before:
bisq_read_1

bisq_read_2

After:

text3

text4

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/shared/ChatView.java
+++ b/desktop/src/main/java/bisq/desktop/main/shared/ChatView.java
@@ -477,7 +477,6 @@ public class ChatView extends AnchorPane {
                             visible = true;
                             icon = AwesomeIcon.OK_SIGN;
                             text = Res.get("support.acknowledged");
-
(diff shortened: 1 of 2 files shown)


ripcurlx (dismissed):
utACK

ripcurlx:
https://github.com/bisq-network/bisq/pull/5332/commits/7e8c9db499ac7461dcdd3193bf4450aa6a5b9fe8 is not signed.

wallclockbuilder (dismissed):
ACK

BtcContributor:
@ripcurlx signing fixed

ripcurlx (approved):
utACK

PR #5336, opened 2021-03-20
[WIP] Maybe init wallet in 1st unlockwallet call (api)

ghubstan:
Unlike the UI, an API daemon cannot capture the user's wallet encryption password at startup, to init an encrypted wallet and update balances.  This change addresses the problem in the 1st call to the api's `unlockwallet` method.


ghubstan:
Commit 8a3a281 resolves the problem for the BTC wallet, but the BSQ wallet init bug is not fixed yet.

PR #5337, opened 2021-03-20
Fix API startup's wallet.init failure when wallet is encrypted

ghubstan:
This bug fix makes sure encrypted wallet initialization happens during startup, without prompting the user for a wallet password.

- Block UI popups in api startup's encrypted wallet init.
Running the UI's `walletPasswordHandler` during wallet initialization will block the API's wallet initialization.  This change ensures an encrypted wallet does not trigger `walletPasswordHandler.run()` if `isApiUser=true` during startup.

- Set `WalletsManager` key in first `CoreWalletsService#unlockWallet` call.
This is necessary because the key cannot be set during api daemon startup, which does happen during the UI's startup.  See commit eb15fda229bd9481b11844ee95f2243613842034.


sqrrm (approved):
utACK

Issue #5329, opened 2021-03-18
BSQ trading fee calculated incorrectly


ghost:
Some offers pay an incorrect BSQ fee.

Here is an example of the discrepancies from today's offer book.  Some have underpaid, others have overpaid.

![image](https://user-images.githubusercontent.com/47253594/111669793-26042b00-87e5-11eb-932a-d0bd9b04c51d.png)

The fee rate is chosen based on the block height at the time the offer is created.  For example, an offer created at block height 673810 would use the fee rate of 7.53 BSQ per bitcoin.

![image](https://user-images.githubusercontent.com/47253594/111671426-ed655100-87e6-11eb-9c4a-f0b7a3d53d68.png)

A small but significant subset of offers have not chosen the correct fee rate.  How can this happen?

Looking at the highlighted onion addresses, it seems that when a node chooses an incorrect rate, it is consistent in choosing the same incorrect fee rate for subsequent offers.

Attn: @chimp1984 @ripcurlx 
[feeDiscrepancies_20210318.xlsx](https://github.com/bisq-network/bisq/files/6166243/feeDiscrepancies_20210318.xlsx) [TxFeeRawData_20210318.txt](https://github.com/bisq-network/bisq/files/6166278/TxFeeRawData_20210318.txt)

chimp1984:
It can be that the user has not a synced DAO state. There are quite some nodes with incorrect BSQ state hash (see DAO monitor in app). We should add some popup to enforce those users to resync DAO state.
You can try to click take offer for those users, then you are connected and at next block you likely get their DAO state and it shows up in the monitor if they have an invalid hash. That would a way how to detect if that is the reason (no guarantee though that u get their hash).

One way to handle that would be like we do for fee receiver address. If the value does not match the expected value we iterate over all past dao param values and if one of those matches we consider it valid. That covers the non-in-sync cases.
As it always can happen that a user gets out of sync, I assume there is no other way around.

pazza83:
> It can be that the user has not a synced DAO state

How does a user know if their DAO state is correctly synced?

![dao](https://user-images.githubusercontent.com/66367547/111857048-c6159d80-8926-11eb-8521-cb66460af4ac.PNG)

I see this most times, are the seed nodes the same as the DAO state?

chimp1984:
If your node is in sync with the seed nodes all is ok.

[...]

PR #5347, opened 2021-03-23
Fee validation leniency using old DAO param values

ghost:
Some Bisq clients are stuck on an invalid DAO state and are paying fees on an old schedule.

Per https://github.com/bisq-network/bisq/issues/5329#issuecomment-803223859 : a paid fee will be considered valid if it can be derived from any of the DAO parameter changelist.  

After implementing this change, the OfferBook discrepancies listed in #5329 are reduced to zero.

---

The essence of the leniency check is this:

image


ripcurlx (approved):
utACK - Code looks fine.

Note for testers:
I guess the easiest/most realistic testing scenario would be to create offers with current fee on Regtest and increase it afterwards through Regtest DAO voting and check if this offer still can be taken afterwards.

PR #5348, opened 2021-03-23
Add issue template config

huey735:
I added an issue template config file to better guide
the user as to where their issue may be more fitting.


ripcurlx (approved):
utACK

PR #5349, opened 2021-03-24
Review changes for Disputes GUI

ghost:
- highlight trade period end using an appropriate CSS alert style
- system messages to show in chat unread message count

---
https://github.com/bisq-network/bisq/pull/5160#issuecomment-801161625

> Better would be e.g. to create a new class .alert without underline and hand, so it doesn't look like a link.

Done.

> And one more thing I missed in the other PR and feels a little bit weird for the user. If a dispute is closed you get the notification popup. When you click on it, it opens the support section and the issue is selected. But it is greyed out and users might not know what to do. Maybe it is better to open in this case automatically the same action as a double click would trigger. WDYT?

I tried this and did not like it.   There are already too many things that pop up, and having the first closed dispute pop open its chat view when entering support is very counter-intuative.  NACK

> Also it would be good to show a (1) badge over the chat icon for the system message, so people are looking into it. In the past the chat was open by default in-line. So at least a notification would be good.

Done.

### DIFF
--- a/core/src/main/java/bisq/core/support/dispute/Dispute.java
+++ b/core/src/main/java/bisq/core/support/dispute/Dispute.java
@@ -409,8 +409,7 @@ public final class Dispute implements NetworkPayload, PersistablePayload {
 
     public long unreadMessageCount(boolean senderFlag) {
         return chatMessages.stream()
-                .filter(m -> m.isSenderIsTrader() == senderFlag)
-                .filter(m -> !m.isSystemMessage())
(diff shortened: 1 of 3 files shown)


ripcurlx (approved):
utACK - Probably having the badge icon now also after a dispute is closed is enough indication that users actually click on the chat icon.

BtcContributor:
Tested ACK

As an improvement I would suggest to highlight the chat icon and badge when there are new messages to be read but the dispute is already closed.

bisq_chat

Edit:
See latest commit aea043e.

PR #5354, opened 2021-03-25
Remove the apparent link for cases where the trade is not available for mediators

BtcContributor:
This is how it appears now the TradeID column for mediators:

bisq_mediator

Before it appeared as an hyperlink that was not actually working.

Note that the trade is only known to the 2 people involved in the trade. 
It is not shared with the Mediator/Arbitrator and for this reason a mediator is expected to not be able to pull trade details as it happens for the peers involved.

EDIT:
WAIT for the merge. Me and @jmacxx are investigating some display issues.

EDIT2:
Now it could be merged.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeView.java
+++ b/desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeView.java
@@ -1139,14 +1139,14 @@ public abstract class DisputeView extends ActivatableView<VBox, Void> {
                                 super.updateItem(item, empty);
 
                                 if (item != null && !empty) {
-                                    field = new HyperlinkWithIcon(item.getShortTradeId());
                                     Optional<Trade> tradeOptional = tradeManager.getTradeById(item.getTradeId());
                                     if (tradeOptional.isPresent()) {
+                                        field = new HyperlinkWithIcon(item.getShortTradeId());
                                         field.setMouseTransparent(false);
                                         field.setTooltip(new Tooltip(Res.get("tooltip.openPopupForDetails")));
                                         field.setOnAction(event -> tradeDetailsWindow.show(tradeOptional.get()));
                                     } else {
-                                        field.setMouseTransparent(true);
+                                        setText(item.getShortTradeId());
                                     }
                                     setGraphic(field);
                                 } else {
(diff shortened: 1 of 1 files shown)


wallclockbuilder (approved):
ACK

Before:
trade id before
Trade ID(Has pseudo hyperlink)

After:
trade id after
TradeID(No more has pseudo hyperlink)

desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeView.java
```diff
@@ -1139,14 +1139,14 @@ public void updateItem(final Dispute item, boolean empty) {
...
                                     Optional<Trade> tradeOptional = tradeManager.getTradeById(item.getTradeId());
                                     if (tradeOptional.isPresent()) {
+                                        field = new HyperlinkWithIcon(item.getShortTradeId());
                                         field.setMouseTransparent(false);
                                         field.setTooltip(new Tooltip(Res.get("tooltip.openPopupForDetails")));
                                         field.setOnAction(event -> tradeDetailsWindow.show(tradeOptional.get()));
                                     } else {
-                                        field.setMouseTransparent(true);
+                                        setText(item.getShortTradeId());
                                     }
```
wallclockbuilder, line 1150: This is all you need.

`Optional<Trade> tradeOptional = tradeManager.getTradeById(item.getTradeId());  `

`if ((item != null && !empty) && tradeOptional.isPresent()) {`

 `  setText(item.getShortTradeId());`

`} else {`

 `  setText("");`

`}`

Every other line related to the hyperlink can be deleted(1135, 1151 - 1155). 
Might want to use getDateColumn() above(1097) for reference. Its also a text with no hyperlink.

wallclockbuilder (changes requested):
Codewise there is still dead code from the hyperlink.
Once thats cleaned up its a go.

desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeView.java
ripcurlx: The hyperlink is needed for the DisputeClientView that is extending DisputeView. So if a trade is available it should link to the trade details popup.

ripcurlx (approved):
utACK - Code looks fine.

PR #5356, opened 2021-03-25
Add api method createcryptopaymentacct

ghubstan:
This change supports creation of `BSQ` `BLOCKCHAIN` payment method accounts.

- Added proto message defs to `grpc.proto`.
- Added grpc server boilerplate to `GrpcPaymentAccountsService`.
- Added server impl to `CoreApi`, `CorePaymentAccountsService`.
- Added `createcryptopaymentacct-help.txt`.
- Added CLI side support for new api method.
- Added opt parsing unit tests to `OptionParsersTest`.

This is the 1st PR in a series, with the goal of supporting the BTC/BSQ trading pair.  Support for other crypto currency payment accounts will be added later.

PR #5357, opened 2021-03-25
Provide more offer & contract detail to CLI

ghubstan:
This change adds offer and trade contract detail to the API's Offer and Trade protos, and improves CLI output formatting.

- Appended missing fields to `OfferInfo` proto message:
	uint64 sellerSecurityDeposit = 20;
	string offerFeePaymentTxId = 21;
	uint64 txFee = 22;
	uint64 makerFee = 23;

- Added new api proto messages `ContractInfo` and `PaymentAccountPayloadInfo`.
	Lighterweight protos are needed because core Trade/Contract classes are not visible to CLI.

- Appended `ContractInfo` field to api proto message `TradeInfo`.

- Added proto / model converters for `ContractInfo` and `PaymentAccountPayloadInfo`, and adjusted `OfferInfo` & `TradeInfo`.

- Improved CLI output formatting.  Added more trade detail to CLI's `gettrade` output, and prepared to support BTC/BSQ trading pair.  _Note a reviewer is advised to look at the CLI outout formatting class files instead getting bogged down in the many commit changes._

This is the 2nd PR in a series, with the goal of supporting the BTC/BSQ trading pair.
PR https://github.com/bisq-network/bisq/pull/5356 must be reviewed/merge before this one.


cli/src/main/java/bisq/cli/TradeFormat.java
```diff
@@ -181,4 +190,32 @@ private static String formatTradeData(String format,
...
+            boolean isBuyerMakerAndSellerTaker = contract.getIsBuyerMakerAndSellerTaker();
+            return isBuyerMakerAndSellerTaker  // (is BTC buyer / maker)
+                    ? contract.getTakerPaymentAccountPayload().getAddress()
+                    : contract.getMakerPaymentAccountPayload().getAddress();
+        } else {
+            return "";
+        }
+    };
+
+    private static boolean shouldShowBqsBuyerAddress(TradeInfo tradeInfo, boolean isTaker) {
```
sqrrm: Spelling `...Bqs...`

core/src/main/java/bisq/core/api/CoreWalletsService.java
```diff
@@ -305,6 +312,41 @@ void sendBtc(String address,
         }
     }
 
+    boolean verifyBsqSentToAddress(String address, String amount) {
```
sqrrm, line 315: There are some issues with this as two trades could have the same amounts, or someone sends the BSQ as two transactions adding up to the total amount. I think this is good enough for now, but it's not fool proof. The problem is the same when trading manually though, just something to keep in mind.

cli/src/main/java/bisq/cli/TradeFormat.java
ghubstan: Fixed in commit bddc273bb27ba73f93ed9e9cea4defa598c9f28c.

core/src/main/java/bisq/core/api/CoreWalletsService.java
ghubstan: I'm trying to be careful...  The mm bot I just removed from https://github.com/bisq-network/bisq/pull/5359 created up to some max bsq payment accounts, to use a different rcv address for each buy-bsq trade.  After max accts have been created, the bot randomly chooses from the existing accts for each trade cycle.

That bot also randomly varies each trade's amt and price by some max deviation size (in sats), to help reduce the risk of re-using the same verifyBsqSentToAddress(address, amt) arguments, even if the same payment acct was used.

sqrrm (approved):
utACK

PR #5370, opened 2021-03-28
Disputes UI improvements

### LINKED ISSUE #5367: The 1.6.0 mediation style is confusing
Hello. I would like to point out that the new mediation style introduced in 1.6.0 is very confusing. Now the chat with the mediator is opened in a separate window, just like chat with the peer. They look the same, so it is very easy to send message to the peer instead of the mediator and the other way around, without even noticing it which may result in unnecessary delays. The previous style, where chat with the mediator had no separate window was immune to this confusion. Thank you.

ghost:
Fixes #5367

- Adds a border for dispute chat.
- Window title displays "`Dispute chat window for trade....`"
- Each chat message indicates who it is from, e.g. `Received [timestamp] From Mediator`

See screenshots.

Screenshot from 2021-03-28 15-05-01

---

Screenshot from 2021-03-28 15-04-24


wallclockbuilder (dismissed):
utACK

desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeChatPopup.java
```diff
@@ -96,7 +96,7 @@ public void openChat(Dispute selectedDispute, DisputeSession concreteDisputeSess
         AnchorPane.setRightAnchor(chatView, 10d);
         AnchorPane.setTopAnchor(chatView, -20d);
         AnchorPane.setBottomAnchor(chatView, 10d);
-
+        pane.setStyle("-fx-background-color: -bs-color-blue-5");
```
ripcurlx: Didn't it work to use a class in that case?

ghost: Thanks for the suggestion, done! :+1:

ghost:
@ripcurlx: I added https://github.com/bisq-network/bisq/pull/5370/commits/23535de4185b3e87654dae0ddd79b81b3fe253b7 to fix something for @refund-agent2; closes the chat window when its associated ticket is closed.

ghost:
Added https://github.com/bisq-network/bisq/pull/5370/commits/a8209591be9c20ac59763f343c5f2eab5e2a1b24 per https://github.com/bisq-network/bisq/discussions/5392

Support agents want the process & chat buttons on the left.  For users it makes sense for the chat button to be on the right, consistent with how it is shown in TradeView.  Hopefully that makes everyone happy! :grinning:

ripcurlx:
@pedromvpg What do you think about the new blue border for distinguishing trader chat from dispute chat?

ripcurlx (approved):
utACK - Code changes are looking fine. If @pedromvpg comes up with some design changes I think we can do it in a follow-up PR.

PR #5372, opened 2021-03-29
Update gradle-witness comment

cd2357:
Specify the algorithm used to calculate and verify the checksums.


cd2357:
I confirmed this by locally generating the SHA-256 of a jar in `bisq/lib` and comparing it with the checksum in `gradle-witness.gradle`.

ghubstan (commented):
ACK

ripcurlx (approved):
utACK

Issue #5366, opened 2021-03-28
Getting java.lang.NumberFormatException and blank/unresponsive Bisq window


m52go:
### Description

Since upgrading my Bisq instance to 1.5.9 about 10 days ago, I've gotten this error about 4-5 times:

`Mar-26 20:03:25.028 [JavaFX Application Thread] ERROR b.c.u.MathUtils: java.lang.NumberFormatException: Character I is neither a decimal digit number, decimal point, nor "e" notation exponential mark.`

When this happens, Bisq becomes unresponsive. Usually the Bisq window becomes a blank white rectangle, but sometimes the Bisq GUI remains visible (but not operational).

Also when this happens, I have to kill Bisq and then open it again -- closing Bisq properly doesn't work. So far there haven't been any errors opening properly afterward.

### Version

1.5.9

### Steps to reproduce

I got this error a couple of times last year, but then it went away with an update so I didn't report it.

But I also didn't report it because there's not very much help I can give on reproducing it......

The error seems to occur randomly, and when it occurs, all logs are overwritten so quickly that any clues of the bug's cause are quickly overwritten:

![bisq-error](https://user-images.githubusercontent.com/735155/112739079-bd454d00-8f60-11eb-8793-f1159f2a3153.gif)

Every time this error has happened so far, I wasn't actively using Bisq. Sometimes I was actively using the machine it was running on (but not using Bisq), and sometimes I wasn't using the machine at all (screen locked but machine still on).

Usually I'm tipped off that the error has occurred when I hear the computer fans going crazy.

### Expected behaviour

N/A

### Actual behaviour

N/A

### Screenshots

See above GIF for an idea of what logs look like streaming in real-time.

As noted above, the Bisq window usually turns into a blank white rectangle.

### Device or machine

Ubuntu 18.04

### Additional info

Can provide logs upon request, but these logs will only show logs _after_ this error occurs (i.e., startup after killing the program).

wallclockbuilder:
Update to v1.6.0 its coming out on Monday(in 48 hrs). `https://github.com/bisq-network/bisq/tree/release/v1.6.0`

cd2357:
@m52go can you look back in the logs at the first time it occurs, and copy paste the stack trace if available?

That would help us better pin-point what is causing it, thanks.

m52go:
> can you look back in the logs at the first time it occurs

[...]

Issue #5362, opened 2021-03-26
GUI not functioning.


RandomAccountt:
### Description

When opening Bisq, the GUI does not start.

#### Version

1.5.9

### Steps to reproduce

1. Open Bisq.

### Expected behaviour

GUI opens.

### Actual behaviour

GUI does not open, but the application exists in the background. Uses resources as normal. The app appears open in the taskbar, but clicking on it does not bring up the GUI.

### Screenshots
This appears in the windows key+tab menu.
![win-tab](https://user-images.githubusercontent.com/30158779/112684808-7de90480-8e6b-11eb-91c8-d51fac366452.png)
#### Device or machine

ASUS TUF FX505DT
Running latest version of Windows 10.

#### Additional info

Issue persists after reinstall. Additionally, I would like to know if there is a way to export my wallet to electrum or any other wallet. I do not have the wallet seed, but I have my backup files. Thanks

wallclockbuilder:
Fastest way around this is to use a VM(e.g. Ubuntu on virtualbox).

cd2357:
Is this the first time you run Bisq? Or did you run it before and it ran fine, but it started behaving like this after a Bisq upgrade / OS upgrade / etc ?

One thing you can try is to disable the use of Hardware Graphics Acceleration in Bisq by doing the following:

1. Using notepad open the file `Bisq.cfg` located in `C:\Users\<your username>\AppData\Local\Bisq\app`
2. Under the section `[JVMOptions]` add the following `-Dprism.order=sw` such that it now looks like this:
```
[JVMOptions]
-Dprism.order=sw
```
3. Close/save the file and start Bisq again

---

Reference threads:
* https://www.reddit.com/r/javahelp/comments/84w6i6/problem_displaying_anything_with_javafx_only/
* https://stackoverflow.com/questions/18754803/how-to-disable-or-bypass-hardware-graphics-accelerationprism-in-javafx

cd2357:
Might be related to #5366 because there

> "the Bisq window usually turns into a blank white rectangle"

RandomAccountt:
> Is this the first time you run Bisq? Or did you run it before and it ran fine, but it started behaving like this after a Bisq upgrade / OS upgrade / etc ?

Nope, it happened after a while using the app. Definitely not from a bisq upgrade, perhaps from a Windows update but I cannot remember.

[...]

PR #5374, opened 2021-03-29
Handle rounding of infinite Doubles

cd2357:
In certain situations we deal with infinite (or NaN) doubles. This causes issues when trying to instantiate a BigDecimal from it, which is an intermediary step in the process of rounding the double.

This commit checks for such non-finite doubles and throws a specific exception if their rounding is attempted.

Related to #5366, #5362


ripcurlx (approved):
utACK

PR #5380, opened 2021-03-31
Use semaphore for SPV resync

### LINKED ISSUE #5335: SPV file is deleted after SPV resync
### Description

User can't complete SPV resync, after the process and restarting it keeps displaying a message that spv file has been removed.

### Version

Bisq v1.5.9 ; Arch linux 5.11.2

### Steps to reproduce

Unknown.
### Expected behaviour

SPV resync completes.

### Actual behaviour

The "SPV chain has been deleted" message appears after SPV resync restart.

### Screenshots
paste (2)
paste

### Additional info
@jmacxx has logs from this case and has investigated a little. The issue was solved by renaming-remove PreferencesPayload file as "there is a flag stored in user preferences that controls if the resync is initiated at Bisq startup".

ghost:
Prevents an infinite SPV resync loop due to preferences not persisting if the app did not start.

Fixes #5335

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqSetup.java
+++ b/core/src/main/java/bisq/core/app/BisqSetup.java
@@ -103,6 +103,7 @@ import javax.annotation.Nullable;
 @Singleton
 public class BisqSetup {
     private static final String VERSION_FILE_NAME = "version";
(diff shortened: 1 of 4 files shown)


paulmercier-bisq (approved):
utACK, looks good! 👍

sqrrm (approved):
ACK

PR #5397, opened 2021-04-06
Add BTC node wubwzaadboxwiffa.onion

devinbileck:
Adding this node to help alleviate the increased network load.

### DIFF
--- a/core/src/main/java/bisq/core/btc/nodes/BtcNodes.java
+++ b/core/src/main/java/bisq/core/btc/nodes/BtcNodes.java
@@ -70,6 +70,7 @@ public class BtcNodes {
                         // Devin Bileck
                         new BtcNode("btc1.bisq.services", "lva54pnbq2nsmjyr.onion", "172.105.21.216", BtcNode.DEFAULT_PORT, "@devinbileck"),
                         new BtcNode("btc2.bisq.services", "qxjrxmhyqp5vy5hj.onion", "173.255.240.205", BtcNode.DEFAULT_PORT, "@devinbileck"),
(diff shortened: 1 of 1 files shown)


core/src/main/java/bisq/core/btc/nodes/BtcNodes.java
```diff
@@ -70,6 +70,7 @@
                         // Devin Bileck
                         new BtcNode("btc1.bisq.services", "lva54pnbq2nsmjyr.onion", "172.105.21.216", BtcNode.DEFAULT_PORT, "@devinbileck"),
                         new BtcNode("btc2.bisq.services", "qxjrxmhyqp5vy5hj.onion", "173.255.240.205", BtcNode.DEFAULT_PORT, "@devinbileck"),
+                        new BtcNode(null, "wubwzaadboxwiffa.onion", null, BtcNode.DEFAULT_PORT, "@devinbileck"),
```
ghubstan, line 73: I notice the hostname and ip-address arguments are null.   Does that matter?

devinbileck: The clearnet host/address is only used for people running regtest or if someone is intentionally running on clearnet only.
This node does not have a static IP so I omitted them.

ghubstan: My mistake.  I should have figured that out.

ripcurlx (approved):
utACK

ripcurlx:
@devinbileck Somehow I'm not able to re-trigger the Travis build for Java12. I always get `Oh no....` Not sure what is the problem. Maybe you could try to force push to trigger it with that?

devinbileck:
@ripcurlx It passed now.

PR #5401, opened 2021-04-09
Wait for AckMessage before completing BuyerSendCounterCurrencyTransferStartedMessage task

ghost:
Fixes `seller-cant-confirm` issue by transitioning past `BuyerSendCounterCurrencyTransferStartedMessage` only after an `AckMessage` response.  Previously it would transition upon mailbox message delivery.

Scenario from log of issue `aTPPev` (and others) : 

- Alice creates an offer to buy.  Bob takes the offer.  Blockchain confirms.
- Alice initiates fiat payment and clicks on Payment sent.
- 12 seconds later Alice clicks Payment sent.
- 11 seconds later Alice clicks Payment sent.
- 12 seconds later Alice clicks Payment sent.
- 9 seconds later Alice log receives "Message arrived at peer" (4 times).
- Alice transitions to the next step and no longer has the option to click Payment sent.
- [10 hours passes], Ack message has **not** been received from Bob.
- Bob is still waiting for `BuyerSendCounterCurrencyTransferStartedMessage`, so he can not click on "Payment received".
- Alice and Bob exchange chat messages and eventually open a Dispute.

In the scenario above, Alice would keep being prompted to "send confirmation again" (until she receives AckMessage) and eventually DM with Bob to ask what the issue is.  

- Bob might say he has **not** received payment notification, he could restart Bisq and maybe the message would go through,
- or Bob might have already transitioned to showing the "Confirm Payment Receipt" button, and therefore could pay out the trade.
- The trade can be paid out normally even if Alice has the Payment started button showing.  As long as Bob processes a `BuyerSendCounterCurrencyTransferStartedMessage` => can click "Confirm Payment Receipt".
- This fix gives more chance for `BuyerSendCounterCurrencyTransferStartedMessage` to be re-sent.

image

tag @chimp1984 pls :eyes:

### DIFF
--- a/core/src/main/java/bisq/core/trade/protocol/tasks/buyer/BuyerSendCounterCurrencyTransferStartedMessage.java
+++ b/core/src/main/java/bisq/core/trade/protocol/tasks/buyer/BuyerSendCounterCurrencyTransferStartedMessage.java
@@ -92,11 +92,10 @@ public class BuyerSendCounterCurrencyTransferStartedMessage extends SendMailboxM
 
     @Override
     protected void setStateArrived() {
-        trade.setStateIfValidTransitionTo(Trade.State.BUYER_SAW_ARRIVED_FIAT_PAYMENT_INITIATED_MSG);
-
-        processModel.getTradeManager().requestPersistence();
-        cleanup();
-        // Complete is called in base class
+        // the message has arrived but we're ultimately waiting for an AckMessage response
+        if (!trade.isPayoutPublished()) {
+            tryToSendAgainLater();
+        }
     }
 
     // We override the default behaviour for onStoredInMailbox and do not call complete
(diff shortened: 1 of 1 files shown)


chimp1984 (approved):
utACK

chimp1984:
Ah great that tackle that issue! Your suggested solution looks good to me.

Seems there are some edge cases where the message arrives but the peer does not process the message (maybe just shuts down app or some edge case bugs). Maybe related with hibernate behaviour at some OS. I saw sometimes offers I could never take because the peer did not respond (in time), so that looked like the network throttling behaviour for apps in hibernate state. Unfortunately hibernate strategies getting more and more complex and instransparant with each new OS version, at least on OSX and Windows will be worse for sure...

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/5401#pullrequestreview-632635019

PR #5402, opened 2021-04-09
Tweak wording for takeOffer.fundsBox.isOfferAvailable

maxim-belkin:
Minor wording tweak. This message is automatically displayed before the buyer can take the offer, so it makes sense to say {that Bisq is} checking if the offer is still available.


ripcurlx:
@m52go ☝️ Could you please review this one? Thanks!

m52go:
Makes sense :+1:

utACK

ripcurlx (approved):
utACK

maxim-belkin:
Thank you both for approving and merging!

PR #5404, opened 2021-04-10
.github/workflows/build.yml: new GitHub Action

maxim-belkin:
A GitHub action that builds Bisq on Ubuntu 18.04.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,33 @@
+name: Build Bisq
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-18.04
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+        ref: ${{ github.event.pull_request.head.sha }}
+        lfs: true
+
+    - name: Set up JDK
+      uses: actions/setup-java@v2
+      with:
+        java-version: '11'
+        distribution: 'adopt'
+
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
+
+    - name: Validate Gradle Wrapper JAR files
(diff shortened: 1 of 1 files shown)


ripcurlx:
> A GitHub action that builds Bisq on Ubuntu 18.04.

Is there an additional benefit doing it like that compared to what we have configured for Travis right now?

maxim-belkin:
> Is there an additional benefit doing it like that compared to what we have configured for Travis right now?

* Travis has a limit of 10,000 credits that doesn't reset/renew ([link](https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works)), so the switch is (more or less) imminent.
* The submitted action has no advantages over what Travis CI is currently set up to do. The benefits of using GH Actions become more obvious when the CI/CD system is a bit more complex.

I can tweak the action to make it a bit more useful (test scripts in the /scripts directory, test different OSes,...) -- just let me know what you'd like to automate.

ripcurlx:
Right now we support Java 10 (for releases) and Java 12 (development). That is why we added both versions. As soon as #4242 gets merged Java 11 will be the sole support version. Does master build for you with Java 11 right now if you run the action on your forked repo?

Regarding the move from travis-ci.org to travis-ci.com: Yes there might be a problem if they don't allow unlimited jobs for OS projects. I think the only thing that is missing on this  action is the Asset check that is in place right now https://github.com/bisq-network/bisq/blob/64fe6c18a4f4bebe85ed0b8a5cd09d640d8d923d/.travis.yml#L15-L16

ripcurlx:
As this use of GitHub actions doesn't introduce any risks I'll merge it as it might be a good alternative if we have to move away from Travis soon.

ripcurlx (approved):
utACK

maxim-belkin:
Thanks for merging, @ripcurlx!

PR #5405, opened 2021-04-11
Update scripts/install_java.sh script

maxim-belkin:
1. Determine package manager without using Bash arrays (also add `dnf` package manager)
2. Use the new way to install Homebrew:
   ```
   /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
   ```
3. Don't call `brew update` -- Homebrew calls it implicitly.
4. Evaluate `brew shellenv` to get access to the `brew` command
5. Re-org commands to follow the download-unpack-install procedure
6. Remove `| sudo bash`


ripcurlx:
@devinbileck Could you please review this PR? Thanks!

maxim-belkin:
Marked this as a WIP for now because I'm not sure I understand why one needs to install cURL on macOS -- AFAIK macOS-es come with one (mine does) and, moreover, Homebrew installation script literally checks if cURL is installed before proceeding
https://github.com/Homebrew/install/blob/84f7c2e7ea0a026fdff8b4563492b6ac06e9b799/install.sh#L324-L330
So, depending on the reasons behind `brew install curl`, it might have to be removed.

devinbileck:
I can't speak for installing curl on macOS, but I know that was explicitly added for Linux as some distros do not come with it by default, so perhaps it was just carried over to the macOS section.
If that is the case where homebrew explicitly checks for it, then should be fine to remove it.

devinbileck:
And by "remove it", I mean the entire homebrew install since it appears to only have been used to install curl.
Potentially we could add a similar check and abort if curl is not already installed.

scripts/install_java_linux.sh
```diff
@@ -0,0 +1,143 @@
+#!/bin/bash
+
+# Install OpenJDK 11.0.2 on Linux
+
+set -eu
+
+# Download and install locations
+JDK_URL=https://download.java.net/java/GA/jdk11/9/GPL/openjdk-11.0.2_linux-x64_bin.tar.gz
```
ripcurlx, line 8: I think it would be good to start with the current version of Java (11.0.11). Do we want to update this script for each Java sub version update?

maxim-belkin: I don't see 11.0.11 at https://jdk.java.net/archive/, but I do see it on AdopOpenJDK website ([here][1], which links to [GitHub][2]) and [here][3]. Which one should I use? (I think the last link is the right one but would be good to confirm).

> Do we want to update this script for each Java sub version update?

We can move these links to another file(s) and make `install_java_*.sh` scripts read those file(s). There is a minor issue with Alpine Linux -- it uses [musl libc][musl], not glibc -- so the only way to install OpenJDK there I found was to use its package manager (`apk`).

[1]: https://adoptopenjdk.net/upstream.html?variant=openjdk11&ga=ga
[2]: https://github.com/AdoptOpenJDK/openjdk11-upstream-binaries/releases/
[3]: https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/tag/jdk-11.0.11%2B9_openj9-0.26.0
[musl]: https://musl.libc.org/

ripcurlx: Weird - Yes, maybe I had it on one VM from AdoptOpenJDK. Looking at my dev (macOS) machine I have openJDK installed via brew on 11.0.10. I just wanted to point out that we should include the latest java version available. No idea why the latest in the archive is only 11.0.2.

BtcContributor:
I want just to notice that a PR of mine that adapts scripts to JDK11 has recently been merged.

scripts/install_java_linux.sh
maxim-belkin: For macOS, we could use Homebrew and the [adoptopenjdk/openjdk](https://github.com/adoptopenjdk/homebrew-openjdk) tap.

maxim-belkin:
> I want just to notice that a PR of mine that adapts scripts to JDK11 has recently been merged.

Thanks for the note, @BtcContributor. I did see your commit and updated this PR accordingly.

scripts/install_java_linux.sh
ripcurlx: > For macOS, we could use Homebrew and the [adoptopenjdk/openjdk](https://github.com/adoptopenjdk/homebrew-openjdk) tap.

That is the setup I use personally as it is easy to use and upgrade multiple JDKs I can switch around with `jenv` when needed.

ripcurlx:
@wiz @Emzy Could one of the seednode operators ACK this new Java install script? Thanks!

Emzy:
utACK
Reviewed the scripts.

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/5405#issuecomment-831897871

PR #5423, opened 2021-04-20
Functionality to duplicate an offer

ghost:
Adds a right-click menu to Portfolio->Open Offers, Open Trades, and History lists allowing those offers to be duplicated.
The create offer screen is then presented with the filled-in details which can be edited if necessary.

### Screenshots

Screenshot from 2021-04-20 15-55-40

image

image

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -583,6 +583,9 @@ portfolio.tab.pendingTrades=Open trades
 portfolio.tab.history=History
 portfolio.tab.failed=Failed
 portfolio.tab.editOpenOffer=Edit offer
+portfolio.tab.duplicateOffer=Duplicate offer
+portfolio.context.offerLikeThis=New offer like this...
+portfolio.context.notYourOffer=You can only duplicate trades where you were the maker.
 
 portfolio.closedTrades.deviation.help=Percentage price deviation from market
 
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -584,8 +584,8 @@ portfolio.tab.history=History
 portfolio.tab.failed=Failed
 portfolio.tab.editOpenOffer=Edit offer
 portfolio.tab.duplicateOffer=Duplicate offer
-portfolio.context.offerLikeThis=New offer like this...
-portfolio.context.notYourOffer=You can only duplicate trades where you were the maker.
+portfolio.context.offerLikeThis=Create new offer like this...
+portfolio.context.notYourOffer=You can only duplicate offers where you were the maker.
 
 portfolio.closedTrades.deviation.help=Percentage price deviation from market
 
@@ -2948,9 +2948,9 @@ popup.shutDownInProgress.headline=Shut down in progress
 popup.shutDownInProgress.msg=Shutting down application can take a few seconds.\nPlease don't interrupt this process.
 
 popup.attention.forTradeWithId=Attention required for trade with ID {0}
-popup.attention.reasonForPaymentRuleChange=Version 1.5.5 introduces a critical trade rule change regarding \
-  the \"reason for payment\" field in bank transfers. Please leave this field empty -- \
-  DO NOT use the trade ID as \"reason for payment\" anymore.
+popup.attention.newFeatureDuplicateOffer=Version 1.6.3 introduces a new feature allowing easy re-entry of offers \
+  by right-clicking on an existing offer or trade and choosing `Create new offer like this`.  This is useful for \
+  traders who frequently make the same offer.
 
 popup.info.multiplePaymentAccounts.headline=Multiple payment accounts available
 popup.info.multiplePaymentAccounts.msg=You have multiple payment accounts available for this offer. Please make sure you've picked the right one.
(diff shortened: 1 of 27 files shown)


ripcurlx:
Code looks fine and it works functionality wise (tested it on Regtest). I only have two remarks:
- Wording: `New offer like this...` For me this sounds a little bit more like filter for offers like this than create offer like this -> @m52go Any ideas about the wording?
- Right click: This is quite a hidden feature. How do we want to communicate it best to the user so they will be actually aware of this? Maybe we could add at those places where the column space is sufficient a duplicate action icon? WDYT?

ripcurlx:
Something like this?
Artboard

ghost:
The thought of cluttering the row with yet another icon is painful, and doing that on `Open trades` view & `History` view more so.  Personally I think a context action is cleaner.  Power users have been pleading for the ability to reload offers.  We could announce the feature once at startup like the existing announcement in 1.5.5 about "reason for payment" change.  WDYT?

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -583,6 +583,9 @@ portfolio.tab.pendingTrades=Open trades
 portfolio.tab.history=History
 portfolio.tab.failed=Failed
 portfolio.tab.editOpenOffer=Edit offer
+portfolio.tab.duplicateOffer=Duplicate offer
+portfolio.context.offerLikeThis=New offer like this...
```
```suggestion
portfolio.context.offerLikeThis=Create new offer like this...
```

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -583,6 +583,9 @@ portfolio.tab.pendingTrades=Open trades
...
+portfolio.context.notYourOffer=You can only duplicate trades where you were the maker.
```
```suggestion
portfolio.context.notYourOffer=You can only duplicate offers where you were the maker.
```

m52go:
Add suggestions to strings above. I agree the "new offer like this" can sound ambiguous, but a verb at the beginning should clarify.

ghost:
image



image


@m52go :eyes: :pray:

ripcurlx (approved):
ACK - Tested it on Regtest

BtcContributor:
A couple of suggestions on this functionality:

1. I would explicitly tell that you can right click only within the *Portfolio* menu, at least for now. I read very quickly the popup at the beginning and I was a bit lost when I did not find the functionality on main Buy Btc / Sell Btc (btw, it could be interesting to add the functionality there as well). 
So I would write "Version 1.6.3 introduces a new feature allowing easy re-entry of offers by right-clicking on an existing offer or trade within Portfolio menu and choosing `Create new offer like this`.  This is useful for traders who frequently make the same offer"
2. The duplication seems to keep fixed just the btc value. I personally think most of the people sell/buy btc on fiat basis so it would be more important the viceversa. Or, best, we could add a new entry in the right-click menu, one for duplicate an offer maintaining the original btc value and another one to maintain the original fiat value.

RefundAgent:
Would it not be possible to have a duplicate offer that does exactly that - without any more clicks needed? One single click to publish an offer identical to the one selected. 

This is more like duplicate and modify, which forces the user to go through looking at and accepting the contract and so on which adds a lot of fully needless clicks.

Or a two-choice menu when right-clicking:
i) duplicate 
ii) duplicate and modify.

ghost:
1-click is possible; I worry about the potential risk though.  Normally anything that spends money should have some kind of confirmation in case it was accidental.  However, perhaps a 1-click offer feature as you want could be considered similar to Amazon's 1-click buy feature.  I would err on the cautious side and require that the user enable the "1-click offer feature" via a settings option.  It would assume that the local wallet be used for funding.  Then we can have two options in the right-click menu as you described, adding `Duplicate this offer (1-click)`.  
@ripcurlx what do you think?

RefundAgent:
I doubt people will accidentally click on duplicate offer and even if they do they can always retract the offer. A 1-click offer is indeed like Amazons 1-click buy which I appreciate so much and I don't think it should be opt-in. Amazon realized that a 1-click offer was so valuable that they tried to patent it. An opt-in once again forces people to click click click and try to find the correct place in the settings.

The present method should be used when making a new offer, i. e. the fields should be prefilled with the numbers from a previous offer, unless it is the very first offer. Most people (and everyone I know) repeat similar offers.

ripcurlx:
I'm not sure if we should duplicate the offer without any confirmation at all as the user might not be aware of the exact fees she is expected to pay at this very moment. What would be a good next step to reduce unnecessary clicks before duplicating it, could be to directly jump to the final offer confirmation window. WDYT?

RefundAgent:
To jump to a final confirmation window is the next best thing if a true 1-click is deemed risky. 
That is, if only one confirmation click is needed. 

Maybe a warning pop-up saying something like "fees are high right now" can be used instead when right-clicking the first time. The user can then continue or stop depending on mood and 1-click can still work.

RefundAgent:
At least four out of 17 selloffers in XMR have been created by duplicate offer, not counting the initial offer. Liquidity has thus already increased about 30% despite the improvements that still can be made. Other markets such as EUR and USD also have many duplicated offers. All XMR selloffers have autoconfirm as well.

Maybe the Refund Agents suggestions to improve the UI should be taken more seriously.

RefundAgent:
I now think 1-click will work. Most traders will duplicate immediately after making an initial offer and the fees will be the same. If the fees rise substantially after some time then duplicate offer could give a warning such that the trader can abort the duplication and make an offer from scratch. 

For an offer started from scratch the fields should be filled with the numbers from the previous offer.  An extra option of "make n copies of this offer" should be given, prefilled with a 1.

PR #5399, opened 2021-04-07
CashByMail show terms and conditions upon taking an offer

ghost:
Fixes #5394 

A more explicit suggestion in CashByMail account creation, suggesting countries & other restrictions to be entered.

Maker's terms and conditions are then displayed to the taker who can then choose to continue with the trade or not after reviewing.

Since the terms and conditions can be quite long, a special popup window had to be used which contains a scrollable TextArea control.  The `GenericMessageWindow` is written to be reusable for other features.

PR #5424, opened 2021-04-20
Size popup proportional to amount of text shown

ghost:
This is a followup to #5399 which cleans up a minor display issue.

CashByMail terms are defined by the offer maker, some users have only 1 line of text, others have a lengthy contract.
For the cases where only 1 line is shown, the large popup taking the whole screen looks a bit ridiculous, so here we vertically size the popup proportional to the number of lines in the message (with limits to not be < 20% or > 70% of allowable height).

---

Here are some example screenshots comparing the results with lots of text, and just one line.

image

image


ripcurlx (approved):
utACK

PR #5425, opened 2021-04-21
Allow Revolut username to be minimum 3 characters

BtcContributor:
Since around a week, Revolut users could change their usernames.
Username could be at minimum of 3 characters so the validation has been adjusted accordingly.

Reference: https://www.revolut.com/en-US/help/making-payments/sending-money-to-another-revolut-account/username-payments/can-i-change-my-username

Maximum length has not been touched to preserve email and phone numbers usernames.


ripcurlx (approved):
utACK

PR #5426, opened 2021-04-21
Pricenode: Increase mempool fee suggestion from medium to high priority

wiz:
Due to the recent high mining variance and mempool congestion, some users are reporting stuck trades due to TX getting stuck in the mempool. To help, this PR increases pricenodes usage of mempool.space fee suggestions from medium to high priority.


Emzy:
I agree. And running it on my pricenode.

ripcurlx (approved):
utACK

Conza88:
So it's just a suggestion? 
Would be cool if UX wise, user could get an indication from mempool and choose the 'medium' or 'high' or 'low' suggestion.

pazza83:
I think this includes the transactions for the trade protocol. Hopefully will reduce the cases of trades being stuck for a few days or longer before confirming. 

In these cases I think users should not be able to choose the fee as it could have a detrimental impact on their counterparty (long wait for trade confirmations).

PR #5444, opened 2021-04-28
Add missing check if tx was in correct phase

chimp1984:
There have been 1 tx in cycle 17 and 2 txs in cycle 23 which have been confirmed outside of the phase.
The vote result had excluded it correctly but for the hash calculation for the blind vote monitor the
phase check was missing which results in an incorrect hash.
By fixing that the past hashes back to cycle 17 are all invalid. Once all seed nodes are running that
patch nodes should be in sync again. As it did not affect consensus but only the monitoring it is not
critical.

We see currently 2 different hashes on seeds. I am not sure why but probably some have filtered those invalid txs already out.

### DIFF
--- a/core/src/main/java/bisq/core/dao/monitoring/BlindVoteStateMonitoringService.java
+++ b/core/src/main/java/bisq/core/dao/monitoring/BlindVoteStateMonitoringService.java
@@ -257,8 +257,8 @@ public class BlindVoteStateMonitoringService implements DaoSetupService, DaoStat
 
         periodService.getCycle(blockHeight).ifPresent(cycle -> {
             List<BlindVote> blindVotes = blindVoteListService.getConfirmedBlindVotes().stream()
-                    .filter(e -> periodService.isTxInPhaseAndCycle(e.getTxId(), DaoPhase.Phase.BLIND_VOTE, blockHeight))
                     .filter(e -> e.getTxId() != null)
+                    .filter(e -> periodService.isTxInPhaseAndCycle(e.getTxId(), DaoPhase.Phase.BLIND_VOTE, blockHeight))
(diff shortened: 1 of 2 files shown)


core/src/main/java/bisq/core/dao/monitoring/BlindVoteStateMonitoringService.java
```diff
@@ -257,8 +257,10 @@ private boolean maybeUpdateHashChain(int blockHeight) {
 
         periodService.getCycle(blockHeight).ifPresent(cycle -> {
             List<BlindVote> blindVotes = blindVoteListService.getConfirmedBlindVotes().stream()
-                    .filter(e -> periodService.isTxInCorrectCycle(e.getTxId(), blockHeight))
-                    .sorted(Comparator.comparing(BlindVote::getTxId)).collect(Collectors.toList());
+                    .filter(e -> periodService.isTxInPhaseAndCycle(e.getTxId(), DaoPhase.Phase.BLIND_VOTE, blockHeight))
+                    .filter(e -> e.getTxId() != null)
```
sqrrm: I think it would be better to reverse the order of the filters to filter out null txid before using it.

Otherwise it looks correct.

sqrrm:
@ManfredKarrer Need your review on this DAO related change.

core/src/main/java/bisq/core/dao/monitoring/BlindVoteStateMonitoringService.java
chimp1984: Ah good point.

sqrrm (approved):
utACK

Still need ACK from @ManfredKarrer

ManfredKarrer (approved):
utACK

PR #5447, opened 2021-04-28
Fix bsq wallet & tx fee rate test cases (api)

ghubstan:
- `BsqWalletTest`  Regtest btc block being generated with this fix, and chain download triggered.   Also adjusted testGetUnusedBsqAddress() for upcoming SegWit BSQ changes.

- `BtcTxFeeRateTest` Minimum tx fee rates move fast, sometimes so low the validation test fails.  This change tries to consistently force validation error by setting fee rate=1.


ripcurlx (approved):
utACK

PR #5449, opened 2021-04-29
Update DAO data store for v1.6.3

ripcurlx:
This is a new updated DAO data store that is synced from last release running a full node with the BSQ SegWit PR applied.
Light node in light mode and full node in dark mode. There is a conflict on both nodes in the blind vote state with one of the seednodes because of a difference in the number of votes, which relies on the fact that two votes were mined too late and were counted as invalid.

Bildschirmfoto 2021-04-29 um 20 18 24
Bildschirmfoto 2021-04-29 um 20 18 29
Bildschirmfoto 2021-04-29 um 20 18 42
Bildschirmfoto 2021-04-29 um 20 18 47
Bildschirmfoto 2021-04-29 um 20 19 06
Bildschirmfoto 2021-04-29 um 20 19 10

### DIFF
--- a/p2p/src/main/resources/DaoStateStore_BTC_MAINNET
+++ b/p2p/src/main/resources/DaoStateStore_BTC_MAINNET
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:d946925601bd28d60abeef0f368f21c455349044ac5cd8ac24557105bfc05828
-size 121812570
+oid sha256:123ebb76206d420d0998a68ead2a7f70184e3bee36eeea713bbf49d66643ce3a
+size 122349144


ripcurlx:
@chimp1984 Please have a look at it as well if this state is fine from your point of view. Thanks!

chimp1984:
Looks ok to me. Maybe u can cross check the DAO state hash with http://46.101.179.224:8080/

ripcurlx:
> Looks ok to me. Maybe u can cross check the DAO state hash with http://46.101.179.224:8080/

The dao state hash seems to be different. Proposal and blind vote is the same.

chimp1984:
> > Looks ok to me. Maybe u can cross check the DAO state hash with http://46.101.179.224:8080/
> 
> The dao state hash seems to be different. Proposal and blind vote is the same.

Oh, I will cross check... Lets wait with release until its confirmed that its ok.

chimp1984:
I was running your branch and I get the same daostate hash:
DAO state block height: 681146
DAO state hash: f0ad4135c637196ebe9a240e4ae1e9799fc0cf93

You have to check if the block height is the same...

sqrrm (approved):
ACK

PR #5451, opened 2021-04-30
Fix string when creating a new offer

BtcContributor:
Now it is similar to the message provided when taking a new offer.

See: https://github.com/bisq-network/bisq/blob/3047cd48d42c50f908f5750a8dccea194c057c69/core/src/main/resources/i18n/displayStrings.properties#L541

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -538,7 +538,7 @@ takeOffer.noPriceFeedAvailable=You cannot take that offer as it uses a percentag
 takeOffer.takeOfferFundWalletInfo.headline=Fund your trade
 # suppress inspection "TrailingSpacesInProperty"
 takeOffer.takeOfferFundWalletInfo.tradeAmount=- Trade amount: {0} \n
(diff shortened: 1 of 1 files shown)


ripcurlx:
@m52go Could you please ACK this change? Thanks!

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -471,7 +471,7 @@ createOffer.placeOfferButton=Review: Place offer to {0} bitcoin
 createOffer.createOfferFundWalletInfo.headline=Fund your offer
 # suppress inspection "TrailingSpacesInProperty"
 createOffer.createOfferFundWalletInfo.tradeAmount=- Trade amount: {0} \n
-createOffer.createOfferFundWalletInfo.msg=You need to deposit {0} to this offer.\n\n\
+createOffer.createOfferFundWalletInfo.msg=You need to deposit {0} for making this offer.\n\n\
```
```suggestion
createOffer.createOfferFundWalletInfo.msg=You need to deposit {0} to make this offer.\n\n\
```

BtcContributor:
@m52go so should we also change this https://github.com/bisq-network/bisq/blob/3047cd48d42c50f908f5750a8dccea194c057c69/core/src/main/resources/i18n/displayStrings.properties#L541 ?

Like that: You need to deposit {0} to take this offer

m52go:
> @m52go so should we also change this 

Yes, that would be better.

m52go:
ACK

ripcurlx (approved):
utACK

PR #5453, opened 2021-04-30
Change mempool.bisq.services onion address

devinbileck:
In the process of migrating the server to a new instance, my secret key
for the onion address seems to have become corrupted. Therefore, I
needed to generate a new onion address.


ripcurlx (approved):
utACK

PR #5469, opened 2021-05-05
Filter banned seednodes with wrong format

ripcurlx:
This prevents incorrect formatted banned seednodes in bisq.properties file to block Bisq from starting.

It will only log an error and filter all banned seednodes that are correctly formatted.
I checked banned BTC nodes and banned price relay nodes and those are handled gracefully already.


sqrrm (approved):
utACK

PR #5473, opened 2021-05-07
Fix focus issue when enter key used to edit offer

### LINKED ISSUE #5460: 1.6.3 Market price distance is always 0.00% when editing an offer
### Description
When you edit an existing offer, the "Above % from market price" section is shown as 0.00% while the real distance is different.

#### Version
1.6.3

### Steps to reproduce
1. close bisq
2. start bisq
3. wait until it's synced
4. go to Portfolio/My open offers
5. click on the pencil icon to edit an offer (this assumes that you have some offers published)
6. notice the percentage of the offer is correct, like "10.41%"
7. change the percentage to, say, "10.15%" and press Enter to save changes
8. notice the percentage on the offers list is shown correctly, like "10.15%"
8. click again on the pencil icon of the offer in Portfolio/My open offers
9. notice the percentage is show as "0.00%" in the Edit offer screen

### Expected behaviour
The market price distance from point 3 should be the same as in the Deviation section of the Portfolio/My open offers screen no matter whether you press Enter or click "Confirm: Edit offer" when editing the offer

### Actual behaviour

The distance from market price is shown as 0.00% when you earlier pressed Enter when editing an offer, it is correct when you clicked on the button "Confirm: Edit offer"

### Screenshots
grafik

#### Device or machine
MS Windows 10

#### Additional info
This bug first occured in 1.6.3. After I have updated, sometimes the shown distance was correct, sometimes it wasn't, now it's exclusively 0.00%, though most of my offers were not taken.

ghost:
If an offer percent is edited and submitted using the `enter` key to accept the changes, the **next time** an offer is edited its price percentage field is always shown as 0%.

Using `enter`, the control with focus is not the submit button, whereas it always is when the mouse is used.  However, the next time the edit offer screen is shown, it does not initialize correctly if the focus is not on the submit button (the dialog is cached, so the previous controls state are still in effect).  Various initialization events get fired that reset the price percentage field to 0%, overwriting the value stored in the data object.  The initialization process for the edit offer screen is the most difficult GUI code in Bisq, so the simplest way to fix this is to place focus on the submit button when its action is invoked.

Note: issue happens in all releases <= 1.6.4
Fixes #5460

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/portfolio/editoffer/EditOfferView.java
+++ b/desktop/src/main/java/bisq/desktop/main/portfolio/editoffer/EditOfferView.java
@@ -200,6 +200,7 @@ public class EditOfferView extends MutableOfferView<EditOfferViewModel> {
         editOfferConfirmationBox.getChildren().add(cancelButton);
 
         confirmButton.setOnAction(e -> {
+            confirmButton.requestFocus();   // fix issue #5460 (when enter key used, focus is wrong)
             if (model.isPriceInRange()) {
                 model.isNextButtonDisabled.setValue(true);
                 cancelButton.setDisable(true);


ripcurlx (approved):
utACK

PR #5475, opened 2021-05-07
Fix the formula used to calculate BSQ amount needed to avoid dust

BtcContributor:
The formula used to calculate the BSQ needed to avoid dust was incorrect.

For example, this was before:

paste

The BSQ amount needed should be 8, not 13. Now it is fixed.


sqrrm (approved):
utACK

pazza83:
I am confused is 5.46 not the dust limit?

bsq min

sqrrm:
5.46 is the dust limit, so each output need to be at least 546 sat. If you have 7 BSQ (700 sat) you can burn 1.54 BSQ and have a change output of 5.46 since the burnt BSQ is used to pay the mining fee which is not an output, just the remainder of all the inputs minus outputs.

Please let me know if this didn't help explain the situation.

PR #5478, opened 2021-05-09
Restore missing displayStrings strings in Settings

BtcContributor:
For mistake I forgot to restore these strings in a previous PR.

Fix instances like this:
image

Thanks to @jmacxx and @wiz for pointing out this and sorry about that.


rafaelpac:
While checking this, I noticed that despite #5224 had raised minfee to 15, I still can get 10 as the real minimum. Cant find why!

ghost:
@rafaelpac  https://github.com/bisq-network/bisq/pull/5235 superseded 5224, in which a dynamic minimum fee rate is obtained from bisq mempool servers.

ripcurlx (approved):
utACK

PR #5483, opened 2021-05-11
Fix custom fee rate field when custom inputs are used

BtcContributor:
If a user chooses to use custom inputs in Funds menu, custom fee rate field is wrongly blanked and this leads to _NumberFormatException_.

Fixed by updating _inputsToggleGroupListener_

Thanks @jmacxx for having pointed out this.


ripcurlx (approved):
utACK

PR #5484, opened 2021-05-11
Speed up "Synchronizing DAO" by ~30%

cd2357:
`BlockParser.parseBlock()` normally takes about 300ms per block.

One part of it is checking if the new block hash is known, which was done by string-comparing it to the hashes of all known blocks. In my case, this was about 110 000 string-comparisons per `BlockParser.parseBlock()` call.

This commit adds a block hash cache in the `DaoState` and replaces the hash string-comparisons by one `HashSet` lookup.

With this change, `BlockParser.parseBlock()` takes around 200ms on my setup.

### DIFF
--- a/core/src/main/java/bisq/core/dao/state/model/DaoState.java
+++ b/core/src/main/java/bisq/core/dao/state/model/DaoState.java
@@ -76,7 +76,9 @@ public class DaoState implements PersistablePayload {
 
     @Getter
     private int chainHeight; // Is set initially to genesis height
-    @Getter
+
+    // We override the getter so callers can't modify the list without also updating
+    // the block caches and indices below
     private final LinkedList<Block> blocks;
     @Getter
     private final LinkedList<Cycle> cycles;
@@ -109,9 +111,9 @@ public class DaoState implements PersistablePayload {
     @JsonExclude
     private transient final Map<String, Tx> txCache; // key is txId
     @JsonExclude
-    private transient final Map<Integer, Block> blockHeightCache; // key is block height
+    private transient final Map<Integer, Block> blocksByHeight; // Blocks indexed by height
     @JsonExclude
-    private transient final Set<String> blockHashCache;
+    private transient final Set<String> blocksByHash; // Blocks indexed by hash
 
     ///////////////////////////////////////////////////////////////////////////////////////////
     // Constructor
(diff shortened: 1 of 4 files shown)


ripcurlx:
As by definition this needs an ACK by @ManfredKarrer to get merged.

core/src/main/java/bisq/core/dao/state/model/DaoState.java
```diff
@@ -109,6 +109,8 @@ public static DaoState getClone(DaoState daoState) {
     @JsonExclude
     private transient final Map<String, Tx> txCache; // key is txId
     @JsonExclude
+    private transient final Map<Integer, Block> blockHeightCache; // key is block height
```
chimp1984: Maybe better name is blocksByHeight.

chimp1984 (changes requested):
@cd2357 Thanks for working on that.
There is missing the chache update when a new block is added in `onNewBlockWithEmptyTxs`.
Maybe its better to create explicit methods for `daoState.getBlocks().addAll` and `daoState.getBlocks().add` in DaoStateService.

ripcurlx:
@chimp1984 Thanks for doing a review as well! But still we'll need in the end the ACK by @ManfredKarrer as well.

chimp1984:
> @chimp1984 Thanks for doing a review as well! But still we'll need in the end the ACK by @ManfredKarrer as well.

Just to be clear, it is a NACK at current state. Cache does not get updated at new blocks.

cd2357:
Done, thanks for the speedy review.

> There is missing the chache update when a new block is added in `onNewBlockWithEmptyTxs`.
> Maybe its better to create explicit methods for `daoState.getBlocks().addAll` and `daoState.getBlocks().add` in DaoStateService.

Good point. I restricted the adding of blocks to two new methods, `addBlock` and `addBlocks`. They make sure to update all the indices as well. Also made sure the `DaoState.blocks` list is not exposed directly anymore, but as an unmodifiable list, so no direct modifications are possible.

core/src/main/java/bisq/core/dao/state/model/DaoState.java
```diff
@@ -109,9 +111,9 @@ public static DaoState getClone(DaoState daoState) {
     @JsonExclude
     private transient final Map<String, Tx> txCache; // key is txId
     @JsonExclude
-    private transient final Map<Integer, Block> blockHeightCache; // key is block height
+    private transient final Map<Integer, Block> blocksByHeight; // Blocks indexed by height
     @JsonExclude
-    private transient final Set<String> blockHashCache;
+    private transient final Set<String> blocksByHash; // Blocks indexed by hash
```
chimp1984: With blocksByHash I would expect a map with hash as key and blocks as value. For a set I would use 'blockHashes'

chimp1984 (dismissed):
Beside the small naming issue utACK

core/src/main/java/bisq/core/dao/state/model/DaoState.java
cd2357: Agree. I had forgotten one is a set, the other a map.

ripcurlx:
So only the ACK by @ManfredKarrer is finally missing :smile: :+1:

ManfredKarrer (approved):
utACK

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/5484#pullrequestreview-660971809

PR #5485, opened 2021-05-11
Return correct security deposit percentage size/minimum deposit when cloning an offer

### LINKED ISSUE #5470: 1.6.3 Different deposit percentage size when cloning an offer
### Description

When you choose the option "Create new offer like this..." all data from the old offer is replicated except for the security deposit size which is newly calculated. It is easy to miss it.

#### Version

1.6.3

### Steps to reproduce

1. Go to Portfolio/Open trades
2. Right click on an offer and choose "Create new offer like this..."
3. "Duplicate offer" section will show up
3. Notice the "Set both traders' security deposit (%)" value is different than in the offer being cloned

### Expected behaviour

In point 3 the value should be like in the original offer which is being cloned.

### Actual behaviour

The security deposit size is newly calculated

### Screenshots

grafik

#### Device or machine

Windows 10

BtcContributor:
Fixes #5470 as security deposit was not previously set.
It works both with minimum security deposit and arbitrary percentage.

Thanks to @jmacxx for letting me work on this.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/portfolio/duplicateoffer/DuplicateOfferDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/portfolio/duplicateoffer/DuplicateOfferDataModel.java
@@ -86,11 +86,12 @@ class DuplicateOfferDataModel extends MutableOfferDataModel {
         setAmount(offer.getAmount());
         setPrice(offer.getPrice());
         setVolume(offer.getVolume());
+        setUseMarketBasedPrice(offer.isUseMarketBasedPrice());
+
         if (offer.getBuyerSecurityDeposit().value == Restrictions.getMinBuyerSecurityDepositAsCoin().getValue())
             setBuyerSecurityDeposit(Restrictions.getMinBuyerSecurityDepositAsPercent());
         else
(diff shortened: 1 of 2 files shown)


desktop/src/main/java/bisq/desktop/main/portfolio/duplicateoffer/DuplicateOfferDataModel.java
```diff
@@ -85,6 +87,12 @@ public void populateData(Offer offer) {
         setPrice(offer.getPrice());
         setVolume(offer.getVolume());
         setUseMarketBasedPrice(offer.isUseMarketBasedPrice());
+
+        if (offer.getBuyerSecurityDeposit().value == Restrictions.getMinBuyerSecurityDepositAsCoin().getValue())
+            setBuyerSecurityDeposit(Restrictions.getMinBuyerSecurityDepositAsPercent());
+        else
+            setBuyerSecurityDeposit(CoinUtil.getAsPercentPerBtc(offer.getBuyerSecurityDeposit(), offer.getAmount()));
+
```
ripcurlx: NIT: we decided to always use brackets, even for one-liners as we had issues in the past.

ripcurlx (changes requested):
NACK

ripcurlx (approved):
utACK

PR #5487, opened 2021-05-11
Fix mac jar hash

cd2357:
Copy the deterministic, signed jar to the binaries folder only after it has been processed. Once copied there, it will be checksum-ed in the last packaging step and the resulting hash will be ready for publishing with the release.

This solves the issue in v1.6.3 and v1.6.4 where the jar was checksum-ed and copied to the binaries folder _before_ being signed and processed (dylib processing, macOS-specific). This resulting checksum would therefore not match anymore with the hash of the signed jar contained in the `dmg`.


cd2357:
Discussion ongoing about how to best solve this.

The `dmg` now requires a signed jar, but the act of signing itself modifies its hash. Therefore, this PR doesn't actually address the problem.

cd2357:
As discussed with @ripcurlx this might be worth trying out for the mac build, even though that jar won't be reproducible on other platforms due to signing.

ripcurlx (approved):
utACK

PR #5489, opened 2021-05-12
Fix the PriceNode lower bounds for MinFee

### LINKED ISSUE #5486: Minimum Fee behaving weirdly
### Description

I believe that there is something wrong with the fee estimation implemented in  #5235
I cant find what it is, but the final behaviour here is that it gets a value that is not coming from the mempool.space api (is it an average?) and it also has a 10sats/vb hardcoded minimum.
The image below shows that 21sats/vb, but the minimumFee was 1sat/vb (as far as I can tell from the code, it should use 2xminimumFee).
If I try 10sats/vb, it will accept it, but not less.

image

#### Version

1.6.4

### Steps to reproduce

Try to change the minumum fee.

### Expected behaviour

Should accept at least 2 x minimumFee, but stucks to 10 or a number that it takes from an unknown place.

ghost:
PriceNode caps the MinFee and the TxFee both to a minimum of 10.  The MinFee cap should be 2, which corresponds to the smallest allowable custom withdrawal fee.

Fixes #5486

### DIFF
--- a/pricenode/src/main/java/bisq/price/mining/FeeRateProvider.java
+++ b/pricenode/src/main/java/bisq/price/mining/FeeRateProvider.java
@@ -26,7 +26,8 @@ import java.time.Duration;
  */
 public abstract class FeeRateProvider extends PriceProvider<FeeRate> {
 
-    public static final long MIN_FEE_RATE = 10; // satoshi/vbyte
+    public static final long MIN_FEE_RATE_FOR_WITHDRAWAL = 2; // satoshi/vbyte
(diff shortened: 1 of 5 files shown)


rafaelpac:
Nice.
But wouldnt it be better to avoid hardcoding the trading fee? What about using the HourFee for trading (and FastestFee for instant payment methods) and 2 x MinimunFee for withdraw?

ghost:
@rafaelpac This is just enforcing established consensus rules, changing them would need approval from the DAO.

wiz (approved):
ACK - thanks for fixing this bug!

ripcurlx (approved):
utACK

PR #5490, opened 2021-05-12
change MIN_FEE_RATE to 2 form 10

Emzy:
The mempool is getting empty and we put the MIN_FEE_RATE = 10 as a temporary fix.


ghost:
The mining fee rate for trading should not go below 10 sats/vB.  The mining fee rate for withdrawals should not go below 2 sats/vB.  That's my understanding of the established rules.  Pricenode was applying the constant MIN_FEE_RATE=10 for both cases.  Changing it to 2 would lead to trades using 2 sats/vB making them vulnerable to sudden fee spikes and the higher change of not confirming.

See #5489.

bisqubutor:
MIN_FEE_RATE

bisqubutor:
MIN_FEE_RATE

PR #5497, opened 2021-05-16
Remove jcenter repository

### LINKED ISSUE #5496: Remove JCenter dependencies from build
JCenter is [shutting down services this year](https://www.infoq.com/news/2021/02/jfrog-jcenter-bintray-closure/). They will continue to serve packages until February 1st 2022.

Bisq currently relies on jcenter in the build process: https://github.com/bisq-network/bisq/blob/e375e9aadb278b0dc9ed659467baf56a97cec840/build.gradle#L2-L4

Therefore, remove the dependency on jcenter and use another repository instead.

cd2357:
Replace the jcenter repo with mavenCentral and jitpack.

One of the dependencies had a slightly different name in jcenter compared to jitpack (`com.github.jengelman.gradle.plugins:shadow` vs `com.github.johnrengelman:shadow`), so that was renamed as well.

Fixes #5496

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,11 +1,12 @@
 buildscript {
     repositories {
-        jcenter()
+        mavenCentral()
+        maven { url "https://jitpack.io" }
     }
     dependencies {
         classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.10'
         classpath 'com.google.gradle:osdetector-gradle-plugin:1.6.0'
-        classpath 'com.github.jengelman.gradle.plugins:shadow:5.2.0'
+        classpath 'com.github.johnrengelman:shadow:5.2.0'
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
utACK - as the lib hash hasn't been changed, so it should still be exact the same lib.

cd2357:
I didn't find the `shadow` dependency hash in `gradle-witness.gradle`, maybe because its not a dependency _per se_ but just a plugin used during building.

As a test, I updated the version to 6.0.0 in https://github.com/bisq-network/bisq/blob/ee6ccc48691929668d9775f4e3d45a86f8da2fa6/build.gradle#L9 and `gradle-witness` didn't complain.

So the hash isn't checked.

I did find a way to preserve the plugin name (by adding another repo to the list, PR incoming).

PR #5500, opened 2021-05-17
Add Account owner full name field to Uphold payment method

### LINKED ISSUE #5493: Uphold payment method issues to fix
### Description

The Uphold payment method could do with a couple of tweaks to make it a smoother trading experience.

- Add 'Account owner full name' field
- Fix error showing a 4 day trade period to 1 day trade period

When payment is received from an Uphold member an email is received to show the full name of the payee. This is also shown in the app when clicking on the payment details. 

There is no way to identify payee from the current field of 'Username or email or phone no'. You can use the payment amount / currency to link payment , but when multiple trades are made this makes it difficult / impossible to achieve. Adding an 'Account owner full name'  field would fix this.

The trade screen also shows the payment time as 4 days. This is incorrect. It should be 1 day. See screen shot below:

uphold bisq

#### Version

1.6.4

### Steps to reproduce

1. Add Uphold as a payment method. There is no field to add 'Account owner full name'. This should be a required field.
2. Start a trade with Uphold. The UI will show 'Max. trade allowed period' as 4 days. It should be 1 day (the actual remaining time and actual trade window is correct at one day' so I think the 4 day is just a typo.

### Expected behaviour

1. Add Uphold as a payment method. User see a field to add 'Account owner full name'. This should be a required field.
2. Start a trade with Uphold. The UI should show 'Max. trade allowed period' as 1 day.

BtcContributor:
Fixes #5493

Now it looks like this:

pazza_uphold_fix3

pazza_uphold_fix

pazza_uphold_fix2

### DIFF
--- a/core/src/main/java/bisq/core/payment/UpholdAccount.java
+++ b/core/src/main/java/bisq/core/payment/UpholdAccount.java
@@ -44,4 +44,13 @@ public final class UpholdAccount extends PaymentAccount {
     public String getAccountId() {
         return ((UpholdAccountPayload) paymentAccountPayload).getAccountId();
     }
+
+    public String getAccountOwner() {
+        return ((UpholdAccountPayload) paymentAccountPayload).getAccountOwner();
+    }
+
+    public void setAccountOwner(String accountOwner) {
+        if (accountOwner == null) accountOwner = "";
+        ((UpholdAccountPayload) paymentAccountPayload).setAccountOwner(accountOwner);
+    }
 }
--- a/core/src/main/java/bisq/core/payment/UpholdAccount.java
+++ b/core/src/main/java/bisq/core/payment/UpholdAccount.java
@@ -50,7 +50,9 @@ public final class UpholdAccount extends PaymentAccount {
     }
 
     public void setAccountOwner(String accountOwner) {
-        if (accountOwner == null) accountOwner = "";
+        if (accountOwner == null) {
+            accountOwner = "";
+        }
         ((UpholdAccountPayload) paymentAccountPayload).setAccountOwner(accountOwner);
     }
 }
--- a/core/src/main/java/bisq/core/payment/payload/UpholdAccountPayload.java
+++ b/core/src/main/java/bisq/core/payment/payload/UpholdAccountPayload.java
@@ -100,14 +100,14 @@ public final class UpholdAccountPayload extends PaymentAccountPayload {
 
     @Override
     public String getPaymentDetailsForTradePopup() {
-                if (accountOwner != "") {
-                    return
-                            Res.get("payment.account") + ": " + accountId + "\n" +
-                                    Res.get("payment.account.owner") + ": " + accountOwner;
-                } else {
+                if (accountOwner.equals("")) {
                     return
                             Res.get("payment.account") + ": " + accountId + "\n" +
                                     Res.get("payment.account.owner") + ": N/A";
+                } else {
+                    return
+                            Res.get("payment.account") + ": " + accountId + "\n" +
+                                    Res.get("payment.account.owner") + ": " + accountOwner;
                 }
     }
 
--- a/core/src/main/java/bisq/core/payment/payload/UpholdAccountPayload.java
+++ b/core/src/main/java/bisq/core/payment/payload/UpholdAccountPayload.java
@@ -39,6 +39,7 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 public final class UpholdAccountPayload extends PaymentAccountPayload {
     private String accountId = "";
+    private String accountOwner = "";
 
     public UpholdAccountPayload(String paymentMethod, String id) {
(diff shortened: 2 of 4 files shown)


proto/src/main/proto/pb.proto
```diff
@@ -1132,7 +1132,8 @@ message OKPayAccountPayload {
 }
 
 message UpholdAccountPayload {
-    string account_id = 1;
+    string account_owner = 1;
+    string account_id = 2;
```
ripcurlx: Never change the IDs of the old values, use incremented ones for new fields (`string account_owner=2`). This would cause issues with existing accounts. Please also run some backwards comp. tests with this change to be sure it works as expected.

ripcurlx (changes requested):
NACK - Please see my comment regarding proto changes

BtcContributor:
Restored proto order and add backward compatibility, like done in https://github.com/bisq-network/bisq/pull/5117

(Not sure why screenshots are yellow...)


uphold_back

uphold_back2

Not sure if we should also update existing account at startup. What do you think @pazza83 and @ripcurlx?

pazza83:
Looks great thanks for this.

I would be in favor of updating existing accounts at startup if not too much additional work

core/src/main/java/bisq/core/payment/UpholdAccount.java
```diff
@@ -44,4 +44,13 @@ public void setAccountId(String accountId) {
     public String getAccountId() {
         return ((UpholdAccountPayload) paymentAccountPayload).getAccountId();
     }
+
+    public String getAccountOwner() {
+        return ((UpholdAccountPayload) paymentAccountPayload).getAccountOwner();
+    }
+
+    public void setAccountOwner(String accountOwner) {
+        if (accountOwner == null) accountOwner = "";
```
ripcurlx: Please use brackets around this conditional.

core/src/main/java/bisq/core/payment/UpholdAccount.java
```diff
@@ -44,4 +44,13 @@ public void setAccountId(String accountId) {
     public String getAccountId() {
         return ((UpholdAccountPayload) paymentAccountPayload).getAccountId();
     }
+
+    public String getAccountOwner() {
```
ripcurlx, line 48: Is there a reason why you don't display the account owner in `addFormForDisplayAccount`?

core/src/main/java/bisq/core/payment/payload/UpholdAccountPayload.java
```diff
@@ -85,12 +95,20 @@ public static UpholdAccountPayload fromProto(protobuf.PaymentAccountPayload prot
...
-        return getPaymentDetails();
+                if (accountOwner.equals("")) {
+                    return
+                            Res.get("payment.account") + ": " + accountId + "\n" +
+                                    Res.get("payment.account.owner") + ": N/A";
+                } else {
+                    return
+                            Res.get("payment.account") + ": " + accountId + "\n" +
+                                    Res.get("payment.account.owner") + ": " + accountOwner;
+                }
```
ripcurlx: Could you please decrease the intention

ripcurlx (changes requested):
NACK - proto and backwrd. comp. are looking fine now, just a couple of minor adjustments needed.

core/src/main/java/bisq/core/payment/UpholdAccount.java
BtcContributor: Nope, my mistake :)

ripcurlx (dismissed):
utACK

core/src/main/java/bisq/core/payment/payload/UpholdAccountPayload.java
```diff
@@ -85,12 +95,20 @@ public static UpholdAccountPayload fromProto(protobuf.PaymentAccountPayload prot
...
     @Override
     public String getPaymentDetails() {
-        return Res.get(paymentMethodId) + " - " + Res.getWithCol("payment.account") + " " + accountId;
+        return Res.get(paymentMethodId) + " - " + getPaymentDetailsForTradePopup().replace("\n", ", ");
     }
 
     @Override
     public String getPaymentDetailsForTradePopup() {
-        return getPaymentDetails();
+        if (accountOwner.equals("")) {
```
skarelin: `"".equals(accountOwner)` ?

ripcurlx: I don't think we need the extra null check for `accountOwner` in this case as it is always initialized with `""`. Probably `accountOwner.isEmpty()` would be more readable for this backwardscomp. check. @BtcContributor WDYT?

BtcContributor: Thanks, good catch!

ripcurlx (approved):
utACK

PR #5502, opened 2021-05-17
Preserve shadow plugin name

cd2357:
Include plugins.gradle.org repository, which contains the shadow plugin of the same artefact group and name as used before in jcenter.

Addresses issue with plugin name mentioned in https://github.com/bisq-network/bisq/pull/5497#issuecomment-842226991


ripcurlx (approved):
utACK - I have added this plugin source myself while trying to figure out how to best load the javaFX modules in an officially supported way

PR #5503, opened 2021-05-18
Replace gradle-witness with gradle verification

cd2357:
Remove dependency on gradle-witness. Add config file for the built-in gradle verification mechanism. Bootstrap initial list of hashes.

I have checked previously known hashes in `gradle-witness.gradle` and confirmed they are present in the new hashes list in `verification-metadata.xml` (this can be double-checked by reviewers as well).

As opposed to `gradle-witness`, which could only verify dependencies, the gradle verification task covers both dependencies and plugins.

In addition to checksum verification, the gradle task also has support for PGP verification.

Usage:
  - the task is called automatically during build. It succeeds if all used dependencies and plugins match the defined hashes, fails otherwise.
  - If new dependencies are added, or a dependency version is changed, use `./gradlew --write-verification-metadata sha256 clean build -x test` to *optimistically* bootstrap the list of hashes. The `clean build` at the end is necessary for gradle to navigate the entire dependency tree of all Bisq submodules and find any new or changed dependency. Alternatively, the hash can be manually retrieved from the dependency's repo and added to `verification-metadata.xml`.

For details, see https://docs.gradle.org/current/userguide/dependency_verification.html

For details on bootstrapping the initial hashes, see https://docs.gradle.org/current/userguide/dependency_verification.html#sec:bootstrapping-verification


ripcurlx:
@cd2357 Did you see the verification failure in [GitHub](https://github.com/bisq-network/bisq/pull/5503/checks?check_run_id=2608302237) and Travis build (mainly javafx related). The problem with this is I guess that we need different hashes for different OSes. Not sure how easy that can be configured. We did exclude in the past the javaFX library from the checks which is bad, but there wasn't an easy solution for this.

gradle/verification-metadata.xml
```diff
@@ -2111,6 +2122,9 @@
          </artifact>
       </component>
       <component group="org.openjfx" name="javafx-base" version="15">
+         <artifact name="javafx-base-15-linux.jar">
```
ripcurlx: But wouldn't that fail now locally when e.g. building a release binary on macOS?

cd2357: No, this simply adds the expected hash for the linux libs to the list. There is a different hash for `javafx-base-15-mac.jar` / `javafx-base-15-linux.jar` / `javafx-base-15-windows.jar`. Depending on which platform the gradle build is run, a different lib is loaded and a different hash is checked. As long as all 3 are defined in this list, it should be fine.

Osimaco:
Greetings

zulkifli25 (changes requested):
github.com exchanges

Osimaco:
Please I want to sale my site are u interred?

cd2357:
Superseded by #5824

cbeams:
> Superseded by #5824

Oops, I hadn't realized this PR was already here!

cd2357:
No problem @cbeams , yours is more comprehensive.

PR #5512, opened 2021-05-19
Withdrawal view: reduce vertical spacing and fix issue with custom fee

### LINKED ISSUE #5457: Unable to send funds from custom inputs (low res screens) (v1.6.3)
On limited resolution screens, it is impossible to select a custom input to send funds, because the table usually at the upper part of the screen is not visible.

I think this screen needs some rework to make it usable.  Too much vertical space is taken up by the bottom 6 fields.
Additionally, it is possible to get the fields drawn over each other when resizing (double click title bar).

---

As far as I understood, Bisq was designed to support a minimum screen resolution of 1020 x 620

https://github.com/bisq-network/bisq/blob/ec4ce6027d9f1eccb1d7aabd291644188a97bee4/desktop/src/main/java/bisq/desktop/util/Layout.java#L23-L24

---

**Screenshots:**
a laptop with resolution 1366 x 768 the table of custom inputs is not reachable at all, there is no scroll bar to access it.

image

ghost:
- Reduce the vertical spacing in order to show the custom inputs table on smaller screens.
- Reduce vertical space used by showing the custom fee toggle to the right of its input box.
- Don't reset the user's chosen fee rate when clicking in the table or the custom input radio button.
- Don't overwrite the user's custom fee with FeeService updates.

Fixes: #5457

Before fix:
image

After fix:
image

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/funds/withdrawal/WithdrawalView.java
+++ b/desktop/src/main/java/bisq/desktop/main/funds/withdrawal/WithdrawalView.java
@@ -86,6 +86,7 @@ import javafx.scene.control.Tooltip;
 import javafx.scene.layout.GridPane;
 import javafx.scene.layout.VBox;
 
+import javafx.geometry.HPos;
 import javafx.geometry.Pos;
 
 import javafx.beans.property.BooleanProperty;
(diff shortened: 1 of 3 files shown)


ripcurlx (approved):
utACK

PR #5514, opened 2021-05-20
Build: Include build platform in fat jar filename

cd2357:
Depending on which platform it was built, the resulting fat jar (and its hash file) will include the platform in the name.

For example on mac:
- `desktop-1.6.5-all-mac.jar`
- `desktop-1.6.5-all-mac.jar.SHA-256`

Depends on #5487


ripcurlx (approved):
utACK

PR #5515, opened 2021-05-20
Package: Open binaries shared folder on finish

cd2357:
If user selects Y at the last packaging step ("copy binaries to shared folder"), open an explorer window at the shared folder location after binaries are copied. Copy binaries to shared folder only after all binary processing is done (including sha generation).


ripcurlx (approved):
utACK

PR #5516, opened 2021-05-20
Fix raspi build

cd2357:
Fix packaging issue which sometimes caused issues with creating the raspi package on Windows (and other environments where a recent `./gradlew :desktop:build` was not run recently).

Builds on top of #5515


ripcurlx (approved):
utACK

PR #5517, opened 2021-05-20
Package: Remove one y/n interaction: Enforce Java 11

cd2357:
Enforce JDK 11 for building and packaging. If user tries to package binaries using any other version, the packaging will fail.


ripcurlx (approved):
utACK

PR #5520, opened 2021-05-21
Reduce the number of connected BTC nodes from 9 to 7

ripcurlx:
After a discussion on #ops on keybase with @chimp1984 @Emzy and @wiz it was decided to reduce the number of connected BTC nodes, in case the provided nodes are used, from 9 to 7. This should reduce unnecessary load on our federated nodes.


ripcurlx:
Although we don't support the public network option right now I've adapted the code to be able to set different defaults for public and provided nodes.

sqrrm (dismissed):
utACK

sqrrm (approved):
utACK

PR #5524, opened 2021-05-24
Build: Change default console to plain

cd2357:
Change default gradle console style to plain. This makes y/n user prompts more readable during installer packaging, without having to specify `--console=plain`.

Now it packaging installers can be directly done with `./gradlew packageInstallers`


ripcurlx (approved):
utACK

PR #5528, opened 2021-05-25
Add pending callbacks to FeeService

sqrrm:
This fix guarantees a callback if there is a pending request rather than just
dropping the fee request.

When requestFee() is called the callbacks are added to a list of callBacks.
Once the next fee request completes all callbacks in the list are called
and the list cleared.

This issue popped up while testing atomic trades with the API and could likely be an issue for other usages of the API. If requestFees() is called while there is an pending request no callback is called. Since it's expected that either the resultHandler or the faultHandler is called it's not clear what happens for callers relying on the callbacks. This PR adds a guarantee one callback is called.

Also return early for the frequency check.

### DIFF
--- a/core/src/main/java/bisq/core/provider/fee/FeeService.java
+++ b/core/src/main/java/bisq/core/provider/fee/FeeService.java
@@ -41,6 +41,8 @@ import javafx.beans.property.SimpleIntegerProperty;
 
 import java.time.Instant;
 
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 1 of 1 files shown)


ghubstan:
Tested ACK

I used an api test to make hundreds of fee requests @ 1 per 0.5s -- no problems.

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/5528#issuecomment-849097983

PR #5531, opened 2021-05-26
Added new test case to OfferTest

ashah1220:
Added a test case to check if resetState() works.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionSha256Sum=11657af6356b7587bfb37287b5992e94a9686d5c8a0a1b60b87b9928a2decde5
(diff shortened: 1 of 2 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionSha256Sum=11657af6356b7587bfb37287b5992e94a9686d5c8a0a1b60b87b9928a2decde5
```
ripcurlx, line 3: Why did you try to remove the hash in this commit?

ripcurlx (changes requested):
NACK - please see my comment

gradle/wrapper/gradle-wrapper.properties
ashah1220: The build fails for me unless I remove line 3 in the gradle properties.

ripcurlx: Could you please post the log output in that case?

PR #5534, opened 2021-05-26
Manually create BSQ outputs

### LINKED ISSUE #5533: [v1.6.5] Change output below dust limit is burned when custom inputs are used
### Description

If you try to send BSQ from the wallet using custom inputs and the change output is below the dust limit, it is burnt and deducted from the senders wallet.

#### Version

v1.6.5 (probably also in older versions)

### Steps to reproduce

- Go to BSQ Wallet > Send
- Select an input (e.g. 999.94 BSQ)
- Set receiver address and amount to e.g. 995 BSQ
- Click `Send BSQ funds`

### Expected behaviour

Warning is shown that an amount below dust limit will be burnt and more inputs have to be added.

### Actual behaviour

Transactions is created and published burning the BSQ amount below the dust limit and it is displayed as Trading Fee transaction in the senders and receivers wallet.

### Screenshots

Bildschirmfoto 2021-05-26 um 11 46 31
Bildschirmfoto 2021-05-26 um 11 46 38

sqrrm:
Avoid relying on bitcoinj completeTx for creation of BSQ change
outputs as dust BSQ would be used for mining fee, thus burning
the BSQ.

Fixes #5533

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/BsqWalletService.java
@@ -52,7 +52,6 @@ import org.bitcoinj.core.TransactionOutput;
 import org.bitcoinj.script.Script;
 import org.bitcoinj.script.ScriptException;
 import org.bitcoinj.wallet.CoinSelection;
-import org.bitcoinj.wallet.CoinSelector;
 import org.bitcoinj.wallet.SendRequest;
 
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
ACK - Tested on Regtest BSQ and non-BSQ (BTC) withdrawal from BSQ wallet with manual and automatic inputs. Does work as expected now :+1:

PR #5535, opened 2021-05-26
Fix security deposit in duplicate offer when minimum deposit is used

BtcContributor:
This should fix the security deposit when a user duplicates an offer with a minimum security deposit set.

Please test that on Mainnet as I saw some differences between Regtest, but on Mainnet should be ok.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/offer/MutableOfferDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/MutableOfferDataModel.java
@@ -106,7 +106,6 @@ public abstract class MutableOfferDataModel extends OfferDataModel implements Bs
     final String shortOfferId;
     private final AccountAgeWitnessService accountAgeWitnessService;
     private final FeeService feeService;
-    private final CoinFormatter btcFormatter;
     private final Navigation navigation;
     private final String offerId;
     private final BalanceListener btcBalanceListener;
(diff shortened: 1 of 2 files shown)


desktop/src/main/java/bisq/desktop/main/offer/MutableOfferDataModel.java
```diff
@@ -121,10 +120,13 @@
     protected final ObjectProperty<Price> price = new SimpleObjectProperty<>();
     protected final ObjectProperty<Volume> volume = new SimpleObjectProperty<>();
     protected final ObjectProperty<Volume> minVolume = new SimpleObjectProperty<>();
+    protected final CoinFormatter btcFormatter;
 
     // Percentage value of buyer security deposit. E.g. 0.01 means 1% of trade amount
     protected final DoubleProperty buyerSecurityDeposit = new SimpleDoubleProperty();
 
+    protected final StringProperty buyerSecurityDepositInBTC = new SimpleStringProperty();
```
ripcurlx, line 128: This property is set, but never used. How can this fix the issue with the security deposit?

ghost: `buyerSecurityDepositInBTC` is already used by a listener in MutableOfferView.java.  Its odd that github diff is showing the line above as "new code", as I think @BtcContributor only added the keyword `protected` here.  So he's granting access for the property to be set from the duplicate offer data model.  :+1:

ghost: NACK. -> upon testing this PR it does not fix the case of duplicate on offers that had 0.006 security deposit.

PR #5537, opened 2021-05-27
Remove Bitpay as a pricenode data provider

Emzy:
Volume on Bitpay is super low and the spread is really high.
That results in a far to low or to high spot price.

Screen Shot 2021-05-27 at 12 18 04


ripcurlx (approved):
utACK

PR #5539, opened 2021-05-27
Fix setting of buyer security deposit

ripcurlx:
The existing code failed if an offer was duplicated that had
a different minimum security deposit set. It is changed so that
it makes sure that the it never falls below the minimum buyer
security deposit and never exceeds the maximum security deposit
percentage.


ripcurlx:
@jmacxx It would be great if you could try it out for your use case. I think this code changes should handle those cases as well.

ghost:
tested, ACK.

sqrrm (approved):

PR #5540, opened 2021-05-28
Revert setting of deposit when cloning an offer

ripcurlx:
As it caused too many issues during testing I'll revert this change for this release.


sqrrm (approved):
utACK

Issue #5526, opened 2021-05-25
Unreliable network delivery causing trading issues.


ghost:
_Opening this to give the problem more visibility and hoping that more eyes can lead to a fix._

This is an attempt to describe a long-standing issue with messaging in Bisq.  Rare, but frequent enough to be an almost daily occurrence in the network.  Some messages supposedly arrive, but don't get seen by the recipient Bisq application layer.  Causing such problems as failed trades and miscommunication between users.

[Update - just noticed #1172 from 2018 describes the same problem.  Seems that many people investigated but it was never resolved.  Looks like a problem with TOR.]

---

Bisq uses TCP for networking which is a "reliable" protocol.  [However, what this actually means is](https://stackoverflow.com/questions/14303198/is-tcp-100-reliable#14303258):

> TCP is reliable in the sense that messages sent will either be received as they were sent (same order, same contents); assuming that the built-in checksum does not happen to fail, which it may, since it is only 16 bits **OR not be received at all**

The small percentage of messages **not received at all** are causing these issues.

---

### Example 1:

- Mediator sends a chat message to Trader, status=Message arrival confirmed by receiver.  All good.
![image](https://user-images.githubusercontent.com/47253594/119432319-c0975380-bcd9-11eb-84bd-2d042e07462a.png)
- Trader sends a chat message to Mediator, status=Message arrived at receiver.
![image](https://user-images.githubusercontent.com/47253594/119432441-f63c3c80-bcd9-11eb-9bea-f7458f52c23e.png)
- Looks good from the sender's side, **however, Mediator never actually received the message**.  

---

### Example 2:

- Alice clicks "Payment started".  (status=`Message arrived at receiver`)
- No `ack` message is received back so Alice is prompted again to click "Payment started".
- And again and again until they ask in support why the button is not working.
- **Bob never receives the Payment started message**.

This used to be known as `Seller cannot confirm` (#4748) since they did not receive Payment started, however since #5401 the burden is shifted to the buyer who is told to try again.  Its still causing a problem but at least there's an option to retry now, whereas before the only option was mediation.  What can be learned from this is that retrying the send message, and even rebooting the sender does not help.  Perhaps rebooting the recipent side would help, but no one has come forth to report that as successful yet.

---

@leo816 has experienced it multiple times and reported both examples above.

`Seller-cannot-confirm` is a high cause of mediation tickets, e.g. 

[...]

PR #5551, opened 2021-06-02
Improve Chat message reliability

ghost:
As described in #5526, there are cases where network messages get the arrived status but no ACK and the receiver never actually saw the message.  This can be cause of confusion because the "arrived" status is generally used by the sender as indication that all is good.  The ACK is generally ignored.

To remedy this for Chat messages, this PR displays the sent message as "In transit" until the ACK is received back, or it times out and displays status describing the problem and asking the user to try again.  The timeout is 5 seconds.  An ACK can still be received after 5 seconds, in which case the status will update to show that the message arrived.

### Screenshots

Sent the message, waiting for the ACK:
image

ACK received back:
image

No ACK received back:
image


ripcurlx:
utACK from code perspective. @m52go Could you please review the text changes? Thanks!

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1137,8 +1137,10 @@ support.closeTicket=Close ticket
 support.attachments=Attachments:
 support.savedInMailbox=Message saved in receiver's mailbox
 support.arrived=Message arrived at receiver
+support.transient=Message is on its way to receiver
 support.acknowledged=Message arrival confirmed by receiver
 support.error=Receiver could not process message. Error: {0}
+support.errorTimeout=Timed-out waiting for ACK, try again
```
m52go: Is this suggestion to "try again" directed at the user?

If so, something like this might be more clear:

```suggestion
support.errorTimeout=timed out. Try sending message again.
```

ripcurlx (approved):
utACK

PR #5552, opened 2021-06-02
Remind user when their DAO state needs to be refreshed

### LINKED ISSUE #5329: BSQ trading fee calculated incorrectly

### LINKED ISSUE #5393: DAO out of sync causes Bisq fee to be validated incorrectly
Since 1.6.0, Bisq has had [this feature](https://github.com/bisq-network/bisq/issues/5119) to validate trading fees paid.  With this model, Bisq trading partners check each other's transactions thus ensuring fees are paid (which is what keeps Bisq running).  

However it is possible for a Bisq node's DAO to be out of sync and thus have a differing view on what the sufficient fee should be.  

For example user Alice's node expected trading partner Bob to have paid a fee of 2.11 BSQ; in fact Bob had paid the correct fee of 1.82 BSQ.  In this case Alice's node had an out of sync DAO which told them that Bob should have used a fee rate from two months ago which is higher than today's fee rate.  Bisq complains to Alice that her counterparty did not pay enough fees and that a support ticket should be opened.

There are already [leniency checks](https://github.com/bisq-network/bisq/pull/5347) built onto the fee validation to allow historical fee rates.  But in this case the appropriate fee rate was unknown to Alice since her DAO was stuck, **and this was only evident by looking at DAO -> Dao Status screen**.

The solution for Alice was to re-sync the DAO.

---

Ideally we would not have the situation where the DAO can silently get stuck.

Related issue: https://github.com/bisq-network/bisq/issues/5329

ghost:
As described in #5393, having a DAO state that is in conflict can cause issues and unnecessary mediation cases.

Unfortunately, users are not notified when their DAO state is in conflict with seednodes, it is only shown by checking DAO -> Network Monitor -> Dao State.

This PR attempts to rectify that by checking the DAO state whenever Bisq has completed the block sync.  If the DAO is reporting a conflict then a popup will be shown to the user asking them to resync .  To avoid annoyance, this popup will only be shown at most once per session (and then only if a resync is needed).

This PR only checks the overall DAO state, not the voting or proposals state.  This is intentional, since proposal and voting states are only applicable to a small subset of users whereas the goal of this PR is to prevent problems in trading.

### Screenshot

image

&nbsp;

Fixes #5329 
Fixes #5393

### DIFF
--- a/core/src/main/java/bisq/core/dao/DaoFacade.java
+++ b/core/src/main/java/bisq/core/dao/DaoFacade.java
@@ -52,6 +52,7 @@ import bisq.core.dao.governance.proposal.reimbursement.ReimbursementConsensus;
 import bisq.core.dao.governance.proposal.reimbursement.ReimbursementProposalFactory;
 import bisq.core.dao.governance.proposal.removeAsset.RemoveAssetProposalFactory;
 import bisq.core.dao.governance.proposal.role.RoleProposalFactory;
+import bisq.core.dao.monitoring.DaoStateMonitoringService;
 import bisq.core.dao.state.DaoStateListener;
 import bisq.core.dao.state.DaoStateService;
 import bisq.core.dao.state.model.blockchain.BaseTx;
@@ -124,6 +125,7 @@ public class DaoFacade implements DaoSetupService {
     private final BallotListPresentation ballotListPresentation;
     private final MyProposalListService myProposalListService;
     private final DaoStateService daoStateService;
+    private final DaoStateMonitoringService daoStateMonitoringService;
     private final PeriodService periodService;
     private final MyBlindVoteListService myBlindVoteListService;
     private final MyVoteListService myVoteListService;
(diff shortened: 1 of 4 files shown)


ripcurlx:
@jmacxx Maybe we could make it even easier if we offer the user a button to jump directly to the DAO -> Network Monitor screen in the popup as well. WDYT?

ghost:
> @jmacxx Maybe we could make it even easier if we offer the user a button to jump directly to the DAO -> Network Monitor screen in the popup as well. WDYT?

@ripcurlx I did the needful, looks better.  Further testing revealed a bug, which has also been resolved.  :sweat_smile:

ripcurlx (approved):
utACK

Issue #5509, opened 2021-05-18
When chosen fee provider is down, Bisq should failover to a new one


ghost:
The Bisq FeeService should automatically select a new provider when the chosen RPC endpoint is down.  It does not, simply keeps querying the same one and writing errors to the log.  Users are left using stale fee rate (or the default: 50 sat/vB if the server was down at Bisq startup).

PriceFeedService has an automatic failover which chooses a new provider.

PriceFeedService and FeeService could be redesigned to share the same provider and failover mechanism.  This would have the added benefit of opening one less connection (currently each Bisq client simultaneously connects to two separate PriceNodes).

ghost:
@wiz suggestion: add mempool fees to the `getAllMarketPrices` endpoint.  This will enable a single request from Bisq clients to process market prices and mempool fees.  Benefits: better network efficiency & more frequent mempool fee updates (every minute instead of every 5 minutes).  PR to follow.

PR #5553, opened 2021-06-04
Add mempool fees to PriceServer getAllMarketPrices endpoint

ghost:
As mentioned in #5509, Bisq client has two connections to price nodes: FeeService and PriceFeedService.  We can optimize Bisq to use PriceFeedService and obtain both sets of information together which provides the following benefits:

- Bisq will use one less network connection and make fewer network API calls,
- corrects the lack of failback mechanism in FeeService,
- increases the frequency of mempool fee updates 5-fold, at no cost.

This change is backwards compatible after PriceNodes are updated:

- old clients can still process `getAllMarketPrices` and `getFees` as before.
- new clients can process the additional mempool fee information returned in the `getAllMarketPrices` data set.

---

This change adds two fields: `bitcoinFeesTs` and `bitcoinFeeInfo` to `getAllMarketPrices`:

image

This PR only modifies the PriceNode.  The idea is to give time to get the PriceNodes upgraded before the Bisq client is rolled out (as noted already, the PriceNode upgrade is compatible with existing Bisq clients).  

A separate PR will be created for Bisq clients to make use of this PriceNode optimization.

### DIFF
--- a/common/src/main/java/bisq/common/config/Config.java
+++ b/common/src/main/java/bisq/common/config/Config.java
@@ -121,9 +121,11 @@ public class Config {
     public static final String API_PORT = "apiPort";
     public static final String PREVENT_PERIODIC_SHUTDOWN_AT_SEED_NODE = "preventPeriodicShutdownAtSeedNode";
     public static final String REPUBLISH_MAILBOX_ENTRIES = "republishMailboxEntries";
+    public static final String LEGACY_FEE_DATAMAP = "dataMap";
     public static final String BTC_TX_FEE = "btcTxFee";
     public static final String BTC_MIN_TX_FEE = "btcMinTxFee";
     public static final String BTC_FEES_TS = "bitcoinFeesTs";
+    public static final String BTC_FEE_INFO = "bitcoinFeeInfo";
     public static final String BYPASS_MEMPOOL_VALIDATION = "bypassMempoolValidation";
 
     // Default values for certain options
--- a/pricenode/src/main/java/bisq/price/mining/FeeRateService.java
+++ b/pricenode/src/main/java/bisq/price/mining/FeeRateService.java
@@ -36,7 +36,7 @@ import org.slf4j.LoggerFactory;
  * High-level mining {@link FeeRate} operations.
  */
 @Service
-class FeeRateService {
+public class FeeRateService {
 
     private final List<FeeRateProvider> providers;
     protected final Logger log = LoggerFactory.getLogger(this.getClass());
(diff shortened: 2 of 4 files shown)


common/src/main/java/bisq/common/config/Config.java
```diff
@@ -124,6 +124,7 @@
     public static final String BTC_TX_FEE = "btcTxFee";
     public static final String BTC_MIN_TX_FEE = "btcMinTxFee";
     public static final String BTC_FEES_TS = "bitcoinFeesTs";
+    public static final String BTC_FEES_INFO = "bitcoinFeeInfo";
```
kaiwolfram:
```suggestion
    public static final String BTC_FEE_INFO = "bitcoinFeeInfo";
```

pricenode/src/main/java/bisq/price/spot/ExchangeRateController.java
```diff
@@ -28,13 +31,29 @@
...
     @GetMapping(path = "/getAllMarketPrices")
     public Map<String, Object> getAllMarketPrices() {
-        return exchangeRateService.getAllMarketPrices();
+        Map<String, Object> retVal = exchangeRateService.getAllMarketPrices();
+
+        // add the fee info to results
+        Map<String, Object> feeInfo = feeRateService.getFees();
+        for (String key : feeInfo.keySet()) {
+            retVal.put(translateFieldName(key), feeInfo.get(key));
+        }
```
kaiwolfram: It's better to either iterate over `feeRateService.getFees().entrySet()` to avoid value lookups for each key and to make your intent clearer or use `feeRateService.getFees().forEach((k, v) -> ...)`

pricenode/src/main/java/bisq/price/mining/FeeRateService.java
```diff
@@ -36,6 +36,7 @@
  * High-level mining {@link FeeRate} operations.
  */
 @Service
+public
 class FeeRateService {
```
kaiwolfram:
```suggestion
public class FeeRateService {
```

pricenode/src/main/java/bisq/price/spot/ExchangeRateController.java
```diff
@@ -28,13 +31,29 @@
...
+
+        return retVal;
+    }
+
+    static String translateFieldName(String name) {
+        if (name.equals("dataMap"))
```
kaiwolfram: `"dataMap"` should be a variable, as it is used across multiple methods

ghost:
Thanks @KaiWitt changes applied.

ripcurlx (approved):
utACK

PR #5556, opened 2021-06-07
Remove Chase QuickPay as a payment method

### LINKED ISSUE #5547: Remove Chase QuickPay as a payment method
### Description

Chase Quickpay is a current payment method in Bisq.

Zelle is also a current payment method in Bisq.

Some users have recently asked questions about the differences between using Chase Quickpay (1 day trade window) and Zelle (4 day trade window).

Looking at the [Chase Quickpay website](https://www.chase.com/personal/quickpay-match/login-quickpay) it has now become [Zelle](https://www.zellepay.com/) 

From both websites most payments arrive instantly but may take up to 3 days.

Therefore, I propose to remove Chase Quickpay as a payment method.

Currently, at the time of writing this, there are 52 Zelle offers and 1 Chase Quickpay offer. Therefore, it appears this would not affect the current offer book too much.

#### Version

v1.6.5

### Steps to reproduce

Both Zelle and Chase Quickpay payment methods can be set up.

### Expected behaviour

Chase Quickpay payment methods should be removed. Existing users can set up a Zelle payment method if they have not already and edit their offers to Zelle or another payment method.

#### Additional info

I have not used either method so let me know if I have missed anything.

andyheko:
Add a deprecated tag to chase QuickPay, and remove it from the list of available payment methods.

resolved #5547

### DIFF
--- a/core/src/main/java/bisq/core/payment/ChaseQuickPayAccount.java
+++ b/core/src/main/java/bisq/core/payment/ChaseQuickPayAccount.java
@@ -24,6 +24,9 @@ import bisq.core.payment.payload.PaymentMethod;
 
 import lombok.EqualsAndHashCode;
 
+// Removed due to QuickPay becoming Zelle
+// Cannot be deleted as it would break old trade history entries
+@Deprecated
 @EqualsAndHashCode(callSuper = true)
 public final class ChaseQuickPayAccount extends PaymentAccount {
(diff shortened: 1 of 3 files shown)


ripcurlx (approved):
utACK - code changes are looking fine, so it shouldn't be possible anymore to create new Chase QuickPay accounts with this change.

PR #5562, opened 2021-06-10
Disable payment received button after mediation has been accepted

ghost:
If the seller has accepted a mediation proposed payout, disable the confirm payment received button.
Intended as a failsafe so users do not accidentally release while waiting for the mediation approval process to finish.

Requested by @leo816

Related to, but does not contradict #4602.


ripcurlx (approved):
utACK

PR #5570, opened 2021-06-12
Remove create CHASE_QUICK_PAY apitest case

ghubstan:
And remove commented core test.

PaymentMethod `CHASE_QUICK_PAY` is no[ longer supported](https://github.com/bisq-network/bisq/pull/5556).

This is the 1st in a short series of PRs to add an `editoffer` api method.


ripcurlx (approved):
utACK

PR #5582, opened 2021-06-22
PriceNode: exclude currencies via config

### LINKED ISSUE #5581: Users in Argentina and Lebanon unable to use Bisq
### Description

Users in Argentina and Lebanon effectively unable to use Bisq as price controls in their countries mean the difference between the official and unofficial exchange rate exceeds >50%.

Looking at [How to run a Bisq price relay node](https://docs.bisq.network/exchange/howto/run-price-relay-node.html) it would appear Bisq uses the [BitcoinAverage](https://bitcoinaverage.com/) API for fiat prices. The prices on Bitcoin average are the official currency rates.

As both Argentina and Lebanon have foreign currency controls in place. The difference between the official and unofficial rate is 60% for Argentina and 280% for Lebanon.

Bisq limits offers to buy/sell BTC to be a max deviation of 50% from the market price. Unfortunately as this market price is the official rate it effectively prevents users in these countries trading on Bisq.

I think it is important for Bisq to enable users in all countries to be able to trade on Bisq. I think it is especially important to allow users to trade on Bisq where there are current currency controls in place and/or their currency in a state of crisis.

#### Version

1.6.5

### Steps to reproduce

Check the price feeds on the Bisq app for Argentina and Lebanon they use the official rates.

Rate of USD in Argentina / Argentine Peso 
https://bluedollar.net/
Difference between offical and unofficial is 60%

Rate of USD in Lebanon / Lebanese Pound
https://lebaneselira.org/
Difference between official and unofficial is 280%

### Expected behaviour

Remove the max deviation of 50% for currencies with capital controls in place i.e a large percentage difference between official and unofficial rates. 

This would allow users in those countries to place buy and sell offers that are economic and likely to be taken.

### Actual behaviour

Users in Argentina and Lebanon unable to use Bisq due to the difference in official/unofficial exchange rates being uneconomical

ghost:
This allows the PriceNode to configure certain currencies as not quoted, i.e. will not be delivered in its service API.  By doing so, those markets can self-discover the price the same way as is already done in BSQ/BTC market.

Fixes #5581 

### Config

Added to `application.properties`

    bisq.price.fiatcurrency.excluded=LBP,ARS
    bisq.price.cryptocurrency.excluded=

### Code changes: 

- get Environment object from Spring framework.
- parse & validate any currencies specified in the `application.properties` file.
- exclude currencies from the exchange supported fiat currency / crypto currency lists.
- log details of currencies excluded via configuration.
- add tests for the above.

### DIFF
--- a/pricenode/src/main/java/bisq/price/spot/ExchangeRateProvider.java
+++ b/pricenode/src/main/java/bisq/price/spot/ExchangeRateProvider.java
@@ -76,12 +76,12 @@ public abstract class ExchangeRateProvider extends PriceProvider<Set<ExchangeRat
     }
 
     public Set<String> getSupportedFiatCurrencies() {
-        if (SUPPORTED_FIAT_CURRENCIES.size() == 0) {         // one-time initialization
+        if (SUPPORTED_FIAT_CURRENCIES.isEmpty()) {         // one-time initialization
             List<String> excludedFiatCurrencies =
                     Arrays.asList(env.getProperty("bisq.price.fiatcurrency.excluded", "")
                             .toUpperCase().trim().split("\\s*,\\s*"));
             String validatedExclusionList = excludedFiatCurrencies.stream()
-                    .filter(ccy -> ccy.length() > 0)
+                    .filter(ccy -> !ccy.isEmpty())
                     .filter(CurrencyUtil::isFiatCurrency)
                     .collect(Collectors.toList()).toString();
             SUPPORTED_FIAT_CURRENCIES = CurrencyUtil.getAllSortedFiatCurrencies().stream()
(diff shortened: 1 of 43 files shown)


pricenode/src/main/java/bisq/price/spot/ExchangeRateProvider.java
```diff
@@ -59,21 +62,55 @@
...
-    public ExchangeRateProvider(String name, String prefix, Duration refreshInterval) {
+    public ExchangeRateProvider(Environment env, String name, String prefix, Duration refreshInterval) {
         super(refreshInterval);
         this.name = name;
         this.prefix = prefix;
+        this.env = env;
+    }
+
+    public Set<String> getSupportedFiatCurrencies() {
+        if (SUPPORTED_FIAT_CURRENCIES.size() == 0) {         // one-time initialization
```
kaiwolfram: `.isEmpty()` is more expressive, same in line 84 `.filter(ccy -> !ccy.isEmpty())` and the crypto currency method

ripcurlx:
@jmacxx It does work as described on RegTest, but what is the difference to using fixed price right now already?

Current state:
Bildschirmfoto 2021-07-13 um 10 02 28

Filtered out state:
Bildschirmfoto 2021-07-13 um 09 58 12

So the only difference is that as a user I have the price of the last trade as indicator what someone was willing to pay.

ghost:
When a market is quoted by a price feed, users are limited to trading at prices within 50% of the reference quote.  If we disable the quote for LBP & ARS the user can trade using the street price.

ripcurlx (approved):
ACK - Tested it on Regtest. The changes for these market should be coordinated with @pazza83 I guess.

pazza83:
I will test it with the new release. I also think it is worth promoting to potential users in these markets.

PR #5593, opened 2021-06-30
Update relay-node address

devinbileck:
With Tor v2 addresses being deprecated soon, now is a good time to
update the relay-node used for mobile notifications with a new v3 onion
address.


devinbileck:
I have kept this as draft for now until I can confirm notifications are working for Android. I have confirmed iOS is working.
This is a new server, so need to check it is functioning correctly.

devinbileck:
@huey735 confirmed that Android notifications are working.

devinbileck:
This should now be ready to go live.
Note, I will keep the old server running likely for at least the next couple months until all users have updated to a newer Bisq with this change.

ripcurlx (approved):
utACK

PR #5599, opened 2021-07-02
Fix bug in chat message status display

ghost:
Reported by @huey735 there is a display bug when chat messages are stored in mailbox awaiting peer to come online.
The bug was introduced by #5551 
The fix is to consider both acknowledged and stored in mailbox statuses when checking if the message timed out.

To reproduce:

- Alice sends chat message to Bob (who is offline).
- Alice's chat window shows `Message saved in receiver's mailbox`.  
- Alice closes chat window, and reopens it again.
- Alice's chat window now says `Error: timed out.  Try sending message again`.   << This is the bug.

Expected behavior:

- Alice sends chat message to Bob (who is offline).
- Alice's chat window shows `Message saved in receiver's mailbox`.  
- Alice closes chat window, and reopens it again.
- Alice's chat window shows `Message saved in receiver's mailbox`.  

If Bob were to come online Alice's window would update to `Message arrival confirmed by receiver`.

### DIFF
--- a/core/src/main/java/bisq/core/support/messages/ChatMessage.java
+++ b/core/src/main/java/bisq/core/support/messages/ChatMessage.java
@@ -302,7 +302,7 @@ public final class ChatMessage extends SupportMessage {
     // each chat message notifies the user if an ACK is not received in time
     public void startAckTimer() {
         UserThread.runAfter(() -> {
-            if (!this.getAcknowledgedProperty().get()) {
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
utACK

ManoloSFC:
I'm having that problem, but I'm not getting any answer from the other part after sending multiple messages, so I'm guessing that is not only a problem of Bisq reporting wrongly the status of the message but the message system is actually broken and messages aren't being sent through.

PR #5602, opened 2021-07-03
Fix custom withdrawal fee bug

### LINKED ISSUE #5575: Transaction was made with 20 sat/vb instead of 2 sat/vb
### Description
Yesterday I wanted to send all of my funds to a cold storage. I chose 1 sat/vb fee and a dialog told me the minimum was 2 sat/vb (no idea why, since mempool was almost empty).  
I then chose 2 sat/vb and continued. IIRC, the confirmation dialog didn't show the fee in sat/vb, only the total fee (if that is right, then I would suggest to add the sat/vb info as it would have saved me).
I saw ~13K sats as fee and thought that was alright, since the transaction had many inputs, but it turns out that was actually a 20 sat/vb fee.

I remember going back to the previous screen and saw 2 sat/vb was still there.

I am not sure this report will be useful, as it might be difficult to reproduce and happened only once (no balance left to try again).

#### Version

1.6.5 downloaded from Arch User Repository (AUR).
`bisq.common.app.Version: Version{VERSION=1.6.5, P2P_NETWORK_VERSION=1, LOCAL_DB_VERSION=1, TRADE_PROTOCOL_VERSION=3, BASE_CURRENCY_NETWORK=0, getP2PNetworkId()=10}`.

### Steps to reproduce

I don't think it can realiably be reproduced.

### Expected behaviour

Create the transaction with the correct fee: 2 sat/vb.

### Actual behaviour

Transaction was created with a different (10x) fee: 20 sat/vb.

#### Device or machine

Arch Linux; Linux 5.12.10.

#### Additional info

Logs don't seem to have anything useful, just thousands of exceptions about some BSQ reimbursement.
Edit: realised there are many log files. Any particular string I could grep?

### LINKED ISSUE #5605: Confusing Withdrawal Pop-up
### Description

The dot at the end of the receivng address showned in the “withdrawal request” pop-up windows is confusing. The dot should be removed as newcomers to Bitcoin could be confused. I’m not a newcomer to Bitcoin, but was still confused although I know there’s not dot in a bitcoin address. However, when dealing with real money, I remain suspicious on anything I find not conventional. Indeed, no dot shown at the end of the address above (From address), but a dot below for the receiving address. This alerted me in making a test transaction to make sure I’ll get the satoshis on the receiving end. So, the dot should be removed.

#### Version

1.6.5

### Steps to reproduce

Withdraw bitcoins to an external wallet.

### Expected behaviour

There should not be a dot at the end of the recipient's address

### Actual behaviour

There is a dot at the end of the recipient's address

### Screenshots

Formatting_Issue

#### Device or machine

MacOS Mojave 10.14.6

ghost:
The wallet code was by default using the value on Setting/Preferences instead of the value chosen on the Withdrawal screen.

Fixes #5575

- Pass the chosen fee rate to the wallet when doing a withdraw.
- Calculate sats/vb in the summary popup instead of showing the value of the input text field.

 
 
### Steps to reproduce the issue (regtest)

- Prerequisite: `Use custom withdrawal fee` is unchecked in Settings/Preferences.
- Go to Funds/Receive funds select a receiving address and copy to clipboard.
- Go to Funds/Send funds, paste your address into the withdraw to address field.
- Take note of the default suggested transaction fee (in this case **50 sats/vb**).
- Click use custom value for fee, enter 2 sats/vbyte or the lowest value permissible.
- Click Withdraw Selected.  Note the mining fee and the sats/vb reported alongside it.
- Click yes to withdraw.
- Now go to the log where it details the transaction, note the weight e.g. 110 virtual bytes, and note the fee paid e.g. 0.00055 BTC ( = 55000 sats).
- Divide the fee paid by virtual bytes to get the actual fee per vb paid, in this case 55000 / 110 = **50 sats/vb**.
- The fee used is the value originally shown in the fee text box, instead of the one the user entered.

---

[update] Refactored the fix to pass the chosen fee rate to the wallet instead of setting the custom withdrawal flag in user preferences.  I think this approach is cleaner and more in line with @BtcContributor original idea.

Also: fixes #5605

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -124,7 +124,7 @@ shared.noDateAvailable=No date available
 shared.noDetailsAvailable=No details available
 shared.notUsedYet=Not used yet
 shared.date=Date
-shared.sendFundsDetailsWithFee=Sending: {0}\nFrom address: {1}\nTo receiving address: {2}.\nRequired mining fee is: {3} ({4} satoshis/vbyte)\nTransaction vsize: {5} vKb\n\nThe recipient will receive: {6}\n\nAre you sure you want to withdraw this amount?
+shared.sendFundsDetailsWithFee=Sending: {0}\nFrom address: {1}\nTo receiving address: {2}\nRequired mining fee is: {3} ({4} satoshis/vbyte)\nTransaction vsize: {5} vKb\n\nThe recipient will receive: {6}\n\nAre you sure you want to withdraw this amount?
 # suppress inspection "TrailingSpacesInProperty"
 shared.sendFundsDetailsDust=Bisq detected that this transaction would create a change output which is below the minimum dust threshold (and therefore not allowed by Bitcoin consensus rules). Instead, this dust ({0} satoshi{1}) will be added to the mining fee.\n\n\n
 shared.copyToClipboard=Copy to clipboard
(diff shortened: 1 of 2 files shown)


ripcurlx (approved):
utACK

MatthewCroughan:
https://github.com/bisq-network/bisq/discussions/5601

PR #5608, opened 2021-07-06
Json dump for Dao state refactoring

chimp1984:
- Write chainHeight
- Write on new blocks or on parse complete from last persisted block height
- Do not delete json dirs at start
- Write txs and txos from block/s to write
-  Add reorg handling to json dump
- Do not load daostate in case the DAO is disabled (not related to json dump aspect)

In case of a reorg we pass the block height of the last persisted snapshot
and delete all blocks and related tx/txos from that block height on.
In case we dont have a snapshot we delete the whole json dir.

It is not tested much yet. 
Will require adoption for the explorer nodes by @softsimon - should be either deployed to a branch or waited for merge to master until @softsimon signals readiness.

### DIFF
--- a/common/src/main/java/bisq/common/file/JsonFileManager.java
+++ b/common/src/main/java/bisq/common/file/JsonFileManager.java
@@ -22,10 +22,13 @@ import bisq.common.util.Utilities;
 import java.nio.file.Paths;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.PrintWriter;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
+import java.util.Scanner;
 import java.util.concurrent.ThreadPoolExecutor;
 
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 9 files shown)


ripcurlx:
@softsimon Are you aware about this open change?

common/src/main/java/bisq/common/file/JsonFileManager.java
```diff
@@ -76,10 +79,26 @@ public void shutDown() {
...
+            return Optional.empty();
+        }
+        StringBuilder json = new StringBuilder();
+        try (Scanner scanner = new Scanner(jsonFile)) {
+            while (scanner.hasNext()) {
+                json.append(scanner.next());
+            }
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
+        }
```
ghubstan, line 97: Shouldn't line 96 be `log.error(e.toString());`  if not a thrown exception?  Stack traces printed to std.err can be out of sync with the logger's buffer flushing -- it can make the timing of the stack trace in the bisq.log file a bit confusing.

core/src/main/java/bisq/core/dao/node/explorer/ExportJsonFilesService.java
```diff
@@ -113,57 +102,86 @@ public void start() {
...
+            chainHeightFileManager.write(String.valueOf(chainHeight), "chainHeight");
+            jsonBlockList.forEach(jsonBlock -> blocksFileManager.write(Utilities.objectToJson(jsonBlock), jsonBlock.getHeight() + "_" + jsonBlock.getHash()));
+            jsonTxs.forEach(jsonTx -> txFileManager.write(Utilities.objectToJson(jsonTx), jsonTx.getId()));
+            jsonTxOutputs.forEach(jsonTxOutput -> txOutputFileManager.write(Utilities.objectToJson(jsonTxOutput), jsonTxOutput.getId()));
+            return null;
+        });
+
+        Futures.addCallback(future, Utilities.failureCallback(throwable -> {
+            log.error(throwable.toString());
+            throwable.printStackTrace();
```
ghubstan, line 220: Isn't this t.printStackTrace() redundant?  You just logged the error, with stack trace.

ghubstan:
I verified json dumps are happening while offers are being created in the api test harness.
Pending file conflict resolution, a pseudo `Tested ACK` from me.

Stepped through json dump dirs created:               OK
Tested chainHeight.json is updated/incremented:	OK
Tested block/<blk-#>_hash.json is created:		OK
Tested tx/<tx-id>.json is created:				OK
Tested txo/<tx-id>.json is created:				OK

Json file contents look OK to me.

chimp1984:
Close as totally outdated...

PR #5609, opened 2021-07-06
Optimize DaoState snapshot behaviour

chimp1984:
- Run persistence call in thread instead of user thread (serialisation is very slow and had blocked user thread)
- Create new snapshot only after persistence is completed to avoid to have 3 daoState objects in memory
- Set DaoState in store to null to let gc remove the old reference (was left there before so we had 3 instances of daoStates in memory)

This change results in about 70 mb less memory usage (10% of about 700 mb normal memory usage).
With vm arguments:
```
-XX:+UseG1GC
-XX:MaxHeapFreeRatio=10
-XX:MinHeapFreeRatio=5
-XX:+UseStringDeduplication
```
memory reduction is about 30% and total memory gets released (not the case with defaults). In my tests I got about 496 MB instead of about 700 with the master version and no vm arguments.

During parsing or other heavier load the total memory will go up again to about 2GB but gets released after some GC rounds once the load is reduced again.

As those gc settings have impact on performance we should observe it first to see if it works well or continue to fine tune it (or add/find other options).

To test the difference one need to wait either enough blocks to get a snapshot (every 20 blocks but I guess one need to wait > 40 blocks to see the full effect) or to hack the code a bit to create the daostate clones at each block (thats how I got my numbers - but I will let it running in both versions, master and the PR over night to see if the improvement really gets confirmed).

I also added that debugging code to the Profiler class to trigger gc fast and to get fast updates about memory usage:
```
 static {
        UserThread.runPeriodically(() -> {
            log.error("UsedMemoryInMB={}, totalMemory={}, freeMemory={}, maxMemory={}",
                    Profiler.getUsedMemoryInMB(),
                    Utilities.readableFileSize(Runtime.getRuntime().totalMemory()),
                    Utilities.readableFileSize(Runtime.getRuntime().freeMemory()),
                    Utilities.readableFileSize(Runtime.getRuntime().maxMemory())
            );
        }, 2);

        UserThread.runPeriodically(() -> {
            System.gc();
        }, 30);
    }
```

I used https://heaphero.io/ for a memory analysis. Here are the results:
[heaphero_results_PR.pdf](https://github.com/bisq-network/bisq/files/6772730/heaphero_results_PR.pdf)
[heap_hero_results_master.pdf](https://github.com/bisq-network/bisq/files/6772731/heap_hero_results_master.pdf)

It shows clearly the lower memory use for the daostate and for strings. Most strings are also from the daostate data (address, pubkeyashex, asm,...)

### DIFF
--- a/core/src/main/java/bisq/core/dao/state/storage/DaoStateStorageService.java
+++ b/core/src/main/java/bisq/core/dao/state/storage/DaoStateStorageService.java
@@ -92,9 +92,8 @@ public class DaoStateStorageService extends StoreService<DaoStateStore> {
         new Thread(() -> {
             Thread.currentThread().setName("Serialize and write DaoState");
             persistenceManager.persistNow(() -> {
-                // After we have written to disk we remove the reference to the daoState. We have in the meantime
-                // already cloned the next daoState and if we would not release the reference we would have 3 times
-                // the daoState data in memory.
+                // After we have written to disk we remove the the daoState in the store to avoid that it stays in
+                // memory there until the next persist call.
                 store.setDaoState(null);
 
                 completeHandler.run();
--- a/core/src/main/java/bisq/core/dao/state/storage/DaoStateStorageService.java
+++ b/core/src/main/java/bisq/core/dao/state/storage/DaoStateStorageService.java
@@ -97,8 +97,6 @@ public class DaoStateStorageService extends StoreService<DaoStateStore> {
                 // the daoState data in memory.
                 store.setDaoState(null);
 
-                // Hint the system for garbage collection. No guarantee if and when it will be executed...
-                System.gc();
                 completeHandler.run();
             });
         }).start();
--- a/core/src/main/java/bisq/core/dao/state/storage/DaoStateStorageService.java
+++ b/core/src/main/java/bisq/core/dao/state/storage/DaoStateStorageService.java
@@ -76,10 +76,32 @@ public class DaoStateStorageService extends StoreService<DaoStateStore> {
         return FILE_NAME;
     }
 
-    public void requestPersistence(DaoState daoState, LinkedList<DaoStateHash> daoStateHashChain) {
+    public void requestPersistence(DaoState daoState,
+                                   LinkedList<DaoStateHash> daoStateHashChain,
+                                   Runnable completeHandler) {
+        if (daoState == null) {
+            completeHandler.run();
+            return;
+        }
+
         store.setDaoState(daoState);
         store.setDaoStateHashChain(daoStateHashChain);
-        persistenceManager.requestPersistence();
+
+        // We let the persistence run in a thread to avoid the slow protobuf serialisation to happen on the user
+        // thread. We also call it immediately to get notified about the completion event.
+        new Thread(() -> {
+            Thread.currentThread().setName("Serialize and write DaoState");
+            persistenceManager.persistNow(() -> {
+                // After we have written to disk we remove the reference to the daoState. We have in the meantime
+                // already cloned the next daoState and if we would not release the reference we would have 3 times
+                // the daoState data in memory.
+                store.setDaoState(null);
+
+                // Hint the system for garbage collection. No guarantee if and when it will be executed...
+                System.gc();
+                completeHandler.run();
+            });
+        }).start();
     }
 
     public DaoState getPersistedBsqState() {
@@ -125,7 +147,7 @@ public class DaoStateStorageService extends StoreService<DaoStateStore> {
 
     @Override
     protected DaoStateStore createStore() {
-        return new DaoStateStore(DaoState.getClone(daoState), new LinkedList<>(daoStateMonitoringService.getDaoStateHashChain()));
+        return new DaoStateStore(null, new LinkedList<>(daoStateMonitoringService.getDaoStateHashChain()));
     }
 
     @Override
--- a/common/src/main/java/bisq/common/persistence/PersistenceManager.java
+++ b/common/src/main/java/bisq/common/persistence/PersistenceManager.java
@@ -26,6 +26,7 @@ import bisq.common.file.FileUtil;
 import bisq.common.handlers.ResultHandler;
 import bisq.common.proto.persistable.PersistableEnvelope;
 import bisq.common.proto.persistable.PersistenceProtoResolver;
+import bisq.common.util.GcUtil;
 import bisq.common.util.Utilities;
 
 import com.google.inject.Inject;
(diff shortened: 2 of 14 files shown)


chimp1984:
Here are the results from 2 apps running over night. I did not use the periodic gc() calls but called it manually after checking the logs.

Version of that PR with G1 GC:
UsedMemoryInMB=1208, totalMemory=1.927 GB, freeMemory=764.248 MB

After calling gc():
UsedMemoryInMB=539, totalMemory=580 MB, freeMemory=40.5 MB


Master version without vm args:
UsedMemoryInMB=836, totalMemory=1.422 GB, freeMemory=619.766 MB

After calling gc():
UsedMemoryInMB=636, totalMemory=698 MB, freeMemory=61.384 MB 

So this PR version had about 100 MB less memory than master which is about 15%.

core/src/main/java/bisq/core/dao/state/storage/DaoStateStorageService.java
```diff
@@ -76,10 +76,32 @@ public String getFileName() {
...
+
+        // We let the persistence run in a thread to avoid the slow protobuf serialisation to happen on the user
+        // thread. We also call it immediately to get notified about the completion event.
+        new Thread(() -> {
+            Thread.currentThread().setName("Serialize and write DaoState");
+            persistenceManager.persistNow(() -> {
+                // After we have written to disk we remove the reference to the daoState. We have in the meantime
+                // already cloned the next daoState and if we would not release the reference we would have 3 times
+                // the daoState data in memory.
+                store.setDaoState(null);
```
sqrrm: It looks to me like the store is reset before completehandler is called which is where the new clone of daoState is made. Comment makes it seem like the clone is made before store is cleared.

sqrrm (commented):
It looks safe, just want to clarify my confusion with the memory reset order

core/src/main/java/bisq/core/dao/state/storage/DaoStateStorageService.java
chimp1984: Yes comment was from earlier version. I updated it in a new commit.

chimp1984:
I tested with dao syncs and there memory goes up quite a lot. To avoid that gc is called too often (takes about 300 ms) I have set the totalMemory trigger to 3gb. This value is used at all the manual calls at areas where memory spikes are expected. Beside that there is a 10 min interval to check if totalMemory is > 1gb and invokes the gc if memory is high. 
With that more conservative approach memory will likely stay around 1 gb, but it avoids that gc gets called too often and as peak memory goes up to about 3gb a 1 gb level at usual operation should be ok.

sqrrm (commented):
At this point the changes look ok. I will run a DAO resync from genesis. The memory management code now looks neater.

chimp1984:
@devinbileck Could you test it on Windows? Would be good to get diff. OS tested before we use it....

devinbileck:
Preliminary testing on Windows 10 using the VM arguments during a DAO resync:

- The memory usage fluctuating as the GC kicks in and it drops significantly. Just watching the task manager shows it getting up to around 2.5GB then dropping sometimes down to 1.5GB or so then ramping up.

- I do notice much more disk activity now as it is persisting to disk rather than remaining in memory, which hopefully will also help with possible corruption and invalid DAO states.

- I am noticing higher CPU usage now (averaging around 80% - up from around 50% or so). Not sure if that will have an impact or not.

I haven't yet evaluated whether there is a noticeable improvement to UI interaction during the DAO resync.

devinbileck:
To visualize the CPU/memory usage better, I used VisualVM during DAO sync:

- master without VM args:
image

- PR with VM args:
image

sqrrm:
I resynced from genesis. Memory usage increasing steadily by 2.5 GB over the period, but looks like 1.7 GB of those are not related to bisq so 800 MB is probably bisq's footprint. I don't have a good memory monitor so it's a bit crude. The choppiness in memory makes sense with the new garbage collection features and I didn't notice any sync issues.
 
image

sqrrm (approved):
ACK

@chimp1984 do you intend to add anything more on this PR or should I merge it?

ghubstan:
1.7 OOME on Linux

- I have had a couple of OOM errors with my live Bisq instance since upgrading to release/v1.7.0.
- I have not had any OOM errors while testing this PR on mainnet, but my testing (mainnet) wallet is always empty, and there is no TX or trade history.

Linux OS Level Profiling (Very Simple)

I have been doing some linux sys-level profiling to see diffs in total physical RAM usage (native + heap).  Native memory use is not shown by java heap profilers. The Linux OS memory profiler I'm using (`htop`) shows total physical mem usage (`RES`).  It includes both on-heap and off-heap (native) memory usage in 1 value.

Addition of periodic `System.gc()` did not reduce overall RAM usage when linux's jvm `MAXRam=4g`.  (Maybe because this 'low' MAXRam value is forcing more frequent GC, making the coded GC redundant and ignored.)

		RES RAM Mb / Tag v1.7 pre PR / MAXRam=4g / no GC opts / 3 trials
			2171
			2041
			2119

		RES RAM Mb / v1.7 PR 5609 commit 8859503 / MAXRam=4g / no GC opts / 3 trials
			2166
			2053
			2099
			
			Note: no effect from added System.gc()	


If you want to merge, don't wait on me.  I think Windows is the biggest worry(?), and I can't help there.  
But I will still do more testing and let you know what I find. 

Today: I am going to see how it is for MAXRam=4g today, with chimp's jvm GC options.
And run tests using jvm opt MAXRam=8g, with and without chimp's jvm GC options.

chimp1984:
> I do notice much more disk activity now as it is persisting to disk rather than remaining in memory

The persistence of dao state has not changed. Only that we do not keep the daostate used for persisting in memory longer as needed.

I have started a full sync from genesis with a full node (OSX), still missing 10 000 blocks, but so far looks all as expected.
Earlier total memory was between 400 MB and 1.7 GB now its about 600 MB and 2.2 GB.

@sqrrm From my side you can merge it. Other potential optimizations can be done later.

Observations from dao parsing performance:
Initially parsing is very fast (10ms), getting the block via rpc is rather slow (about 1 sec.) - what I remember it was faster with the pre-segwit rpc library, but not sure. But a much bigger improvement might be done to request not each single block but a batch.
Over time when dao state gets bigger parsing gets slower as well. Not at block 680000 its about 200-300 ms, though that might be also because of the very long running process (runs since > 48 h). But there is likely more optimization options in that area - usually to add a hashmap as cache to avoid iterations.

sqrrm:
@ManfredKarrer needs to ACK this PR before merge.

ManfredKarrer (approved):
utACK

PR #5614, opened 2021-07-10
Fix payment account deserialize issue (e.g. CHASE_QUICK_PAY)

### LINKED ISSUE #5613: Payment accounts deleted on upgrade to 1.7.0
#### Version

1.7.0

### Steps to reproduce

Have payment accounts set up, including Chase Quickpay.
Upgrade to 1.7.0

### Expected behaviour

Payment accounts should be retained.

### Actual behaviour

Offers are greyed out / not takeable.
Check the Payment Accounts screen, all fiat & altcoin accounts have gone.

#### Additional info

From the log:

    Jul-10 07:46:30.344 [PersistenceManager-read-PreferencesPayload] INFO  b.c.p.PersistenceManager: Reading PreferencesPayload completed in 112 ms 
    Jul-10 07:46:30.404 [PersistenceManager-read-UserPayload] ERROR b.c.p.PersistenceManager: Reading UserPayload failed with Not supported PaymentMethod: PaymentMethod(id=N/A, maxTradePeriod=0, maxTradeLimit=0). 
    Jul-10 07:46:30.413 [PersistenceManager-read-UserPayload] ERROR bisq.common.app.DevEnv: java.lang.RuntimeException: Not supported PaymentMethod: PaymentMethod(id=N/A, maxTradePeriod=0, maxTradeLimit=0) 

---

I am investigating a fix for the issue.

ghost:
Fixes #5613

When loading payment accounts which have been deprecated, an exception is thrown which is not handled.  This causes the entire set of payment accounts to be lost.

To fix this we check for an exception when deserializing a payment account, and log a warning with the details.  The erroneous payment account will therefore be dropped from the list, which is what is expected.

I tested using the following procedure on both regtest & mainnet Bisq.

### Testing:

- Create a new Bisq account using v1.6.5 (or earlier).
- Add Amazon Giftcard and Chase QuickPay accounts for the user.
- Close Bisq, log in again using the same version and check that both accounts are listed.
- Close Bisq, upgrade to (v1.7.0 + this PR) and run it.
- Check that the Chase QuickPay account is no longer in the list, but the Amazon Giftcard account is present.

Check that a log warning is written, like this: 

    Jul-10 10:12:05.316 [PersistenceManager-read-UserPayload] WARN  b.core.payment.PaymentAccount: Could not load account: CHASE_QUICK_PAY, exception: java.lang.RuntimeException: Not supported PaymentMethod: PaymentMethod(id=N/A, maxTradePeriod=0, maxTradeLimit=0)

(It will actually be logged 2 times since UserPreferences deserializes 2 times).

----

### DIFF
--- a/core/src/main/java/bisq/core/user/UserPayload.java
+++ b/core/src/main/java/bisq/core/user/UserPayload.java
@@ -33,6 +33,7 @@ import bisq.common.proto.persistable.PersistableEnvelope;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -132,6 +133,7 @@ public class UserPayload implements PersistableEnvelope {
                 ProtoUtil.stringOrNullFromProto(proto.getAccountId()),
                 proto.getPaymentAccountsList().isEmpty() ? new HashSet<>() : new HashSet<>(proto.getPaymentAccountsList().stream()
                         .map(e -> PaymentAccount.fromProto(e, coreProtoResolver))
+                        .filter(Objects::nonNull)
                         .collect(Collectors.toSet())),
                 proto.hasCurrentPaymentAccount() ? PaymentAccount.fromProto(proto.getCurrentPaymentAccount(), coreProtoResolver) : null,
                 proto.getAcceptedLanguageLocaleCodesList().isEmpty() ? new ArrayList<>() : new ArrayList<>(proto.getAcceptedLanguageLocaleCodesList()),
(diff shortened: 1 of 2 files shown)


core/src/main/java/bisq/core/user/UserPayload.java
```diff
@@ -132,6 +132,7 @@ public static UserPayload fromProto(protobuf.UserPayload proto, CoreProtoResolve
                 ProtoUtil.stringOrNullFromProto(proto.getAccountId()),
                 proto.getPaymentAccountsList().isEmpty() ? new HashSet<>() : new HashSet<>(proto.getPaymentAccountsList().stream()
                         .map(e -> PaymentAccount.fromProto(e, coreProtoResolver))
+                        .filter(e -> e != null)
```
ripcurlx:
```suggestion
                        .filter(Objects::nonNull)
```

ripcurlx: NIT ☝️

ghost: Thanks!  change applied.

ripcurlx (approved):
utACK

PR #5615, opened 2021-07-11
Enable lost payment accounts to be imported from backups

ghost:
See #5613.  This feature depends on the bugfix included in #5614.

Some users lost their payment accounts when upgrading to v1.7.0.  This temporary feature allows the payment accounts to be re-imported from the automatic backups.

If you had previously done an Export Accounts you do not need this feature; you can simply re-import that file. 

Otherwise, this feature can be used as a last resort to import from Bisq's automatic backups.  Caveat: only 10 rolling backups are kept so if you have restarted Bisq more than 10 times since losing your accounts in v1.7.0, the older backups containing the data have probably been replaced.

If you have done [backups of your data directory](https://bisq.wiki/Backing_up_application_data) before upgrading to v1.7.0 you can also restore the accounts from there in the same way as described below, just select the appropriate path within your backup when prompted.

---

To use: 

- go to Account -> National Currency Accounts -> Import Accounts
- In the file picker, navigate to the directory location of your Bisq periodic backups (see [Backups Location](backups-location) below) and pick a file to import.  The file will have a name like `1625921084736_UserPayload`.  Choose one with a recent date from before you upgraded to v1.7.0.  The ones with larger file size will most likely contain your lost accounts.
 
e.g.

image

- Click on `Open` and it will tell you the accounts which were imported, as follows.  If you already have the accounts in Bisq it will not import them twice, as each account has a unique identifier.  If no accounts were imported it will say so.

image

### Backups Location

- (Windows) : `C:\Users\<username>\AppData\Roaming\Bisq\btc_mainnet\db\backup\backups_UserPayload`
- (Mac) : `/Users/<username>/Library/Application Support/Bisq/btc_mainnet/db/backup/backups_UserPayload`
- (Linux) : `/home/<username>/.local/share/Bisq/btc_mainnet/db/backup/backups_UserPayload`

### DIFF
--- a/desktop/src/main/java/bisq/desktop/util/GUIUtil.java
+++ b/desktop/src/main/java/bisq/desktop/util/GUIUtil.java
@@ -52,6 +52,7 @@ import bisq.core.trade.txproof.AssetTxProofResult;
 import bisq.core.user.DontShowAgainLookup;
 import bisq.core.user.Preferences;
 import bisq.core.user.User;
+import bisq.core.user.UserPayload;
 import bisq.core.util.FormattingUtils;
 import bisq.core.util.coin.BsqFormatter;
 import bisq.core.util.coin.CoinFormatter;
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
utACK

KISBITCOIN:
That fix does not work for me. I get same result as when I try to import my fiat payment backup file. I'm running Bisq on Mac osx 10.15.7

ghost:
Did you first apply PRs 5614 and 5615 to your Bisq?  You can DM me on Keybase (same user name) and I'll help you.

ghost:
Summary of testing performed:


- start with a fresh accounts for Bob and Alice, using Bisq v1.6.5 or earlier.
- create accounts Chase QuickPay, Amazon Giftcard, and Zelle.
- Bob creates an offer to Sell 0.01 BTC using Zelle.
- make backups of the data directory.
- upgrade to 1.7.0.  Notice that they both lose all accounts.
- upgrade to 1.7.2. (containing #5614 and #5615)
- go to Accounts, and import old accounts from a backup as described in the OP above.
- now Bob and Alice both have restored their original Amazon Giftcard and Zelle accounts.
- Alice can successfully take Bob's original Zelle offer and complete the trade.

PR #5618, opened 2021-07-11
Improve Validation of Deposit Inputs

jimmyneutront:
Improve the accuracy and readability of error messages that arise during the validation of deposit inputs, and stop throwing unnecessary errors when a trade's reserved-for-trade wallet has been funded by an external wallet

### DIFF
--- a/core/src/main/java/bisq/core/trade/TradeDataValidation.java
+++ b/core/src/main/java/bisq/core/trade/TradeDataValidation.java
@@ -356,23 +356,46 @@ public class TradeDataValidation {
     }
 
     public static void validateDepositInputs(Trade trade) throws InvalidTxException {
-        // assumption: deposit tx always has 2 inputs, the maker and taker
-        if (trade == null || trade.getDepositTx() == null || trade.getDepositTx().getInputs().size() != 2) {
-            throw new InvalidTxException("Deposit transaction is null or has unexpected input count");
+        /*
+         * deposit tx usually has 2 inputs, the maker and taker
+         * however, if the maker funds the reserved for trade wallet from
+         * an external wallet, the deposit tx will have more than 2 inputs
+         */
+        if (trade == null || trade.getDepositTx() == null) {
+            throw new InvalidTxException("Deposit transaction is null");
+        }
+        if (trade.getDepositTx().getInputs().size() != 2) {
+            log.warn("Deposit transaction has unusual input count: " + trade.getDepositTx().getInputs().size() + " rather than 2");
         }
         Transaction depositTx = trade.getDepositTx();
-        String txIdInput0 = depositTx.getInput(0).getOutpoint().getHash().toString();
-        String txIdInput1 = depositTx.getInput(1).getOutpoint().getHash().toString();
+        boolean hasFoundMakerTx = false;
+        boolean hasFoundTakerTx = false;
         String contractMakerTxId = trade.getContract().getOfferPayload().getOfferFeePaymentTxId();
         String contractTakerTxId = trade.getContract().getTakerFeeTxID();
-        boolean makerFirstMatch = contractMakerTxId.equalsIgnoreCase(txIdInput0) && contractTakerTxId.equalsIgnoreCase(txIdInput1);
-        boolean takerFirstMatch = contractMakerTxId.equalsIgnoreCase(txIdInput1) && contractTakerTxId.equalsIgnoreCase(txIdInput0);
-        if (!makerFirstMatch && !takerFirstMatch) {
-            String errMsg = "Maker/Taker txId in contract does not match deposit tx input";
-            log.error(errMsg +
-                "\nContract Maker tx=" + contractMakerTxId + " Contract Taker tx=" + contractTakerTxId +
-                "\nDeposit Input0=" + txIdInput0 + " Deposit Input1=" + txIdInput1);
-            throw new InvalidTxException(errMsg);
+        for (TransactionInput input : depositTx.getInputs()) {
+            if (contractMakerTxId.equalsIgnoreCase(input.getOutpoint().getHash().toString())) {
+                hasFoundMakerTx = true;
+            } else if (contractTakerTxId.equalsIgnoreCase(input.getOutpoint().getHash().toString())) {
+                hasFoundTakerTx = true;
+            }
+            if (hasFoundMakerTx && hasFoundTakerTx) {
+                break;
+            }
+        }
+        if (!hasFoundMakerTx || !hasFoundTakerTx) {
+            String errMsg = "For trade with id " + trade.getId() + ":\n";
+            if (!hasFoundMakerTx) {
+                errMsg = errMsg + "Maker txId in contract does not match any deposit tx input" +
+                        "\nContract Maker tx=" + contractMakerTxId;
+            }
+            if (!hasFoundTakerTx) {
+                errMsg = errMsg + "Taker txId in contract does not match any deposit tx input" +
+                        "\nContract Taker tx=" + contractTakerTxId;
+            }
+            for (int index = 0; index < depositTx.getInputs().size(); index++) {
+                errMsg = errMsg + "\nDeposit Input" + index + "=" + depositTx.getInput(index).getOutpoint().getHash().toString();
+            }
+            log.warn(errMsg);
         }
     }
 


core/src/main/java/bisq/core/trade/TradeDataValidation.java
```diff
@@ -356,23 +356,46 @@ public static void validatePayoutTxInput(Transaction depositTx,
     }
 
     public static void validateDepositInputs(Trade trade) throws InvalidTxException {
-        // assumption: deposit tx always has 2 inputs, the maker and taker
-        if (trade == null || trade.getDepositTx() == null || trade.getDepositTx().getInputs().size() != 2) {
-            throw new InvalidTxException("Deposit transaction is null or has unexpected input count");
+        /*
+         * deposit tx usually has 2 inputs, the maker and taker
```
chimp1984, line 360: Can you provide more detail when that would be expected? I do not see any case where that can happen.
Funding goes to the fee txs, the deposit tx always has the 1 output from the fee tx. The deposit can have 2 outputs in case the trade amount is < offer amount, as one output would be change to the seller/maker then.

chimp1984 (changes requested):
NACK. I think it is wrong. We always have 2 inputs to the deposit tx. If you have a clear case where that is not the case, please post all the trade transactions and describe how you could reproduce it. Thanks.

core/src/main/java/bisq/core/trade/TradeDataValidation.java
jimmyneutront: This change will allow me to bring much more liquidity to Bisq; specifically: more sell offers at much better prices.

The deposit transaction may have more than two inputs in the following scenario:

The maker/seller publishes a sell offer, and pays the maker fee. Then, it gets an address, flags it as "RESERVED_FOR_TRADE", and associates it with the new sell offer's ID. However, it does NOT send the seller security deposit amount + offer amount to the address that has been flagged as "RESERVED_FOR_TRADE" and associated with the new sell offer's ID.

The taker/buyer attempts to take the offer, sending an InputsForDepositTxRequest to the maker/seller, and awaits a InputsForDepositTxResponse.

As soon as the maker/seller receives the InputsForDepositTxRequest, it completes the MakerProcessesInputsForDepositTxRequest TradeTask. Then, it sends the seller security deposit amount + offer amount to the address that has been flagged as "RESERVED_FOR_TRADE" and associated with the id of the offer being taken. It may send multiple transactions that sum to this amount, and thus the deposit transaction will have more than 2 inputs as a result. (The maker/seller can purchase an amount of BTC equal to the seller security deposit amount + offer amount from a centralized exchange, and then immediately withdraw it to the address flagged as "RESERVED_FOR_TRADE" and associated with the id of the offer being taken.)  

Then, the maker/seller completes the following TradeTasks:
ApplyFilter
VerifyPeersAccountAgeWitness
MakerVerifyTakerFeePayment (although this does nothing)
MakerSetsLockTime
MakerCreateAndSignContract

Then, the maker/seller completes the SellerAsMakerCreatesUnsignedDepositTx TradeTask, and in the process it uses UTXOs at the "RESERVED_FOR_TRADE" address associated with the id of the offer being taken to fund the deposit transaction. Since the security deposit amount + offer amount may have been sent to that address split across multiple transactions (as described 2 paragraphs above) the deposit transaction will thus have more than 2 inputs, though it will otherwise be perfectly valid. The maker/seller will then complete the SellerAsMakerSendsInputsForDepositTxResponse TradeTask, sending an InputsForDepositTxResponse to the taker/buyer.

As soon as the taker/buyer receives the InputsForDepositTxResponse, it completes the following TradeTasks:
ApplyFilter
VerifyPeersAccountAgeWitness
TakerVerifyAndSignContract
TakerPublishFeeTx (this may be problematic, read on for details)
	
Then, the taker/buyer completes the BuyerAsTakerSignsDepositTx TradeTask, and in the process it verifies that the unsigned deposit transaction received from the maker/seller is properly formed. If the unsigned deposit transaction is not properly formed, it throws an exception. However, in this case, no exception is thrown, because the deposit transaction created by the maker/seller is valid. BuyerAsTakerSignsDepositTx doesn't count the number of inputs to the deposit transaction, because it is irrelevant. The only code that performs this action is the validateDepositInputs method in the TradeDataValidation class, and again, it is an unnecessary validation that is irrelevant. Furthermore, the validateDepositInputs function is only run AFTER the offer has been successfully taken and has become a trade; at that point, the taker/buyer has already paid the taker fee and locked up the buyer security deposit amount.

After the taker/buyer completes the BuyerAsTakerSignsDepositTx TradeTask, it then completes the BuyerAsTakerSendsDepositTxMessage TradeTask, and the trade process carries on normally, with no problems.

Because the maker/seller funded the deposit transaction with UTXOs sent from an external address, the maker fee transaction id will not match any of the transaction ids used to fund the deposit transaction. Although again: this validation is unnecessary and irrelevant. If there is anything wrong with the deposit transaction that could cause the trade to fail, it will be discovered by the taker/buyer's BuyerAsTakerSignsDepositTx TradeTask. Furthermore, this validation is currently performed by validateDepositInputs only AFTER the offer has been successfully taken and has become a trade with no problems.

If the maker/seller doesn't fund the "RESERVED_FOR_TRADE" address associated with the id of the offer being taken and send an InputsForDepositTxResponse to the taker/buyer in time:
The offer taking process will timeout and the trade will fail. However, because the taker/buyer doesn't publish the taker fee transaction until AFTER an InputsForDepositTxResponse has been received, the seller will not lose anything.

If the maker/seller sends an invalid unsigned deposit transaction to the taker/buyer:
The trade will fail. Additionally, because the seller/taker completes the TakerPublishFeeTx TradeTask before it completes the BuyerAsTakerSignsDepositTx TradeTask, which verifies that the unsigned deposit transaction received from the maker/seller is valid, the taker/buyer may lose their taker fee, unless the maker/seller resolves the issue with their deposit transaction. However, we cannot be sure the maker/seller will do this. Therefore, I believe we should cause the BuyerAsTakerSignsDepositTx TradeTask to be completed before the TakerPublishFeeTx TradeTask is completed.

I know that much of the maker/seller's behavior that I have described here isn't possible with Bisq at the moment. However, I have a private Bisq repo where I am working on this functionality myself. Again: the change I am proposing poses no new risk to the offer taker. It simply removes an unnecessary validation step and will allow me to bring much more liquidity to Bisq.

Concerning the issue with the TakerPublishFeeTx TradeTask: I believe this is a separate issue that needs attention. A malicious user can send invalid unsigned deposit transactions to takers, causing takers to lose their taker fee amount. I will look into this further.

chimp1984: I welcome your initiative, but I dislike that you have not been open and clear initially. The short comment you added did not at all point out the real intention and motivation.
I consider that a trade protocol change, even it touches only the fee tx and preparation part. I have not read now the whole post, would take me more time... But I suspect that there might be issues with the maker fee payments. 
In any case such a change require very critical review and you should have made it much more clear what it is about.

jimmyneutront: My apologies for not being more clear.

Could you elaborate a bit more on your suspicions about issues with maker fee payments? It appears that the taker never thoroughly validates the maker fee transaction by ensuring that it sent the correct amount to BTC to a FeeReceiver address. As long as the String in the OfferPayload's offerFeePaymentTxId field matches the id of the transaction that the maker used to fund the deposit transaction, the taker's Bisq instance raises no errors. And even if these don't match, validateDepositInputs() only raises an error AFTER the trade has been taken and funds are locked in multisig. I tested this by completing a trade for which the seller/maker's Bisq instance was built with line 159 of [TradeWalletService.java](https://github.com/bisq-network/bisq/blob/030a76522cb1d9325f45656c7814043b809be68e/core/src/main/java/bisq/core/btc/wallet/TradeWalletService.java) commented out, so the seller/maker paid no maker fee. The buyer/taker was able to take the offer and complete the entire trade protocol with no errors.

How could my proposed change cause any new problems? Additionally, I apologize that my  comment was not as clear and open as it should have been. I'd be happy to amend the commit and clarify the comment, if you would like me to; what information would you like me to include?

chimp1984: I do not have enough time atm to review it and my motivation to spend more time on it decreased due the misleading PR comment. Hope some other dev can review it with the care required.

jimmyneutront: I believe we may have had a miscommunication: the only code I am asking to be reviewed is in the one file I changed in this PR. All the other code that enables the functionality that I described in the long comment above is in a private repository, and I don't plan on ever submitting a pull request for it. I am highly confident that this PR will have no negative effects whatsoever, and that it will allow me to bring much more affordable sell offers to Bisq.

Again, I sincerely apologize that I was not initially as clear as I should have been, and I am more than happy to amend my commit with a better comment and answer any questions whatsoever to expedite and facilitate the review process.

Since you do not have the time to review this change, do I need to request a review from another developer myself, or will another one be assigned to this PR?

PR #5628, opened 2021-07-18
Add don't show again option to DAO resync popup

ghost:
As requested @ripcurlx 
Same logic as #5552, and user is able to turn off subsequent showings.

Screenshot from 2021-07-18 18-46-17


ripcurlx (approved):
utACK

Issue #5550, opened 2021-06-01
Java Heap Space error


ThisMightHelp:
### Description

 Java Heap Space error. No other info.

#### Version

 v1.6.4

### Steps to reproduce

 One time occurence

### Expected behaviour

Unknown

### Actual behaviour

Unknown

### Screenshots

 
[SS java heap space error 6-1-2021.odt](https://github.com/bisq-network/bisq/files/6580851/SS.java.heap.space.error.6-1-2021.odt)

#### Device or machine

Win 10 Pro

#### Additional info

I saved the log file, but can't find it. How do I redisplay it from within the UI? Thanks.

ripcurlx:
@ThisMightHelp You can find the log file here: https://bisq.wiki/Data_directory#Windows

Unfortunately I wasn't able to display the screenshot file you posted. For how long did your client run before this happened?
How much RAM do you have available on your machine?

marekbedy:
I have the same problem on Linux Ubuntu with Oracle Hotspot Java / OpenJDK JRE.

![image](https://user-images.githubusercontent.com/5310638/120493694-768a2e00-c3bb-11eb-9799-ab4f6167e92f.png)

From the logs:

```
Jun-02 08:04:51.803 [Write-DaoStateStore_to-disk] ERROR b.c.s.CommonSetup: Uncaught Exception from thread Write-DaoStateStore_to-disk
Jun-02 08:04:51.803 [Write-DaoStateStore_to-disk] ERROR b.c.s.CommonSetup: throwableMessage= Java heap space
Jun-02 08:04:51.803 [Write-DaoStateStore_to-disk] ERROR b.c.s.CommonSetup: throwableClass= class java.lang.OutOfMemoryError
Jun-02 08:04:51.803 [Write-DaoStateStore_to-disk] ERROR b.c.s.CommonSetup: Stack trace:
java.lang.OutOfMemoryError: Java heap space
```

Usually it happens 1-2x a day, my machine runs 6GB RAM. 
Linux bisq 5.11.0-18-generic #19-Ubuntu SMP Fri May 7 14:22:03 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

ThisMightHelp:
Thanks ripcurlz. The screenshot was the same as marekbedy's. It happened once before, but otherwise Bisq seems ok. Bisq was running for a few hours when the error popped up. Running Win 10 Pro and 32GB RAM.

pazza83:
[...]

PR #5629, opened 2021-07-19
Update Java options as described in #5609 and #5550

ripcurlx:
See https://github.com/bisq-network/bisq/issues/5550 and https://github.com/bisq-network/bisq/pull/5609.


ripcurlx:
@chimp1984 @sqrrm @jmacxx Do we want to go higher on the max ram setting than 6G? @pazza83 had to add 32G to make it work for him.

chimp1984:
Maybe 8G? @ghubstan What do you think?

ghubstan:
> Maybe 8G? @ghubstan What do you think?

People using computers w/ just 4g RAM are SOL.  That said, 8g is OK.   Much better than 6g.

I'm using a 10g setting, and total native memory (on-heap + off-heap) use does not go over ~ 3.7g (so far).

sqrrm (approved):
utACK

PR #5632, opened 2021-07-23
Keep chatview listener open until ack, mailbox, or error

ghost:
#5599 had an side effect that did not manifest until tested on mainnet (with tor, latency etc).

Chat message status was not picked up by the display, even if it was set correctly on the object (scrolling up and down on the chat box would update the display to correct status).  This happened because the ChatView stopped listening for updates prematurely.  Changed it to wait for acknowledged status instead of arrived status.

The timeout for displaying a `Try sending message again` error message had to be increased from 5 seconds to 60.  In mainnet it routinely takes 7 - 16 seconds to open a tor connection so the try again message was being shown all the time unless a connection was already established.  60 seconds gives enough time for the message to be acked, or saved to mailbox.  Then only the rare outliers should get the try again message when the message apparently arrives but is not acked (#5526).


ripcurlx (approved):
utACK

PR #5639, opened 2021-07-27
Allow LTC accounts to use segwit addresses

ghost:
Fixes issue described by a user here:

https://old.reddit.com/r/bisq/comments/omf3l3/cannot_add_ltc_account/

> Under "altcoin accounts" I'm trying to add the address of my LTC core wallet.
> Neither regular nor Bech32 addresses works. Anyone know How I'd go about adding LTC?

---

### Implementation:

- The address validator specified the bitcoin magic number for wrapped segwit, 5 -- it should be 50 for LTC.
- The address validator expected base58 only (legacy format).  Changed to use an address validator that supports bech32 and base58.

### Testing:

Added tests for wrapped segwit and bech32 address formats.

---

image

image

### DIFF
--- a/assets/src/main/java/bisq/asset/coins/Litecoin.java
+++ b/assets/src/main/java/bisq/asset/coins/Litecoin.java
@@ -17,19 +17,20 @@
 
 package bisq.asset.coins;
 
-import bisq.asset.Base58AddressValidator;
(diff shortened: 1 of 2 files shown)


ripcurlx (approved):
utACK

PR #5640, opened 2021-07-27
Update expired public key

ripcurlx:
I did update my expired key already locally some time ago, but forgot to update it in the repo as well.
I just updated the public key used for verifying the downloaded files in the release.


sqrrm (approved):
utACK

PR #5643, opened 2021-07-30
Prevent negative suggested mining fee when burning BSQ for fees

ghost:
Problem: Negative mining fees could be shown in the make/take offer summary popup when BSQ is used for fees and the BTC sats value of the BSQ is greater than the calculated transaction fee.  

Using BSQ to pay trade fees burns the BSQ effectively paying miners with those sats.  A bug surfaced whereby negative mining fees could be displayed, giving a false/confusing impression to users.  Note that this is a display issue only, there were no problems with the trading transactions themselves.

After this fix, the displayed mining fee cannot be negative. 

image

Screenshot from 2021-07-29 21-34-53

In these examples the BSQ fee rate (Param.java) was artificially increased to 100 BSQ/BTC producing larger than usual BSQ fees and triggering the bug in order to validate the fix.


sqrrm (approved):
utACK

PR #5644, opened 2021-07-30
Fix issue for deprecated payment account imported from backup

ghost:
When importing payment accounts from a backup file, we have to guard for the fact that deprecated accounts may be present so test for null.  This is the same as #5614 except applied to the account backup file.

Thanks to @KISBITCOIN for discovering this.


ripcurlx (approved):
utACK

PR #5649, opened 2021-08-02
Avoid creating more than one refund payout

sqrrm:
Double clicking the close ticketbutton creates two payout transactions.
This fix makes sure only one payout transaction is created for the
dispute.

Restarting the client allows for creating another refund transaction
for the dispute if needed.

Refund agent reported a case of a refund being done twice. This PR makes sure that two transactions are not created for the same refund during one session. It's still possible to restart the client to make another refund for an already closed ticket if needed.

I managed to reproduce at least one case of double refund transactions by double clicking the close ticket button, there might be other ways to trigger the refund payout twice, but with this fix it will only create one refund transaction no matter which way the payout was triggered.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/overlays/windows/DisputeSummaryWindow.java
+++ b/desktop/src/main/java/bisq/desktop/main/overlays/windows/DisputeSummaryWindow.java
@@ -673,6 +673,13 @@ public class DisputeSummaryWindow extends Overlay<DisputeSummaryWindow> {
     }
 
     private void showPayoutTxConfirmation(Contract contract, DisputeResult disputeResult, ResultHandler resultHandler) {
+        if (dispute.isPayoutDone()) {
+            new Popup().headLine(Res.get("disputeSummaryWindow.close.alreadyPaid.headline"))
+                    .confirmation(Res.get("disputeSummaryWindow.close.alreadyPaid.text"))
+                    .closeButtonText(Res.get("shared.cancel"))
+                    .show();
+        }
+
         Coin buyerPayoutAmount = disputeResult.getBuyerPayoutAmount();
         String buyerPayoutAddressString = contract.getBuyerPayoutAddressString();
         Coin sellerPayoutAmount = disputeResult.getSellerPayoutAmount();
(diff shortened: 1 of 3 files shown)


sqrrm:
@jmacxx could you review this one? Would be good to build a special release for refund agent.

desktop/src/main/java/bisq/desktop/main/overlays/windows/DisputeSummaryWindow.java
```diff
@@ -673,6 +673,13 @@ private void addButtons(Contract contract) {
     }
 
     private void showPayoutTxConfirmation(Contract contract, DisputeResult disputeResult, ResultHandler resultHandler) {
+        if (dispute.isPayoutDone()) {
+            new Popup().headLine(Res.get("disputeSummaryWindow.close.alreadyPaid.headline"))
```
ghost, line 677: The `payoutDone` guard works fine, and it issues the log warning about avoiding double payouts.  However I cannot get this warning popup to show - so it seems redundant.

ghost (changes requested):
It solves the issue of being able to make multiple payouts by double clicks on the `Close Ticket` button, ~but the new warning popup seems redundant as it is never shown~. [edit] all good.

desktop/src/main/java/bisq/desktop/main/overlays/windows/DisputeSummaryWindow.java
sqrrm: You get this warning if you try to do a payout on the same user twice without closing the counterparty ticket first.

ghost (approved):
ACK (sqrrm explained how to get the warning popup to show).

ripcurlx (approved):
utACK

refund-agent2:
Many thanks for the fix, during the reimbursements I did today I got the alert message about 2 payouts being generated and the need to restart the client.

sqrrm:
@refund-agent2 good to know it's all working as expected now. It's strange this hasn't happened before as it was an old bug, not something introduced recently.

PR #5651, opened 2021-08-03
Fix issue with extraData field when editing offer

### LINKED ISSUE #5597: account info in orderbook was not changed to the related trading account
account info in orderbook was not changed to the related trading account

### Description

i created an offer to buy btc, but wrongly with the F2F account, after changing to SEPA, the payment method is changed correct to SEPA but the account info in orderbook is still show the info from the F2F account which has no trades and is not signed

#### Version

since 1.6.4 and still exist with 1.7

#### Device or machine

manjaro linux

ghost:
Reported by @stakoholic, when editing an offer some information was not being transferred.

`extraData` is used for the account age and other custom fields like F2F city.
So when editing an offer the extraData should be obtained from the edited offer not the original one.

Fixes #5597
FYI @ghubstan this one line fix also needs carrying over to the current editoffer API work.

---

### Testing:

You have an Amazon gift card account which is 4 days old (for example).
Create an F2F account which specifies city as 'Paris' and additional information as "ask for Freddie".

**Prior version: (bug)**

- Create offer to buy 0.01 BTC using Amazon gift card.
- Edit the offer changing payment account to F2F.
- Look at the offer in the order book.  
- It should show the F2F city ('Paris') but it is blank.
- It should show additional information ("ask for Freddie") but it is blank.
- It should show the account age as 0 days old but instead shows it as 4 days old.
 
image

**Version with fix applied:**

- Create offer to buy 0.01 BTC using Amazon gift card.
- Edit the offer changing payment account to F2F.
- Look at the offer in the order book.  
- It shows the F2F city ('Paris').
- It shows additional information as "ask for Freddie".
- It shows the account age as 0 days old.

image

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/portfolio/editoffer/EditOfferDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/portfolio/editoffer/EditOfferDataModel.java
@@ -223,7 +223,7 @@ class EditOfferDataModel extends MutableOfferDataModel {
                 offerPayload.getUpperClosePrice(),
                 offerPayload.isPrivateOffer(),
                 offerPayload.getHashOfChallenge(),
-                offerPayload.getExtraDataMap(),
+                newOfferPayload.getExtraDataMap(),
                 offerPayload.getProtocolVersion());
 
         final Offer editedOffer = new Offer(editedPayload);


ghubstan:
>> FYI @ghubstan this one line fix also needs carrying over to the current `editoffer` API work.

@jmacxx Thanks for the heads up.  I do need to adjust the api to this change.  

You seem to already know it only affects new api method `editoffer`, which has not been merged into master yet.  And, api method `editoffer` v1 cannot change payment acct details -- just offer price details.  But I plan to make the change this week.

ripcurlx (approved):
utACK

PR #5659, opened 2021-08-11
Add UI OfferBook list add/remove checks

ghubstan:
Problem:  Using the API's CLI to edit offers causes add/remove messages to be received on peers in the same envelope batch, then passed to the UI in random order.  When `OfferBook` receives a remove(offer) event after an add(offer) event, the edited offer will be removed immediately after being added to the UI's `OfferBook` list.  

Fix:  This change uses storage entry sequence number and storage entry payload hash comparisons to avoid the problem.

- `OfferBookListItem` Added new constructor taking `P2PDataStorage.ByteArray hashOfPayload`,
  and `int sequenceNumber` params.  Defined a `toString` method.
  
- `OfferBook` Added new checks on `OfferBookListItem`'s `hashOfPayload` and `sequenceNumber` while
  determining if offer candidates should be added or removed from the UI's OfferBook List.
  See OfferBook contructor's implementation of OfferBookChangedListener#onAdded and
  OfferBookChangedListener#onRemoved.  Added many (maybe too many) comments explaining the add/remove rules,
  and plenty of debug statements (maybe too many) to help trace the add/remove event process.
  
- `OfferBookService#OfferBookChangedListener` Added new `P2PDataStorage.ByteArray` hashOfPayload,
  and `int sequenceNumber` params to listener's `onAdded` and `onRemoved` method signatures.
  Added these two new parameter values to listener.onAdded and listener.onRemoved calls.
  
- `TakeOfferDataModel` Replaced unused `tradeManager` param in `offerBook.removeOffer`
  with `P2PDataStorage.ByteArray hashOfPayload` (null), and `int sequenceNumber` (-1) params.
  `OfferBook` will remove the candidate offer as before.

- MarketAlerts` Adjusted `onAdded` & `onRemoved` listener method signatures.

This PR is based on https://github.com/bisq-network/bisq/pull/5577, which should be reviewed and merged before this one.

PR #5663, opened 2021-08-14
Display Enabled=PENDING in CLI 'createoffer' output

ghubstan:
A newly created offer has no OpenOffer+State (`AVAILABLE` || `DEACTIVATED`) when displayed in the CLI's console.  This change adds a `bool isMyPendingOffer`  to the `OfferInfo` proto + wrapper, and the CLI's console offer output formatter uses it to determine if it should display an offer's `Enabled` column value as `PENDING`, instead of an ambiguous `NO` value.

This PR is based on https://github.com/bisq-network/bisq/pull/5659, which should be reviewed and merged before this one.


ghubstan:
File conflicts resolved in child PR https://github.com/bisq-network/bisq/pull/5666

PR #5666, opened 2021-08-16
Adjust API 'editoffer' to PR 5651 (include extraData field when editing offer)

ghubstan:
PR https://github.com/bisq-network/bisq/pull/5651 ensures payment acct changes are saved when the UI edits an offer.

Although the API `editoffer` command does not yet support editing payment account details, this series of PRs must adjust to 5651 (merged today) to avoid file conflicts.

This PR is based on https://github.com/bisq-network/bisq/pull/5663, which should be reviewed and merged before this one.


desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBook.java
```diff
@@ -83,44 +85,113 @@ public void onAdded(Offer offer) {
...
+            return;
+        }
+
+        OfferBookListItem candidate = candidateWithMatchingPayloadHash.get();
+
+        // Remove the candidate only if the candidate's offer payload hash matches the
+        // onRemoved hashOfPayload parameter.  We may receive add/remove messages out of
+        // order (from api's 'editoffer'), and use the offer payload hash to
+        // ensure we do not remove an edited offer immediately after it was added.
+        if ((candidate.getHashOfPayload() == null || candidate.getHashOfPayload().equals(hashOfPayload))) {
```
sqrrm, line 173: How can there be candidates with null hash?

ghubstan: There can be a null hash each time a user clicks on the view, and calling `OfferBook#fillOfferBookListItems()`.  That method loads the view from `offerBookService.getOffers()`, and the returned list elements are mapped to new `OfferBookListItem(offer)`s with a null hash code.

We discussed this on keybase, when you mentioned the idea of adding a `transient var hash` to OfferPayload.
At the time, I also mentioned that without the cached OfferPayload hash,  OfferBook line 208 might be 
        `.map(o -> new OfferBookListItem(o, get32ByteHashAsByteArray(o.getOfferPayload()))`

There is a comment about null hashes in `OfferBookListItem`:  _A null value indicates the item's payload hash has not been set by onAdded or onRemoved since the most recent OfferBook view refresh._

desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBook.java
```diff
@@ -83,44 +85,112 @@ public void onAdded(Offer offer) {
...
 
-    public void removeOffer(Offer offer, TradeManager tradeManager) {
+    private void removeDuplicateItem(OfferBookListItem newOfferBookListItem) {
+        String offerId = newOfferBookListItem.getOffer().getId();
+        // We need to remove any view items with a matching offerId before
+        // a newOfferBookListItem is added to the view.
+        List<OfferBookListItem> duplicateItems = offerBookListItems.stream()
+                .filter(item -> item.getOffer().getId().equals(offerId))
+                .collect(Collectors.toList());
+        if (duplicateItems.size() > 0) {
```
sqrrm: This check is not needed since `forEach` won't do anything if there are no items in the list.

desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBook.java
```diff
@@ -83,44 +85,112 @@ public void onAdded(Offer offer) {
...
+            duplicateItems.forEach(oldOfferItem -> {
+                offerBookListItems.remove(oldOfferItem);
+                if (log.isDebugEnabled()) {
+                    log.debug("onAdded: Removed old offer {}\n"
+                                    + "\twith payload hash {} from list.\n"
+                                    + "\tThis may make a subsequent onRemoved( {} ) call redundant.",
+                            offerId,
+                            oldOfferItem.getHashOfPayload() == null ? "null" : oldOfferItem.getHashOfPayload().getHex(),
+                            oldOfferItem.getOffer().getId());
+                }
```
sqrrm: There are still a lot of debug logs. Do you want to keep them for the first release with these changes? If so, I think it would be good to add a TODO to remove them later as I don't think they're that valuable in the long run.

desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBook.java
sqrrm: Does this mean an item with null hash will be removed if only the id matches? Seems like a way to get around the hash checks.

desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBook.java
ghubstan: Removed list.size check in commit 4889da6bba6ad155436a2dede5182643f706ac6d.

desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBook.java
ghubstan: >> RE  _a lot of debug logs_

TODO comments are added in commit 84036bd86212d0eab8bd48a1114985c5b87b14e1.

desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBook.java
ghubstan: It is a way to get around the hash check if the list contains an item loaded by `OfferBook#fillOfferBookListItems()` (where items are created with null hashes).

If we do create a transient OfferPayload hash field, fillOfferBookListItems() can load the view with non-null hashes, and this null hash check can be removed.

sqrrm (approved):
ACK

PR #5671, opened 2021-08-24
Pricenode: Remove the Exmo exchange as a price provider

wiz:
Since it seems liquidity on Exmo has dried up for XMR/BTC pair, we should stop using them as a price provider


ghost (approved):

devinbileck (approved):
ACK

ripcurlx (approved):
utACK

PR #5672, opened 2021-08-24
Add SWIFT payment method

### LINKED ISSUE #1789: Add SWIFT (int. wire transfer) as payment method
I am wondering if we should add international wire transfer (SWIFT).

There are some issues thought:
- Banking fees are rather high (50-100 USD) which makes that very expensive if used with low trade limits (about 400 USD atm for new accounts).
- Fees are not clearly determined and receiver will also get charged by a fee. Furthermore some banks allow to define the fee model. There are 3 options: 'Sender pays', 'receiver pays' or fee is 'shared'. I think default is shared but users who manage to set that to 'receiver pay fee' can save much money on cost of the peer.
- It seems that the receivers address is required which adds some privacy and security risks to the traders. But not sure if the address is verified. Many users change real address when moving house and do not update address at bank (often the bank webpage does not support that anyway).
- Transfer takes 3-5 working days but can take even longer
- It is not very clear to me how to find out if an intermediary bank is needed. I assume the banks figure it out themself and add on demand intermediary banks (and increase total fees by that) or it need to be defined by the users, but that will prob. not work as you don't know in advance the combination of the sellers and buyers banks.
- If the bank cannot complete the transfer more undfined costs may apply to the sender. 

So all in all it seems like a can of worms and thats why we never added it but I am still wondering if we should offer it and make it very clear to the user which risks and issues are connected to it. There are countries with no Bisq market developed yet and that payment method could help to connect those isolated users to trade between them. Also in such countries they pay a huge premium anyway on LBTC or other exchanges, so the high fees might be not an issue for them. It could also help to bootstrap a national tranfer market once there are a few traders using Bisq there.

I think chargeback risk is very low, so that is a good part.

We would need to use a high trade limit otherwise the costs are way too high (25% of the trade if the costs are 100 USD for a 400 USD tx). We could allow 0.25 BTC for new accounts (about 1600 USD) then the 100 USD fee would be "only" 6%. Once the account has 2 months then its 1 BTC and then the fee with 1.5% is actaully pretty reasonable.

Another problem is how to deal with the fee mode. There is an incentive that the sender tries to let the receiver pay the fee if his bank offers that option. So maybe the best would be that we use that as the default case. The extra costs for the fee for the Fiat receiver can be priced in the offer price.   There can be though then cases where the sender cannot select that and then the seller has the disadvantage. 

Arbitration with the fee issues migth become an issue. But I think if we use the model that the reciver pays all fees and communicate it clearly that fees are not determined it should be ok. User has to be aware of an extra risk factor here. 
A sender who pays more fee because his bank does not allow selection of the model or is instranparent with fees has to complain at his bank. 

@alfsbs What do you think?

ghost:
Implements SWIFT international wire transfer per spec here: https://github.com/bisq-network/bisq/issues/1789#issuecomment-903172680

Closes #1789

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -3280,6 +3280,27 @@ payment.secret=Secret question
 payment.answer=Answer
 payment.wallet=Wallet ID
 payment.capitual.cap=CAP Code
+
+# suppress inspection "UnusedProperty"
+payment.swift.headline=International SWIFT Wire Transfer
+payment.swift.title.bank=Receiving Bank
+payment.swift.title.intermediary=Intermediary Bank (click to expand)
+payment.swift.country.bank=Receiving Bank Country
+payment.swift.country.intermediary=Intermediary Bank Country
+payment.swift.swiftCode.bank=Receiving Bank SWIFT Code
+payment.swift.swiftCode.intermediary=Intermediary Bank SWIFT Code
+payment.swift.name.bank=Receiving Bank name
+payment.swift.name.intermediary=Intermediary Bank name
+payment.swift.branch.bank=Receiving Bank branch
+payment.swift.branch.intermediary=Intermediary Bank branch
+payment.swift.address.bank=Receiving Bank address
+payment.swift.address.intermediary=Intermediary Bank address
+payment.swift.address.beneficiary=Beneficiary address
+payment.swift.phone.beneficiary=Beneficiary phone number
+payment.swift.account=Account No. (or IBAN)
+payment.swift.use.intermediary=Use Intermediary Bank
+payment.swift.showPaymentInfo=Show Payment Information...
+
 payment.amazon.site=Buy giftcard at
 payment.ask=Ask in Trader Chat
 payment.uphold.accountId=Username or email or phone no.
@@ -3410,6 +3431,36 @@ payment.account.amazonGiftCard.addCountryInfo={0}\n\
   Please enter your Amazon Gift Card Country to update your account data.\n\
   This will not affect your account age status.
 payment.amazonGiftCard.upgrade.headLine=Update Amazon Gift Card account
+
+payment.swift.info=Please make sure to include as much information as possible for SWIFT payments. Not all the information \
+may be required for some payments, however as SWIFT is international you will not know from which country your offer is \
+taken and therefore, what information is needed. As such please provide as much information to make the process as smooth \
+as possible. \n\
+\n\
+Most banks charge fees for sending and receiving SWIFT payments. Bisq requires BTC Buyers to send SWIFT payments using \
+the shared fee model (SHA). If you send a payment using another mode for example Beneficiary pays fees (BEN) you will \
+likely be penalized. Therefore, please take care to select the correct fee model when sending payment. \n\
+\n\
+SWIFT payments require more information than any other payment method on Bisq. Please ensure you communicate with your \
+trading peer effectively and promptly to ensure a smooth trading experience for all. As is the case for all payment \
+methods please do leave any reference blank. If a reference is required please use your 'Account Owner Name'.
+
+payment.swift.info.buyer=Most banks charge fees for sending and receiving SWIFT payments. Bisq requires BTC Buyers to \
+send SWIFT payments using the shared fee model (SHA). If you send a payment using another mode for example Beneficiary \
+pays fees (BEN) you will likely be penalized. Therefore, please take care to select the correct fee model when sending \
+payment and make yourself aware of what fees you will likely pay in advance of taking the trade.\n\
+\n\
+SWIFT payments require more information than any other payment method on Bisq. Please ensure you communicate with your \
+trading peer effectively and promptly to ensure a smooth trading experience for all. As is the case for all payment \
+methods please do leave any reference blank. If a reference is required please use your 'Account Owner Name'.
+
+payment.swift.info.seller=Most banks charge fees for sending and receiving SWIFT payments. Bisq requires BTC Buyers to \
+send SWIFT payments using the shared fee model (SHA). If you receive a SWIFT payment that does not use the SHA model, \
+for example Beneficiary pays fees (BEN), please open a mediation ticket. Please ensure you are aware of what fees you \
+might incur as the beneficiary of a SWIFT payment.\n\
+\n\
+SWIFT payments require more information than any other payment method on Bisq. Please ensure you communicate with your \
+trading peer effectively and promptly to ensure a smooth trading experience for all.
 
 payment.usPostalMoneyOrder.info=Trading using US Postal Money Orders (USPMO) on Bisq requires that you understand the following:\n\
 \n\
@@ -3581,6 +3632,8 @@ AMAZON_GIFT_CARD=Amazon eGift Card
 BLOCK_CHAINS_INSTANT=Altcoins Instant
 # suppress inspection "UnusedProperty"
 CAPITUAL=Capitual
+# suppress inspection "UnusedProperty"
+SWIFT=SWIFT International Wire Transfer
 
 # Deprecated: Cannot be deleted as it would break old trade history entries
 # suppress inspection "UnusedProperty"
@@ -3639,6 +3692,8 @@ AMAZON_GIFT_CARD_SHORT=Amazon eGift Card
 BLOCK_CHAINS_INSTANT_SHORT=Altcoins Instant
 # suppress inspection "UnusedProperty"
 CAPITUAL_SHORT=Capitual
+# suppress inspection "UnusedProperty"
+SWIFT_SHORT=SWIFT
 
 # Deprecated: Cannot be deleted as it would break old trade history entries
 # suppress inspection "UnusedProperty"
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -3432,35 +3432,27 @@ payment.account.amazonGiftCard.addCountryInfo={0}\n\
   This will not affect your account age status.
 payment.amazonGiftCard.upgrade.headLine=Update Amazon Gift Card account
 
-payment.swift.info=Please make sure to include as much information as possible for SWIFT payments. Not all the information \
-may be required for some payments, however as SWIFT is international you will not know from which country your offer is \
-taken and therefore, what information is needed. As such please provide as much information to make the process as smooth \
-as possible. \n\
+payment.swift.info=Carefully review the core guidelines for using SWIFT on Bisq:\n\
 \n\
-Most banks charge fees for sending and receiving SWIFT payments. Bisq requires BTC Buyers to send SWIFT payments using \
-the shared fee model (SHA). If you send a payment using another mode for example Beneficiary pays fees (BEN) you will \
-likely be penalized. Therefore, please take care to select the correct fee model when sending payment. \n\
+- fill all fields completely and accurately \n\
+- buyer must send payment in currency specified by the offer maker \n\
+- buyer must use the shared fee model (SHA) \n\
+- buyer and seller may incur fees, so they should check their bank's fee schedules beforehand \n\
 \n\
-SWIFT payments require more information than any other payment method on Bisq. Please ensure you communicate with your \
-trading peer effectively and promptly to ensure a smooth trading experience for all. As is the case for all payment \
-methods please do leave any reference blank. If a reference is required please use your 'Account Owner Name'.
+SWIFT is more sophisticated than other payment methods, so please take a moment to review full guidance on the wiki [HYPERLINK:https://bisq.wiki/SWIFT].
 
-payment.swift.info.buyer=Most banks charge fees for sending and receiving SWIFT payments. Bisq requires BTC Buyers to \
-send SWIFT payments using the shared fee model (SHA). If you send a payment using another mode for example Beneficiary \
-pays fees (BEN) you will likely be penalized. Therefore, please take care to select the correct fee model when sending \
-payment and make yourself aware of what fees you will likely pay in advance of taking the trade.\n\
+payment.swift.info.buyer=To buy bitcoin with SWIFT, you must:\n\
 \n\
-SWIFT payments require more information than any other payment method on Bisq. Please ensure you communicate with your \
-trading peer effectively and promptly to ensure a smooth trading experience for all. As is the case for all payment \
-methods please do leave any reference blank. If a reference is required please use your 'Account Owner Name'.
+- send payment in the currency specified by the offer maker \n\
+- use the shared fee model (SHA) to send payment\n\
+\n\
+Please review further guidance on the wiki to avoid penalties and ensure smooth trades [HYPERLINK:https://bisq.wiki/SWIFT].
 
-payment.swift.info.seller=Most banks charge fees for sending and receiving SWIFT payments. Bisq requires BTC Buyers to \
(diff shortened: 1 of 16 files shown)


m52go:
@jmacxx thanks for doing this! Just ran through it on regtest.

Would it be possible to have the "Special info about any of the inputs required for creating the payment account" text appear when the blank input form shows up? As it is, the user will be confronted with a long form and many inputs, but the guidance on how to handle those inputs comes after they're done filling out the inputs.

With that said, I'm just now noticing that this is the case for all payment methods...seems kind of backwards to have the helper pop-up appear after the user is done with the form. Maybe that's a different task / PR, not sure.

@pazza83 I think the wiki should be updated with all the text in this PR, at a minimum, along with any other information you think would be helpful. That would also allow us to replace some of the text in the pop-ups with a link to the wiki.

pazza83:
Thanks for this @jmacxx 

@m52go yes agreed. I will create a wiki for SWIFT and include the relevant info.

The link to the wiki page will be: https://bisq.wiki/SWIFT

ghost:
> Would it be possible to have the "Special info about any of the inputs required for creating the payment account" text appear when the blank input form shows up?

@m52go no problem, made the change for SWIFT via commit https://github.com/bisq-network/bisq/pull/5672/commits/53bc2b2ec9e79b51bc14252774ab3474fd456469

m52go:
Super :+1:

I would like to refine some of the text in the modals and add links but that can only be done once @pazza83 finishes wiki.

Since this is a rather complex payment method for large amounts, it would be great to get more eyes on it. Below are screenshots. 

@dav1dpgit @mpolavieja @conza88 would you mind taking a look to see if it all looks good to you?

### Popup shown when a user chooses to create new SWIFT account

Screenshot from 2021-08-24 13-03-39

### New account creation (1/3)

Screenshot from 2021-08-24 13-10-49

### New account creation; this part is optional (2/3)

Screenshot from 2021-08-24 13-11-04

### New account creation (3/3)

Screenshot from 2021-08-24 13-11-17

### Message shown to buyer

Screenshot from 2021-08-24 13-02-25

### Message shown to seller

Screenshot from 2021-08-24 13-02-47

Conza88:
A quick comment — have done a few of these. The bank A I was sending from to bank B, was meant to go AUD to stay AUD although in a foreign bank. 

When the funds arrived, receiver noticed less than should have been there. Upon his bank B internal investigation, it was routed via a US bank as intermediary. The bastard's (banks) had erroneously converted to USD then back to AUD in the middle. Taking a cut. SMH. Took like 4 days to get them traversed, bank B had to send back to intermediary bank.

The recommendation was to put in description when sending in future: DO NOT CONVERT. This is how retarded they are. 

Now if that's worth including lol.. maybe a big giveaway, increase in 10x people "saying don't convert" 🤷‍♂️. But could be added a request in optional comments.

Potentially confusing the no trade or ID for other trades, but message about don't leave anything blank. 

I also originally was confused about address, if was receivers but actually their bank address info etc.

WillyGoldman:
Swift should be a great addition.

Wordsmithing first modal dialog:

-------------
Please make sure to include as much information as possible for SWIFT
payments.  Some information may not be required for this SWIFT payment- 
provide this infromation anyway. SWIFT is international and different 
countries require different information. To be safe, include as much 
information as possible.

Most banks charge fees for sending or receiving SWIFT payments. Bisq requires
BTC Buyers to send SWIFT payments using the "shared fee model" (SHA). Do not use
another mode.  Modes like "Beneficiary pays fees" (BEN) violate the rules of Bisq
and can be penzliazed from the buyers bitcoin secuirty deposit.

SWIFT payments require more information than any other payment method on Bisq.
Please communicate with your trading peer quickly and clearly. Communication will
ensure a good trading experience. All Bisq payment methods should always leave
the "reference" or "note" field blank.  If this field is required please 
use your own "Account owner name".
-------------------------

Sorry, third paragraph is misformatted because SWIFT in caps invokes bold....

My goal in all edits is to simplify vocabulary and sentence structure to ensure clear understanding for
users that have English as a second language.  Shorter sentences, less pronouns, simpler vocabulary is my goal.

I think the other dialogs are essentially the same text rearanged. 

Swift looks like it will help many people!

dav1dpgit:
Some comments:

* Receiving banks usually convert incoming SWIFT wires into the domestic currency at a steep premium.  If this is not desired by the recipient he/she must instruct the SWIFT sender to include "DO NOT CONVERT" in the comment field.  In this situation the recipient **must** be able to handle multicurrency incoming SWIFT payments.  Per the comment from @Conza88 it may be useful to notify at the SWIFT account setup something along the lines of "Be aware:  Your receiving bank may convert incoming SWIFT wires to the domestic currency without prior notice unless specifically instructed not to do so by the SWIFT sender and/or receiver. Bank SWIFT wire conversions may result in a different quantity and currency received vs. the bisq trade quantity and currency.  Discuss with your bank how to prevent this from occurring before initiating a SWIFT transaction."

* I see at the end of other thread (#1789 ) and above in 1 of 3, when setting up the bisq SWIFT settlement method the bisq user must choose the currency.  Will this necessitate a separate SWIFT settlement method for every currency one is willing to receive?  I do not think this is optimal, particularly as SWIFT should be multinational/multicurrency.  Some SWIFT users can accept multicurrency SWIFT payments, and others will simply let their bank automatically convert SWIFT payments into the domestic currency.  I suggest a checkbox set of currencies acceptable (similar to the REVOLUT) when setting up the SWIFT settlement method.

ghost:
Functionality seems ready by now.  Popup text still needs polishing per @m52go :

> I would like to refine some of the text in the modals and add links but that can only be done once @pazza83 finishes wiki.

pazza83:
I have completed the first draft of the [SWIFT wiki](https://bisq.wiki/SWIFT).

@jmacxx please can you add a reference to the wiki on each of the pop ups. Cheers.

Also if you can post a full screen shot of the account addition process I will replace the current one I used for the wiki: https://bisq.wiki/File:SWIFT_1.png

ghost:
> ... if you can post a full screen shot of the account addition process ...

Here u go @pazza83 


image

Screenshot from 2021-09-07 21-37-12

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -3411,6 +3432,36 @@ payment.account.amazonGiftCard.addCountryInfo={0}\n\
...
+taken and therefore, what information is needed. As such please provide as much information to make the process as smooth \
+as possible. \n\
+\n\
+Most banks charge fees for sending and receiving SWIFT payments. Bisq requires BTC Buyers to send SWIFT payments using \
+the shared fee model (SHA). If you send a payment using another mode for example Beneficiary pays fees (BEN) you will \
+likely be penalized. Therefore, please take care to select the correct fee model when sending payment. \n\
+\n\
+SWIFT payments require more information than any other payment method on Bisq. Please ensure you communicate with your \
+trading peer effectively and promptly to ensure a smooth trading experience for all. As is the case for all payment \
+methods please do leave any reference blank. If a reference is required please use your 'Account Owner Name'.
```
```suggestion
payment.swift.info=Carefully review the core guidelines for using SWIFT on Bisq:\n\
\n\
- fill all fields completely and accurately \n\
- buyer must send payment in currency specified by the offer maker \n\
- buyer must use the shared fee model (SHA) \n\
- buyer and seller may incur fees, so they should check their bank's fee schedules beforehand \n\
\n\
SWIFT is more sophisticated than other payment methods, so please take a moment to review full guidance on the wiki [HYPERLINK:https://bisq.wiki/SWIFT].
```

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -3411,6 +3432,36 @@ payment.account.amazonGiftCard.addCountryInfo={0}\n\
...
+
+payment.swift.info.buyer=Most banks charge fees for sending and receiving SWIFT payments. Bisq requires BTC Buyers to \
+send SWIFT payments using the shared fee model (SHA). If you send a payment using another mode for example Beneficiary \
+pays fees (BEN) you will likely be penalized. Therefore, please take care to select the correct fee model when sending \
+payment and make yourself aware of what fees you will likely pay in advance of taking the trade.\n\
+\n\
+SWIFT payments require more information than any other payment method on Bisq. Please ensure you communicate with your \
+trading peer effectively and promptly to ensure a smooth trading experience for all. As is the case for all payment \
+methods please do leave any reference blank. If a reference is required please use your 'Account Owner Name'.
```
```suggestion
payment.swift.info.buyer=To buy bitcoin with SWIFT, you must:\n\
\n\
- send payment in the currency specified by the offer maker \n\
- use the shared fee model (SHA) to send payment\n\
\n\
Please review further guidance on the wiki to avoid penalties and ensure smooth trades [HYPERLINK:https://bisq.wiki/SWIFT].
```

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -3411,6 +3432,36 @@ payment.account.amazonGiftCard.addCountryInfo={0}\n\
...
+
+payment.swift.info.seller=Most banks charge fees for sending and receiving SWIFT payments. Bisq requires BTC Buyers to \
+send SWIFT payments using the shared fee model (SHA). If you receive a SWIFT payment that does not use the SHA model, \
+for example Beneficiary pays fees (BEN), please open a mediation ticket. Please ensure you are aware of what fees you \
+might incur as the beneficiary of a SWIFT payment.\n\
+\n\
+SWIFT payments require more information than any other payment method on Bisq. Please ensure you communicate with your \
+trading peer effectively and promptly to ensure a smooth trading experience for all.
```
```suggestion
payment.swift.info.seller=SWIFT senders are required to use the shared payment model (SHA) to send payments.\n\
\n\
If you receive a SWIFT payment that does not use SHA, open a mediation ticket.\n\
\n\
Please review further guidance on the wiki to avoid penalties and ensure smooth trades [HYPERLINK:https://bisq.wiki/SWIFT].
```

m52go:
I've edited the [wiki article](https://bisq.wiki/SWIFT) and offered suggestions for the copy in this pull request.

Graphics are below.

Notes:
- As I understand @dav1dpgit's comment, "DO NOT CONVERT" should go in the comments for the wire transfer, not in the Bisq payment account's "additional info" section.

## Create payment account

Screenshot from 2021-09-08 14-49-41

## Buying BTC

Screenshot from 2021-09-08 14-50-28

## Selling BTC

Screenshot from 2021-09-08 14-50-54

desktop/src/main/java/bisq/desktop/components/paymentmethods/SwiftForm.java
```diff
@@ -0,0 +1,346 @@
...
+    private LengthValidator swiftValidator = new LengthValidator(11, 11);
+    private LengthValidator accountNrValidator = new LengthValidator(2, 40);
+    private LengthValidator addressValidator = new LengthValidator(1, 100);
+
+    public SwiftForm(PaymentAccount paymentAccount,
+                     AccountAgeWitnessService accountAgeWitnessService,
+                     InputValidator defaultValidator, GridPane gridPane, int gridRow, CoinFormatter formatter) {
+        super(paymentAccount, accountAgeWitnessService, defaultValidator, gridPane, gridRow, formatter);
+        this.formData = ((SwiftAccount) paymentAccount).getPayload();
+        this.useIntermediaryCheck = new CheckBox(Res.get("payment.swift.use.intermediary"));
```
ripcurlx:
```suggestion
        this.useIntermediaryCheck = new AutoTooltipCheckBox(Res.get("payment.swift.use.intermediary"));
```
Should be 
Bildschirmfoto 2021-09-09 um 10 21 18
Instead of 
Bildschirmfoto 2021-09-09 um 10 19 33

desktop/src/main/java/bisq/desktop/components/paymentmethods/SwiftForm.java
```diff
@@ -0,0 +1,346 @@
...
+        label = Res.get(Res.get(PHONE + BENEFICIARYPOSTFIX));
+        InputTextField beneficiaryPhoneField = addInputTextField(gridPane, ++gridRow, label);
+        beneficiaryPhoneField.setPromptText(label);
+        beneficiaryPhoneField.setValidator(defaultValidator);
+        beneficiaryPhoneField.textProperty().addListener((ov, oldValue, newValue) -> {
+            formData.setBeneficiaryPhone(newValue.trim());
+            updateFromInputs();
+        });
+
+        label = Res.get("payment.shared.optionalExtraInfo");
```
ripcurlx:
```suggestion
        label = Res.get("payment.shared.optionalExtra");
```
Wrong key
Bildschirmfoto 2021-09-09 um 10 27 58

ripcurlx (changes requested):
NACK - Please see my comments. As I have the changes locally already I'll push them to the branch as soon as I'm done with the Regtest test run. So please wait for doing the changes on your own.

ripcurlx:
NIT: would be great to get rid of the grey line at the bottom and to be able to copy the fields to prevent any typos.

Bildschirmfoto 2021-09-09 um 10 37 59

ripcurlx:
Bildschirmfoto 2021-09-09 um 10 40 12
As I'm not familiar with SWIFT: Is it necessary to display on the BTC seller side more information from the buyer as just the name to confirm the payment received?

ripcurlx:
Besides the issues mentioned above I was able to complete a trade as expected. I'm not 100% happy with the default design of the TitledPane, but it doesn't look too bad in dark mode and it is just on account creation anyways.

ghost:
Thanks @ripcurlx, requested changes made.

>  Is it necessary to display on the BTC seller side more information from the buyer as just the name to confirm the payment received?

Buyer name and amount is ok AFAIK.

Conza88:
Just checking hitting "copy" it gets just the name, not the full spiel about international swift transfer?

ghost:
> Just checking hitting "copy" it gets just the name, not the full spiel about international swift transfer?

"Copy" copies the contents of the field next to the copy button, per standard GUI practice.  The same functionality exists on all payment methods, for example:

image

image

Conza88:
The whole purpose of the "copy" is to paste into the fiat on ramp / bank transfer to not make a mistake.

Replicating the Swift part is redundant and annoying for customer to remove. 

That's the full customer experience. UX design standards take a back seat to this end purpose. Or at least should, because the UX is meant to aide experience not hinder.

ghost:
This is the seller confirming payment received.  Same principle for all payment methods.

Conza88:
Ah ok. My bad. Apologies.
As long as sender of fiat, first screen is not the full spiel just the necessary account name etc, then all good

ghost:
@m52go when placing a buy offer, user is reminded to use shared fee model:

image

When sending payment, we instruct users to pay own bank fees, (a default message for most payment methods unless we customize it).

image

`If your bank charges you any fees to make the transfer, you are responsible for paying those fees`

Might this be confusing or contradictory, do you think the wording of that line needs to be made clearer for SWIFT?

m52go:
Shoot, good catch. If there is there a way to make a custom version of [that line](https://github.com/bisq-network/bisq/blob/b8f2ac7a68e67d6108c76349af6b48618f565751/core/src/main/resources/i18n/displayStrings.properties#L661) just for SWIFT...ideally it should be replaced with something like `Make sure to use the SHA (shared fee model) to send the SWIFT payment. See more details at [HYPERLINK:https://bisq.wiki/SWIFT#Use_the_correct_fee_option].`

If customizing that line isn't feasible, let me know, and we might have to get more creative.

ghost:
Change made.

image

m52go:
Awesome, thank you @jmacxx!

ripcurlx (approved):

PR #5676, opened 2021-08-28
Adopt is no more, long live Zulu!

geertjanw:

Your PR description here.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
       uses: actions/setup-java@v2
       with:
         java-version: '11'
-        distribution: 'adopt'
+        distribution: 'zulu'
 
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew


ripcurlx:
@geertjanw Is there a big benefit using Zulu openJDK version instead of Adopt?

geertjanw:
Adopt no longer exists. Zulu does. That's a big difference.

geertjanw:
It would also be good to not only set '11', i.e., that is whatever is the latest release, but also a specific version, e.g., 11.0.3, because then if a build fails, and you have 11.0.3 green and 11 red, it will mean that the problem is not with your code but the latest version of the JDK (and Zulu supports not only the latest JDK at this time, while Temurin supports only the latest JDK version at this time).

Here's an example of best practise, i.e., using a fixed version as well as the latest version of the JDK:

https://github.com/HdrHistogram/HdrHistogram/blob/master/.github/workflows/maven.yml

ghubstan:
I assume "Adopt no longer exists" is referring to its move to the [Eclipse Foundation](https://blog.adoptopenjdk.net/2021/03/transition-to-eclipse-an-update).  But Adopt still exists, and the Bisq desktop packager's Adopt JDK dist URLs still work as of today.

All I know so far about [Zulu](https://www.azul.com/downloads/openjdk) is that it is an OpenJDK build.   I didn't know about it until last week, when I looked into the possibility of using [JReleaser](https://github.com/jreleaser/jreleaser) for packaging Bisq with bundled JDK.  I had some problems downloading the Adopt JDK distributions with the  `org.jreleaser.jdks` plugin as per https://github.com/jreleaser/jreleaser/issues/363.

aalmiray:
FWIW regarding jreleaser/jreleaser#363 a fix is in progress but will become available at the end of September when JReleaser 0.7.0 is scheduled for release. 

Switching to Zulu builds will make the `org.jreleaser.jdks` plugin work as expected. Why? Because Zulu builds unpack with an unique directory name per platform where as most JDK vendors (Oracle, Adopt, BellSoft, Amazon, Alibaba) use the same directory name (the JDK version) without any platform information in the name.

ghubstan:
> Switching to Zulu builds will make the `org.jreleaser.jdks` plugin work as expected.

Thanks, @aalmiray , for the quick response to the issue and remark here.

Yes, the `org.jreleaser.jdks` plugin works great with Zulu builds, as described in @aalmiray 's [documentation](https://jreleaser.org/guide/latest/tools/jdks-gradle.html).

ripcurlx (approved):
As I don't see any drawbacks using Zulu (I used it myself in the past already) and jreleaser is using it also it might be a good first start to try it out here first.

ripcurlx:
> It would also be good to not only set '11', i.e., that is whatever is the latest release, but also a specific version, e.g., 11.0.3, because then if a build fails, and you have 11.0.3 green and 11 red, it will mean that the problem is not with your code but the latest version of the JDK (and Zulu supports not only the latest JDK at this time, while Temurin supports only the latest JDK version at this time).
> 
> Here's an example of best practise, i.e., using a fixed version as well as the latest version of the JDK:
> 
> https://github.com/HdrHistogram/HdrHistogram/blob/master/.github/workflows/maven.yml

Feel free to provide a PR for it :smile:

PR #5681, opened 2021-09-01
Allow user to initiate arbitration once locktime has expired

### LINKED ISSUE #4101: Allow for opening of arbitration after a certain time
### Description

Currently one can't open Arbitration without first going through Mediation.

#### Version

Bisq v.1.2.9

### Steps to reproduce

Alice and Bob enter into a trade and one of them opens Mediation. If for any reason the Mediator doesn't act, the traders can't move into arbitration. 

### Expected behaviour

After the locked payout becomes available (10 days for altcoins, 20 for national currencies) the user should be able to bypass Mediation and open Arbitration.

### Actual behaviour

The user gets stuck waiting for Mediation. If for some reason the Mediator is unresponsive the user is left without option inside the software

#### Additional info

[Relevant case](https://bisq.community/t/request-arbitration-with-unresponsive-mediator/9155)

ghost:
### Problem description:

Normal sequence of events is user opens mediation, mediation completes and then user has option to open arbitration once locktime has expired.  In the unusual cases, users might not be able to open mediation, or mediation might not close (as described in #4101).  When that happens, arbitration cannot be initiated.

### Ruleset for opening mediation or arbitration

CTRL-O opens a **mediation ticket** when either:

- trade is not in dispute,
- mediation has been initiated and the locktime has not expired.

else CTRL-O opens an **arbitration ticket** when either:

- trade mediation has closed,
- mediation result was rejected,
- locktime has expired.

To accomodate the fact that users may need to open arbitration manually, this change allows it upon locktime expiry.

Closes #4101

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/PendingTradesDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/PendingTradesDataModel.java
@@ -493,12 +493,13 @@ public class PendingTradesDataModel extends ActivatableDataModel {
         }
         Trade.DisputeState disputeState = trade.getDisputeState();
         DisputeManager<? extends DisputeList<Dispute>> disputeManager;
-        boolean useMediation;
-        boolean useRefundAgent;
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
utACK

PR #5682, opened 2021-09-01
Reduce # of hash calculations in UI OfferBook view

ghubstan:
We can cache an offer payload hash as soon as its `offerFeePaymentTxId`is set.  (The payload hash cannot be calculated until the object can be transformed into a protobuf message, which requires a non-null `offerFeePaymentTxId`.)

Another benefit is removal of the payload hash argument from the`OfferBookListItem` constructor.

Changes include

- `OfferPayload` Added `transient byte[] hash` field + getter method (where hash is calculated and cached).

- `OfferBookService` Removed `P2PDataStorage.ByteArray hashOfPayload` parameter from `OfferBookChangedListener` listener methods `onAdded` & `onRemoved`.  (Hash is cached in `OfferPayload`.)

- `P2PDataStorage` Added null check to `ByteArray` class constructor.

- `OfferBook` Adjusted for change to `OfferBookChangedListener`, and removed redundant payload hash null checks.

- `TakeOfferDataModel` and `MarketAlerts` Adjusted for change to `OfferBookChangedListener`.

- `OfferBookListItem` Removed overloaded constructor with `@Nullable P2PDataStorage.ByteArray hashOfPayload` parameter.  (Field value is set from cached offer payload hash.)

- `OfferBookViewModelTest` and `OfferMaker`  Adjusted test and test fixture:  do not attempt to create offer payloads without an `offerFeePaymentTxId`.

_Follow up to the [last PR in series](https://github.com/bisq-network/bisq/pull/5666)  to support API `editoffer` method._

### DIFF
--- a/core/src/main/java/bisq/core/notifications/alerts/market/MarketAlerts.java
+++ b/core/src/main/java/bisq/core/notifications/alerts/market/MarketAlerts.java
@@ -32,8 +32,6 @@ import bisq.core.provider.price.PriceFeedService;
 import bisq.core.user.User;
 import bisq.core.util.FormattingUtils;
 
-import bisq.network.p2p.storage.P2PDataStorage;
-
 import bisq.common.crypto.KeyRing;
 import bisq.common.util.MathUtils;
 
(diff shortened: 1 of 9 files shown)


ripcurlx (approved):
ACK - Tested it on Mainnet and everything seems to be as before.

PR #5683, opened 2021-09-02
Override/disable gRPC call rate meters in test harness driver

ghubstan:
Ad-hoc API testers need to be able to run the test harness without interference from the production api method call rate meters.  This change overrides and disables most call rate meters when the test harness is run from the `ApiTestMain` driver (no jupiter tests).


ripcurlx (approved):
utACK

PR #5685, opened 2021-09-06
Let API user set currencies in some payment account json forms

ghubstan:
Several payment methods support multiple trade currencies and a selected trade currency, but the API has not let CLI users specify them in the json form passed to the `createpaymentacct` command.

This change adds `tradeCurrencies` and `selectedTradeCurrency` fields to the appropriate json forms.

Summary of changes:

- Added support for reading/writing trade currency info from/to json forms to `PaymentAccountTypeAdapter`.

- Added new boolean convenience methods to `PaymentAccount` to aid in correctly defining and validating pay account json from currency lists and selections.

- Adjusted `CorePaymentAccountsService`, `CreateOfferService` `MutableOfferDataModel`, `MutableOfferView`, `MutableOfferViewModel`, `TakeOfferDataModel` to `PaymentAccount` refactoring.

- Added `Res # List<String> getWrappedAsList(String key, int wrapLength)` and `Utilities # List<String> toListOfWrappedStrings(String s, int wrapLength)` to convert long, multi-line i18n strings to wrapped strings the API can use.

- Added constants to apitest's `AbstractPaymentAccountTest` for new `SwiftAccount` test case to be added after [PR 5672](https://github.com/bisq-network/bisq/pull/5672) is merged.

- Added asserts to existing  `CreatePaymentAccountTest` cases:  check new account instances' `tradeCurrencies` and `selectedTradeCurrency` field values written to payment account creation json forms.

- Added new `CreatePaymentAccountTest` cases, including commented out case `testCreateSwiftAccount` --  in anticipation of the [new Swift payment method](https://github.com/bisq-network/bisq/pull/5672) merge.

- Added `apitest/src/test/resources/logback.xml` for configuring apitest case logging.

- Removed dead/commented code in some of the touched classes.

Based on branch `detault-api-testharness-callrate-config`, which was merged earlier today in [PR 5683](https://github.com/bisq-network/bisq/pull/5683)

### DIFF
--- a/apitest/src/test/java/bisq/apitest/method/payment/AbstractPaymentAccountTest.java
+++ b/apitest/src/test/java/bisq/apitest/method/payment/AbstractPaymentAccountTest.java
@@ -146,6 +146,7 @@ public class AbstractPaymentAccountTest extends MethodTest {
                 Object.class);
         assertNotNull(emptyForm);
 
+        // TODO remove 'false' condition to enable creation of SWIFT accounts in future PR.
         if (false && paymentMethodId.equals("SWIFT_ID")) {
             assertEquals(getSwiftFormComments(), emptyForm.get(PROPERTY_NAME_JSON_COMMENTS));
         } else {
--- a/apitest/src/test/java/bisq/apitest/method/payment/AbstractPaymentAccountTest.java
+++ b/apitest/src/test/java/bisq/apitest/method/payment/AbstractPaymentAccountTest.java
@@ -1,6 +1,7 @@
 package bisq.apitest.method.payment;
 
 import bisq.core.api.model.PaymentAccountForm;
+import bisq.core.locale.FiatCurrency;
 import bisq.core.locale.Res;
 import bisq.core.locale.TradeCurrency;
 import bisq.core.payment.PaymentAccount;
@@ -17,10 +18,13 @@ import java.io.IOException;
 import java.io.OutputStreamWriter;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import lombok.extern.slf4j.Slf4j;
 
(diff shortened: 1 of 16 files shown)


core/src/main/java/bisq/core/api/model/PaymentAccountTypeAdapter.java
```diff
@@ -350,14 +378,61 @@ private boolean didReadTradeCurrenciesField(JsonReader in,
...
+        else if (account.hasPaymentMethodWithId(UPHOLD_ID))
+            return getTradeCurrenciesInList(currencyCodes, getAllUpholdCurrencies());
+        else
+            return Optional.empty();
+    }
+
+    private boolean didReadSelectedTradeCurrencyField(JsonReader in,
+                                                      PaymentAccount account,
+                                                      String fieldName) {
+        if (fieldName.equals("selectedTradeCurrency")) {
```
sqrrm: Prefer return early, eg
```
if (!fieldName.equals("selectedTradeCurrency")) {
  return false;
}
...
```

core/src/main/java/bisq/core/api/model/PaymentAccountTypeAdapter.java
```diff
@@ -350,14 +378,61 @@ private boolean didReadTradeCurrenciesField(JsonReader in,
...
+            String fieldValue = nextStringOrNull(in);
+            if (fieldValue != null && !fieldValue.isEmpty()) {
+                Optional<TradeCurrency> tradeCurrency = getTradeCurrency(fieldValue.toUpperCase());
+                if (tradeCurrency.isPresent()) {
+                    account.setSelectedTradeCurrency(tradeCurrency.get());
+                } else {
```
sqrrm: It does read the field, but should it return true even when it's not defined?

apitest/src/test/java/bisq/apitest/method/payment/AbstractPaymentAccountTest.java
```diff
@@ -125,7 +145,13 @@ protected final void verifyEmptyForm(File jsonForm, String paymentMethodId, Stri
                 PAYMENT_ACCOUNT_FORM.toJsonString(jsonForm),
                 Object.class);
         assertNotNull(emptyForm);
-        assertEquals(PROPERTY_VALUE_JSON_COMMENTS, emptyForm.get(PROPERTY_NAME_JSON_COMMENTS));
+
+        if (false && paymentMethodId.equals("SWIFT_ID")) {
```
sqrrm, line 150: Maybe a TODO if this is a placeholder for future swift test

apitest/src/test/resources/logback.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sqrrm, line 1: What does this file do?

sqrrm (commented):
Some comments, looks generally good

apitest/src/test/resources/logback.xml
ghubstan: The `<root level="XXX">` sets the log.level in the test case classes in `apitest/src/test/java/bisq/apitest`, when you run tests within the IDE.  There are some log.debug statements in the jupiter tests.  This file determines whether I see those log.debug outputs or not.  This file has no effect on any other logging, outside of `apitest/src/test/java/bisq/apitest`.

apitest/src/test/java/bisq/apitest/method/payment/AbstractPaymentAccountTest.java
ghubstan: Done.  Commit 429e943a363d76c6a8953d2905c5ab33dee8f6bc

core/src/main/java/bisq/core/api/model/PaymentAccountTypeAdapter.java
ghubstan: If the json field is not defined, it can't be read.   If it is defined, it is always read.

`if (fieldName.equals("selectedTradeCurrency")) ` fails, this method needs to return false so the calling read loop does not  `continue` to the next json field.  

Look at `PaymentAccount read(JsonReader in)`  [Line 215-254]:

If `didReadSelectedTradeCurrencyField()` returns false, the `currentFieldName` might be read in `didReadCommonField()` or `didReadCountryField()`. 

If `didReadSelectedTradeCurrencyField()` returns true, the `currentFieldName` cannot be read in `didReadCommonField()` or `didReadCountryField()`.

core/src/main/java/bisq/core/api/model/PaymentAccountTypeAdapter.java
ghubstan: Done, commit 1a9cdfbdd343050de93df2f2db18b7fd6548cf07

ghubstan: Done.  Commit 1a9cdfbdd343050de93df2f2db18b7fd6548cf07

sqrrm (approved):
utACK

PR #5694, opened 2021-09-13
Improve japanese bank branch validation

### LINKED ISSUE #5688: japanese.validation.regex should accept U+3007
### Description
`JapanBankData.getString("japanese.validation.regex")` only lets kanji/kana pass. As a kanji, currently `U+4E00`  … `U+9FAF` will be seen as valid (according to `JapanBankData.java`), which may be practically good enough. However, there is one special character out of the above range that should be also accepted. Namely:

`U+3007` [ 〇 ] IDEOGRAPHIC NUMBER ZERO

This letter (the kanji version of number 0) is used in many branch names of a specific Japanese bank. Thus one should add the following to that RegEx:

"〇" + // IDEOGRAPHIC NUMBER ZERO (0x3007)

#### Version
1.7.3

### Steps to reproduce
As a test, try to input a random (hypothetical) Japanese Bank account info, with a branch name e.g. "〇一二". This branch name is well-formed but will be refused with an error message.

ripcurlx:
Fixes #5688


wiz:
The diff is so big I have no idea what you actually changed...

wiz:
I think you should probably only change the bottom part of the file where the regex patterns actually changed, otherwise you're rewriting the entire code

ripcurlx:
> I think you should probably only change the bottom part of the file where the regex patterns actually changed, otherwise you're rewriting the entire code

Did you look at the commits :wink:. I've put the formatting fixes in a separate commit and the other two changes in two further commits. So it is easy to see what actually changed.

Liisachan:
What's significant is the 2nd commit c781394dbbd44742eb50ec4f83cb638a2fc7a0a4. It's so simple, no regression seems possible. The 3rd commit 4d62136d6218d4785bc678bf10daf22c4d85f960 is even more trivial, like just improving comments.

The 1st commit 851ef0781209db66a813c9446c4859c1136d7bda is less transparent, but after hiding whitespace changes, I see there are only 2 actual edits:
1) `import java.util.HashMap;` and `import bisq.core.locale.Country;` deleted.
2) `new ArrayList<String>();` -> `new ArrayList<>();`

I guess 1 is simply unused "include"s, and 2 is something like `auto` in C++? Idk, a real Java programmer can review this.

sqrrm (approved):
The formatting and style changes are good. I cannot tell if the regex changes are good but from @wiz and @Liisachan they sound correct.

sqrrm:
utACK

PR #5704, opened 2021-09-18
CLI bug fix: show trade's contract volume, not moving offer volume

ghubstan:
This fixes a an API CLI `gettrade` bug that could have resulted in loss of funds (explained below).  
There are two minor changes:

- Show correct, frozen `Buyer Cost` (trade volume, not offer volume) in CLI console.
This corrects the CLI's displayed fiat trade cost value, which should be trade.volume, not offer.volume.  Offer volume varies with BTC volatility, and the CLI should be showing the trade.volume value instead, frozen when the contract was made.

- Show correct, frozen altcoin `Amount`  (trade volume, not offer volume) in CLI console.
This corrects the CLI's displayed altcoin trade amount value, which should be trade.volume, not offer.volume.  The bug has been hidden by the stability of the BSQ price, and exposed while testing API support for XMR trades.

**The bug could have resulted in loss of funds for traders who used the API to buy BTC in the following scenario:**

	(1) A trader made or took a market-price-margin based offer to buy BTC, using the API.
	
	(2) The BTC price was volatile between the time the trade was taken and the payment was sent.
	
	(3) The payer did not pay the amount shown by the API CLI at the time the trade was created.

		She waited, then returned to the API to find out how much to pay -- minutes to days 
		after the contract was created -- when the trade price (volume) was frozen.  The CLI
		would have shown her a different amount in the `Buyer Cost` column than it showed 
		immediately after the trade was created.

The CLI's `gettrade` command has been showing the moving offer volume in the `Buyer Cost` column -- how 
much the buyer pays.  This value varies with BTC price if the offer is market-price-margin based.

The CLI should be showing the constant trade volume, frozen at the time a trade contract is written.

This bug does not affect BSQ trading because all BSQ trades are at fixed prices.
This bug would have affected XMR trading if the fix was not applied before XMR trading support in the API is released.

To check to see if you paid too much or too little, you could use the desktop app to view your trade history,
and match the correct payment amounts to your bank transfer records.

### DIFF
--- a/cli/src/main/java/bisq/cli/CliMain.java
+++ b/cli/src/main/java/bisq/cli/CliMain.java
@@ -781,7 +781,7 @@ public class CliMain {
             stream.println();
             stream.format(rowFormat, editoffer.name(), "--offer-id=<offer-id> \\", "Edit offer with id");
             stream.format(rowFormat, "", "[--fixed-price=<price>] \\", "");
-            stream.format(rowFormat, "", "[--market-price=margin=<percent>] \\", "");
+            stream.format(rowFormat, "", "[--market-price-margin=<percent>] \\", "");
             stream.format(rowFormat, "", "[--trigger-price=<price>] \\", "");
             stream.format(rowFormat, "", "[--enabled=<true|false>]", "");
             stream.println();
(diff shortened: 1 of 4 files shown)


ripcurlx (approved):
utACK

PR #5708, opened 2021-09-20
Call 'keepfunds' (close trade) on both sides of trade simulation

ghubstan:
UI would prompt Bob to 'keep funds' if started with `apitest` data dirs.
Trade simulation script should complete both sides' protocol instead.


ripcurlx (approved):
utACK

PR #5711, opened 2021-09-20
Fix data handling if dao is deactivated

chimp1984:
Adds a toggle for deactivating the DAO and fixes an issue with P2P data processing if DAO is deactivated.


chimp1984:
Adjusted the DAO view in case DAO is deactivated.

@m52go Could you have a look to the text?
```
dao.news.daoInfo.description=To participate in the Bisq DAO and to use BSQ for discounted trading fees you need to \
  enable the DAO. If the DAO is enabled, it downloads all the missing blocks and verifies the BSQ transactions. \
  This can take some time and consumes 100% of the available resources during the verification process.
```

DAO view in case DAO is deactivated:
Screen Shot 2021-09-23 at 19 00 58

Settings when DAO is deactivated. Shows activate toggle.
Screen Shot 2021-09-23 at 19 00 49

Settings when DAO is activated. 
Screen Shot 2021-09-23 at 19 01 42

m52go:
Suggestion:

```
dao.news.daoInfo.description=To participate in the Bisq DAO and to use BSQ for discounted trading fees, you need to \
  enable the DAO. When the DAO is enabled, Bisq downloads all missing blocks and verifies BSQ transactions. \
  This verification process requires time, during which you may see Bisq use a lot of memory and processing power. This is normal.
```

ghubstan:
`Tested ACK`

Tested DAO view is adjusted for `daoActivated` true or false.		OK

Tested `bisq.properties` editing (daoActivated=true/false).		OK

Tested `--daoActivated` program argument is set  it has higher priority and property file value is ignored. OK
		
Tested dao activation/deactivation from UI.						OK

Tested `appendOnlyDataStoreListeners` called only after  successful` AppendOnlyDataStoreService.put(hash,payload)` into `Optional<MapStoreService>`.  OK

Tested _we expect only one matching service which can handle the payload_.  OK
		
Ran modified test cases / all passed.  OK

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/5711#issuecomment-926887351

PR #5718, opened 2021-09-22
Add custom bisq.properties to API test harness

ghubstan:
This change adds a `bisq.properties` file to `apitest/src/main/resources`, and makes sure it is copied to regtest/dev bisq instances' app data dirs before they are started.  By default it is empty, but can be used to override default BisqApp options during ad-hoc testing.

The change was necessary because the core's `bisq.properties` resource is not copied to test harness bisq instance's regtest data dirs during a normal build.

This PR's branch is based off `master`, is not part of any series of PRs requiring reviewing and merging in chronological order.


ripcurlx (approved):
utACK

PR #5719, opened 2021-09-23
Show a warning if PC has been in sleep mode

ghost:
Per action item in https://github.com/bisq-network/projects/issues/60#issuecomment-922818568,

> advise users when their PC has been in sleep mode.

clockwatcher

Though I'm not sure if it is useful .. direct PR to bitbucket if necessary.  :wink:


ripcurlx (approved):
utACK

PR #5727, opened 2021-09-27
Remove trade id reference for Amazon gift card

### LINKED ISSUE #5726: payment.amazonGiftCard.info = Inconsistent with wiki about Message Field
### Description

In `\core\src\main\resources\i18n\displayStrings.properties` (and its translations), `payment.amazonGiftCard.info` states: “you must **include the trade ID** in the gift card''s message field. Please see the wiki”.

In https://bisq.wiki/Amazon_eGift_card#Step_2._Customize_eGift_card however, it is said that the Message field **should be left blank**.

This can confuse the user as to whether or not they should include the trade ID there. Maybe writing the ID there is usually not necessary nor desirable...?

It's not really a bug - more like a documentation issue.

#### Version
v1.7.3

ripcurlx:
Fixes #5726


ghubstan (approved):
utACK

sqrrm (approved):

PR #5728, opened 2021-09-27
Disable GC calls if fullDaoNode is set (like in case of seed nodes)

chimp1984:
Seed nodes have 4 GB allocated so the attempt to reduce memory
footprint is not needed for those cases. We use the fullDaoNode
prog arg as its expected that any full node will have allocated
sufficient memory and thus the intended reduction of memory by
calling the GC is not required and counter productive.


ripcurlx (approved):
utACK

ghost:
Are we sure there is no more edges cases like that where GC should be disabled?

PR #5729, opened 2021-09-27
[WIP] Atomic bsq trades

sqrrm:
Replaces https://github.com/bisq-network/bisq/pull/5121/, https://github.com/bisq-network/bisq/pull/5127

Major refactoring of OfferPayload to accommodate offers with no maker fee tx.

A lot of copy paste and remove extra stuff to place and take atomic offers.

Trade domain refactoring to allow for atomic trades.


core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
```diff
@@ -425,7 +426,7 @@ private long getTradeLimit(Coin maxTradeLimit,
                                PaymentMethod paymentMethod) {
         if (CurrencyUtil.isCryptoCurrency(currencyCode) ||
                 !PaymentMethod.hasChargebackRisk(paymentMethod, currencyCode) ||
-                direction == OfferPayload.Direction.SELL) {
+                direction == OfferPayloadI.Direction.SELL) {
```
ripcurlx: The naming of the interface `OfferPayloadI` somehow feels weird. How about calling the interface `OfferPayload` and the implementation `OfferPayloadImpl` and coding as much as possible against the interface? Or call the implementation something other more specific (I'm to early in the commits to suggest something better - `OfferWithFeePayload`?)

ripcurlx: Or is this about the proto resolving to be backwards compatible?

core/src/main/java/bisq/core/offer/AtomicOfferPayload.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    @Override
+    public long getMinAmount() {
+        return 0;
+    }
+
+    @Override
+    public String getBaseCurrencyCode() {
+        return "BSQ";
```
ripcurlx: as this is an BSQ atomic offer payload shall we call it AtomicBsqOfferPayload in case we want to support more Atomic offers in the future (XMR?)

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -409,6 +412,35 @@ public void placeOffer(Offer offer,
...
+                                 ErrorMessageHandler errorMessageHandler) {
+        checkArgument(offer.getOfferPayloadI() instanceof AtomicOfferPayload);
+        var atomicModel = new AtomicPlaceOfferModel(offer,
+                offerBookService,
+                tradeStatisticsManager,
+                daoFacade,
+                user,
+                filterManager);
+
+        var atomicPlaceOfferProtocol = new AtomicPlaceOfferProtocol(atomicModel,
```
ripcurlx: Maybe `PlaceAtomicOfferProtocol` instead of `AtomicPlaceOfferProtocol`?

core/src/main/java/bisq/core/offer/placeoffer/atomic/AtomicPlaceOfferModel.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.bitcoinj.core.Transaction;
+
+import lombok.Getter;
+import lombok.Setter;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+@Getter
+public class AtomicPlaceOfferModel implements Model {
```
ripcurlx: Same here `PlaceAtomicOfferModel`?

desktop/src/main/java/bisq/desktop/main/offer/AtomicOfferDataModel.java
```diff
@@ -0,0 +1,591 @@
...
+                log.error(t.toString());
+            }
+        }
+    }
+
+    private Volume calculateVolumeForAmount(ObjectProperty<Coin> minAmount) {
+        Volume volumeByAmount = price.get().getVolumeByAmount(minAmount.get());
+
+        // For HalCash we want multiple of 10 EUR
+        if (paymentAccount.isHalCashAccount())
```
ripcurlx: I guess the model needs some clean-up after copy & pasting :wink:

desktop/src/main/java/bisq/desktop/main/offer/AtomicOfferView.java
```diff
@@ -0,0 +1,753 @@
...
+        boolean currencyForMakerFeeBtc = model.getDataModel().isCurrencyForMakerFeeBtc();
+        tradeFeeInBtcToggle.setSelected(currencyForMakerFeeBtc);
+        tradeFeeInBsqToggle.setSelected(!currencyForMakerFeeBtc);
+
+        if (!DevEnv.isDaoActivated()) {
+            tradeFeeInBtcToggle.setVisible(false);
+            tradeFeeInBtcToggle.setManaged(false);
+            tradeFeeInBsqToggle.setVisible(false);
+            tradeFeeInBsqToggle.setManaged(false);
+        }
```
ripcurlx: I guess all those trade fee UI components won't be necessary anymore for atomic BSQ trades.

core/src/main/java/bisq/core/trade/Trade.java
```diff
@@ -30,6 +30,7 @@
 import bisq.core.support.dispute.refund.RefundResultState;
 import bisq.core.support.messages.ChatMessage;
 import bisq.core.trade.protocol.ProcessModel;
+import bisq.core.trade.protocol.ProcessModelI;
```
ripcurlx: Same here I find adding the I to the end to mark the interface kind of confusing. It could be mistaken as the short form of `Impl`.

chimp1984:
I just looked mostly at the UX and looks good to me overall. Congrats!

Here some remarks:
I think the field in red should be hidden.
Screen Shot 2021-09-28 at 21 37 09

Maybe better to call it "Atomic swaps" instead of "Atomic trades" 
Also in the take offer screen maybe better to use: SWAP instead of ATOMIC!!
And "BSQ swap" instead of "Atomic BSQ" in offerbook

In general maybe better not use `atomic` at all in the UI. For non techies it might sound confusion (nuclear - might create more environmental concerns ;-)). Maybe just call it BSQ swaps (swap might be too generic if we once add cross chain swaps)?

I got that expection after having one swap done and started another without creating a new block in between. I guess the BSQ change was not confirmed....
```
java.lang.IllegalArgumentException: Failed to prepare maker inputs
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:142)
	at bisq.core.trade.atomic.protocol.tasks.maker.AtomicMakerVerifiesTakerInputs.run(AtomicMakerVerifiesTakerInputs.java:58)
	at bisq.common.taskrunner.TaskRunner.next(TaskRunner.java:69)
	at bisq.common.taskrunner.TaskRunner.handleComplete(TaskRunner.java:85)
	at bisq.common.taskrunner.Task.complete(Task.java:58)
	at bisq.core.trade.protocol.tasks.AtomicTradeTask.complete(AtomicTradeTask.java:45)
	at bisq.core.trade.atomic.protocol.tasks.maker.AtomicMakerVerifiesAmounts.run(AtomicMakerVerifiesAmounts.java:85)
```
After creating a block it worked. Could take same offer.

In the code maybe better use BsqSwap instead of Atomic. AtomicProcessModel sounds confusing as the ProcessModel is not atomic but a ProcessModel for an atomic bsq swap.

Also would avoid the postfix I like in ProcessModelI.
Better rename the ProcessModel with some short prefix (DefaultProcessModel if no better found)

I think addP2PMessageFilter will not work as filterManager.getFilter() is likley null at startup. I would also use a concrete type for message not Object. And would invert the result as now false is expected that the predicate is evaluated true.

ripcurlx:
Closing as superseded and merged by #5775

PR #5740, opened 2021-10-02
Move volume formatting from DisplayUtils to VolumeUtil

ghubstan:
In practice, presentation/display logic should be kept in UIs, but the core api will need to duplicate a very small part of the code in `DisplayUtils` if a very small part of this refactoring is not done.  Instead of moving a tiny bit of the volume formatting methods, all of them are moved to core's `VolumeUtil`.

This PR's branch is based off `master`, and is not part of any PR chain requiring review/merge of any parent PRs.


ripcurlx (approved):
utACK

PR #5741, opened 2021-10-03
Add payment methods Bizum and Pix

### LINKED ISSUE #208: New payment method from Brazil - PIX
There is a new method of transferring securities available in Brazil called PIX. The recipient only needs to share a phone number, e-mail or Social Security Number. The advantage over the National Bank Transfer payment method is that the PIX works 24 hours a day, 7 days a week and the transfer takes place in a few minutes. To prevent users from creating orders in different ways with the intention of receiving through PIX (which would make it difficult to accept orders), I believe it would be important to create a PIX payment method where you only need to inform one field, which would be the link , a phone or email

### Why

_Why would Bisq benefit from adding this payment method? Is is popular in a particular region? More convenient? More safe?Is a payment method that works in the same way as National Bank Transfer for security, but uses a new protocol established by the Brazilian Central Bank and allows transfers to take place in a few minutes and works 24/7, which allows for faster negotiations

### Region

_Who can use this payment method?_
works throughout Brazil

### Chargeback risk

_How easy is it to cancel a payment once it's made? This is the most important characteristic of a payment method._
There is a possibility of chargeback if the payer disputes the payment at the Bank, as it is already done with National Bank Transfers. PIX follows the same rules, only uses a new protocol that works 24/7

### Data requirements

_What information **must** a user provide to a counterparty to accept a payment?_
the recipient needs to fill in a single field with the information he prefers (according to the registration made at PIX):
Social Security Number (called CPF in Brazil) OR
Phone number OR
Email
Only one of these 3 information is sufficient for payment to be made

### Verification

_How can payments be verified? Examples: TLSNotary is typically used for electronic payments, and receipt scans are typically used for money orders._
it is possible to generate a proof of transaction via internet banking

### Duration

_How long do payments take? Please provide a range, as the advertised best-case scenario rarely accounts for edge cases._
as PIX works 24/7 there is no need to worry about weekends for example. So I believe that 1 day is enough for the negotiation to be concluded

### Fees

_Does it cost anything to make a transfer using this payment method? If so, how much?_
There are no costs

### Fraud risk

_How easy or difficult is fraud with this payment method?_
it is possible for the recipient to inform a link in the field with the necessary data for payment (email, telephone or SSN described above). Link is also a valid possibility within the PIX protocol, but for the user's security within Bisq I believe that this option should not be used, because the link could be to a malicious destination. Outside this point, the risks are the same as for a National Bank Transfer

### LINKED ISSUE #222: Add Bizum as a new payment method (popular in Spain)
## Bizum

Resources: 

[Bizum website](https://bizum.es/en/) 
[Bizum Wiki](https://es.wikipedia.org/wiki/Bizum)
[Bizum FAQs](https://bizum.es/en/frequently-asked-questions/)

Bizum is a mobile payment provider service based in Spain. Bizium is a result of the collaboration of almost all of the Spanish banks to create a system of instant payments between individuals and shopping in stores. 

### Region

Spain. You need is a bank account (IBAN) in Spain with any of the entities which offer the service and a mobile phone number.

The Bizum service is not yet available for bank accounts outside Spain. Insofar as the telephone is concerned, Bizum has no restrictions, reason why it would not be a problem to make a Bizum to foreign mobile phone numbers which have a Spanish bank account associated with those phone numbers

### Currencies

EUR

### Chargeback risk

Low

### Size of user base

13 million users

### Data requirements

The service allows payments between users using the mobile application and knowing only the recipient's phone number.

### Verification

TLSNotary can not be used as payment method is app based.  Screen recording / screenshots might be able to be used to provide payment proof.

### Duration

Instant payments, less than 10 seconds. 

When the user wants to send money, they will receive a verification text message that, along with a code, they will have to enter in the application. Once this is done, the transaction is done and the other user receives the funds instantly. The recipient also receives a text message confirming the receipt of the transferred funds.

### Fees

Free to send and receive

### Fraud risk

Low

### Payment amounts

Can be used for amounts between 0.50 and 1000 per payment

The maximum amount of transactions you can receive using Bizum is 2000 Euros per day.

- Minimum amount per operation: €0.50
- Maximum amount per operation: €1,000
- Amount of operations received per client per day: €2,000
- Number of operations issued and received by a client in a month: 150 operations
- Maximum number of recipients to be included in a multiple request/remittance: 30 recipients
- Duration of pending remittance to a Non-Client: 2 days
- Duration of the pending request to a Non-Client: 7 days

Each bank however may establish its own limits, within these limits, for its clients.

ghost:
Fixes https://github.com/bisq-network/growth/issues/222
Fixes https://github.com/bisq-network/growth/issues/208

---

<details>
  <summary>Screenshots for Bizum</summary>
  
bizum_create_popup
bizum_create
bizum_buy_popup
bizum_buy_offer
bizum_sell_popup
bizum_sell
bizum_trade_buyer
bizum_trade_seller

</details>

---

<details>
  <summary>Screenshots for Pix</summary>
  
pix_create_popup
pix_create
pix_sell_popup
pix_sell_offer
pix_buy_popup
pix_buy_takeoffer
pix_trade_pay
pix_trade_confirm

</details>

---


ripcurlx (approved):
utACK - NIT: Please start your commit messages with an uppercase (https://chris.beams.io/posts/git-commit/)

PR #5747, opened 2021-10-11
Allow standard mark & copy functionality for displayed text fields

ghost:
Instead of a "Copy" icon on every field, allow the text to be selected & copied.  Replace `addConfirmationLabelLabel` with `addConfirmationLabelTextField`.

- Gives the ability to mark & copy all text fields in Trade and Dispute tickets.
- Gives the ability to mark & copy text from `TextFieldWithIcon`

Resolves https://github.com/bisq-network/bisq/discussions/5706

---

Screenshots:

image

image

image

image

### DIFF
--- a/desktop/src/main/java/bisq/desktop/components/TextFieldWithIcon.java
+++ b/desktop/src/main/java/bisq/desktop/components/TextFieldWithIcon.java
@@ -29,13 +29,9 @@ import javafx.scene.text.TextAlignment;
 
 import javafx.geometry.Pos;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import lombok.Getter;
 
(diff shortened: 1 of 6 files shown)


ripcurlx:
I do agree that it is better to stick to common expected UI features compared to adding more copy icons to do such task.
For completeness this also affects the state string in step 3 on buyer side.
Bildschirmfoto 2021-10-12 um 10 25 38

I've pushed one little logger fix as well, that was introduced by mistake and popped up while searching for references.

ripcurlx (approved):
ACK

BtcContributor:
I noticed that in the trade detail window if you try to right click on "green" strings like all the txid to copy them, the context menu will open but then instantly it is opened also the mempool.space link.

I know that a user should use the copy button but maybe we could block right click on strings that already have a copy button?

PR #5748, opened 2021-10-11
Fix logic of SEPA country comparison routine

ghost:
Minor bugfix: there is logic to display `All Euro countries` in place of listing out all 23 country codes (see screeshots below).  This was broken because the comparison routine was comparing the lists out of order.

---
Account details:
image

Dispute details:
image


ripcurlx (approved):
utACK

PR #5753, opened 2021-10-13
Fix exception in offer entry UI

ghost:
Fixes an exception in offer entry when fixed prices are used.

## Steps to reproduce issue:

First, select the option of fixed price:

image

- Click on create offer to Buy.
- Select the Fixed Price option (per screenshot above).
- Close the enter offer dialog.  This saves the setting.
- Again, click on create offer to Buy.
- Enter 0.01 (for BTC amount), [TAB], 0, [TAB], 0, [TAB], [TAB].
- Exception error popup message is shown.

image

image


ripcurlx (approved):
utACK

PR #5736, opened 2021-09-30
Revert "Community tab"

ripcurlx:
Reverts bisq-network/bisq#5680

Because of concerns raised in https://github.com/bisq-network/bisq/issues/5732 I temporarily revert this PR until remaining issues are agreed upon and fixed.

PR #5760, opened 2021-10-16
Fix typo (SHA)

### LINKED ISSUE #5759: Typo in SWIFT payment method
https://github.com/bisq-network/bisq/blob/31cc79428d0f5cde30ce31755392b1c392ddc593/core/src/main/resources/i18n/displayStrings.properties#L3463

To be consistent with other occurrences of this expression, `shared payment model (SHA)` should be `shared fee model (SHA)`.

Jakub-CZ:
Fixes #5759

The string `shared payment model (SHA)` should be `shared fee model (SHA)`.

Also fixing a few more typos found.


core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -3649,16 +3649,16 @@ Users who have registered with only their email, name, and phone number have the
...
   ● $100 maximum per payment\n\n\
-Users can increase their limits by providing Strike more information have the following limits:\n\n\
+Users can increase their limits by providing Strike with more information. These users have the following limits:\n\n\
   ● $1,000 maximum per deposit\n\
   ● $1,000 maximum total deposits per week\n\
   ● $1,000 maximum per payment\n\n\
 If you trade over the above limits your trade might be cancelled and there could be a penalty.
 payment.strike.info.buyer=Please send payment only to the BTC Seller's Strike username as provided in Bisq.\n\n\
-The maximum trade size is $1,000 maximum per payment.\n\n\
+The maximum trade size is $1,000 per payment.\n\n\
```
Jakub-CZ, line 3658: Is this really "**per payment**"? Or is this supposed to be "**per transaction**" like with some of the Indian payment methods?

E.g.:
https://github.com/bisq-network/bisq/blob/adf0562152def9eb1948f9f46bb512d49f3b9282/core/src/main/resources/i18n/displayStrings.properties#L3476-L3477

m52go:
Aside from the SHA error, this is another good reminder to make sure strings get a second look before merging. Something similar happened with [the community tab](https://github.com/bisq-network/bisq/pull/5736) recently.

core/src/main/resources/i18n/displayStrings.properties
pazza83: Thanks for spotting the SHA error.

For Strike

> Is this really "per payment"? Or is this supposed to be "per transaction" like with some of the Indian payment methods?
Only $1000 USD can be deposited to a users account per week. 
The maximum payment size is $1000 USD.

Therefore it is not possible to buy or sell on Strike over $1000 USD per week. Users should limit their activity to trades of this size.

For Strike the limits are:

Users who have registered with only their email, name, and phone number have the following limits:

$100 maximum per deposit
$1,000 maximum total deposits per week
$100 maximum per payment
Users who increase their limits by providing more information have the following limits:

$1,000 maximum per deposit
$1,000 maximum total deposits per week
$1,000 maximum per payment

For IMPS the limits are:

Maximum of 200,000 INR can be sent per transaction. This is approximately $2,700 USD.
Maximum of 1,000,000 INR can be sent per day. This is approximately $13,500 USD.
Minimum payment is 1 INR ~ $0.01 USD

Therefore, trade sizes can be more fluid than Strike as multiple payments can be made to complete a trade.

Plan for both would be to include a link to the wiki once completed.

ripcurlx (approved):
utACK

PR #5763, opened 2021-10-18
chore: Bump tor to v0.4.5.10

cd2357:
Update netlayer and tor-binary dependencies.


ripcurlx (approved):
utACK - verification of hashes took place in other repositories already

PR #5771, opened 2021-10-20
Option to not repeat popup message of locked up funds from failed trade

ghost:
When Bisq detects funds locked up from a failed trade it pops up a message alert at frequent intervals (for example every block).  This can get frustrating especially if more than one trade is affected.

This PR allows users to selectively choose to stop receiving these popups **on a trade by trade basis**, via the "Don't show again" checkbox.

failed_trade_popup

Requested by @leo816 and users in keybase support channel.


ripcurlx (approved):
utACK

PR #5770, opened 2021-10-20
BSQ swaps: Low risk refactorings [2]

chimp1984:
Based on https://github.com/bisq-network/bisq/pull/5769

Starts at: e75296557739f6b8ae86bd3004781fd268036a9e

Contains low risk but potential functional changes as preparation for BSQ swap PR.
Contains preparations for BsqSwap which might not make sense at the moment of the commit (e.g. unused mehtods, params,...) but are done to reduce change set while getting closer to the full version.

EDIT:
Merged https://github.com/bisq-network/bisq/pull/5769 (which was merged and rebased with master). But rebase failed each time I tried...

PR #5775, opened 2021-10-21
Bsq swap: Implementation [3]

chimp1984:
Based on https://github.com/bisq-network/bisq/pull/5770

Starts at 9c56202

This adds the whole BSQ swap implementation. I did not see much value in breaking it up into many commits. I think for review its best to check it out and do a full compare to the previous PR to see the changes.

All that work is derived from the work of @sqrrm  at https://github.com/bisq-network/bisq/pull/5729 

EDIT:
Merged and rebased  https://github.com/bisq-network/bisq/pull/5770


chimp1984:
No idea why the gradle dependencyVerification fails. It is the same content as in the release branch.

chimp1984:
Seems its a jitpack issue. Devs on keybase complained also about problems.

desktop/src/main/java/bisq/desktop/main/dao/DaoView.java
```diff
@@ -145,8 +145,12 @@ public void initialize() {
     protected void activate() {
         if (DevEnv.isDaoActivated()) {
 
-            // Hide dao new badge if user saw this section
-            preferences.dontShowAgain(DaoPresentation.DAO_NEWS, true);
+            if (preferences.showAgain(DaoPresentation.DAO_NEWS)) {
+                preferences.dontShowAgain(DaoPresentation.DAO_NEWS, true);
+                new Popup().headLine(Res.get("news.bsqSwap.title"))
+                        .information(Res.get("news.bsqSwap.description"))
```
ghubstan, line 151: TODO Write and publish [BSQ Swaps wiki.](https://bisq.wiki/BSQ_Swaps)

core/src/main/java/bisq/core/offer/placeoffer/bsqswap/PlaceBsqSwapOfferModel.java
```diff
@@ -0,0 +1,46 @@
...
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.core.offer.placeoffer.bsqswap;
```
ghubstan: Some packages are named `bsq_swap`, but this one is `bsqswap`.  Could the `bsq_swap` pkgs be renamed `bsqswap` too?

desktop/src/main/java/bisq/desktop/main/overlays/windows/BsqSwapOfferDetailsWindow.java
```diff
@@ -0,0 +1,320 @@
...
+        String btcDirectionInfo = "";
+        OfferDirection direction = offer.getDirection();
+        String currencyCode = offer.getCurrencyCode();
+        String offerTypeLabel = Res.get("shared.offerType");
+        String toReceive = " " + Res.get("shared.toReceive");
+        String toSpend = " " + Res.get("shared.toSpend");
+        String minus = " - ";
+        String plus = " + ";
+        String minerFeePostFix = Res.get("tradeDetailsWindow.txFee");
+        String tradeFeePostFix = Res.get("shared.tradeFee");
```
ghubstan, line 160: Is there no way to tell the bsq swap offer creator what the trade fee will be
how-much-is-trade-fee
?

ghubstan: The taker can see the trade fee before taking the offer, but the maker still doesn't know yet.

ghubstan: After taking the offer, the maker still can't find the trade fee, until he checks the swap tx in a block explorer, or exporting swap trade to CSV file.

desktop/src/main/java/bisq/desktop/main/dao/DaoView.java
chimp1984: I pinged @m52go . I think he had already the content but I dont see a PR yet.

core/src/main/java/bisq/core/offer/placeoffer/bsqswap/PlaceBsqSwapOfferModel.java
chimp1984: Ops. Renamed in a new commit.

desktop/src/main/java/bisq/desktop/main/overlays/windows/BsqSwapOfferDetailsWindow.java
chimp1984: When creating the offer the second step shows all the fees. I was a bit lazy to add it here as well and as the tx fee are calculated based on the inputs/change its not that trivial. But yes could be added.

ripcurlx:
I just tried both scenarios to Buy and Sell BTC for BSQ  successfully! 🚀  These are my comments for the UI parts of the trade process.

1) I think it would be good to mention in the popup or in wiki article linked in the `NEW` popup (https://bisq.wiki/BSQ_swaps), that a hidden BSQ Swap account is created: `BSQ Swap (BSQ, BSQ Swap)`. It wasn't a big hurdle when I tried to create a Buy BTC offer that I need to select the new BSQ Swap account, but I think it would be good to mention that this account is created automatically and that it is not visible in the account section as all other accounts.

Bildschirmfoto 2021-11-03 um 10 16 23

2) When creating a Buy BTC (Sell BSQ) offer I recognized that after creating the offer the filter by currency combobox was selected and open (I have the same open issue in the buy BSQ branch). Is this a known problem?
Bildschirmfoto 2021-11-03 um 10 21 26

3) I'm not sure if it is necessary to change the label of the tab to `Create BSQ Swap offer` from `Create offer` when a BSQ swap account is selected, as it is also placed under `My open offers`. 
Bildschirmfoto 2021-11-03 um 10 20 23

4) When taking a BSQ swap offer, after taking the offer successfully I get following success message.

Bildschirmfoto 2021-11-03 um 10 24 50

This leaves me with the impression that the trade is already completed.  I was expecting in that case that I'll find  the trade in my portfolio > history tab. But as mentioned in the overlay above there is a new BSQ SWAP TRADES tab (Should be in lower case to match other navigation references: `BSQ swap trades`).  So I though that this is because the transaction doesn't have already a block confirmation.

Bildschirmfoto 2021-11-03 um 10 25 43
Bildschirmfoto 2021-11-03 um 10 26 37
After sufficient confirmations the tab is still there. Is there a reason why we didn't call this tab `Open BSQ swap trades` and move it to history after the trade has at least one block confirmation? I personally would appreciate having all my trades in one history, or is there a technical reason for this?

cd2357:
> No idea why the gradle dependencyVerification fails. It is the same content as in the release branch.

The latest build (re-triggered by @ripcurlx 20 mins ago) was successful. Should also work locally.

ripcurlx:
Bildschirmfoto 2021-11-03 um 11 28 54
Bildschirmfoto 2021-11-03 um 11 29 04

I also just recognized that all altcoin/bsq swap offers have the deviation value greyed out, but not Fiat offers. This is a live problem already. I just recognized it for the first time.

ripcurlx:
Just read the code, that it is done like this to indicate if it is a fixed price or a market price offer. This is not intuitive IMO. I'd rather add the icon as in the offer book and keep the opacity to `1`. @chimp1984 Did you do this to save horizontal space?

ripcurlx:
> > No idea why the gradle dependencyVerification fails. It is the same content as in the release branch.
> 
> The latest build (re-triggered by @ripcurlx 20 mins ago) was successful. Should also work locally.

This is very weird right now. Sometimes it builds, sometimes it fails because of the verification error.

chimp1984:
@ripcurlx 
Re 1: The BSQ account creation is mentioned in the news popup:
Screenshot 2021-11-03 at 12 03 01

Re 2:
I cannot reproduce that anymore, but had that issue in the past. It seems some bug with the combobox and with focus. I added a requestFocus call but might be that there are timing/race conditions where the combobox still wins... Maybe some delayed requestFocus can make it more reliable.

Re 3:
Yes, no strong opinion here. But adding a bit explicit naming might help that people understand its different to normal offers/trades. For open offers there is no beavioural difference beside that edit offer is disabled and the disable/enable column which reacts to wallet balance. So that was rather easy to add to the existing one. For trades I think its more justified to add another tab as the process of pending trades->closed trades is not happening for BSQ swaps. 
To add too many tabs might overload a bit the UI and I think its more convenient to have all offers in one place. 

Re 4:
Yes I can rename "BSQ SWAP TRADES".
The trade is basically completed. You cannot do anything to stop it from confirming beside doing a doublespend.
I think we should not give the impression the trade is still in a pending state just because its not confirmed. For BSQ though the funds don't show up before confirmations. So yes that could be improved...
To add the trade after confrimation to history should be possible but comes with some effort. I would prefer to do that in a new PR if considered important.

Re: "I also just recognized that all altcoin/bsq swap offers have the deviation value greyed out, but not Fiat offers. This is a live problem already."
Yes might need a review...

I will add a commit for the rename and add another delayed requestFocus (hope that helps). For the BSQ swap account issue I think it should be covered in the news popup suficiently and maybe we can add/emphasise it in the wiki (have not looked up how its there). If you have a concrete suggestion where to add it additionally let me know.
But would prefer to get that merged and add other changes in new PRs.

chimp1984:
Btw. When creating a BSQ altcoin account one get also a popup explaining that there is a default BSQ swaps account...

ripcurlx:
NIT: If there is no BSQ swap I think we should not show the `Unconfirmed...` tab.

Bildschirmfoto 2021-11-04 um 09 56 44

ripcurlx (approved):
ACK - Besides the NIT I mentioned above I think it is good to move to some extended developer testing on master. Everyone running on master please be aware that the offer propagation might not be very stable until our seednodes have been updated (hint to @Emzy). Besides that BSQ swaps were thoroughly manually and automated tested on Regtest and on Mainnet.

PR #5782, opened 2021-10-26
Redesign dao state monitoring [4]

### LINKED ISSUE #5779: Redesign DAO state monitoring to avoid performance degradation

chimp1984:
Based on https://github.com/bisq-network/bisq/pull/5775
Starts at 3ed6914

Fixes https://github.com/bisq-network/bisq/issues/5779

Spec see at https://github.com/bisq-network/bisq/issues/5779#issuecomment-951912926

EDIT: Merged and rebased with master


core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -2531,6 +2531,8 @@ dao.monitor.isInConflictWithSeedNode=Your local data is not in consensus with at
   Please resync the DAO state.
 dao.monitor.isInConflictWithNonSeedNode=One of your peers is not in consensus with the network but your node \
   is in sync with the seed nodes.
+dao.monitor.isDaoStateBlockChainNotConnecting=You dao state chain is not connecting with the new data. \
+  Please resync the DAO state.
```
ripcurlx: Should be `Your dao state ...` and is there a reason why you used first `dao` in lower case and afterwards `DAO` in uppercase?

desktop/src/main/java/bisq/desktop/main/dao/monitor/StateMonitorView.java
```diff
@@ -400,6 +406,7 @@ public void updateItem(final BLI item, boolean empty) {
         column.setComparator(Comparator.comparing(e -> e.getStateBlock().getInConflictMap().size()));
         tableView.getColumns().add(column);
 
+
```
ripcurlx: Unnecessary new blank line.

core/src/main/java/bisq/core/dao/monitoring/DaoStateMonitoringService.java
```diff
@@ -338,59 +361,93 @@ private void updateHashChain(Block block) {
...
-            if (this.isInConflictWithSeedNode)
-                log.warn("Conflict with seed nodes: {}", conflictMsg);
-            else if (this.isInConflictWithNonSeedNode)
-                log.debug("Conflict with non-seed nodes: {}", conflictMsg);
+    private void processPeersDaoStateHashes(List<DaoStateHash> stateHashes, Optional<NodeAddress> peersNodeAddress) {
+        boolean useDaoMonitor = preferences.isUseFullModeDaoMonitor();
+        stateHashes.forEach(peersHash -> {
+            Optional<DaoStateBlock> optionalDaoStateBlock;
+            // If we do not add own hashes during initial parsing we fill the missing hashes from the peer and create
+            // the at the last block our own hash.
```
ripcurlx: `the at the`  should be `at the`

core/src/main/java/bisq/core/dao/state/DaoStateSnapshotService.java
```diff
@@ -89,25 +90,76 @@ public DaoStateSnapshotService(DaoStateService daoStateService,
         this.storageDir = storageDir;
     }
 
+
```
ripcurlx, line 93: Unnecessary blank line

core/src/main/java/bisq/core/dao/state/DaoStateSnapshotService.java
```diff
@@ -89,25 +90,76 @@ public DaoStateSnapshotService(DaoStateService daoStateService,
...
+        // and we got called back from daoStateMonitoringService once the hashes are created from peers data.
+        if (!preferences.isUseFullModeDaoMonitor()) {
+            // We register a callback handler once the daoStateMonitoringService has received the missing hashes from
+            // the seed node and applied the latest hash. After that we are ready to make a snapshot and persist it.
+            daoStateMonitoringService.setCreateSnapshotHandler(() -> {
+                // As we did not have created any snapshots during initial parsing we create it now. We cannot use the past
+                // snapshot height as we have not cloned a candidate (that would cause quite some delay during parsing).
+                // The next snapshots will be created again according to the snapshot height grid (each 20 blocks).
+                // This also comes with the improvement that the user does not need to load the past blocks back to the last
+                // snapshot height. Thought it comes also with the small risk that in case of re-orgs the user need to do
```
ripcurlx: `Though` instead of `Thought`

core/src/main/resources/i18n/displayStrings.properties
chimp1984: Fixed

desktop/src/main/java/bisq/desktop/main/dao/monitor/StateMonitorView.java
chimp1984: Fixed

core/src/main/java/bisq/core/dao/monitoring/DaoStateMonitoringService.java
chimp1984: Fixed

core/src/main/java/bisq/core/dao/state/DaoStateSnapshotService.java
chimp1984: Fixed

core/src/main/java/bisq/core/dao/state/DaoStateSnapshotService.java
chimp1984: Fixed

ripcurlx (approved):
ACK - reviewed the code changes, ran it on Mainnet with a light and full node and compared hashes.

Bildschirmfoto 2021-11-09 um 12 10 56
Bildschirmfoto 2021-11-09 um 12 10 52

PR #5790, opened 2021-10-31
Extract persistence of BSQ blocks out of DaoStateStore [5]

### LINKED ISSUE #5783: Improve DAO state peristence
The DAO state is getting larger with each new block and is a major issue for memory consumption of Bisq.
Currently it is about 150 MB, where 135 MB are the blocks. We load it at startup which takes about 2-4 sec on my machine and then after parsing is complete we write snapshots to disk (every 20 blocks) which also takes a few seconds. Persistence happens on background threads so it is not a major problem but still quite slow.

I was wondering what we could do to improve that situation.
Best would be if we could avoid to load 150 MB into memory but I fear that is not feasible without either a huge and risky effort to optimize and customize the data access or try to use some DB but I am pretty sure that would degrade performance a lot as we need most of the data for different forms of access (the large chunk of data are transactions and we need to look up those for various properties in the DAO context). In Bitcoin it is easier as there the spendability is the only real aspect to consider, so pruning of data which is not relevant for furture use cases can be done easier. In Bisq the DAO has many different use cases where old data is still relevant, so pruning is likely not really possible. The only way I see is to create all kind of data views (lookup maps) for the various use cases we have. E.g. If we need to iterate all txs for finding potential locked outputs we could just make a lookup map where the txOutputType we are interested in is used as key and we hold only the few relevant txs there. If all use cases could be optimized that way we could release the blocks/txs and only keep the relevant views, which likely will be much smaller. That would be one approach but I am not 100% sure if it would really lead to less data and it comes with considerable effort and risks for the DAO consensus.

But the 150 MB in memory are not the only concern. At snapshot creation we have to keep a copy of that in memory. Here we could optimize with much less risks and less effort.
We could extract the blocks from the daoState into a dedicated data store and split that up either automatically in buckets of e.g. 1000 blocks or use a similar versioning model as we use for trade statistics (historical data store). The automatically managed approach would be probably better as it would not require work from the release manager. As blocks are immutable we could treat all old buckets as immutable closed data structures and therefor wont need to persist them. They are only loaded for getting the data and concatenated with all the other buckets to form the full BSQ block chain. We only persist the current one which will be much faster as data size will be a few MB not 150 MB.
The other DAO state is about 15 MB so that is also ok for fast persistence.
There is still a smaller issue with re-org. Say we just moved to a new bucket with a new block and then there is a 3 block deep re-org so it changes our blocks in the previous bucket which we considered as immutable/closed. But that is a minor issue and there are ways to deal with that, like considering the previous bucket not as closed but preliminarily closed so we still can mutate it. After deleting the reorged blocks we would resync from there. Thats similar as we do a resync from the last snapshot in case of reorgs in the current model. We do not need to support arbitrarily deep re-orgs as realistically reorgs are never deeper than 3-4 blocks. 

Another smaller improvement is that I guess it would decrease memory requirements if we serialize the cloned snapshot candidate immediately so we avoid the overhead for Java data sturctures and required memory should be smaller. Have to try out how much difference that makes.

So I think that approach is feasible to implement and comes with little risks for DAO consensus issues.

Any other ideas? Feedback?

chimp1984:
Based on https://github.com/bisq-network/bisq/pull/5782
(will merge/rebase once other PRs are in master)
Starts at ff702ae

Fixes https://github.com/bisq-network/bisq/issues/5783

EDIT: Merged and rebased with https://github.com/bisq-network/bisq/pull/5782


chimp1984 (commented):
<strike>utACK</strike> Upps was on the wrong tab...

common/src/main/java/bisq/common/persistence/db/FixedHash.java
```diff
@@ -0,0 +1,418 @@
...
+import java.nio.channels.FileChannel;
+
+import java.io.File;
+import java.io.RandomAccessFile;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * @author ebruno
```
ripcurlx: Where is this code copied from?

common/src/main/java/bisq/common/persistence/db/NoHeapDB.java
```diff
@@ -0,0 +1,380 @@
...
+
+            log.error("write took {} ms", System.currentTimeMillis() - ts);
+            ts = System.currentTimeMillis();
+            for (int i = 0; i < size; i++) {
+                String key = c + sep + i;
+                String val = store.getString(key);
+                //log.error("read key/val {}/{}", key, val);
+            }
+            log.error("read took {} ms", System.currentTimeMillis() - ts);
+            c++;*/
```
ripcurlx: Why is this code part commented out?

common/src/main/java/bisq/common/util/GcUtil.java
```diff
@@ -59,20 +60,25 @@ private static void maybeReleaseMemory(long trigger) {
...
+                    totalGCTime / 1000d);
+           /* if (DevEnv.isDevMode()) {
                 try {
                     // To see from where we got called
                     throw new RuntimeException("Dummy Exception for print stacktrace at maybeReleaseMemory");
                 } catch (Throwable t) {
                     t.printStackTrace();
                 }
-            }
+            }*/
```
ripcurlx, line 81: If commented out, better to remove it completely.

core/src/main/java/bisq/core/dao/state/storage/BlocksPersistence.java
```diff
@@ -0,0 +1,205 @@
...
+    private Path usedTempFilePath;
+
+    public BlocksPersistence(File storageDir, String fileName, PersistenceProtoResolver persistenceProtoResolver) {
+        this.storageDir = storageDir;
+        this.fileName = fileName;
+        this.persistenceProtoResolver = persistenceProtoResolver;
+
+       /* if (!storageDir.exists()) {
+            storageDir.mkdir();
+        }*/
```
ripcurlx: Why is this commented out?

ripcurlx: I guess it can be safely removed as it is checked during the write procedure anyways.

core/src/main/java/bisq/core/dao/state/storage/DaoStateStorageService.java
```diff
@@ -74,60 +79,118 @@ public String getFileName() {
...
-        store.setDaoStateAsProto(null);
-        store.setDaoStateHashChain(null);
-        GcUtil.maybeReleaseMemory();
+    @Override
+    protected void readFromResources(String postFix, Runnable completeHandler) {
+        new Thread(() -> {
+            Thread.currentThread().setName("copyBsqBlocksFromResources");
+            bsqBlocksStorageService.copyFromResources(postFix);
+
+            // We read daoState and blocks ane keep them in fields in store and
```
ripcurlx: ...and keep them

common/src/main/java/bisq/common/persistence/db/FixedHash.java
chimp1984: Ah that whole `common/src/main/java/bisq/common/persistence/db/` dir should have been deleted. There is a later commit removing it... was from the MappedByteBuffer experiments.

common/src/main/java/bisq/common/persistence/db/NoHeapDB.java
chimp1984: Ah that whole common/src/main/java/bisq/common/persistence/db/ dir should have been deleted. There is a later commit removing it... was from the MappedByteBuffer experiments.

core/src/main/java/bisq/core/dao/state/storage/BlocksPersistence.java
chimp1984: Got enabeld in a later commit again as it was required at some cases (resync from resources i guess, or regtest fresh start)

core/src/main/java/bisq/core/dao/state/storage/DaoStateStorageService.java
chimp1984: Outdated... not there in final version

common/src/main/java/bisq/common/util/GcUtil.java
chimp1984: I left it as its still helpful for memory/performance profiling to see from where we call it too often...
But Ok to me to remove it as well...

chimp1984:
I will wait until #5782 before merging/rebasing it...

core/src/main/java/bisq/core/dao/state/DaoStateSnapshotService.java
```diff
@@ -265,6 +273,8 @@ public void applySnapshot(boolean fromReorg) {
                         "We remove all dao store files and shutdown. " +
                         "After a restart resource files will be applied if available.");
                 resyncDaoStateFromResources();
+            } else {
+                log.info("No Bsq blocks in DaoState. Expected if not data are provided yet from resources or persisted data.");
```
ripcurlx: Expected if **no data is** provided...

ripcurlx:
I think we should move the block snapshots to Git LFS. If I'm not mistaken adding `p2p/src/main/resources/BsqBlocks_BTC_MAINNET/* filter=lfs diff=lfs merge=lfs -text` to the `.gitattributes` file should do the trick.

ripcurlx (dismissed):
utACK - besides the Git LFS config changes. As soon as it is in and rebased to master after merging #5782 I'll run it again against Mainnet before merging.

ripcurlx:
There are also two minor Codacy complaints regarding an import statement: https://github.com/bisq-network/bisq/pull/5790/checks?check_run_id=4111754368

chimp1984:
Rebases on maser.

core/src/main/java/bisq/core/dao/state/DaoStateSnapshotService.java
chimp1984: Fixed

chimp1984:
@ripcurlx 
Re Codacy complaints: Unnecessary use of fully qualified name 'protobuf.BaseBlock' due to existing import 'protobuf.BaseBlock' 
I intentionally kept the protobuf package to make it clear that we dont use the java model here but the proto data.
So I prefer to keep it. I find those compaints anyway questionable. Leaving package qualifiers can help in certain context to improve readibility IMO.

ripcurlx (dismissed):
utACK

ripcurlx (approved):
utACK

PR #5794, opened 2021-11-02
Optimize trade chart view [6]

chimp1984:
Based on https://github.com/bisq-network/bisq/pull/5790

Starts at ebf5eac

This optimizes performance and memory usage of the Trade Charts view (market).
It uses about 620MB (in JProfiler) when visible and 600 MB when at other view.
Before it was about 680-700 MB and memory did not really go lower when moving away (as we kept the data structures for caching).

It uses CompletableFutures and moved all heavy calculation code to static methods. So at view activation there is nothing heavy left on the UI thread and it can render faster, and as soon the tasks are completed the UI gets updated to render the table and charts. 
I think that pattern can be used elsewhere (like BSQ Dashboard which also uses the trade stats and is therefore heavy). But it requires caution to not call JavaFx controls from a non UI thread.

EDIT: Merged and rebased with  https://github.com/bisq-network/bisq/pull/5790

### DIFF
--- a/common/src/main/java/bisq/common/util/CompletableFutureUtils.java
+++ b/common/src/main/java/bisq/common/util/CompletableFutureUtils.java
@@ -1,0 +1,37 @@
+package bisq.common.util;/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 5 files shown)


ripcurlx:
I'm starting to review from https://github.com/bisq-network/bisq/pull/5794/commits/6cc4265517ff6d0c28fa9b936bb99958f9b355a8 onwards.

ripcurlx (dismissed):
ACK - Reviewed code and tested it on Mainnet. Will wait for final test and merge after rebase to master.

chimp1984:
Merged/Rebased with master

ripcurlx (dismissed):
ACK

ripcurlx (approved):
ACK

PR #5805, opened 2021-11-03
GitHub build: Cache gradle dependencies

cd2357:
Enable gradle cache to speed up builds and become less affected by various public repository issues.

See https://github.com/marketplace/actions/setup-java-jdk#caching-packages-dependencies


ripcurlx (approved):
utACK

PR #5808, opened 2021-11-04
Confirm button state for BTC seller to be disabled in mediation

ghost:
### Current Scenario:

- Bob makes offer to buy XMR, Alice takes it.
- Alice initiates transfer of XMR and clicks Payment Started
- Bob does not respond for 24 hours, so Alice opens a Dispute.
- Mediator issues payout recommendation with a 5% trade penalty to Bob for not responding within the trade period.
- Bob does not respond for as long as he wishes, then finally clicks "Confirm Payment Receipt" ending the trade and skirting the penalty.
- Bob skirted mediation rules and as a result currently **mediators are banning such users from trading**.

That is not good for morale or liquidity.

&nbsp;

### Proposed Scenario:

- Bob makes offer to buy XMR, Alice takes it.
- Alice initiates transfer of XMR and clicks Payment Started
- Bob does not respond for 24 hours, so Alice opens a Dispute.
- Mediator issues payout recommendation with a 5% trade penalty to Bob for not responding within the trade period.
- Bob sees that his only options are to accept the mediation resolution or go for arbitration.

It already works this way for Fiat trades, and the penalty system seems to operate fairly.  We should adopt the same process for Altcoins.  Hence this PR.

---

image

### DIFF
--- a/core/src/main/java/bisq/core/trade/model/bisq_v1/SellerTrade.java
+++ b/core/src/main/java/bisq/core/trade/model/bisq_v1/SellerTrade.java
@@ -91,11 +91,6 @@ public abstract class SellerTrade extends Trade {
 
     @Override
     public boolean confirmPermitted() {
-        // For altcoin there is no reason to delay BTC release as no chargeback risk
-        if (CurrencyUtil.isCryptoCurrency(getOffer().getCurrencyCode())) {
-            return true;
-        }
-
         switch (getDisputeState()) {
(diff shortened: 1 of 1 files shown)


chimp1984 (approved):
utACK

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/5808#pullrequestreview-801812675

PR #5811, opened 2021-11-05
Remove apitest's create bsq-swap account method

ghubstan:
If not present, a default BSQ swap account is saved when a `User` object is initialized.  Use the existing default account in API bsq-swap test cases, and rename the legacy BSQ payment account fixtures to distinguish them from the new default 'BSQ Swap' payment account.


chimp1984 (approved):
utACK

PR #5812, opened 2021-11-07
Deprecate and replace hard-coded CLI console output formatters

ghubstan:
This PR adds a new CLI console output formatting api, which dynamically  formats and justifies console output via a table of columns, populated from protobuf messages.  New test cases are included to verify there are no differences between the old console output "api" and the new one.

Some refactoring of apitest trading test cases was done to reduce class size.

Based on PR https://github.com/bisq-network/bisq/pull/5811


chimp1984 (approved):
utACK

ripcurlx (approved):
utACK based on https://github.com/bisq-network/bisq/pull/5812#pullrequestreview-801818291

Issue #5574, opened 2021-06-16
Trade failed but security deposit is still locked


SoldadoBTC:
### Description

I made a transaction and it has failed (it got stucked in OPEN TRADES), i moved it to FAILED TRADES but the security deposit is still locked.

Already tried to remove PendingTrades files in db folder and do a spv resync only to remove invalid invalid transactions, but it did not work. Actually the locked funds disapeared as well as the transaction (it is not in FAILED nor in the OPEN TRADES anymore)

Bellow you can find the original screenshots of the transaction.

#### Version

1.6.5

### Steps to reproduce

I can't reproduce

### Expected behaviour

Locked funds to return to the available balance.

### Actual behaviour

Funds Locked

#### Device or machine

WINDOWS 10

#### Additional info
![Capturar](https://user-images.githubusercontent.com/85647096/122293941-ecb78600-cecd-11eb-893f-194ec837cf6a.PNG)
![Capturar2](https://user-images.githubusercontent.com/85647096/122293951-f04b0d00-cecd-11eb-8f9c-a4208b33edc0.PNG)
![Capturar3](https://user-images.githubusercontent.com/85647096/122293966-f50fc100-cecd-11eb-8667-59a73bd73217.PNG)
![Capturar4](https://user-images.githubusercontent.com/85647096/122293988-f8a34800-cecd-11eb-9802-0a96a80c10b3.PNG)

ripcurlx:
> Already tried to remove PendingTrades files in db folder and do a spv resync only to remove invalid invalid transactions, but it did not work. Actually the locked funds disapeared as well as the transaction (it is not in FAILED nor in the OPEN TRADES anymore)

So you are saying, after deleting the pending trades file, you don't see the locked deposit amount in your available balance anymore? Could you please verify if the deposit transaction got included in a block or not?

SoldadoBTC:
Yes, after deleting the pending trades file and resync i don't see the locked deposit neither in my available balance nor in my locked balance anymore it simply desapeard.

The deposit transaction (ef3126e5bceacb6e00ee673d1392e84084ffe6f8c48db51b93f6b93ec9ad7211 - showed in red in one of theprevious  screenshots) was not included in any block as screenshot below:
![Capturar5](https://user-images.githubusercontent.com/85647096/122467833-61072d80-cf91-11eb-95e9-0cb5f63d6f96.PNG)

ripcurlx:
[...]

PR #5813, opened 2021-11-08
Fix calculation & display of Locked Funds

### LINKED ISSUE #5703: Locked balance amount in menu bar is not consistent with "Locked Funds" tab
### Description

The "Locked BTC" in the upper right corner permanently shows amount of 0.006 BTC to me although I don't have any open trades neither do I have any open offer and nothing is shown in the "Locked Funds" tab. 
See the discrepancy on the screenshot:

image

#### Version

1.7.3

### Steps to reproduce

Alas I can't provide the steps for reproducing this because I don't know them. But it *might* possibly be related to the fact that once I had a failed transaction?

image

But I'm not very confident that this is related, because if I move that failed transaction back to the open ones and then I click the "Details" icon, I get this message which clearly states that no funds have been locked:

image

### Expected behaviour

I would expect that
 - either the "Locked" amount in the top right corner is an exact sum of the partial amounts on the "Locked Funds" tab.
 - or the "Locked Funds" tab lists all the sources which the "Locked" amount in the top right corner consists of.

### Actual behaviour

I see a mysterious 0.006 BTC value as the Locked amount and I don't know it thats a visual-only bug OR if I there is really 0.006 BTC locked which I could somehow reclaim.

#### Device or machine

Windows 10

ghost:
Fixes #5703, #5574

### Scenario:

If for any reason the deposit transaction is rejected by bitcoin nodes, Bisq thinks those funds are locked in a trade.  Bisq does not receive any notification about rejected transactions -- once they are broadcast via BitcoinJ they are assumed to be valid even though this may not be the case.  This can happen when for example if the user's wallet is out of sync and the deposit transaction is attempting to use already spent funds.  The only indication that something is wrong is that the deposit transaction never confirms (and if they look it up on a blockchain explorer the txId will not be found).  At this point Bisq will erroneously show the deposit amount in "Locked Funds" on the main screen.

This can be fixed by considering funds to be locked once the deposit transaction is **confirmed** on the blockchain.

----

### Code change:

We change one line in the method `isFundsLockedIn()` replacing `isDepositPublished` with `isDepositConfirmed`.

https://github.com/bisq-network/bisq/blob/851bdcf9ff0133462aa6e4e742fe59989a2d0db1/core/src/main/java/bisq/core/trade/model/bisq_v1/Trade.java#L921

In order to determine the possible impacts of this, here is a usage analysis of the altered method 'isFundsLockedIn'.
It is used in the following 3 locations:

**Funds -> Locked Funds Tab:**
    TradeManager.java +769 method getTradesStreamWithFundsLockedIn
        -> LockedView.java +224
            -> updates the amounts shown in Funds -> Locked Funds (active trades only)

**Main Screen Locked Balance:**
    TradeManager.java +769 method getTradesStreamWithFundsLockedIn
    ClosedTradableManager.java +156 method getTradesStreamWithFundsLockedIn
    FailedTradesManager.java +117 method getTradesStreamWithFundsLockedIn
        -> Balances.java +123
            -> sets the locked balance amount shown at the top right of the main screen

**Locked Funds Popup:**
    ClosedTradableManager.java +156 method getTradesStreamWithFundsLockedIn
    FailedTradesManager.java +117 method getTradesStreamWithFundsLockedIn
        Each of the above are used in the following two use cases:
        -> TradeManager.java +772 method getSetOfFailedOrClosedTradeIdsFromLockedInFunds
            -> BisqSetup.java  method checkForLockedUpFunds pops up a user directed warning message "you have locked up funds from a failed trade"

### DIFF
--- a/core/src/main/java/bisq/core/trade/model/bisq_v1/Trade.java
+++ b/core/src/main/java/bisq/core/trade/model/bisq_v1/Trade.java
@@ -919,8 +919,8 @@ public abstract class Trade extends TradeModel {
     }
 
     public boolean isFundsLockedIn() {
-        // If no deposit tx was published we have no funds locked in
-        if (!isDepositPublished()) {
+        // If no deposit tx was confirmed we have no funds locked in
+        if (!isDepositConfirmed()) {
             return false;
         }
 


chimp1984:
NACK

This will not fix it as the call checks if the state.ordinal is >= DEPOSIT_PUBLISHED so DEPOSIT_CONFIRMED is included in that as well.

You probably need to check the tx from the wallet to see if its really confirmed/available


```
   public boolean isDepositPublished() {
        return getTradePhase().ordinal() >= Phase.DEPOSIT_PUBLISHED.ordinal();
    }


 public enum Phase implements TradePhase {
        INIT,
        TAKER_FEE_PUBLISHED,
        DEPOSIT_PUBLISHED,
        DEPOSIT_CONFIRMED,
        FIAT_SENT,
        FIAT_RECEIVED,
        PAYOUT_PUBLISHED,
        WITHDRAWN;
```

ripcurlx:
@jmacxx Did you have time already to look at @chimp1984 's remark?

ghost:
I did test this fix while developing it.  I was going to return and re-do the tests before commenting again.  They do take a few hours to run and I still have a PR to finish work on before I can get to this.

ghost:
The easiest consistent way to reproduce the locked up funds issue described in #5703, is by not signing one of the deposit transaction inputs causing it to be rejected by bitcoind.  Achieve that in test commenting out `TradeWalletService` L645-L646.

You end up with a trade that looks ok, but its deposit tx never gets a confirm.

At that point you'll decide to do an SPV resync, reconciling BitcoinJ state with the blockchain.  Upon restarting Bisq you'll encounter the following:

- Locked funds amount showing the trade deposit value.
- A popup which says the deposit tx has missing inputs.
- Red failed icons in the pending trades screen.
- Nothing shown in Funds -> Locked funds tab.

<details>
  <summary>Screenshots - click to expand</summary>

failed_trade_popup_missing_deposit


failed_trade_locked_funds

failed_trade_details


failed_trade_move_to




</details>


Moving that trade to Failed Trades shows the dialog which says there are no locked funds, contradicting what is shown in the top right of the screen at Locked Funds display.

I re-did my testing to obtain these screenshots.

---

After the patch is applied, the Locked Funds no longer displays the fake locked amount.

**confirmed** deposit == locked funds.

image

chimp1984:
Ah sorry my earlier comment was wrong. As we use the negation its a difference if the state is published or confirmed for that case. 
Could you check what state the trade is in? Probably in pusblished state but not confirmed state...

ghost:
Yes published, not confirmed.

chimp1984 (approved):
utACK

chimp1984:
> Yes published, not confirmed.

Ok. From my understanding it should be ok then.

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/5813#issuecomment-968335776

pazza83:
Hi @jmacxx will this fix users with historic locked funds showing? Assuming they will have to do an spv resync?

ghost:
Yes.

PR #5815, opened 2021-11-09
Fix Git LFS pointers

ripcurlx:


chimp1984 (approved):
utACK

bisq-github-admin-3 (approved):
ACK - Tested it with a fresh repo clone

PR #5816, opened 2021-11-09
Fix missing handling for non-mainnet use case

chimp1984:


ghubstan (approved):
Tested ACK

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/5816#pullrequestreview-801958332

PR #5819, opened 2021-11-10
Fix reorg handling

### LINKED ISSUE #5809: Bug at reorgs
When a reorg happens the app gets in a loop of deleting daoState to get it replaced from resources and requesting new blocks which triggers the cycle again... 

I will look into fixing it after the open PRs are merged.
Simple version would be to enforce a restart. Other more complex option is to replace daoState with resources without restart. But that might cause issues in clients if they do not listen to that event and not getting updated. 
Not sure how often reorgs happen in Bitcoin, I guess very rarely. If so the restart options seems ok and carries less risk.

chimp1984:
Fixes https://github.com/bisq-network/bisq/issues/5809

How to reproduce:
- Start a new regtest setup with a seed, a full node (Alice) and a lite node (Bob)
- Make a backup of that start setup version. Assume block height is 113 here.
- Create 1 block
- Create 1 tx 
- Create 1 block. Block height is 115.
- Stop BTC core replace the blocks and regtest folders in btc core data dir with the folders from the start setup backup. This is used for simulating a reorg. 
- Create 2 blocks. Still no reorg as fork chain has same height 115 as existing chain.
- Create 1 more block triggering the reorg at 116.

In master the Alice nad Bob nodes get caught in an endless loop requesting new blocks. The seed node is shut down automatically. Bob will start the endless loop only after the seed has restarted and the next block is propagated.

With the PR version there is just the shutdown at seed node and the shutdown popup at desktop apps but no loops.
The litenode will not recognize the new block as the full nodes are shut down or inactive. Only after restarting the seed and creating a new block the lite node will detect the reorg. The block propagation is delayed with a random delay so it can take a bit... 

To avoid that all seed nodes shut down at the same time when a reorg happens we add a detereministic delay derived from the index of the seed in the seed node list. So shut down is delayed by 30 sec * index, distributing the shutdown of 12 seeds over 6 minutes.

### DIFF
--- a/core/src/main/java/bisq/core/dao/node/full/FullNode.java
+++ b/core/src/main/java/bisq/core/dao/node/full/FullNode.java
@@ -107,7 +107,18 @@ public class FullNode extends BsqNode {
 
     @Override
     protected void startParseBlocks() {
-        requestChainHeadHeightAndParseBlocks(getStartBlockHeight());
+        int startBlockHeight = getStartBlockHeight();
+
+        log.info("startParseBlocks: startBlockHeight={}", startBlockHeight);
+        rpcService.requestChainHeadHeight(chainHeight -> {
+                    // If our persisted block is equal to the chain height we have startBlockHeight 1 block higher,
+                    // so we do not call parseBlocksOnHeadHeight
+                    log.info("startParseBlocks: chainHeight={}", chainHeight);
+                    if (startBlockHeight <= chainHeight) {
+                        parseBlocksOnHeadHeight(startBlockHeight, chainHeight);
+                    }
+                },
+                this::handleError);
     }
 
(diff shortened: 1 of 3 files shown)


ghost (approved):
utACK

ripcurlx (approved):
ACK - Tested the process on Regtest and was able to simulate the behavior as described.

- Seednode shut down
- Full node got notified to shut down
Bildschirmfoto 2021-11-12 um 09 40 13
- Light node didn't recognize the reorg yet
- Seednode and Alice (fullnode) were restarted
- After creating new blocks, Bob (light node) recognized the reorg
Bildschirmfoto 2021-11-12 um 10 43 51

After restart of Light node everything went back to normal.

PR #5831, opened 2021-11-15
Fix display of missing Failed Trade tab

ghost:
The addition of BSQ Swaps tab under portfolio view inadvertently caused the Failed Trades tab to not appear since it was hard coded to expect 3 tabs before it.  

image


ripcurlx (approved):
utACK - Thanks for catching this one @jmacxx 👍

PR #5834, opened 2021-11-16
Add penalty calculation to dispute agent UI

ghost:
Requested by @leo816 and @pazza83.
@refund-agent2 please review.

- Adds penalty calculation options to the disputes UI.  
- Enables agents to apply penalties 
 
## Specifications

- Formulas defined per https://bisq.wiki/Table_of_penalties
- [mediationPenaltyCalc.xls](https://github.com/bisq-network/bisq/files/7543131/mediationPenaltyCalc.xls)
 

## Screenshots

image

image


leo816:
@refund-agent2 please review

refund-agent2:
I can't review properly to make sure it works ok, but it seems ok to make it easier to apply the Penalties table. The only problem I find for my use of it is that it doesn''t include arbitration fees.

ghost:
Thanks for your review comments here and on keybase.
I have made the suggestions WRT increasing the height of the summary text area to 8 lines, and removing 'loser as publisher' as it is no longer used.  We can address arbitration fees with a future PR after discussion has taken place.

arbitration_screen

ripcurlx (approved):
utACK

PR #5838, opened 2021-11-18
Fix broken test for new build

ripcurlx:
It seems that the new build process exposed a timing issue in the setting of the locale that wasn't existing before.
This PR sets the locale additionally in the GlobalSettings.
This bug only surfaces for non-english setups.


ripcurlx:
Waiting for Gradle Verification fix before I can rebase this test fix on master.

cbeams (dismissed):
ACK. This builds fine for me locally. The CI build failure here is a red herring. The underlying cause was just fixed in https://github.com/bisq-network/bisq/pull/5839.

bisq-github-admin-3 (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/5838#pullrequestreview-809818750

PR #5824, opened 2021-11-12
Upgrade to Gradle 7.3

cbeams:
This PR successfully upgrades the build from Gradle 6.6.1 to 7.3, but with a number of caveats that need to be worked out:

- [x] Our Gradle Witness plugin fork needs to be updated such that it no longer depends on the now-removed `compile` configuration. The witness plugin has been disabled in the meantime, and it should be re-enabled when fixed.
- [x] There is quite a bit of duplication now in repeated `implementation` dependency declarations, now that transitive `compile` declarations are no longer supported. This may be fine in the end (the explicitness is a good thing), but it's worth considering if we want to tidy them up in some way.
- [x] The pricenode should be run and thoroughly tested, as its underlying Spring Boot dependency went through a major upgrade from 1.5.10-RELEASE => 2.5.6. As part of this upgrade, the `/getVersion` endpoint has been removed entirely. @bisq-network/pricenode-operators, is this endpoint something you actually use, or is it fine to remove?

Background: the reason I set out on this upgrade was part of the process of getting Bisq to build on an M1 Mac. That still appears impossible due to a bug in JFoenix's progress meter. See https://github.com/cbeams/bisq/commit/3f4fba9 for details. In the meantime, building on an M1 Mac does work so long as you pass `-x :desktop:test`. Actually running the app, e.g. with `./gradlew :desktop:run`, however, does not and will not work until that bug is fixed. In any case, it will be easier to get Bisq building on M1 with this PR merged.

PR #5839, opened 2021-11-18
Add missing metadata for jackson-base pom 2.12.1

cbeams:
Problem: When merging #5824, the absence of this entry caused a build
failure at dependency verification time against JDK11 and JDK15 on
Ubuntu-latest [1]. It may also cause failures on other JDK/OS
combinations, but the GitHub workflow was aborted before those failures
couldhave occurred. In any case, this omission did not create build
failures on any of the local development machines that tested the
aforementioned PR. Reasons for this discrepancy are unknown.

Solution: manually fetch the pom from [2], run `sha256sum` on it locally
and commit the result to the verification metadata file.

[1]: https://github.com/bisq-network/bisq/runs/4249640611?check_suite_focus=true#step:6:33
[2]: https://repo1.maven.org/maven2/com/fasterxml/jackson/jackson-base/2.11.1/jackson-base-2.11.1.pom


ripcurlx (approved):
utACK

PR #5849, opened 2021-11-20
Add checks for trade statistics

chimp1984:
Add check for trade statistic for max amount and if currency exists in CurrencyUtil


ripcurlx (approved):
utACK

PR #5850, opened 2021-11-21
Attempt to fix guava dependency issue

cbeams:
This is a quick attempt to fix the issue described by @jmacxx in Keybase
similar to what's described at
https://github.com/jeremylong/DependencyCheck/issues/3221.

It needs to be tested in the environment where the issue was observed
(Windows).

Fixes #replaceWithIssueNr, fixes #replaceWithIssueNr

Your PR description here.

PR #5851, opened 2021-11-21
Fix guava dependency issue

cbeams:
Problem: a

    NoSuchMethodError: 'java.util.stream.Collector
    com.google.common.collect.ImmutableMultiset.toImmutableMultiset()'

exception was being thrown when testing the previously-merged upgrade to
Gradle 7.3, as described at keybase://chat/bisq#testing/2466.

Solution: This problem is similar to the issue reported at
jeremylong/DependencyCheck#3221. The source of the problem was multiple
conflicting guava jars on the runtime classpath. This commit upgrades to
guava 30.1.1-jre which ensures a single jar on the classpath.

NOTE: This PR supersedes #5850, which was submitted incorrectly.


ghubstan:
Tested ACK

See my notes about guava version conflict on Ubuntu:  https://github.com/bisq-network/bisq/pull/5850#issuecomment-974844250

ripcurlx (approved):
utACK - based on feedback from @jmacxx and @ghubstan. On macOS it wasn't reproducable for me.

PR #5852, opened 2021-11-22
Support and test creation of Swift accounts via API

ghubstan:
The Swift account is one of the more complicated setups, and the API adds instructions for defining them in a payment account json form.


ripcurlx (approved):
ACK - Tested it on Regtest. Everything worked as expected.

PR #5853, opened 2021-11-22
Improve TradeStatistics validation

chimp1984:
Only apply maxTradeLimit from PaymentMethod if the PaymentMethod is found in the active list. Otherwise the 2 BTC default is used.
We get TradeStatistics3 objects from old retired PaymentMethods
which are not found in the active list.

### DIFF
--- a/core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java
+++ b/core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java
@@ -437,7 +437,12 @@ public final class TradeStatistics3 implements ProcessOncePersistableNetworkPayl
         }
         long maxTradeLimit = Coin.COIN.multiply(2).value;
         try {
-            maxTradeLimit = PaymentMethod.getPaymentMethodById(getPaymentMethod()).getMaxTradeLimitAsCoin(currency).value;
+            // We cover only active payment methods. Retired ones will not be found by getActivePaymentMethodById.
+            String paymentMethod = getPaymentMethod();
+            Optional<PaymentMethod> optionalPaymentMethodById = PaymentMethod.getActivePaymentMethodById(paymentMethod);
+            if (optionalPaymentMethodById.isPresent()) {
+                maxTradeLimit = optionalPaymentMethodById.get().getMaxTradeLimitAsCoin(currency).value;
+            }
         } catch (Exception ignore) {
         }
         return amount > 0 &&
--- a/core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java
+++ b/core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java
@@ -398,7 +398,7 @@ public final class TradeStatistics3 implements ProcessOncePersistableNetworkPayl
         refundAgent = null;
     }
 
-    public String getPaymentMethod() {
+    public String getPaymentMethodId() {
         try {
             return PaymentMethodMapper.values()[Integer.parseInt(paymentMethod)].name();
         } catch (Throwable ignore) {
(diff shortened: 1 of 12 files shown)


core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java
```diff
@@ -437,7 +437,12 @@ public boolean isValid() {
         }
         long maxTradeLimit = Coin.COIN.multiply(2).value;
         try {
-            maxTradeLimit = PaymentMethod.getPaymentMethodById(getPaymentMethod()).getMaxTradeLimitAsCoin(currency).value;
+            // We cover only active payment methods. Retired ones will not be found by getActivePaymentMethodById.
+            String paymentMethod = getPaymentMethod();
```
ghubstan: `utACK` with a comment about line 441's var name not being `paymentMethodId`.  

It's clear from line 401's method declaration `public String getPaymentMethod()` that this is an ID (name).  But maybe line 401 could also be changed to `public String getPaymentMethodId()`?

core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java
```diff
@@ -437,7 +437,12 @@ public boolean isValid() {
...
+            Optional<PaymentMethod> optionalPaymentMethodById = PaymentMethod.getActivePaymentMethodById(paymentMethod);
+            if (optionalPaymentMethodById.isPresent()) {
+                maxTradeLimit = optionalPaymentMethodById.get().getMaxTradeLimitAsCoin(currency).value;
+            }
         } catch (Exception ignore) {
         }
```
ghubstan, line 448: Maybe `log.warn(ignored)` exception,  just in case it hides a bug?

core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java
chimp1984: Actually its not the ID (ending with _ID) but the name of the payment method. It was used so far only for display. Would have been better to use the ID...

core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java
chimp1984: Was thinking as well but the only area where it could throw is when the payment method is not in our defined list inside tradestats, which is an expected case when a user has not updated and new payment methods got added.

chimp1984:
``` public String getPaymentMethod() {
        try {
            return PaymentMethodMapper.values()[Integer.parseInt(paymentMethod)].name();
        } catch (Throwable ignore) {
            return paymentMethod;
        }
    }```

Comment (chimp1984) [reply core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java:448]: Added a warn....

File: core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java (hunk 1 above)
Comment (chimp1984) [reply core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java]: Renamed it...

File: core/src/main/java/bisq/core/trade/statistics/TradeStatistics3.java [hunk 3]
```diff
@@ -443,7 +443,8 @@ public boolean isValid() {
             if (optionalPaymentMethodById.isPresent()) {
                 maxTradeLimit = optionalPaymentMethodById.get().getMaxTradeLimitAsCoin(currency).value;
             }
-        } catch (Exception ignore) {
+        } catch (Exception e) {
+            log.warn("Error at isValid(). " + e.toString(), e);
```
ghubstan: I'm prettry sure `e.toString(),  e` will result in a duplicated stack trace in the log.   Maybe `log.warn("Error at isValid(). {}", e);` is enough?

ripcurlx: I think so too. The first object that is passed is called `.toString` upon. I'll push a fix to this branch.

ripcurlx: I think even `log.warn( "Error at isValid().", e );` will be sufficient.

ripcurlx (approved):
utACK

PR #5854, opened 2021-11-22
Improve FilterManager

chimp1984:
Only try to remove invalid filter from network if it's our own. This avoids unnecessary/confusing warning logs in data storage.


ghubstan (approved):
utACK

ripcurlx (approved):
utACK

PR #5857, opened 2021-11-23
Enforce inclusion of TradeLimits for headless apps

chimp1984:
PaymentMethod use an instance of TradeLimits and expect that it
has been injected, which is the case for desktop but not for
headless apps, so we enforce injection in the app base classes
used for headless apps.

The validation of trade statistics use a method in PaymentMethod where that
dependency is required.

Tha hack how the PaymentMethod use TradeLimits is not nice, but
would require more effort for refactoring.

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqHeadlessAppMain.java
+++ b/core/src/main/java/bisq/core/app/BisqHeadlessAppMain.java
@@ -17,6 +17,8 @@
 
 package bisq.core.app;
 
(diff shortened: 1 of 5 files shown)


ripcurlx (dismissed):
ACK - Tested on Regtest.

chimp1984:
@ripcurlx  Please ignore the Codacy complaints.

ripcurlx (approved):
utACK

PR #5858, opened 2021-11-23
Provide ASIC resistant PoW scheme for BSQ swaps

stejbac:
This PR adds an implementation of the [Equihash](https://eprint.iacr.org/2015/946.pdf) memory-hard Proof of Work scheme for DoS/spam attack protection, to replace the present simple Hashcash scheme. It also adds a simple versioning system to make it easier to migrate from one algorithm to another in case of failure to provide adequate DoS attach protection, that is, if an attacker spamming the BSQ swap offer book has too much of a speed/efficiency advantage over an honest user, e.g. due to use of FPGAs, GPUs, etc.

The Equihash scheme is _asymmetric_, meaning that producing a PoW by solving a hashing puzzle is slow and consumes up to MBs of memory, but verification is instant. The puzzle solutions are larger than that of the present Hashcash scheme (where it is just a 64-bit counter/nonce) but still fairly small - 10's of bytes to a few kB in size, depending on the puzzle parameters. So they shouldn't bloat the BSQ swap offer payloads significantly.

--

Note that one of the commits (5da8df2) touches the DAO packages (VoteResultService), although the code change is trivial.

Also, the last commit (e0595aa) makes a breaking change the proto serialisation, by changing the Filter/PoW difficulty field to a linear-scale floating point -- that commit isn't essential to the PR, though it would cause less disruption to include it before the 1.8.0 release.


proto/src/main/proto/pb.proto
```diff
@@ -882,7 +882,7 @@ message ProofOfWork {
     bytes payload = 1;
     int64 counter = 2;
     bytes challenge = 3;
-    bytes difficulty = 4;
+    double difficulty = 4;
```
chimp1984, line 885: This cause an error popup for users running master. The `new BigInteger(difficulty)` at `getNumLeadingZeros` with a empty byte array fails.
Only happens with published BSQ offers (I was only one who had a BSQ offer and I just removed it). So we could ignore it.... Or we deprecate that existing fields and add a new one for the double type? But I guess its not worth to pollute the protofile with that short term tiny risk.... Would also only affect users running current master before that PR is merged.

proto/src/main/proto/pb.proto
```diff
@@ -770,6 +770,7 @@ message Filter {
     bool disable_mempool_validation = 28;
     bool disable_pow_message = 29;
     int32 pow_difficulty = 30;
+    repeated int32 enabled_pow_versions = 35;
```
chimp1984: I understand that it makes sense to keep it contextually close to the other pow field but I think we follow the convention that we keep the fields ordered by the field index. So I think it would be better to add it at the end of the fields. But no strong opinion either....

core/src/main/java/bisq/core/filter/FilterManager.java
```diff
@@ -493,13 +494,23 @@ public boolean isProofOfWorkValid(Offer offer) {
...
         checkArgument(offer.getBsqSwapOfferPayload().isPresent(),
                 "Offer payload must be BsqSwapOfferPayload");
-        return HashCashService.verify(offer.getBsqSwapOfferPayload().get().getProofOfWork(),
-                HashCashService.getBytes(offer.getId() + offer.getOwnerNodeAddress().toString()),
+        ProofOfWork pow = offer.getBsqSwapOfferPayload().get().getProofOfWork();
+        var service = ProofOfWorkService.forVersion(pow.getVersion());
+        if (!service.isPresent() || !getEnabledPowVersions().contains(pow.getVersion())) {
+            return false;
+        }
+        return service.get().verify(offer.getBsqSwapOfferPayload().get().getProofOfWork(),
```
chimp1984: We could use the `pow` variable from line 497....

common/src/main/java/bisq/common/crypto/EquihashProofOfWorkService.java
```diff
@@ -0,0 +1,76 @@
...
+            log.info("Completed minting proofOfWork: {}", proofOfWork);
+            return proofOfWork;
+        });
+    }
+
+    @Override
+    public byte[] getChallenge(String itemId, String ownerId) {
+        checkArgument(!StringUtils.contains(itemId, '\0'));
+        checkArgument(!StringUtils.contains(ownerId, '\0'));
+        return Hash.getSha256Hash(checkNotNull(itemId) + "\0" + checkNotNull(ownerId));
```
chimp1984: Why are the 2 strings concatenated with a null char?

common/src/main/java/bisq/common/crypto/EquihashProofOfWorkService.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    boolean verify(ProofOfWork proofOfWork) {
+        double difficulty = adjustedDifficulty(proofOfWork.getNumLeadingZeros());
+
+        var puzzle = new Equihash(90, 5, difficulty).puzzle(proofOfWork.getChallenge());
+        return puzzle.deserializeSolution(proofOfWork.getPayload()).verify();
```
chimp1984: I think it would be better to extends the `ProofOfWork` fields to support the storage of the pow solution instead using the payload.
For the use case of the offers it might not matter as the payload is kind of irrelevant as the challenge contains already the offerID as well, but for other potential use cases (e.g. p2p network dos protection) it would be more useful to have a field for the payload and one for the challenge (which could be an interactive challenge nonce) and then have once field for the pow solution (counter in hashcash). Maybe to keep counter as informational field for the number of iterations but not clear yet to me if that makes sense for Equihash...

chimp1984:
@stejbac 
I think the predicates in FilterManager and HashCashService are just complicating and bloating the APIs. The intention was that they might be useful for testing but at the end they are unnecessary. I would prefer to inline them. 
I created 2 commits which I will post here... Feel free to add them if you agree to the change. I think its easier that way as that I make a PR to you repo (can do that as well if you prefer).

chimp1984:
```
Index: common/src/main/java/bisq/common/crypto/ProofOfWorkService.java
IDEA additional info:
<+>UTF-8
diff --git a/common/src/main/java/bisq/common/crypto/ProofOfWorkService.java b/common/src/main/java/bisq/common/crypto/ProofOfWorkService.java
--- a/common/src/main/java/bisq/common/crypto/ProofOfWorkService.java	(revision e0595aa284ce18ab7602f1126d23014da336ef69)
+++ b/common/src/main/java/bisq/common/crypto/ProofOfWorkService.java	(revision e0e70bd804fec36844987bc1a29051a24e3cbdc4)
@@ -19,9 +19,9 @@
 
 import com.google.common.base.Preconditions;
 
+import java.util.Arrays;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import java.util.function.BiPredicate;
 
 import lombok.Getter;
 
@@ -58,15 +58,13 @@
     public boolean verify(ProofOfWork proofOfWork,
                           String itemId,
                           String ownerId,
-                          double controlDifficulty,
-                          BiPredicate<byte[], byte[]> challengeValidation,
-                          BiPredicate<Double, Double> difficultyValidation) {
+                          double controlDifficulty) {
 
         Preconditions.checkArgument(proofOfWork.getVersion() == version);
 
         byte[] controlChallenge = getChallenge(itemId, ownerId);
-        return challengeValidation.test(proofOfWork.getChallenge(), controlChallenge) &&
-                difficultyValidation.test(proofOfWork.getDifficulty(), controlDifficulty) &&
+        return Arrays.equals(proofOfWork.getChallenge(), controlChallenge) &&
+                proofOfWork.getDifficulty() >= controlDifficulty &&
                 verify(proofOfWork);
     }
 }
Index: core/src/main/java/bisq/core/filter/FilterManager.java
IDEA additional info:
<+>UTF-8
diff --git a/core/src/main/java/bisq/core/filter/FilterManager.java b/core/src/main/java/bisq/core/filter/FilterManager.java
--- a/core/src/main/java/bisq/core/filter/FilterManager.java	(revision e0595aa284ce18ab7602f1126d23014da336ef69)
+++ b/core/src/main/java/bisq/core/filter/FilterManager.java	(revision e0e70bd804fec36844987bc1a29051a24e3cbdc4)
@@ -37,9 +37,9 @@
 import bisq.common.app.Version;
 import bisq.common.config.Config;
 import bisq.common.config.ConfigFileEditor;
-import bisq.common.crypto.ProofOfWorkService;
 import bisq.common.crypto.KeyRing;
 import bisq.common.crypto.ProofOfWork;
+import bisq.common.crypto.ProofOfWorkService;
 
 import org.bitcoinj.core.ECKey;
 import org.bitcoinj.core.Sha256Hash;
@@ -58,7 +58,6 @@
 
 import java.math.BigInteger;
 
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
@@ -66,7 +65,6 @@
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.function.BiPredicate;
 import java.util.function.Consumer;
 
 import java.lang.reflect.Method;
@@ -88,12 +86,6 @@
     private static final String BANNED_SEED_NODES = "bannedSeedNodes";
     private static final String BANNED_BTC_NODES = "bannedBtcNodes";
 
-    private final BiPredicate<byte[], byte[]> challengeValidation = Arrays::equals;
-    // We only require a new pow if difficulty has increased
-    private final BiPredicate<Double, Double> difficultyValidation =
-            (value, controlValue) -> value - controlValue >= 0;
-
-
     // Listener
@@ -501,9 +493,7 @@
         }
         return service.get().verify(offer.getBsqSwapOfferPayload().get().getProofOfWork(),
                 offer.getId(), offer.getOwnerNodeAddress().toString(),
-                filter.getPowDifficulty(),
-                challengeValidation,
-                difficultyValidation);
+                filter.getPowDifficulty());
     }
 
     public List<Integer> getEnabledPowVersions() {

```

chimp1984:
```
Index: common/src/main/java/bisq/common/crypto/HashCashService.java
IDEA additional info:
<+>UTF-8
diff --git a/common/src/main/java/bisq/common/crypto/HashCashService.java b/common/src/main/java/bisq/common/crypto/HashCashService.java
--- a/common/src/main/java/bisq/common/crypto/HashCashService.java	(revision e0e70bd804fec36844987bc1a29051a24e3cbdc4)
+++ b/common/src/main/java/bisq/common/crypto/HashCashService.java	(revision d54f1aafc04601ec350b083d06b24b4774a08d6b)
@@ -22,24 +22,18 @@
 
 import java.nio.charset.StandardCharsets;
 
-import java.util.Arrays;
 import java.util.concurrent.CompletableFuture;
-import java.util.function.BiPredicate;
 
 import lombok.extern.slf4j.Slf4j;
 
 /**
  * HashCash implementation for proof of work
- * It doubles required work by difficulty increase (adding one leading zero).
+ * It doubles required work by log2Difficulty increase (adding one leading zero).
  *
  * See https://www.hashcash.org/papers/hashcash.pdf
  */
 @Slf4j
 public class HashCashService extends ProofOfWorkService {
-    // Default validations. Custom implementations might use tolerance.
-    private static final BiPredicate<byte[], byte[]> isChallengeValid = Arrays::equals;
-    private static final BiPredicate<Integer, Integer> isDifficultyValid = Integer::equals;
-
     HashCashService() {
         super(0);
     }
@@ -50,105 +44,35 @@
         return mint(payload, challenge, difficulty);
     }
 
-    @Override
-    public byte[] getChallenge(String itemId, String ownerId) {
-        return getBytes(itemId + ownerId);
-    }
-
-    static CompletableFuture<ProofOfWork> mint(byte[] payload,
+    public CompletableFuture<ProofOfWork> mint(byte[] payload,
                                                byte[] challenge,
                                                double difficulty) {
-        return HashCashService.mint(payload,
-                challenge,
-                difficulty,
-                HashCashService::testDifficulty);
-    }
-
-    @Override
-    boolean verify(ProofOfWork proofOfWork) {
-        return verify(proofOfWork,
-                proofOfWork.getChallenge(),
-                toNumLeadingZeros(proofOfWork.getDifficulty()));
-    }
-
-    static boolean verify(ProofOfWork proofOfWork,
-                          byte[] controlChallenge,
-                          int controlLog2Difficulty) {
-        return HashCashService.verify(proofOfWork,
-                controlChallenge,
-                controlLog2Difficulty,
-                HashCashService::testDifficulty);
-    }
-
-    static boolean verify(ProofOfWork proofOfWork,
-                          byte[] controlChallenge,
-                          int controlLog2Difficulty,
-                          BiPredicate<byte[], byte[]> challengeValidation,
-                          BiPredicate<Integer, Integer> difficultyValidation) {
-        return HashCashService.verify(proofOfWork,
-                controlChallenge,
-                controlLog2Difficulty,
-                challengeValidation,
-                difficultyValidation,
-                HashCashService::testDifficulty);
-    }
-
-    private static boolean testDifficulty(byte[] result, int log2Difficulty) {
-        return HashCashService.numberOfLeadingZeros(result) > log2Difficulty;
-    }
-
-
-    ///////////////////////////////////////////////////////////////////////////////////////////
-    // Generic
-    ///////////////////////////////////////////////////////////////////////////////////////////
-
-    static CompletableFuture<ProofOfWork> mint(byte[] payload,
-                                               byte[] challenge,
-                                               double difficulty,
-                                               BiPredicate<byte[], Integer> testDifficulty) {
         return CompletableFuture.supplyAsync(() -> {
             long ts = System.currentTimeMillis();
             int log2Difficulty = toNumLeadingZeros(difficulty);
-            byte[] result;
+            byte[] hash;
             long counter = 0;
             do {
-                result = toSha256Hash(payload, challenge, ++counter);
+                hash = toSha256Hash(payload, challenge, ++counter);
             }
-            while (!testDifficulty.test(result, log2Difficulty));
+            while (numberOfLeadingZeros(hash) <= log2Difficulty);
             ProofOfWork proofOfWork = new ProofOfWork(payload, counter, challenge, difficulty, System.currentTimeMillis() - ts, 0);
             log.info("Completed minting proofOfWork: {}", proofOfWork);
             return proofOfWork;
         });
     }
 
-    static boolean verify(ProofOfWork proofOfWork,
-                          byte[] controlChallenge,
-                          int controlLog2Difficulty,
-                          BiPredicate<byte[], Integer> testDifficulty) {
-        return verify(proofOfWork,
-                controlChallenge,
-                controlLog2Difficulty,
-                HashCashService.isChallengeValid,
-                HashCashService.isDifficultyValid,
-                testDifficulty);
-    }
-
-    static boolean verify(ProofOfWork proofOfWork,
-                          byte[] controlChallenge,
-                          int controlLog2Difficulty,
-                          BiPredicate<byte[], byte[]> challengeValidation,
-                          BiPredicate<Integer, Integer> difficultyValidation,
-                          BiPredicate<byte[], Integer> testDifficulty) {
-        return challengeValidation.test(proofOfWork.getChallenge(), controlChallenge) &&
-                difficultyValidation.test(toNumLeadingZeros(proofOfWork.getDifficulty()), controlLog2Difficulty) &&
-                verify(proofOfWork, testDifficulty);
-    }
-
-    private static boolean verify(ProofOfWork proofOfWork, BiPredicate<byte[], Integer> testDifficulty) {
-        byte[] hash = HashCashService.toSha256Hash(proofOfWork.getPayload(),
+    @Override
+    boolean verify(ProofOfWork proofOfWork) {
+        byte[] hash = toSha256Hash(proofOfWork.getPayload(),
                 proofOfWork.getChallenge(),
                 proofOfWork.getCounter());
-        return testDifficulty.test(hash, toNumLeadingZeros(proofOfWork.getDifficulty()));
+        return numberOfLeadingZeros(hash) > toNumLeadingZeros(proofOfWork.getDifficulty());
+    }
+
+    @Override
+    public byte[] getChallenge(String itemId, String ownerId) {
+        return getBytes(itemId + ownerId);
     }
 
 
Index: common/src/test/java/bisq/common/crypto/HashCashServiceTest.java
IDEA additional info:
<+>UTF-8
diff --git a/common/src/test/java/bisq/common/crypto/HashCashServiceTest.java b/common/src/test/java/bisq/common/crypto/HashCashServiceTest.java
--- a/common/src/test/java/bisq/common/crypto/HashCashServiceTest.java	(revision e0e70bd804fec36844987bc1a29051a24e3cbdc4)
+++ b/common/src/test/java/bisq/common/crypto/HashCashServiceTest.java	(revision d54f1aafc04601ec350b083d06b24b4774a08d6b)
@@ -77,7 +77,7 @@
         List<ProofOfWork> tokens = new ArrayList<>();
         for (int i = 0; i < numTokens; i++) {
             byte[] challenge = UUID.randomUUID().toString().getBytes(StandardCharsets.UTF_8);
-            tokens.add(HashCashService.mint(payload, challenge, difficulty).get());
+            tokens.add(new HashCashService().mint(payload, challenge, difficulty).get());
         }
         double size = tokens.size();
         long ts2 = System.currentTimeMillis();

```

chimp1984:
@stejbac 
As far as I understand (and tested) the behaviour in case of  2 pow version (e.g. [0,1] or [1,0] set as enabled versions) is that the first is the preferred one and both are valid. So if I had an offer with hashcash pow and we add Equihash as first item I dont need to recreate the pow. But for a new offer Equihash will be used. If we switch from [0] to [1] then all my offers need to recreate the pow.
If difficulty from filter decreases there is no need to recreate the pow of an existing offer, only for increases pow need to be redone. 

The difficulty value to be set is 2 pow x, e.g. 2 pow 18 = 262144 means 18 leading zeros for hashcash. It is roughly the number if iterations for hashcash. Values in the range of 300 000-800 000 produce on my machine a pow duration of about 0.5-4 sec.

Could we add some documentation to the Filter so that the Filter maintainer finds quickly some guidance when he sets or changes the values?

proto/src/main/proto/pb.proto
```diff
@@ -769,7 +769,7 @@ message Filter {
     bool disable_api = 27;
     bool disable_mempool_validation = 28;
     bool disable_pow_message = 29;
-    int32 pow_difficulty = 30;
+    double pow_difficulty = 30;
```
chimp1984, line 772: The values are rather large (e.g. 500 000). I understand that in Equihash a double is used, but wouldn't it be sufficiet to convert an integer value to double and leave the proto field type as int32?

stejbac:
@chimp1984 I've just pushed those two patches now (as a single commit).

Yes, the list of enabled version numbers is in order of preference, so that the first is preferred when creating PoWs and they are both allowed when verifying PoWs. Adding Equihash to the list (in any order) won't cause Hashcash PoWs to be recreated, unless Hashcash is also removed from the list.

Generation times for Hashcash & Equihash are intended to be roughly the same for any given difficulty (at least when it's a power-of-two), with a difficulty of _2 pow n_ meaning the difficulty of finding > _n_ leading zeros for Hashcash (or equivalently finding a hash with exactly _n_ leading zeros), so _2 * (2 pow n)_ Hashcash iterations on average, i.e. 524288 average iterations for a difficulty of 2 pow 18 = 262144.

I can attempt to update the BSQ Swaps wiki, as information there about the difficulty would be out of date with this PR.

stejbac:
I just discovered a trivial bug in the iterator returned by `Equihash.IntListMultimap::get`, which was causing puzzle solutions (and hence Equihash proofs-of-work) to be invalid about 3% of the time. I've committed a fix (0ee6175), pushed above.

chimp1984 (changes requested):

proto/src/main/proto/pb.proto
stejbac: That would possibly be OK. But on a fast CPU running the PoW search in parallel (if we made an enhancement to do that), it might take around 1 microsecond per unit difficulty or less, so a difficulty set to INT_MAX would correspond to an average search time of less that 35 minutes on a fast machine. I'm not sure we would ever want to set the difficulty that high in case of heavy DoS attack, but it would still make posting BSQ swap offers just barely usable for people with fast machines.

common/src/main/java/bisq/common/crypto/EquihashProofOfWorkService.java
stejbac: I just thought it less likely that a string with an embedded null would ever be passed to the method than a string containing whitespace, but I guess any reasonable separator would do.

chimp1984: Ah ok.

proto/src/main/proto/pb.proto
chimp1984: Ah ok. Yes then we should not limit it by choosing a int type. What about a long type? Would it hurt anything on the Equihash side? But no strong opinion, just was wondering why a double is used.

common/src/main/java/bisq/common/crypto/EquihashProofOfWorkService.java
stejbac: Perhaps it would be better to do something like the following:

>◌◌◌◌public◌byte[]◌getChallenge(String◌itemId,◌String◌ownerId)◌{
>◌◌◌◌◌◌◌◌String◌escapedItemId◌=◌itemId.replace("◌", "◌◌");
>◌◌◌◌◌◌◌◌String◌escapedOwnerId◌=◌ownerId.replace("◌", "◌◌");
>◌◌◌◌◌◌◌◌return◌Hash.getSha256Hash(escapedItemId◌+◌"◌"◌+◌escapedOwnerId);
>◌◌◌◌}

That way `getChallenge` accepts all pairs of non-null string arguments and it is collision-resistant over them.

core/src/main/java/bisq/core/filter/FilterManager.java
stejbac: Yes, sorry I missed that.

proto/src/main/proto/pb.proto
stejbac: OK, I'll change that.

common/src/main/java/bisq/common/crypto/EquihashProofOfWorkService.java
stejbac: Oops, `("◌", "")` & `("", "◌")` map to the same hash. Would need to use a second character, e.g. ",", like:

> return◌Hash.getSha256Hash(escapedItemId◌+◌",◌"◌+◌escapedOwnerId);

chimp1984:
@stejbac Is there a commit outstanding or is the PR ready for merge?

proto/src/main/proto/pb.proto
stejbac: I was attempting to base it loosely on the Bitcoin difficulty, which is floating point. E.g. in the RPC source code of Bitcoin Core, https://github.com/bitcoin/bitcoin/blob/master/src/rpc/blockchain.cpp, they internally use a double to represent difficulty.

common/src/main/java/bisq/common/crypto/EquihashProofOfWorkService.java
stejbac: I've just added a commit (db6025a) to address this. It adds a `ProofOfWork.solution` field to hold the Equihash puzzle solution bytes.

chimp1984 (approved):
utACK

proto/src/main/proto/pb.proto
stejbac: (I have just left this as it is for now.)

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/5858#pullrequestreview-824737436

PR #5859, opened 2021-11-23
Add date for applying filter

chimp1984:
We had historically higher trade limits and assets which are not in the
currency list anymore, so we apply the filter only for data after
Nov 1st 2021.


ripcurlx (approved):
utACK

PR #5864, opened 2021-11-24
Update package.gradle: rely on local JDK 15

cd2357:
Enforce the use of JDK 15 when packaging. This removes the need to download and unpack JDK 15 just for the use of jpackager.

Depends on bisq-network/bisq#5836


ripcurlx (approved):
ACK - tested it with a new notarized build on macOS. Everything works like before 👍

PR #5870, opened 2021-11-27
Improve user experience once mediation has been accepted by both parties

### LINKED ISSUE #5546: Improve user experience once mediation has been accepted by both parties
### Description

When mediation for a given trade is accepted by both parties the trade moves from 'open trades' to 'history' and the payout transaction ID is broadcast.

This is how it should be but the user experience could be achieved by keeping the trade in 'open trades' for longer and give the user the option to withdraw funds to Bisq wallet or external wallet. This would add an additional step that is the same as what happens currently when trades in 'open trades' are successfully completed ie, the user is notified and then they make the choice to keep funds in Bisq wallet or withdraw them. This would improve the user experience consistency.

#### Version

v 1.6.5

### Steps to reproduce

**From the perspective of 'Trader 1 (initial trader to accept mediation)'**

A trade goes to mediation.

1. Trader 1 accepts mediation proposal.
2. Trader 2 accepts mediation proposal.

Trader 1 is not informed mediation has been accepted, there is no notification. Next time they log in the trade will have moved from open trades to history and the broadcast payout tx id will hopefully have been paid out. Trader 1 will need to check funds > transactions to see the status of the trade.

**From the perspective of 'Trader 2 (second trader to accept mediation)'**

A trade goes to mediation.

1. Trader 1 accepts mediation proposal.
2. Trader 2 accepts mediation proposal.

Trader 2 is not informed mediation has been accepted, there is no notification. The trade will have moved from open trades to history and the payout tx id will be broadcast.

### Expected behaviour

**From the perspective of 'Trader 1 (initial trader to accept mediation)'**

A trade goes to mediation.

1. Trader 1 accepts mediation proposal.
2. Trader 2 accepts mediation proposal.

Trader 1 is not informed mediation has been accepted, there is a notification. The mediated trade will still be in open trades and the Trader 1 will have the option to keep the funds in their Bisq wallet or withdraw to an external wallet. The trade will then move to history.

**From the perspective of 'Trader 2 (second trader to accept mediation)'**

A trade goes to mediation.

1. Trader 1 accepts mediation proposal.
2. Trader 2 accepts mediation proposal.

Trader 2 is informed mediation has been accepted, by a notification. The trade will still be in open trades and Trader 2 will have the option to keep the funds in their Bisq wallet or withdraw to an external wallet.  The trade will then move to history.

ghost:
Closes #5546

&nbsp;
Transitions to Trade Step 4 (previously it would close/remove the trade).


ghost:
### Implement suggestion from @leo816 here: https://github.com/bisq-network/bisq/discussions/5410


image

---


image

ripcurlx:
I agree that the withdraw to external wallet convenience option might be confusing for new users. But as this is a bigger change in the trade flow, I'd appreciate an ACK from @chimp1984 or @sqrrm as well before merging.

ripcurlx:
@jmacxx I think we should also rename the wording of those two popups

Before
Bildschirmfoto 2022-01-11 um 10 42 04
After
Bildschirmfoto 2022-01-11 um 10 51 21
Before
Bildschirmfoto 2022-01-11 um 10 42 23
After
Bildschirmfoto 2022-01-11 um 10 51 32

I'll push this changes from my local branch.

ripcurlx (dismissed):
ACK - Tested it on Regtest and pushed some minor improvements

bisq-github-admin-3:
I'm still merging this PR although as Codacy has permission issues (again)

pazza83:
@jmacxx thanks for this. Could something like this be done for arbitration?

I think currently when arbitration is closed the trade just disappears? I might be wrong though.

ghost:
This PR had the same effect for both mediated & arbitrated trades, they wait until the user presses close.

PR #5878, opened 2021-11-29
Bugfix: dispute ticket reopened by mailbox message

ghost:
Some dispute tickets would show as open on the client side after being closed by dispute agent.

This can happen when

- dispute agent writes a chat message before closing the ticket,
- dispute client is offline (so will receive mailbox notification at startup),
- dispute client receives from mailbox the DisputeResult before the chat message.

The DisputeResult message is processed, closing the ticket.  Then the (earlier) chat mailbox message is received, causing the ticket to be reopened.

The solution, when checking if a chat message reopens the ticket, is to sort them by timestamp.


ripcurlx (approved):
utACK

PR #5884, opened 2021-12-01
Substitute Unconfirmed for Processing BSQ swap

MwithM:
*Unconfirmed* BSQ swap seems like something failed, while *Processing* is used by some wallets for unconfirmed transactions and has no negative implications.

Bildschirmfoto 2021-12-01 um 11 10 50
Bildschirmfoto 2021-12-01 um 11 10 58

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -576,7 +576,7 @@ takeOffer.success.info=You can see the status of your trade at \"Portfolio/Open 
 takeOffer.error.message=An error occurred when taking the offer.\n\n{0}
 
 takeOffer.bsqSwap.success.headline=Your BSQ swap trade is completed
-takeOffer.bsqSwap.success.info=You can see your completed trade at \"Portfolio/Unconfirmed BSQ Swaps\"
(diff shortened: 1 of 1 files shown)


ripcurlx:
@chimp1984 WDYT ☝️

chimp1984:
Hm... I never saw processing being used in a BTC wallet and it comes with me with the wrong meaning that someone (company) is processing it. I think confirmed/unconfirmed is for Bitcoiners a standard term. But no super strong opinion either.

MwithM:
"Pending" would not have third party implications and it's used at ETH/RSK wallets.
Anyway, it's probably better to leave it as Unconfirmed because this is the specific, technical term and wait if it's really confusing once implemented. Probably we will get used to it.

ripcurlx:
I think if we keep the naming we might extend the `Your BSQ swap trade is completed` with:

`Until your trade is included in a block you can see your completed trade at "Portfolio/Unconfirmed BSQ swaps"`

chimp1984:
> "Pending" would not have third party implications and it's used at ETH/RSK wallets.

For ETH: Yes they depend usually heavily on 3rd parties (e.g. Infura), so "processing" in their context is more adequate IMO.

bisq-github-admin-3 (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/5884#issuecomment-985325925

PR #5888, opened 2021-12-02
Fix NPEs in Portfolio View wrt. BSQ swap

ghost:
1. Alice creates a BSQ swap offer
2. Bob takes the offer
3. While the trade is unconfirmed, Bob or Alice closes Bisq and then re-starts.
4. Go to Portfolio -> Unconfirmed BSQ Swap
5. Type into the Filter input box.
6. An NPE occurs.
7. Confirm the trade.
8. Go to Portfolio -> History
9. Type into the Filter input box.
10. An NPE occurs.

Screenshot from 2021-12-02 12-52-15
 
Screenshot from 2021-12-02 12-53-13

Screenshot from 2021-12-02 13-07-48


ripcurlx (approved):
ACK - Tested on Regtest

PR #5889, opened 2021-12-02
Fix BSQ swap buyer tx fee theft vulnerability

stejbac:
Prevent the seller from stealing the combined tx fee as change by lying about the value of one or more of his BTC inputs, which are passed to the buyer as raw inputs in the `BsqSwapFinalizeTxRequest` message.

To this end, add a `RawTransactionInput::validate` method to check the `value` field against the output value of the respective spending tx and run it on every raw seller input in `ProcessBsqSwapFinalizeTxRequest`, so that the buyer is no longer just trusting those numbers.

Additionally, check that the spending txIds from the raw BTC inputs supplied by the seller actually match those of his signed inputs in the accompanying partially signed tx, thus tying the raw input values to the seller's tx.

--

This should block an attack by the BTC seller that I was able to carry out on regtest. If the seller is the taker, the attack can be made worse by creating a swap trade with the tx fee set at just under twice that seen by the maker, so that it is still within tolerance. So the seller-as-taker could wait until the tx fee spikes to, say, 50 sat/vbyte (but the mempool is not too full) and then take an offer with the tx fee set to 99 sat/vbyte. In this way, they could steal around 12000 sat or so as change and still leave 1 sat/vbyte for the fee so that the transaction eventually confirms. The attack is made worse the more inputs the maker uses or the higher the broadcast tx fee.

### DIFF
--- a/core/src/main/java/bisq/core/btc/model/RawTransactionInput.java
+++ b/core/src/main/java/bisq/core/btc/model/RawTransactionInput.java
@@ -144,6 +144,11 @@ public final class RawTransactionInput implements NetworkPayload, PersistablePay
         long outputValue = tx.getOutput(index).getValue().value;
         checkArgument(value == outputValue,
                 "Input value (%s) mismatches connected tx output value (%s).", value, outputValue);
+        var scriptPubKey = tx.getOutput(index).getScriptPubKey();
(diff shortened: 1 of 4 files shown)


chimp1984 (approved):
utACK

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/5889#pullrequestreview-822880143

PR #5894, opened 2021-12-04
Fix premature validation call

chimp1984:
When syncing from genesis the number of blocks are limited so we get the
`onParseBlockCompleteAfterBatchProcessing` called each time when the received
 blocks are processed, and as we are not at wallet height we repeat requesting
 blocks. But the new check for the BTC recipient triggers a resync from resource call.
We add now a check that we do this check only once the wallet is synced and our
block height from dao state matches wallet blockheight.

Fixes https://github.com/bisq-network/bisq/issues/5882

PR #5898, opened 2021-12-05
Fix issue with resource files not loaded on Ubuntu Linux

ghost:
On Ubuntu Linux the attempt to enumerate DAO block files in a JAR resource directory fails.

See https://github.com/bisq-network/bisq/pull/5894#issuecomment-986164443

All the block data is then downloaded from the seednodes.

This PR does the enumeration using `nio.File.FileSystem` so that block files in resources can be used.

### DIFF
--- a/core/src/main/java/bisq/core/dao/state/storage/BsqBlocksStorageService.java
+++ b/core/src/main/java/bisq/core/dao/state/storage/BsqBlocksStorageService.java
@@ -21,6 +21,7 @@ import bisq.core.dao.state.GenesisTxInfo;
 import bisq.core.dao.state.model.blockchain.Block;
 
 import bisq.common.config.Config;
+import bisq.common.file.FileUtil;
 import bisq.common.proto.persistable.PersistenceProtoResolver;
 
 import protobuf.BaseBlock;
@@ -33,8 +34,14 @@ import org.apache.commons.io.FileUtils;
 
 import java.net.URL;
 
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
 import java.io.File;
 
+import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


core/src/main/java/bisq/core/dao/state/storage/BsqBlocksStorageService.java
```diff
@@ -121,22 +128,21 @@ void copyFromResources(String postFix) {
...
                 return;
             }
-            File dir = new File(dirUrl.toURI());
-            String[] fileNames = dir.list();
-            if (fileNames == null) {
-                log.info("No files in directory. {}", dir.getAbsolutePath());
+            FileSystem fileSystem = FileSystems.newFileSystem(dirUrl.toURI(), Collections.emptyMap());
+            List<Path> filePaths = Files.walk(fileSystem.getPath(resourceDir), 1).collect(Collectors.toList());
+            if (filePaths.size() <= 1) {
+                log.info("No files in directory. {}", dirUrl.toString());
```
ripcurlx, line 134: toString is not necessary.

```suggestion
                log.info("No files in directory. {}", dirUrl);
```

ripcurlx:
Doesn't work now on macOS unfortunately
```
java.lang.IllegalArgumentException: Path component should be '/'
	at java.base/sun.nio.fs.UnixFileSystemProvider.checkUri(UnixFileSystemProvider.java:81)
	at java.base/sun.nio.fs.UnixFileSystemProvider.newFileSystem(UnixFileSystemProvider.java:90)
	at java.base/java.nio.file.FileSystems.newFileSystem(FileSystems.java:338)
	at java.base/java.nio.file.FileSystems.newFileSystem(FileSystems.java:287)
	at bisq.core.dao.state.storage.BsqBlocksStorageService.copyFromResources(BsqBlocksStorageService.java:131)
	at bisq.core.dao.state.storage.DaoStateStorageService.lambda$readFromResources$4(DaoStateStorageService.java:119)
	at java.base/java.lang.Thread.run(Thread.java:832)
```

ripcurlx (changes requested):
NACK - https://github.com/bisq-network/bisq/pull/5898#issuecomment-986659019

stejbac:
I was debugging the same issue on Windows a couple of days ago, caused by the expression `new File(dirUrl.toUri())`. I ended up adapting the following method from StackOverflow, which seems to work:

https://stackoverflow.com/questions/3923129/get-a-list-of-resources-from-classpath-directory/48190582#48190582

It looks like you have to use two different code paths to list the resource directory, depending on the environment.

(I created a [draft patch for the fix](https://github.com/stejbac/bisq/commit/c8d0c49545bf0688e5789bec946102037c0151f6), but never got round to creating a PR out of it. The foreach loop to do the copying needs modifying to work on Windows as well, just like in this PR.)

ghost:
@stejbac Looks like we were both debugging the same issue.  As this one does not work for MacOs, can you make your PR?

stejbac:
@jmacxx OK, sure.

chimp1984:
@stejbac Ah great that you tackled that as well. Could we extract the handling in a FileUtil class to make it re-usable. A pity that simple filehandling from resources is such a mess in java...
We have verious other loadFromResource calls. Maybe worth to re-check if there are issues as well.

PR #5900, opened 2021-12-07
Fix resync BSQ blocks from resources

stejbac:
Prevent a _"URI is not hierarchical"_ `IllegalArgumentException` from the expression, `new File(dirUrl.toURI())`, which occurs on Linux & Windows when listing the resource directory of BSQ blocks. Adapt a solution from StackOverflow (together with the earlier `java.nio.file.FileSystem`-based approach in #5898), which uses two separate code paths depending on the environment, into the new method `FileUtil::listResourceDirectory`.

The issue is caused by the resource URL taking one of the two forms:

1.  file:/Users/[USER]/Java/bisq/bisq/p2p/out/production/resources/BsqBlocks_BTC_MAINNET
2.  jar:file:...p2p.jar!/BsqBlocks_BTC_MAINNET

depending on whether the system is OSX or not.

--

I tested the PR on Linux & Windows, but have not been able to test the absence of a regression on OSX (though I don't expect this, as it should use a similar code path to the original).

### DIFF
--- a/common/src/main/java/bisq/common/file/FileUtil.java
+++ b/common/src/main/java/bisq/common/file/FileUtil.java
@@ -24,6 +24,11 @@ import com.google.common.io.Files;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.net.URLDecoder;
+
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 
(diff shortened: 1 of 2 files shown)


stejbac:
I think the Codacy check is going to fail, due to a missed `//NOPMD` suppression on a line containing an unavoidable fully qualified name.

Amending and force-pushing the last commit to fix.

ghost (approved):
ACK for Ubuntu Linux.  :+1:

ripcurlx (approved):
ACK from macOS Monterey. 👍

PR #5901, opened 2021-12-07
Add Exception handling at RpcService

chimp1984:
At shutdown the executor can throw a RejectedExecutionException which is expected as we shut down the executor immediately.
If not thrown at shutdown or exception is not a RejectedExecutionException
we log a warning and re-throw the exception (to avoid change of behaviour
of current version). The exception is likely not handled by callers and goes up to
the uncaught exception handler.


ripcurlx (approved):
ACK - Tested it on Mainnet

bisq-github-admin-3:
Still merging this as the Codacy complaint isn't correct in this scenario.

Issue #4328, opened 2020-06-22
Bisq does not detect when the Tor network is rate limiting Bisq P2P connections


wiz:
### Description

Tor can rate limit Bisq for various reasons, but Bisq does not handle this case or indicate it to the user in any way

#### Version

v1.3.5

### Steps to reproduce

1) Start Bisq with max connections set to 100
2) Enable Tor debug logging
3) Wait for Tor to rate limit Bisq and Tor log debug prints to indicate this

### Expected behaviour

After step 3, Bisq should have a ⚠️ icon in the status bar, and when you mousehover over the icon it should indicate to the user that "Tor Network is rate-limiting your Bisq node"

### Actual behaviour

Bisq does not detect when the Tor network is rate limiting Bisq P2P connections or indicate it to the user in any way

#### Additional info
```
Jun-22 19:57:07.189 [TorControlParser] DEBUG o.b.n.tor.Tor: hiddenService: HS_DESC FAILED epfaacm6s4cf4xvx NO_AUTH $A1D8BE264B369180006CDA03998879B49CFAE93B~reaigauebnap uhhz3uvcbp46dsx5wwkqi272bjm3nufu REASON=NOT_FOUND 
Jun-22 19:57:07.189 [TorControlParser] DEBUG o.b.n.tor.Tor: hiddenService: HS_DESC_CONTENT uhhz3uvcbp46dsx5wwkqi272bjm3nufu NO_DESCRIPTOR, as in epfaacm6s4cf4xvx uhhz3uvcbp46dsx5wwkqi272bjm3nufu $A1D8BE264B369180006CDA03998879B49CFAE93B~reaigauebnap 
Jun-22 19:57:07.189 [TorControlParser] DEBUG o.b.n.tor.Tor: hiddenService: HS_DESC FAILED epfaacm6s4cf4xvx NO_AUTH UNKNOWN REASON=QUERY_RATE_LIMITED 
Jun-22 19:57:07.189 [TorControlParser] DEBUG o.b.n.tor.Tor: hiddenService: HS_DESC_CONTENT usqzva7jkdx5c5fl57ic4qm6pwib26u4 NO_DESCRIPTOR, as in epfaacm6s4cf4xvx usqzva7jkdx5c5fl57ic4qm6pwib26u4 UNKNOWN 
[...]

PR #5903, opened 2021-12-08
Clean tor directory at startup

chimp1984:
This PR is an intent to fight the messaging issues. We had in the past sometimes issues that Tor became instable and deleting the tor directory (excluding the hidden service dir) solved that. So it might be worth to see if it helps if we reset at each startup the tor directory. 
The startup time for Tor is a bit slower as it need to download about 10MB of data. 

Tests showed that it takes about 8 seconds longer. On a slow internet connection its about  15-25 sec slower.

I dont have a very strong opinion if we should use that PR as the chances that it fixes the problem are not super high, but its a relatively cheap attempt which comes with only a small delay at startup. Give a thumbs up/down if you support that PR.
I researched also to find if there are reported bugs with corrupted/outdated tor data but could not find much beside that at shutdown the decriptor files might get corrupted, but I assume that would lead to problems as starting Tor.

Normal:
```
bisq.log:Tor started after 6736 ms. Start publishing hidden service.
bisq.log:Tor started after 6182 ms. Start publishing hidden service.
bisq.log:Tor started after 5802 ms. Start publishing hidden service.
bisq.log:Tor started after 6286 ms. Start publishing hidden service.
bisq.log:Tor started after 6303 ms. Start publishing hidden service.
```

With that PR:
```
bisq.log:Tor started after 14819 ms. Start publishing hidden service.
bisq.log:Tor started after 13833 ms. Start publishing hidden service.
bisq.log:Tor started after 12840 ms. Start publishing hidden service.
bisq.log:Tor started after 13760 ms. Start publishing hidden service.
```

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqSetup.java
+++ b/core/src/main/java/bisq/core/app/BisqSetup.java
@@ -147,6 +147,7 @@ public class BisqSetup {
     private final MediationManager mediationManager;
     private final RefundManager refundManager;
     private final ArbitrationManager arbitrationManager;
+    private final TorSetup torSetup;
 
     @Setter
     @Nullable
@@ -234,7 +235,8 @@ public class BisqSetup {
                      Socks5ProxyProvider socks5ProxyProvider,
                      MediationManager mediationManager,
                      RefundManager refundManager,
-                     ArbitrationManager arbitrationManager) {
+                     ArbitrationManager arbitrationManager,
+                     TorSetup torSetup) {
         this.domainInitialisation = domainInitialisation;
         this.p2PNetworkSetup = p2PNetworkSetup;
         this.walletAppSetup = walletAppSetup;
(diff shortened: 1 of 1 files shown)


wiz:
Hmm, I am not sure this is a good idea as there are tradeoffs to configure - for example, it might cause people to have better Tor connectivity in the short-term but it might cause users to get rate-limited from Tor more often (i.e. banned) in the long-run. Are we detecting when Bisq gets rate limited by Tor now? Originally reported in #4328

wiz:
As a comparison, Tor Browser has the following buttons in the browser:
Screen Shot 2021-12-11 at 14 42 13
Perhaps we could expose these buttons to allow users to manually clear Tor files when necessary, and also clear Tor files when certain states are detected (such as rate-limiting), or at least less frequently and more time based such as once per day or once per week, but not always on every startup?

alkum:
I tend to agree with @wiz, the tradeoff doesn't seem worth it.

How about having an optional startup flag that could trigger this, if necessary?

chimp1984:
Yes I also agree to better not touch that.
The Tor settings have already the option to clean the tor dir. But it is not easy to find and not available at startup, so that is a UI issue which could relatively easy be solved. I close that for now but any dev is welcome to pick that idea up from here.

PR #5904, opened 2021-12-08
Increase trade protocol timeout from 60 sec to 120 sec

chimp1984:
Another attempt to deal with the messaging issues....

At one investigated case we saw that a timeout got triggered followed by `Hidden Service xxx.onion has been announced to the Tor network.` log from the Netwlay lib which gets logged if the hidden service gets an `UPLOADED` event, which usually happen at startup. So it might be Tor related. Unfortunately the low level tor logs are not eabled by default so we cannot see more details.


ripcurlx (approved):
utACK

PR #5905, opened 2021-12-08
Improve shutdown process

chimp1984:
I observed several times that the app shutdown got blocked for about 1 minute. I could not find out what blocks it, but I suspect its in the network/tor layer. I added more logs to make it easier to spot the issue and improved the shutdown process as well as reduced the timeouts. There was also a incorrect handling of the tor shutdown process which got fixed.


ripcurlx (approved):
utACK

PR #5907, opened 2021-12-08
OfferBook: Show Complete Text Of Action Button If Enough Space Available

### LINKED ISSUE #5862: Adapt column resizing policy in offer book
### Description

Launching the app in another language (e.g. German) reveals a problem with the column resizing in the offerbook

#### Version

v1.7.5-

### Steps to reproduce

- Switch to German
- Open offer book

### Expected behaviour

Column should resize properly if enough space is available so that no text is truncated.

### Actual behaviour

Text stays truncated

### Screenshots

Bildschirmfoto 2021-11-24 um 11 07 08

#### Device or machine

Not OS related

alvasw:
Even if the app is maximized it does not show the full text of the
action button if the text is long. Switching Bisq to German and
maximizing the window reveals the problem.

Fixes #5862


ripcurlx (approved):
ACK - Tested it on Mainnet

Bildschirmfoto 2021-12-09 um 10 18 31

Congrats to your first merged PR in Bisq!

PR #5909, opened 2021-12-09
Dont log error if BSQ blocks directory does not exist in resources

chimp1984:
In case the BSQ blocks directory does not exist in resources (as in case of REG_TEST)

we do not want to log an error stacktrace but just an info message.

Use `File.separator` instead of "/"

Your PR description here.

### DIFF
--- a/core/src/main/java/bisq/core/dao/state/storage/BsqBlocksStorageService.java
+++ b/core/src/main/java/bisq/core/dao/state/storage/BsqBlocksStorageService.java
@@ -22,6 +22,7 @@ import bisq.core.dao.state.model.blockchain.Block;
 
 import bisq.common.config.Config;
 import bisq.common.file.FileUtil;
+import bisq.common.file.ResourceNotFoundException;
 import bisq.common.proto.persistable.PersistenceProtoResolver;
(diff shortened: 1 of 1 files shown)


core/src/main/java/bisq/core/dao/state/storage/BsqBlocksStorageService.java
```diff
@@ -123,9 +123,11 @@ void copyFromResources(String postFix) {
             }
             for (String fileName : fileNames) {
                 File destinationFile = new File(storageDir, fileName);
-                FileUtil.resourceToFile(resourceDir + "/" + fileName, destinationFile);
+                FileUtil.resourceToFile(resourceDir + File.separator + fileName, destinationFile);
```
stejbac: This doesn't appear to work on Windows. It has to be "/", not "\\".

chimp1984: Ah really? Weird... Will revert that. Javas resource handling is really a pain.

ripcurlx (approved):
utACK

PR #5914, opened 2021-12-13
Fix fat finger protection in Offer Entry screen

ghost:
Normally, for currencies that have a market price feed, Bisq prevents the user from entering offers that are more than 50% away from spot.  It recently came to light that this fat finger protection in Bisq has a flaw.

A certain sequence of focus operations in the enter offer screen causes fat finger protection to turn off, and then the user is no longer protected from making offers that are significantly out of the market.

**To reproduce the issue:**

1. Go to BUY (or SELL) BTC for an market priced asset, e.g. ETH
2. Click Create New Offer
3. Enter 0.25 BTC
4. Tab to the next field.
5. Click the up/down arrow icon to make sure the fixed price edit box is selected at the top.  The price and deviation gets auto-populated to market price and 0% deviation.
6. Enter a fixed price that is 10x higher than market value.
7. Click "Next" and the fat finger protection warning DOES NOT pop up.
8. This is because the deviation still shows 0%.
 

**Video example:**

fat_finger

---

_The two lines of code removed by this PR seem to be a relic from < 2018 so its hard to locate the original intent.  Regardless, it will require a good regression test to check that this fix does not affect any other path in the offer UI._

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/offer/bisq_v1/MutableOfferViewModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/bisq_v1/MutableOfferViewModel.java
@@ -812,6 +812,7 @@ public abstract class MutableOfferViewModel<M extends MutableOfferDataModel> ext
     }
 
     void onFixPriceToggleChange(boolean fixedPriceSelected) {
+        inputIsMarketBasedPrice = !fixedPriceSelected;
         updateButtonDisableState();
         if (!fixedPriceSelected) {
             onTriggerPriceTextFieldChanged();


ripcurlx:
@jmacxx Can you re-produce this all the time? I tried to, but failed so far (see recording)


https://user-images.githubusercontent.com/170962/146765004-5c8fcf51-23b6-4505-8211-126dc7af6018.mov

ghost:
It is reproducible consistently, whenever the % field is by default shown at the top when the offer dialog is opened.  If it is not at the top, make it so, then close the dialog and reopen to reproduce the issue.

ripcurlx:
> It is reproducible consistently, whenever the % field is by default shown at the top when the offer dialog is opened. If it is not at the top, make it so, then close the dialog and reopen to reproduce the issue.

Now I was able to re-produce it as well. I've to check why I've added this bit in 2018 🤔

ripcurlx:
Luckily there was a test for it otherwise I wouldn't know it anymore

https://user-images.githubusercontent.com/170962/146929707-c9f25a66-68e5-495e-8c49-e3427b2e36d3.mov

This change would break the behavior above, so we need to solve it somehow else.

ghost:
Here is an alternative fix that does not break the test for https://github.com/bisq-network/bisq/pull/5914#issuecomment-998737756

ripcurlx (approved):
utACK - code change looks fine to me now 👍

PR #5918, opened 2021-12-14
Add @KanoczTomas btc node with onion v3 address

KanoczTomas:
Fixes #5714

Added TOR hidden service v3 address to btc node.


ripcurlx:
@KanoczTomas There is an error in your code: https://github.com/bisq-network/bisq/runs/4527362870?check_suite_focus=true#step:6:69

ripcurlx (approved):
ACK - checked both hidden service and IP access.

PR #5928, opened 2021-12-20
Upgrade log4j 2.15.0 => 2.17.0

cbeams:
This change upgrades log4j to patch fixes for recently documented
CVE-2021-45046 CVE-2021-45105 vulnerabilities related to the Log4Shell
exploit.

Like the earlier fix, Bisq does not appear to be vulnerable to these
exploits because it does not use log4j directly, only transitively
depends on it. Nevertheless, the upgrade is still the safe bet.


wiz (approved):
Tested ACK @ 07a139c927f3c4c6f57f5567b536cbaa9a89a44e
Running OK in production on wizpriceje6q5tdrxkyiazsgu7irquiqjy2dptezqhrtu7l2qelqktid

ripcurlx (approved):
utACK

PR #5930, opened 2021-12-20
SEPA Account - editable account name and accepted countries

### LINKED ISSUE #5692: Not possible to modify accepted countries
### Description

Hey guys👋 I’ve noticed that once I create a payment method (bank transfer) I can’t then modify the list of accepted countries other than by deleting it.  

#### Version

 1.7.3

### Steps to reproduce

Create a payment method which involves bank transfers (like SEPA)

### Idea

Of course there are some important info defining the details of the payment method that should not be changed in order to prevent the workaround of the “account signing” functionality. Indeed, this avoids that a user could wait for the signing of a payment method and then change its info.

However, I think that the the modification of the “accepted countries” field does not pose a risk in this sense. Thus, what about adding an option to modify this field so that there’s not need delete and recreate the payment method?

This would allow users to change their preferences in case they want to limit/change countries from which they can receive payments.

Hope you find this idea helpful. Please let me know what you think about it

ghost:
Video:
update-account

Fixes #5692

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/account/content/fiataccounts/FiatAccountsDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/account/content/fiataccounts/FiatAccountsDataModel.java
@@ -124,6 +124,10 @@ class FiatAccountsDataModel extends ActivatableDataModel {
         accountAgeWitnessService.signAndPublishSameNameAccounts();
     }
 
+    public void onUpdateAccount(PaymentAccount paymentAccount) {
+        user.requestPersistence();
+    }
+
     public boolean onDeleteAccount(PaymentAccount paymentAccount) {
         boolean usedInOpenOffers = openOfferManager.getObservableList().stream()
                 .anyMatch(openOffer -> openOffer.getOffer().getMakerPaymentAccountId().equals(paymentAccount.getId()));
--- a/desktop/src/main/java/bisq/desktop/main/account/content/fiataccounts/FiatAccountsDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/account/content/fiataccounts/FiatAccountsDataModel.java
@@ -125,6 +125,7 @@ class FiatAccountsDataModel extends ActivatableDataModel {
     }
 
     public void onUpdateAccount(PaymentAccount paymentAccount) {
+        paymentAccount.onPersistChanges();
         user.requestPersistence();
     }
 
--- a/core/src/main/java/bisq/core/payment/PaymentAccount.java
+++ b/core/src/main/java/bisq/core/payment/PaymentAccount.java
@@ -58,6 +58,9 @@ public abstract class PaymentAccount implements PersistablePayload {
     public PaymentAccountPayload paymentAccountPayload;
     @Setter
     protected String accountName;
+    @Setter
+    protected String persistedAccountName;
+
     protected final List<TradeCurrency> tradeCurrencies = new ArrayList<>();
     @Setter
     @Nullable
@@ -117,6 +120,7 @@ public abstract class PaymentAccount implements PersistablePayload {
             account.setId(proto.getId());
             account.setCreationDate(proto.getCreationDate());
             account.setAccountName(proto.getAccountName());
+            account.setPersistedAccountName(proto.getAccountName());
             account.getTradeCurrencies().addAll(tradeCurrencies);
             account.setPaymentAccountPayload(coreProtoResolver.fromProto(proto.getPaymentAccountPayload()));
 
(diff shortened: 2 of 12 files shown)


desktop/src/main/java/bisq/desktop/main/account/content/fiataccounts/FiatAccountsDataModel.java
```diff
@@ -124,6 +125,14 @@ public void onSaveNewAccount(PaymentAccount paymentAccount) {
         accountAgeWitnessService.signAndPublishSameNameAccounts();
     }
 
+    public void onUpdateAccount(PaymentAccount paymentAccount) {
+        user.requestPersistence();
+        SepaAccount account = (SepaAccount) paymentAccount;
+        account.getAcceptedCountryCodes();
+        log.info("ON UPDATE, ACCOUNT: " + account.getAcceptedCountryCodes());
+        // TODO: persist changes
```
ripcurlx: Is this PR already ready for review?

ripcurlx:
@xyzmaker123 Did you try out what happens when you do this on an already signed account? I fear that this will invalidate the account signing state as it changes the hash as well.

desktop/src/main/java/bisq/desktop/main/account/content/fiataccounts/FiatAccountsDataModel.java
ghost: Sorry, now it's okay. This TODO was just leftover.

ghost:
@ripcurlx I didn't tried it on an already signed account - I don't know how to invoke this scenario on localnet dev setup. IIUC acceptedCountries are not used for hash calculation - only [iban, bic](https://github.com/bisq-network/bisq/blob/087a780d024d68814a994e14390d9a87b0cee7aa/core/src/main/java/bisq/core/payment/payload/SepaAccountPayload.java#L162-L167), [countryCode](https://github.com/bisq-network/bisq/blob/cbb59c155b688cf51433b5f4070f70aa3d3fb24f/core/src/main/java/bisq/core/payment/payload/CountryBasedPaymentAccountPayload.java#L69-L72) and [paymentMethodId](https://github.com/bisq-network/bisq/blob/f02772e3a22e9f4c0a2a7cc91c6376e48f2437cc/core/src/main/java/bisq/core/payment/payload/PaymentAccountPayload.java#L135-L137)

ghost:
Edge case scenario I tested:
1) Bob has single SEPA account, and accept trades from AT and DE
2) Alice has single SEPA account, and "country of bank set" to DE
3) Bob created offer to buy BTC with euro
4) Alice accepted this offer
5) Bob removed DE from list of accepted countries
6) Alice and Bob were able to finish transaction
Accepted countries are checked in time of taking offer, there is no check later

pazza83:
Hi @xyzmaker123 if this is possible it would be great. SEPA countries have changed over the last 12 months (more have been added) so it would be good for accounts set up prior to this to be able to add more countries they are happy to trade with.

I would also propose they should also be able to edit the 'account name' as essentially this is just the nickname given to the account and only used by the users local Bisq instance.

ghost:
@pazza83 This pull request support this case. These countries are unchecked when someone open edit account, and it's possible to check them.

Sure - I'll extend this pull request with possibility to edit the 'account name' (CC: @ripcurlx)

ripcurlx:
> Hi @xyzmaker123 if this is possible it would be great. SEPA countries have changed over the last 12 months (more have been added) so it would be good for accounts set up prior to this to be able to add more countries they are happy to trade with.
> 
> I would also propose they should also be able to edit the 'account name' as essentially this is just the nickname given to the account and only used by the users local Bisq instance.

Just to make sure: In this PR the SEPA countries are not extended/changed:
- getAllSepaEuroCountries: `"AT", "BE", "CY", "DE", "EE", "FI", "FR", "GR", "IE","IT", "LV", "LT", "LU", "MC", "MT", "NL", "PT", "SK", "SI", "ES", "AD", "SM", "VA"`
- getAllSepaNonEuroCountries:`"BG", "HR", "CZ", "DK", "GB", "HU", "PL", "RO","SE", "IS", "NO", "LI", "CH", "JE"`

ripcurlx (approved):
ACK - Tested it on Regtest.
- with signed account, where countries were changed --> account still signed
- created offer with account, changed countries afterwards, created another offer --> both offers/trades are valid as countries are part of the offer and won't be changed at a later stage

bisq-github-admin-3:
I'm still merging this PR although as Codacy has permission issues (again)

Bl5ckj5ck:
Great job @xyzmaker123 ! Thanks

pazza83:
Hi @xyzmaker123 thanks for this. Does this also work for SEPA Instant? Apologies just thought of this now :)

ghost:
@pazza83 Yes - I also included SEPA Instant accounts during implementation

PR #5943, opened 2021-12-30
CI Security: Use Github actions via sha1 (not tags) and keep them updated via dependabot

dbast:
Tags are mutable and can change unexpectedly. Referencing actions via sha1 is more secure in that regard. Dependabot helps to automatically update to newer versions without the need to manually deal with sha1s.

Merging the PR opens update PRs similar to those shown here https://github.com/dbast/bisq/pulls 

Each update PR lists the according changes to each action and allows giving instructions to dependabot via e.g. `@dependabot ignore this major version`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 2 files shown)


ripcurlx (approved):
ACK - reviewed the commit hashes used for each action. It is true that Github seems to update their major version action tags all the time.

PR #5826, opened 2021-11-14
Improve fee handling

chimp1984:

PR #5949, opened 2022-01-03
Fix trade fee validation bug

ghost:
A support issue identified a bug in the trade fee validation when two BSQ UTXOs are burned to pay the fee.

- This PR addresses the issue by checking if the expected fee is larger than the first UTXO value.
- A new test has been added to validate this.
- Other tests have been updated as a result of the changes in https://github.com/bisq-network/bisq/pull/5826 where fee parameter is now supplied from a filter; this results in more accurate fee checking.


ripcurlx (approved):
utACK - code changes are looking fine

PR #5951, opened 2022-01-05
Rename API method keepfunds -> closetrade (#1)

ghubstan:
Trade proceeds and deposits have already been transferred to Bisq wallets before the `keepfunds` command is (was) executed;  `keepfunds` merely moves open trade to closed trade lists and persistence files.  Renaming `keepfunds` as `closetrade` makes its purpose clear to API users.

The commit modifies only method names and comments in api server+cli classes, apitest cases, and api trade simulation scripts.

Based on `master`


ripcurlx (approved):
utACK

PR #5960, opened 2022-01-07
Enable fractional percentages in dispute payout calculation field

ghost:
@leo816 requested the ability to enter fractional percentages in the dispute payout calculation window.
Either a comma (1,23) or a period (1.23) may be used as the fractional delimiter.

image

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/overlays/windows/DisputeSummaryWindow.java
+++ b/desktop/src/main/java/bisq/desktop/main/overlays/windows/DisputeSummaryWindow.java
@@ -963,7 +963,7 @@ public class DisputeSummaryWindow extends Overlay<DisputeSummaryWindow> {
 
         Coin penalizedPortionOfTradeAmount = Coin.ZERO;
         try {
-            disputeResult.setPayoutAdjustmentPercent(compensationOrPenalty.getText().replaceAll("[^0-9]", ""));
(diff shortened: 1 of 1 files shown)


desktop/src/main/java/bisq/desktop/main/overlays/windows/DisputeSummaryWindow.java
```diff
@@ -963,7 +963,7 @@ private void applyUiControlsToDisputeResult(Toggle selectedTradeAmountToggle) {
 
         Coin penalizedPortionOfTradeAmount = Coin.ZERO;
         try {
-            disputeResult.setPayoutAdjustmentPercent(compensationOrPenalty.getText().replaceAll("[^0-9]", ""));
+            disputeResult.setPayoutAdjustmentPercent(compensationOrPenalty.getText().replaceAll("[^0-9,.]", ""));
```
ghost, line 966: What about changing type of `payoutAdjustmentPercent` to `double` instead of `String`, and handling this `replaceAll` logic inside `ParsingUtils.parsePercentStringToDouble` method?

ghost: @xyzmaker123 thankyou for the review.  However I sincerely believe in this case that it would be better remaining as is.

ripcurlx (approved):
utACK

PR #5961, opened 2022-01-07
Add API methods 'failtrade', 'unfailtrade' + test case (#2)

ghubstan:
Prerequisite for next PR:  **Add API method `gettrades`.**

The `gettrades` method implementation will show 'open', 'closed', and 'failed' trades.  Users already needed to be able to fail and unfail trades for the same reasons they do in the UI, and `apitest` cases will need to be able to fail and unfail trades to check correct behavior of the `gettrades` method.

Based on https://github.com/bisq-network/bisq/pull/5951


ripcurlx (approved):
utACK

PR #5963, opened 2022-01-12
Add missing filtering on lists

### LINKED ISSUE #5307: Add a filter to the 'Funds' section in Bisq to able searching of; Trade IDs, Tx IDs, BTC amount, etc
### Description

To make it easier to search for; Trade IDs, Tx IDs, BTC amount, etc in the Bisq client their is the possibility to add a search filter in the following sections:

- Portfolio > My Open Offers
- Portfolio > History'
- Support > Mediation
- Support > Arbitration 

This makes it easy to find information quickly.

The ability to search for ; Trade IDs, Tx IDs, BTC amount, etc is missing from the following sections:

- Funds > Receive Funds
- Portfolio > Send Funds
- Support > Locked Funds
- Support > Transactions

If you have lots of transactions this makes it difficult to find information quickly in the funds section.

Therefore, a filter should be added to the 'Funds' section in Bisq

#### Version

v1.5.9

### Steps to reproduce

Go to the Funds section in Bisq and try and search for ; Trade IDs, Tx IDs, BTC amount. It has to be done manually. There is no search option.

### Expected behaviour

Go to the Funds section in Bisq and be able to search for ; Trade IDs, Tx IDs, BTC amount. It has to be done manually. There is no search option.

### Screenshots

portfolio filter - Copy
Example of funds filter

funds filter - Copy - Copy
Example of missing funds filter

ghost:
Fixes #5307

PR introduce possibility to filtering these lists:
- Portfolio -> Open Trades
- Funds -> Receive Funds
- Funds -> Send Funds
- Funds -> Reserved Funds
- Funds -> Locked Funds
- Funds -> Transactions

Code from Funds -> Transactions was also simplified to be more consistent with whole code base.


desktop/src/main/java/bisq/desktop/main/funds/withdrawal/WithdrawalView.java
```diff
@@ -392,6 +407,19 @@ protected void deactivate() {
     // UI handlers
     ///////////////////////////////////////////////////////////////////////////////////////////
 
+    private void applyFilteredListPredicate(String filterString) {
+        filteredList.setPredicate(item -> {
+            if (filterString.isEmpty())
+                return true;
+
+            if (item.getBalance().toString().contains(filterString)) {
```
chimp1984: This does not cover formatted string. E.g. if balance is 0.01 BTC and you type in 0.01 it is not found.
Better create a new string field in the itme (formatter.formatCoin(this.balance) which reflects the displayed string

desktop/src/main/java/bisq/desktop/main/funds/deposit/DepositView.fxml
```diff
@@ -23,12 +23,22 @@
...
+<?import javafx.scene.layout.HBox?>
+<?import bisq.desktop.components.AutoTooltipLabel?>
+<?import bisq.desktop.components.InputTextField?>
 <VBox fx:id="root" fx:controller="bisq.desktop.main.funds.deposit.DepositView"
       spacing="10" xmlns:fx="http://javafx.com/fxml">
     <padding>
         <Insets bottom="15.0" left="15.0" right="15.0" top="15.0"/>
     </padding>
-
+    <HBox spacing="5.0">
```
chimp1984: The usage of FXML is a old left-over from early dev state. We should avoid adding new FXML items and add it in the normal View classes instead. This makes code easier to reuse (e.g. you can make a filter component which is then used every time when you show that filter). Beside better code reusablility and flexibility with more dynamic use cases FXML is slow when getting rendered, thats why we move away from it. We just never spent the time to cleanup the left overs and it would require some more work to remove it completely, so we use it as container still but do all the rest in the view class.

desktop/src/main/java/bisq/desktop/main/funds/locked/LockedView.java
```diff
@@ -219,6 +237,61 @@ protected void deactivate() {
...
+                }
+
+                Contract contract = trade.getContract();
+
+                boolean isBuyerOnion = false;
+                boolean isSellerOnion = false;
+                boolean matchesBuyersPaymentAccountData = false;
+                boolean matchesSellersPaymentAccountData = false;
+                if (contract != null) {
+                    isBuyerOnion = contract.getBuyerNodeAddress().getFullAddress().contains(filterString);
```
chimp1984: Those predicates appear several times. Would be good to make it reusable code

desktop/src/main/java/bisq/desktop/main/funds/transactions/TransactionsView.java
```diff
@@ -251,6 +299,81 @@ protected void deactivate() {
...
+                                } else if (tradable instanceof TradeModel) {
+                                    return new TransactionAwareTrade(
+                                            (TradeModel) tradable,
+                                            arbitrationManager,
+                                            refundManager,
+                                            btcWalletService,
+                                            pubKeyRing
+                                    );
+                                } else {
+                                    return new DummyTransactionAwareTradable(tradable);
```
chimp1984: I know that is from old code, but as that case never gets executed I think we should return null here and then filter out null in the next step. Tradable is only extended from OpenOffer and TradeModel (one other case is not relevant here). The DummyTransactionAwareTradable can be removed then.

chimp1984 (changes requested):

chimp1984:
I have not looked closely to the other filter predicates. But basically it shoud always take the display string as that is what users would expect to be able to filter. Maybe there are some cases where additional domain data filtering is useful as well, but should be additional to the display string case.
As we have not plenty of filters in the app it would be good to have them re-usable also with reusable predicates where it makes sense.

I am not so 100% sure if we should show the filter fields by default or extend it just when the user demand it. E.g. add a filter icon and when the user opens it show the field. That would help to not overload more the already overloaded UI.
We can persist the selection of hte icon, so users who prefer to have the filter get it automatically displayed at next restart.

I know the earlier filter implementations have been done the same way (by myself), but as we extend it now to more screens I think its worth to put a bit more effort into it. 

Beside that great to see that filter added to more screens.

desktop/src/main/java/bisq/desktop/main/funds/deposit/DepositView.fxml
ghost: I agree - I'm just not sure if this pull request is appropriate place for doing it. `DepositView` and others similar already have FXML items declared. It's probably not good idea to mix both concepts (adding filtering `HBox` in java code while keeping `TableView` and `GridPane` in fxml). To keep this PR compact and monothematic I prefer to keep it as it is and prepare complex refactoring in dedicated pull request after merging this one.

However I'm a bit curious about performance issue with fxml rendering. I thought that thanks of [CachingViewLoader](https://github.com/bisq-network/bisq/blob/7f73df5dfd761a26392fb6ae16adc937ef6bf9da/desktop/src/main/java/bisq/desktop/common/view/CachingViewLoader.java) usage views are parsed once while loading and later it doesn't affect performance. If you know any resources that could help me to better understand this problem I'd be thankful.

chimp1984: Ok with me.
Re performance: Yes the CachingViewLoader was introduced to avoid the extra costs each time the view is opened. But there is still the penalty for the first time. If you have a very component heavy UI screen in FXML and then in plain Java code you can see/measure the difference. Parsing FXML does not come for free...

ghost: As a former frontend dev I always prefer to separate UI from logic - from my perspective it makes code easier to read and maintain. You're right it comes with some costs, don't know how big these costs are - just want to be sure we are not falling into the trap of [premature optimization](https://wiki.c2.com/?PrematureOptimization).

If I'll find time I'll try to measure this problem a bit during refactoring we talked about here.

chimp1984: Yes the UI code in Bisq is pretty messy unfortunately. Clear separation can also be done in Java code of course.
It was not premature but a result of experience of laggy UI and then got changed and observed the performance improvement. In Bisq 2 we will not use FXML at all.

ghost: I've created [dedicated issue](https://github.com/bisq-network/bisq/issues/5989) for refactoring efforts. I'm gonna resolve this conversation.

ghost: > In Bisq 2 we will not use FXML at all.

I quickly looked into [bisq2 components code](https://github.com/bisq-network/bisq2/tree/main/desktop/src/main/java/bisq/desktop/components) and they looks really promising!

desktop/src/main/java/bisq/desktop/main/funds/withdrawal/WithdrawalView.java
ghost: [Fixed](https://github.com/bisq-network/bisq/pull/5963/commits/35f01a6ca3f9a457471d7c9e9eb0a0645d315547)

desktop/src/main/java/bisq/desktop/main/funds/transactions/TransactionsView.java
ghost: [Fixed](https://github.com/bisq-network/bisq/pull/5963/commits/5aae20fa3aa230ad21c7edb31519b6ad11f35da5)

desktop/src/main/java/bisq/desktop/main/funds/locked/LockedView.java
ghost: [Fixed](https://github.com/bisq-network/bisq/pull/5963/commits/2f6c249e3197f8f2a007b717daf6132b3bf7fed9)

ghost:
@chimp1984 I fixed these small issues you mentioned on particular code lines.

### 1.
 Regarding code reusability I've moved it forward, details:
- I created [FilteringUtils](https://github.com/xyzmaker123/bisq/blob/e010a98402a05c1e1a86e50eb98722734aee5b32/desktop/src/main/java/bisq/desktop/util/filtering/FilteringUtils.java) that supports filtering upon certain entities
- I created generic [FilterBox](https://github.com/xyzmaker123/bisq/blob/b280b0ac5c93ca832ab81a9dfbc22fe970aa41e4/desktop/src/main/java/bisq/desktop/components/list/FilterBox.java) component that encapsulates filtering layout, listener and common logic


The main problem I have now is some inconsistency in code - fields are rendered in two ways:
A) with model usage, like [here](https://github.com/bisq-network/bisq/blob/8fa66ea7aafd4ebfdcda4b51a2f0eee7a576d2cd/desktop/src/main/java/bisq/desktop/main/portfolio/bsqswaps/UnconfirmedBsqSwapsViewModel.java#L107-L124)
B) inside item, like [here](https://github.com/bisq-network/bisq/blob/c4c231a8182421602e7e4ff8ec0423c3cdf75af9/desktop/src/main/java/bisq/desktop/main/funds/locked/LockedListItem.java#L111-L121)

I think we should unify it (personally I think B approach is better). What do you think?


I'll appreciate your feedback about changes I made so far.


## 2.
> basically it shoud always take the display string as that is what users would expect to be able to filter. Maybe there are some cases where additional domain data filtering is useful as well, but should be additional to the display string case.

Let's polish it after all of the work regarding reusability will be done

## 3.
> I am not so 100% sure if we should show the filter fields by default or extend it just when the user demand it. E.g. add a filter icon and when the user opens it show the field. That would help to not overload more the already overloaded UI.
We can persist the selection of hte icon, so users who prefer to have the filter get it automatically displayed at next restart.

It seems to be quite complex - I prefer not doing it in the scope of this particular pull request.

chimp1984:
> I think we should unify it (personally I think B approach is better). What do you think?

Yes agree. Don't remember the reasons why that was in the ViewModel, maybe because it accessed other ViewDataModel code or as it was already used from earlier code. But I also dislike that.

Rest looks good to me. Not sure if I find time for a full review... If @ripcurlx does not find time ping me again. 

Thanks.

ghost:
>> I think we should unify it (personally I think B approach is better). What do you think?

> Yes agree.

This part is finished - [`FilterBox`](https://github.com/xyzmaker123/bisq/blob/b280b0ac5c93ca832ab81a9dfbc22fe970aa41e4/desktop/src/main/java/bisq/desktop/components/list/FilterBox.java) is now used consistently across all of the lists.

ripcurlx:
I'll try to review it tomorrow. Could you please resolve the conflict of this PR with master which was caused by a merge today. Thanks!

ghost:
Thanks @ripcurlx. It's rebased.

desktop/src/main/java/bisq/desktop/components/list/FilterBox.java
```diff
@@ -0,0 +1,56 @@
+package bisq.desktop.components.list;
```
ripcurlx, line 18: Copyright notice is missing.

ripcurlx:
I'm in the middle of an in-depth code review and started now to test it in the client itself.
When entering send funds it fails with following exception:
```
java.lang.NullPointerException: Cannot invoke "bisq.desktop.components.list.FilterBox.initialize(javafx.collections.transformation.FilteredList, javafx.scene.control.TableView)" because "this.filterBox" is null
	at bisq.desktop.main.funds.withdrawal.WithdrawalView.initialize(WithdrawalView.java:190)
	at bisq.desktop.common.view.InitializableView.initialize(InitializableView.java:41)
	at javafx.fxml.FXMLLoader.loadImpl(FXMLLoader.java:2655)
	at javafx.fxml.FXMLLoader.loadImpl(FXMLLoader.java:2548)
	at javafx.fxml.FXMLLoader.load(FXMLLoader.java:2517)
	at bisq.desktop.common.fxml.FxmlViewLoader.loadFromFxml(FxmlViewLoader.java:107)
```

ripcurlx:
I think it would be great to ignore lower upper case differentiation in transactions on the Details column as users are expecting it like that IMO

ripcurlx:
> I think it would be great to ignore lower upper case differentiation in transactions on the Details column as users are expecting it like that IMO

Similar behavior would be great for:
- `My Open Offers`  for Date/Time, Market, payment method and offer type
- `Open Trades` for Date/Time, Market,Payment method and My role
- `Unconfirmed BSQ Swaps` for Date/Time,Market and Offer type (actually not sure why we display the market at all in BSQ Swaps)
- `History` for Date/Time, Market, Offer type and status
- `Failed` for Date/Time, Market, Offer type and status

But great to see this option now for all lists 👍

desktop/src/main/java/bisq/desktop/main/portfolio/bsqswaps/UnconfirmedBsqSwapsView.java
```diff
@@ -242,25 +231,23 @@ protected void activate() {
...
+                columns[ColumnNames.TX_FEE.ordinal()] = item.getTxFeeAsString();
+                columns[ColumnNames.TRADE_FEE.ordinal()] = item.getTradeFeeAsString();
+                columns[ColumnNames.OFFER_TYPE.ordinal()] = item.getDirectionLabel();
+                columns[ColumnNames.CONF.ordinal()] = String.valueOf(item.getConfidence());
                 return columns;
             };
 
             GUIUtil.exportCSV("bsqSwapHistory.csv", headerConverter, contentConverter,
-                    new UnconfirmedBsqSwapsListItem(), sortedList, (Stage) root.getScene().getWindow());
+                    null, sortedList, (Stage) root.getScene().getWindow());
```
ripcurlx, line 248: As you changed it here and in other places. Do you now why an empty object was used before? 
☝️  @chimp1984

desktop/src/main/java/bisq/desktop/components/list/FilterBox.java
ripcurlx: and on a couple of other places. I've adapted the Codacy checks so it should be checked from now on.

desktop/src/main/java/bisq/desktop/main/portfolio/bsqswaps/UnconfirmedBsqSwapsView.java
ghost: It is passed into [`headerConverter`](https://github.com/xyzmaker123/bisq/blob/9e753458e4e248ac9b0852e05112b79c9dac42c1/desktop/src/main/java/bisq/desktop/util/GUIUtil.java#L330), but `headerConverter` doesn't use it, so passing null doesn't breaks anything. I think we could refactor `exportCSV` a bit deeper - I was planning to do it in dedicated pr.

desktop/src/main/java/bisq/desktop/components/list/FilterBox.java
ghost: [Fixed](https://github.com/bisq-network/bisq/pull/5963/commits/4268d5e451036d3c2f041f897d147dd95286499c)

ghost:
@ripcurlx Thanks for review - I either fixed or put additional explanation on comments

desktop/src/main/java/bisq/desktop/main/portfolio/bsqswaps/UnconfirmedBsqSwapsView.java
chimp1984: > Do you now why an empty object was used before? 

That was required for the exportCSV code... dont remember details but i think with null there was issues

ghost: I tested export csv for bsq swaps with null passed as `emptyItem` and everything went smoothly.

ripcurlx: It also worked for me. I was just curious why it was needed in the past. Let's keep an eye on this.

ripcurlx (approved):
ACK - Tested it on Regtest

ghost:
Column sorting by date is broken in `portfolio -> closed trades view`.  It used to sort by date, now it sorts by date string.

pazza83:
@xyzmaker123 thanks for this. It is working well for me.

PR #5969, opened 2022-01-16
Fix issue with edit button visibility

ghost:
Pull request fixes issue founded during testing:

Screenshot_2022-01-16_22-02-41

seems like using [setManaged](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/Node.html#setManaged-boolean-) wasn't enough.


w0000000t:
Just tested it on my side by editing the relevant file and rebuilding, made sure to add enough offers on both Alice's and Bob's side so scrolling was necessary, and switching between tabs... appears as fixed to me, can't reproduce the artifacts that I noticed during first test.

ripcurlx (approved):
ACK - It is also good practice to reset the button action if not used (see my commmit https://github.com/bisq-network/bisq/pull/5969/commits/fd34053b6b3bb4cf0efa68e678f6e7a771133d58)

PR #5974, opened 2022-01-18
Not create and publish account age witnesses for BSQ swap accounts

ripcurlx:
While creating updated data stores I recognized that the amount of new daily account age witnesses increased by a factor of 10.
The bug was caused as BSQ swap accounts do not extend AssetAccount and are slipping through a re-publish check.

### DIFF
--- a/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
+++ b/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
@@ -26,6 +26,7 @@ import bisq.core.offer.Offer;
 import bisq.core.offer.OfferDirection;
 import bisq.core.offer.OfferRestrictions;
 import bisq.core.payment.AssetAccount;
+import bisq.core.payment.BsqSwapAccount;
 import bisq.core.payment.ChargeBackRisk;
 import bisq.core.payment.PaymentAccount;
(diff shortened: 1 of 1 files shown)


core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
```diff
@@ -228,7 +229,7 @@ private void onBootStrapped() {
     private void republishAllFiatAccounts() {
         if (user.getPaymentAccounts() != null)
             user.getPaymentAccounts().stream()
-                    .filter(account -> !(account instanceof AssetAccount))
+                    .filter(account -> !(account instanceof AssetAccount) && !(account instanceof BsqSwapAccount))
```
chimp1984: Alternatively you could use the `isFiat` in `PaymentMethod`

chimp1984 (approved):
utACK

chimp1984 (dismissed):
utACK

core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
ripcurlx: True - I'll adapt the code.

bisq-github-admin-3 (approved):
ACK - based on https://github.com/bisq-network/bisq/pull/5974#pullrequestreview-855624815

PR #5977, opened 2022-01-18
Show proposal txId with explore and copy icons.

ghost:
Fixes #5923

### Before / After comparison:

Screenshot from 2022-01-18 13-19-09


ripcurlx (approved):
utACK - 👍

PR #5981, opened 2022-01-19
update hostname of BSQ explorer mempool.emzy.de

Emzy:
The new version 2.3.0 of mempool.space supports BSQ exporer only on a separate domain.
update URL to http://bisq.mempool.emzy.de/....

There is a redirect from the old location to the new one. 
So nothing is broken and the PR is low prio.


ripcurlx (approved):
utACK

PR #5983, opened 2022-01-20
Prevent reset offer values after switch to same currency payment account

### LINKED ISSUE #5973: CONFIRM: EDIT OFFER button disabled after changing payment account
### Description

Button to save edited offer is disabled after changing only payment method.

#### Version

1.8.0 and 1.8.1
### Steps to reproduce

Have an open offer, click to edit it, in the following screen change payment method leaving all other fields alone.
The button to save the edited offer will become greyed out, preventing the user from being able to save the edited offer.
Solution is editing the deviation %, and revert it back, with this the button will become clickable
### Expected behaviour

After changing payment method, button to save edits is clicable.
#### Device or machine

Windows 10 and Xubuntu VM

ghost:
Fixes #5973

When we switch to same currency account we don't need to reset `volume`, `minVolume`, `price` and `marketPriceMargin`. 

When we switch to different currency account [onCurrencySelected](https://github.com/bisq-network/bisq/blob/e993773d9f9a8c4534653381264c1917901e99d7/desktop/src/main/java/bisq/desktop/main/offer/bisq_v1/MutableOfferDataModel.java#L391) is invoked, so these params are being reset.

Pull request applies to "create offer", "edit offer" and "duplicate offer" views.

Demo:

1) Before:
before

2) After:
after

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/offer/bisq_v1/MutableOfferDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/bisq_v1/MutableOfferDataModel.java
@@ -329,10 +329,6 @@ public abstract class MutableOfferDataModel extends OfferDataModel implements Bs
 
     void onPaymentAccountSelected(PaymentAccount paymentAccount) {
         if (paymentAccount != null && !this.paymentAccount.equals(paymentAccount)) {
-            volume.set(null);
-            minVolume.set(null);
-            price.set(null);
-            marketPriceMargin = 0;
             preferences.setSelectedPaymentAccountForCreateOffer(paymentAccount);
             this.paymentAccount = paymentAccount;
 


chimp1984 (commented):
I think that requires good testing. The validation checks are rather complex and carry already a few bugs. Resetting the values made it easier to not carry over stale validation state after changes. So I think its good to change that, but just want to emphisis the risks.

ghost:
I agree code around `MutableOfferView` is quite complex. When I was debugging [original issue](https://github.com/bisq-network/bisq/issues/5973) I set debugger breakpoint on [`updateButtonDisableState`](https://github.com/bisq-network/bisq/blob/d901faa29e88df81448e74bd16732c87646ec6bf/desktop/src/main/java/bisq/desktop/main/offer/bisq_v1/MutableOfferViewModel.java#L1297) method and I saw it was invoked 5 or 6 times after changing trading account (in ideal world there should be single call).

Sure - I'll try to prepare testing scenarios that covers changes in this pull request. If you see any edge cases that I should include don't hesitate to share them here.

ripcurlx:
I did try some scenarios and removing this reset code didn't make anything worse, but I found a different bug.

bug

If you switch in `create offer` between payment methods that share the same supported currency and the new payment method has a different default currency set, the price related fields are not updated properly.

@xyzmaker123 Do you want to prepare a PR for this as well?

ripcurlx (approved):
ACK - Tested it on Regtest

ghost:
@ripcurlx Sure - I can try to fix this issue. If you don't mind please prepare appropriate issue and assign me (for better tracking).

PR #5984, opened 2022-01-20
Filter out the BSQ swap payment method from dropdown on edit/duplicat…

### LINKED ISSUE #5982: Open offer edit dialogue disappears after switching payment method to Bisq Swap
### Description

When opening the edit open offer dialogue made with a fiat payment account, if switching the payment method to BSQ Swap the editing dialogue will vanish and go back to the previous panel, giving out the impression of some sort of bug.
#### Version

1.8.0
### Steps to reproduce

Open edit dialogue for open offer to trade with fiat, change payment method to BSQ Swap.
### Expected behaviour

If intended to not let the user switch from fiat payment method to BSQ Swap, then a dialogue would be appropriate to explain this, and then, closing the dialog, the user will be presented with the same edit offer panel, where the previous payment method has been restore in the selection dropdown

### Actual behaviour

Edit panel just vanishes

#### Device or machine

Windows 10

ghost:
Fixes #5982


ripcurlx (approved):

PR #5986, opened 2022-01-20
Fix issue of % price being zero when editing offer

ghost:
Fixes regression issue caused by https://github.com/bisq-network/bisq/pull/5914


ghost (approved):
ACK

ripcurlx (approved):
utACK

PR #5987, opened 2022-01-20
Cleanup: remove unused classes

ghost:
During code exploration I sometimes saw info in intellij idea that certain class is never used. I've tried to run inspection to find all of places like that but it didn't worked properly - issue is [already reported](https://youtrack.jetbrains.com/issue/IDEA-247897) in jetbrains youtrack.

2 classes in this pr I've found manually, 3 others were found by inspection. Probably there is more cases like that in application.

Maybe this pull request will be a starting point to the discussion on a more sophisticated static code analysis solution for bisq.


ripcurlx (changes requested):
NACK - I'd leave InfoDisplay, NewBadge and GUIProfiler as it is used on demand. For the remaining ones feel free to remove them, but also take care that the linked translations are removed as well. Thanks!

ghost:
Fixed

ripcurlx (changes requested):
NACK - Just one last change request. Please only remove translations in the displayStrings.properties file as all other files will be replaced anyways when the translations are updated from Transifex.

ghost:
Fixed

ripcurlx (approved):

PR #5991, opened 2022-01-22
Fix csv export issue on open offers view

ghost:
Icon column for duplicate offer was added [recently](https://github.com/bisq-network/bisq/commit/daf762f8faab8bd9722cb1c6faee89bd8611d599#diff-f38dd85004debdfdecc0b0bbc3d434ac2e62ff6df69d3176a848471b7a4edd42R101) but was not excluded from csv columns calculation. This resulted with NPE during export.

Bug was not reported so far - I've found it during my work on missing filters above lists.


ripcurlx:
> 

Good catch - I missed to check the export. To prevent this in the future I'd rather solve it differently (Please see my commit) WDYT?

ghost:
LGTM

ripcurlx (approved):
ACK

PR #5998, opened 2022-01-25
Update copyright string for Windows

w0000000t:
Substitute 2021 with 2022, and add date range
for consistency with MacOS notice.


ripcurlx (approved):
utACK

PR #6002, opened 2022-01-28
Fix sorting on bonded roles view

### LINKED ISSUE #6000: DAO > Bonding > Bonded roles, ordering by column not working
### Description

Open app panel as per title, try ordering columns by `Name` (easier on the eyes to check), no matter the ordering, it will be all over the place. Same goes for all other columns.

#### Version

1.8.2

### Steps to reproduce

See `Description` above

### Expected behaviour

Rows are ordered by the selected column, either ascending or descending

### Actual behaviour

Rows get somehow "ordered", just the rationale is not apparent, but it's not by the selected column; sometimes usernames are grouped together, but not completely still.

### Screenshots

image
#### Device or machine

Windows 10

ghost:
Fixes #6000


alkum (approved):
ACK. Tested and looks good, sorting works on all columns.

As a side note: would have been easier to review if it was split over multiple commits instead of one big one.

ripcurlx (approved):

PR #6015, opened 2022-01-31
Allow dispute to be closed if trade has paid out

### LINKED ISSUE #6008: Allow trader to self close mediation tickets when trade funds have been sent to donation address
### Description

- Trader opens mediation.
- Mediation ticket fails to open in mediator instance for whatever reason
- Trader waits for time lock to expire and escalates trade to arbitration 

In the instances above a mediation ticket will be open on the users Bisq and it will be unable to be self closed.

Capture

#### Version

1.8.2

### Expected behaviour

Trader's should be able to self close mediation tickets when they have initiated arbitration and sent the funds to the donation address.

### Actual behaviour

Trader's are unable to self close mediation tickets when they have initiated arbitration and sent the funds to the donation address.

### Other info

I could not find the GiHub issue that allowed for trader's to self close mediation tickets. Maybe @jmacxx will be able to provide a link?

ghost:
Fixes #6008 

If a dispute is closed by the mediator but the user's side does not receive that message, user is left with an open mediation ticket that they cannot close.  This change allows the user to close it if the trade no longer has funds locked in.  Possible scenarios are the mediation paid out or the funds went to arbitration.

During testing I encountered a separate existing bug where a self-closed dispute could automatically reopen.  Made a fix to this so that the auto reopening logic additionally checks for an unread message count.  The scenario being that the when the mediator reopens a ticket, the user is notified by a system generated chat message.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeView.java
+++ b/desktop/src/main/java/bisq/desktop/main/support/dispute/DisputeView.java
@@ -399,7 +399,9 @@ public abstract class DisputeView extends ActivatableView<VBox, Void> implements
                 ObservableList<ChatMessage> chatMessages = dispute.getChatMessages();
                 // If last message is not a result message we re-open as we might have received a new message from the
                 // trader/mediator/arbitrator who has reopened the case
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
utACK

pazza83:
Tested this and it works well. Many thanks.

PR #6018, opened 2022-02-02
Fix exception: heightOfLastBlock must match chainHeight

### LINKED ISSUE #6016: error when opening the Bisq application: heightOfLastBlock must match chainHeight
### Description

 when i open the application I recieve this error message "heightOfLastBlock must match chainHeight"

ghost:
Fixes #6016 

If DaoStateStore gets out of sync with BsqBlocks, Bisq ends up in a cycle where it cannot start up.

This PR handles the issue by:
- clearing the DAO state to be rebuilt from resources on the next startup
- triggering a popup informing the user that a restart is needed.

image

---

To reproduce the "heightOfLastBlock must match chainHeight" problem, replace your DaoStateStore file with one from a different instance of Bisq that has not been started in a while, so that it is out of sync with your BsqBlocks.


ripcurlx (approved):
utACK - code changes are looking fine 👍

PR #6019, opened 2022-02-02
Column sorting - fix string comparators

ghost:
Fixes issue with sorting mentioned [here](https://github.com/bisq-network/bisq/pull/5963#issuecomment-1027014554)


ripcurlx (approved):
utACK - code changes are looking fine 👍

PR #6040, opened 2022-02-11
Fix issue selecting specific BSQ UTXO for send.

### LINKED ISSUE #6023: DAO - BSQ Wallet - Select Inputs popup does not accommodate multiple inputs.
### Description

The recently added select Inputs for the DAO BSQ wallet it a great feature. However, if the user has inputs that go over one page then the displayed pop-up is unusable and must be force closed with `ESC`.

#### Version

1.8.2

### Steps to reproduce

With an active BSQ wallet with multiple inputs Go to > DAO > BSQ Wallet > Select Inputs 

### Expected behaviour

Some sort of limitation to window pop-up size with a scroll should inputs exceed screen size.

### Actual behaviour

Window pop-up size fills with inputs and becomes unusable.

### Screenshots

BSQ inputs
Low amount of inputs

multiple inputs
High amount of inputs

ghost:
Fixes #6023

### UI Issue:

- The UI for selecting UTXO inputs was being resized to the number of UTXO available, which caused problems when there is a large set.  Solution is to size the UI to accomodate 3 to 15 rows, which provides a reasonable screen display and a scrollbar for the case of a large UTXO set.

### Transaction issue:

- The `BisqDefaultCoinSelector` is invoked multiple times when a transaction is being built.  Problem was that the coin selection was being reset within, ultimately leading to coin selection being completely ignored.  Solution is to keep the coin selection, resetting after the transaction has been built.

---

<details>
 <summary>Screenshots</summary>

example no UTXO

image

example 2 utxo

image

fabricated example 100 utxo with scrollbar

image

</details>

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/BisqDefaultCoinSelector.java
+++ b/core/src/main/java/bisq/core/btc/wallet/BisqDefaultCoinSelector.java
@@ -56,7 +56,7 @@ public abstract class BisqDefaultCoinSelector implements CoinSelector {
     // We reset the value to null just after we have applied it inside the select method.
     @Nullable
     @Setter
-    protected Set<TransactionOutput> utxoCandidates;
+    private Set<TransactionOutput> utxoCandidates;
 
     public CoinSelection select(Coin target, Set<TransactionOutput> candidates) {
         return select(target, new ArrayList<>(candidates));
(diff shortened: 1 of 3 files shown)


ripcurlx (approved):
utACK

PR #6041, opened 2022-02-12
Update currency filtering logic

### LINKED ISSUE #5300: Currency filtering is inconsistent
### Description

Currency filter (selection drop-down menu) is not working consistently. For instance, there are offers for Zcash and for Doge. I can filter only for Zcash, not for Doge. LBC coin was previously not selectable too, but after an LBC altcoin account was 

created. Either I should be able to filter after I have created the specific altcoin account, or I should be able to filter by all altcoins no matter which accounts I have. I assume, the ladder should apply.

#### Version

1.5.9

### Steps to reproduce

Use the drop-down menu to filter for an altcoin. Some you can filter (in my case for instance Zcash), some you cannot (in my case for instance DOGE). 

### Expected behaviour

Currency filter should be consistent, either with after having created that altcoin account, or filtering through all altcoins no matter which altcoin accounts you have

### Actual behaviour

Not possible to filter for certain altcoins.

#### Device or machine

Windows 64 bit

ghost:
Fixes #5300

### DIFF
--- a/desktop/src/main/java/bisq/desktop/components/AutocompleteComboBox.java
+++ b/desktop/src/main/java/bisq/desktop/components/AutocompleteComboBox.java
@@ -35,6 +35,7 @@ import javafx.collections.ObservableList;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 /**
  * Implements searchable dropdown (an autocomplete like experience).
@@ -44,8 +45,9 @@ import java.util.List;
  * @param <T>  type of the ComboBox item; in the simplest case this can be a String
  */
 public class AutocompleteComboBox<T> extends JFXComboBox<T> {
-    private ArrayList<T> completeList;
-    private ArrayList<T> matchingList;
+    private List<? extends T> list;
+    private List<? extends T> extendedList;
+    private List<T> matchingList;
     private JFXComboBoxListViewSkin<T> comboBoxListViewSkin;
 
     public AutocompleteComboBox() {
@@ -65,13 +67,18 @@ public class AutocompleteComboBox<T> extends JFXComboBox<T> {
     /**
      * Set the complete list of ComboBox items. Use this instead of setItems().
      */
-    public void setAutocompleteItems(List<? extends T> items) {
-        completeList = new ArrayList<>(items);
-        matchingList = new ArrayList<>(completeList);
+    public void setAutocompleteItems(List<? extends T> items, List<? extends T> allItems) {
+        list = items;
+        extendedList = allItems;
+        matchingList = new ArrayList<>(list);
         setValue(null);
         getSelectionModel().clearSelection();
         setItems(FXCollections.observableList(matchingList));
         getEditor().setText("");
+    }
+
+    public void setAutocompleteItems(List<? extends T> items) {
+        setAutocompleteItems(items, null);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


alkum:
ACK.

All currencies are enabled by default only for new installations, but that's fine.

ripcurlx:
For this change I'd like to get more consensus as it will show for the initial dropdown a list with only one market that has actual offers. (AUD has some Mainnet offers)
Bildschirmfoto 2022-02-28 um 13 09 41

I agree that this pre-selection is not such a big thing anymore since we have the search possibility in the combobox we haven't had in the past. Personally I'd be fine with this change if we also change the sorting in the comboxbox from sorting by name to the currency code (first column value) which is more intuitive IMO.

@chimp1984 @m52go @UX-P @leo816

m52go:
Is there any way to sort currencies by trading activity? Maybe not real-time but even updated manually would work since the overall trends don't change very much.

pazza83:
> Is there any way to sort currencies by trading activity? Maybe not real-time but even updated manually would work since the overall trends don't change very much.

It is done in a similar way as proposed; number of current offers, in Market > Offer Book

This seems to work well. Could this be implemented for users going via Buy / Sell ?


Market - offer book

m52go:
Interesting. Mine appear to be sorted alphabetically in spite of having the sort-by-number-of-offers option ticked on in settings. But I do remember my drop-down looking like yours at some point in the past...so I'm not sure what's going on.

pazza83:
@m52go 

I think the option is in Setting > Preferences

Sort market lists with no. of offers / trades

m52go:
Yes, as I said, that option is ticked on but not working for some reason...seems like a bug.

chimp1984:
NACK

I think to spam the whole list with a multitude of currencies never traded is decreasing usability.
The related issue also do not suggest that solution but only complains inconsistency.

I think a better solution would be to show main currencies as we have, but use for the filter predicate the whole list of all currencies so that If I search for a currency which is not in the default list (or user defined list) it still is found.

ghost:
> The related issue also do not suggest that solution but only complains inconsistency.

That solution was suggested by @pazza83 [in discussion](https://github.com/bisq-network/bisq/issues/5300#issuecomment-1034748685) on issue, however I also [wasn't sure](https://github.com/bisq-network/bisq/issues/5300#issuecomment-1034806517) if it's the best one.

> I think a better solution would be to show main currencies as we have, but use for the filter predicate the whole list of all currencies so that If I search for a currency which is not in the default list (or user defined list) it still is found.

I like it, I'll check how difficult it is to implement it that way.

ghost:
I pushed solution suggested by @chimp1984 :
- when user type single letter default/user defined list is used for search
- when user type two or more letters list of all currencies is used for search

leo816 (dismissed):

chimp1984:
> I pushed solution suggested by @chimp1984 :
> 
>     * when user type single letter default/user defined list is used for search
> 
>     * when user type two or more letters list of all currencies is used for search

I think that is confusing.
Just show the user defined list when opening the dropdown and apply the search to all currencies.

chimp1984:
NACK

ghost:
> Just show the user defined list when opening the dropdown and apply the search to all currencies.

[Fixed](https://github.com/bisq-network/bisq/pull/6041/commits/5b4599253b2d7b686154e86a3aacafa16ac0c992)

ripcurlx (approved):
ACK - Tested it on Mainnet

All suggested changes in comments have been implemented.

PR #6044, opened 2022-02-14
Update currency related fields after switch of payment account

### LINKED ISSUE #5999: Switching between payment accounts doesn't update currency related fields
bug

### Description

When switching between payment accounts in e.g. create offer the updating of currency related fields don't work as expected

### Steps to reproduce

- Go to Create Offer
- Select a e.g. SEPA Account and fill in amount and market price margin
- Switch to e.g. Revolut (if EUR is a supported currency in this account and USD was selected before for this payment method) 

### Expected behaviour

Currency fields are changed to USD

### Actual behaviour

EUR is still displayed although USD is displaced in the currency dropdown

ghost:
Fixes #5999

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/offer/bisq_v1/MutableOfferDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/bisq_v1/MutableOfferDataModel.java
@@ -53,7 +53,6 @@ import bisq.core.util.coin.CoinUtil;
 
 import bisq.network.p2p.P2PService;
 
-import bisq.common.app.DevEnv;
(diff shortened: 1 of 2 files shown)


ripcurlx:
All those changes not related to the actual problem makes it really hard to review. I also found a bug that was introduced by those changes (switching between same payment methods will not take the available currency, but rather the trade currency as you refactored out `setTradeCurrencyFromPaymentAccount` ). Please have a look at my adapted PR with the required changes to fix this issue.

Superseded by https://github.com/bisq-network/bisq/pull/6097.

PR #6045, opened 2022-02-14
Fix failed trades view when contract is null

### LINKED ISSUE #5634: Cannot invoke "bisq.core.trade.Contract.getMyNodeAddress(bisq.common.crypto.PubKeyRing)" because the return value of "bisq.core.trade.Trade.getContract()" is null

ghost:
Fixes #5634


ripcurlx (approved):
utACK

PR #6048, opened 2022-02-15
Obtain trade fee settings from filter

ghost:
Fixes issue where user with `DAO disabled` paid lower trade fees.

---

**Before:**

image

**After:**

image


ripcurlx (approved):
utACK

PR #6051, opened 2022-02-16
Fix API FeeService NPE

ghubstan:
Return fee from dao state service if `CoreContext` singleton's `isApiUser`==true.

JFX property `filterManager.getFilter()` is null in apitest cases.

Related to 1e4e7c52846ab4b6f2171ad6d6f4ff82e960e25d.

Based on `master`.

PR #6052, opened 2022-02-16
Fix API FeeService NPE (Alternate)

ghubstan:
Return fee from dao state service if JFX property filterManager.getFilter() == null.

JFX property `filterManager.getFilter()` is null in apitest cases.

This is change may be preferred over https://github.com/bisq-network/bisq/pull/6051.

Related to 1e4e7c52846ab4b6f2171ad6d6f4ff82e960e25d.

Based on `master`.


ghost (approved):
utACK.  I prefer this, simpler solution.

ghost:
This NPE is quite prominent in regtest (if no filter setup), for any offer entry.  (my mistake  :grimacing: )

ripcurlx (approved):
utACK

PR #6053, opened 2022-02-16
Show signed status when applicable in Trade & Dispute info screens

### LINKED ISSUE #6036: Move info from 'Peer Info' to 'Trade Information Screen' and 'Detail Data'
### Background

Following a recent [Bisq support call](https://github.com/bisq-network/support/issues/916#issuecomment-1034033469) I was investigating why a user might have been able to take a SEPA offer of over 0.01 BTC without having a signed account. 

AFAIK signed status for trades in disputes is not information that can be seen in the mediator Bisq instance. This hampered my investigation :(

This lead me to write the issue

### Description

Currently the following are viable on the `peer info` screen

- Onion address
- Number of completed trades
- Payment account age
- Account signed age
- Set tag for that peer

I propose removing from this screen the following:

- Number of completed trades
- Payment account age
- Account signed age

This info should instead be added to:

- `Trade Information`
- `Detail Data`

The main reason for this is I think it would be beneficial for signed status to be shown in the trade information screen and be visible to mediators and arbitrators should the trade enter dispute resolution (currently signed status is not visible to mediators or arbitrators).

Signed status being visible to mediators and arbitrators would help ensure things such as trades were not being take over the btc limits by un-signed accounts, and help with the identification of any nefarious activity.

Also with the recent change to [clear payment account payload info from closed trades](https://github.com/bisq-network/bisq/pull/6001) it would seem to make sense to also clear the above information from the `peer info` screen.

I would propose that double clicking on an avatar just displays:

- Onion address
- Set tag for that peer

I would not propose making any changes to the avatar. Could still show number of completed trades in small circle as current.

#### Version

1.8.2

### Steps to reproduce

Double click a peer's avatar to see what information is displayed. Information displayed varies depending on national payment method (signing required or not required) or if it is an altcoin trade.

### Expected behaviour

Double clicking on a peers avatar just displays:

- Onion address
- Set tag for that peer

From the Trade Information` and `Detail Data` screens the following information can be ascertained*:

- Number of completed trades
- Payment account age 
- Account signed age

*My preference would be for the above to be a snapshot at the time the trade being completed as opposed to updating as time progresses, number of trades increase (as happens now).

ghost:
Fixes #6036

### DIFF
--- a/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
+++ b/core/src/main/java/bisq/core/account/witness/AccountAgeWitnessService.java
@@ -286,7 +286,7 @@ public class AccountAgeWitnessService {
         return new AccountAgeWitness(hash, new Date().getTime());
     }
 
-    Optional<AccountAgeWitness> findWitness(PaymentAccountPayload paymentAccountPayload,
+    public Optional<AccountAgeWitness> findWitness(PaymentAccountPayload paymentAccountPayload,
                                             PubKeyRing pubKeyRing) {
         if (paymentAccountPayload == null) {
             return Optional.empty();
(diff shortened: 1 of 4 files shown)


ripcurlx (approved):
utACK

pazza83:
Thanks @jmacxx working well for me

PR #6054, opened 2022-02-17
Add ability to copy popup text to clipboard.

### LINKED ISSUE #6047: Make all content that appears in pop-ups clickable
### Description

Not all content that appear in user 'pop-ups' are clickable (able to be copied). 

Optimal user experience would allow for any content users are shown in pop-ups / information / confirmation messages etc to be copied.

Examples:

- when a trade experiences issues and mediation is requested it should be easy for the user to copy the content of the pop-up and paste to the mediator. 
- When a user is shown a btc address or transaction id this should be able to be copied.

#### Version

1.8.2

### Steps to reproduce

Here is an example of a transaction id being shown that is not able to be selected.

Emergency Multisig tool

### Expected behaviour

All content in user displayed messages is able to be copied.

### Actual behaviour

Not all content in user displayed messages is able to be copied.

ghost:
Fixes #6047

Adds copy icon to popup dialogs:

image

image

&nbsp;
**Result:**

    Warning
    Security deposit transaction on trade XHykat is still unconfirmed after 56 hours. Check the deposit transaction at a blockchain explorer; if it has been confirmed but is not being displayed as confirmed in Bisq: 
    ● Make a data backup [1] 
    ● Do an SPV resync. [2]

    Contact Bisq support [3] if you have doubts or the issue persists.
    [https://bisq.wiki/Backing_up_application_data, https://bisq.wiki/Resyncing_SPV_file, https://bisq.chat]

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -130,6 +130,7 @@ shared.sendFundsDetailsWithFee=Sending: {0}\nFrom address: {1}\nTo receiving add
 # suppress inspection "TrailingSpacesInProperty"
 shared.sendFundsDetailsDust=Bisq detected that this transaction would create a change output which is below the minimum dust threshold (and therefore not allowed by Bitcoin consensus rules). Instead, this dust ({0} satoshi{1}) will be added to the mining fee.\n\n\n
 shared.copyToClipboard=Copy to clipboard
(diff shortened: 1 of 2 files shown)


alkum:
ACK.

ripcurlx (approved):
utACK

pazza83:
Thanks I have used it a few times already. This is a useful addition!

PR #6057, opened 2022-02-18
Change API grpc.proto triggerPrice field to string [#3]

ghubstan:
As a long, it created confusion and extra client side code for handling display of BTC prices to 8 decimal places, and fiat prices to 4 decimal places.

- Change existing `uint64 triggerPrice` field type in `grpc.proto` to `string triggerPrice`.
- Adjust to `string triggerPrice` change in API core classes.
- Adjust to `string triggerPrice` change in API cli classes.
- Adjust to `string triggerPrice` change in API test cases.

Based on branch `2-change-offerinfo-price-to-string`, PR https://github.com/bisq-network/bisq/pull/6056

PR #6060, opened 2022-02-18
Change API OfferInfo volume/minVolume proto fields to string [#4]

ghubstan:
As longs, they created confusion and extra client side code for handling display of altcoin volumes to 2 decimal places, and fiat volumes to 0 decimal places.
				
- Change `OfferInfo` proto's volume fields' type to string.
- Adjust cli module to `grpc.proto`'s `volume` & `minVolume` field type change (string).
- Add test asserts on volume/min-volume strings from server.

Based on branch `3-change-grpc-proto-triggerprice-to-string`, PR https://github.com/bisq-network/bisq/pull/6057


ripcurlx (approved):

PR #6062, opened 2022-02-19
Change TradeInfo proto's tradePrice & tradeVolume fields to string [#5]

ghubstan:
As longs, they created confusion, and extra client side code for handling display of trade prices and volumes.  

Java/Python API client devs will not have to try to figure out how long values are representing fractional fiat/altcoin prices and volumes.

Just changing volume _or_ price field to string would leave the PR in an incomplete, messy state, so changed both in same PR.  

Summary:

- Change `TradeInfo` .proto's `tradePrice` & `tradeVolume` fields from long to string.

- Adjust `core.api.model` .proto wrappers to price & volume type change.

- Adjust `cli` to `TradeInfo` .proto's price & volume type change.  Quite a bit of refactoring in `cli` module.  A lot of number-string formatting was removed.	
  
- Adjust apitest cases to `TradeInfo` .proto's price & volume type change.

- XMR amounts/volumes are now shown with a precision of 8 in the CLI console.

Based on branch `4-change-grpc-proto-volume-to-string`, PR https://github.com/bisq-network/bisq/pull/6060


ripcurlx (approved):

PR #6063, opened 2022-02-20
Improve BSQ swap error messaging & DAO resync popup.

ghost:
In the event of a failed BSQ swap:  An informative message to the user suggests that checking whether the DAO state may need rebuilding, or SPV may need to be resynced.

Re-enable popup alert indicating when DAO needs to be resynced.

---

image


ripcurlx (approved):
utACK

PR #5985, opened 2022-01-20
Add support/help links on the UI

ghost:
Fixes #5972

---
## Screenshots:
s3
s1
s2
screen4
screen5
screen6

PR #6064, opened 2022-02-20
Fix bug with text rendering on trade step info

ghost:
Fixes https://github.com/bisq-network/bisq/pull/5985#issuecomment-1046070063


ripcurlx (approved):
utACK

PR #6067, opened 2022-02-21
Cosmetic cleanup for popup message copy function.

ghost:
For some popups it is not appropriate to show the copy to clipboard icon as there is no text message.  (i.e. not all popups are informational messages).  An example would be the BSQ Wallet "Select Inputs" popup.

So the [copy icon](https://github.com/bisq-network/bisq/pull/6054) will not be shown in that case.


ripcurlx (approved):
utACK

PR #6065, opened 2022-02-20
Clarify meaning of API/Offer related % fields [#6]

ghubstan:
This change fixes some inconsistencies in offer related API fields representing percentages:  
- Rename double `marketPriceMargin` -> `marketPriceMarginPct`.
- Rename double `buyerSecurityDeposit` -> `buyerSecurityDepositPct`.

In both cases, these fields' values represent "literal" percent values, e.g., 2.55 means 2.55%.

It also fixes inconsistencies between input-pct values and output-pct values. For example, when `createoffer` was passed a `marketPriceMargin`=1.0 (%) param, the `getoffer` response contained an unexpected decimal representation of the pct literal:  `marketPriceMargin`=0.01.  Now `createoffer` input equals `getoffer` output: 1.0 (%).

Summary of commits:
- Disambiguate `grpc.proto` offer pct literal field names.
- Adjust `GrpcOffersService` to new .proto *Pct field names.
- Adjust `core` api pkg classes to new .proto *Pct field names.
- Adjust `cli` module to new .proto *Pct field names.
- Adjust `apitest` cases to new .proto *Pct field names.

Based on branch `5-change-tradeinfo-proto-prc-n-vol-to-string`, PR https://github.com/bisq-network/bisq/pull/6062

PR #6068, opened 2022-02-21
Continue documenting grpc.proto for API reference site [#7]

ghubstan:
This completes a first pass of `grpc.proto` documentation, after some .proto field name and field type changes were made in previous PRs.  Comments have not been added to `pb.proto` defs referenced by the API site, such as all the `*PaymentAccount` messages.  These may be commented later, but providing `CLI`, `Java` and `Python` code examples for the API reference doc is more important at this time.

While still a work in progress, the most recent version of the [slate based API Reference document](https://ghubstan.github.io/slate) could be checked for correctness, missing detail, and to see if more .proto rpc method and/or field types or names need to be changed.

Based on branch `6-disambiguate-api-pct-field-names`, PR https://github.com/bisq-network/bisq/pull/6065.


ghubstan:
I think the markdown->html transformation by the slate engine works pretty well, but there are two problems so far.

1.  When the browser is showing the lowest parts of the page, the nav links default to the last header: 
 https://github.com/slatedocs/slate/issues/1440

2. Reloading the html page can result in navigation to a different internal link.  This seems related to the presence of code examples in the generated markdown.  I'll learn more as I start including more code snippets.

ripcurlx (approved):

PR #6001, opened 2022-01-28
Clear payment account payload info from closed trades.

ghost:
- Removes sensitive payment account information and trader chat messages from closed trades and disputes.
- User can specify in preferences how long to wait before the information is removed, defaults to 20 days.
- The feature is enabled after the user clicks `I UNDERSTAND` shown in the dialog below.
 
Closes #5396

----

### Screenshots:

image

**Settings:**

image

**The highlighted payment account details get removed:**

image

**Trade contract showing payment account information removed.**

image

image

---

### Invocation:

The list of closed trades need to be occasionally evaluated to see if any are eligible for having their sensitive data cleared.  This is primarily done at app initialization, but also is done periodically whenever a trade is added to the list shown in portfolio history.  That is to handle the case of clients which do not restart daily.

Likewise, the list of dispute tickets is evaluated at initialization and also whenever a dispute ticket closed message is processed.

Whenever the evaluation is done, the following log messages are written:

    INFO  b.c.t.ClosedTradableManager: checking closed trades eligibility for having sensitive data cleared 
    INFO  b.c.s.dispute.DisputeManager: ArbitrationManager checking closed disputes eligibility for having sensitive data cleared 
    INFO  b.c.s.dispute.DisputeManager: MediationManager checking closed disputes eligibility for having sensitive data cleared 
    INFO  b.c.s.dispute.DisputeManager: RefundManager checking closed disputes eligibility for having sensitive data cleared

If any actions are taken, they are logged, for example:

    INFO  b.c.trade.model.bisq_v1.Trade: cleared sensitive data from contract;contractAsJson; of trade rfvUYPIx 
    INFO  b.core.support.dispute.Dispute: cleared sensitive data from chat messages; of dispute for trade IOAMZBWH

PR #6069, opened 2022-02-21
[1.8.3 fix] Get payment method ID from offerPayload

ghost:
Fixes bug related to privacy enhancement in 1.8.3.

Was getting an NPE viewing closed trades older than v1.7.0.
makerPaymentMethodId is null for older trades.
makerPaymentAccountPayload can be null for trades where we have removed account info for privacy reasons (see #6001).
The payment method ID can always be obtained from offerPayload.

---

To reproduce: 

- Enable the privacy setting "Clear sensitive data after (days)"
- Restart Bisq for the change to have effect
- Open trade details of a trade done prior to July 2021.
- Error shown.  Closed trades lost.
- Need to restore from backup.


ripcurlx (approved):
utACK

PR #6072, opened 2022-02-22
[1.8.3 fix] Get payment method ID from offerPayload pt.II

ghost:
A persistence issue was revealed during testing of data sets containing old trades pre-1.7.0.
Once the account payload cleared there is a null pointer error.
To rectify, the payment method ID is obtained from offerPayload for pre-1.7.0 trades.
Similar issue to #6069


ripcurlx (approved):
utACK

PR #6079, opened 2022-02-26
Remove GetTradeStatistics service from grpc.proto [#8]

ghubstan:
This protobuf definition and service stub has been in place since the start of work on the API, but was never fully implemented, nor intended to be included in the API beta & v1 releases.

Its presence added a useless section to the [gRPC API Reference](https://ghubstan.github.io/slate).

Based on branch `7-more-grpcproto-comments`, PR https://github.com/bisq-network/bisq/pull/6068.


ripcurlx (approved):

PR #6085, opened 2022-03-03
Use a new key for settings badge.

ghost:
- Rename the key constant variable to SETTINGS_BADGE_KEY to clarify that its value should be different each time it is used for a new feature.
- Use a key which describes the new feature being promoted.

image


ripcurlx (approved):
utACK

Issue #6073, opened 2022-02-23
Unsynced DAO state forcing to restart Bisq in order to fix sync


w0000000t:
As per today's [support weekly call](https://github.com/bisq-network/support/issues/916#issuecomment-1049036040) the issue of "untakeable BSQ swap offers" reported by @jmacxx was further discussed. He explained how there is a limit to the amount of DAO data that peers are "allowed" to download, and swaps introduce a significant data traffic increase, that often will lead to a peer not downloading all the available data, thus resulting out of consensus. This, in turn, will prevent a swap maker to accept takers, or a swap taker to be accepted by makers.
The solution would be to resync DAO state, and what was before an "annoying popup" that got disabled, is being reinstated to this purpose, with the risk of resuming the flow of support requests regarding the aforementioned popup.

My take on this issue, from my understanding, is that the restart of the application will allow Bisq to "resume" the download of the additional data missing from the previous sync attempt, thus restoring the sync state; when the missing data is significant, more consecutive restarts will be prompted by the popup until sync is finally achieved.
Regarding this, why is it needed for the application to be restarted? Is it not possible to have Bisq, on one side, limiting the amount of downloaded data in one go, and on the other side, resume at a later moment (after N minutes, for example) downloading the next batch of allowed data, until sync is finally complete, in a transparent way for the user, and especially not requiring a restart?

As an alternative, if the above is not technically feasible, it would be a nice addition to make the "annoying unsynced DAO popup" be manageable by the user.
For example, briefly explain what happened, that it might affect the user's ability to participate in swaps (and if this is needed, it's necessary to restart Bisq until sync is successful), and that you can ignore the error if you are not interested in swaps; additionally, have a checkbox to "not show this again".

ghost:
The significant data increase of account age witness records was a bug, fixed by https://github.com/bisq-network/bisq/pull/5974.  So that may have settled down by now, and it could be that we're back to the previous "normal" rate of consensus errors.  

The reminder popup was coded to show a maximum of once per bisq session; if we put a "do not show again" option, it would defeat the purpose which is to get all nodes in consensus.  I agree though, wording could be made nicer to explain cause and effect as was done in a related way here: https://github.com/bisq-network/bisq/pull/6063

The other questions @chimp1984 is best suited to answer.

chimp1984:
Agree with @jmacxx 
The data limitation issues should be fixed by now. In the logs it can be seen if that is an issue, but I highly doubt.
[...]

PR #6087, opened 2022-03-04
Fix DAO needs resync popup.

ghost:
The resync popup was being shown when flag `isInConflictWithSeedNode` was set, but there is also a different relevant flag
`isDaoStateBlockChainNotConnecting` which can happen in certain situations.  For that case we also need to show the popup.

Ref:
https://github.com/bisq-network/bisq/issues/6073#issuecomment-1059487173


ripcurlx (approved):
utACK

PR #6086, opened 2022-03-03
Move get (cryptocurrency) offer filter to daemon [No. 1]

ghubstan:
Some API reference & Python bot examples exposed an API bug the Java CLI has been hiding. Due due this Bisq v1 `Offer` entity design:

- In fiat offers, the `baseCurrencyCode`=BTC, `counterCurrencyCode`=FiatCode
- In altcoin offers, `baseCurrencyCode`=AltcoinCode, `counterCurrencyCode`=BTC,

new API examples were not doing what the CLI has been doing for several months, which is get (cryptocurrency) all BTC offers from the server, and filter on the altcoin code.  The CLI side filtering should have been done on the server, as it is in this commit.

A lot of dead gRPC client side offer filtering code is removed as well.

Based on `master`.

PR #6088, opened 2022-03-05
Send meaningful io.grpc.Status.Code to gRPC clients [No. 2]

ghubstan:
Exceptions thrown by the Core API services up to the daemon's Grpc*Services have to be converted into gRPC `StatusRuntimeException`s before being sent to  gRPC clients.  Most of these gRPC `StatusRuntimeException`s had a gRPC `Status.Code.UNKNOWN`, which not helpful to client error handlers.

This change partially resolves the issue by sending more meaningful gRPC response status codes to clients, where possible.  But it is not as comprehensive as it can be for a webapp because HTTP has so many more possible response status codes than the gRPC library (sixteen).   See:  https://github.com/grpc/grpc-java/blob/master/api/src/main/java/io/grpc/Status.java

There are three types of changes:

- Create custom exceptions in `bisq.core.api.exception`.

- Map any custom `bisq.core.api.exception` to a meaningful `io.grpc.Status.Code `within daemon Grpc*Service classes.

- Adjust apitest cases to new grpc status codes.  

Based on branch `move-cli-crypto-offer-filter-to-server`, PR https://github.com/bisq-network/bisq/pull/6086

### DIFF
--- a/apitest/src/test/java/bisq/apitest/method/MethodTest.java
+++ b/apitest/src/test/java/bisq/apitest/method/MethodTest.java
@@ -26,6 +26,9 @@ import bisq.common.util.Utilities;
 
 import bisq.proto.grpc.BalancesInfo;
 
+import io.grpc.Status;
+import io.grpc.StatusRuntimeException;
+
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintWriter;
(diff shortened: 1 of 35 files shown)


ripcurlx (approved):
utACK

PR #6094, opened 2022-03-07
API bug-fix for incorrect BsqSwap's OfferInfo.isActivated value [No. 3]

ghubstan:
I think this bug was introduced when deprecating `GrpcOffersService.getMyOffer(id)`, in favor of using only `getOffer(id)` for 'my' and 'available' offers.  This change explicitly sets the proto's `isActivated` flag in the `OfferInfo` factory methods, and adds checks to api offer test cases.
    
Based on branch `2-improve-grpc-exception-status-code-mapping`, PR https://github.com/bisq-network/bisq/pull/6088


ripcurlx (approved):
utACK

PR #6098, opened 2022-03-09
Exclude persistedAccountName from API payment acct json forms

ghubstan:
This recently added field  (1da71f6c66e790b8124e226467e78d68e7c5452a) is automatically set inside `PaymentAccount`.  It should not be included in the payment acct json forms used by API clients.

Based on branch `master`.


ripcurlx (approved):
utACK

PR #6109, opened 2022-03-17
Update trigger price validation

### LINKED ISSUE #5979: Negative "Below % from market price" + "Deactivate offer if" doesn't work in some situation
Let's say the current market price is 40000 and you are making an offer to buy BTC, with a negative "Below % from market price", say -10% for simplicity, meaning you are willing to pay 10% *more* than the market price for some (possibly unusual) reason, i.e. you're fine with the rate 44000. It would be reasonable for you to think, "But any rate higher than 45100 will be too expensive."

In this situation, you should be able to set the "Deactivate offer if market price is above" = 41000, since 45100 is 110% of 41000. That is to say, you're willing to pay 10% more but only if the market price stays below 41000; you want your offer to be disabled if the market price becomes 41000 or higher (when you'd have to pay 45100 or higher).

However, Bisq doesn't seem to let you do this, showing an error message which looks like: "Value must be higher than 44000."
It is indeed a bit counter-intuitive that the offer should be deactivated at 41000 when you're making a new offer at 44000 - apparently already *higher* than your limit. But with a negative deviation, this 41000 (a possible future market price) actually means 45100 (a price you'd have to then pay).

One might think -10% is unrealistic but the values above are simplified just to illustrate the issue. A similar problem can happen in various realistic situations, where you're willing to (or you have no other choice but to) pay a slightly higher price than the market price. 

Correct me if I'm misunderstanding something, as how these parameters are supposed to interact is somewhat confusing when the deviation is negative.

#### Version
1.8.0

### Steps to reproduce
Not 100% sure, but...
(1) Try to make an offer, with a negative value B for "Below % from market price", e.g. -5%, -10%, etc.
(2) Let M be the current Market price, and N be the price at which a New offer will be made: N = M(1 - B/100) > M with B < 0. Try to use the "Deactivate offer if" value D greater than M but less than N.

### Expected behaviour
A new offer will be made with the initial deviation B (negative), and with the "Deactivate offer if" value D. Note that although D is already less than N, it is greater than the current market price M.

### Actual behaviour
A new offer is not made, refused with an error message "Value must be higher than N".

### PS
https://bisq.community/ is still using v2 onion http://dnkeu7zlmu7vlyox.onion/ which is no longer supported by the current version of Tor. A new Tor Browser user, who doesn't know the situation, may be scared to see the error message from the browser. To avoid various potential problems and confusion, I'd suggest a v3-onion option or no-onion option at all for https://bisq.community/

ghost:
Fixes #5979

### DIFF
--- a/core/src/main/java/bisq/core/util/PriceUtil.java
+++ b/core/src/main/java/bisq/core/util/PriceUtil.java
@@ -72,7 +72,8 @@ public class PriceUtil {
     }
 
     public static InputValidator.ValidationResult isTriggerPriceValid(String triggerPriceAsString,
-                                                                      Price price,
+                                                                      double marketPrice,
+                                                                      String currencyCode,
                                                                       boolean isSellOffer,
                                                                       boolean isFiatCurrency) {
         if (triggerPriceAsString == null || triggerPriceAsString.isEmpty()) {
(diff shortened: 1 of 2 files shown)


alkum:
ACK.

Tested with a -5% market deviation and cutoff price of 41 000 USD.

Before this PR:
Screenshot from 2022-03-17 20-15-16

After PR:
Screenshot from 2022-03-17 20-32-43

core/src/main/java/bisq/core/util/PriceUtil.java
```diff
@@ -72,7 +72,8 @@ public static MonetaryValidator getPriceValidator(boolean isFiatCurrency) {
     }
 
     public static InputValidator.ValidationResult isTriggerPriceValid(String triggerPriceAsString,
-                                                                      Price price,
```
ripcurlx, line 75: I'd rather use the MarketPrice object exposing the currencyCode property instead of adding another parameter. WDYT?

ripcurlx:
@xyzmaker123 Did you also check the trigger price actual execution code? I'd surprised, if only changing the UI validation will do the work as expected.

core/src/main/java/bisq/core/util/PriceUtil.java
ghost: Good idea - [fixed](https://github.com/bisq-network/bisq/pull/6109/commits/55c55dbf472e14d4bea938469c1e54c9cd75df63)

ghost:
> Did you also check the trigger price actual execution code? I'd surprised, if only changing the UI validation will do the work as expected.

@ripcurlx I roughly reviewed [it](https://github.com/bisq-network/bisq/blob/f7fcaa915e0467ce51a41688685ae6d557b39443/core/src/main/java/bisq/core/offer/bisq_v1/TriggerPriceService.java#L116-L185). I think changing validation criteria is enough - we basically add more flexibility to user to set trigger price possible values, but execution logic doesn't change.

ripcurlx:
> 

True - just looked at the code myself. There doesn't seem to be any special handling that matches the one of the validation before.

ripcurlx (approved):
ACK - tested UI validation on Regtest. Behavior for positiv difference to market does not change only for negative difference (against your advantage).

PR #6110, opened 2022-03-17
Separate transaction details in funds views

### LINKED ISSUE #5308: Separate Transaction Details and Trade ID in 'Reserved Funds', 'Locked Funds' and 'Transactions' sections in Bisq
### Description

The Funds section in Bisq becomes a little difficult to use when there is a large amount of transactions.

One proposal to make it more usable is to [add a filter to the 'Funds' section in Bisq to able searching of; Trade IDs, Tx IDs, BTC amount.](https://github.com/bisq-network/bisq/issues/5307)

Another suggestion is to separate transaction details and trade ID in the following sections

- Reserved Funds
- Locked Funds
- Transactions

This would allow for effective sorting of columns and make it easier to sort Trade IDs and transaction details both natively in the Bisq app, but also when exporting to CSV.

#### Version

v1.5.9

### Steps to reproduce

Try sorting by Trade ID in the above sections. You will not be able to. Want to find a specific trade ID? You will need to spend some time to do so if you have a large number of transactions.

Try exporting the above sections to a CSV file. You still will not be able to sort by trade ID without some additional work.

### Expected behaviour

It should be easy to sort by trade ID.

### Actual behaviour

It is impossible to sort by trade ID

### Screenshots

As an example my current Bisq wallet has almost 2,500 transactions. This makes finding a specific trade ID a pain. Here is a screen shot showing the column that combines the transaction details with the Trade ID. These should be separated into two columns.

transactions2

ghost:
Fixes #5308

Reserved view:
01-reserved-funds

Locked view:
02-locked-funds

Transactions:
03-transactions

A few questions / doubts:
- should I show short Id or full Id in list?
- is columns order proper?
- there is no always tradeId available in transactions, but txId - should I show it and maybe use different column name?
- transactions list is crowded - maybe remove some columns (f.x. Transaction Id)?

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -1125,16 +1125,16 @@ funds.tx.direction.sentTo=Sent to:
 funds.tx.direction.receivedWith=Received with:
 funds.tx.direction.genesisTx=From Genesis tx:
 funds.tx.txFeePaymentForBsqTx=Miner fee for BSQ tx
-funds.tx.createOfferFee=Maker and tx fee: {0}
-funds.tx.takeOfferFee=Taker and tx fee: {0}
-funds.tx.multiSigDeposit=Multisig deposit: {0}
-funds.tx.multiSigPayout=Multisig payout: {0}
-funds.tx.disputePayout=Dispute payout: {0}
-funds.tx.disputeLost=Lost dispute case: {0}
-funds.tx.collateralForRefund=Refund collateral: {0}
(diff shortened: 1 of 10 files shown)


alkum:
ACK. Looks good and makes those views more usable.

As to your questions / doubts: I'd leave it as it is. These views are "info heavy" by design, they contain as much detail as possible to help the user categorize / search / document the entries. Therefore, removing columns seems counter-productive. Beautifying and re-arranging I don't think is necessary.

ripcurlx (approved):
utACK - code changes are looking fine. utACK also based on https://github.com/bisq-network/bisq/pull/6110#issuecomment-1071191790

PR #6111, opened 2022-03-18
Trigger price truncation of prompt text

### LINKED ISSUE #5189: Trigger price truncation of prompt text not working

ghost:
Fixes #5189


alkum:
ACK

ripcurlx (approved):
ACK - Tested it on Regtest and it works so far. Of course the perfect solution would have been to only truncate if it exceeds the input field, but that it is a little more complicated to achieve in that case and I think it is sufficient so far as is.

PR #6121, opened 2022-03-28
Show correct buy BSQ state on activation for take offer

ripcurlx:
Looks like this for take offer right now:
Bildschirmfoto 2022-03-28 um 12 45 41
But should look like this:
Bildschirmfoto 2022-03-28 um 12 30 12

It only works for create offer right now as intended.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/offer/bisq_v1/takeoffer/TakeOfferView.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/bisq_v1/takeoffer/TakeOfferView.java
@@ -336,11 +336,9 @@ public class TakeOfferView extends ActivatableViewAndModel<AnchorPane, TakeOffer
             tradeFeeInBsqToggle.setManaged(false);
             buyBsqBox.setVisible(false);
             buyBsqBox.setManaged(false);
-        }
-
-        if (!model.isShowBuyBsqHint()) {
-            buyBsqBox.setVisible(false);
-            buyBsqBox.setManaged(false);
+        } else {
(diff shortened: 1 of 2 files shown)


ripcurlx:
@jmacxx Could you please ACK this? Thanks!

desktop/src/main/java/bisq/desktop/main/offer/bisq_v1/takeoffer/TakeOfferView.java
```diff
@@ -336,11 +336,9 @@ protected void activate() {
             tradeFeeInBsqToggle.setManaged(false);
             buyBsqBox.setVisible(false);
             buyBsqBox.setManaged(false);
-        }
-
-        if (!model.isShowBuyBsqHint()) {
```
ghost, line 341: I see similar code fragment in [MutableOfferView](https://github.com/bisq-network/bisq/blob/00dc7998ef3e77496605def8111eae8a91b7faae/desktop/src/main/java/bisq/desktop/main/offer/bisq_v1/MutableOfferView.java#L277-L280) - shouldn't it be updated similarly?

ripcurlx: It works there as the listener is triggered, but yes I guess it is better to have the same code there as well. I'll push an additional commit.

ghost (approved):
utACK

ghost (approved):
ACK

image

bisq-github-admin-3 (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/6121#pullrequestreview-923476350

PR #6123, opened 2022-03-28
Support Bitcoin Core v22 RPC for JSON export

### LINKED ISSUE #6119: Bitcoin Core v22 deprecated addresses is not handled
### Description
Since Bitcoin Core v22, the Bitcoin API has changed the format of how addresses are returned. So since v22 `deprecatedrpc=addresses` is required to get the old format and in v23 it's removed completely.

Bisq JSON blocks data export does not seem to handle this transition which breaks the Bisq block explorer if upgrading to the latest versions as addresses are not visible.

### Steps to reproduce
Run Bitcoin Core v22 or later with default settings.
Latest Bisq version v1.8.4

### Expected behaviour
The `address` field should be present

Screen Shot 2022-03-24 at 17 35 38

### Actual behaviour

`address` is now missing in the JSON data that is dumped.

Screen Shot 2022-03-24 at 17 33 57

ghost:
Fixes #6119 
Transaction address not provided from bitcoind v22 RPC call.
Get address using BitcoinJ to decode scriptPubKey.

@chimp1984

### DIFF
--- a/core/src/main/java/bisq/core/dao/node/full/RpcService.java
+++ b/core/src/main/java/bisq/core/dao/node/full/RpcService.java
@@ -31,9 +31,11 @@ import bisq.core.user.Preferences;
 import bisq.common.UserThread;
 import bisq.common.config.Config;
 import bisq.common.handlers.ResultHandler;
+import bisq.common.util.Hex;
 import bisq.common.util.Utilities;
 
 import org.bitcoinj.core.Utils;
+import org.bitcoinj.script.Script;
 
 import com.google.inject.Inject;
 
@@ -75,7 +77,7 @@ public class RpcService {
     private static final int ACTIVATE_HARD_FORK_2_HEIGHT_MAINNET = 680300;
     private static final int ACTIVATE_HARD_FORK_2_HEIGHT_TESTNET = 1943000;
     private static final int ACTIVATE_HARD_FORK_2_HEIGHT_REGTEST = 1;
-    private static final Range<Integer> SUPPORTED_NODE_VERSION_RANGE = Range.closedOpen(180000, 210100);
+    private static final Range<Integer> SUPPORTED_NODE_VERSION_RANGE = Range.closedOpen(180000, 220100);
 
     private final String rpcUser;
     private final String rpcPassword;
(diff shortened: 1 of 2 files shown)


core/src/main/java/bisq/core/dao/node/full/RpcService.java
```diff
@@ -350,9 +352,14 @@ private static RawTx getTxFromRawTransaction(RawDtoTransaction rawDtoTx,
                                     }
                                 }
                             }
-                            // We don't support raw MS which are the only case where scriptPubKey.getAddresses()>1
-                            String address = scriptPubKey.getAddresses() != null &&
```
chimp1984: Is the addresses field in `DtoPubKeyScript` still valid/supported?

chimp1984 (dismissed):
ACK

Tested with Bitcoin Core 19 and 22 and tested basic DAO use cases.

Left a comment, not sure if that is relevant to address.

core/src/main/java/bisq/core/dao/node/full/RpcService.java
ghost: AFAIK the addresses field in `DtoPubKeyScript` is populated when bitcoind is older than v22; `null` if >= v22.

I think it can be removed if that is what you prefer, I will test it out with an older `bitcoind` to make sure that is ok.  Noticed that removing it from code also cascades to DAO `PubKeyScript` and `JsonScriptPubKey` so will require some more testing.

chimp1984: Better leave it, just to be sure that it does not cause an exception, but seems not.

chimp1984: Btw. I have not tested with mainnet, would be good to get some full node setup with mainnet tested as well

ripcurlx: @jmacxx Have you tested it with Mainnet as well?

ghost: I am working on mainnet tests, it is going slow due to syncing fullnodes and repeatedly running out of disk space.  Will update here when tests complete.

ghost: The test failed, DAO hashes no longer matched with the seednodes.  Investigation showed that the `addresses` field makes its way into the DAO state records and thus when it is null, affects the DAO state hash.  The way to solve this is to generate the addresses field if necessary at the point we get DtoPubKeyScript from bitcoind.

ghost:
Testing:

- PRE: using v21 bitcoind or earlier.
- Setup genesis DAO environment for Alice/Bob.
- Go through 1 cycle with 2 comp requests & 2 votes.
- Make a BSQ transfer from Bob -> Alice.
- Shutdown, take a bitcoind chain snapshot/copy for future test (label it as v21).
- Run JSON seednode export using v21 bitcoind and set the JSON directory aside (marked as v21_run1).
- Repeat above step and set the JSON directory aside (marked as v21_run2).
- Verify JSON directories match using meld: v21_run1 == v21_run2.
- Switch bitcoind to v22 (POST environment).
- Run JSON seednode export using v22 bitcoind and set the JSON directory aside (marked as v22_run1).
- Repeat above step and set the JSON directory aside (marked as v22_run2).
- Verify JSON directories match using meld: v22_run1 == v22_run2.
- **Compare JSON directories using meld: v21_run1 == v22_run1**.

ghost:
This PRE/POST test :point_up: is successful with the revised code solution https://github.com/bisq-network/bisq/pull/6123/commits/60db9440c7c1a7d617390846a7741cbd5f2dd4e7.  If the code is acceptable I can redo the more laborious mainnet test.

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/6123#pullrequestreview-925096726 and testing by @jmacxx

PR #6130, opened 2022-04-04
Clarify DAO status shown when in sync & peers have conflicts.

ghost:
This is a minor wording change to a status message.  The PRE wording was confusing both for new users and experienced alike, this is cleaned up to let the user know that their DAO state is ok even if peers may not be.

The main effect of the clarification is to show the primary information first (you are in consensus with seed nodes), and the secondary information after (some of your peers are not in consensus).

**PRE:**

Screenshot from 2022-04-04 09-29-21

**POST:**

Screenshot from 2022-04-04 08-51-29


ripcurlx (approved):
ACK

PR #6133, opened 2022-04-05
Fix pause between fee requests

ghost:
`requestFees` method contained logic that prevents too frequent requests. Unfortunately actual code execution was not stopped.


ripcurlx (approved):
utACK - True it seems to be missing this return statement. NIT: the initialization of the variable is not necessary IMO as it is 0 anyways.

PR #6134, opened 2022-04-05
Fix the avatar tradecount tag to include BSQ Swap trades.

ghost:
The counter shown on a peer's avatar indicates how many trades you have done with that peer.  BSQ swaps are currently not reflected in that count, so if you were looking on a peer avatar it would indicate either:

- that you have not traded with that peer (even if you have done BSQ swaps with them), or
- that you have traded with them 'n' times, representing the number of **non-BSQ swap** trades.
 
This PR fixes the issue by counting the number of regular trades and BSQ swap trades.


ripcurlx (approved):
utACK

PR #6136, opened 2022-04-06
Prevent multiple payout screens when agent repeat clicks on close ticket.

ghost:
This PR fixes an issue raised by @refund-agent2

- In Refund Agent ticket view, click process ticket (the gavel icon).
- Select a payout option (e.g. BTC buyer gets trade amount) and populate the notes field.
- Click `Close Ticket`.  Payout screen pops up.
- Without closing the payout screen, click `Close Ticket` again.  Another Payout screen is shown beneath the summary window.
- Click `Close Ticket` again.  A third Payout screen is shown beneath the summary window.
- Click `Cancel`.  The stack of Payout screens is revealed.

Although there is already a guard to [prevent multiple actual funds payouts from happening](https://github.com/bisq-network/bisq/pull/5649), we can also prevent multiple payout prompts to be shown.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/overlays/windows/DisputeSummaryWindow.java
+++ b/desktop/src/main/java/bisq/desktop/main/overlays/windows/DisputeSummaryWindow.java
@@ -132,6 +132,7 @@ public class DisputeSummaryWindow extends Overlay<DisputeSummaryWindow> {
     private ChangeListener<Toggle> tradeAmountToggleGroupListener;
     private ChangeListener<String> compensationOrPenaltyListener;
     private boolean updatingUi = false;
+    private Popup payoutPromptOnDisplay = null;
 
     ///////////////////////////////////////////////////////////////////////////////////////////
     // Public API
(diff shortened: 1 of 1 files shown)


desktop/src/main/java/bisq/desktop/main/overlays/windows/DisputeSummaryWindow.java
```diff
@@ -132,6 +132,7 @@
     private ChangeListener<Toggle> tradeAmountToggleGroupListener;
     private ChangeListener<String> compensationOrPenaltyListener;
     private boolean updatingUi = false;
+    private boolean payoutPromptOnDisplay = false;
```
ghost: I see alternative approach in code base in [`DaoStateMonitorView`](https://github.com/bisq-network/bisq/blob/8e3f95383e1c1784cf7cb735b345e2a73b8456c9/desktop/src/main/java/bisq/desktop/main/dao/monitor/daostate/DaoStateMonitorView.java#L65) and [`TradeStepView`](https://github.com/bisq-network/bisq/blob/ee374c3c39720e540a79311d48a1d6daa93d8abd/desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/steps/TradeStepView.java#L106) where `Popup` instance is stored instead of boolean. To be more consistent we could choose single way.

ghost (dismissed):
utACK (with small refactoring suggestion)

Maybe it would be good to search for other places where multiple popups could be shown (search for "`new Popup()`" in code base). More generic solution could be also considered.

desktop/src/main/java/bisq/desktop/main/overlays/windows/DisputeSummaryWindow.java
ghost: Good suggestion!  I'll adopt that approach.

ripcurlx (commented):
utACK

ripcurlx (approved):
utACK

PR #6145, opened 2022-04-11
Add logo to README

MarnixCroes:
This makes the README look better, imo

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,8 @@
 </p>
 
 <h3 align="center">
-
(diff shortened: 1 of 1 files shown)


pazza83:
Do you mean the .exe file? If so I agree.

ripcurlx:
I just recognized that we also need to remove the Travis build icon and replace it with the Github Actions svg.
I think `[Actions Status]`
should do the trick. Could you please add this to your PR as well?

MarnixCroes:
> Do you mean the .exe file? If so I agree.

@pazza83 
which one?

ripcurlx (approved):
ACK - tested it locally

Bildschirmfoto 2022-04-13 um 10 58 16

PR #6147, opened 2022-04-12
Add inbound Tor connectivity test

### LINKED ISSUE #6141: Misleading error dialog when firewall is misconfigured
### Description

The error dialog appearing when an offer is open but the firewall is misconfigured to block incoming connections is misleading.

#### Version

Bisq v1.8.4

### Steps to reproduce

1. Install Bisq on Whonix as per the instructions on the Whonix wiki, but forget to configure the firewall on the Whonix-Workstation VM (thus, incoming onion connections from the Tor daemon to Bisq will be dropped).
2. Create an open offer.
3. Wait a week.

### Expected behaviour

The error dialog from Bisq should suggest to check the firewall settings (in addition to the other things it suggests).

### Actual behaviour

The error dialog from Bisq says:

> Important private notification!
> 
> You might need to clear your tor cache / restart your Bisq instance, as your 2 offers have not been reachable for at least a week.
> 
> To do that from the Bisq application menus:
> Settings -> Network Info -> Open Tor Settings -> Delete Outdated Tor Files and Shutdown

It does not suggest checking the firewall settings, thus users who have a misconfigured firewall will go on a wild goose chase and not find the actual problem.

### Screenshots

N/A; issue is with text, not graphical.

#### Device or machine

Haswell i7; Whonix 16 inside Qubes 4.1.

#### Additional info

In my experience, firewall misconfiguration is a very common newbie error, and it's often not obvious to newbies that this is something they need to check.

### LINKED ISSUE #6142: Unreachable local Bisq onion takes way too long to detect
### Description

If the local Bisq P2P onion service is unreachable due to a firewall misconfiguration, it takes a week for Bisq to notify the user of this, while it should be trivially easy to detect immediately.

#### Version

Bisq v1.8.4

### Steps to reproduce

1. Install Bisq on Whonix as per the instructions on the Whonix wiki, but forget to configure the firewall on the Whonix-Workstation VM (thus, incoming onion connections from the Tor daemon to Bisq will be dropped).
2. Create an open offer.

### Expected behaviour

Bisq should immediately try to open a TCP connection to the local Bisq P2P onion service (routed through Tor), and when it finds that the connection is dropped, it should quickly convey this to the user.

### Actual behaviour

Bisq waits 1 week before displaying the following error:

> Important private notification!
> 
> You might need to clear your tor cache / restart your Bisq instance, as your 2 offers have not been reachable for at least a week.
> 
> To do that from the Bisq application menus:
> Settings -> Network Info -> Open Tor Settings -> Delete Outdated Tor Files and Shutdown

That's a long delay to notify the user about a problem that can be detected instantly.

### Screenshots

N/A; issue is in timing of graphical display, not the contents of the graphical display.

#### Device or machine

Haswell i7; Whonix 16 inside Qubes 4.1.

#### Additional info

N/A.

ghost:
A user set up Bisq on Whonix.  While Bisq appears to run correctly, their open offers were not able to be taken - rejected with the error message "Maker is offline".  After a while the user was contacted by the support team with a private notification suggesting they clear the Tor cache and restart (the normal suggested resolution for connectivity issues).

The user made a good suggestion that Bisq should do a test attempting to connect back to itself over Tor.  If that fails then show a message informing them to check their firewall configuration.  This PR implements that.  The test runs after Bisq has fully initialized.

Screenshot from 2022-04-12 13-24-34

Fixes #6142
Fixes #6141

### Testing

- In a normal mainnet setup, ran Bisq and the connection to self was successful.
- On Whonix, ran Bisq and the connection to self was unsuccessful resulting in the above popup.

### DIFF
--- a/core/src/main/java/bisq/core/alert/PrivateNotificationManager.java
+++ b/core/src/main/java/bisq/core/alert/PrivateNotificationManager.java
@@ -22,6 +22,11 @@ import bisq.network.p2p.NodeAddress;
 import bisq.network.p2p.P2PService;
 import bisq.network.p2p.SendMailboxMessageListener;
 import bisq.network.p2p.mailbox.MailboxMessageService;
+import bisq.network.p2p.network.Connection;
+import bisq.network.p2p.network.MessageListener;
+import bisq.network.p2p.network.NetworkNode;
+import bisq.network.p2p.peers.keepalive.messages.Ping;
+import bisq.network.p2p.peers.keepalive.messages.Pong;
 
 import bisq.common.app.DevEnv;
 import bisq.common.config.Config;
@@ -36,6 +41,10 @@ import javax.inject.Inject;
 import javax.inject.Named;
 
 import com.google.common.base.Charsets;
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.google.common.util.concurrent.SettableFuture;
 
 import javafx.beans.property.ObjectProperty;
 import javafx.beans.property.ReadOnlyObjectProperty;
@@ -45,16 +54,20 @@ import java.security.SignatureException;
 
 import java.math.BigInteger;
 
+import java.util.Random;
 import java.util.UUID;
+import java.util.function.Consumer;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import org.jetbrains.annotations.NotNull;
 
 import javax.annotation.Nullable;
 
 import static org.bitcoinj.core.Utils.HEX;
 
-public class PrivateNotificationManager {
+public class PrivateNotificationManager implements MessageListener {
     private static final Logger log = LoggerFactory.getLogger(PrivateNotificationManager.class);
 
     private final P2PService p2PService;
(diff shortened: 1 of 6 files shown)


JeremyRand:
Concept ACK.

ripcurlx (approved):

Liisachan:
1) Feedback: An accidental "false warning" (?) was experienced on a machine where offers can be taken.

May-11 ... [JavaFX Application Thread] INFO  b.c.a.PrivateNotificationManager: Send Ping to peer MYSELF.onion:9999, nonce=AAA
May-11 ... [NetworkNode:SendMessage-to-MYSELF.onion:9999] WARN  b.n.p.n.NetworkNode: We are sending a message to ourselves 
May-11 ... [JavaFX Application Thread] INFO  b.c.a.BisqSetup: Sending ping to MYSELF failed. That is expected if the peer is offline.
	ping=Ping(nonce=AAA, lastRoundTripTime=0).
	Exception=Cannot connect to hidden service 

*Tor chache cleared & restarted*
May-11 ... [JavaFX Application Thread] INFO  b.c.a.PrivateNotificationManager: Send Ping to peer MYSELF.onion:9999, nonce=BBB
May-11 ... [NetworkNode:SendMessage-to-MYSELF.onion:9999] WARN  b.n.p.n.NetworkNode: We are sending a message to ourselves 
May-11 ... [JavaFX Application Thread] INFO  b.c.a.PrivateNotificationManager: Received Pong! Pong(requestNonce=BBB) from MYSELF.onion:9999 

2) A possible idea: an "Incoming connection OK/Bad" indicator (e.g. a small green/red circle) could be added on the status bar, showing not just a warning on error, but an ok notification on success.

ghost:
Thanks @Liisachan good suggestion, I'll work on that.  And the self-ping could be attempted a few times rather than just once.

PR #6166, opened 2022-04-26
Fix issue with deleting newly created account

### LINKED ISSUE #6161: [v1.9.0] Newly created SEPA account can't be deleted
### Description

A newly created payment account can't be deleted immediately

#### Version

v1.9.0

### Steps to reproduce

1. Create SEPA account
2. Delete SEPA account

### Expected behaviour

Account is deleted and removed from list.

### Actual behaviour

It is not deleted from list.

#### Additional info

Existing accounts can be deleted.

ghost:
Fixes #6161

`setPersistedAccountName(getAccountName())` was invoked after `paymentAccountsAsObservable.add(paymentAccount);`. `HashSet` is caching hashcodes when adding, so they didn't matched when trying to invoke `paymentAccountsAsObservable.remove(paymentAccount)`.

More details:
 https://stackoverflow.com/a/43554123

### DIFF
--- a/core/src/main/java/bisq/core/payment/PaymentAccount.java
+++ b/core/src/main/java/bisq/core/payment/PaymentAccount.java
@@ -61,6 +61,7 @@ public abstract class PaymentAccount implements PersistablePayload {
     @Setter
     protected String accountName;
     @Setter
+    @EqualsAndHashCode.Exclude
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
ACK - works for me now

PR #6167, opened 2022-04-26
Fix problems with editing account names

### LINKED ISSUE #6162: [v1.9.0] Trying to change account name of same bank account fails with Nullpointer exception
### Description

Use existing transfer with same bank account and try to change the account name.

#### Version

v1.9.0

### Steps to reproduce

1. Use existing account (e.g. Transfer with same bank)
2. Click into Account name and try to change it.

### Expected behaviour

You are able to change the account name.

### Actual behaviour

It throws a Nullpointer exception

### Screenshots

Bildschirmfoto 2022-04-25 um 13 05 01

#### Additional info

It does work e.g. with Perfect Money.

ghost:
Fixes #6162

### DIFF
--- a/desktop/src/main/java/bisq/desktop/components/paymentmethods/GeneralBankForm.java
+++ b/desktop/src/main/java/bisq/desktop/components/paymentmethods/GeneralBankForm.java
@@ -10,6 +10,8 @@ import bisq.core.account.witness.AccountAgeWitnessService;
 import bisq.core.locale.BankUtil;
 import bisq.core.locale.Res;
 import bisq.core.payment.PaymentAccount;
+import bisq.core.payment.payload.BankAccountPayload;
+import bisq.core.payment.payload.CashDepositAccountPayload;
 import bisq.core.payment.payload.CountryBasedPaymentAccountPayload;
 import bisq.core.util.coin.CoinFormatter;
 import bisq.core.util.validation.InputValidator;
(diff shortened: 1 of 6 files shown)


ripcurlx (changes requested):
NACK - Still broken when switching back to not use a custom account name

Bildschirmfoto 2022-04-26 um 11 00 07

```
Apr.-26 11:00:03.709 [JavaFX Application Thread] ERROR bisq.common.setup.CommonSetup: Uncaught Exception from thread JavaFX Application Thread 
Apr.-26 11:00:03.709 [JavaFX Application Thread] ERROR bisq.common.setup.CommonSetup: throwableMessage= Cannot invoke "bisq.desktop.components.InputTextField.getText()" because "this.bankIdInputTextField" is null 
Apr.-26 11:00:03.709 [JavaFX Application Thread] ERROR bisq.common.setup.CommonSetup: throwableClass= class java.lang.NullPointerException 
Apr.-26 11:00:03.709 [JavaFX Application Thread] ERROR bisq.common.setup.CommonSetup: Stack trace:
java.lang.NullPointerException: Cannot invoke "bisq.desktop.components.InputTextField.getText()" because "this.bankIdInputTextField" is null
	at bisq.desktop.components.paymentmethods.GeneralBankForm.autoFillAccountTextFields(GeneralBankForm.java:145)
	at bisq.desktop.components.paymentmethods.BankForm.autoFillNameTextField(BankForm.java:401)
	at bisq.desktop.components.paymentmethods.PaymentMethodForm.lambda$addAccountNameTextFieldWithAutoFillToggleButton$2(PaymentMethodForm.java:144)
```

desktop/src/main/java/bisq/desktop/components/paymentmethods/GeneralBankForm.java
```diff
@@ -136,26 +138,61 @@ void updateHolderIDInput(String countryCode, boolean requiresHolderId) {
...
+
+
+        BankAccountPayload bankAccountPayload = null;
+        CashDepositAccountPayload cashDepositAccountPayload = null;
+        if (paymentAccountPayload instanceof BankAccountPayload) {
+            bankAccountPayload = (BankAccountPayload) paymentAccountPayload;
+        } else if (paymentAccountPayload instanceof CashDepositAccountPayload) {
+            cashDepositAccountPayload = (CashDepositAccountPayload) paymentAccountPayload;
+        }
+
```
ripcurlx: I don't like it so much having those checks in the abstract class for specific implementations. I think it would be better to have a getBankId, getBranchId, getBankName and getAccountNr implementation in the CashDeposit and BankAccount forms WDYT?

ghost: Fixed in latest commit.

Code could be even more simplified when `CashDepositAccountPayload` extends `BankAccountPayload` - do you see any obstacles from doing it (in this pull request)?

ripcurlx: I think - let's keep it to a minimum for now. I'll test it now.

ripcurlx (approved):

PR #6172, opened 2022-04-27
Refactor CashDepositAccountPayload to extends BankAccountPayload

ghost:


ripcurlx (approved):
ACK - Tested on Regtest

PR #6176, opened 2022-04-28
Bugfix: corrupted MaxTradePeriod when offer payment method changed

### LINKED ISSUE #6174: I am in the middle of a CashByMail trade..... and my trading partner and I are both wondering....  Why does Bisq tell us both,  "Max allowed trade period: 4 days"  ......on this CashByMail exchange.....?
### Description

I am in the middle of a CashByMail trade..... and my trading partner and I are both wondering....  Why does Bisq tell us both,  "Max allowed trade period: 4 days" 
......on this CashByMail exchange.....?

Why does Maximum Allowed Trade Time say 4 days...... on ALL of my CashByMail trades?

### Screenshots

photo1651095069

ghost:
Fixes #6174 

`maxTradeLimit` and `maxTradePeriod` were not being set in the `OfferPayload` when an offer was edited, instead they kept their values from the original offer.  So, for example, changing a Zelle offer to CashByMail would result in the trade period being displayed wrong (4 days when it should be 8).

_As a side note, I cannot find any reason for these two fields being in the offer payload at all.  Perhaps they are legacy and should be marked as deprecated?  Most of the code seems to just pick them from the payment method, there only seems to be this one spot in TradeStepView that uses maxTradePeriod from the offer._  ( -> fixed by https://github.com/bisq-network/bisq/commit/e0c13df89f5328779e2f71274568a04e67d7a557)


ripcurlx (approved):
utACK - Thanks for catching that - must be out there for quite a while already 😬

PR #6099, opened 2022-03-09
Dispute subsystem enhancements 

ghost:
Requested by @leo816 and @pazza83

- In mediation summary, show payout suggestion and penalty when applicable
- Add feature enabling traders to upload their chat session to the dispute agent.

---

<details>
  <summary>Screenshots</summary>

&nbsp;

**Show payout suggestion and penalty:**

Payout suggestion text is from the option chosen in the dispute ticket.

image

image

&nbsp;

**Upload trader chat to dispute agent:**

image

&nbsp;

**Dispute agent receives trader chat as an attachment:**

image

</details>

PR #6181, opened 2022-04-29
[1.9.0] Fix uninitialized variable causing NPE in mediation ticket.

ghost:
Fixes https://github.com/bisq-network/bisq/pull/6099#issuecomment-1113103278


ripcurlx (approved):
utACK - @w0000000t Thanks for catching this bug and @jmacxx for fixing! Will prepare v1.9.1 after cherry-picking the remaining fixes.

PR #6182, opened 2022-04-29
Check current fee receiver list whenever a fee tx is validated.

ghost:
Had an issue in support: the current [BTC fee receiver address](https://mempool.space/address/38bZBj5peYS3Husdz7AH3gEUiUbYRD951t) was not recognized when checking a taker fee payment.

This caused mediation to start, analysis of logs showed that the first seednode did not provide any filter objects, and so caused the validator to be initialized without the fee address which comes from the filter.  (The data downloaded from the second seednode did contain 2 filter objects, but by that time initialization had already passed).

This change addresses the issue by obtaining the fee addresses each time a fee payment is validated, instead of a snapshot upon service initialization.  There's no overhead since the fee addresses are already stored in memory, and the benefit is that whenever a new filter is received, its contents are used.


ripcurlx (approved):
utACK

PR #6184, opened 2022-05-01
Add REST API for DAO data

chimp1984:
This is a REST API prototype for DAO data required for Bisq 2.
In its current form it uses a super simple http server and lack swagger support.
The PR is not intended for merge, but rather that some dev use that as base for a production ready implementation.

The REST API could be used also for BSQ explorer use case, but should get some input from @wiz and Simon what would be the best approach.

Also there are some rough plans to add a REST API for normal Bisq trade use case. I would prefer to postpone that as a separate project as it would be a much larger project with a different security context.

This API is intended for private node use case only (e.g. Bisq2 bridge client running on same server, BSQ explorer,...), not for public access. In case a public access use case is desired it should be limited to read only API functionality, thus limiting the risks to dos.

@ghubstan Would you have bandwidth to work on that? If so, can you make a rough estimation and roadmap for the missing parts? See various TODOs in the code for the parts which require more work.


chimp1984:
Superseded by https://github.com/bisq-network/bisq/pull/6190

PR #6188, opened 2022-05-05
Mark user's own offers in Market -> Offer book

### LINKED ISSUE #6171: In Market/OfferBook: Highlight user's own offers
### Description
In Market/OfferBook,  the user's own offers are visually identical in format wih everyone else's. It would be nice to visually distinguish them in some way, e.g. highlighting or outlining.

Why? Because the icons currently used to identify users are too small to visually differentiate without zooming in. It's not easy to tell who is "you".  Many icons appear nearly identical from typical screen viewing distance. Highlighting one's own offer lines would be helpful. (Or, as a more general solution: Replace the existing user icons with a more orthogonal and visually differentiable set, so that you can distinguish your own icon more easily from others)

#### Version
1.8.4

### Steps to reproduce
Make offer. Open Market/OfferBook view.

### Desired behaviour
Highlight or somehow distinguish user's own offers in Market/OfferBook view

### Actual behaviour
User's own offers appear idential to everyone else's (afaict)

#### Device or machine
Arch linux, commodity laptop.

ghost:
Fixes #6171

Description:
Shows "Me" label in buyer/seller column when current user is offer maker.

Preview:
offer-book-me

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -223,6 +223,7 @@ shared.unconfirmedTransactionsLimitReached=You have too many unconfirmed transac
 shared.numItemsLabel=Number of entries: {0}
 shared.filter=Filter
 shared.enabled=Enabled
+shared.me=Me
 
 
 ####################################################################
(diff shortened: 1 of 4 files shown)


ripcurlx (approved):
utACK

PR #6197, opened 2022-05-11
Bugfix: Signed account must show signed age.

### LINKED ISSUE #6196: Trade's popup details report wrong time of account signing
1. In `PORTFOLIO` > `OPEN TRADE` or in `PORTFOLIO` > `HISTORY`, click on a trade's ID to show the popup with the trade's details

2. Hover the mouse over the field "BTC buyer payment details" or "BTC seller payment details"

3. The tooltip that is shown reports wrong time since signing at the end: `(...) / Time since signing: X days`

The time reported is the time that this particular account was created, not signed. In my case for my SEPA account, hovering over "BTC buyer payment details" tells me that my account was signed over 60 days ago (which is my account age), but it was only signed yesterday.

ghost:
Fixes #6196


ripcurlx (approved):
utACK

PR #6200, opened 2022-05-13
Refactor BSQ fee validation to handle multiple txins.

### LINKED ISSUE #6175: False alert shown when trade fees are paid with 3 or more BSQ UTXOs.
Doesn't happen very often, but causes an un-necessary mediation when it does.

I'll fix it.

ghost:
Fixes #6175 

Previously the BSQ trade fee validation would handle either 1 or 2 BSQ UTXOs as input to maker/taker fee txs.
Here it has been refactored to handle any number of BSQ UTXOs as input.
New tests have been added to cover various scenarios.


ripcurlx (approved):
utACK

PR #6201, opened 2022-05-13
Fix refresh issue with x-axis tick labels (Market/Trade charts)

### LINKED ISSUE #6151: Invalid dates displayed in "Market/Trades" view
### Description
The price and volume charts in the "Market/Trades" view often show invalid dates on the time axis. See example screenshot.

Unable to reproduce consistently, but have seen it many times. The invalid date is often the unix epoch 0 date (Dec. 31, 1969), as in the attached example. 

#### Version
1.8.4, Arch linux, from package bisq-bin 1.8.4-1

### Steps to reproduce

Unable to find the magic incantation that causes it, but have seen it many times during normal use. It is not a rare occurrance.

### Expected behaviour

Correct dates shown.

### Actual behaviour

Wrong dates shown. See time axis on screenshot

### Screenshots

attached

#### Device or machine

i86_64 commodity laptop
bisq_ui_date_bug

ghost:
Fixes #6151


ripcurlx (approved):
utACK

Issue #6170, opened 2022-04-26
ERROR PRICE_OUT_OF_TOLERANCE 


bitcointerra:
abr-26 [BisqDaemonMain] WARN  b.core.offer.OpenOfferManager: Received AckMessage with error state for OfferAvailabilityResponse with offerId ORDER ID and errorMessage=An error occurred at task: ProcessOfferAvailabilityResponse
Take offer attempt rejected because of: PRICE_OUT_OF_TOLERANCE 

This error occurs when accepting my offer.

Version 1.8.4

ghost:
@bitcointerra  What is the currency of the offer?

bitcointerra:
Hi, XMR

ghost:
During times of high price volatility this can happen, that your peers price is more than 2% different to yours.  Sometimes this issue can be caused by a pricenode that has gone stale, but all 5 of our pricenodes are displaying a consistent XMR price.

I was not able to find any XMR offers that give the PRICE_OUT_OF_TOLERANCE error at this moment. 

If the problem still happening for you, restart Bisq and try again.  Do you get this error from several offers?  If it is with one particular offer, or user, consider posting the offerId here where it can be looked into.

bitcointerra:
This error occurs when accepting my offer. I'm using bisq_daemon

ghost:
@bitcointerra Ok, so sounds like your node is rejecting the taker's price.  Is yours a fixed price or variable (market price) offer?

Please check your log and let us know what you find, looking for similar messages to these:

`Takers price does not match offer price.  `
`Price at take-offer time:`
`Taker's trade price is too far away from our calculated price based on the market price.`

Please report the values it tells you for takers price and makers price in the log at that point, and the offer id.
Is the onion address of the taker always the same?

bitcointerra:
Hi,
Variable (market price) offer


log below

mai-05 [BisqDaemonMain] WARN  b.c.o.OpenOfferManager: Received AckMessage with error state for OfferAvailabilityResponse with offerId and errorMessage=An error occurred at task: ProcessOfferAvailabilityResponse
Take offer attempt rejected because of: PRICE_OUT_OF_TOLERANCE 

Is the onion address of the taker always not the same! I tested it with another bisq client.

bitcointerra:
Hi,

ID Offer ieveqhar-db1d612e-9918-4a33-9ee3-647f750fc653-184
All offers are having the same problem.

ghost:
[...]

PR #6202, opened 2022-05-14
Fix API editoffer bug: set fixed-price=0 on margin offers

ghubstan:
This is causing erroneous PRICE_OUT_OF_TOLERANCE errors when trying to take offers having (fixed) `price`!=0, and `isUsingMktPriceMargin`=true in the payload.

The API daemon `editoffer's` treatment of (fixed) `price` and `isUsingMktPriceMargin` flag in the API has been inconsistent with the UI.

With this change:  when isUsingMktPriceMargin=true, (fixed) price is set to 0 on the server.  API clients, however, still receive and must show the calculated `price` when isUsingMktPriceMargin=true, making this fix hard to test in the client.  The server will now throw an exception if (fixed) price and isUsingMktPriceMargin flag in the API are not properly set in the API server.

This fix is intended to prevent issues such as https://github.com/bisq-network/bisq/issues/6170 from happening for this reason.  The offer maker edited offers with API, creating inconsistent state described above.  It is hoped the user's offers can be fixed by editing them in the UI.

Based on `master`.

### DIFF
--- a/core/src/main/java/bisq/core/api/CoreOffersService.java
+++ b/core/src/main/java/bisq/core/api/CoreOffersService.java
@@ -464,22 +464,35 @@ class CoreOffersService {
         // in OfferPayload.
         Offer offer = openOffer.getOffer();
         String currencyCode = offer.getCurrencyCode();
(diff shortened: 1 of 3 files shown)


ghubstan:
This PR remains a draft pending further testing.

ripcurlx (approved):
utACK

PR #6207, opened 2022-05-15
Fix API CLI gettrades bug: category param can be uppercase

ghubstan:
When category param value was uppercase, e.g.,  `$ ./bisq-cli --password=xyz --port=9998 gettrades --category=OPEN`,
CLI validation failed:  `Error: No enum constant bisq.cli.opts.GetTradesOptionParser.CATEGORY.OPEN`.
(`gRPC` java/python clients were OK because they have to use the protobuf enum instead of a string parameter value.)
    
This fixes the bug, and `--category=open|closed|failed` now passes CLI validation.
    
Based on `master`.


ripcurlx (approved):
utACK

PR #6211, opened 2022-05-17
Prevent shutdown if trade is in process of being taken.

### LINKED ISSUE #6205: ERROR: Deposit transaction ID N/A
### Description

while exiting my bisq client a buyer took one of my sell offer, this caused an error in the multisig deposit transaction. 
The buyer was able to send me the payment but all my button are grayed out and i cannot release the BTC using my client. Already tried resyncing SPV, restoring a backup and unistalling/reinstalling.
We already went trough mediation, the mediator proposed a solution, but i didn't get it in my client.

#### Version

1.9.1

### Steps to reproduce

Exit bisq when an offer is being accepted.

### Expected behaviour

not sure

### Actual behaviour

bisq thinks i sent btc out of my wallet instead of sending it to a multisig wallet for the trade.

#### Additional info

multisig deposit: 59f5f0702d4ddda5239c4c0475abcd0a39a12df51ad7b0af3984fe2cdf74694b
logs may contain sensible data, i can share it on matrix
@dgcfus:matrix.org

ghost:
Fixes #6205 

We've had a few trades that do not contain a multisig deposit TxId, even though the deposit tx is valid on the blockchain.  An example is #6205.  (Maybe @pazza83 can give an idea how often this happens).  The deposit TxId gets written to the trade once BitcoinJ tells us the tx has been seen by peer nodes, but that can sometimes be delayed by network issues, and the remainder of the P2P negotiation can continue successfully.  This leaves the trade looking ok, but if the user closes Bisq the trade will forever be without a deposit TxId.  The result is a trade which cannot be completed, nor can mediation be opened by the affected trader because the deposit TxId is missing.

This PR makes a check of open trades when the user is closing the app.  If any recent trades are found (less than 5 minutes old), and still in the phase TAKER_FEE_PUBLISHED, it prompts the user to please wait a bit more before closing the app.

image

(The popup can list more multiple trades' information in the rare case that more than one was affected - if you want a screenshot of that scenario let me know).

Anyone (e.g. @w0000000t) if the popup message does not appear clear to you, please feel free to suggest improvements.

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -3080,6 +3080,10 @@ popup.info.shutDownWithOpenOffers=Bisq is being shut down, but there are open of
   they will be re-published to the P2P network the next time you start Bisq.\n\n\
   To keep your offers online, keep Bisq running and make sure this computer remains online too \
   (i.e., make sure it doesn't go into standby mode...monitor standby is not a problem).
+popup.info.shutDownWithTradeInit=Bisq is being shut down, but there are trades still being initialized. \n\n\
+  Specific trade affected: {0}\n\n\
+  If a trade is still in process of being initialized, shutting down can result in the trade missing some important state.\n\n\
+  Do not close Bisq if this trade has not had enough time to initialize correctly.  One minute should be sufficient time.
 popup.info.qubesOSSetupInfo=It appears you are running Bisq on Qubes OS. \n\n\
   Please make sure your Bisq qube is setup according to our Setup Guide at [HYPERLINK:https://bisq.wiki/Running_Bisq_on_Qubes].
 popup.info.firewallSetupInfo=It appears this machine blocks incoming Tor connections. \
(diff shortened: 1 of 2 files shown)


ripcurlx (approved):
utACK - So as soon as there is one trade that had the taker fee tx published within the last five minutes we show this warning. Would be great if we can get rid of those nasty support cases 👍

w0000000t:
Thanks for pinging me @jmacxx 
Just my 2 sats, for something like this, which is "major" IMO (that is, trade being seriously messed up if user proceeds) I would expect something flashier.
My selective blindness saw the attached image and just scrolled past, "hey where's this popup jmacxx was talking about?", then I realized, "is it that innocuous looking notice?" and yes it was 😄 

Here goes my suggestions:
1. Make the popup a red error-ish one so it catches user attention
2. cut the text to a minimum so people actually take the time to read and maybe understand what's writter in there
3. remove the "don't show again" checkbox, we want users to actually go proactive at closing bisq by ignoring the issue, otherwise they shouldn't be able to, so double choice: "Close anyway (DANGER!)" and "Wait a while and try again in a minute"
4. make it so that at next attempt to close Bisq, warning will be shown once again if the initialization wasn't completed yet

Idea about texts:
Popup content
> Trade [blahblah] has not finished initializing; closing now will probably make it corrupted. Please wait a minute and try again.
> (if this is the Nth time you tried and still see this, click <here> to download the log to your favorite location and attach it on (GitHub issues/bisq.chat)

Button text:
> [RISK: close anyway] [wait]

ghost:
:smile: @w0000000t  That's very good feedback - I'll work on adopting it.  Thankyou.

PR #6213, opened 2022-05-18
Improve wording & appearance of trade shutdown dialog.

ghost:
Implements suggestions [here](https://github.com/bisq-network/bisq/pull/6211#issuecomment-1129985098) from @w0000000t .

image

_I considered using  an `Error` box which automatically includes a button to open 'bisq.log', but the error box has text saying this is a bug; whereas this case is simply a network delay.  So I opted for `Warning` box instead.  Simply waiting a bit as prompted will solve the issue in most cases, the rare ones where the condition is permanent can make it to the support channels anyway._


ripcurlx (approved):
utACK 👍

PR #6214, opened 2022-05-19
Remove legacy bitcoind txindex param from apitest harness

ghubstan:
It was put there thinking it would be needed by the test harness and suites, which work fine without it, and it must be removed before API test harness can run against bitcoin-core v23.

See https://github.com/bitcoin/bitcoin/pull/22626#issuecomment-893220371

Based on `master`.


ripcurlx (approved):
utACK

PR #6221, opened 2022-05-22
Block API takeoffer attempt if !sufficient btc in wallet

ghubstan:
This API bug was relying on offer availability checks, but those do not check the taker's wallet.  The take offer model makes the check convenient, and a core.api NotAvailableException can be thrown from CoreTradesService, then mapped to the appropriate gPRC UNAVAILABLE exception sent to clients.

A new test case is added for this change:  Bob's wallet is emptied, he fails to take an offer, and Alice returns Bob's BTC.

Based on `master`.


ripcurlx (approved):
utACK

PR #6224, opened 2022-05-25
Increase seednode RAM limit from 4G to 8G, and BISQ_MAX_MEMORY to match

wiz:
It seems my Bisq seednode `wizseeds` has crashed due to OOM:
```
[InputHandler] ERROR b.c.s.CommonSetup: Stack trace: java.lang.OutOfMemoryError: Java heap space
```

I suppose it is time to increase our RAM limits from 4G to 8G @Emzy @chimp1984 @ripcurlx

### DIFF
--- a/seednode/bisq.env
+++ b/seednode/bisq.env
@@ -5,7 +5,7 @@
 JAVA_HOME=/usr/lib/jvm/openjdk-10.0.2
 
 # java memory and remote management options
-JAVA_OPTS="-Xms4096M -Xmx4096M"
+JAVA_OPTS="-Xms8192M -Xmx8192M"
 
 # bitcoin rpc credentials
 BITCOIN_RPC_USER=__BITCOIN_RPC_USER__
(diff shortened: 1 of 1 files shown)


Emzy:
Mine is running fine with following setting:
JAVA_OPTS="-Xms4096M -Xmx4096M -XX:+UseG1GC -XX:MaxHeapFreeRatio=10 -XX:MinHeapFreeRatio=5 -XX:+UseStringDeduplication"

wiz:
Making the GC more aggressive feels a bit like kicking the can down the road, since I imagine we're just going to have more OOM crashes later on with things like DAO state growing larger over time. Double the the RAM limits for seednodes from 4G to 8G feels reasonable, any reason not to do it?

Emzy:
No. But there was this strange observation that more ram made it run more unstable. So we should test this extra long before activating on all seed nodes.

Emzy:
FWIW I see no "java.lang.OutOfMemoryError" in the last week on my seed nodes

ripcurlx:
@Emzy Just let me know when you guys want this new configuration merged into master. Thanks!

chimp1984:
From 12 nodes there is only 1 which had repeated problems since longer. I do not recommend to just blindly increase RAM without knowing whats the problem. 4GB should be enough.

PR #6227, opened 2022-05-27
Pricenode: configurable exclusion by provider currency

ghost:
The provider exchange Huobi has added BRL with an inaccurate rate such that users have noticed and complained that the Bisq index rate for BRL is wrong.

This PR adds the ability to exclude a currency feed from a specific provider, such that we can ignore the bad BRL rate from Huobi.  In essence it is a follow-on from #5582 which allowed global currency exclusions; this is a provider level exclusion.

@Emzy would you be ok to try out this PR?

---

### Config

Added to `application.properties`.  This is the suggested value to fix the Huobi issue:

    bisq.price.fiatcurrency.excludedByProvider=HUOBI:BRL

Multiple providers can have exclusions, and more than one ccy can be excluded as follows (a comma separated list of items in the format `PROVIDER_NAME:CCY`)

    bisq.price.fiatcurrency.excludedByProvider=HUOBI:BRL,HUOBI:GBP,BINANCE:SEK,BINANCE:AUD


ghost:
Added commit https://github.com/bisq-network/bisq/pull/6227/commits/031095896712ecff3a13b6af62fe01da7179b2fc at request of @Emzy, to completely remove Huobi as a provider.  This is because they started feeding a bad EUR rate, 40% below other providers.  So without Huobi we'll not currently need to exclude by provider currency, but the feature can still be useful in the future.  The `application.properties` setting will do no harm referencing a non-existent provider -- it can remain as indication of how the feature can be used.

Emzy:
I'm running this pr on my Pricenode right now. Because of a very wrong EUR price from Huobi this is necessary for now.

ripcurlx (approved):
utACK

PR #6230, opened 2022-05-31
Pricenode: Add JVM flag ExitOnOutOfMemoryError

### LINKED ISSUE #6225: Bisq seednode exits cleanly on fatal error, causing systemd to not automatically restart it
In the `/etc/systemd/system/bisq.service` file I correctly enabled `Restart=on-failure`, but since Bisq exits with signal 0 on a fatal Out Of Memory heap error, systemd did not restart it and my seednode `wizseeds` was offline until I manually restarted it.

Bisq should exit with error code 1 if such fatal errors such as Out Of Memory occur so that systemd can automatically restart it.

alkum:
Add flag that causes the JVM to exit with exit code 3 whenever there is an OutOfMemory exception for the pricenode service, which causes the systemd bisq-pricenode service to restart.

Can be tested by running the pricenode on a VM with very little memory, or by artificially limiting the JVM memory in addition to this flag: `-Xmx10m -XX:+ExitOnOutOfMemoryError`. The tests should result in an endless restart cycle for the pricenode service (`sudo systemctl status bisq-pricenode` should show service uptime of max 1-2 seconds each time it's called).

Fixes #6225

### DIFF
--- a/pricenode/bisq-pricenode.env
+++ b/pricenode/bisq-pricenode.env
@@ -1,1 +1,1 @@
-JAVA_OPTS=""
+JAVA_OPTS="-XX:+ExitOnOutOfMemoryError"
--- a/seednode/bisq.env
+++ b/seednode/bisq.env
@@ -5,7 +5,7 @@
 JAVA_HOME=/usr/lib/jvm/openjdk-10.0.2
 
 # java memory and remote management options
-JAVA_OPTS="-Xms4096M -Xmx4096M"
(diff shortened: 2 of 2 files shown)


ghost:
You could do the same for seednode, to resolve issue #6225.

https://github.com/bisq-network/bisq/blob/ad15fcd5c4947325dfbcca16c7ec940cc85ec8b3/seednode/bisq.env#L1-L8

alkum:
Done.

ripcurlx (approved):
utACK

wiz:
Thank you 🙏

PR #6231, opened 2022-05-31
Clean up tor files at seed node startup

chimp1984:


ripcurlx (approved):
utACK

Emzy:
I have no exact measurements but I see no changes on my seednode running this change.
It seems to change nothing for the reliability of tor.

chimp1984:
> I have no exact measurements but I see no changes on my seednode running this change. It seems to change nothing for the reliability of tor.

Thanks for the feedback.

PR #6235, opened 2022-06-04
Prevent account creation for banned Revolut GBP sort code 040075

### LINKED ISSUE #6218: Ban Revolut Sort Code '040075' from being used for Faster Payment trades.
### Description

Recently there have been a handful of cases where trades on the GBP/BTC have been unsuccessful due to BTC Seller using Revolut to receive payment via Faster Payments.

As anyone with a Revolut account has access to a Faster Payment account (for immediate GBP debits and credits) it seems that UK banks have issues with people making large ~£1,000+ payments to Revolut Faster Payments accounts. These can be identified as they all use the sort code '040075'.

What usually happens is the BTC Buyer gets an automated message or call from their bank asking for more details about the payment. Following this they may or may not allow the payment.

This has lead to a number of unsuccessful trades.

Of course it is unknown how many trade complete successfully but there are a fair few ones that end up not being successful.

SEPA accounts for Revolut have been banned from Bisq. I assume this is for similar reasons. Ie banks do not like making SEPA payments to Revolut accounts without some additional checks.

Therefore, I am proposing that Revolut accounts are banned from trading on Faster payments.

This would involve not allowing users to add Faster Payment accounts with the sort code: 040075

#### Version

1.9.1

### Current behaviour

Traders can add Revolut Faster Payment accounts

### Expected behaviour

Traders will not be able to add Revolut Faster Payment accounts

#### Alternatives

Allow trading with Revolut accounts but:

- If the Buyer is unable to send GBP to the seller then the Seller either has to provide an alternative Faster Payment account OR compensate the seller 2% of the trade amount to cover the buyers fees.
- If the Buyer using a Revolut account is unable to send GBP to the seller then the Buyer can be penalized for cancelling the trade.

ghost:
Fixes #6218.  As discussed here https://github.com/bisq-network/support/issues/916#issuecomment-1143861666.
- New GBP accounts cannot be created referencing sort code 040075 which is Revolut.
- Prevents new sellers from selling GBP using Revolut.
- Issues with existing sellers will be handled via mediation.
- Faster Payments and National Bank Transfer payment methods affected.

<details>
<summary>Screenshots</summary>

image

image

</details>

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -4250,6 +4250,7 @@ validation.bic.letters=Bank and Country code must be letters
 validation.bic.invalidLocationCode=BIC contains invalid location code
 validation.bic.invalidBranchCode=BIC contains invalid branch code
 validation.bic.sepaRevolutBic=Revolut Sepa accounts are not supported.
+validation.gbp.revolutBranch=Revolut Branch accounts are not supported.
 validation.btc.invalidFormat=Invalid format for a Bitcoin address.
 validation.bsq.invalidFormat=Invalid format for a BSQ address.
 validation.email.invalidAddress=Invalid address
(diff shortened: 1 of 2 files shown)


ripcurlx (approved):
utACK

PR #6241, opened 2022-06-08
Start payment screen reftext message & sizing fixes.

### LINKED ISSUE #6238: Payment started button not shown on small res screens
User in support reported with domestic wire transfer payment method and screen resolution 1920x1080.
Problem also applies to other payment methods when the screen is sized to smallest.

The payment started button is not accessible (it is below the bottom of the window).
Workaround is to click on any field that can receive focus, then press [ENTER] to invoke the default button.

**Proposed fix:**
The payment details sub-pane should have a scrollbar when needed.

<details>
<summary>screenshots:</summary>

image

Same issue for other payment methods e.g. cash by mail:

image

</details>

ghost:
Fixes #6238 "Payment started button not shown on small res screens"
Fixes #6212 "Add "LEAVE PAYMENT REASON BLANK" to payment details panel"

See screenshots for details.

For payment methods that [show the reftext warning](https://github.com/bisq-network/bisq/discussions/6212), it will be displayed together with the payment details and "Payment Started" button.

On machines where the actual vertical resolution is < 800 pixels, it will show a scroll bar for the make payment step, to prevent a situation where the start payment button was not able to be put into view.

Other changes to save vertical space are to limit the vertical size of address textareas somewhat as they are scrollable anyway.

image

### DIFF
--- a/core/src/main/java/bisq/core/payment/payload/AssetsAccountPayload.java
+++ b/core/src/main/java/bisq/core/payment/payload/AssetsAccountPayload.java
@@ -74,6 +74,9 @@ public abstract class AssetsAccountPayload extends PaymentAccountPayload {
     }
 
     @Override
+    public boolean showRefTextWarning() { return false; }
+
+    @Override
     public byte[] getAgeWitnessInputData() {
         return super.getAgeWitnessInputData(address.getBytes(StandardCharsets.UTF_8));
     }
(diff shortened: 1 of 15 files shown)


ripcurlx (approved):
utACK

w0000000t:
I was thinking more on the lines of having a field like the others for `Amount`, `Account`, `Bank name`, this one titled `Payment reason`, and inside `LEAVE EMPTY!!!11ONE`
But I suppose someone would end up writing just that in the payment reason 😅 so probably have the field empty under `Payment reason`, and under it, in small red characters (like when you try and use a Revolut routing number when creating faster payment account) it says "leave empty! or agree with peer in trader chat about a specific reason"

w0000000t:
In general, I feel that the way this warning is implemented, it's not relevant enough to a user, they wouldn't perceive how important it really is to leave the reason blank.

PR #6247, opened 2022-06-14
Conditionally block API's send payment sent/rcvd msgs

ghubstan:
Fixes API bug incorrectly permitting payment sent / received msgs before trade deposit tx is confirmed.  Also blocks payment rcvd confirmation msgs from being sent if payment sent msgs have not been sent (yet).  API cannot depend on silent, UI purposed task runner errors if `confirmpaymentstarted` or `confirmpaymentreceived` msgs are sent outside the proper context.  Throw a FailedPreconditionException instead, to be  translated into the proper gPRC `StatusRuntimeException` before being sent to client.

The important change is in `bisq.core.api.CoreTradesService`:

- For BTC buyers and sellers, block the messaging attempt if the taker deposit tx has not been confirmed.
- For BTC sellers, block an attempt to send a `confirmpaymentreceived` message until after receiving a `confirmpaymentstarted` message has been sent from the buyer.

Also apitest send payment sent/rcvd precondition checks:

- verifyPaymentSentMsgIsFromBtcBuyerPrecondition
- verifyPaymentReceivedMsgIsFromBtcSellerPrecondition
- verifyPaymentSentMsgDepositTxConfirmedPrecondition
- verifyPaymentReceivedMsgDepositTxConfirmedPrecondition
- verifyPaymentReceivedMsgAfterPaymentSentMsgPrecondition 

Based on 'master'.

### DIFF
--- a/apitest/src/test/java/bisq/apitest/method/trade/AbstractTradeTest.java
+++ b/apitest/src/test/java/bisq/apitest/method/trade/AbstractTradeTest.java
@@ -43,10 +43,16 @@ public class AbstractTradeTest extends AbstractOfferTest {
     @Getter
     protected static String tradeId;
 
-    protected final Supplier<Integer> maxTradeStateAndPhaseChecks = () -> isLongRunningTest ? 10 : 2;
+    protected final Supplier<Integer> maxTradeStateAndPhaseChecks = () ->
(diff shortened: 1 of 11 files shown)


ripcurlx (approved):
utACK

PR #6249, opened 2022-06-16
Add API method 'getnetwork'

ghubstan:
Returns name of the BTC network API daemon is connected to (`mainnet`, `testnet3`, or `regtest`).

I am working on some Java bot examples in another repo, and want to stop worrying about accidentally simulating trade payments on mainnet, despite all the warnings I add in comments.  The `getnetwork` method allows bots to know what network the API daemon is connected to, and gives API bot devs and users a way to avoid one potential source of costly mistakes.

API CLI manpage and apitest case are included in this PR.

Based on https://github.com/bisq-network/bisq/pull/6248, branch `reverse-get-altcoin-offers-list-order`.

PR #6250, opened 2022-06-17
Add API method 'getavgbsqprice'

ghubstan:
Returns volume weighted average BSQ price in BTC and USD over N days.

The need for this arose while creating a Java BSQ Swap bot example (PR pending).  API bots can use this to determine whether or not to take swap offers based on their price distance above or below the 30 (or 90) day trade price average.

Based on https://github.com/bisq-network/bisq/pull/6249, branch `add-api-wallet-service-getnetwork-method`.


ripcurlx (approved):
utACK

PR #6253, opened 2022-06-20
Block API attempt to swap BSQ with insufficient funds

ghubstan:
This API bug was relying on the UI purposed `TaskRunner` to bubble up an insufficient funds error to the client, but it cannot do that.  This change blocks the `takeoffer` attempt if sufficient funds are not found after the swap model calculates the taker's trade amounts.

Related: https://github.com/bisq-network/bisq/pull/6221
Based on https://github.com/bisq-network/bisq/pull/6250, branch `add-api-price-service-getaverageprice`.


ripcurlx (approved):
utACK

PR #6260, opened 2022-06-23
add tutorial to clone exclusively specific branch

w0000000t:
This is intended both as self reference, and introductory help for possible new testers, so it is possible to only fetch the needed files, and at the same time have the current latest release checked out to test it.


ripcurlx (approved):
utACK

PR #6262, opened 2022-06-23
Fix snake_case field names

ghubstan:
Be consistent, and x-language.

API reference doc needs snake_case proto msg field names (and so do non-java client devs who want to follow their language's field name conventions).

Based on `master`.


ripcurlx (approved):
utACK

PR #6135, opened 2022-04-06
Improve navigation structure

ripcurlx:
Based on https://github.com/bisq-network/bisq/discussions/6092 this is the PR that implements all required changes.
I hope I have covered everything, but it is a lot of testing.

Bildschirmfoto 2022-04-06 um 18 03 16
Bildschirmfoto 2022-04-06 um 18 02 33
Bildschirmfoto 2022-04-06 um 18 02 28
Bildschirmfoto 2022-04-06 um 18 02 24
Bildschirmfoto 2022-04-06 um 18 02 19

Anyone who wants to give it a shot and provide feedback is more than welcome.
This PR also includes improvements in other areas as well (e.g. only Payment methods are shown in filter for selected currency) and many more.

PR #6266, opened 2022-06-28
Correct wording of altcoin payto address label.

ghost:
Wording of the make altcoin payment prompt is confusing, legacy from when everything was in relation to BTC buyer and seller.  However since the [navigation menus have been adjusted](https://github.com/bisq-network/bisq/pull/6135) to be more user friendly, user has perspective of being altcoin buyer / seller.

For example user is selling XMR (buying BTC):
Currently prompted:
    `Seller's Monero address = [monero address]`
Should be prompted:
    `Monero Buyer's address = [monero address]`

---

**Screenshot Pre / Post :**

image


ripcurlx (approved):
utACK - Thanks for catching that. I'll cherry-pick it into v1.9.3

PR #6270, opened 2022-06-29
[v1.9.3] Reduce toggle width to prevent horizontal scrollbar in min width

ripcurlx:
Before:
Bildschirmfoto 2022-06-29 um 11 31 32

After:
Bildschirmfoto 2022-06-29 um 11 22 12

It does break the alignment of the columns, but I think the horizontal scrollbar is more ugly IMO.

@chimp1984: WDYT?

### DIFF
--- a/desktop/src/main/java/bisq/desktop/components/chart/ChartView.java
+++ b/desktop/src/main/java/bisq/desktop/components/chart/ChartView.java
@@ -372,7 +372,7 @@ public abstract class ChartView<T extends ChartViewModel<? extends ChartDataMode
         hBox.setSpacing(10);
         collection.forEach(series -> {
             AutoTooltipSlideToggleButton toggle = new AutoTooltipSlideToggleButton();
-            toggle.setMinWidth(200);
+            toggle.setMinWidth(180);
             toggle.setAlignment(Pos.TOP_LEFT);
             String seriesId = getSeriesId(series);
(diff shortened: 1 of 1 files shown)


chimp1984:
Maybe we need a 4th row instead? Rows should contain a logical group...

chimp1984 (dismissed):
Code utACK
Concept: I think a 4th row would be better

ripcurlx:
> Code utACK Concept: I think a 4th row would be better

Yes, but we would still run into the same issue if we want to keep the content of the first three main columns untruncated. 
ATM it is structured as 3 HBoxes on top of each other and all related charts are next to the total value. So there is no easy solution to add a fourth "column" and place one item of the 2nd row into the first row. Maybe we could make "column" 4 and 5 smaller than the first three so we only have truncations in the last two columns. I think I can make some little tweak that might solve our problem.

ripcurlx:
Bildschirmfoto 2022-06-29 um 12 24 42

@chimp1984 I think that should do the trick.

chimp1984:
I will make a PR for it. I think it needs anyway a better grouping of whats related.

chimp1984:
I will group the series more clearly and add tooltip with the details.

Screenshot 2022-06-29 at 14 50 27

chimp1984:
Superseeded by https://github.com/bisq-network/bisq/pull/6274

Issue #6269, opened 2022-06-29
API: Only getting SWAP offers (Python)


dainelli98:
### Description

When I perform the getOffers operation against the API, I am only able to recurve BTC <-> BSQ SWAP offers, while when trying other currencies I receive empty offer lists.

I am using Python with the example described here:
https://bisq-network.github.io/slate/?python#rpc-method-getoffers

#### Version

1.9.2

### Steps to reproduce

Execute above-referenced code using USD, EUR,... currency codes.

### Expected behaviour

Getting available offers.

### Actual behaviour

Getting empty offer list.

#### Device or machine

https://hub.docker.com/layers/python/library/python/3.10.5-bullseye/images/sha256-850b7f7626e5ca9822cc9ac36ce1f712930d8c87eb31b5937dba4037fe204034?context=explore

ghubstan:
Can you give me a cleaner URL?

Are you referring to the Python [get_offers.py](https://bisq-network.github.io/slate/?python#rpc-method-getoffers) example ?

I do not think it is possible to receive BSQ swap offers using that example.  I receive available offers to SELL BTC (for USD) offers, or an empty list if I have not set up a USD payment account / method matching any USD offers created with the same payment method.

dainelli98:
Yep, I was referring to that example (I modified the url in the initial post).

From your comment I understand that as of now in order to check offers for a certain currency you need to add a payment method for that currency.

I did not see any references to that specific aspect in the documentation, so in case that it is not explained there, maybe it would be great to add that explanation or make it more visible.

Furthermore, maybe in future releases that restriction could be removed. I do not see the necessity of providing a payment method if you are using the API only for exploratory purposes.

ghubstan:
>> I understand that as of now in order to check offers for a certain currency you need to add a payment method for that currency.

Yes, `getoffers` will return only offers you "can take", implying you need to have the same payment method / accounts set up for the offers you want to look at.  And you are right:  this restriction is not mentioned in [RPC Method GetOffers](https://bisq-network.github.io/slate/?python#rpc-method-getoffers) doc.  I need to fix that in several places.

>> Furthermore, maybe in future releases that restriction could be removed...

This deserves consideration.

ghubstan:
@dainelli98 I created #6272 that includes an update to the protobuf file used to generate the [API reference](https://bisq-network.github.io/slate/).

[...]

PR #6272, opened 2022-06-29
Update API test-guide and gRPC method docs

ghubstan:
A warning was added to the API test guide:  "Warn user not to run API daemon and GUI at same time".

The grpc.proto `takeoffer` and `getoffer(s)`  rpc method comments were updated to explain that they only work as expected when the user's node has payment account/methods matching the available offers' own payment methods.
    
- `getoffers` returns only the offers with payment methods matching those set up on the user's node
    
- `getoffer` will only return the offer if the user's node has a payment method matching the payment method associated with the requested offer
    
- `takeoffer` will fail if the user's node does not have a matching payment method

Based on `master`.

### DIFF
--- a/apitest/docs/api-beta-test-guide.md
+++ b/apitest/docs/api-beta-test-guide.md
@@ -43,6 +43,14 @@ $ ./gradlew clean build :apitest:installDaoSetup            # if you want to run
 
 ## Running Api Test Harness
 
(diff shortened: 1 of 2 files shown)


ripcurlx (approved):
utACK

ghubstan:
Addresses https://github.com/bisq-network/bisq/issues/6269

PR #6275, opened 2022-06-30
Reset Tor directory when a new release is downloaded.

ghost:
Requested by @chimp1984 : 

When user downloads a new release, they are prompted to open the download directory and shut down.  A flag will be set in order to have the Tor directory reset (sans HiddenService) at next startup.  The Tor cache is cleared by this action, which may improve connectivity.

open_download_directory_and_shut_down

    Jun-29 21:25:34.898 [JavaFX Application Thread] INFO  bisq.desktop.app.BisqAppMain: Tor directory reset

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/overlays/windows/downloadupdate/DisplayUpdateDownloadWindow.java
+++ b/desktop/src/main/java/bisq/desktop/main/overlays/windows/downloadupdate/DisplayUpdateDownloadWindow.java
@@ -17,6 +17,7 @@
 
 package bisq.desktop.main.overlays.windows.downloadupdate;
 
+import bisq.desktop.app.BisqApp;
 import bisq.desktop.components.AutoTooltipButton;
 import bisq.desktop.components.AutoTooltipLabel;
 import bisq.desktop.components.BusyAnimation;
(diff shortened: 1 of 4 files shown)


desktop/src/main/java/bisq/desktop/main/overlays/windows/downloadupdate/DisplayUpdateDownloadWindow.java
```diff
@@ -260,6 +266,10 @@ private void addContent() {
                                             .onAction(() -> {
                                                 try {
                                                     Utilities.openFile(new File(Utilities.getDownloadOfHomeDir()));
+                                                    // We set a flag to clear tor cache files at re-start. We cannot clear it now as Tor is used and
+                                                    // that can cause problems.
+                                                    user.getCookie().putAsBoolean(CookieKey.CLEAN_TOR_DIR_AT_RESTART, true);
```
chimp1984: I think we need to apply it also at close. So better move it directly into the `else` case so its set once download is completed.

chimp1984 (changes requested):

chimp1984 (approved):
utACK

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/6275#pullrequestreview-1024758992

PR #6284, opened 2022-07-08
Inventory monitor: show last successful response time

ghost:
For the monitor e.g. http://46.101.179.224:8080/

In the case where a seednode has gone offline it would be useful to know when it was last reachable.

Add `Last response received: [timestamp]`
If a seednode has never been reachable, it will show `Last response received: none`


chimp1984 (approved):
utACK

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/6284#pullrequestreview-1033344358

PR #6298, opened 2022-07-19
Set desktop version property in resources file

ghubstan:
Several repos are to be extracted from the main bisq repo, and _some_ of their distribution versions will need to match the bisq desktop's version.

This change moves the desktop version property definition to a file that can be read from Gradle builds having the bisq git submodule dependency, i.e., 

             version = file("bisq/desktop/src/main/resources/version.txt").text.trim()

Based on `master`.


ripcurlx (approved):
utACK

PR #6303, opened 2022-07-28
Allow mediated payout signature message to be re-sent

ghost:
Recently with the tor DDoS attack, there has been a [significant uptick](https://github.com/bisq-network/support/issues/916#issuecomment-1197206497) in the number of mediation proposals that could not be accepted because one signing participant's trade signature message did not get through.  The app prevents any re-try of sending the message as it is coded to assume reliable delivery.

This caused the mediators to do a lot more manual payouts (v. time consuming), and as such they are asking for the ability for users to re-send the dispute payout signature message.

Testing shows that removal of the gating, allowing the message to be resent, is an acceptable solution.


ripcurlx (approved):
utACK

PR #6307, opened 2022-07-31
Fix/improve clickable areas of toggle buttons

### LINKED ISSUE #6263: Enable Offers toggle-button activated when not clicking it
### Description

Enable/Disable Offers toggle-button activated if clicked outside of it, to the right.

#### Version

v1.9.2 (`master` branch)

1)  Go to PORTFOLIO -> MY OPEN OFFERS
2)  Click wide to right of the `Enabled` toggle-button (not in the toggle-button area).
3)  Offers are disabled/enabled if clicking outside of the toggle-button area (all the way to the edge of the window).

### Expected behaviour

Offers are not enabled or disabled if not clicking inside the toggle-button.

### Actual behaviour

Offers are  enabled or disabled if clicking way to the right of the toggle-button.

Not a big deal, but confused me before I understood why all my offers were disabled.

### Screenshots

Offers-Enable-Disable-Button

#### Device or machine

Ubuntu 20.04.4 LTS

Android-X13:
Fixes #6263

### DIFF
--- a/desktop/src/main/java/bisq/desktop/components/chart/ChartView.java
+++ b/desktop/src/main/java/bisq/desktop/components/chart/ChartView.java
@@ -177,28 +177,29 @@ public abstract class ChartView<T extends ChartViewModel<? extends ChartDataMode
 
         // Put all together
         VBox timelineNavigationBox = new VBox();
-        double paddingLeft = 15;
+        double paddingTop = 0;
         double paddingRight = 89;
+        double paddingLeft = 15;
         // Y-axis width depends on data so we register a listener to get correct value
         yAxisWidthListener = (observable, oldValue, newValue) -> {
             double width = newValue.doubleValue();
             if (width > 0) {
                 double rightPadding = width + 14;
-                VBox.setMargin(timeIntervalBox, new Insets(0, rightPadding, 0, paddingLeft));
-                VBox.setMargin(timelineNavigation, new Insets(0, rightPadding, 0, paddingLeft));
-                VBox.setMargin(timelineLabels, new Insets(0, rightPadding, 0, paddingLeft));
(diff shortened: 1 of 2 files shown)


ghost (dismissed):
utACK

ghost:
@Android-X13 looks great.  May I request an additional thing?
Could you fix the clickable area of toggle buttons in DAO -> Facts & Figures -> Bsq Supply screen?  There its possible to click on a toggle button and it toggles the one below.

image

Android-X13:
Hi @jmacxx,
The author of the relevant code resorted to a hack in order to decrease the height of the switches and the vertical gap between them:

switch hack

The switches work best in their default height although the gaps increase quite enough which isn't what was wanted.
Unless there's a better way to shrink the toggle's height symmetrically instead of doing hacks, my only solution to this would be to decrease the overlapping between them. But this would mean that their vertical gaps would increase a bit. And the clickable area still won't be ideal (but it will be better than now).
If you still want this, should I push it in this branch?

ghost:
> the clickable area still won't be ideal (but it will be better than now).

Yes, I think an improvement would be worth it.  To my mind it is similar enough to include in this branch.  Thanks.

Android-X13:
Before:

Before

After:

After

The clickabe area is better now and the vertical distance between the switches isn't increased (too) much.
It could be better but the gaps would also grow more.. I think the compromise between height and clickable areas is acceptable now

Android-X13:
Kindly disregard my ramblings in the last comment above. The vertical gaps between toggles in BSQ SUPPLY do increase like shown, but their clickable area is ideal now.

Also, somehow I managed to screw up this PR. Really sorry about that. I'm new to git...

ghost (approved):
ACK

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/6307#pullrequestreview-1072180475

PR #6309, opened 2022-08-03
Show Peer Info window only when left-clicking on avatar

Android-X13:
Fixes minor glitch when right-clicking on a peer's avatar in PORTFOLIO

Steps to reproduce:

- PORTFOLIO > HISTORY > Right-click on a peer's avatar > will show both PeerInfoWithTagEditor window and the "Create new offer like this..." context menu overlapping:

- Click on "Create new offer like this..." > if you were maker, this will open the duplicate offer tab while still keeping the PeerInfoWithTagEditor popup visible:

- Right-clicking to another avatar instead > will show a duplicate "Create new offer like this..." context menu without closing the previous one:

### DIFF
--- a/desktop/src/main/java/bisq/desktop/components/PeerInfoIcon.java
+++ b/desktop/src/main/java/bisq/desktop/components/PeerInfoIcon.java
@@ -36,6 +36,7 @@ import javafx.scene.canvas.GraphicsContext;
 import javafx.scene.control.Label;
 import javafx.scene.control.Tooltip;
 import javafx.scene.image.ImageView;
+import javafx.scene.input.MouseButton;
 import javafx.scene.layout.Pane;
 import javafx.scene.paint.Color;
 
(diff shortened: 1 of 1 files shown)


desktop/src/main/java/bisq/desktop/components/PeerInfoIcon.java
```diff
@@ -170,23 +171,27 @@ protected void addMouseListener(int numTrades,
...
-                .onSave(newTag -> {
-                    preferences.setTagForPeer(fullAddress, newTag);
-                    updatePeerInfoIcon();
-                    if (callback != null) {
-                        callback.avatarTagUpdated();
-                    }
-                })
-                .show());
+        setOnMouseClicked(e -> {
+            if (e.getButton().equals(MouseButton.PRIMARY)) {
```
ghost, line 175:
```suggestion
            if (e.isPrimaryButtonDown()) {
```

ghost (approved):
ACK - tested and works perfectly, just a small code suggestion

desktop/src/main/java/bisq/desktop/components/PeerInfoIcon.java
Android-X13: Thank you @xyzmaker123. `isPrimaryButtonDown` doesn't work correctly in this case and Peer Info window won't show up with a simple left click. It will only show up if left button is pressed the exact moment you release an other mouse button on the avatar

ghost (approved):
ACK

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/6309#pullrequestreview-1071882369

PR #6310, opened 2022-08-04
Remove subproject inventory

ghubstan:
This has been moved into its own repo: https://github.com/bisq-network/bisq-inventory

Based on `master`.


ripcurlx (approved):
utACK

PR #6312, opened 2022-08-06
API takeoffer:  Let user choose intended trade amount

ghubstan:
The API's `takeoffer` method was not allowing users to define an intended trade amount, anywhere between the offer's min-amount and (max) amount.  This change adds support for an `--amount` parameter in the `takeoffer` request, aligning this method's behavior with the UI's.

Based on `master`.


ripcurlx (approved):
utACK

PR #6314, opened 2022-08-08
Possibility to copy addresses / transaction ids

### LINKED ISSUE #4452: Add copy icon to addresses and tx IDs if not there
At lists we do not use the copy icon for copying the data to clipboard. It would be good to add that as well so that users who want to open a blockexplorer can copy it without opening automatically the browser by clicking on it.

### LINKED ISSUE #5278: DAO and Fund section in Bisq seems to be missing ability to copy links
### Description

A minor annoyance of mine but the whole DAO and Fund section in Bisq seems to be missing ability to copy links.

Links all have a little link icon. They all seem to be missing the little copy icon.

This is most annoying when doing a transaction and then providing the Tx ID. Currently I users have to visit the website and then copy the Tx ID from the URL or the webpage. It would be much easier if users could just press the little 'copy icon'. It would also reduce the risk of user less error, 

#### Version

v1.5.9

### Steps to reproduce

Go to the DAO section is Bisq (No option to copy links in multiple sections)
Go to the Funds section is Bisq (No option to copy links in multiple sections)

You can copy from places such as Trade ID details screen

### Expected behaviour

Go to the DAO section is Bisq (option to copy links in multiple sections)
Go to the Funds section is Bisq (option to copy links in multiple sections)

Use the same icon as in the Trade ID details screen

### Screenshots

funds - Copy
No option to copy Tx ID

bonds - Copy
No option to copy Tx Id or GitHub issue

ghost:
Fixes #5278, Fixes #4452

Preview:
copy

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -130,7 +130,6 @@ shared.date=Date
 shared.sendFundsDetailsWithFee=Sending: {0}\nFrom address: {1}\nTo receiving address: {2}\nRequired mining fee is: {3} ({4} satoshis/vbyte)\nTransaction vsize: {5} vKb\n\nThe recipient will receive: {6}\n\nAre you sure you want to withdraw this amount?
 # suppress inspection "TrailingSpacesInProperty"
 shared.sendFundsDetailsDust=Bisq detected that this transaction would create a change output which is below the minimum dust threshold (and therefore not allowed by Bitcoin consensus rules). Instead, this dust ({0} satoshi{1}) will be added to the mining fee.\n\n\n
-shared.copy=Copy
 shared.copyToClipboard=Copy to clipboard
 shared.copiedToClipboard=Copied to clipboard!
 shared.language=Language
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -130,6 +130,7 @@ shared.date=Date
 shared.sendFundsDetailsWithFee=Sending: {0}\nFrom address: {1}\nTo receiving address: {2}\nRequired mining fee is: {3} ({4} satoshis/vbyte)\nTransaction vsize: {5} vKb\n\nThe recipient will receive: {6}\n\nAre you sure you want to withdraw this amount?
 # suppress inspection "TrailingSpacesInProperty"
 shared.sendFundsDetailsDust=Bisq detected that this transaction would create a change output which is below the minimum dust threshold (and therefore not allowed by Bitcoin consensus rules). Instead, this dust ({0} satoshi{1}) will be added to the mining fee.\n\n\n
+shared.copy=Copy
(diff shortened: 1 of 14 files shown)


ghost:
@xyzmaker123 Appreciate the PR, but I have a question about a possible stylistic improvement.

The clickable href and chain icon at the moment do the same thing which seems redundant.  How about switching the chain icon with a copy icon, using that to copy to the clipboard.  So if the user wants to open in a browser, they click the href or if just want to copy then click on the icon?

ghost:
Thanks @jmacxx - for me replacing link icon with copy icon sounds good. I implemented current solution based on @ripcurlx and @pazza83 [comments](https://github.com/bisq-network/bisq/issues/5278#issuecomment-933273539) - I'd like to hear their opinion before reimplementing it

pazza83:
@xyzmaker123 Replacing the link icon with a copy icon sounds good to me.

Do this PR fix both the funds and the DAO section?

ghost:
Let's wait for @ripcurlx confirmation that replacing icon is okay - concept with icon for the link seems to be really old in bisq and it's used also for different kind of links - it probably has some UX grounds.

> Do this PR fix both the funds and the DAO section?

Yes

ripcurlx:
> Let's wait for @ripcurlx confirmation that replacing icon is okay - concept with icon for the link seems to be really old in bisq and it's used also for different kind of links - it probably has some UX grounds.
> 
> > Do this PR fix both the funds and the DAO section?
> 
> Yes

ATM the chain icon is (should) be an indication next to all links that will guide the user to an external url. I'm not sure if it is worth in this case to replace it with a copy icon.

ghost:
> ATM the chain icon is (should) be an indication next to all links that will guide the user to an external url. I'm not sure if it is worth in this case to replace it with a copy icon.

Yeah, I think current solution with copy menu after click right mouse button is good enough.

...

Pushed conflicts fix

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -130,6 +130,7 @@ shared.date=Date
 shared.sendFundsDetailsWithFee=Sending: {0}\nFrom address: {1}\nTo receiving address: {2}\nRequired mining fee is: {3} ({4} satoshis/vbyte)\nTransaction vsize: {5} vKb\n\nThe recipient will receive: {6}\n\nAre you sure you want to withdraw this amount?
 # suppress inspection "TrailingSpacesInProperty"
 shared.sendFundsDetailsDust=Bisq detected that this transaction would create a change output which is below the minimum dust threshold (and therefore not allowed by Bitcoin consensus rules). Instead, this dust ({0} satoshi{1}) will be added to the mining fee.\n\n\n
+shared.copy=Copy
```
ripcurlx: Why not reuse copyToClipiboard?

ghost: Fixed

ripcurlx (approved):
utACK

PR #6316, opened 2022-08-09
Send funds - don't use all available funds by default

ghost:
Fixes https://github.com/bisq-network/bisq/discussions/5866

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/funds/withdrawal/WithdrawalView.java
+++ b/desktop/src/main/java/bisq/desktop/main/funds/withdrawal/WithdrawalView.java
@@ -154,7 +154,7 @@ public class WithdrawalView extends ActivatableView<VBox, Void> {
     private ChangeListener<Number> transactionFeeChangeListener;
     private ToggleGroup feeToggleGroup, inputsToggleGroup;
     private ToggleButton useCustomFee;
-    private final BooleanProperty useAllInputs = new SimpleBooleanProperty(true);
+    private final BooleanProperty useAllInputs = new SimpleBooleanProperty(false);
     private boolean feeExcluded;
     private int rowIndex = 0;
     private final FeeService feeService;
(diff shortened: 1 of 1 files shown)


desktop/src/main/java/bisq/desktop/main/funds/withdrawal/WithdrawalView.java
```diff
@@ -364,7 +364,7 @@ protected void activate() {
             feeToggleGroup.selectToggle(feeIncludedRadioButton);
 
         if (inputsToggleGroup.getSelectedToggle() == null)
-            inputsToggleGroup.selectToggle(useAllInputsRadioButton);
+            inputsToggleGroup.selectToggle(useAllInputs.get() ? useAllInputsRadioButton : useCustomInputsRadioButton);
```
ghost, line 367: Legacy nit: current Bisq coding standards dictate use of curly braces for all conditional statements even one liners.

ghost (changes requested):

desktop/src/main/java/bisq/desktop/main/funds/withdrawal/WithdrawalView.java
ghost: I think it would be better to prepare all files fix in separated pull request. If you'd like I can do research regarding codacy configuration that would force developers to always use curly braces.

ripcurlx: > I think it would be better to prepare all files fix in separated pull request. If you'd like I can do research regarding codacy configuration that would force developers to always use curly braces.

I fear that would be quite a big PR to review 😬 - I'll check why Codacy didn't catch this in their checks.

ripcurlx: AvoidInlineConditionals of Checkstyle doesn't seem to be triggered by this 🤔. I don't remember anymore exactly, but I thought the ? : conditionals are still ok to be used. No strong opinion from my side on this.

ghost: > I fear that would be quite a big PR to review

That's true. I think we can omit this for now and merge this pull request as it is.

ripcurlx (approved):
utACK

PR #6324, opened 2022-08-14
Updating DAO->Governance menu to indicate vote possible

### LINKED ISSUE #3744: DAO->Governance menu not updating to indicate vote possible
### Description

DAO->Governance->Menu still has "BROWSE OPEN PROPOSALS" during blind vote phase even though a vote could happen.

#### Version

dbf621ab4
### Steps to reproduce

1. Start localnet cluster
2. Create proposal to change parameter
3. Advance blocks to blind vote
4. See asymmetrical menus from the proposal creator and the voter

### Expected behaviour

Button should be "VOTE ON PROPOSALS"

### Actual behaviour

Button was "BROWSE OPEN PROPOSALS"

### Screenshots

vote_on_proposals_button

#### Device or machine

Ubuntu 19.10

#### Additional info
No change after restarting Bob.

[alice-bisq.log](https://github.com/bisq-network/bisq/files/3922928/bisq.log)
[bob-bisq.log](https://github.com/bisq-network/bisq/files/3922931/bisq.log)

ghost:
Fixes #3744

Replacing `ChangeListener` with `EasyBind.Subscription` causes immediate invoke change at the time of binding creation, which fixes posted issue

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/dao/governance/GovernanceView.java
+++ b/desktop/src/main/java/bisq/desktop/main/dao/governance/GovernanceView.java
@@ -46,7 +46,8 @@ import javafx.scene.control.ToggleGroup;
 import javafx.scene.layout.AnchorPane;
 import javafx.scene.layout.VBox;
 
-import javafx.beans.value.ChangeListener;
+import org.fxmisc.easybind.EasyBind;
+import org.fxmisc.easybind.Subscription;
 
 import java.util.Arrays;
 import java.util.List;
@@ -67,7 +68,7 @@ public class GovernanceView extends ActivatableView<AnchorPane, Void> implements
     private AnchorPane content;
 
     private Class<? extends View> selectedViewClass;
-    private ChangeListener<DaoPhase.Phase> phaseChangeListener;
+    private Subscription phaseChangeSubscription;
     private ToggleGroup toggleGroup;
 
     @Inject
@@ -89,13 +90,6 @@ public class GovernanceView extends ActivatableView<AnchorPane, Void> implements
             loadView(selectedViewClass);
         };
 
-        phaseChangeListener = (observable, oldValue, newValue) -> {
-            if (newValue == DaoPhase.Phase.BLIND_VOTE)
-                open.setLabelText(Res.get("dao.proposal.menuItem.vote"));
-            else
-                open.setLabelText(Res.get("dao.proposal.menuItem.browse"));
-        };
-
         toggleGroup = new ToggleGroup();
         List<Class<? extends View>> baseNavPath = Arrays.asList(MainView.class, DaoView.class, GovernanceView.class);
         dashboard = new MenuItem(navigation, toggleGroup, Res.get("shared.dashboard"),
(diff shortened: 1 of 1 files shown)


desktop/src/main/java/bisq/desktop/main/dao/governance/GovernanceView.java
```diff
@@ -142,7 +136,9 @@ protected void activate() {
     @SuppressWarnings("Duplicates")
     @Override
     protected void deactivate() {
-        daoFacade.phaseProperty().removeListener(phaseChangeListener);
+        if (phaseChangeSubscription != null)
+            phaseChangeSubscription.unsubscribe();
```
ghost, line 140: As we had issues in the past concerning using no brackets for one line conditionals we decided to use always {}

desktop/src/main/java/bisq/desktop/main/dao/governance/GovernanceView.java
```diff
@@ -177,6 +173,13 @@ private void loadView(Class<? extends View> viewClass) {
         else if (view instanceof ProposalsView) toggleGroup.selectToggle(open);
         else if (view instanceof VoteResultView) toggleGroup.selectToggle(result);
     }
+
+    private void setupOpenMenuButtonText(DaoPhase.Phase phase) {
+        if (phase == DaoPhase.Phase.BLIND_VOTE)
+            open.setLabelText(Res.get("dao.proposal.menuItem.vote"));
+        else
+            open.setLabelText(Res.get("dao.proposal.menuItem.browse"));
+    }
```
ghost: As we had issues in the past concerning using no brackets for one line conditionals we decided to use always {}

ghost (changes requested):

desktop/src/main/java/bisq/desktop/main/dao/governance/GovernanceView.java
ghost: Fixed

desktop/src/main/java/bisq/desktop/main/dao/governance/GovernanceView.java
ghost: Fixed

ghost:
@xyzmaker123 while testing in Bisq prior to this PR I'm unable to recreate the problem.  


- Start localnet cluster
- Create proposal to change parameter
- Advance blocks to blind vote
- See asymmetrical menus from the proposal creator and the voter



Both menu buttons immediately change to `VOTE ON PROPOSALS` when the block advances to the blind vote stage.
I'd like to recreate the issue in order to test the fix.  Please advise.

Screenshot:
image

ghost:
@jmacxx You need to be on another view during advancing blocks to blind vote and then switch to DAO->Governance.

ghost (approved):
ACK

Regtest Testing note: the DAO facade `phase` property stays as `Blind vote` for 1 block into the break period.  But it seems minor and does not really cause any trouble except the button says `vote on proposals` for one extra block actually in the break period.  Possibly a minor bug in the DAO.

ripcurlx:
As this is touching the DAO code it is necessary to get a ACK from @ManfredKarrer as well prior merging.

ghost:
@ripcurlx It doesn't touch any domain code - only simple switch from `ChangeListener` to `EasyBind.Subscription` on UI level.

ripcurlx:
> As this is touching the DAO code it is necessary to get a ACK from @ManfredKarrer as well prior merging.

Yes, I'm aware of that, but we had cases in the past where also UI code caused some issues, so this is just an extra safety net. I'll try to ping @ManfredKarrer for a review.

ManfredKarrer (approved):
utACK

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/6324#pullrequestreview-1114520718

PR #6335, opened 2022-08-22
Remove sn2bisqad7... Seeednode

Emzy:
Talked to Miker and confirmed that he has no time to fix this Seednode.
So we should remove it.


ripcurlx (approved):
utACK

PR #6337, opened 2022-08-24
Fix bugs in BSQ Supply screen

ghost:
@pazza83 asked me to investigate bugs in the BSQ Supply screen, primarily that the total supply seems incorrect.  Here are the results, and associated fixes.  tldr; _With this change, the Total BSQ Supply shown on this chart matches exactly the Total available BSQ shown on the Dashboard screen, (i.e. the UTXO set)._

### 1. Date range slider does not work

Bring up the BSQ Supply page.  Click and drag the left or right date range sliders and notice that they do not let you change the boundaries of the data shown in the chart, nor do they display human readable dates.

### 2. Total BSQ supply is calculated incorrectly

In the BSQ Supply page enable the `Total BSQ supply` toggle button and check the values by hovering over each data point in the monthly chart.  The latest value should match what is shown in Dashboard `Total available BSQ` but currently shows 3742605 instead of 4194070.  Total supply should be genesis value plus **cumulative** supply changes to date.

### 3. Periods with zero issuance are not handled correctly

Cycle 20 (December 2020) had 0 issuance, 87K burnt.  Change in supply should be -87K but instead is calculated as +87K. <sup>[1]</sup>

### 4. Miscellaneous burnt BSQ is not taken into account

A reconciliation <sup>[1]</sup> showed that more BSQ is burnt than reported in the BSQ Supply page.  This can be attributed to burns from participating in the DAO via proposals, requests, voting, and asset listing fees.  The difference currently stands at about 5K more BSQ burnt than reported on the screen.  There should also be an option to view these miscellaneous burnt BSQ amounts.

### 5. Historical data that applies to mainnet only

There exists a table of hardcoded historical values for compensation and reinbursement requests prior to them being separated.  This data only applies to mainnet, and should not be included for regtest environments.

---

[1] This spreadsheet was developed to aid the analysis above:  [bsqAnalysisBurning.xls](https://github.com/bisq-network/bisq/files/9410860/bsqAnalysisBurning.xls)

---


ripcurlx:
As agreed this PR needs an ACK from @ManfredKarrer to get merged. ☝️

ghost:
I've applied the code review suggestions, thanks!

ripcurlx:
> I've applied the code review suggestions, thanks!

Could you maybe rebase and force push it again? It seems to be some git lfs issue right now in the CI.

ghost:
Rebased.

ripcurlx (dismissed):
utACK

ripcurlx:
@jmacxx Please sign your commit messages - thanks!

ghost:
Oof.. rebased and signed. :sweat_smile:

ripcurlx (approved):
utACK

PR #6346, opened 2022-09-02
[WIP] Provide API trading statistics

ghubstan:
## Why API trading stats should be collected

The decision to approve the collection of API stats will weigh the cost of the coding, and network & memory overhead, against the need to know if the API is being used, and how much.  At the moment, we know little of API usage.

## Implementation summary

To summarize the impact of API stats collection: it adds two pieces of information to Bisq v1 storage and its network:

- Boolean field `is_maker_api_user` is added to Offer model and proto msg (was offer created by API).

- Boolean field `is_taker_api_user` is added to Trade model and proto msg (was offer taken by API).

- A new domain class `ApiTradeStatistics` was added to core, plus classes `ApiTradeStatisticsStore`,  `ApiTradeStatisticsStorageService`, and `ApiTradeStatisticsManager` (analogs of domain & store management classes for `TradeStatistics3`).

- Lifecyle for the new store and manager singletons is handled the same way as it is for `TradeStatistics3`.

## Who will use API trading stats

Bisq team members are the intended users of the first cut of this feature.  If the standard `--dumpStatistics=true` option is passed to the `bisq-daemon` and `bisq-desktop` startup scripts, API trading stats will be dumped to the appDataDir's db storage directory every time the API was used to by either side of a new trade, and the final dump during bisq shutdown.

### API trade statistics format

At this early stage, I chose to skip the work on dumping API stats as Json (and implementing code to consume it).  The dumped CSV data can be sliced and diced in many ways after importing them into spreadsheet apps, and this will give ideas on how to report API trading stats in API clients and the desktop.

When API stats are requested or dumped, `ApiTradeStatistics` is joined to `TradeStatistics3` (by `hash`), and the resulting CSV rows file will contain two flags: `is_maker_api_user`, `is_taker_api_user`, followed by the `TradeStatistics3` field values.

## How API stats are collected, stored, and broadcast

To summarize, just like `TradingStatistics3` are collected, stored, and broadcast, with the following differences:

- `ApiTradeStatistics` payloads are broadcast only after a BSQ swap buyer publishes a `TradeStatistics3` payload (see `PublishTradeStatistics`), and (a) the API was used to create the offer, (b) the API was used to take the offer, or (c) the API was used to make and take the offer.  Same is true for fiat and XMR trades (see `SellerPublishesTradeStatistics`).

- Each new `ApiTradeStatistics` instance contains the matching trade's `TradeStatistics3.hash` plus a byte determining how the API was used in the trade (by maker, taker, or both).

- When statistics are dumped, each `ApiTradeStatistics` object is _lazily_ joined to its matching `TradeStatistics3` on the hash, and each `TradeStatistics3`'s field values are written to CSV.

##  How much data is added to the network and storage

An `ApiTradeStatistics` payload on the wire is 25 bytes, roughly half that of a `TradeStatistics3` payload.

- 20 bytes are used for the hash used to join on its matching `TradeStatistics3`.
- 1 byte determines whether the offer was created by the API, taken by the API, or both.
- 4 bytes are protobuf object overhead

Any number of stored `ApiTradeStatistics` protos will be ~ 1/2 the size of an equal number of stored TradeStatistics3 protos.

## Testing

- `ApiTradeStatistics` serialization unit test:  `ApiTradeStatisticsTest`.

- API trade statistics store read/write (integration) test: `ApiTradeStatisticsIntegrationTest`.

- I am testing `ApiTradeStatisticsManager` via API regtest trade simulation runs, and `apitest` cases.

## TODO

- Implement `ApiTradeStatisticsManager.maybeRepublishApiTradeStatistics(...)`.

- Discuss how historical storage snapshots are created with @chimp1984 and/or @ripcurlx.  (It is still a bit magical to me.) 
 Implement what is needed to support it.

- Create a new API method `getapistats` with appropriate request params to dump stats to csv (and/or json?)

- Confer with @chimp1984 about a backward compatibility issue at the bottom of  core's `OfferPayload` class.  See its static class `JsonSerializer`.  The question is if the new field `isMakerApiUser` should NOT be serialized, to maintain backward compatibility.

### DIFF
--- a/apitest/src/main/java/bisq/apitest/config/ApiTestConfig.java
+++ b/apitest/src/main/java/bisq/apitest/config/ApiTestConfig.java
@@ -76,6 +76,7 @@ public class ApiTestConfig {
     static final String CONFIG_FILE = "configFile";
     static final String ROOT_APP_DATA_DIR = "rootAppDataDir";
     static final String API_PASSWORD = "apiPassword";
+    static final String DUMP_STATISTICS = "dumpStatistics";
     static final String RUN_SUBPROJECT_JARS = "runSubprojectJars";
     static final String BISQ_APP_INIT_TIME = "bisqAppInitTime";
     static final String SKIP_TESTS = "skipTests";
@@ -109,6 +110,7 @@ public class ApiTestConfig {
     public final String bitcoinRpcPassword;
     // Daemon instances can use same gRPC password, but each needs a different apiPort.
     public final String apiPassword;
+    public final boolean dumpStatistics;
     public final boolean runSubprojectJars;
     public final long bisqAppInitTime;
     public final boolean skipTests;
@@ -206,6 +208,12 @@ public class ApiTestConfig {
                 parser.accepts(API_PASSWORD, "gRPC API password")
                         .withRequiredArg()
                         .defaultsTo("xyz");
+
+        ArgumentAcceptingOptionSpec<Boolean> dumpStatisticsOpt =
+                parser.accepts(DUMP_STATISTICS, "Dump node statistics")
+                        .withRequiredArg()
+                        .ofType(Boolean.class)
+                        .defaultsTo(false);
 
         ArgumentAcceptingOptionSpec<Boolean> runSubprojectJarsOpt =
                 parser.accepts(RUN_SUBPROJECT_JARS,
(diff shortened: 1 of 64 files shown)


chimp1984:
Why create a new data structure and not just add it to the `TradeStatistics3`? There is the hashmap for additional data to not break backward compatibility.

ghubstan:
> Why create a new data structure and not just add it to the `TradeStatistics3`? There is the hashmap for additional data to not break backward compatibility.

I did not use the hashmap because of the field comment `Should be only used in emergency case...`  We talked about this about a month ago via DM.  Can you check my response there?

PR #6348, opened 2022-09-03
Refresh all avatars upon setting a peer's tag

### LINKED ISSUE #4660: Tagging peer only updates one row instead of all affected rows before restart
### Description

Tagging a peer only applies that tag to the selected row of that peer instead of all their rows, until a restart is done.

#### Version
1.3.9

### Steps to reproduce
Tag an untagged peer who has multiple rows on the offer sheet. 

### Expected behaviour
All rows of this peer would update to reflect the new tag.

### Actual behaviour
Only the selected row of this peer updates to reflect the new tag.

### Screenshots
Before restarting:
before_restart

After restarting (now that all rows are reloaded):
after_restart

#### Device or machine
Ubuntu 20.04

#### Additional info
It's a very minor issue, but I can attempt to solve it as my first issue to get familiar with Bisq repository.

Android-X13:
Fixes #4660

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBookView.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBookView.java
@@ -28,7 +28,7 @@ import bisq.desktop.components.AutocompleteComboBox;
 import bisq.desktop.components.ColoredDecimalPlacesWithZerosText;
 import bisq.desktop.components.HyperlinkWithIcon;
 import bisq.desktop.components.InfoAutoTooltipLabel;
-import bisq.desktop.components.PeerInfoIcon;
+import bisq.desktop.components.PeerInfoIconMap;
 import bisq.desktop.components.PeerInfoIconTrading;
 import bisq.desktop.components.TitledGroupBg;
 import bisq.desktop.main.MainView;
@@ -113,7 +113,6 @@ import javafx.util.Callback;
 import javafx.util.StringConverter;
 
 import java.util.Comparator;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 
@@ -154,7 +153,8 @@ abstract public class OfferBookView<R extends GridPane, M extends OfferBookViewM
     private static final int SHOW_ALL = 0;
     private Label disabledCreateOfferButtonTooltip;
     protected VBox currencyComboBoxContainer;
-    private final Map<String, PeerInfoIcon> avatarMap = new HashMap<>();
+    @SuppressWarnings("MismatchedQueryAndUpdateOfCollection")
+    private final PeerInfoIconMap avatarMap = new PeerInfoIconMap();
 
     ///////////////////////////////////////////////////////////////////////////////////////////
     // Constructor, lifecycle
@@ -1284,7 +1284,7 @@ abstract public class OfferBookView<R extends GridPane, M extends OfferBookViewM
                                             model.accountAgeWitnessService,
                                             useDevPrivilegeKeys);
                                     String key = offer.getId();
-                                    peerInfoIcon.setAvatarMapAndKey(avatarMap, key);
+                                    avatarMap.put(key, peerInfoIcon);
                                     setGraphic(peerInfoIcon);
                                 } else {
                                     setGraphic(null);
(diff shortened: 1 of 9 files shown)


desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBookView.java
```diff
@@ -1289,6 +1281,32 @@ public void updateItem(final OfferBookListItem newItem, boolean empty) {
         return column;
     }
 
+    private PeerInfoIconTrading findOrCreateAvatar(Integer tableRowId, Offer offer) {
```
ripcurlx: Did you intend to find the avatar and create it only if not found in the map, as you create it all the time right now?

Android-X13: @ripcurlx no - I would use a name like `createAvatar`, but I kept this method name to keep it similar with the one in DisputeView.
Do you want me to change it (and maybe the one in DisputeView too)?

ripcurlx: Yes, please change the other one as well, as it looks like as if some code parts are missing based on the method name.

Android-X13: OK
I will commit tomorrow because I want to test something first

Android-X13: I improved it a bit:
Changed the key that is stored in the map to only contain the offer's ID. Row number + maker's address are not sufficient because the offers keep changing rows/positions (it still worked fine previously but now the map isn't increasing unnecessarily).
The map is cleared on deactivate() to avoid holding entries that may not exist anymore (needed for OfferBookView which keeps changing frequently, not so much for DisputeView).

ripcurlx (changes requested):
NACK - I reviewed the code and tested it quickly on Regtest to confirm my assumption that you also need to do some changes in UnconfirmedBsqSwapsView, ClosedTradesView and PendingTradesView which use the PeerInfoIconTrading which extends PeerInfoIcon

Android-X13:
I rewrote this using a different approach to eliminate the interface and the multiple identical overrides (since it's now used by many classes). Of course I can revert to using the old approach if you prefer

ripcurlx:
How about having something like a PeerInfoIconMap that holds all PeerInfoIcons, listens to a save event of each PeerInfoIcon and calls refreshTag on all instances in the internal Map accordingly. Controlling behavior of other PeerInfoIcons from within a PeerInfoIcon doesn't feel right to me. WDYT?

Android-X13:
No problem, re-re-wrote this :sweat_smile:
Maybe I misinterpreted you, but I suppose you didn't mean one universal avatar map holding all PeerInfoIcons from all views from different sections because I think that would introduce other problems.
I'm using one custom map in each view like before, which now listens for tag change events, close to what you requested.
The only thing I'm not sure about is where to place the new PeerInfoIconMap class in the vast directory structure of this project. I've put it in components along with the other PeerInfoIcon classes although it's not exactly a component...

ripcurlx:
> No problem, re-re-wrote this 😅 Maybe I misinterpreted you, but I suppose you didn't mean one universal avatar map holding all PeerInfoIcons from all views from different sections because I think that would introduce other problems. I'm using one custom map in each view like before, which now listens for tag change events, close to what you requested. The only thing I'm not sure about is where to place the new PeerInfoIconMap class in the vast directory structure of this project. I've put it in components along with the other PeerInfoIcon classes although it's not exactly a component...

Yes, that is what I was thinking about 👍 . Location of the extended HashMap is fine IMO. Maybe we could have put all PeerInfoIcon components into one sub-package, but I think it is fine as it is right now. I'm doing one quick Regtest test and will merge it afterwards.

ripcurlx (dismissed):

ripcurlx:
@Android-X13 Your last commits haven't been signed somehow. Could you update them and force push it and I'll merge this PR. Thanks!

ripcurlx:
@Android-X13 Your last commits haven't been signed somehow. Could you update them and force push it and I'll merge this PR. Thanks!

ripcurlx:
@Android-X13 I don't know why, but your first two commits are still not signed. You can see it if the 'Verified' label is not shown next to the commit.

ripcurlx (approved):
ACK

PR #6352, opened 2022-09-08
Prevent NPE in Open Proposals when last column is empty

Android-X13:
DAO > GOVERNANCE > BROWSE OPEN PROPOSALS > Click on the last column's header, when column is empty


ripcurlx (approved):
utACK - Code changes are looking fine, but we'll need to wait for an ACK of @ManfredKarrer as it touches the DAO package.

ManfredKarrer (approved):
utACK

Issue #6355, opened 2022-09-10
BSQ balance after swap doesn't reconcile properly


pmartincalvo:
### Description

After executing a BSQ Swap (I buy BSQ, sell BTC) on a daemon with a clean transaction history, the BSQ balance doesn't add up to the purchased BSQ.

#### Version

1.9.4 daemon + cli

### Steps to reproduce

---
#### Note
`bisq_cli` is an alias for `java -jar cli.jar --password=mypassword`.

---

I checked available offers with `bisq_cli getoffers --direction=BUY --currency-code=BSQ`. I decided to pick the following one:

```
Buy/Sell            Price in BTC for 1 BSQ           BTC(min - max)    BSQ(min - max)  Payment Method  Creation Date (UTC)   ID
Sell BSQ (Buy BTC)              0.00004680               0.00060000             12.82  BSQ Swap        2022-07-20T20:49:31Z  PJNjEX-0461e540-c6ed-49f7-b1f7-8584f8aefb69-192  
```

Took it successfully. I'm paying the trade fee with BTC since I have no BSQ:
```
bisq_cli takeoffer --offer-id=PJNjEX-0461e540-c6ed-49f7-b1f7-8584f8aefb69-192 --fee-currency=BTC

trade PJNjEX-0461e540-c6ed-49f7-b1f7-8584f8aefb69-192 successfully taken
```

After the trade is finished, I go and check my BSQ balance

```
bisq_cli getbalance --currency-code=bsq

Available Confirmed Balance  Unverified Balance  Unconfirmed Change Balance  Locked For Voting Balance  Lockup Bonds Balance  Unlocking Bonds Balance
                      12.74                0.00                        0.00                       0.00                  0.00                     0.00
```

And here is the surprise. If I took a trade to buy 12.82, why do I only have 12.74?

Note that this is the only trade this wallet has executed.

### Expected behaviour

Obtaining 12.82 BSQ after the trade.

### Actual behaviour

Obtaining 12.74 BSQ after the trade.

### Screenshots

NA

#### Device or machine

Debian 11 on a VPS.

#### Additional info

NA

ghost:
Trading fees are paid from the BSQ purchase.  
Current taker fees are 126.02 BSQ per BTC traded.
0.0006 * 126.02 = 0.08 BSQ

In the GUI the fee is shown before committing to the trade.

pmartincalvo:
Thanks for your comment. I still don't fully understand how this works:
[...]

PR #6357, opened 2022-09-11
API: Fix BSQ buyer's displayed trade fee for gettrade(s) methods

ghubstan:
API clients looking at their BSQ swap trades -- BSQ buyers specifically -- were seeing a 0.00 BSQ trade fee, due to an incorrect assumption BSQ sellers paid the entire BSQ trade fee for both sides.  BSQ Swap execution via API has been working just like the GUI (same code-base), but the reported trade fee for the BSQ buy side was wrong.  This change shows the correct BSQ trade fee.

Partially addresses issue https://github.com/bisq-network/bisq/issues/6355

Based on https://github.com/bisq-network/bisq/pull/6356	and branch `doc-takeoffer-fee-curnency-is-ignored-for-swaps`.


ripcurlx (approved):
utACK

PR #6360, opened 2022-09-15
Not use git lfs anymore

ripcurlx:
We introduced git lfs in PR https://github.com/bisq-network/bisq/pull/4114

Back then we always updated all data stores not having diffs in place for trade statistics and DAO blocks as we do have now.

There are a couple of reasons for me to go back to track the binaries as before:
- One step less for devs to build the repo
- No more purchase of data packs on GitHub. We are already paying €360/year for 3,600 GB Bandwidth and we already consumed nearly half of it in this cycle (resets in 244 days)
- We don't need to have any payment information on GitHub
- We handle most of the data store diffs now ourselves anyways


cbeams (approved):
Concept ACK. @ripcurlx and I discussed this offline. This change unfortunately means revisions of Bisq between the merging of #4114 in commit 2e3e811 on Apr 29, 2020 and the revision prior to this PR being merge will not be able to be built anymore. However, given all the considerations on the table, it seems like the right trade-off.

PR #6361, opened 2022-09-15
Not use platform specific binary signing and notarization

ripcurlx:
Based on the reasoning layed out in https://github.com/bisq-network/bisq/discussions/6341 the introduction of platform specific signing and notarization did improve
user UX during installation, but also introduced a kill switch for Bisq.

This PR removes platform specific binary signing and notarization.

For this to work we need to add information on the wiki and the download page to guide users how to install Bisq on macOS and Windows despite the security warnings.

### DIFF
--- a/desktop/package/package.gradle
+++ b/desktop/package/package.gradle
@@ -1,7 +1,6 @@
-import java.time.LocalDateTime
 import org.apache.tools.ant.taskdefs.condition.Os
 
-import static groovy.io.FileType.*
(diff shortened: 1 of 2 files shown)


cd2357 (approved):
LGTM

bisq-github-admin-3 (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/6361#pullrequestreview-1111977013

PR #6362, opened 2022-09-19
Set the minimum payout at mediation to be 2.5% of trade amount

ghost:
If 2.5% is less than the original minimum 0.0005 BTC payout, then fall back to 0.0005 BTC.

2.5% was requested by @refund-agent2 and the support team in order to incentivize acceptance of mediation proposals.


ripcurlx (approved):
utACK - code changes are looking fine

PR #6364, opened 2022-09-28
Remove c4talys7 seednodes and readd wiz seednodes

Emzy:
c4talys7 is stepping down as Seednode operator and Wiz offered to keep his Seednodes running.

Wiz's seednodes are looking fine for now. We will try to fix his OOM problems on his nodes. 
For me it seems to be clearly related to the Tor DDoS. And only happens then.


ripcurlx (approved):
utACK

PR #6369, opened 2022-10-05
Offers from a banned node should not be displayed

ghost:
There's someone spamming using fake F2F offers (25 of them across different currencies).
They are banned (offers cannot be taken), but the offers still show in the market & offer book
This PR addresses that.


ripcurlx (approved):
utACK

PR #6378, opened 2022-10-17
Update missing hashes for tor 0.7.5 update

ripcurlx:
Not sure why this wasn't added when I ran it last time, but it is missing in the verification metadata file right now.


ghost (approved):
ACK.  This is why new PRs have failing CI.

cd2357 (approved):
utACK

There were some major changes to the workflow (of upgrading tor) a few months ago. I don't fully understand what has changed and why, but dependencies and poms look different for both tor and netlayer.

When I reviewed the previous PR I went back and I tried to follow these changes, commit by commit, to understand the new tor / netlayer pom structure. But after a few hours I stopped, as there were too many changes in too many places at once.

So I don't think I can really ACK or NACK in a meaningful way. I can check general syntax and see if it works, but thats basically it ATM.

PR #6383, opened 2022-10-18
Include option for non-URI addresses in QR codes

### LINKED ISSUE #6377: QR Code Generator
Hello,

The QR Code Generator on BISQ is using this format to generate a QR Code:

**_"bitcoin:address?amount=0&label=Fund your wallet"_**

However the two different external wallets I use will not accept this format. After talking to their support, they need to see it in this format:

**_"bitcoin:address"_**

Apparently some wallets will take the expanded QR code. However ALL external wallets should take the simplified code that has just the wallet address in it.

Is it possible to change the format that BISQ uses to generate the QR generator to not have the extra parameters?

Android-X13:
Resolves #6377

Includes an option in Preferences to control the way bitcoin addresses are encoded (either in bitcoin URI format or in plain non-URI format) and affects all views that generate/display a QR code.

- When the bitcoin URI format is selected (default), addresses are encoded like before: `bitcoin:address?amount=<AMOUNT>&label=<LABEL>`. In DepositView however the `amount` parameter will only be included if the amount is larger than zero. This will fix the particular issue mentioned above (`amount=0` is probably wrong because it can be interpreted as a request for 0 BTCs, so literally a request for no transaction). It won't make any difference in other views like TakeOfferView where the `amount` parameter is present because it's always greater than 0.

- When the plain format is selected, addresses are encoded simply as: `address`. This will allow all external wallets to read the generated QRs if they're facing issues (some can only read simple `bitcoin:address` URIs without any extra parameters, some cannot handle the URI scheme format altogether even when not having parameters, but all can read plain addresses).

I also added a simple text formatter to the amount text field in DepositView, mainly to prevent the exception-like message that pops up when you input a negative number:

negative

The textfield will now only accept positive integer or positive dot-decimal numerical values.

### NOTE
After reading through BIP21 and on the topic of coin control, in my opinion a more "correct" format of Bisq's bitcoin URIs (ie. one that strictly adheres to the standard) would be the following:

For URIs generated in DepositView:
`bitcoin:<ADDRESS>?amount=<AMOUNT>&label=Bisq&message=Fund wallet` (instead of `&label=Fund Bisq wallet`)

For URIs generated when creating/taking offer:
`bitcoin:<ADDRESS>?amount=<AMOUNT>&label=Bisq&message=Trade with ID XXXXX` (instead of `&label=Bisq trade with ID X`)

Because the `label` of BIP21 seems to denote names/entities instead of payment reasons. This would allow the transactions to be tagged as intended by the few wallets that can parse the `label` parameter and that can also apply coin control.

I didn't change the format, just noting this here.

### DIFF
--- a/core/src/main/java/bisq/core/user/Preferences.java
+++ b/core/src/main/java/bisq/core/user/Preferences.java
@@ -830,6 +830,11 @@ public final class Preferences implements PersistedDataHost, BridgeAddressProvid
         requestPersistence();
     }
 
+    public void setUseBitcoinUrisInQrCodes(boolean value) {
+        prefPayload.setUseBitcoinUrisInQrCodes(value);
+        requestPersistence();
+    }
+
 
     ///////////////////////////////////////////////////////////////////////////////////////////
     // Getter
@@ -1147,6 +1152,8 @@ public final class Preferences implements PersistedDataHost, BridgeAddressProvid
         void setNotifyOnPreRelease(boolean value);
 
         void setUseFullModeDaoMonitor(boolean value);
+
+        void setUseBitcoinUrisInQrCodes(boolean value);
 
         void setClearDataAfterDays(int value);
 
--- a/core/src/main/java/bisq/core/user/PreferencesPayload.java
+++ b/core/src/main/java/bisq/core/user/PreferencesPayload.java
@@ -140,6 +140,7 @@ public final class PreferencesPayload implements PersistableEnvelope {
     private boolean denyApiTaker;
     private boolean notifyOnPreRelease;
     private boolean useFullModeDaoMonitor;
+    private boolean useBitcoinUrisInQrCodes = true;
 
     ///////////////////////////////////////////////////////////////////////////////////////////
     // Constructor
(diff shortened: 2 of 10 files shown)


ghost:
Rather than yet another obscure setting in PreferencesView that no one will know about, can the toggle be shown along with the QR code, enabling a user to turn on/off the feature where it is used?  I would also suggest that it not need to be persistable.

Android-X13:
@jmacxx Initially I also tried to figure other ways to implement this... but the problem is that QR codes are displayed in more than one view with different controls around them and it could clutter up the UI, so a toggle along with the QR might look somewhat okay in one view but maybe not so fine in another:

qr1

qr2

I had also considered another way, i.e. making the large QR code window to be a tabbed one, with each tab showing a different QR (one for the URI, one for the plain address). Or it could be like the current one but also contain the toggle there, which is probably better. But in both scenarios the problem would be that one would not be able to switch formats in the main view but only inside the large QR window after clicking on it. And when switching format inside the window, should that be also reflected in the small QR in the main view? Also if the user has typed an amount in the text field of DepositView, when switching to the non-URI format from inside the window that detail would need to be ignored, whereas now it's more straightforward: if the setting is selected in non-URI format, the amount text field is disabled (and enabled otherwise).

Another way would be to have a right-click context menu on the view's QR code, but this wouldn't be so straightforward as a toggle, and no explanation other than the menu item's text would be shown.

This is why I went with the current approach with a global setting and I tried to make it very obvious by applying a tooltip with a small explanation:

setting

Of course if a toggle was to always be shown along with the QR instead (either in the main view or in the large QR window) then I agree that it shouldn't be persistable, only to be enabled by default i.e. first show the bitcoin URI format like previously. But consider the above

Android-X13:
Resolved conflicts

ghost (approved):
ACK

alejandrogarcia83 (approved):
utACK

PR #6384, opened 2022-10-19
[v1.9.6] Resize action column if it contains own offers

ripcurlx:
I think it would be better not to have truncation with our default with when having own offers.

Bildschirmfoto 2022-10-19 um 13 24 45
Bildschirmfoto 2022-10-19 um 13 24 54


ripcurlx:
@jmacxx WDYT?

ghost (approved):
utACK

PR #6396, opened 2022-10-30
Store the build's commit hash in Jarfile manifest

ghost:
The seednode monitor displays a misleading commit hash, just a hardcoded string which never changes.

This PR fixes that by obtaining the git HEAD commit hash at build time, storing it in the Jarfile manifest information.
Seednodes then report the correct commit hash so that the monitor will indicate what commit they are running.
This is useful to know what source the seednode was built from.

e.g:

image

### DIFF
--- a/common/src/main/java/bisq/common/app/Version.java
+++ b/common/src/main/java/bisq/common/app/Version.java
@@ -17,8 +17,12 @@
 
 package bisq.common.app;
 
+import java.net.URL;
+
 import java.util.Arrays;
 import java.util.List;
+import java.util.jar.Attributes;
+import java.util.jar.Manifest;
 
 import lombok.extern.slf4j.Slf4j;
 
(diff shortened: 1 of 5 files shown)


common/src/main/java/bisq/common/app/Version.java
```diff
@@ -135,6 +139,16 @@ public static void printVersion() {
                 '}');
     }
 
+    public String readCommitHash() {
```
HenrikJannsen, line 142: Maybe better to use a static method.

core/src/main/java/bisq/core/network/p2p/inventory/GetInventoryRequestHandler.java
```diff
@@ -151,7 +150,7 @@ public void onMessage(NetworkEnvelope networkEnvelope, Connection connection) {
 
             // node
             inventory.put(InventoryItem.version, Version.VERSION);
-            inventory.put(InventoryItem.commitHash, RequestInfo.COMMIT_HASH);
+            inventory.put(InventoryItem.commitHash, new Version().readCommitHash());
```
HenrikJannsen, line 153: A static method would be more natural here.

HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6404, opened 2022-11-03
Improve dispute validation

HenrikJannsen:
Contains more validation for dispute data and further preparations for https://github.com/bisq-network/proposals/issues/390 implementation

PR #6405, opened 2022-11-04
Distribute burningman role to contributors who burned bsq

HenrikJannsen:
Implements https://github.com/bisq-network/proposals/issues/390

Based on https://github.com/bisq-network/bisq/pull/6404 

Starts at commit [33c08c7](https://github.com/bisq-network/bisq/pull/6405/commits/33c08c7baf6ff173f60201da26f32d34e962cb95)


HenrikJannsen:
Rebased on master

HenrikJannsen:
From my side its complete now. I continue on another new PR for accounting of the received BTC but that should not have implications on that code. I will do a final in depth testing once all is complete but I tested that PR to some good extent.

Please do not merge before DAO voting has accepted the proposal.

core/src/main/java/bisq/core/support/dispute/refund/RefundManager.java
```diff
@@ -238,4 +252,63 @@ public void onDisputeResultMessage(DisputeResultMessage disputeResultMessage) {
...
+            }
+            if (!takerFeeTxFoundAtInputs) {
+                takerFeeTxFoundAtInputs = fundingTxId.equals(takerFeeTx.getTxId().toString());
+            }
+        }
+        checkArgument(makerFeeTxFoundAtInputs, "makerFeeTx not found at depositTx inputs");
+        checkArgument(takerFeeTxFoundAtInputs, "takerFeeTx not found at depositTx inputs");
+        checkArgument(depositTx.getInputs().size() >= 2,
+                "DepositTx must have at least 2 inputs");
+        checkArgument(depositTx.getOutputs().size() == 1,
```
sqrrm: I think the can be change in the deposit tx outputs if taker didn't take the max amount of a seller as maker that was advertising a range.

HenrikJannsen:
Proposal got accepted by DAO voting: https://github.com/bisq-network/proposals/issues/390#issuecomment-1319267561

HenrikJannsen:
Before merging please coordinate. There will be another PR and the activation data and some other data need to be finalized before merge.

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -2734,6 +2734,11 @@ disputeSummaryWindow.reason=Reason of dispute
 disputeSummaryWindow.tradePeriodEnd=Trade period end
 disputeSummaryWindow.extraInfo=Extra information
 disputeSummaryWindow.delayedPayoutStatus=Delayed Payout Status
+disputeSummaryWindow.requestingTxs=Requesting blockchain transactions from block explorer...
+disputeSummaryWindow.requestTransactionsError=Requesting the 4 trade transactions failed. Error message: {0}.\n\n\
+  Please verify the transactions manually before closing the dispute.
+disputeSummaryWindow.delayedPayoutTxVerificationFailed=Verification of the delayed payout transaction failed. Error message: {0}.\n\n\
+  Please do not make the payout but get in touch with developers to clearify the case.
```
sqrrm:
``` suggestion
  Please do not make the payout but get in touch with developers to clarify the case.
```

core/src/main/java/bisq/core/support/dispute/DisputeValidation.java
```diff
@@ -46,6 +56,72 @@
...
+            checkArgument(contract.getOfferPayload().getId().equals(dispute.getTradeId()), "Invalid tradeId");
+            checkArgument(dispute.getContractAsJson().equals(JsonUtil.objectToJson(contract)), "Invalid contractAsJson");
+            checkArgument(Arrays.equals(Objects.requireNonNull(dispute.getContractHash()), Hash.getSha256Hash(checkNotNull(dispute.getContractAsJson()))),
+                    "Invalid contractHash");
+
+            Optional<Transaction> depositTx = dispute.findDepositTx(btcWalletService);
+            if (depositTx.isPresent()) {
+                checkArgument(depositTx.get().getTxId().toString().equals(dispute.getDepositTxId()), "Invalid depositTxId");
+                checkArgument(depositTx.get().getInputs().size() >= 2, "DepositTx must have at least 2 inputs");
+                checkArgument(depositTx.get().getOutputs().size() == 1, "DepositTx must have exactly 1 output");
```
sqrrm: Same here with number of outputs for depositTx when less than max amount taken.

core/src/main/java/bisq/core/util/validation/InputValidator.java
```diff
@@ -24,18 +24,25 @@
...
+
 public class InputValidator {
+    @Setter
+    public boolean allowEmpty;
 
     public ValidationResult validate(String input) {
         return validateIfNotEmpty(input);
     }
 
     protected ValidationResult validateIfNotEmpty(String input) {
```
sqrrm, line 37: This name is a bit odd now that there is allowEmpty

core/src/main/java/bisq/core/dao/DaoFacade.java
```diff
@@ -646,6 +649,24 @@ public int getNumIssuanceTransactions(IssuanceType issuanceType) {
         return daoStateService.getIssuanceSetForType(issuanceType).size();
     }
 
+    public Set<Issuance> getIssuanceSetForType(IssuanceType issuanceType) {
+        return daoStateService.getIssuanceSetForType(issuanceType);
+    }
+
+    public Map<TxOutputKey, Optional<String>> getAddressByOutputKeyMap() {
+        return daoStateService.getUnorderedTxOutputStream()
+                .collect(Collectors.toMap(BaseTxOutput::getKey, txOutput -> Optional.ofNullable(txOutput.getAddress())));
```
sqrrm: Maybe this should be cashed, although it wouldn't be current when new blocks come in

core/src/main/java/bisq/core/dao/burningman/BurningManService.java
```diff
@@ -0,0 +1,638 @@
...
+    public Map<String, BurningManCandidate> getBurningManCandidatesByName(int chainHeight) {
+        Map<String, BurningManCandidate> burningManCandidatesByName = new HashMap<>();
+        Map<P2PDataStorage.ByteArray, Set<TxOutput>> proofOfBurnOpReturnTxOutputByHash = getProofOfBurnOpReturnTxOutputByHash(chainHeight);
+
+        // Add contributors who made a compensation request
+        daoStateService.getIssuanceSetForType(IssuanceType.COMPENSATION).stream()
+                .filter(issuance -> issuance.getChainHeight() <= chainHeight)
+                .forEach(issuance -> {
+                            getCompensationProposalsForIssuance(issuance).forEach(compensationProposal -> {
+                                String name = compensationProposal.getName();
```
sqrrm, line 132: What happens when a proposal has been accepted with the name misspelled?

core/src/main/java/bisq/core/dao/burningman/BurningManService.java
```diff
@@ -0,0 +1,638 @@
...
+                                                        return txOutputs.get(1).getAddress();
+                                                    }
+                                                }))
+                                        .ifPresent(address -> {
+                                            int issuanceHeight = issuance.getChainHeight();
+                                            long issuanceAmount;
+
+                                            // There was a reimbursement for a conference sponsorship with 44776 BSQ. We remove that as well.
+                                            // See https://github.com/bisq-network/compensation/issues/498?utm_source=desktop-client&utm_medium=in-app-link&utm_campaign=language_en
+                                            boolean wasSponsorship = issuance.getTxId().equals("01455fc4c88fca0665a5f56a90ff03fb9e3e88c3430ffc5217246e32d180aa64");
```
sqrrm: I suppose this is correct, but seems not necessary since it's outside of range for compensation requests.

core/src/main/java/bisq/core/dao/burningman/BurningManService.java
```diff
@@ -0,0 +1,638 @@
...
+                                            if (wasSponsorship) {
+                                                issuanceAmount = 119400; // That was the compensation part
+                                            } else {
+                                                issuanceAmount = issuance.getAmount();
+                                            }
+                                            int cycleIndex = getCycleIndex(issuanceHeight);
+                                            // Up to cycle 15 the RefundAgent made reimbursement requests as compensation requests. We filter out those entries.
+                                            // As it is mixed with RefundAgents real compensation requests we take out all above 3500 BSQ.
+                                            boolean isReimbursementOfRefundAgent = name.equals("RefundAgent") && cycleIndex <= 15 && issuanceAmount > 350000;
```
sqrrm: This also seems unnecessary since it won't count for the 24 cycle range.

core/src/main/java/bisq/core/dao/burningman/BurningManService.java
```diff
@@ -0,0 +1,638 @@
...
+        if (issuanceHeight > currentBlockHeight)
+            throw new IllegalArgumentException("issuanceHeight must not be larger than currentBlockHeight. issuanceHeight=" + issuanceHeight + "; currentBlockHeight=" + currentBlockHeight);
+        if (currentBlockHeight < 0)
+            throw new IllegalArgumentException("currentBlockHeight must not be negative. currentBlockHeight=" + currentBlockHeight);
+        if (amount < 0)
+            throw new IllegalArgumentException("amount must not be negative. amount" + amount);
+        if (issuanceHeight < 0)
+            throw new IllegalArgumentException("issuanceHeight must not be negative. issuanceHeight=" + issuanceHeight);
+
+        double factor = Math.max(0, (issuanceHeight - firstBlockHeight) / (double) (currentBlockHeight - firstBlockHeight));
```
sqrrm: The denominator is changing in value from block to block, starting with 24 cycles + 1 block, up to 25 cycles - 1 block. The numerator stays fixed for the cycle until it's degraded one cycle. This will lead to odd step function behavior.

I think it would be better with something like
```
        double factor = Math.max(0,1 - (currentBlockHeight - issuanceHeight) / (double) (currentCycleHeight - firstBlockHeight));
```
That would result in a linear decay.

core/src/main/java/bisq/core/dao/burningman/BurningManService.java
```diff
@@ -0,0 +1,638 @@
...
+
+    // Linear decay between currentBlockHeight (100% of amount) and issuanceHeight (firstBlockOffset % of amount)
+    // Values below firstBlockHeight will use the firstBlockOffset as factor for the amount.
+    // E.g. if firstBlockOffset is 0.1 the decay goes to 10% and earlier values stay at 10%.
+    @VisibleForTesting
+    static long getDecayedAmount(long amount,
+                                 int issuanceHeight,
+                                 int currentBlockHeight,
+                                 int firstBlockHeight,
+                                 double firstBlockOffset) {
```
sqrrm: I think this argument should be removed since it's not being used, just adding confusion.

core/src/main/java/bisq/core/dao/burningman/DelayedPayoutTxReceiverService.java
```diff
@@ -0,0 +1,175 @@
...
+                .sorted(Comparator.<Tuple2<Long, String>, Long>comparing(tuple -> tuple.first)
+                        .thenComparing(tuple -> tuple.second))
+                .collect(Collectors.toList());
+        long totalOutputValue = receivers.stream().mapToLong(e -> e.first).sum();
+        if (totalOutputValue < spendableAmount) {
+            long available = spendableAmount - totalOutputValue;
+            // If the available is larger than DPT_MIN_REMAINDER_TO_LEGACY_BM we send it to legacy BM
+            // Otherwise we use it as miner fee
+            if (available > DPT_MIN_REMAINDER_TO_LEGACY_BM) {
+                receivers.add(new Tuple2<>(available, burningManService.getLegacyBurningManAddress(burningManSelectionHeight)));
```
sqrrm, line 161: The `available` will be positive whenever a burningman is excluded due to minOutputAmount filtering. It would be better to recalibrate the receiver shares after the filtering of the minOutPut. Each receiver that isn't already getting the capped would get their weighted share of `available` amount. Otherwise legacy BM will be guaranteed to keep getting some small amounts.

core/src/main/java/bisq/core/support/dispute/refund/RefundManager.java
```diff
@@ -308,4 +315,25 @@ public void verifyTradeTxChain(List<Transaction> txs) {
...
+
+        NetworkParameters params = btcWalletService.getParams();
+        for (int i = 0; i < delayedPayoutTx.getOutputs().size(); i++) {
+            TransactionOutput transactionOutput = delayedPayoutTx.getOutputs().get(i);
+            Tuple2<Long, String> receiverTuple = delayedPayoutTxReceivers.get(0);
+            checkArgument(transactionOutput.getScriptPubKey().getToAddress(params).toString().equals(receiverTuple.second),
+                    "output address does not match delayedPayoutTxReceivers address. transactionOutput=" + transactionOutput);
+            checkArgument(transactionOutput.getValue().value == receiverTuple.first,
+                    "output value does not match delayedPayoutTxReceivers value. transactionOutput=" + transactionOutput);
+        }
```
sqrrm, line 339: While this looks correct, wouldn't it be easier to just create the DPT and compare txid the same way it's done in the trade protocol?

desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
```diff
@@ -463,6 +511,30 @@ private void onSendBtc() {
...
+        byte[] opReturnData;
+        try {
+            // If preImage is hex encoded we use it directly
+            opReturnData = Hex.decode(preImageAsString);
+        } catch (Throwable ignore) {
+            opReturnData = preImageAsString.getBytes(Charsets.UTF_8);
+        }
+
+        // If too long for OpReturn we hash it
+        if (opReturnData.length > 80) {
```
sqrrm, line 524: Is 80 bytes going to be relayed?

core/src/main/java/bisq/core/dao/burningman/BurningManService.java
HenrikJannsen: Then it is a considered as a different burningman candidate. But if your app has both contribution requests you can use both. So nothing lost besider its less convenient.

core/src/main/java/bisq/core/dao/burningman/BurningManService.java
HenrikJannsen: Yes, true. Still prefer to keep it.

core/src/main/java/bisq/core/dao/burningman/BurningManService.java
HenrikJannsen: True. Maybe once we have finalized the paramaters we remove that and the other one...

core/src/main/java/bisq/core/support/dispute/refund/RefundManager.java
HenrikJannsen: That was outdated and got fixed in the meantime.

core/src/main/java/bisq/core/support/dispute/DisputeValidation.java
HenrikJannsen: That was outdated and got fixed in the meantime.

core/src/main/java/bisq/core/util/validation/InputValidator.java
HenrikJannsen: Youo mean `validateIfNotEmpty`? Yes, agree. But don't have a good suggestion for improvement. If you have please share...

core/src/main/java/bisq/core/dao/DaoFacade.java
HenrikJannsen: Was outdated. Was removed in a later commit as not used anymore.

core/src/main/java/bisq/core/dao/burningman/BurningManService.java
HenrikJannsen: I will add a TODO for now and we can decide later.

core/src/main/java/bisq/core/dao/burningman/BurningManService.java
HenrikJannsen: I will add a TODO for now and we can decide later.

core/src/main/java/bisq/core/dao/burningman/BurningManService.java
HenrikJannsen: We are doing that already because the firstBlockHeight is the currentCycleHeight - numCycles*cycleDuration.
I rename it and add a comment to make it more clear.
See cyclesInDaoStateService.getChainHeightOfPastCycle

core/src/main/java/bisq/core/dao/burningman/BurningManService.java
HenrikJannsen: Ok, will remove it.

core/src/main/java/bisq/core/dao/burningman/DelayedPayoutTxReceiverService.java
HenrikJannsen: Agree. Will change it.

core/src/main/java/bisq/core/support/dispute/refund/RefundManager.java
HenrikJannsen: The lockTime is not present for the arbitrator. We would need to store that in the dispute as well.

desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
HenrikJannsen: https://www.talkcrypto.org/blog/2016/12/30/op_return-40-to-80-bytes/

https://www.oreilly.com/library/view/mastering-bitcoin/9781491902639/ch05.html

```
OP_RETURN was initially proposed with a limit of 80 bytes, but the limit was reduced to 40 bytes when 
the feature was released. In February 2015, in version 0.10 of Bitcoin Core, the limit was raised back to 
80 bytes. Nodes may choose not to relay or mine OP_RETURN, or only relay and mine OP_RETURN 
containing less than 80 bytes of data.
```

Not 100% clear if there might be issues with nodes not relaying, but seems that 80 bytes is standard limit.

HenrikJannsen:
Rebased on master.

core/src/main/java/bisq/core/dao/burningman/BurningManService.java
HenrikJannsen: Actually we use that for showing all comp requests back to genesis in the UI when selecting a BM candidate.

core/src/main/java/bisq/core/dao/burningman/BurningManService.java
HenrikJannsen: Actually we use that for showing all comp requests back to genesis in the UI when selecting a BM candidate.

core/src/main/java/bisq/core/dao/burningman/DelayedPayoutTxReceiverService.java
```diff
@@ -130,11 +130,24 @@ public List<Tuple2<Long, String>> getReceivers(int burningManSelectionHeight,
...
+                    return amount < minOutputAmount ? cappedBurnAmountShare : 0d;
+                })
+                .sum();
 
         List<Tuple2<Long, String>> receivers = burningManCandidates.stream()
                 .filter(candidate -> candidate.getMostRecentAddress().isPresent())
-                .map(candidate -> new Tuple2<>(Math.round(candidate.getCappedBurnAmountShare() * spendableAmount),
-                        candidate.getMostRecentAddress().get()))
+                .map(candidate -> {
+                    double cappedBurnAmountShare = candidate.getCappedBurnAmountShare() / adjustment;
```
sqrrm, line 147: Should probably be called `adjustedBurnAmountShare` for clarity. It's now possible that some receivers get more than their max allowed, although only by a few satoshis.

core/src/main/java/bisq/core/dao/burningman/BurningManService.java
sqrrm: I didn't realize the `chainHeightOfPastCycle` was `firstBlockOfPastCycle + getOffsetFromFirstBlockInCycle(chainHeight)`. So my suggestion is basically the same as the current code.

desktop/src/main/java/bisq/desktop/main/dao/wallet/send/BsqSendView.java
sqrrm: I saw this comment:

https://stackoverflow.com/questions/24845429/maximum-size-of-data-bitcoin-op-return-tx-can-handle

> Please note that the up to 80 part is misleading: one data push can handle up to 75 bytes. Therefore OP_RETURNs with 76-to-80 bytes need special handling: encoded with OP_PUSHDATA1 with 1-byte length. More details: [bitcoin.stackexchange.com/questions/116080/…](https://bitcoin.stackexchange.com/questions/116080/tx-with-op-return-with-77-bytes-of-data-refused-by-node/116083#comment132105_116083) – 
> [Adam B](https://stackoverflow.com/users/6284023/adam-b)
> [Nov 22 at 14:32](https://stackoverflow.com/questions/24845429/maximum-size-of-data-bitcoin-op-return-tx-can-handle#comment131569988_63675019)

https://bitcoin.stackexchange.com/questions/116080/tx-with-op-return-with-77-bytes-of-data-refused-by-node/116083#comment132105_116083

> UPDATE: The first version of this answer suggested using two data pushes, updated to use OP_PUSHDATA1.
> 
> Further research has clarified the issue:
> 
>     OP_RETURN max data length is 80 bytes, however
>     Maximum bytes in one push data is 75 bytes.
>     Therefore data of sizes between 76 to 80 need to be encoded with OP_PUSHDATA1!
> 
> The script in this transaction is in fact invalid, 6a 4d 53 57 should be interpreted as OP_RETURN OP_PUSHDATA2 53 57, after which 22355 (0x5753) bytes are expected.
> 
> Correct encoding would be to use OP_PUSHDATA1 for data of length 76 to 80 bytes.
> 
> 50 6a 4c 4d <77 bytes> OP_RETURN OP_PUSHDATA1 4d <77 bytes>
>

core/src/main/java/bisq/core/util/validation/InputValidator.java
sqrrm: Maybe `validateForEmpty` or something like it. Less descriptive but less misleading when empty is allowed. No good ideas so I think we'll leave it like this.

core/src/main/java/bisq/core/support/dispute/refund/RefundManager.java
sqrrm: Ah, then this is ok.

core/src/main/java/bisq/core/dao/burningman/DelayedPayoutTxReceiverService.java
HenrikJannsen: Yes, I think it does not really add a risk as its only small amounts

HenrikJannsen:
Ok, I reduce the opReturn to 40 bytes. Its anyway just a new feature without concrete use case so far and an only UI thing.
The opReturn size limit issue got heavily mis-managed by the Bitcoin community. Would not like to know how many dev hours have been wasted for that...

core/src/main/java/bisq/core/util/validation/InputValidator.java
HenrikJannsen: I tried to improve it a bit.

HenrikJannsen:
Rebased on master

sqrrm (commented):
No more comments at the moment, will have to test it but code wise I think it looks good.

ghost:
In regtest, the first person to burn is has a suggested amount in the range of a few thousand BSQ.  Thereafter, burn targets are much smaller than what the first BM burned.  e.g. Alice burns 2000 BSQ.  All other BMs will only be allowed to burn 220 BSQ.  Soon after, no one can burn anything, even if they go through several cycles of compensation.
It does not seem right, and maybe its because I'm missing some setup in regtest.  Could it be due to there not being any proof of burns from the legacy BM?  How would I go about simulating burns from the legacy BM, address `2MzBNTJDjjXgViKBGnatDU3yWkJ8pJkEg9w` ?

core/src/main/java/bisq/core/dao/burningman/BurningManPresentationService.java
```diff
@@ -0,0 +1,322 @@
...
+    public LegacyBurningMan getLegacyBurningManForDPT() {
+        if (legacyBurningManDPT.isPresent()) {
+            return legacyBurningManDPT.get();
+        }
+
+        // We do not add the legacy burningman to the list but keep it as class field only to avoid that it
+        // interferes with usage of the burningManCandidatesByName map.
+        // Those are the hashes used by legacy BM for burning BTC received from DPT.
+        Set<String> hashes = Set.of("1701e47e5d8030f444c182b5e243871ebbaeadb5e82f",
+                "1701293c488822f98e70e047012f46f5f1647f37deb7");
```
ghost, line 262: Why are these burningman hashes hard-coded?

core/src/main/java/bisq/core/dao/burningman/BurningManPresentationService.java
```diff
@@ -0,0 +1,322 @@
...
+    public LegacyBurningMan getLegacyBurningManForBtcFees() {
+        if (legacyBurningManBtcFees.isPresent()) {
+            return legacyBurningManBtcFees.get();
+        }
+
+        // We do not add the legacy burningman to the list but keep it as class field only to avoid that it
+        // interferes with usage of the burningManCandidatesByName map.
+        // The hash used by legacy BM for burning BTC received from BTC trade fees.
+        Set<String> hashes = Set.of("1701721206fe6b40777763de1c741f4fd2706d94775d");
+        LegacyBurningMan legacyBurningManBtcFees = getLegacyBurningMan(LEGACY_BURNING_MAN_BTC_FEES_ADDRESS, hashes);
```
ghost, line 278: Same again, hardcoded?  I have to replace it with the relevant hashes from my database in order to pick up the legacy burningman's activity.

core/src/main/java/bisq/core/dao/burningman/BurningManPresentationService.java
HenrikJannsen: Those are the hashes in the opReturn at burn txs of the BM to tag different use cases (btc fee, del payout tx to refund agent, del payout tx to trader)

core/src/main/java/bisq/core/dao/burningman/BurningManPresentationService.java
HenrikJannsen: You mean for testing it with regtest?

ghost: Yes, Regtest.  I'm confused as to the necessity for hardcoding those hashes: if BM3 does a fee burn, I would have expected it to get integrated in real time, instead of pasting into the sourcecode.

HenrikJannsen: The burn tx opReturn data are arbitrary bytes. One can add some preimage which gets hashed in the UI, and that is what the BM is doing to tag the diff. burn txs to groups. So this helps us to filter for burn txs from btc fees and those for DPTs.

HenrikJannsen:
Close that as #6423 is th final PR and based on that.

PR #6411, opened 2022-11-13
Do not pay out the security deposit of the trade peer to the arbitration case winner

ghost:
Change arbitration payout UI to implement https://github.com/bisq-network/proposals/issues/386.

Amounts are filled out based on proposal 386:

- If BTC buyer is selected as case winner they will get trade amount + buyer security deposit.
- If BTC seller is selected as case winner they will get trade amount + seller security deposit.
- If custom payout is selected arbitrator can specify custom amounts as they wish.

Reviewed possible scenarios with @pazza83 on a jitsi call.

@refund-agent2  ping.

### Screenshots

image

image


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6419, opened 2022-11-18
Prevent NPE in NetworkSettingsView when no connected peers are found

### LINKED ISSUE #6390: Cannot invoke "java.util.List.isEmpty()" because "peers" is null
First time bug report, not sure how it works. Got this bug:

Cannot invoke "java.util.List.isEmpty()" because "peers" is null

How can I fix that?

Android-X13:
Fixes #6390

To reproduce (not always, for some reason):

1. Start one Bisq instance (Bob) without deploying a complete localnet
2. Navigate to Settings > NETWORK INFO

While trying to set the value of `chainHeightTextField` it will pass a null list of peers to `getMostCommonChainHeight` which will throw the NPE. Sometimes it passes a non-null empty list instead, so it doesn't throw any NPE.


Android-X13:
@jmacxx I have just realized that you have included a similar fix in #6376 (which has not yet been merged). That will fix the NPE described in the linked issue above.
I will close this PR

PR #6420, opened 2022-11-21
Minor inclusions in deploy & undeploy targets (Makefile)

Android-X13:
If one wants to undeploy but accidentally makes the mistake to run `make deploy` while the localnet is already deployed, `make undeploy` will not work anymore. Instead of having to manually kill the detached screens (and also having to figuring out how to do it if you're not a screen user), I added a line in the `deploy` target to first check if the localnet is already deployed.

In the `undeploy` target I added a line to remove the dead screen(s) that are hanging around after `make undeploy` (they keep increasing in number after more deploying and undeploying).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,6 +150,8 @@ localnet: .localnet
 # user, you'll need to manually run each of the targets listed below
 # commands manually in a separate terminal or as background jobs.
 deploy: setup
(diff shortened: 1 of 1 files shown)


ghost (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6421, opened 2022-11-21
Fix horizontal centering of overlays on Windows

Android-X13:
Horizontal centering of overlays is not accurate on Windows; this is more noticeable when the application window isn't maximized.

Linux is probably unaffected by this (I've confirmed this isn't an issue on a Debian GNOME system but I'm not sure about other flavors) and I don't know about MacOS, so for now this is meant to be a Windows-only fix.

Before/after examples:

Center1a

Center1b

Center2a

Center2b


ghost (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6422, opened 2022-11-21
Fix:  incorrect wallet tx handling

ghost:
### Fix: altcoins/sepa instant BTC buyer's trade timer started too early

This is problematic in mainnet scenarios when a block is not published for 1 hour.  The user is prompted to start payment after the first confirm, and they are also prompted that the trade has expired and to open mediation.

Issue description / steps to reproduce:

- Alice makes an offer to sell XMR (buy BTC) using Altcoins Instant payment method.
- Bob takes the offer.  Note the time (time "X").
- Wait ~10 minutes.
- In regtest, generate the first trade confirm.  Note the time (time "Y").
- Alice's trade display will show that the trade started at time "X".
- Bob's trade display will show that the trade started at time "Y".
- This is wrong, both traders displays should show that the trade started at time "Y"; the time of the first confirm.
 

---

There was an issue with the transaction having a non-initialized blockTime because on the buyer's side the transaction was not obtained from the wallet, instead the serialized transaction was used which does not have a timestamp.

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/WalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/WalletService.java
@@ -837,17 +837,12 @@ public abstract class WalletService {
                                                  Wallet wallet,
                                                  TransactionConfidence.Source source) throws VerificationException {
         Transaction tx = new Transaction(wallet.getParams(), serializedTransaction);
-        Transaction walletTransaction = wallet.getTransaction(tx.getTxId());
-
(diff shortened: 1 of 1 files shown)


HenrikJannsen (approved):
utACK

HenrikJannsen:
@jmacxx Maybe rename the title to make it more clear that its about incorrect wallet tx handling

alejandrogarcia83 (approved):
utACK

PR #6424, opened 2022-11-22
Increase mouse-wheel vertical scrolling speed in ScrollPane

### LINKED ISSUE #3017: Mouse-wheel scrolling doesn't work well for full-pages that need scrolling
The settings page seems to be the only one that has a full-page scroll bar when fully expanded on my wider screen. But I can make other pages have scroll bars by making the window smaller. On my Windows 7 machine, trying to scroll with the mouse wheel scrolls only 1 or two lines for a full wheel scroll, whereas I would expect it to scroll much farther. Everything else in the app that has a scroll bar (other than full pages - eg the Available offers list) scrolls normally. It doesn't help that Bisq lags like hell and seems to refresh at about 1 frame per second...

Android-X13:
Fixes #3017

This PR increases the mouse-wheel scrolling "speed" inside scrollpanes (as stated in the linked issue above the scrolling is slow only inside ScrollPane but not in other components like TableView). I believe that the speed now is more or less of the degree that is usually expected universally in any kind of application.

Apparently the scroll amount in a ScrollPane depends on the content's full height, so the speed usually differs between views. If the content's height is large enough then the scrolling speed will be faster, but if the content is only slightly larger than the viewable area then the scrolling speed is slower. Example: NetworkSettingsView, where the scrolling speed is painfully slow when the window is maximized. The technique used here makes it so the speed is almost (but not exactly) equal in all scrollable views.

Initally I made this to apply only on Windows machines but after testing on a Debian GNOME system there doesn't seem to be any difference, so I removed the OS check. **I have not tested this on MacOS.**

I'm using a custom `BisqScrollPane` component here, an alternative way would be to use helper methods like setFasterScrollSpeed in GUIUtil. I've tested that code too so I can switch to that approach if requested.

Before:

https://user-images.githubusercontent.com/76814540/203327741-87160f13-608e-4e05-8206-383cf62d3f9d.mp4

After:

https://user-images.githubusercontent.com/76814540/203327778-d1efdd5a-d91c-46ef-9af9-c3fc07437c93.mp4

### DIFF
--- a/desktop/src/main/java/bisq/desktop/components/BisqScrollPane.java
+++ b/desktop/src/main/java/bisq/desktop/components/BisqScrollPane.java
@@ -1,0 +1,72 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.desktop.components;
+
+import bisq.common.util.Utilities;
+
+import javafx.scene.Node;
+import javafx.scene.control.ScrollPane;
+
+import javafx.beans.value.ChangeListener;
+import javafx.beans.value.ObservableValue;
+
+/**
+ * Custom scroll pane that uses a workaround to fix slow vertical scrolling.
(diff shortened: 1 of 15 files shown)


Android-X13:
Signed the commit

HenrikJannsen:
Is that tested with all OS? On OSX I don't have any issues with scroll speed and it works with accelleration. In doubt apply it only to the OS where you obsereved issues.

Android-X13:
It's tested on Windows and Debian (with GNOME), not on MacOS. I can exclude MacOS and maybe even Linux (although in Debian it's an issue - the motion you see in the "Before" video, which was recorded on Windows, is the same in that distro).

ghost (dismissed):
Tested on Ubuntu.  The mouse wheel scrolling fix is a big improvement.  :+1: 

ACK.

Android-X13:
I'll probably need to add an OS check to exclude MacOS then, unless someone can test it there as well

HenrikJannsen:
Better exclude it on OSX as there is no problem with it.

Android-X13:
OSX and everything else: excluded

ghost (approved):

alejandrogarcia83 (approved):
utACK

PR #6425, opened 2022-11-22
Validate BSQ fee payment using DAO tx info.

ghost:
Fixes an issue with BSQ fee validation which has been prompting users to open mediation tickets, as reported by @pazza83

---

Previously the BSQ fee payment was determined by parsing a raw tx without relying on the DAO.  Unfortunately this turned out to be problematic, so with this change the BSQ fee paid is obtained from the DAO tx, as originally preferred by @chimp1984.

Unconfirmed transactions will not be able to have their BSQ fees checked so early requests for validation will skip the fee check. This is not a problem since maker fee validation is done by takers and in the majority of cases will be already confirmed; taker fee validation is done after the first confirm at trade step 2.


HenrikJannsen:
Why do we need the explorer lookup then? We know if the fee payment was BTC or BSQ, at least for maker so we could use the exclusively the DAO data. 
But would postpone that until the BM proposal is deployed as only then we can be sure that all traders have the DAO activated.

ghost:
I went with the simplest minimal change.  Yes, the fee validation could probably be refactored as you suggest.

HenrikJannsen:
> I went with the simplest minimal change. Yes, the fee validation could probably be refactored as you suggest.

I think it would be better to reduce the code to that whats needed now.

ghost:
Refactored as suggested above.  Rebased to latest master.

> But would postpone that until the BM proposal is deployed as only then we can be sure that all traders have the DAO activated.

Marked as [DO NOT MERGE] for now.

pazza83:
Hi would be good to get this merged soon. I have a user in support with multiple offers that are going offline. Shared the logs with @jmacxx who informed me that this PR will fix their issue.

alejandrogarcia83:
@HenrikJannsen Could you ACK this?

core/src/main/java/bisq/core/provider/mempool/TxValidator.java
```diff
@@ -56,12 +57,11 @@ public class TxValidator {
...
     private final DaoStateService daoStateService;
     private final FilterManager filterManager;
-    private long blockHeightAtOfferCreation; // Only set for maker.
+    private long feePaymentBlockHeight; // applicable to maker and taker fees
     private final List<String> errorList;
     private final String txId;
     private Coin amount;
-    @Nullable
-    private Boolean isFeeCurrencyBtc = null;
+    private Boolean isFeeCurrencyBtc = true;
```
HenrikJannsen: As isFeeCurrencyBtc is only used by one constructor and otherwise is not set (and not used) I think keeping it a nullable Boolean is better. In the use cases where its accessed the constructor is use with a non null value, so no need to initialize it with true.

HenrikJannsen: There is one unused constructor as well. Better to remove that.

core/src/main/java/bisq/core/provider/mempool/TxValidator.java
```diff
@@ -85,13 +85,13 @@ public TxValidator(DaoStateService daoStateService,
                        String txId,
                        Coin amount,
                        @Nullable Boolean isFeeCurrencyBtc,
```
HenrikJannsen: Better remove the @Nullable and use primitive boolean value as the callers always provide a non null value.

HenrikJannsen (changes requested):
Beside small mentioned issues utACK

HenrikJannsen (dismissed):

HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

core/src/main/java/bisq/core/provider/mempool/MempoolService.java
```diff
@@ -96,26 +96,36 @@ public void validateOfferMakerTx(OfferPayload offerPayload, Consumer<TxValidator
...
+        if (txValidator.getIsFeeCurrencyBtc() != null && txValidator.getIsFeeCurrencyBtc()) {
+            if (!isServiceSupported()) {
+                UserThread.runAfter(() -> resultHandler.accept(txValidator.endResult("mempool request not supported, bypassing", true)), 1);
+                return;
+            }
+            MempoolRequest mempoolRequest = new MempoolRequest(preferences, socks5ProxyProvider);
+            validateOfferMakerTx(mempoolRequest, txValidator, resultHandler);
+        } else {
+            // using BSQ for fees
+            UserThread.runAfter(() -> resultHandler.accept(txValidator.validateBsqFeeTx(true)), 1);
```
ghost, line 108: This BSQ fee check should be gated by a call to `isServiceSupported()`

alvasw: We don't need it there. The `MempoolService` is not used to validate the BSQ fee because `txValidator.validateBsqFeeTx(true)` gets the transaction from the `DaoStateService`.

ghost: With the change to have BSQ fees checked using the DAO database rather than with data obtained from mempool.space, `MempoolService` is incorrectly named.  (_In fact, it was unfortunate that it was named referring to an implementation feature rather than functionality_).  

It would be more appropriately called `TradeFeeCheckingService`.  Fee checking should be gated by `isServiceSupported()`, so that the filter and config settings can do their job in enabling/disabling the fee checking feature.

PR #6431, opened 2022-11-27
Increase trading limits as an opt-in setting

ghost:
Implements [proposal 398](https://github.com/bisq-network/proposals/issues/398) and [proposal 395](https://github.com/bisq-network/proposals/issues/395)

- User's initial max limit is 0.1 BTC, can be self-increased upto the max trade limit DAO param.
- More restrictive limits will still apply based on payment method.
- XMR Auto-Confirm will default to enabled for new users.

It is intended to avoid that a new users who do not fully understand the process of a Bisq trade to cause an arbitration case with high amounts and therefore higher risks and costs for the DAO.

---

image

image

image

image

image

image

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -1319,6 +1319,19 @@ setting.preferences.autoConfirmEnabled=Enabled
 setting.preferences.autoConfirmRequiredConfirmations=Required confirmations
 setting.preferences.autoConfirmMaxTradeSize=Max. trade amount (BTC)
 setting.preferences.autoConfirmServiceAddresses=Monero Explorer URLs (uses Tor, except for localhost, LAN IP addresses, and *.local hostnames)
+setting.preferences.tradeLimits=Trade Limits
+setting.preferences.tradeLimitsEnabled=Increase Trade Limits
+setting.preferences.tradeLimitMax=Max. trade amount (BTC)
+setting.preferences.tradeLimitBlurb=Trading limits are set to reduce risk of costly disputes and losing funds for new \
+  users. Raise trading limits at your own risk. Every trader has their own pace, but it's recommended to complete at \
+  least 5 trades before raising trade limits.\n\n\
+  By raising trading limits you acknowledge to be familiar with the basic use of Bisq, like backing up your data \
+  [HYPERLINK:https://bisq.wiki/Backing_up_application_data], \
+  trading rules [HYPERLINK:https://bisq.wiki/Trading_rules], dispute resolution \
+  [HYPERLINK:https://bisq.wiki/Dispute_resolution], penalties [HYPERLINK:https://bisq.wiki/Table_of_penalties] or find \
+  support [HYPERLINK:https://bisq.wiki/User_support].\n\n\
+  Account signing limits [HYPERLINK:https://bisq.wiki/Account_signing] are still applied to buy BTC with most National \
+  Account payment methods [HYPERLINK:https://bisq.wiki/Payment_methods], to reduce chargeback risk.
 setting.preferences.deviationToLarge=Values higher than {0}% are not allowed.
 setting.preferences.txFee=BSQ Withdrawal transaction fee (satoshis/vbyte)
 setting.preferences.useCustomValue=Use custom value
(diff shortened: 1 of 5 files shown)


HenrikJannsen (dismissed):
utACK

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1319,6 +1319,15 @@ setting.preferences.autoConfirmEnabled=Enabled
...
+setting.preferences.tradeLimits=Trade Limits
+setting.preferences.tradeLimitsEnabled=Increase Trade Limits
+setting.preferences.tradeLimitMax=Max. trade amount (BTC)
+setting.preferences.tradeLimitBlurb=Trading limits are set to reduce risk of costly disputes and losing funds for new \
+  users. Raise trading limits at your own risk. Every trader has their own pace, but it''s recommended to complete at \
+  least 5 trades before raising trade limits.\n\n\
+  By raising trading limits you acknowledge to be familiar with the basic use of Bisq, like backing up your data [HYPERLINK:https://bisq.wiki/Backing_up_application_data], \
+  trading rules [HYPERLINK:https://bisq.wiki/Trading_rules], dispute resolution [HYPERLINK:https://bisq.wiki/Dispute_resolution], penalties [HYPERLINK:https://bisq.wiki/Table_of_penalties] or find support [HYPERLINK:https://bisq.wiki/User_support].\n\n\
+  Account signing limits [HYPERLINK:https://bisq.wiki/Account_signing] are still applied to most National Account payment methods [HYPERLINK:https://bisq.wiki/Payment_methods], to reduce chargeback risk.
```
MwithM: [...] are still applied to buy BTC with most National Account payment [...]
Add **buy BTC with** for clarity.

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1319,6 +1319,15 @@ setting.preferences.autoConfirmEnabled=Enabled
 setting.preferences.autoConfirmRequiredConfirmations=Required confirmations
 setting.preferences.autoConfirmMaxTradeSize=Max. trade amount (BTC)
 setting.preferences.autoConfirmServiceAddresses=Monero Explorer URLs (uses Tor, except for localhost, LAN IP addresses, and *.local hostnames)
+setting.preferences.tradeLimits=Trade Limits
+setting.preferences.tradeLimitsEnabled=Increase Trade Limits
+setting.preferences.tradeLimitMax=Max. trade amount (BTC)
+setting.preferences.tradeLimitBlurb=Trading limits are set to reduce risk of costly disputes and losing funds for new \
+  users. Raise trading limits at your own risk. Every trader has their own pace, but it''s recommended to complete at \
```
MwithM: `it''s` typo

ghost:
Rebased, resolved conflicts and applied @MwithM's review suggestions.

alejandrogarcia83 (approved):
utACK

ripcurlx:
I just tested this feature and I'm not sure if it might be confusing for existing users. E.g. someone who has already signed accounts and traded e.g. 0.25 BTC on SEPA or 2 BTC on XMR will see this new limit applied when creating an offer.

Bildschirmfoto 2022-12-19 um 09 54 01
Bildschirmfoto 2022-12-19 um 09 53 51

Now they have no idea how to change this user set trade limit except trying to reach out in forums, searching through the app settings. I think we should add an informational popup on startup and a NEW badge on the Settings menu item.

@jmacxx @HenrikJannsen @pazza83, @MwithM: WDYT?

ripcurlx:
Bildschirmfoto 2022-12-19 um 10 01 38

I also think this popup needs to be adapted pointing out this new user limit feature, wdyt?

ghost:
@ripcurlx For an existing user we could initialize the limit to the size of their largest prior trade.  #6473

ripcurlx:
> @ripcurlx For an existing user we could initialize the limit to the size of their largest prior trade. #6473

Yes, maybe that would be even better, so existing users wouldn't recognize anything at all. 👍

MwithM:
@ripcurlx I agree that this pop-ups (and the wiki), needs to be updated, but the assumption is that if a user doesn't know how to increase the limits, they won't know other basic stuff that generates trade going to mediation, like where to ask for support.

On that pop-up I would only add the bold text:
> To limit chargebacks **on fiat payment methods**, Bisq sets per-trade limits...

Then, the [account limits](https://bisq.wiki/Account_limits) wiki article nedds to be edited to point exactly how to increase limits for alts. That way, is more probable that users who raised the limit at least know about the wiki.

PR #6433, opened 2022-11-30
User defined trade limit, Part II.

ghost:
Following on from #6431, some commits were misplaced during a code rebase.

- The user limit is applied to offer entry UI. 
- Offers are filtered according to limits in OfferFilterService. 
- If limit changed, the cache in OfferFilterService must be cleared.

### DIFF
--- a/core/src/main/java/bisq/core/offer/OfferFilterService.java
+++ b/core/src/main/java/bisq/core/offer/OfferFilterService.java
@@ -48,7 +48,7 @@ public class OfferFilterService {
     private final FilterManager filterManager;
     private final AccountAgeWitnessService accountAgeWitnessService;
     private final Map<String, Boolean> insufficientCounterpartyTradeLimitCache = new HashMap<>();
-    private Map<String, Boolean> myInsufficientTradeLimitCache = new HashMap<>();
+    private final Map<String, Boolean> myInsufficientTradeLimitCache = new HashMap<>();
 
     @Inject
     public OfferFilterService(User user,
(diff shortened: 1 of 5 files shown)


core/src/main/java/bisq/core/offer/OfferFilterService.java
```diff
@@ -200,10 +200,12 @@ public boolean isMyInsufficientTradeLimit(Offer offer) {
...
-        boolean result = offer.isFiatOffer() &&
-                accountOptional.isPresent() &&
-                myTradeLimit < offerMinAmount;
+        boolean result = accountOptional.isPresent() && myTradeLimit < offerMinAmount;
         myInsufficientTradeLimitCache.put(offerId, result);
         return result;
     }
+
+    public void resetTradeLimitCache() {
+        myInsufficientTradeLimitCache = new HashMap<>();
```
alvasw: You could call `myInsufficientTradeLimitCache.clear()`. This would allow the field to stay final.


This would allow you to call `resetTradeLimit()` above (see line 66) too.
https://github.com/bisq-network/bisq/blob/7ec86c3dc3433aa178c867d11050cfda7c9d967e/core/src/main/java/bisq/core/offer/OfferFilterService.java#L63-L67

alvasw (dismissed):
ACK (tested on mainnet)

alvasw:
When the user defined trade limit is activated this pops up when a user tries to take an offer that he's not allowed to take. Do we want to add "user defined trade limit" to the list or is it hidden on purpose so that only experienced traders know about this feature?

bisq_jmacxx_pr

alvasw (approved):
ACK

alejandrogarcia83 (approved):
utACK

PR #6435, opened 2022-12-02
Prompt to confirm before shutting down.

ghost:
Implements request from @pazza83 and https://github.com/bisq-network/bisq/discussions/6427.  Prevents accidental shutdown.

image

There are other checks which may cause an alternate prompt at shutdown:

- Open offers
- Trades which have not finished initializing
- Dispute messages still waiting for an ack

At shutdown either one or the other will be shown, as long has user has not opted to bypass it.

image


ripcurlx (approved):
utACK

PR #6442, opened 2022-12-06
Reenable verbose dependency verification (failure report)

alvasw:
The built-in Gradle dependency verification XML writer does not find all our libraries for some unknown reason. I had to compute and add multiple hashes manually. It seems like it is OS related and the CI output helped to fix the problem. The console output makes it hard to fix issues because we cannot access the generated HTML file. This change reenables the verbose failure report.


alejandrogarcia83 (approved):
utACK

PR #6443, opened 2022-12-06
Enforce gradle dependency verification

alvasw:
We had some issues with the Gradle Dependency Verification and nobody noticed it because it was not enforced by Gradle. Supply chain attack are sadly a real threat that should not be underestimated.


alejandrogarcia83 (approved):
utACK

PR #6449, opened 2022-12-09
Introduce Gradle Version Catalog

alvasw:
To simplify the build.gradle file, this change moves all version variables to Gradle's version catalog.


HenrikJannsen (approved):
ACK

alejandrogarcia83 (approved):
utACK

PR #6456, opened 2022-12-13
Remove connection:keep-alive header

HenrikJannsen:
The connection:keep-alive header is restricted in java and would require a jvm option to allow it.


ghost (approved):
ACK

ripcurlx (approved):
utACK - based on https://github.com/bisq-network/bisq/pull/6456#pullrequestreview-1214721846

PR #6459, opened 2022-12-13
Handle exception at send

HenrikJannsen:
This PR carries maybe a bit of risk as we did not handle exceptions at send message before. Now the onFailure handler is called. Before the onSuccess was called. There might be some unintended effects from that. If discovered we should fix that as now it uses the correct behaviour.

Based on #6458


ripcurlx:
I'd be fine merging this into the release branch and do more extensive release testing.

ripcurlx (approved):
utACK

PR #6466, opened 2022-12-16
Fix divide by zero errors in Trade history summary.

### LINKED ISSUE #6464: show summary under portfolio history error, log attached

ghost:
Fixes #6464 

Trade history summary was giving div/0 exception for user with no trade history.

image


HenrikJannsen (commented):
NACK

I would not suggest to do that. I think its better to add checks when there are divisions to not have a 0 value. Setting the result to 0 would be inverse to the infinite result from a division by 0.

ghost:
Re-written as suggested in https://github.com/bisq-network/bisq/pull/6466#pullrequestreview-1221645094

By the way, the desired result for the trade fee percentage is zero when there are no trades, as seen in the screenshot above.

HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6468, opened 2022-12-16
Use ThreadPoolExecutor with custom set queueCapacity instead of CachedThreadPool

HenrikJannsen:
The previously used newCachedThreadPool carries higher risk for execution exceptions if exceeded.

Originally we had only one executor with a corePoolSize of 15 and a maximumPoolSize of 30 and queueCapacity was set to maximumPoolSize. This was risky when the 15 corePool threads have been busy and new messages or connection creation threads are queued up with potentially significant delay until getting served leading to timeouts. Now we use (if maxConnections is 12) corePoolSize of 24, maximumPoolSize 36 and queueCapacity 10. This gives considerable headroom. We also have split up the executors in 2 distinct ones.


alejandrogarcia83:
@alvasw Can you ACK this?

sqrrm (approved):
utACK

PR #6469, opened 2022-12-16
Reduce Response size

HenrikJannsen:
Reduce getDataResponse size from 9 MB to 6 MB
Reduce number of blocks at GetBlocksResponse from 4000 to 3000

No need to get that into the release branch as it affects seed nodes only.


alejandrogarcia83 (approved):
utACK

PR #6473, opened 2022-12-19
Base initial limit from user's max past trade size if applicable.

ghost:
Addressing feedback received in https://github.com/bisq-network/bisq/pull/6431#issuecomment-1357305934


ripcurlx (approved):
utACK

PR #6476, opened 2022-12-20
[v1.9.7] Wallet password broken: Not use additional `-%d` in name format because of refactoring

ripcurlx:
Because of the changes in https://github.com/bisq-network/bisq/blob/f66e6936051b587ac6271576f895c1991a8ac2ff/common/src/main/java/bisq/common/util/Utilities.java#L145
we don't need the duplicate Guava placeholder for the unique identifier anymore.


ripcurlx:
@HenrikJannsen ☝️

bisq-github-admin-3 (approved):
utACK

PR #6477, opened 2022-12-20
Add back distribution tar generation for daemon and cli

ripcurlx:
This is required in finalize.sh script for creating the release bundle


bisq-github-admin-3 (approved):
utACK

Issue #6480, opened 2022-12-21
Error Task TrustedListenableFutureTask@7d185bf2[status=PENDING, info=[task=[running=[NOT STARTED YET]


pprvggc6um:
### Description
Error windows opens with the following text

Error
Task TrustedListenableFutureTask@7d185bf2[status=PENDING, info=[task=[running=[NOT STARTED YET], bisq.network.p2p.network.NetworkNode$$Lambda$1211/0x000000001fff47c0@2bbeee47]]] rejected from java.util.concurrent.ThreadPoolExecutor@114159ed[Running, pool size = 36, active threads = 36, queued tasks = 10, completed tasks = 125192]

#### Version

1.9.7

### Steps to reproduce

Error appears sometime after application startup

### Expected behaviour

Error window should not appear

### Actual behaviour

Error window is shown

#### Device or machine

HP Elitebook 8470p
Windows 7 Pro

#### Additional info
error log available, but too large to paste into form

ghost:
@pprvggc6um could you attach the log, or paste the stack trace from the log which is immediately following the `ERROR` line in the log.  100 or so lines from the log before and after the error might be useful for context also.  Thanks

pprvggc6um:
[bisq.log](https://github.com/bisq-network/bisq/files/10280208/bisq.log)

ghost:
@HenrikJannsen could you look into this?

    Dec-21 05:44:27.602 [JavaFX Application Thread] ERROR b.c.s.CommonSetup: Uncaught Exception from thread JavaFX Application Thread 
    Dec-21 05:44:27.626 [JavaFX Application Thread] ERROR b.c.s.CommonSetup: throwableMessage= Task TrustedListenableFutureTask@28630e7c[status=PENDING, info=[task=[running=[NOT STARTED YET], bisq.network.p2p.network.NetworkNode$$Lambda$1211/0x000000001fff47c0@39a9487a]]] rejected from java.util.concurrent.ThreadPoolExecutor@114159ed[Running, pool size = 36, active threads = 36, queued tasks = 10, completed tasks = 99013] 
    Dec-21 05:44:27.626 [JavaFX Application Thread] ERROR b.c.s.CommonSetup: throwableClass= class java.util.concurrent.RejectedExecutionException 
    Dec-21 05:44:27.626 [JavaFX Application Thread] ERROR b.c.s.CommonSetup: Stack trace:
[...]

PR #6481, opened 2022-12-22
Increase queue capacity from 10 to 30.

HenrikJannsen:
https://github.com/bisq-network/bisq/issues/6480 reported reaching the limit with the current settings.


ripcurlx (approved):
utACK

PR #6488, opened 2022-12-27
Add threadpool for broadcaster

### LINKED ISSUE #6480: Error Task TrustedListenableFutureTask@7d185bf2[status=PENDING, info=[task=[running=[NOT STARTED YET]

HenrikJannsen:
Fixes https://github.com/bisq-network/bisq/issues/6480

At broadcast there can be many messages been sent and if the tor connections are slow the sendMessage threadpool can get exhausted. We did not handle the RejectedExecutionException so it escalated as uncaught exception to the user in from of an error popup (pre 1.9.7 it was swallowed in a debug log, so I assume we had RejectedExecutionException aslo earlier).
We catch now RejectedExecutionException at the sendMessage method as well at the uncaughtExceptionHandler to only log an error but not open an error popup.
We add a dedicated threadpool to the broadcaster so that we can separate that pool from the common sendMessage pool. In case a burst of broadcasts would exhaust the broadcaster pool we get dropped some of the broadcast messages which has little impact and risk. The common sendMessage pool will be not affected in such a scenario, thus important messages like trade messages are not at risk.
We reduced also the keepAliveTime to 30 sec to lower resource comsumption.

### DIFF
--- a/common/src/main/java/bisq/common/setup/CommonSetup.java
+++ b/common/src/main/java/bisq/common/setup/CommonSetup.java
@@ -35,6 +35,7 @@ import java.net.URISyntaxException;
 
 import java.nio.file.Paths;
 
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
 
 import ch.qos.logback.classic.Level;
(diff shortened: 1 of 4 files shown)


ghost (approved):
utACK

PR #6489, opened 2022-12-28
Increase capacity for thread pool queue at ExportJsonFilesService

HenrikJannsen:
One seed node using the ExportJsonFilesService had RejectedExecutionExceptions as before we had only permitted a thread pool queue capacity of 1.


ghost (approved):
utACK

PR #6494, opened 2023-01-02
Fix trade fee validation with new BM

HenrikJannsen:
Add all addresses of all burningman who have burned BSQ to the list of addresses who are permitted to receive trade fees.


ghost (approved):
ACK

alejandrogarcia83 (approved):
utACK

PR #6495, opened 2023-01-03
Fix bug with calculating the miner fee for the DPT.

HenrikJannsen:
We used all potential BM instead only the ones who have a positive cappedBurnAmountShare.


ghost (dismissed):
ACK.

- For a trade of 0.15 BTC the DPT fee observed now is `4350` sats and the legacy BM receives an output along with all other BM.
- For a small trade of 0.001 BTC the DPT fee observed now is `20660` sats and no payout to the legacy BM.  This is less than DPT_MIN_REMAINDER_TO_LEGACY_BM according to spec.

HenrikJannsen:
Did 2 live trades with tiny amount and 0.1 BTC based on https://github.com/HenrikJannsen/bisq/tree/prepare_v1-9-9 and both behaved as expected.

alejandrogarcia83 (approved):
utACK

PR #6504, opened 2023-01-05
Add check for dao state hash conflict with seed node

HenrikJannsen:
Add to trade protocol a check if dao state is in sync.
Improve DaoPresentation and show popup in case dao state is not in sync.
Add more info to logs when DPT check fails


ghost (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6509, opened 2023-01-07
Avoid multiple popups

HenrikJannsen:
Avoid that at repeated onDaoStateHashesChanged calls that we show multiple popups.


alejandrogarcia83 (approved):
utACK

Chelo5479 (commented):
Is void now

HenrikJannsen:
@Chelo5479 Did your review find issues? Still good to report is there is any to fix in a new PR...

PR #6525, opened 2023-01-17
Fix UI filter box.

### LINKED ISSUE #6475: Status bar not updated
### Description

When using the filter in *Porfolio / History* the status bar indicating the number of entries is not updated.

#### Version
v1.9.6

### Steps to reproduce

Switching to *Porfolio / History* and start typing in the field *Filter*.

### Expected behaviour

The number of entries indicated in the status bar is updated and reflects the number of entries shown.

### Actual behaviour

The number is not updated while using the filter.

### Screenshots
bisq-update-number

ghost:
Fixes #6475 

As [described](https://github.com/bisq-network/bisq/issues/6475) in the issue, the problem being fixed is: The row count number (summary of items being shown) is not updated while using the filter, only after changing tabs.  It should always reflect the number of records shown.

    Select HISTORY tab and remember the number. Example 100
    Apply a Filter and count the elements manually. Example 2
    Change to the MY OPEN OFFERS tab
    Change back to the HISTORY tab and check the number again.

    In our example the number is 2 and not 100. So the number changed to reflect the applied filter, but it is applied only after changing tabs.

**Applies to:** 

- Funds/  Transactions, Locked, Reserved.
- Portfolio/  Open Offers, Unconfirmed Bisq Swaps, History, Failed.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/funds/reserved/ReservedView.java
+++ b/desktop/src/main/java/bisq/desktop/main/funds/reserved/ReservedView.java
@@ -132,7 +132,6 @@ public class ReservedView extends ActivatableView<VBox, Void> {
 
     @Override
     public void initialize() {
-        filterBox.initialize(filteredList, tableView);
         dateColumn.setGraphic(new AutoTooltipLabel(Res.get("shared.dateTime")));
         offerIdColumn.setGraphic(new AutoTooltipLabel(Res.get("shared.offerId")));
         detailsColumn.setGraphic(new AutoTooltipLabel(Res.get("shared.details")));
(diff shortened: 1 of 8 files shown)


alejandrogarcia83 (approved):
ACK - Tested it on Mainnet
Congrats to your first merged PR in Bisq!

desktop/src/main/java/bisq/desktop/main/funds/reserved/ReservedView.java
```diff
@@ -173,16 +172,17 @@ public void onBalanceChanged(Coin balance, Transaction tx) {
 
     @Override
     protected void activate() {
-        filterBox.activate();
         openOfferManager.getObservableList().addListener(openOfferListChangeListener);
         tradeManager.getObservableList().addListener(tradeListChangeListener);
         sortedList.comparatorProperty().bind(tableView.comparatorProperty());
         tableView.setItems(sortedList);
         updateList();
+        filterBox.initializeWithCallback(filteredList, tableView, () ->
```
HenrikJannsen, line 180: @yonson2023 Why is that not inside initialize()?

PR #6527, opened 2023-01-17
Warn user when free disk space is too low.

ghost:
If there is insufficient free disk space to persist settings, user's database can be corrupted.
Check periodically for adequate free disk space, and show a warning message if necessary.

disk_space_warning

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqSetup.java
+++ b/core/src/main/java/bisq/core/app/BisqSetup.java
@@ -552,10 +552,10 @@ public class BisqSetup {
     }
 
     private void checkFreeDiskSpace() {
-        long REQUIRED_FREE_DISK_SPACE = 2147483648L;
+        long TWO_GIGABYTES = 2147483648L;
         long usableSpace = new File(Config.appDataDir(), VERSION_FILE_NAME).getUsableSpace();
-        if (usableSpace < REQUIRED_FREE_DISK_SPACE) {
-            String message = Res.get("popup.warning.diskSpace", formatBytes(usableSpace), formatBytes(REQUIRED_FREE_DISK_SPACE));
+        if (usableSpace < TWO_GIGABYTES) {
+            String message = Res.get("popup.warning.diskSpace", formatBytes(usableSpace), formatBytes(TWO_GIGABYTES));
             log.warn(message);
             if (diskSpaceWarningHandler != null) {
                 diskSpaceWarningHandler.accept(message);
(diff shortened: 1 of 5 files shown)


core/src/main/java/bisq/core/app/BisqSetup.java
```diff
@@ -547,6 +551,18 @@ private void checkForInvalidMakerFeeTxs() {
         });
     }
 
+    private void checkFreeDiskSpace() {
+        long REQUIRED_FREE_DISK_SPACE = 2147483648L;
```
alejandrogarcia83: Renaming `REQUIRED_FREE_DISK_SPACE` to something like `2GbInBytes` would make the code easier to read.

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -3202,6 +3202,10 @@ popup.warning.openOfferWithInvalidMakerFeeTx=The maker fee transaction for offer
   Please go to \"Settings/Network info\" and do a SPV resync.\n\
   For further help please contact the Bisq support channel at the Bisq Matrix Space.
 
+popup.warning.diskSpace=Your free disk space is critically low.\n\
```
alejandrogarcia83: IMO the error message is not user-friendly. What do you think about "You have less than 2 GB free disk space left. Please free up some disk space to continue running Bisq."

alejandrogarcia83 (changes requested):
Hi @yonson2023!
Thank you for implementing this. Code wise it looks good!

1) If we're low on disk space it doesn't make any sense continue running. However, we cannot shut down Bisq without showing that popup and pausing all operations is non-trivial. So what do you think about replacing the close button with a shutdown button?
You could call:
```
new Popup().warning(msg).
    .useShutDownButton()
    .hideCloseButton()
    .show()
```

2) We had a bug in the past where a new "DAO is out of sync" popup kept popping up and the user was not able to click the "resync" button. Some users create offers and leave Bisq open for an extended period of time. Here we are opening a new popup every time a new block gets created (if we're low on disk space). I would suggest to check whether a popup is already present.

ghost:
> 1. replacing the close button with a shutdown button

Shutting down when disk space is exhausted is what causes corruption of Bisq's DB files.  What the user needs to do is free up some disk space urgently in this case.

I applied your other suggestions, thanks!

alejandrogarcia83 (approved):
> Shutting down when disk space is exhausted is what causes corruption of Bisq's DB files. What the user needs to do is free up some disk space urgently in this case.

I agree with you. Nothing we can really do because it's most likely too late for long-running instances.

ACK. Tested on regtest.

The popup check is not optimal because another popup (dao out of sync, trade notification, ...) could be present already, but it is a rare case.

PR #6528, opened 2023-01-18
Refactor closing ticket logic.

### LINKED ISSUE #6517: RA - Payout is done before Tx validation.
- Upon pressing `close ticket` the refund agent is prompted to confirm the payout.
- Afterwards, the DPT is checked.

It is not the correct order of operation, it should check the DPT first.

I think the workflow should be as follows:

image

ghost:
Fixes #6517


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6531, opened 2023-01-20
Remove clearnet address for BTC node "devinbtcyk"

devinbileck:
I switched hosts and currently do not have an accessible clearnet address.


alejandrogarcia83 (approved):
utACK

PR #6535, opened 2023-01-22
Fix: allow funds received in trade payout to be used for trading.

### LINKED ISSUE #6534: Available funds cannot be used for trading until prior trade moved to closed.
### Description

You've completed a trade and received funds are confirmed, but you have not yet moved the trade to "Closed Trades".
Your funds are prevented from being used in a new trade even though "Available Funds" indicates they are.

#### Version

1.9.9 and earlier.

### Steps to reproduce

Alice completes a trade to buy BTC, but does not move the trade to "Closed Trades" yet.
Alice's available funds are ~0.25 BTC.
Alice creates an offer to sell 0.2 BTC, but is prevented from doing so with the message:

    Warning
    You don't have enough funds in your Bisq wallet for this transaction—0.23002440 BTC is needed but only 0.09895374 BTC is available.
    Please add funds from an external wallet, or fund your Bisq wallet at Funds > Receive Funds.

### Expected behaviour

0.25 BTC is available, so the offer should go through without error.

### Actual behaviour

Alice cannot create a new offer until the old trade is moved to "Closed Trades"

### Screenshots

image

image

image

image

image

ghost:
Fixes #6534 

There is no need to wait for user to press the `Close` button before marking in funds as available.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/steps/buyer/BuyerStep4View.java
+++ b/desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/steps/buyer/BuyerStep4View.java
@@ -134,12 +134,11 @@ public class BuyerStep4View extends TradeStepView {
                     .autoClose()
                     .show();
         }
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
utACK

PR #6541, opened 2023-01-27
DaoStateStorageService: Don't call executorService.shutdown() twice

alvasw:
MoreExecutors.shutdownAndAwaitTermination(executorService, ...) calls executorService.shutdown() before starting its force shutdown timer.


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6542, opened 2023-01-27
Connection: Close protoOutputStream before closing socket

alvasw:
The protoOutputStream depends on the sockets OutputStream. Closing the socket before closing the protoOutputStream will always trigger an IOException.


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6543, opened 2023-01-27
Sychronized proto output stream try graceful shutdown first

alvasw:
ProtoOutputStream is not thread-safe. This change tries to close the stream on the same thread first. Before forcefully killing it after a timeout.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/network/SynchronizedProtoOutputStream.java
+++ b/p2p/src/main/java/bisq/network/p2p/network/SynchronizedProtoOutputStream.java
@@ -19,12 +19,15 @@ package bisq.network.p2p.network;
 
 import bisq.common.proto.network.NetworkEnvelope;
 
+import com.google.common.util.concurrent.MoreExecutors;
+
 import java.io.OutputStream;
 
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


p2p/src/main/java/bisq/network/p2p/network/SynchronizedProtoOutputStream.java
```diff
@@ -62,8 +65,18 @@ void writeEnvelope(NetworkEnvelope envelope) {
...
         try {
-            executorService.shutdownNow();
-            super.onConnectionShutdown();
+            // ProtoOutputStream is not thread-safe that's why try to close the stream
+            // on the same thread first.
+            executorService.submit(super::onConnectionShutdown);
+
+            //noinspection UnstableApiUsage
+            boolean terminatedSuccessfully = MoreExecutors.shutdownAndAwaitTermination(
+                    executorService, Connection.getShutdownTimeout() * 2L, TimeUnit.MILLISECONDS);
```
HenrikJannsen, line 74: What is the reasoning to use the double of the connection shutdown time here?

HenrikJannsen (approved):
utACK

p2p/src/main/java/bisq/network/p2p/network/SynchronizedProtoOutputStream.java
alvasw: `MoreExecutors.shutdownAndAwaitTermination` first disables acceptance of new submitted tasks waits half of the timeout. Afterwards it cancels pending tasks and interrupts running tasks (see [1]).

[1] https://guava.dev/releases/23.5-jre/api/docs/com/google/common/util/concurrent/MoreExecutors.html#shutdownAndAwaitTermination-java.util.concurrent.ExecutorService-long-java.util.concurrent.TimeUnit-

alejandrogarcia83 (approved):
utACK

PR #6544, opened 2023-01-27
Fix withdrawal fee calculation & the min relay fee problem.

### LINKED ISSUE #6493: Withdrawal TX not accepted by peers. Fee too low.
### Description

I ran into a problem where I tried to withdraw some coins I bought into my hardware wallet. I set the fee rate to 1sat/vbyte as I am not in a rush and want to save on fees. The transaction shows up on the transaction list inside of bisq but in the confirmations column it shows "Seen by 0 peers" forever and the transaction is never mined or viewable on any blockchain explorer. 
However the balance I wanted to withdraw is subtracted from my available balance and is not spendable anymore.
Only way to regain these funds is to Delete SPV file and resync.

#### Version

v1.9.8

### Steps to reproduce

In the 'send funds' dialog select a single native segwit UTXO as input, spend the entire UTXO to another native segwit address, check 'Amount includes mining fee' and set a custom withdrawal transaction fee rate of 1sat/vbyte

### Expected behaviour

- Create a transaction with a fee rate of around 1.02sat/vbyte so that it surpasses the minRelayTxFee of 1000sat/kB
- allow floating point input for fee rate so a user can enter something between 1-2sat/vbyte and create a TX that is not rejected by standard mempool settings
- If a transactions is not accepted by any peers after a defined period of time scrap it and make the inputs spendable again. Also alert the user of the scrapped TX and the possible reason.

### Actual behaviour

Transaction shows up as broadcasted but seen by 0 peers. Funds are marked as spent.
In my testing I calculated that my withdrawal TX hat a fee rate of exactly 1000 sat/kB whereas [this post](https://bitcoin.stackexchange.com/questions/48235/what-is-the-minrelaytxfee#48246) says it needs to be **greater than** 1000 to be valid.

#### Device or machine

amd64 Ubuntu

ghost:
**_1.0 sat/vB withdrawals were sometimes sent at 0.995 sat/vB and rejected by the network.  Fixes #6493_** 

### Description of the problem:

Fee estimation for withdrawals has a bug which can be seen if you repeatedly request the withdrawal of the same UTXO and observe the fee shown; it varies slightly.

Worse than that is when a withdrawal is started (after you've seen the preview and confirmed), Bisq re-calculates the tx using the fee calculated earler and the actual receive address, but the tx can end up being larger than what it used in the preview.  

The main problem here is that in the case where user chose 1 sat/vB, the sent tx can end up being below the min relay fee (less than 1 sat/vB) and is rejected by nodes.  The user is left with a corrupted wallet and has to do an SPV resync.  See example reported in #6493.

Why does this happen?  See the fee estimation code here:

https://github.com/bisq-network/bisq/blob/d682d956a16f9a1b21cd3303456ad755378c5ca1/core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java#L1092-L1101

It creates a random receive address for purposes of calculating the fee.  The problem with this is two-fold:

- The address used for estimation can end up taking less tx space than the actual receiver address, resulting in a send tx which uses a smaller fee than requested (as the above code comment says)
- The receiver address, being random, will result in a signature that can vary in length (signatures are sometimes 72 bytes sometimes 71 depending on the content being signed).  That is why when you repeatedly do a "send preview" the shown fee rate varies by a small percentage.

### Fix implemented here:

The withdrawal fee estimation should use the actual address the user has chosen to send to, therefore eliminating all variability in results.  The sent tx will be exactly the same as the tx used in the preview, therefore fixing the problem of lower than 1 sat/vB transactions.

### DIFF
--- a/core/src/main/java/bisq/core/api/CoreWalletsService.java
+++ b/core/src/main/java/bisq/core/api/CoreWalletsService.java
@@ -289,6 +289,7 @@ class CoreWalletsService {
             //  See WithdrawalView # onWithdraw (and refactor).
             Transaction feeEstimationTransaction =
                     btcWalletService.getFeeEstimationTransactionForMultipleAddresses(fromAddresses,
+                            address,
                             receiverAmount,
                             txFeePerVbyte);
             if (feeEstimationTransaction == null)
--- a/core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
@@ -1056,13 +1056,7 @@ public class BtcWalletService extends WalletService {
     }
 
     public Transaction getFeeEstimationTransactionForMultipleAddresses(Set<String> fromAddresses,
-                                                                       Coin amount)
-            throws AddressFormatException, AddressEntryException, InsufficientFundsException {
-        Coin txFeeForWithdrawalPerVbyte = getTxFeeForWithdrawalPerVbyte();
-        return getFeeEstimationTransactionForMultipleAddresses(fromAddresses, amount, txFeeForWithdrawalPerVbyte);
-    }
-
-    public Transaction getFeeEstimationTransactionForMultipleAddresses(Set<String> fromAddresses,
+                                                                       String toAddress,
                                                                        Coin amount,
                                                                        Coin txFeeForWithdrawalPerVbyte)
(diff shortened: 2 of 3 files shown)


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6547, opened 2023-01-30
Make ProtoOutputStream thread-safe

alvasw:
Before SynchronizedProtoOutputStream created a new thread and made a blocking call to ProtoOutputStream.writeEnvelope(...). Each connection had an instance of SynchronizedProtoOutputStream, so we had a redundant thread per connection. This change makes ProtoBufOutputStream thread-safe and reduces the number of thread per connection as a side effect.


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6548, opened 2023-01-31
DaoStateStorageService: Fix shutdown data corruption bug

alvasw:
The DaoStateStorageService submits tasks to a daemon thread executor service. The JVM continues to execute threads until all non-daemon threads have terminated. So the JVM will kill the DaoStateStorageService during write requests leading to data corruption.

This change uses a non-daemon thread executor service at waits during shutdown until all data is written to disk.


alejandrogarcia83 (approved):
utACK

PR #6549, opened 2023-01-31
Fix MailboxMessageService thread leak/explosion

alvasw:
Each time when MailboxMessageService.onAdded(...) got called with multiple mailbox entries a new thread got created. That thread was never shutdown. This change explicitly creates a new Thread and sets its result with a SettableFuture. After its computation the thread terminates.


alejandrogarcia83 (approved):
utACK

PR #6552, opened 2023-02-02
Speed up loading and scrolling of trade history

stejbac:
This PR fixes a quadratic time bug while filtering the list of BSQ swap trades for their tx confirmation status, as needed by `ClosedTradesView`. It also fixes a (sort-of) cubic time bug caused by the repeated invocation of `ClosedTradableManager.getNumPastTrades` (once for each list item). It does this by caching the peer node address counts over all the closed trades & confirmed BSQ swaps, as well as caching a mapping of txIds to txs in `WalletService` to avoid repeated copying and scanning of the wallet tx list while checking for confirmation status by txId.

These issues were making the closed trades view load extremely slow on my Bisq instance (taking more than 20s), which has around 1300 past trades, a bit under half of which are BSQ swaps. It was also making scrolling and paging almost unusable. The following is a screenshot of the hotspots from JProfiler, prior to the changes:
Screenshot from 2023-02-02 10-53-51

### DIFF
--- a/core/src/main/java/bisq/core/btc/wallet/WalletService.java
+++ b/core/src/main/java/bisq/core/btc/wallet/WalletService.java
@@ -72,6 +72,7 @@ import org.bitcoinj.wallet.listeners.WalletReorganizeEventListener;
 
 import javax.inject.Inject;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMultiset;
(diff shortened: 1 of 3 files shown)


alejandrogarcia83:
Hey @stejbac  ,
thank you for the awesome contributions!! Caching will definitely improve the performance and many computations were redundant.

alejandrogarcia83 (approved):
ACK

PR #6556, opened 2023-02-05
ScryptUtil: Fix thread leak

alvasw:
Each time someone called the static method `ScryptUtil.deriveKeyWithScrypt(...)` a new worker pool was created. The worker pool was never shutdown.


alejandrogarcia83 (approved):
utACK

PR #6579, opened 2023-02-12
Further speed up transactions view load

stejbac:
This PR is a collection of optimisations to the Funds/Transactions display, via the `TransactionsView.updateList` method. The much earlier PR #5120 also attempted to optimise the transactions view load, but did not completely fix the main quadratic time issue (cost ~ transations count * tradable count, due to this many calls to `TransactionAwareTradable.isRelatedToTransaction`) and since then it has slowed down again a bit.

The main optimisation of this PR is to partition the wallet transactions pseudo-randomly into 64 buckets and use a bit array structure (which could be considered a degenerate Bloom filter with _k_, the number of hash functions, equal to 1), to filter the list of all tradable-transaction pairs prior to calling `isRelatedToTransaction` on each of them. While technically still quadratic time, it cuts down the number of calls (though not the runtime) roughly 40 times (at least when there are similar numbers of BSQ swap trades and escrow trades).

The other main speedup is fixing an issue where the set of all tradables was needlessly rebuild for every transaction listed. In addition to that, some further modest optimisations to `isRelatedToTransaction` and the `TransactionsListItem` constructor were made by introducing some appropriate simple caching, cleaning up the code slightly and removing an unused listener.

The following is a call tree of the transactions view load from JProfiler, prior to the optimisations (in which the load was taking 5-15 seconds on my old Bisq instance with ~1300 trades and ~3000 transactions):
Screenshot-2023-2-12 Call Tree

The following is the same call tree after the optimisations, repeatedly tabbing to the view 10 times (instead of just once):
Screenshot-2023-2-12 Call Tree(1)

### DIFF
--- a/core/src/main/java/bisq/core/util/FormattingUtils.java
+++ b/core/src/main/java/bisq/core/util/FormattingUtils.java
@@ -7,6 +7,7 @@ import bisq.core.monetary.Altcoin;
 import bisq.core.monetary.Price;
 
 import bisq.common.util.MathUtils;
+import bisq.common.util.Tuple3;
 
 import org.bitcoinj.core.Coin;
 import org.bitcoinj.core.Monetary;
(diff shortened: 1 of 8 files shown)


alejandrogarcia83 (approved):
utACK

Excellent job!

PR #6582, opened 2023-02-12
Remove unused ConnectionListener.onError(...) method

alvasw:


ripcurlx (approved):
utACK - Looked at the code parts and indeed it doesn't seem to be called in InboundConnection nor OutboundConnection.

PR #6593, opened 2023-02-28
Limit offer min/max amount deviation to 50%.

ghost:
Requested by @refund-agent2 in https://github.com/bisq-network/projects/issues/67

- At offer entry, limit the potential difference between min/max BTC amounts.
- Avoids cases where the security deposit is the majority loss in arbitration due to a tiny trade amount taken.

---

minAmountWarning


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6600, opened 2023-03-08
Speed up BSQ tx view load

stejbac:
Apply some fairly simple optimisations to speed up the BSQ tx list display under _DAO_ / _BSQ Wallet_ / _Transactions_:

1. Avoid needlessly calling `updateList()` twice in `BsqTxView`;
2. Build a `Map<String, BsqSwapTrade>` from txIds to swap trades, to fix a quadratic-time performance regression whereby all the tradables were repeatedly scanned to identify the possible matching swap trade for each BSQ tx;
3. Lazily load tooltips and confidence indicators, as some of the other list views do.

Additionally fix a related bug in the BTC transactions CSV export discovered while working on 3 above, and optimise the BTC funds deposit view slightly with some short circuit logic in `DepositListItem`, to make it more consistent with change 3 above.

--

With a large number (~1600) of BSQ txs in my Bisq wallet, the following call tree was produced by JProfiler by tabbing to the BSQ tx view 10 times:
Screenshot-2023-3-9 Call Tree

After the optimisations in this PR, the new call tree (from tabbing to the view 10 times) was produced:
Screenshot-2023-3-9 Call Tree(1)

Thus the average load time has gone from around 8.2 seconds to around 170 ms.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/components/TxConfidenceListItem.java
+++ b/desktop/src/main/java/bisq/desktop/components/TxConfidenceListItem.java
@@ -26,50 +26,78 @@ import bisq.core.btc.wallet.BsqWalletService;
 import org.bitcoinj.core.Transaction;
 import org.bitcoinj.core.TransactionConfidence;
 
+import com.google.common.base.Suppliers;
+
 import javafx.scene.control.Tooltip;
 
-import lombok.Data;
+import java.util.function.Supplier;
 
+import lombok.Getter;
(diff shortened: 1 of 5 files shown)


alvasw (approved):
Excellent work!

ACK

HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6604, opened 2023-03-11
Add missing segwit BSQ keychain path to wallet info

stejbac:
Add the new account path `44'/142'/1'` for segwit BSQ to the wallet info view, which was missed from PR #5109 making the wallet & UI changes to implement segwit BSQ. Also format all the paths found by iterating over the active wallet keychains, which are set to the constant lists defined in `BisqKeyChainGroupStructure`, instead of using hard-coded string literals. In this way, the four paths are only defined in one place (though it is extremely unlikely they would ever change).


ripcurlx (approved):
utACK

PR #6605, opened 2023-03-13
Feat: OCO Offers

ghost:
OCO (One cancels other) offers as defined by @apemithrandir
Ref: https://github.com/bisq-network/proposals/issues/402.  Demo [here](https://github.com/bisq-network/proposals/issues/402#issuecomment-1456438998).

### Features:

- Offers can be cloned using context menu ("Duplicate as OCO").
- Cloning allows the reserved amount to be shared amongst multiple offers.
- Cloned offers can be switched to different markets or payment methods using the existing "Edit Offer" functionality.
- Spam is prevented by requiring cloned offers be on different payment methods or markets before they can be enabled.
- Number of clones is limited to 10 per reserved UTXO.
- Offer clones are recognizable by a group ID in portfolio view.
- When an OCO offer is taken, all its clones are automatically removed (hence the name OCO).


ripcurlx:
As this is a high impact PR I'd like to see som ACK from core contributors like @HenrikJannsen,...

core/src/main/java/bisq/core/btc/model/AddressEntryList.java
```diff
@@ -210,7 +210,17 @@ public void swapToAvailable(AddressEntry addressEntry) {
 
         log.info("swapToAvailable addressEntry to swap={}", addressEntry);
         boolean setChangedByRemove = entrySet.remove(addressEntry);
-        boolean setChangedByAdd = entrySet.add(new AddressEntry(addressEntry.getKeyPair(),
+
+        // check if the ADDRESS still has any existing entries, only if not do the add to available.
+        boolean entryWithSameContextAlreadyExist = entrySet.stream().anyMatch(e -> {
+            if (addressEntry.getAddressString() != null) {
+                return addressEntry.getAddressString().equals(e.getAddressString()) &&
+                    addressEntry.getContext() == addressEntry.getContext();
```
HenrikJannsen: I guess you wanted: `ddressEntry.getContext() == e.getContext()`

core/src/main/java/bisq/core/btc/model/AddressEntryList.java
```diff
@@ -210,7 +210,17 @@ public void swapToAvailable(AddressEntry addressEntry) {
...
+            }
+            return false;
+        });
+        boolean setChangedByAdd = !entryWithSameContextAlreadyExist && entrySet.add(
```
HenrikJannsen: The setChangedByAdd flag should not be mixed with the new logic and its not very transparent that the entrySet.add is only called when entryWithSameContextAlreadyExist is false.
I think its better to return the code if entryWithSameContextAlreadyExist is true.

core/src/main/java/bisq/core/btc/model/AddressEntryList.java
```diff
@@ -210,7 +210,17 @@ public void swapToAvailable(AddressEntry addressEntry) {
 
         log.info("swapToAvailable addressEntry to swap={}", addressEntry);
         boolean setChangedByRemove = entrySet.remove(addressEntry);
-        boolean setChangedByAdd = entrySet.add(new AddressEntry(addressEntry.getKeyPair(),
+
+        // check if the ADDRESS still has any existing entries, only if not do the add to available.
+        boolean entryWithSameContextAlreadyExist = entrySet.stream().anyMatch(e -> {
```
HenrikJannsen: This will always return false as the addressEntry got removed in line 212. We need to do that check before the remove.

core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
```diff
@@ -630,6 +630,12 @@ public Optional<AddressEntry> getAddressEntry(String offerId,
                 .findAny();
     }
 
+    public AddressEntry createAddressEntryForOcoOffer(AddressEntry orgAddressEntry, String offerId) {
+        AddressEntry newEntry = new AddressEntry(orgAddressEntry.getKeyPair(), orgAddressEntry.getContext(), offerId, true);
```
HenrikJannsen: Isn't a check needed if the entry was already added?

core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
```diff
@@ -630,6 +630,12 @@ public Optional<AddressEntry> getAddressEntry(String offerId,
                 .findAny();
     }
 
+    public AddressEntry createAddressEntryForOcoOffer(AddressEntry orgAddressEntry, String offerId) {
```
HenrikJannsen: I think its better not to add higher level domain information to that lower level domain in the methodname. Oco might be also confusing if a code reader is not familiar with the feature.
Maybe just call it getOrCreateAddressEntry as its just another variation of creation method for an AddressEntry but with different params. A comment to add some context might be useful.

core/src/main/java/bisq/core/btc/Balances.java
```diff
@@ -110,11 +111,11 @@ private void updateAvailableBalance() {
     }
 
     private void updateReservedBalance() {
-        long sum = openOfferManager.getObservableList().stream()
-                .map(openOffer -> btcWalletService.getAddressEntry(openOffer.getId(), AddressEntry.Context.RESERVED_FOR_TRADE)
-                        .orElse(null))
-                .filter(Objects::nonNull)
-                .mapToLong(addressEntry -> btcWalletService.getBalanceForAddress(addressEntry.getAddress()).value)
+        long sum = btcWalletService.getAddressEntriesForOpenOffer().stream()
+                .collect(Collectors.toMap(AddressEntry::getAddress, p -> p, (p, q) -> p))
```
HenrikJannsen, line 115: Why you map to a Map and then use the keySet only? Just use the stream of getBalanceForAddress.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -381,20 +381,30 @@ public void onAwakeFromStandby() {
     public void placeOffer(Offer offer,
                            double buyerSecurityDeposit,
                            boolean useSavingsWallet,
+                           boolean useOco,
```
HenrikJannsen: Either use a more self explaining name or add a comment. oco is not understandable for any reader who is not familiar with the feature request.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -381,20 +381,30 @@ public void onAwakeFromStandby() {
...
+            errorMessageHandler.handleErrorMessage("PlaceOffer prevented because cloned OCO offers count is " + numClones);
+            return;
+        }
+
         Coin reservedFundsForOffer = createOfferService.getReservedFundsForOffer(offer.getDirection(),
                 offer.getAmount(),
                 buyerSecurityDeposit,
                 createOfferService.getSellerSecurityDepositAsDouble(buyerSecurityDeposit));
 
+        offer.setPriceFeedService(priceFeedService);
```
HenrikJannsen: This is set already for normal offers. The OcoOffer should set that when its created to have a similar structure. If not possible (have not looked into all the code yet) then only apply it for ocoOffers with a if statement to keep changes to non-oco offers minimal.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -409,6 +419,7 @@ public void placeOffer(Offer offer,
                 model,
                 transaction -> {
                     OpenOffer openOffer = new OpenOffer(offer, triggerPrice);
+                    openOffer.setState(useOco ? OpenOffer.State.DEACTIVATED : OpenOffer.State.AVAILABLE);
```
HenrikJannsen: Better use a if statement and only apply it to ocoOffers to reduce changes for normal offers.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -465,6 +476,10 @@ public void activateOpenOffer(OpenOffer openOffer,
             return;
         }
 
+        if (isSpam(openOffer)) {
+            return;
```
HenrikJannsen, line 485: I guess an error message should be set here.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -595,13 +611,27 @@ private void onRemoved(OpenOffer openOffer, ResultHandler resultHandler, Offer o
...
+                offerBookService.removeOffer(openOffer.getOffer().getOfferPayloadBase(),
+                        () -> log.trace("Successfully removed offer"),
+                        log::error);
+            });
+        } else {
+            // offer taken may have been OCO, in which case all its clones need to be removed
+            getOpenOffersByMakerFeeTxId(offer.getOfferFeePaymentTxId()).forEach(openOffer -> {
+                removeOpenOfferFromList(openOffer);
+                openOffer.setState(OpenOffer.State.CLOSED);
+                if (!offer.getId().equalsIgnoreCase(openOffer.getId())) {
```
HenrikJannsen: equalsIgnoreCase is not needed as we use strict case requirement for offer ID

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -626,6 +656,11 @@ public Optional<OpenOffer> getOpenOfferById(String offerId) {
         return openOffers.stream().filter(e -> e.getId().equals(offerId)).findFirst();
     }
 
+    public List<OpenOffer> getOpenOffersByMakerFeeTxId(String makerFeeTxId) {
+        String safeSearch = makerFeeTxId == null ? "" : makerFeeTxId;
+        return openOffers.stream().filter(e -> !e.getOffer().isBsqSwapOffer() && e.getOffer().getOfferFeePaymentTxId().equals(safeSearch)).collect(Collectors.toList());
```
HenrikJannsen: Better break stream to multiline as its quite long.
safeSearch is not a good var name and i think its easier to just add the null check in the filter.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -1143,7 +1178,33 @@ private boolean isBsqSwapOfferLackingFunds(OpenOffer openOffer) {
                 openOffer.isBsqSwapOfferHasMissingFunds();
     }
 
+    public boolean isSpam(OpenOffer newOffer) {
+        // an offer is spam if the user has another open offer on the same ccy, payment method, and reserved UTXO
+        long matchingOffers = openOffers.stream()
+                .filter((openOffer -> !openOffer.getOffer().isBsqSwapOffer()))
+                .filter(openOffer -> !openOffer.isDeactivated())
+                .filter(openOffer -> !openOffer.getShortId().equalsIgnoreCase(newOffer.getShortId()))
```
HenrikJannsen: Why use shortId instead of Id?

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -1143,7 +1178,33 @@ private boolean isBsqSwapOfferLackingFunds(OpenOffer openOffer) {
...
+                .filter(openOffer1 -> {
+                    Offer newOffer1 = newOffer.getOffer();
```
HenrikJannsen: Its confusing to name the openOffer from the param newOffer and then the Offer var newOffer1.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -1143,7 +1178,33 @@ private boolean isBsqSwapOfferLackingFunds(OpenOffer openOffer) {
                 openOffer.isBsqSwapOfferHasMissingFunds();
     }
 
+    public boolean isSpam(OpenOffer newOffer) {
+        // an offer is spam if the user has another open offer on the same ccy, payment method, and reserved UTXO
+        long matchingOffers = openOffers.stream()
+                .filter((openOffer -> !openOffer.getOffer().isBsqSwapOffer()))
+                .filter(openOffer -> !openOffer.isDeactivated())
+                .filter(openOffer -> !openOffer.getShortId().equalsIgnoreCase(newOffer.getShortId()))
+                .filter(openOffer1 -> {
```
HenrikJannsen: Why use openOffer1 instead of openOffer?

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -1143,7 +1178,33 @@ private boolean isBsqSwapOfferLackingFunds(OpenOffer openOffer) {
                 openOffer.isBsqSwapOfferHasMissingFunds();
     }
 
+    public boolean isSpam(OpenOffer newOffer) {
+        // an offer is spam if the user has another open offer on the same ccy, payment method, and reserved UTXO
```
HenrikJannsen: ccy is not clear to a non domain expert.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -1143,7 +1178,33 @@ private boolean isBsqSwapOfferLackingFunds(OpenOffer openOffer) {
...
+                        offer1.getOfferFeePaymentTxId().equalsIgnoreCase(newOffer1.getOfferFeePaymentTxId()) &&
+                        offer1.getPaymentMethodId().equalsIgnoreCase(newOffer1.getPaymentMethodId()) &&
+                        offer1.getCounterCurrencyCode().equalsIgnoreCase(newOffer1.getCounterCurrencyCode()) &&
+                        offer1.getBaseCurrencyCode().equalsIgnoreCase(newOffer1.getBaseCurrencyCode());
+                })
+                .count();
+        if (matchingOffers > 0) {
+            log.info("{} is considered spam", newOffer.getShortId());
+        }
+        return matchingOffers > 0;
```
HenrikJannsen: A anyMatch would do it as well. The info log would be in the caller and can be a warning log.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -1143,7 +1178,33 @@ private boolean isBsqSwapOfferLackingFunds(OpenOffer openOffer) {
                 openOffer.isBsqSwapOfferHasMissingFunds();
     }
 
+    public boolean isSpam(OpenOffer newOffer) {
```
HenrikJannsen: isSpam is a too generic and overloaded term. Better use some more descriptive method name or something with "isValid"

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOfferListItem.java
```diff
@@ -134,6 +134,14 @@ public String getTriggerPriceAsString() {
         }
     }
 
+    public String getOcoGroupAsString() {
+        Offer offer = getOffer();
+        if (offer.isBsqSwapOffer()) {
+            return "";
+        }
+        return offer.getOfferFeePaymentTxId().substring(0, 4);
```
HenrikJannsen: Does this need to be unique? There might be collisions with using only 4 chars.

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersView.java
```diff
@@ -101,6 +107,7 @@ private enum ColumnNames {
         VOLUME(Res.get("shared.amountMinMax")),
         PAYMENT_METHOD(Res.get("shared.paymentMethod")),
         DIRECTION(Res.get("shared.offerType")),
+        GROUP("Group"),
```
HenrikJannsen: Use translation strings

HenrikJannsen: Group is also very generic. Maybe find a better descriptive name.

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersView.java
```diff
@@ -197,20 +209,27 @@ public void initialize() {
...
         tableView.getSortOrder().add(dateColumn);
 
         tableView.setRowFactory(
                 tableView -> {
                     final TableRow<OpenOfferListItem> row = new TableRow<>();
                     final ContextMenu rowMenu = new ContextMenu();
                     MenuItem duplicateItem = new MenuItem(Res.get("portfolio.context.offerLikeThis"));
-                    duplicateItem.setOnAction((event) -> onDuplicateOffer(row.getItem().getOffer()));
+                    duplicateItem.setOnAction((event) -> onDuplicateOffer(row.getItem()));
+                    MenuItem duplicateItemOco1 = new MenuItem("Duplicate as OCO");
```
HenrikJannsen: Use translation stings. Maybe find a more user friendly term (batch-offer,...).

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersView.java
```diff
@@ -299,6 +319,14 @@ protected void activate() {
 
     private void updateNumberOfOffers() {
         numItems.setText(Res.get("shared.numItemsLabel", sortedList.size()));
+        groupColumn.setVisible(ocoIsInUse());
```
HenrikJannsen: Why do that in a method called `updateNumberOfOffers`? It has nothing to do with updating the Number of offers. We need to take more care of code quality otherwise Bisq 1 code gets un-manageable....

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersView.java
```diff
@@ -299,6 +319,14 @@ protected void activate() {
...
+    }
+
+    private boolean ocoIsInUse()
+    {
+        return sortedList.stream()
```
HenrikJannsen: Better use the domain data and not a display string for logical results.

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersView.java
```diff
@@ -581,6 +673,36 @@ public void updateItem(final OpenOfferListItem item, boolean empty) {
...
+                            public void updateItem(final OpenOfferListItem item, boolean empty) {
+                                super.updateItem(item, empty);
+                                getStyleClass().removeAll("offer-disabled");
+                                if (item != null) {
+                                    if (item.isNotPublished()) getStyleClass().add("offer-disabled");
+                                    Label label = new AutoTooltipLabel(item.getOcoGroupAsString());
+                                    if (openOfferManager.isSpam(item.getOpenOffer())) {
+                                        Text icon = getRegularIconForLabel(MaterialDesignIcon.EYE_OFF, label, "opaque-icon");
+                                        label.setContentDisplay(ContentDisplay.RIGHT);
+                                        Tooltip.install(icon, new Tooltip("Change ccy or payment method to enable offer."));
```
HenrikJannsen: Use translations

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersView.java
```diff
@@ -359,32 +387,37 @@ private void onActivateOpenOffer(OpenOffer openOffer) {
...
-                        Res.get("popup.warning.removeOffer", item.getMakerFeeAsString());
-                new Popup().warning(message)
-                        .actionButtonText(Res.get("shared.removeOffer"))
-                        .onAction(() -> doRemoveOpenOffer(openOffer))
-                        .closeButtonText(Res.get("shared.dontRemoveOffer"))
-                        .dontShowAgainId(key)
-                        .show();
-            } else {
+            if (openOfferManager.safeRemovalOfOcoClone(openOffer)) {
                 doRemoveOpenOffer(openOffer);
```
HenrikJannsen, line 390: Why do we skip the popup here?

HenrikJannsen (changes requested):

core/src/main/java/bisq/core/btc/model/AddressEntryList.java
ghost: The remove always needs to be done first as was the case before and now.  Say you had 5 offers sharing a reserved amount, then you cancel one offer.  The canceled offer's address entry is removed, the other 4 remain open and a new addressEntry(AVAILABLE) should not be added.

core/src/main/java/bisq/core/btc/Balances.java
ghost: This is so that a reserved amount shared across offers is not summed per address entry, but per address.

desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersView.java
ghost: Because there is no cost to removing a cloned offer.  In the demonstration video, it shows the user cloning an offer 5 times and then deleting a few that were not needed.  Having a popup at every delete would be too annoying, so the software checks if the removal is safe (i.e. does not incur a potential cost).  Conversely, an unsafe removal is one that would cause loss of paid fees and that warns the user with a popup.

core/src/main/java/bisq/core/btc/model/AddressEntryList.java
```diff
@@ -209,21 +209,25 @@ public void swapToAvailable(AddressEntry addressEntry) {
         }
 
         log.info("swapToAvailable addressEntry to swap={}", addressEntry);
-        boolean setChangedByRemove = entrySet.remove(addressEntry);
-
-        // check if the ADDRESS still has any existing entries, only if not do the add to available.
-        boolean entryWithSameContextAlreadyExist = entrySet.stream().anyMatch(e -> {
+        if (entrySet.remove(addressEntry)) {
```
HenrikJannsen, line 212: Do we want to remove the address entry if entryWithSameContextStillExists is true?

core/src/main/java/bisq/core/btc/wallet/BtcWalletService.java
```diff
@@ -630,10 +630,19 @@ public Optional<AddressEntry> getAddressEntry(String offerId,
                 .findAny();
     }
 
-    public AddressEntry createAddressEntryForOcoOffer(AddressEntry orgAddressEntry, String offerId) {
-        AddressEntry newEntry = new AddressEntry(orgAddressEntry.getKeyPair(), orgAddressEntry.getContext(), offerId, true);
-        addressEntryList.addAddressEntry(newEntry);
-        return newEntry;
+    // when a new offer needs to share the reserved amount info from parent offer's address entry
+    public AddressEntry getOrCreateAddressEntry(AddressEntry orgAddressEntry, String offerId) {
```
HenrikJannsen, line 634: The param name orgAddressEntry is unclear. If it means original better spell it out.

HenrikJannsen: Maybe better name the method getOrCreateSharedAddressEntry

core/src/main/java/bisq/core/btc/Balances.java
HenrikJannsen: Then map from addressEntry to address and use streams distinct method.

core/src/main/java/bisq/core/offer/placeoffer/bisq_v1/tasks/CloneMakerFeeOco.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    // AddressEntry and TxId are not linked, so do a reverse lookup
+    private Optional<String> getTxIdFromAddress(BtcWalletService walletService, Address address) {
+        List<Transaction> txns = walletService.getRecentTransactions(10, false);
+        for (Transaction txn : txns) {
+            for (TransactionOutput output : txn.getOutputs()) {
+                if (walletService.isTransactionOutputMine(output) && WalletService.isOutputScriptConvertibleToAddress(output)) {
+                    String addressString = WalletService.getAddressStringFromOutput(output);
+                    assert addressString != null;
```
HenrikJannsen, line 71: We do not use assert but checkNotNull.

core/src/main/java/bisq/core/offer/placeoffer/bisq_v1/tasks/CloneMakerFeeOco.java
```diff
@@ -0,0 +1,81 @@
...
+                    // make sure the output is still unspent
+                    if (addressString.equalsIgnoreCase(address.toString()) && output.getSpentBy() == null) {
```
HenrikJannsen, line 73: Why equalsIgnoreCase?

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -657,8 +662,10 @@ public Optional<OpenOffer> getOpenOfferById(String offerId) {
     }
 
     public List<OpenOffer> getOpenOffersByMakerFeeTxId(String makerFeeTxId) {
-        String safeSearch = makerFeeTxId == null ? "" : makerFeeTxId;
-        return openOffers.stream().filter(e -> !e.getOffer().isBsqSwapOffer() && e.getOffer().getOfferFeePaymentTxId().equals(safeSearch)).collect(Collectors.toList());
+        return openOffers.stream()
+                .filter(e -> !e.getOffer().isBsqSwapOffer() && e.getOffer().getOfferFeePaymentTxId()
+                        .equals(makerFeeTxId == null ? "" : makerFeeTxId))
```
HenrikJannsen, line 667: Right side of equals can be null.

core/src/main/java/bisq/core/offer/OpenOfferManager.java
```diff
@@ -1178,33 +1185,23 @@ private boolean isBsqSwapOfferLackingFunds(OpenOffer openOffer) {
...
-        }
-        return matchingOffers > 0;
+    public boolean canBeEnabled(Offer newOffer) {
+        // does the user have another open offer on the same currency, payment method, and reserved UTXO?
+        return openOffers.stream().noneMatch(e ->
+                        !e.getOffer().isBsqSwapOffer() &&
+                        !e.isDeactivated() &&
+                        !e.getId().equals(newOffer.getId()) &&
+                        e.getOffer().getOfferFeePaymentTxId().equals(newOffer.getOfferFeePaymentTxId()) &&
+                        e.getOffer().getPaymentMethodId().equalsIgnoreCase(newOffer.getPaymentMethodId()) &&
```
HenrikJannsen, line 1195: No need to use equalsIgnoreCase. It suggests that the strings do not have a defined case, but those data have not arbitrary upper/lower case.

HenrikJannsen:
After the cloned offers have been created it is unclear what need to be edited to allow activation. There is no feedback at clicking activate and changing the price does not enable it as well, so a user not familiar with the feature would be confused.
We need to shop popups when activating but not permitted to give info whats needed. Also after having created then we should show a popup with info why they are deactivated and what need to be done (e.g. change payment method).

There should be an icon as well next to edit and duplicate offer. Not sure if the 5x option is needed. There will not be that many users who have that many payment methods and its just a simple click to clone. The right click context menu is not a standard UX feature Bisq uses so it would be hidden to most users.

core/src/main/java/bisq/core/btc/model/AddressEntryList.java
HenrikJannsen: Looked closer into that now. The check with entryWithSameContextAlreadyExist is not needed and we can revert it as we have a hashset and the add will not add if we try to add the same entry.

ghost:
I don't think there should be another icon next to edit and duplicate offer - that crowds the screen and confuses casual users.  This feature is not for casual users, it is for advanced market makers.  As such I expect the people using this feature would be the ones who requested it and/or read about it in the forums or bisq.wiki.  Those users being experienced I think they do not need to be spoon fed by popups either.

HenrikJannsen (changes requested):

core/src/main/java/bisq/core/btc/model/AddressEntryList.java
HenrikJannsen: Ah yes, true.

apemithrandir:
Has this feature been included in any of the recent releases? @jmacxx

MwithM:
Yes. https://github.com/bisq-network/bisq/pull/6675
https://bisq.wiki/Cloning_an_offer

PR #6608, opened 2023-03-19
Tests for OfferUtil

helixx87:
A few unit tests for static methods from [OfferUtil](https://github.com/bisq-network/bisq/blob/381985ec3ae09241e06414ad266564c41ef967af/core/src/main/java/bisq/core/offer/OfferUtil.java).


alejandrogarcia83 (approved):
utACK

PR #6615, opened 2023-03-23
BSQ trade fee validation NACK if tx unconfirmed for a long time

### LINKED ISSUE #6603: Offers with invalid maker fees are shown live on the Bisq offer book resulting in failed trades and lost fees for the taker
### Description

When a maker with an out of sync bitcoin wallet creates a trade to buy or seller bitcoin if their maker transaction fails due to a problem with their wallet the trade is still published live in the Bisq offer book.

This results in a taker coming along taking the offer only for the trade to fail and them to lose miner and trade fees.

I believe their was a filter in place previously that prevented this by checking the offers to validate the maker transactions before publishing the offer to the Bisq offer book.

For examples of these types of failed trades please see:

4 days ago: https://github.com/bisq-network/support/issues/1352
2 weeks ago: https://github.com/bisq-network/support/issues/1348

Both of these trades are avoidable if the trades would have been validated before being shown on Bisq.

#### Version

v1.9.9

### Steps to reproduce

- Make an offer with an out of sync bitcoin wallet
- You will not pay any trade fees or mining fees 
- Your offer will be shown live on Bisq for another user to take

### Expected behaviour

- Make an offer with an out of sync bitcoin wallet
- You will not pay any trade fees or mining fees 
- Your offer will not be shown live on Bisq for another user to take because it will fail the validation test

### Actual behaviour

Offers are shown in the offer book that will fail when taken.

ghost:
Fixes #6603 

When a user's wallet is out of sync they may create offers where the fee tx will never be confirmed, and would cause a failed trade if offer is taken.

Maker fees paid in BSQ cannot be checked until the tx confirms and is in the DAO database:

- If an offer is confirmed, check the BSQ fee as usual.
- If an offer is unconfirmed for more than 8 hours assume there is a wallet problem and NACK the validation, which will prevent the offer from being taken.
- If an offer confirms after 8 hours the validation will be able to check the fee.

Taker fees paid in BSQ are checked after the 1st confirmation.

### DIFF
--- a/core/src/main/java/bisq/core/provider/mempool/TxValidator.java
+++ b/core/src/main/java/bisq/core/provider/mempool/TxValidator.java
@@ -63,7 +63,6 @@ public class TxValidator {
     private Coin amount;
     @Nullable
     private Boolean isFeeCurrencyBtc;
-    @Nullable
     private Long chainHeight;
     @Setter
     private String jsonTxt;
@@ -79,6 +78,7 @@ public class TxValidator {
         this.amount = amount;
         this.isFeeCurrencyBtc = isFeeCurrencyBtc;
         this.feePaymentBlockHeight = feePaymentBlockHeight;
+        this.chainHeight = (long) daoStateService.getChainHeight();
         this.filterManager = filterManager;
         this.errorList = new ArrayList<>();
         this.jsonTxt = "";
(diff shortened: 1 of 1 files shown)


ripcurlx (approved):
utACK

PR #6638, opened 2023-04-07
Update data stores for v1.9.10

ripcurlx:
I've updated the data stores from my full node and found a difference in the BsqBlocks that might have caused the error @alejandrogarcia83 experienced.

I've tested this branch with a fresh client from IntelliJ and also from a new built macOS binary and a fresh client.
I waited for a full sync and restarted the client afterwards again and everything seems to be working for me now.

@alejandrogarcia83 Could you please test this PR with your setup, if you still run into this issue?


alejandrogarcia83 (approved):
utACK

PR #6640, opened 2023-04-08
Update offer book UI

### LINKED ISSUE #6639: Market Depth (order book) graph shows inverted colors
### Description
Market Depth shows **GREEN** for sell offers and **RED** for buy offers, the correct is the opposite. 
Market Depth (order book) graph shows inverted colors. This is very confusing for who uses the graph (traders).

image

#### Version
v1.9.9

### Steps to reproduce
1. Open app
2. Click on Market
3. Check Order Book tab

### Expected behaviour
Should show **RED** for **selling offers** and **GREEN** for **buying offers**. As you can see in any other platform that provides the market depth:

image

#### Device or machine
Mac OS

helixx87:
Fixes #6639

PR inverts colors on market depth chart, as suggested in reported issue:
chart-colors

[Second commit](https://github.com/bisq-network/bisq/commit/50e3b44f7fb686288b0ccca5d5ec655233d9fcb3) contains rename of class from `#charts` into `#charts-offer-book` - `#charts` seemed to be too generic.

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -304,8 +304,8 @@ market.tabs.spreadPayment=Offers by Payment Method
 market.tabs.trades=Trades
 
 # OfferBookChartView
-market.offerBook.sellOffersHeaderLabel=Sell {0} to
-market.offerBook.buyOffersHeaderLabel=Buy {0} from
+market.offerBook.sellOffersHeaderLabel=Sell offers
+market.offerBook.buyOffersHeaderLabel=Buy offers
 market.offerBook.buy=I want to buy bitcoin
 market.offerBook.sell=I want to sell bitcoin
 
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -306,8 +306,6 @@ market.tabs.trades=Trades
 # OfferBookChartView
 market.offerBook.sellOffersHeaderLabel=Sell offers
 market.offerBook.buyOffersHeaderLabel=Buy offers
-market.offerBook.buy=I want to buy bitcoin
-market.offerBook.sell=I want to sell bitcoin
 
 # SpreadView
 market.spread.numberOfOffersColumn=All offers ({0})
--- a/desktop/src/main/java/bisq/desktop/bisq.css
+++ b/desktop/src/main/java/bisq/desktop/bisq.css
@@ -1756,19 +1756,19 @@ textfield */
 }
 
 #charts .default-color0.chart-area-symbol {
-    -fx-background-color: -bs-sell, -bs-background-color;
-}
-
-#charts .default-color1.chart-area-symbol, #charts-dao .default-color0.chart-area-symbol {
     -fx-background-color: -bs-buy, -bs-background-color;
 }
 
+#charts .default-color1.chart-area-symbol, #charts-dao .default-color0.chart-area-symbol {
+    -fx-background-color: -bs-sell, -bs-background-color;
+}
+
 #charts .default-color0.chart-series-area-line {
-    -fx-stroke: -bs-sell;
+    -fx-stroke: -bs-buy;
 }
 
 #charts .default-color1.chart-series-area-line, #charts-dao .default-color0.chart-series-area-line {
-    -fx-stroke: -bs-buy;
+    -fx-stroke: -bs-sell;
 }
 
 /* The .chart-line-symbol rules change the color of the legend symbol */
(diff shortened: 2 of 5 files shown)


pazza83:
Thanks, I think that looks more confusing than the existing.

helixx87:
> Thanks, I think that looks more confusing than the existing.

That's true, however I think it's a good direction. Let's discuss it a bit on original issue - for now I'm marking this PR as draft.

helixx87:
I moved it a bit forward - now offer book looks like that:
offer-book

I think it's more similar to how other platforms works.

HenrikJannsen (dismissed):
utACK

HenrikJannsen:
> utACK

my utACk refers to the code change not the conceptual question. I leave it to others to discuss that.

HenrikJannsen:
If that PR requires more discussion can we convert it to draft?

helixx87:
From my perspective this pull request is ready to be merged - implemented solution brings us closer to generally accepted standards in the trading world.

HenrikJannsen:
@pazza83 What is your opinion?

ghost:
NACK - 

1. Bring up market view.  Colors look good. :heavy_check_mark: 
2. Buy offers are on the left pane, sell offers are on the right. :heavy_check_mark: 
3. Click in the right pane, expressing my interest in an offer that **I want to take/buy**. :heavy_check_mark: 
4. I'm taken to a list of offers, inviting me to sell. :x:

pazza83:
> @pazza83 What is your opinion?

I am happy with the color changes per say. Just wanted to see how it would look in practice prior to approving it, but I suppose it is not really needed if other can check on the code and are happy with it.

Sounds like there might be some work to do with where the buttons take you when clicked.

HenrikJannsen:
We have to be careful with changes here. The color codes and BUY/SELL usage have a long history of changes and it not always turned out to become better. Bisq is not a the typical pro-market UI but rather a P2P exchange which comes with different domain terms and behaviour. Pro finance pages and its terms are not always intuitive for non-experts. Bisq users are more in the non-finance expert group I would assume. Still of course we should find the best compromise to serve both, just wanted to voice some caution as its a complex and difficult UX problem.

helixx87:
> 4. I'm taken to a list of offers, inviting me to sell.

@jmacxx I agree this behavior is misleading - would it be enough to change direction [here](https://github.com/bisq-network/bisq/blob/b622620cac6b47cbac4a0537f3a147dc68ec942c/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java#L283-L284) or we should invoke take action after click?

ghost:
> I agree this behavior is misleading - would it be enough to change direction [here](https://github.com/bisq-network/bisq/blob/b622620cac6b47cbac4a0537f3a147dc68ec942c/desktop/src/main/java/bisq/desktop/main/market/offerbook/OfferBookChartView.java#L283-L284) or we should invoke take action after click?

@helixx87 Trying it now, and the former suggestion would address that concern.  :heavy_check_mark: 

But there are nagging issues:
 
- BTC / fiat market view: where it says `Sell offers`, there is a big red button that says `SELL BTC`.  That makes me feel very uneasy about what I'm doing, because I'm trying to **buy** btc in an area where it says `SELL BTC`.

image


- BTC / altcoin market view: I think the by adopting the wording Buy offers / Sell offers and changing the buttons text to always BTC you've confused the UX.  For example, screenshot comparison below, what is a buy offer?  BTC or LTC?  In the current Bisq, it says clearly when you click on the right you are buying LTC.

**Current Bisq, LTC market:**
image

**Your PR, LTC market:**
image


---

So, I remain negative about this change because when trying to use Bisq in the way I'm accustomed, it jarrs my brain.  

I think if this PR were to change the histogram color **only** (as requested [here](https://github.com/bisq-network/bisq/issues/6672)), it may have a chance.

helixx87:
> I think if this PR were to change the histogram color **only** (as requested https://github.com/bisq-network/bisq/issues/6672), it may have a chance.

First two commits was doing that, after [suggestion](https://github.com/bisq-network/bisq/pull/6640#issuecomment-1500983584) from @pazza83 I moved it a bit forward and reorganized also buttons and labels - I can remove these additional commits. 

I'm looking for some consensus here - I am also open to closing this pull request if that is the preferred course of action by team.

pazza83:
Hi @helixx87 

Are you able to post a demo (for example a screen recording gif or video) of the issue on the issue: https://github.com/bisq-network/bisq/issues/6639

That might allow for people like myself who are unable to run Bisq from source with your changes to give feedback.

alvasw:
I found some time to look at your changes. First, imagine you want to buy BTC and look at the buy offers. Pressing the "BUY BTC" button takes you to the Buy BTC tab (correct behavior). However, the data in the table is from the SELL BTC tab (the column in the header even says Buyer) and clicking on an entry in the table takes me to the SELL BTC tab.

offerbook_swap_pr

Second, it's confusing when I change the currency to BSQ or XMR. The buttons still say "BUY BTC" and "SELL BTC", but I want to buy or sell BSQ/XMR.

PR #6641, opened 2023-04-08
Fix close button behaviour on successful offer creation popup

### LINKED ISSUE #6629: Move to portfolio when clicked close in popup
After creating an offer there is a popup asking to move to open offers in the portfolio screen. When clicking close it should do nothing, but it still moves to portfolio screen.

helixx87:
Fixes #6629

Make close button not moving to portfolio:
cancel

(as an alternative we can hide close button)


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6644, opened 2023-04-10
Make processing burningman accounting data optional

HenrikJannsen:
Processing burningman accounting data an take quite some time and CPU. As most Bisq users are likely not interested in those data, this PR deactivates it by default.
Users who are interested in that data can enable it in the settings at the DAO section (requires restart) or when selecting a BM in the table they get a popup to enable it.
If not enabled the UI shows a "(accounting data not updated)" post fix to the "Selected contributor" header. The displayed data is the one from the last update or the one shipped by the release.
If enabled and still processing data we show a "(still processing data...)" post fix to the "Selected contributor" header.
We do not use a busy animation as this is quite CPU intense.
Once all data is processed the post fix is removed.


sqrrm (approved):
utACK

core/src/main/java/bisq/core/dao/burningman/accounting/node/AccountingNode.java
```diff
@@ -160,6 +164,10 @@ public void onParseBlockChainComplete() {
 
     @Override
     public void addListeners() {
+        if (!preferences.isProcessBurningManAccountingData()) {
```
ghost, line 167: A problem (bug) here is it prevents seednodes from processing Accounting data!  :scream:

alvasw: I can confirm the bug!

HenrikJannsen: Saw it too late, but seem its fixed by @jmacxx already

PR #6646, opened 2023-04-11
Support "all" attribute in getoffers API method

### LINKED ISSUE #6645: CLI doesn't return all offers
When I invoke:
```
./bisq-cli getoffers
```
I'm getting only offers that matches my accounts. I'd like to have possibility to get all of the offers.

In the UI we have switch for that:
matching-my-accounts-off

In the API we could add boolean parameter, for example:
```
./bisq-cli getoffers --all=true
```

helixx87:
Fixes #6645

Possibility to return all offers, not only these matching my accounts:
```
./bisq-cli --password=xyz getoffers --direction=sell --currency-code=EUR --all=true
```


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6670, opened 2023-04-30
Allow leniency of 50% when validating trade fees.

### LINKED ISSUE #6669: Offer keeps being disabled
### Description
A created offer is being disabled after a few seconds from enabling it, with the message:
` [JavaFX Application Thread] INFO  b.c.o.b.TriggerPriceService: Deactivating open offer xxxx due to mempool validation`

#### Version

1.9.10

### Steps to reproduce

1. Create an offer. Not every offer will behave like this.

### Expected behaviour

Enabled offer should stay enabled.

### Actual behaviour

Enabled offer disables itself after a few seconds. Sometimes a dozen or so.

#### Device or machine

Linux

#### Additional info

I can send logs in a private message, I'm "akira45" on Matrix.

ghost:
Fixes #6669.


pazza83:
@jmacxx 

Thanks for this. Assuming the trade fee filter should still be updated every time the trade fees are updated, and when implemented only changes in fees over 50% in BTC or BSQ terms would cause the trades posted in the interim to not show?

HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6679, opened 2023-05-06
Remove edit button for BSQ offers

### LINKED ISSUE #6668: Edit button/pen symbol stops working for altcoin offers requiring restart
Sometimes the EDIT button stops working for altcoin offers. If I click on it for my own offers when in the SELL/BUY views, I just get switched to the MY OPEN OFFERS tab and if I click the pen symbol in there nothing happens whatsoever. It only seems to happen when I mess around with BSQ offers so it seems BISQ gets confused and starts treating every offer as a BSQ offer when it comes to edit ability. The only remedy is to restart BISQ.
Version: 1.9.9

napoly:
Fixes #6668

We use a simple 'if' check to determine if our offer is for a BSQ swap. If it is, we do not show the button for editing.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBookView.java
+++ b/desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBookView.java
@@ -1169,13 +1169,18 @@ abstract public class OfferBookView<R extends GridPane, M extends OfferBookViewM
                                         button.setStyle(CssTheme.isDarkTheme() ? "-fx-text-fill: white" : "-fx-text-fill: #444444");
                                         button.setOnAction(e -> onRemoveOpenOffer(offer));
 
-                                        iconView2.setId("image-edit");
-                                        button2.updateText(Res.get("shared.edit"));
-                                        button2.setId(null);
-                                        button2.setStyle(CssTheme.isDarkTheme() ? "-fx-text-fill: white" : "-fx-text-fill: #444444");
-                                        button2.setOnAction(e -> onEditOpenOffer(offer));
-                                        button2.setManaged(true);
-                                        button2.setVisible(true);
+                                        if (offer.isBsqSwapOffer()) {
+                                            button2.setManaged(false);
+                                            button2.setVisible(false);
+                                        } else {
+                                            iconView2.setId("image-edit");
+                                            button2.updateText(Res.get("shared.edit"));
+                                            button2.setId(null);
+                                            button2.setStyle(CssTheme.isDarkTheme() ? "-fx-text-fill: white" : "-fx-text-fill: #444444");
+                                            button2.setOnAction(e -> onEditOpenOffer(offer));
+                                            button2.setManaged(true);
+                                            button2.setVisible(true);
+                                        }
                                     } else {
                                         boolean isSellOffer = OfferViewUtil.isShownAsSellOffer(offer);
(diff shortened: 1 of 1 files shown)


desktop/src/main/java/bisq/desktop/main/offer/offerbook/OfferBookView.java
```diff
@@ -1169,13 +1169,15 @@ public void updateItem(final OfferBookListItem item, boolean empty) {
...
                                         button.setOnAction(e -> onRemoveOpenOffer(offer));
 
-                                        iconView2.setId("image-edit");
-                                        button2.updateText(Res.get("shared.edit"));
-                                        button2.setId(null);
-                                        button2.setStyle(CssTheme.isDarkTheme() ? "-fx-text-fill: white" : "-fx-text-fill: #444444");
-                                        button2.setOnAction(e -> onEditOpenOffer(offer));
-                                        button2.setManaged(true);
-                                        button2.setVisible(true);
+                                        if (!offer.isBsqSwapOffer()) {
```
helixx87: We probably need to do something like that:
```
if (!offer.isBsqSwapOffer()) {
    // ...
} else {
  button2.setManaged(false);
  button2.setVisible(false);
}
```
IIRC `updateItem` is re-rendering items on list during scrolling.

napoly: Thanks for reviewing! I refactored accordingly.

helixx87: Thank you - looks good now!

helixx87 (approved):

HenrikJannsen (changes requested):
NACK
As far I remember the edit button worked for BSQ offers as @pazza83 commented in https://github.com/bisq-network/bisq/issues/6668#issuecomment-1527427004. We should figure out what caused it to not work anymore.

napoly:
I did some digging since I'm new to the code base, and it appears that the functionality is not available. This is supported by the fact that it was removed from even the open offers, as shown here: https://github.com/bisq-network/bisq/blob/master/desktop/src/main/java/bisq/desktop/main/portfolio/openoffer/OpenOffersView.java#L1007

ghost (approved):

alvasw (changes requested):
NACK
After further debugging, I can confirm that the edit BSQ offer functionality is present.

When I press the edit button on a BSQ offer first it opens the edit tab:
bisq_bsq_edit


However, there are several bugs:
- Pressing to fast on the edit button trigger a `java.util.ConcurrentModificationException` exception
- Pressing the edit button on an BSQ offer breaks other offers edit buttons
- Sometimes the edit button doesn't even open the edit tab

ghost:
**This bug needs to get fixed, and @napoly has done correctly what was requested/required.**

I will second what @ripcurlx said in https://github.com/bisq-network/bisq/issues/6668#issuecomment-1519674678 that  there is **no edit of BSQ swap offers** and the best thing would be to remove the edit button for BSQ offers, as implemented above.  The screenshot Alva produced is the edit dialog for original (multisig) offers which is incompatible with BSQ swaps.  There's not much sense trying to adapt it for BSQ swaps, as the offer UI code is complex / fragile / sensitive to change.

alvasw:
@jmacxx You are totally right!! Sorry.

alejandrogarcia83 (approved):
utACK

PR #6680, opened 2023-05-06
API: gettransactions method

helixx87:
```
./bisq-cli --password=xyz gettransactions
```

### DIFF
--- a/cli/src/main/java/bisq/cli/table/builder/TransactionTableBuilder.java
+++ b/cli/src/main/java/bisq/cli/table/builder/TransactionTableBuilder.java
@@ -20,8 +20,7 @@ package bisq.cli.table.builder;
 import bisq.proto.grpc.TxInfo;
 
 import java.util.List;
-
-import javax.annotation.Nullable;
+import java.util.stream.Collectors;
 
 import static bisq.cli.table.builder.TableBuilderConstants.*;
 import static bisq.cli.table.builder.TableType.TRANSACTION_TBL;
@@ -47,6 +46,7 @@ class TransactionTableBuilder extends AbstractTableBuilder {
     private final Column<Long> colOutputSum;
     private final Column<Long> colTxFee;
     private final Column<Long> colTxSize;
+    private final Column<String> colMemo;
 
     TransactionTableBuilder(List<?> protos) {
         super(TRANSACTION_TBL, protos);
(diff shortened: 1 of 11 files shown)


cli/src/main/java/bisq/cli/table/builder/TransactionTableBuilder.java
```diff
@@ -56,32 +56,22 @@ class TransactionTableBuilder extends AbstractTableBuilder {
...
+        this.colMemo = new StringColumn(COL_HEADER_TX_MEMO);
     }
 
     public Table build() {
-        // TODO Add 'gettransactions' api method & show multiple tx in the console.
-        //  For now, a tx tbl is only one row.
-        TxInfo tx = (TxInfo) protos.get(0);
-
-        // Declare the columns derived from tx info.
+        List<TxInfo> transactions = protos.stream().map(p -> (TxInfo) p).collect(Collectors.toList());
```
alvasw: You can avoid creating the `List` here by replacing this line with:
```
Stream<TxInfo> transactions = protos.stream()
                .map(p -> (TxInfo) p);
```

core/src/main/java/bisq/core/api/CoreWalletsService.java
```diff
@@ -413,6 +414,10 @@ TxFeeRateInfo getMostRecentTxFeeRateInfo() {
                 feeService.getLastRequest());
     }
 
+    List<Transaction> getTransactions() {
+        return btcWalletService.getTransactions(false).stream().collect(Collectors.toList());
```
alvasw: `btcWalletService.getTransactions(false)` returns a `Set`. You can pass that `Set` to `GetTransactionsReply.newBuilder().addAllTxInfo(...)` because it accepts any `Iterable<? extends TxInfo>`.

alvasw (changes requested):
Nit

cli/src/main/java/bisq/cli/table/builder/TransactionTableBuilder.java
helixx87: Great catch - fixed it.

core/src/main/java/bisq/core/api/CoreWalletsService.java
helixx87: Fixed

helixx87:
Thanks @alvasw for review and valuable suggestions regarding code quality. I introduced them here using `git commit --amend`.

alvasw (changes requested):
Code changes look good to me! :+1: 

When I run `bisq-cli --help`, I cannot find the `gettransactions` method in the list.
```
gettxfeerate                                                                 Get current tx fee rate in sats/byte

settxfeerate             --tx-fee-rate=<sats/byte>                           Set custom tx fee rate in sats/byte

unsettxfeerate                                                               Unset custom tx fee rate

gettransaction           --transaction-id=<transaction-id>                   Get transaction with id

createoffer              --payment-account=<payment-account-id> \            Create and place an offer
                         --direction=<buy|sell> \                            
                         --currency-code=<currency-code> \                   
                         --amount=<btc-amount> \                             
```

alvasw (approved):
ACK

alejandrogarcia83 (approved):
utACK

PR #6697, opened 2023-05-10
Speed up burningman and statistics view loads

stejbac:
This PR makes the trade statistics set held by `TradeStatisticsManager` into a sorted collection (a `TreeSet`), in order to make certain optimisations possible which speed up the Burning Man view, the Trades Charts view and the BSQ Dashboard view (plus a few other minor areas of the application). The PR also applies quite a lot of miscellaneous optimisations to those three views, plus a simple related optimisation to the BSQ Supply view.

The two main benefits of holding the trade statistics in a sorted set (with chronological ordering) are:

1) Lightweight caches can be used in timezone, time interval and similar calculations much more easily, since consecutive trade statistics objects tend to be close to each other in time;
2) It is possible to obtain a date range view of the trade statistics set very cheaply (in _O(log n)_ time), avoiding the need to stream over the entire collection in `AveragePriceUtil` and elsewhere. For this purpose, a `RangeUtils` class is provided, which works round the limitation that `java.util.NavigableSet` only allows searching with keys of the collection type, not mapped keys such as dates.

Note that even though a tree has somewhat slower lookups than a hash table, it doesn't matter here since there are no lookups into the trade statistics set, only streaming & filtering of the elements, plus occasional insertions. (Also I believe `TreeSet`s are a little more memory efficient than `HashSet`s.)

--

The biggest speedup is probably to the burning man view (Under _DAO_ / _Proof of Burn_ / _Burningmen_). Tabbing to it five times gives the following hotspots from JProfiler before the changes:
Screenshot-2023-5-10 Hot Spots
As can be seen, by far the biggest hotspot comes from `AveragePriceUtil.getAveragePriceTuple`, which filters the entire trade statistics set for BSQ & USD trades in a given one-month (as specified) period.

After the changes, tabbing to the view five times gives the following hotspots from JProfiler:
Screenshot-2023-5-10 Hot Spots(1)

### DIFF
--- a/common/src/main/java/bisq/common/util/ComparableExt.java
+++ b/common/src/main/java/bisq/common/util/ComparableExt.java
@@ -1,0 +1,65 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.common.util;
+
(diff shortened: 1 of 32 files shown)


HenrikJannsen:
Wow! Great thanks for the impressive improvements!

HenrikJannsen (approved):
utACK - Very impressive work!!!

HenrikJannsen:
I tested the branch and the BM view and BSQ Facts&Figures views are much faster now!

alejandrogarcia83 (approved):
utACK

Thank you!! This is a massive improvement.

PR #6700, opened 2023-05-17
Fix BM receive address selection

### LINKED ISSUE #6699: Bug with selecting the distributed BM receiver address
There is a bug in the way how we select the most recent address for the distributed burningman.

We take the latest address and override any custom defined address (in comp. requests) by the change address of a new CR (which does not have a custom address defined).

For deploying the bugfix we need to maintain the old buggy version until all users have updated.
We can do that by using an activation date when the new `BurningManCandidate.mostRecentAddress` is used and enforce the new version with the Filter.

BM who prefer to get BTC funds on their custom address can do that by adding the custom address at every cycles CR until the bugfix is activated (will take about 2-3 months).

HenrikJannsen:
Fixes https://github.com/bisq-network/bisq/issues/6699


alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6702, opened 2023-05-17
Remove date check

HenrikJannsen:
Remove USE_FEE_FROM_FILTER_ACTIVATION_DATE and new Date().before(USE_FEE_FROM_FILTER_ACTIVATION_DATE) as it would always be false


alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6705, opened 2023-05-17
Remove warning popup when UI thread is blocked for > 20 sec.

### LINKED ISSUE #6703: Remove "Your computer was asleep" popup
At startup when not being up to data with blockchain and DAO data we often get into the "Your computer was asleep for {0} seconds." popup. The Don't show again does not work there, so even you click it you get repeated the popups (probably they got queued up already before the checkbox has effect).

The intention of that popup was to detect sleep mode. We do not really know if that is still an issue, but we get that misleading and annoying popups way too often. So I suggest that we remove that and do a warn logging instead.

The reason why the UI thread gets blocked at blockchain/DAO sync need to be be investigated.
See https://github.com/bisq-network/bisq/issues/6704

HenrikJannsen:
Fixes https://github.com/bisq-network/bisq/issues/6703


alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6709, opened 2023-05-20
Add available BSQ amount to top menu bar

### LINKED ISSUE #6660: Add available BSQ amount to top menu bar in client
### Description

Bisq has two native wallets, BTC and BSQ.

Users of Bisq see there BTC balances on the top right of the Bisq client (available, reserved and locked), but to see their BSQ balance they need to go to  > DAO > Bisq wallet.

It would improve usability if users could see their available BSQ balance alongside their BTC balances in the top right menu bar on Bisq.

I would propose adding it on the far right. 

#### Version

1.9.9

### Steps to reproduce

To see your BSQ balance you have to go through a couple of screens.

### Expected behaviour

Be able to see your BSQ balance on every screen in the top right alongside BTC balances.

helixx87:
Fixes #6660 

bsq-balance

topbar-2

topbar-2b

topbar-3

### DIFF
--- a/core/src/main/java/bisq/core/btc/Balances.java
+++ b/core/src/main/java/bisq/core/btc/Balances.java
@@ -18,7 +18,9 @@
 package bisq.core.btc;
 
 import bisq.core.btc.listeners.BalanceListener;
+import bisq.core.btc.listeners.BsqBalanceListener;
 import bisq.core.btc.model.AddressEntry;
+import bisq.core.btc.wallet.BsqWalletService;
 import bisq.core.btc.wallet.BtcWalletService;
 import bisq.core.offer.OpenOffer;
 import bisq.core.offer.OpenOfferManager;
@@ -51,6 +53,7 @@ import lombok.extern.slf4j.Slf4j;
 public class Balances {
     private final TradeManager tradeManager;
     private final BtcWalletService btcWalletService;
+    private final BsqWalletService bsqWalletService;
     private final OpenOfferManager openOfferManager;
     private final ClosedTradableManager closedTradableManager;
     private final FailedTradesManager failedTradesManager;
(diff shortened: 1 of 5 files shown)


pazza83:
@helixx87 

Thanks for this.

How does this display if a user has say 5 figures in BSQ?

helixx87:
@pazza83 I've added a few more screenshots that present this feature in a different scenarios

pazza83:
@helixx87 

It looks great. Many thanks for doing this. 

Looking forward to seeing this added to Bisq.

HenrikJannsen:
Is the layout still ok with the max. number of decimals?
An alternative could be to let BTC and BSQ balance flip on mouse click and if not clicked to let them cross fade with a animation. so the user can see the other balace every 3 sec or so. we can do that only if there is a non zero BSQ balance.

pazza83:
> we can do that only if there is a non zero BSQ balance.

I think having the amount fixed top right works well.

Makes casual users aware that BSQ exists. That was part of the reason behind the creation of the issues. Increase BSQ awareness.

alvasw:
I understand the motivation to show the BSQ balance in the top menu bar, but I think it clutters the UI. IMO it's better to show the makers (at offer creation) and takers (when taking an offer) how much they would save by using BSQ. Nevertheless, I'm not against this change.

pazza83:
I created the issue following new user feedback saying they would like to see their BSQ balance at a glance. 

Another user had sent them BSQ to funds there trade fees for their first few trades but they could not find what their balance was.

I do think it makes sense to show the BSQ balance in the top right rather than a few screens deep as it currently is.

I think there is enough space in the top bar (when using in full screen mode) for the addition of the BSQ balance not to add clutter.

I think this change is good as part of a number of changes needed to make BSQ more prominent in Bisq.

HenrikJannsen:
The BSQ field is truncated even with small amounts:
Screenshot 2023-05-24 at 07 54 00

And more truncated with large amounts:
Screenshot 2023-05-24 at 07 51 51


If the app has min. screen width the balance fields are truncated:
Screenshot 2023-05-24 at 07 50 53

Without the new field we also have truncated the balance fields, which is also not great.
Screenshot 2023-05-24 at 07 58 50

One option to solve that would be to pack the available/reserved and locked balance into a combobox, similar as the price box and show available by default. We could do that only if the screen width is too small, otherwise we can expand the fields as it is now.

pazza83:
Would be good to get this added still @helixx87 where are you up to with this?

alejandrogarcia83:
I'm closing this pull request for now. Feel free to reopen it when ready!

PR #6712, opened 2023-05-24
Exclude mostRecentAddress from EqualsAndHashCode

HenrikJannsen:
The getter was called by EqualsAndHashCode which throws an exception as it is is not intended to get used anymore.


alejandrogarcia83 (approved):
utACK

PR #6716, opened 2023-05-31
Fix broken gson version enforcement

alvasw:
We strictly enforce version 2.8.5 for all modules, but the cli module transitively depends on version 2.8.6.


alejandrogarcia83 (approved):
utACK

PR #6723, opened 2023-06-05
Allow dispute agent to resend ticket.

### LINKED ISSUE #6690: Add option for mediator to resend PeerOpenedDisputeMessage to non-resposive peer.
### Description

Described [here](https://github.com/bisq-network/bisq/pull/6376#issuecomment-1532963851) by @pazza83, user may never receive mediation ticket if issued in the middle of them performing an SPV resync.  This is because their Bisq instance is up / receptive but the infrastructure components are not active / listening, so any protocol messages received during an SPV resync are dropped.

- Alice and Bob have a trade in process.
- Bob restarts Bisq for an SPV resync (which is a lengthy process), then Alice opens a ticket with the mediator.
- The mediator relays the ticket to Bob by sending `PeerOpenedDisputeMessage` **while he is in the middle of syncing the chain**.
- Bob never receives the mediation ticket message, (message is received by the low level Connection class but there is no listening infrastructure to process it).

Bob is unaware that mediation was started, and mediator is unable to communicate with Bob.
We need Bob to get a mediation ticket in order to resolve his trade.

[From @pazza83:](https://github.com/bisq-network/bisq/pull/6376#issuecomment-1536782210)

> Mediator will always try to send a message again if a trader does not reply.
> Could something be done on the subsequent attempts? Maybe checking to see if a mediation ticket exists on the traders instance and opening one if that is not the case.

### Proposed solution

Allow the Mediator's instance to:

1. Detect that there is a problem communicating with Bob on the ticket (no dispute chat msgs have been ACKd).
2. Prompt to re-send the dispute ticket `PeerOpenedDisputeMessage` to Bob.

### Considerations

This proposed change, being a special case (as with most bug fixes) would slightly increase complexity in the DisputeManager code.  I would like concept approval from a reviewer before proceeding with the solution.

ghost:
Fixes #6690

Dispute Agent will get this prompt when it detects that a trader has not acknowledged any dispute chats.

Applies both in mediation and arbitration, as the ticket opening process is the same.

Screenshot from 2023-06-05 15-00-43

### DIFF
--- a/core/src/main/java/bisq/core/support/dispute/DisputeManager.java
+++ b/core/src/main/java/bisq/core/support/dispute/DisputeManager.java
@@ -363,6 +363,24 @@ public abstract class DisputeManager<T extends DisputeList<Dispute>> extends Sup
     // Message handler
     ///////////////////////////////////////////////////////////////////////////////////////////
 
+    public boolean agentCheckDisputeHealth(Dispute disputeToCheck) {
+        // checking from the agent perspective only
+        if (disputeToCheck.getChatMessages().stream().anyMatch(ChatMessage::isSenderIsTrader)) {
+            return true;
+        }
+        // consider only messages which have been transmitted
+        List<ChatMessage> transmittedMessages = disputeToCheck.getChatMessages().stream()
+                .filter(e -> !e.isSystemMessage())
+                .filter(e -> !e.getStoredInMailboxProperty().get())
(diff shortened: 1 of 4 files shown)


alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6724, opened 2023-06-06
Fix bisq.properties pickup in ApiTest

alyokaz:
This change copies the 'bisq.properties' file in apitest/src/main/resources to each instance's DataDir on test harness startup.

This was necessary as #5718 was not copying the file into the instance's data dir but leaving it in 'build/resources/main' where it wasn't picked up by the instance.

### DIFF
--- a/apitest/src/main/java/bisq/apitest/Scaffold.java
+++ b/apitest/src/main/java/bisq/apitest/Scaffold.java
@@ -29,6 +29,7 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 
+import java.util.Arrays;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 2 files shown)


apitest/src/main/java/bisq/apitest/Scaffold.java
```diff
@@ -242,28 +243,52 @@ public void installDaoSetupDirectories() {
...
             // by a Bisq instance:  delete the 'bisq.properties' file automatically copied
             // to the 'apitest/build/resources/main' directory during IDE or Gradle build.
             // Note:  there is no way to prevent this deleted file from being re-copied
             // from 'src/main/resources' to the buildDataDir each time you hit the build
             // button in the IDE.
-            BashCommand rmRedundantBisqPropertiesFile =
-                    new BashCommand("rm -rf " + buildDataDir + "/bisq.properties");
-            if (rmRedundantBisqPropertiesFile.run().getExitStatus() != 0)
-                throw new IllegalStateException("Could not delete redundant bisq.properties file");
+            deleteRedundantBisqPropertiesFile();
```
alvasw: I know, you didn't write this code, but do we need to delete the `bisq.properties` file? This breaks the Gradle Build Cache.

alvasw (changes requested):
Do we need to delete the `bisq.properties` file in the root directory? The deletion breaks the Gradle Build Cache. It's smarter to include/exclude the file explicitly [1], or even have them multiple times (for all instances) in the resources directory.

I know that you didn't write the initial file-deletion code.

[1] https://docs.gradle.org/current/dsl/org.gradle.language.jvm.tasks.ProcessResources.html

apitest/src/main/java/bisq/apitest/Scaffold.java
AllyKaz: I think it seems fine to leave. It's presence can act as a sign the files were copied from a single 'master' file and it's absence as a sign of explicit inclusion. 

I've made the build file copy the bisq.properties file into the relevant directories if it's present and have removed the equivalent logic from Scaffold along with that which deletes the file.

I've also refactored the dao-setup.gradle file to remove some redundant copy and renaming.

alyokaz: I think it's fine to keep, its presence signals it's being used by all instances, and its absence signals explicit inclusion.

I've removed `deleteRedundantBisqPropertiesFile()` and done some refactoring on `dao-setup.gradle`, removing some redundant copying and renaming.

apitest/src/main/java/bisq/apitest/Scaffold.java
```diff
@@ -29,6 +29,7 @@
 import java.io.FileNotFoundException;
 import java.io.IOException;
 
+import java.util.Arrays;
```
alvasw: This is import is not needed anymore.

apitest/src/main/java/bisq/apitest/Scaffold.java
```diff
@@ -242,28 +243,36 @@ public void installDaoSetupDirectories() {
...
-
         } catch (IOException | InterruptedException ex) {
             throw new IllegalStateException("Could not install dao-setup files from " + daoSetupDir, ex);
         }
     }
 
+    private void copyBisqPropertiesFileToAppDataDir() throws IOException, InterruptedException {
+        copyBisqPropertiesFileToAppDataDir(alicedaemon);
+        copyBisqPropertiesFileToAppDataDir(bobdaemon);
+    }
```
alvasw, line 262: Please reformat this to:
```diff
     }
+
     private void copyBisqPropertiesFileToAppDataDir(BisqAppConfig bisqAppConfig) throws IOException, InterruptedException {
-        if(!new File(config.baseSrcResourcesDir + "/bisq.properties").exists()) return;
+        if (!new File(config.baseSrcResourcesDir + "/bisq.properties").exists()) return;
 
         String instanceDataDir = config.rootAppDataDir + "/" + bisqAppConfig.appName;
```

alvasw (changes requested):
Looks good!

apitest/src/main/java/bisq/apitest/Scaffold.java
alyokaz: Amended as instructed.

alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6725, opened 2023-06-07
Remove hardcoding of enableBisqDebugging in ApiTest

alyokaz:
Hardcode enableBisqDebugging in the Api tests prevents the option to set it from the apitest.properties file as any hardcoded value will be set as a command line option which will always take precedence over any value placed in the apitest.properties file. This makes it impossible to turn on debugging without altering the code in various places. It is also redundant as hardcoded value is false, which is the default anyway. 

This change removes the hardcoding of the value, allowing the test hardness to fallback on the default of debugging being turned off, with the option to turn it on from the apitest.properties file.


alvasw (approved):
utACK

Please sign your commits. See: https://github.com/bisq-network/bisq/pull/6732#issuecomment-1615186608

alyokaz:
Rebased and signed this and all other commits.

alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6762, opened 2023-07-10
Change bond roles names for Bisq 2 roles

HenrikJannsen:
Repurpose not used bonded roles for Bisq 2 roles.

We prefer to not add new roles to avoid risks with DAO consensus issues.


alejandrogarcia83 (approved):
utACK

PR #6783, opened 2023-07-24
Add new md files

HenrikJannsen:
Extract 3 sections from the release notes as markdown pages.
The release notes page is quite a lot of text and linking to those md files seems a better approach.

Update slightly the text where appropriate.


alejandrogarcia83 (approved):
utACK

PR #6788, opened 2023-07-31
Feature: customizable trade greetings

ghost:
I implemented this feature claiming a bounty from @anongit111.  They have requested the feature be added to the official Bisq build.  See discussion https://github.com/bisq-network/bisq/discussions/6764

---

- Greeting messages can be defined from the Settings screen `EDIT GREETINGS`.
- If applicable, greeting will be sent in a trader chat message when an offer is taken.

---

image

image

### DIFF
--- a/core/src/main/java/bisq/core/support/traderchat/TraderChatManager.java
+++ b/core/src/main/java/bisq/core/support/traderchat/TraderChatManager.java
@@ -125,7 +125,7 @@ public class TraderChatManager extends SupportManager {
             List<ChatMessage> chatMessages = trade.getChatMessages();
             if (chatMessages.stream().noneMatch(m -> m.getUid().equals(message.getUid()))) {
                 if (chatMessages.isEmpty()) {
-                    addSystemMsg(trade);
+                    tradeManager.addSystemMsg(trade);
                 }
                 trade.addAndPersistChatMessage(message);
                 tradeManager.requestPersistence();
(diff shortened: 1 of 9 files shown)


anongit111:
Made four trades with the patched version and it works as intended . As maker this is quite useful imo .
ACK from my side .

alejandrogarcia83:
This is a valid point https://github.com/bisq-network/bisq/discussions/6764#discussioncomment-6491327 from @pazza83 :
> I have some concerns about adding a defined opening chat message for traders.
The reason is I think the payment account set up should give the buyer's everything they need to make the payment.
There are cases where traders can 'agree' to use a specific reference or for the payment to be made a certain way. But having a message sent automatically by a trader might give a request the appearance of authority to a new buyer.
For example messages like:
"Hi thanks for taking this trade please make payment with a reference related to the purchase of an electronic good."
"Hi thanks for taking this trade. When you pay please add your name as the payment reference"
Whilst it is fine to request these things they should be in agreement with the buyer and they should be able to refuse and send the payment with no reference if they prefer.
The buyer could always message manually but the nature of Bisq at least means that the makers of the offer are not always present when the trade is taken, meaning they should expect to receive payment as per the usual trading protocols.
> > I ask because it seems that many banks auto create a reason for payment if left blank. I as maker don't want strange reason for payment.
> > > If someone was to send a payment with a strange reference this is already against trade rules and could be punished.

@alvasw @HenrikJannsen WDYT?

HenrikJannsen:
Concept NACK. I will comment at https://github.com/bisq-network/bisq/discussions/6764 why.

PR #6807, opened 2023-08-11
Forth phase at seller (part 6)

HenrikJannsen:
Based on https://github.com/bisq-network/bisq/pull/6806


Godson1986:
> 
# first-level heading
## second-level heading

PR #6810, opened 2023-08-15
Allow Tx validation requests when the DAO state is ready.

ghost:
A bug was raised whereby sometimes at startup, offers were being disabled.

It happens if the offer validation is sooner than the DAO being ready, in which case the valid BM fee receiver addresses are not known. 

---

TriggerPriceService [validates open offers at startup](https://github.com/bisq-network/bisq/blob/cc183ca77dc413c2739a537ff01976af09135354/core/src/main/java/bisq/core/offer/bisq_v1/TriggerPriceService.java#L169), by testing `mempoolService.canRequestBeMade()` and if so, `mempoolService.validateOfferMakerTx()`.  Now, `canRequestBeMade()`  will check the DAO state, and if not ready the checking will be postponed.

### DIFF
--- a/core/src/main/java/bisq/core/provider/mempool/MempoolService.java
+++ b/core/src/main/java/bisq/core/provider/mempool/MempoolService.java
@@ -88,7 +88,7 @@ public class MempoolService {
     }
 
     public boolean canRequestBeMade() {
-        return outstandingRequests < 5; // limit max simultaneous lookups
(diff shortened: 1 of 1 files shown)


alvasw (approved):
ACK

Good catch!! :-)

alejandrogarcia83 (approved):
utACK

PR #6814, opened 2023-08-16
Fix: BTC maker/taker fee validation

### LINKED ISSUE #6813: Error message: Validation of Taker Tx returned: [fee address: bc1qj62pl9d06c2f4dzutnqsn7nv0gvy7k7gwhlv3c was not a known BTC fee receiver, Taker tx validation]
Open Support ticket
There is an issue with a missing or invalid transaction.

Please do NOT send the fiat or altcoin payment.

Open a support ticket to get assistance from a Mediator.

Error message: Validation of Taker Tx returned: [fee address: bc1qj62pl9d06c2f4dzutnqsn7nv0gvy7k7gwhlv3c was not a known BTC fee receiver, Taker tx validation]

Sorry if I'm at the wrong place to ask this.

#### Additional info

[bisq 1.log](https://github.com/bisq-network/bisq/files/12355278/bisq.1.log)

ghost:
Fixes #6813

- User took an offer at startup, before DAO had finished parsing blockchain.
- BM addresses were not yet known, so fee validation erroneously flagged the taker fee as unrecognized.
 
It is another edge case similar to #6810, so protect from false flagging a trade if the DAO is not yet ready.

### DIFF
--- a/core/src/main/java/bisq/core/provider/mempool/MempoolService.java
+++ b/core/src/main/java/bisq/core/provider/mempool/MempoolService.java
@@ -300,6 +300,10 @@ public class MempoolService {
             log.info("MempoolService only supports mainnet");
             return false;
         }
+        if (!canRequestBeMade()) {
+            log.info("Tx Validation bypassed as service is not ready");
+            return false;
+        }
         return true;
     }
 }


ghost:
To clarify: Bisq validates fees paid in Trade Step 2, prior to payment.  In this case the app was started up with an existing trade and the validation was allowed to occur too early, before the DAO was ready, causing the erroneous warning.

alvasw (changes requested):
Nit

In this PR, you check with `canRequestBeMade()` in `isServiceSupported` whether the request can be made or not. This should cover the change from your previous PR (#6810) too. Therefore, we can safely remove `mempoolService.canRequestBeMade(offerPayload)` in the `TriggerPriceService`. Is this correct or am I missing something?

TriggerPriceService:
```java
            OfferPayload offerPayload = offer.getOfferPayload().orElseThrow();
            if (openOffer.getMempoolStatus() < 0 &&
                    mempoolService.canRequestBeMade(offerPayload)) {
                mempoolService.validateOfferMakerTx(offerPayload, (txValidator -> {
                    openOffer.setMempoolStatus(txValidator.isFail() ? 0 : 1);
                }));
            }
```

Here, `mempoolService.validateOfferMakerTx(...)`, calls `validateOfferMakerTx(...)`, and `validateOfferMakerTx()` calls `isServiceSupported()` before creating the mempool request.

MempoolService:
```java
    public void validateOfferMakerTx(TxValidator txValidator, Consumer<TxValidator> resultHandler) {
        if (txValidator.getIsFeeCurrencyBtc() != null && txValidator.getIsFeeCurrencyBtc()) {
            if (!isServiceSupported()) {
                UserThread.runAfter(() -> resultHandler.accept(txValidator.endResult("mempool request not supported, bypassing", true)), 1);
                return;
            }
            MempoolRequest mempoolRequest = new MempoolRequest(preferences, socks5ProxyProvider);
            validateOfferMakerTx(mempoolRequest, txValidator, resultHandler);
        } else {
            // using BSQ for fees
            UserThread.runAfter(() -> resultHandler.accept(txValidator.validateBsqFeeTx(true)), 1);
        }
    }
```

ghost:
`canRequestBeMade()` ensures that the DAO is ready and we are not flooding too many simultaneous requests.

`canRequestBeMade(offer specific)` in addition ensures that a new offer is not checked immediately, as the info retrieved from mempool.space may not be available until the Tx has time to propagate.


If TriggerPriceService was simplified to only call `validateOfferMakerTx` you have the possibility that the offer gets flagged as valid just because the DAO was not yet ready.  Or the possibility that a valid offer gets disabled (as missing Tx) before its Tx had chance to propagate.

So I think the validation has to be done after the system becomes ready (i.e. the way it is now).

alvasw:
@jmacxx Thank you very much for the explanation!

alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6818, opened 2023-08-18
Arbitration broadcast DPT and dispute message in parallel.

### LINKED ISSUE #6797: Allow users to Open Arbitration ticket if first attempt fails
I think there is currently a bug in Bisq where once a user broadcasts the DPT, and if for whatever reason, an arbitration ticket is not opened on their instance, they are then unable to press the arbitration button again as they get an error message due to the DPT being successful even though the opening of an arbitration ticket has not been.

This is annoying for the user as they will have to wait longer for the payout. It is also causes more work for the mediator and arbitration as a less secure off Bisq reimbursement is required.

An example of a recent error message a user received is:

```
Error
An error occurred at task: PublishedDelayedPayoutTx
TxBroadcastException{
     txId='d7b32b1f6602b0830492d6caebffc3b8ad4c417f7de91bb8f327284553b0eaf6'
} bisq.core.btc.exceptions.TxBroadcastException: We got broadcastTx called with a tx which has an open timeoutTimer. txId=d7b32b1f6602b0830492d6caebffc3b8ad4c417f7de91bb8f327284553b0eaf6

```

I have the logs from the user to provide if needed.

The ideal solution would be for the arbitration button to remain clickable based not on the DPT transaction alone but on if an arbitration ticket if is open on the user's instance. Or alternatively just not give an error message and just attempt to send the DPT AND open an arbitration ticket.

ghost:
Fixes #6797

> attempt to send the DPT AND open an arbitration ticket

Previously the dispute message would only be sent once the DPT had finished propagation (which can take a few seconds or more on mainnet).  To a user it would appear as if their clicking the button had no effect and they would attempt to click again. That would cause the DPT broadcast task to fail (error: `broadcastTx called with a tx which has an open timeoutTimer`) and it would result in no happy path therefore no Dispute message / ticket being opened.

Testing confirmed there is no requirement for the dispute to be chained on the successful propagation of the DPT to nodes. The send to arbitration can be attempted again if necessary, by pressing CTRL-O / Open support ticket.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/PendingTradesDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/PendingTradesDataModel.java
@@ -647,11 +647,10 @@ public class PendingTradesDataModel extends ActivatableDataModel {
             dispute.setBurningManSelectionHeight(trade.getProcessModel().getBurningManSelectionHeight());
             dispute.setTradeTxFee(trade.getTradeTxFeeAsLong());
 
-            ((DisputeProtocol) tradeManager.getTradeProtocol(trade)).onPublishDelayedPayoutTx(() -> {
-                        log.info("DelayedPayoutTx published and message sent to peer");
-                        sendOpenDisputeMessage(disputeManager, resultHandler, dispute);
-                    },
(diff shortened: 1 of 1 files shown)


alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6822, opened 2023-08-22
Remove unused dependency on Pricenode metadata strings.

### LINKED ISSUE #6820: Bisq loses market price feed when hardcoded exchange names are not present
There is code in the Bisq client to expect the following hard coded metadata items from the Price Node:

https://github.com/bisq-network/bisq/blob/34d61d1f56486e8971fb713177bd7ddaa12e0cd7/core/src/main/java/bisq/core/provider/price/PriceProvider.java#L67-L69

A [comment](https://github.com/bisq-network/bisq-pricenode/blob/ce8e04d911cd921bb7cccc2fac9e629bca4cc0ea/src/main/java/bisq/price/spot/ExchangeRateService.java#L64-L67) in the PriceNode says "_Specific metadata fields for specific providers are expected by the client, mostly for historical reasons_".

If any of them are not present, an Null Pointer Exception occurs and the price node is switched out for an alternate.  If all price nodes are doing the same thing, then Bisq will not have any market price feed.

This is currently a problem because the poloniex feed is down, and therefore `poloniexTs` is not present in metadata, unless the pricenode has cached results from previously when the feed was working.  The observed effect is that if you reboot a pricenode with the poloniex feed not working, it will not provide metadata that Bisq clients rely on.

---

Temporarily, here is a [bisq-pricenode](https://github.com/bisq-network/bisq-pricenode/) patch to restore the missing metadata:
https://gist.github.com/jmacxx/eff7705656640d7845e15bad60ad11d3

Ultimately, the Bisq client should not expect exchange specific tags in the feed, because exchanges are not permanent.

--- 

Credit to Matrix user **paulj** for working through this issue and https://github.com/bisq-network/bisq-pricenode/issues/18

ghost:
Fixes #6820 

Bisq client rejects the pricenode if certain metadata exchange timestamps are not present in the feed.
However they are not used since https://github.com/bisq-network/bisq/commit/189431ed7ea72930f000bbee4b24aa8cbf440e35 so remove the hardcoded dependency.

### DIFF
--- a/core/src/main/java/bisq/core/provider/price/PriceFeedService.java
+++ b/core/src/main/java/bisq/core/provider/price/PriceFeedService.java
@@ -394,21 +394,17 @@ public class PriceFeedService {
         }
 
         priceRequest = new PriceRequest();
-        SettableFuture<Tuple2<Map<String, Long>, Map<String, MarketPrice>>> future = priceRequest.requestAllPrices(provider);
+        SettableFuture<Map<String, MarketPrice>> future = priceRequest.requestAllPrices(provider);
         Futures.addCallback(future, new FutureCallback<>() {
             @Override
-            public void onSuccess(@Nullable Tuple2<Map<String, Long>, Map<String, MarketPrice>> result) {
+            public void onSuccess(@Nullable Map<String, MarketPrice> result) {
                 UserThread.execute(() -> {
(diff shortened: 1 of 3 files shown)


alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6840, opened 2023-08-30
Add info popup about Argentinian Peso market prices.

ghost:
@HenrikJannsen [requested](https://github.com/bisq-network/bisq-pricenode/pull/15#issuecomment-1666597874) we display some info about the new ARS blue rate to users.

----

This will show at application startup and can be dismissed with the "Don't show again" option.

bluedollar

----

FYI 50% of [pricenodes](https://github.com/bisq-network/roles/issues/14) are currently providing blue market rates for ARS, the remaining 50% have not yet updated.

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -3411,6 +3411,13 @@ popup.info.torMigration.msg=Your Bisq node is currently using a Tor v2 address. 
   [HYPERLINK:https://bisq.wiki/Changing_your_onion_address] or in this video [HYPERLINK:https://bitcointv.com/videos/watch/f96adc20-4092-4253-84c0-1b18088b4b95]. \n\n\
   Make sure to back up your data directory beforehand.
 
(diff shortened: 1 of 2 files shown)


alvasw:
This won't affect the majority of our users. We can show the dialog at start-up to users with an ARS payment account and additionally to everyone when they navigate to the ARS currency. What do you think?

ghost:
I would think showing the news popup at startup helps attract users to the ARS market!  (Personally I don't think its an inconvenience, especially since there's a `Don't show again` option).

> show the dialog at start-up to users with an ARS payment account

Likely very few users would see it; bear in mind there have only been 10 lifetime trades.  Also weigh the challenge of testing all possible payment methods for ARS.

Happy enough to implement those changes if that's what the community wants.  @rodvar @pazza83 your thoughts?

alvasw:
I agree with you!

alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

rodvar:
Awesome guys!!!

PR #6848, opened 2023-09-04
Improve flow in create/take offer process.

ghost:
#### Summary: Cuts make/take offer process from 4 screens to 2 by opting to fund from Bisq wallet.

Requested by @HenrikJannsen #6824.

When creating or taking an offer the user gets the choice to fund the trade from an external wallet for potentially better privacy (avoid coin merge if inputs from ext. wallet are kept separated). From feedback it seems that most users are not using that option and then they need to do that extra click on the "Fund from internal Bisq wallet" button.

We offer a way to avoid that extra step if the user chooses to use the internal wallet. We show at the first time a popup which give them an option to prefer funding from Bisq wallet in future, with background info why funding from an external wallet could improve privacy.  In the settings we add a toggle to enable/disable the feature.  The funding and review screens are skipped if user chooses this feature, leading to `data entry > confirm` (2 clicks) rather than `data entry > fund > review > confirm` (4 clicks).

(Note that if the Bisq wallet lacks sufficient funds, it will revert to prompting the user to fund the trade as before).

----

- If the user clicks on "Transfer Funds From Bisq Wallet"..

image

----

- ..it asks if they would like to choose this by default in the future:

image

----
- Setting to control the feature:
 
image

### DIFF
--- a/core/src/main/java/bisq/core/user/PreferencesPayload.java
+++ b/core/src/main/java/bisq/core/user/PreferencesPayload.java
@@ -151,6 +151,7 @@ public final class PreferencesPayload implements PersistableEnvelope {
     // Added at 1.9.11
     private boolean isFullBMAccountingNode = false;
 
+    private boolean useBisqWalletFunding = false;
 
     ///////////////////////////////////////////////////////////////////////////////////////////
     // Constructor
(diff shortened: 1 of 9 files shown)


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6868, opened 2023-09-11
Fix codacy coverage reporter Java version

alejandrogarcia83:
@alvasw's PRs enforce Java 11 but the codacy coverage reporter yml file uses Java 15.


gabernard (approved):
utACK

PR #6879, opened 2023-09-24
Fix seednode startup script.

ghost:
Per @alvasw, use `bash` instead of `sh`.
Allows seednode to start correctly in v1.9.13.


alvasw (approved):
ACK

alejandrogarcia83 (approved):
utACK

PR #6899, opened 2023-10-08
Add filter option for delayed payout

HenrikJannsen:


desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/PendingTradesDataModel.java
```diff
@@ -695,5 +702,33 @@ private void sendOpenDisputeMessage(DisputeManager<? extends DisputeList<Dispute
...
+            tradeManager.requestPersistence();
+        }
+    }
+
+    public long getSellerConfirmedPaymentReceiptDate() {
+        return Objects.requireNonNull(getTrade()).getSellerConfirmedPaymentReceiptDate();
+    }
+
+    public Date getDelayedPayoutDate() {
+        return new Date(Objects.requireNonNull(getTrade()).getSellerConfirmedPaymentReceiptDate() + getPayoutDelay());
```
sqrrm: Better to use local convenience function `getSellerConfirmedPaymentReceiptDate()`

desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/steps/seller/SellerStep3View.java
```diff
@@ -475,14 +487,60 @@ else if (paymentAccountPayload instanceof AmazonGiftCardAccountPayload)
...
+    private void updateConfirmButton() {
+        if (model.dataModel.requiresPayoutDelay() &&
+                model.dataModel.getSellerConfirmedPaymentReceiptDate() > 0) {
+            if (payoutDelayTimer == null) {
+                confirmButton.setText(Res.get("portfolio.pending.step3_seller.releaseBitcoin"));
+                confirmButton.setOnAction(e -> onReleaseBitcoin());
+                payoutDelayTimer = UserThread.runPeriodically(this::updateConfirmButtonDisableState, 1, TimeUnit.SECONDS);
+            }
+            if (!model.dataModel.requiredPayoutDelayHasPassed()) {
+                new Popup().warning(Res.get("portfolio.pending.step3_seller.delayedPayout",
```
sqrrm: This pops up every time the view is activated. I think it should only pop up once after confirming payment receipt.

sqrrm (changes requested):
Everything works. Removing the filter immediately allows traders to release bitcoins. The recurring popup should be fixed before releasing, otherwise all looks good.

desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/PendingTradesDataModel.java
HenrikJannsen: ok

desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/steps/seller/SellerStep3View.java
HenrikJannsen: I preferred to do it that way as we do not show additional info in the UI. Users who quickly close without reading (they exist, I am such one ;-)) might get lost otherwise. Alternatively we could add a "dont show again" checkbox per trade.

sqrrm: Fair point on clicking away popups, don't show again, per trade, sounds like a reasonable compromise.

HenrikJannsen: I added the dont show again checkbox

sqrrm (dismissed):
utACK

sqrrm (dismissed):

ghost (changes requested):
Confirm button here now looks odd, because we have uppercase text for all other buttons in the app.   Previously it was uppercase.

image

image

HenrikJannsen:
> Confirm button here now looks odd, because we have uppercase text for all other buttons in the app. Previously it was uppercase.

Hmm... did not change that, but added uppercase to both now.

alejandrogarcia83 (approved):
utACK

ghost (commented):
ACK

PR #6931, opened 2023-10-18
Refactor & simplify fee service.

### LINKED ISSUE #5509: When chosen fee provider is down, Bisq should failover to a new one

ghost:
- Use fees retrieved from getAllMarketPrices.  Results in fee and price updated at the same rate, each minute.
- Saves resources: one socket connection and one service threadpool are eliminated.
- Failover to new node works ([previously did not](https://github.com/bisq-network/bisq/issues/5509)). 
- Uses POJO data transfer object instead of parsing Json in a tree. 
- Code footprint is reduced.
- ~For now the FeeService interface remains unchanged to make this PR easy to read; a further PR could~ clean up [client calls](https://github.com/bisq-network/bisq/blob/175a0946e55de1de105030791efc459cd7a7b00f/core/src/main/java/bisq/core/offer/bisq_v1/TakeOfferModel.java#L169-L171) as an async/blocking interface is not needed.  (-> https://github.com/bisq-network/bisq/pull/6931/commits/6b42bb9657ac3018b020bf228c9cec7e122eb5ca)

Fixes #5509


core/src/main/java/bisq/core/api/CoreApi.java
```diff
@@ -413,17 +413,12 @@ public boolean verifyBsqSentToAddress(String address, String amount) {
         return walletsService.verifyBsqSentToAddress(address, amount);
     }
 
-    public void getTxFeeRate(ResultHandler resultHandler) {
-        walletsService.getTxFeeRate(resultHandler);
+    public void setTxFeeRatePreference(long txFeeRate   ) {
```
alvasw: You accidentally inserted some whitespace `(long txFeeRate   ) {`.

alvasw (changes requested):
Great!! :-)

Nit

alejandrogarcia83 (approved):
utACK

PR #6946, opened 2023-11-14
Download all P2P data before announcing P2P ready.

### LINKED ISSUE #6823: Fresh client always prompted DAO out of sync after initial bootstrap.
#### Version

1.9.12

### Steps to reproduce

- Start Bisq as a new user with no data directory.
- Wait for it to finish sync.
- It pops up a message that the DAO is out of sync, prompting you to go to DAO Network Monitor.

image

image

### Additional info

The same thing happens to a seednode.  Its DAO state does not match consensus.
Workaround in that case is to shut down, delete `DaoStateStore` and `BsqBlocks` then restart.

ghost:
BsqBlocks were being downloaded and processed as soon as the initial P2P data response was received, however there can be subsequent P2P data batches received which contain information needed by the DAO.  Resulting in DAO out of sync error for new users.

Solution is to announce P2P ready once all the data has been received.

Closes #6823 

---

### Another problem identified & not yet fixed:

If the BTC blockchain is still being sync'd (typical for a new user) the download of BsqBlocks stalls after the first batch.  Resulting in "Synchronizing DAO" not completing until the user restarts.

### DIFF
--- a/p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataManager.java
+++ b/p2p/src/main/java/bisq/network/p2p/peers/getdata/RequestDataManager.java
@@ -359,17 +359,17 @@ public class RequestDataManager implements MessageListener, ConnectionListener, 
                                     checkNotNull(listener).onUpdatedDataReceived();
                                 }
 
-                                checkNotNull(listener).onDataReceived();
-
                                 if (wasTruncated) {
                                     if (numRepeatedRequests < 10) {
                                         log.info("DataResponse did not contain all data, so we repeat request until we got all data");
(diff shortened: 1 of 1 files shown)


alejandrogarcia83:
@alvasw @HenrikJannsen Could any of you ACK this PR? Thank you!

alvasw:
@HenrikJannsen should review this becaue he is more familiar with the P2P/DAO layer.

HenrikJannsen:
The code change looks good to me but when testing with a fresh app i never got the DAO synced (stays in DAO synchronizing). This might be another bug, but we should figure it out before applying that.

PR #6947, opened 2023-11-20
Fix initial data request handling

djing-chan:
This should fix the issues with dao hash conflicts when starting a node which was missing many blocks.

Now we wait until the updated data requests do not report that data is still missing (truncated flag). Only after that (when at least one seed responded with a non-truncated response) we consider the network initialized and trigger client code.

In the dao block requests there are some fixes as well which should make the block deliver more stable.

This work is based on @jmacxx draft PR https://github.com/bisq-network/bisq/pull/6946.

### DIFF
--- a/core/src/main/java/bisq/core/account/sign/SignedWitnessService.java
+++ b/core/src/main/java/bisq/core/account/sign/SignedWitnessService.java
@@ -140,7 +140,7 @@ public class SignedWitnessService {
         } else {
             p2PService.addP2PServiceListener(new BootstrapListener() {
                 @Override
-                public void onUpdatedDataReceived() {
+                public void onDataReceived() {
                     onBootstrapComplete();
                 }
             });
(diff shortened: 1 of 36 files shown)


HenrikJannsen (approved):
ACK

ghost (approved):
ACK
Tested for 1.5 days (mainnet) :

- new data directory, lite DAO node
- new data directory, full DAO node
- existing data directory
- checked the case where new blocks are mined during initial blocks download
- verify self calculated DAO hashes after sync
- seed node
- seed node, new data directory.

--- 

There is one issue where the trade statistics do not appear updated in the GUI until after restart -> that is an **existing issue** not something introduced by this PR change.  I mention it here because it is likely to be noticed by others when testing this.

djing-chan:
@jmacxx Great thanks for the extensive testing!!!

ghost:
**NB:** I think this PR needs to be thoroughly soak-tested for several weeks before considering safe to merge.  I've started that soak testing and encourage others to do so as well.

alejandrogarcia83 (approved):
utACK

PR #6949, opened 2023-11-26
Change burning man capping algorithm

stejbac:
Change the algorithm used to adjust & cap the burn share of each BM candidate to use an unlimited number of 'rounds', as described in https://github.com/bisq-network/proposals/issues/412. That is, instead of capping the shares once, then distributing the excess to the remaining BM, then capping again and giving any excess to the Legacy Burning Man, we cap-redistribute-cap-redistribute-... an unlimited number of times until no more candidates are capped. This has the effect of reducing the LBM's share and increasing everyone else's, alleviating the security risk of giving too much to the LBM (who is necessarily uncapped).

#### Implementation details

Instead of implementing the new algorithm directly, we simply enlarge the set of candidates who should be capped to include those who would eventually be capped by the new algorithm, in order to determine how much excess burn share should go to the remaining BM. Then we apply the original method, `candidate.calculateCappedAndAdjustedShares(..)`, to set each share to be equal to its respective cap or uniformly scaled upwards from the starting amount accordingly.
    
To this end, the static method `BurningManService.imposeCaps` is added, which determines which candidates will eventually be capped, by sorting them in descending order of burn-share/cap-share ratio, then marking all the candidates in some suitable prefix of the list as capped, iterating through them one-by-one & gradually increasing the virtual capping round (starting at zero) until the end of the prefix is reached. (The method also determines what the uncapped adjusted burn share of each BM should be, but that only affects the BM view & burn targets.) In this way, the new algorithm runs in guaranteed _O(n * log n)_ time.

#### Activation date

To prevent failed trades, the new algorithm is set to activate at time `DelayedPayoutTxReceiverService.PROPOSAL_412_ACTIVATION_DATE`, with a placeholder value of _12am, 1st January 2024 (UTC)_. This simply toggles whether the for-loop in `imposeCaps` should stop after capping round 0, since doing so will lead to identical behaviour to the original code (even accounting for FP rounding errors).

### DIFF
--- a/core/src/main/java/bisq/core/dao/burningman/BurningManService.java
+++ b/core/src/main/java/bisq/core/dao/burningman/BurningManService.java
@@ -73,10 +73,10 @@ public class BurningManService {
     private static final double GENESIS_OUTPUT_AMOUNT_FACTOR = 0.1;
 
     // The number of cycles we go back for the decay function used for compensation request amounts.
-    private static final int NUM_CYCLES_COMP_REQUEST_DECAY = 24;
+    static final int NUM_CYCLES_COMP_REQUEST_DECAY = 24;
 
     // The number of cycles we go back for the decay function used for burned amounts.
-    private static final int NUM_CYCLES_BURN_AMOUNT_DECAY = 12;
+    static final int NUM_CYCLES_BURN_AMOUNT_DECAY = 12;
 
     // Factor for boosting the issuance share (issuance is compensation requests + genesis output).
     // This will be used for increasing the allowed burn amount. The factor gives more flexibility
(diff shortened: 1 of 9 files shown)


stejbac:
I've just pushed another fairly minor commit to this branch, to clean up the test classes a little by replacing the deprecated `initMocks` method.

alejandrogarcia83:
@alvasw @HenrikJannsen @jmacxx Could any of you review this PR please? Thanks!

alejandrogarcia83:
@alvasw @HenrikJannsen @jmacxx Could anyone of you review this PR please? Thanks!

HenrikJannsen:
I am traveling at but will review as soon I find time.

ghost (approved):
utACK

HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6952, opened 2023-11-28
Display the reason for auto-disabling an open offer.

### LINKED ISSUE #6127: "Deactivating open offer due to mempool validation" does not produce any GUI explanation
### Description

The "Deactivating open offer due to mempool validation" error (introduced by https://github.com/bisq-network/bisq/pull/5160 ) does not result in any GUI prompt telling the user what's wrong.  The user must check the log file to gain any understanding of why their offer spontaneously disabled itself (and the GUI does not instruct the user to check the log).  This is a UX issue.

#### Version

v1.8.4.

### Steps to reproduce

1. Have an open offer that fails mempool validation.
2. Enable it by clicking the Status toggle.
3. Wait a few minutes.

### Expected behaviour

When the offer disables itself, the GUI should display a message somewhere (perhaps as an icon with popup next to the offer) that explains why the offer was disabled, and preferably give some suggestions on how to fix it.

### Actual behaviour

No indication in the GUI is given as to what's wrong.  The offer just disables itself without explanation.  User has to think of checking the logs to figure out what happened.

### Screenshots

N/A (issue is lack of GUI, not a bug in GUI that exists).

#### Device or machine

Haswell i7, Whonix 16 inside Qubes 4.1.

#### Additional info

N/A.

### LINKED ISSUE #6959: BSQ fee validation should check mempool.space in addition to the DAO.
In https://github.com/bisq-network/bisq/pull/6425, we switched validation of BSQ fee payments from mempool.space the DAO database.

> Unconfirmed transactions will not be able to have their BSQ fees checked so early requests for validation will skip the fee check. This is not a problem since maker fee validation is done by takers and **in the majority of cases will be already confirmed**; taker fee validation is done after the first confirm at trade step 2.

---

### Problem scenario:

- Maker has bad SPV wallet state.
- Maker creates an offer but their tx fails to propagate because of their bad SPV state.
- Taker initiates trade.  BSQ fees cannot be checked because DAO does not have the info.  
- Taker tx is published.
- Deposit tx fails due to missing maker input.

**Possible solution is to reintroduce querying mempool.space as part of the BSQ validation prior to trade.**

---

Issue raised by @suddenwhipvapor

ghost:
Fixes #6127.  Users have complained that when fee validation flags & disables their offer, they have no idea why.
This PR adds a status which is shown to the UI when an offer has been disabled.
The disabled reason is shown in Open Offers via the shield icon tooltip and globally via a notification.

Reasons an offer can be flagged:

- BTC fee tx not found
- BSQ fee tx not found
- Maker Fee underpaid
- Taker Fee underpaid
- Fee receiver not a recognised address
- JSON parsing failure

An offer can also become disabled if the user chooses a Trigger price  #4998.  This reason is also shown on the shield icon and notification. 

### Key features:

- Adds an enum of potential fee validation reasons.
- Shows the reason to Maker in Open Offers if there was an issue with the maker fee.  Previously would silently disable the offer.
- Shows the reason to Taker in Take Offer, if there was an issue with the maker fee.  Previously would say `maker validation failed`.
- Shows the reason to Maker & Taker in Trade Step 2, if there was an issue with the taker fee.  Previously would say `taker validation failed`.

### Bug Fix:

In prior releases, sometimes there would be a false flagging of the offer when checked at startup because the BSQ tx was not yet available.  This PR includes a fix to postpone the initial check until a minute after the boostrapping.

### Screenshots:

Bisq_2023-11-24_11-13-12

&nbsp;

Bisq_2023-11-28_10-13-07

&nbsp;

Bisq_2023-11-28_13-18-41

---

Fixes #6959

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqHeadlessApp.java
+++ b/core/src/main/java/bisq/core/app/BisqHeadlessApp.java
@@ -79,6 +79,7 @@ public class BisqHeadlessApp implements HeadlessApp {
         bisqSetup.setSpvFileCorruptedHandler(msg -> log.error("onSpvFileCorruptedHandler: msg={}", msg));
         bisqSetup.setChainFileLockedExceptionHandler(msg -> log.error("onChainFileLockedExceptionHandler: msg={}", msg));
         bisqSetup.setDiskSpaceWarningHandler(msg -> log.error("onDiskSpaceWarningHandler: msg={}", msg));
+        bisqSetup.setOfferDisabledHandler(msg -> log.error("onOfferDisabledHandler: msg={}", msg));
         bisqSetup.setChainNotSyncedHandler(msg -> log.error("onChainNotSyncedHandler: msg={}", msg));
         bisqSetup.setLockedUpFundsHandler(msg -> log.info("onLockedUpFundsHandler: msg={}", msg));
         bisqSetup.setShowFirstPopupIfResyncSPVRequestedHandler(() -> log.info("onShowFirstPopupIfResyncSPVRequestedHandler"));
--- a/core/src/main/java/bisq/core/app/BisqSetup.java
+++ b/core/src/main/java/bisq/core/app/BisqSetup.java
@@ -164,7 +164,7 @@ public class BisqSetup {
             filterWarningHandler, displaySecurityRecommendationHandler, displayLocalhostHandler,
             wrongOSArchitectureHandler, displaySignedByArbitratorHandler,
             displaySignedByPeerHandler, displayPeerLimitLiftedHandler, displayPeerSignerHandler,
-            rejectedTxErrorMessageHandler, diskSpaceWarningHandler, chainNotSyncedHandler;
+            rejectedTxErrorMessageHandler, diskSpaceWarningHandler, offerDisabledHandler, chainNotSyncedHandler;
     @Setter
     @Nullable
     private Consumer<Boolean> displayTorNetworkSettingsHandler;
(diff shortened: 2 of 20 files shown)


ghost:
Added https://github.com/bisq-network/bisq/pull/6952/commits/265ca17e327fc659e433474bdc4b5eb0468bc956 Fixes #6959

ghost:
I created this PR as a result of numerous requests during the year, both via #6127 and Matrix chat.
Maintainers, would you prefer the two included bugfixes moved into separate PRs?

pazza83:
Looks like a great addition.

I think the trade amount being higher than a user's Bisq's 'trade limit' settings is also a common reason why offers are not shown.

ghost:
Not true @pazza83.  A user's offer is de-activated (a) while a user edits it. (b) if the trigger price is breached (c) if the fee validation fails.

Indicating the reason for an auto-disable is a nice-to-have, but this PR has included two bug fixes which are higher priority, i.e. they fix issues that can cause failed trades.  It would be disappointing if the bug fixes missed the `1.9.15` release due to this larger feature being left on the shelf.  Hence my question above, should the bug fixes be separated out so they can be reviewed with higher priority?

pazza83:
Ok, thanks for clarifying. I was confused about how the trade limit setting works

alvasw (approved):
utACK

alejandrogarcia83:
@alvasw Thank you for reviewing this PR! :-)

alejandrogarcia83 (approved):
utACK

PR #6964, opened 2023-12-11
Update Wise payment account (was TransferWise)

### LINKED ISSUE #6934: rename TransferWise to Wise
### Steps to reproduce

Account->National Currency Accounts->Add New Account-> (payment method)

### Expected behaviour

- TransferWise should be renamed to Wise
- Transferwise-USD should be renamed to Wise-USD
- In the info popup dialog when opening Transferwise-USD method: word "formally" should be replaced with "formerly"

ghost:
Fixes #6934 

- Rename TransferWise to Wise.
- Rename TransferWise-USD to Wise-USD.
- Update the list of supported currencies (Wise).
- Add HolderName, mandatory for new account creation.


core/src/main/java/bisq/core/payment/payload/TransferwiseAccountPayload.java
```diff
@@ -94,6 +96,8 @@ public String getPaymentDetailsForTradePopup() {
 
     @Override
     public byte[] getAgeWitnessInputData() {
-        return super.getAgeWitnessInputData(email.getBytes(StandardCharsets.UTF_8));
+        return super.getAgeWitnessInputData(ArrayUtils.addAll(
+                email.getBytes(StandardCharsets.UTF_8),
+                getHolderName().getBytes(StandardCharsets.UTF_8)));
```
HenrikJannsen, line 103: Doesn't that break the Account age witness verification of existing accounts?

HenrikJannsen (changes requested):
See comment to clearify

core/src/main/java/bisq/core/payment/payload/TransferwiseAccountPayload.java
HenrikJannsen: Ah sorry, overlooked that you set an empty string which results as an empty byte array at getBytes() thus having not effect on the byte array we use for the witness.

HenrikJannsen: Maybe a comment regarding the backward compatibility and the means to enable it would be helpful.

core/src/main/java/bisq/core/payment/payload/PaymentAccountPayload.java
```diff
@@ -129,6 +131,14 @@ public void setSalt(byte[] salt) {
         excludeFromJsonDataMap.put(SALT, Utilities.encodeToHex(salt));
     }
 
+    public String getHolderName() {
+        return excludeFromJsonDataMap.getOrDefault(HOLDER_NAME, "");
+    }
+
+    public void setHolderName(String holderName) {
+        excludeFromJsonDataMap.compute(HOLDER_NAME, (k, v) -> Strings.emptyToNull(holderName));
```
HenrikJannsen, line 140: Maybe a comment why to use compute here would be helpful (I see the reason, to ensure an empty string will result in the removal of the map entry as a null result in the mapping removes the entry).

HenrikJannsen (approved):
utACK

ghost:
Code comments added, per code review.

HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

alvasw:
Hi @yonson2023,
I found some time to review this PR to make sure everything is working. Unfortunately, there are some issues: the seller doesn't see the buyer's holder name, the holder name is missing on the trade details popup, and traders can't copy the holder name (all other fields have a copy button next to its field).

### Seller doesn't see the buyer's holder name
1. Create new Wise accounts with an e-mail address and a holder name on Alice's and Bob's Bisq instances.  
2. Start a trade between Alice and Bob  
3. Press the payment started button on the buyer's side  
4. The seller doesn't see the holder name of the buyer 
seller_holder_missing

### Holder name is missing on both traders trade details popup
trade_details_everything_missing

### Copy button missing (next to the holder name text field)
copy_button_missing

@yonson2023 Do you have time to fix these issues? Otherwise another dev can fix it.

PR #6965, opened 2023-12-11
Update Pix payment account, add HolderName

ghost:
- Add HolderName, mandatory for new accounts.
- Older, existing Pix accounts do not have HolderName, and retain their age status.
- If buying from an old account that does not have HolderName, Trader Chat can be used.

@pazza83 

---

pix1

---

pix2

---

pix3

PR #6968, opened 2023-12-16
Update MoneyBeam account, add HolderName

ghost:
Same procedure as #6965.

- Add HolderName, mandatory for new accounts.
- Older, existing MoneyBeam accounts do not have HolderName, and retain their age status.
- If buying from an old account that does not have HolderName, Trader Chat can be used.

@pazza83

---

moneybeam-1

---

moneybeam-3

---

moneybeam-2

### DIFF
--- a/core/src/main/java/bisq/core/payment/payload/MoneyBeamAccountPayload.java
+++ b/core/src/main/java/bisq/core/payment/payload/MoneyBeamAccountPayload.java
@@ -92,8 +92,8 @@ public final class MoneyBeamAccountPayload extends PaymentAccountPayload {
 
     @Override
     public String getPaymentDetailsForTradePopup() {
-        return (getHolderName().isEmpty() ? "" : Res.getWithCol("payment.account.owner") + " " + getHolderName() + "\n") +
-                Res.getWithCol("payment.account") + " " + accountId;
+        return Res.getWithCol("payment.account") + " " + accountId + "\n" +
+                Res.getWithCol("payment.account.owner") + " " + getHolderNameOrPromptIfEmpty();
     }
 
     @Override
--- a/core/src/main/java/bisq/core/payment/payload/MoneyBeamAccountPayload.java
+++ b/core/src/main/java/bisq/core/payment/payload/MoneyBeamAccountPayload.java
@@ -21,6 +21,8 @@ import bisq.core.locale.Res;
 
 import com.google.protobuf.Message;
 
+import org.apache.commons.lang3.ArrayUtils;
+
 import java.nio.charset.StandardCharsets;
 
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


alvasw (changes requested):
Thank you for opening this PR @yonson2023!

Here are the needed changes to get your PR merged:

1. The start payment screen shows an empty "Account owner full name" text field for old accounts during a trade. This could confuse users and increase the number of support requests. It'd be better to hide this field for old accounts (without a holder name). 
start_payment_hide_account_owner_field_if_empty

2. The "Account owner full name" text field is empty for old accounts on the "Manage accounts" screen (Accounts -> Manage Accounts -> Select an old MoneyBeam account). It's better to hide the text field for empty holder names (old accounts). 
dont_show_holder_name_test_field_if_old_account

3. The text fields on the "Start payment using ..." screen are below each other. This is inconsistent to our other account types. There should be two text fields next to each other (see SEPA example). 

MoneyBeam:

moneybeam_new_screen_fields_below_each_other

SEPA:

sepa_text_fields_next_each_other

ghost:
>  text fields on the "Start payment using ..." screen are below each other. This is inconsistent to our other account types. There should be two text fields next to each other (see SEPA example).

Fixed.

----


Regarding cases when an old account supplies no Account Holder Name.  I think it is **important** to show that the field was not supplied (is empty) rather than hiding it from view.  It prompts the user to either (a) update/recreate his own account with the necessary info, or (b) contact his trading peer via trader chat to request the info for successful payment verification (which also lets the trading peer know they need to update/recreate their account information).

pazza83:
If the account has not provided a name because their account was created prior to name being required please can we show "Ask trader to provide account name if needed"

This could shown in the account summary, payment receipt, payment confirmation, and support ticket screens etc.

alvasw:
> Regarding cases when an old account supplies no Account Holder Name. I think it is important to show that the field was not supplied (is empty) rather than hiding it from view. It prompts the user to either (a) update/recreate his own account with the necessary info, or (b) contact his trading peer via trader chat to request the info for successful payment verification (which also lets the trading peer know they need to update/recreate their account information).

What do you think about writing into text field that the user can ask the other party his holder name? Additionally, we can add text on the trader's side with the old MoneyBeam account that he has to tell the other party his holder name when asked.

> If the account has not provided a name because their account was created prior to name being required please can we show "Ask trader to provide account name if needed"
> 
> This could shown in the account summary, payment receipt, payment confirmation, and support ticket screens etc.

I looked into the account aging/signing code. We could migrate all "MoneyBeam" accounts to "MoneyBeam accounts with holder name" while retaining the old account age and adding the holder name to the salt. Afterward, we can show a popup before starting/taking a trade to users with old accounts to enter their holder name. But it's too much for this PR.

ghost:
Done.

moneybeam1

alvasw (approved):
ACK

alejandrogarcia83 (approved):
utACK

PR #6972, opened 2023-12-20
Bugfix: avoid standby mode should adhere to the user preference setting.

ghost:
On systems not supporting direct power management it plays the audio file regardless of the user preference setting.  Ref: #6971

### Fix:

- `init()` was called before preferences had been loaded causing the audio file to be played even if not desired.
- A line of code in the user preference listener was missing, causing audio to be played but never stopped.


alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #6980, opened 2023-12-29
core: Don't add old filter to invalid filters if new signer banned

alvasw:
Before this change, the currentFilter will be added to the invalidFilters list after comparing both filters creation date and
before checking whether the new filter was created by a banned signer.

Note: The invalid filters list is only used on the security manager Bisq instances. This list is never read by regular user clients.


alejandrogarcia83 (approved):
utACK

Issue #6948, opened 2023-11-25
False deposit confirmation messages gives option for user to move the trade to 'Failed Trades' increasing likelihood of trade ending in arbitration.


pazza83:
### Description

On occasion Bisq client does not recognize the deposit confirmation ID.

This seems to happen in the following circumstances:

- Deposit confirmation takes over 4 hours to confirm, I think Bisq stops checking after 4 hours. Might be another number. I think this is in the code. As a result it recommends an SPV resync.
- User does an SPV resync at some point during the trade process that gives a false error. Perhaps related to above. Eg user gets error message due to deposit not confirming in 4 hours, recommends them to resync. They do an SPV resync and this shows them to trade is failed. 
 
User is promoted to move the trade to failed trade despite their being a confirmed deposit confirmation.

Example with screenshots:

![image](https://github.com/bisq-network/bisq/assets/66367547/136c9d89-cfff-43bf-be32-a082bab2eaf3)
Trade showing as failed when it has not in open offers.

![image2](https://github.com/bisq-network/bisq/assets/66367547/a87c64ef-3ab0-4ff5-a776-98e6d5cf3206)
Trade information screen showing trade has failed when it has not.

Deposit Transaction ID for the trade was: [7e7189d09c3e11803650be541ef079ff8744b971c3d02e50aa1a852aa6385e46](https://mempool.space/tx/7e7189d09c3e11803650be541ef079ff8744b971c3d02e50aa1a852aa6385e46)

This causes some users to move the trade to failed. This results in:

1a. Trader peer if seller not receiving payment.
1b. Trade peer if buyer sending payment and not getting their bitcoin released.
2. Trade peer opening mediation
3. Mediator not able to contact the user who put the trade in failed
4. Trade peer not able to contact the user who put the trade in failed
5. Mediator makes proposal (unresponsive peer)
6. Trade goes to arbitration
7. Arbitration ticket not opened for user who put the trade in failed
8. User with false error message often oblivious to the fact they lost funds. They might receive a partial payout from the arbitrator into their Bisq wallet but either not notice or be oblivious to where the bitcoin has come from. Either way they have lost their security deposit.

#### Version

1.9.4

### Steps to reproduce

One of the issues with the above is it is more likely to happen when the mempool is busy, as trade confirmation can often take over 4 hours. Bisq prompts the user to do an SP resync and as a result the trade may show as failed.

[...]

PR #6994, opened 2024-01-05
Change unconfirmed popup to 24h after trade start.

ghost:
Ref: #6948
@pazza83 as discussed, here is the change to the unconfirmed trade popup:  

- Instead of 4 hours now is 24 hours.
- A button to resync is included.

Let me know if you want any wording, etc changes.

resync_popup

Existing conirmation prompt is shown before starting resync process:

resync_popup2

### DIFF
--- a/core/src/main/resources/i18n/displayStrings.properties
+++ b/core/src/main/resources/i18n/displayStrings.properties
@@ -1564,7 +1564,7 @@ settings.net.peer=Peer
 settings.net.inbound=inbound
 settings.net.outbound=outbound
 settings.net.reSyncSPVChainLabel=Resync SPV chain
-settings.net.reSyncSPVChainButton=Delete SPV file and resync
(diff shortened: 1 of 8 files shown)


pazza83:
Thanks @jmacxx 

With regards the following warning message:

294575239-2b4205f9-55c5-49a2-8c84-84d02fa9c810

Would it be possible for there to be a deposit transaction lookup prior to showing the message?

- If the deposit transaction was still unconfirmed no error message needs to be shown
- If the deposit transaction was confirmed on the blockchain but not showing in Bisq than an error message can be shown prompting the user to do an SPV resync

ghost:
> Would it be possible for there to be a deposit transaction lookup prior to showing the message?

@pazza83 - done.

alvasw (approved):
Good!

utACK

alejandrogarcia83 (approved):
utACK

PR #6995, opened 2024-01-06
Create GetBlocksRequestHandler's reply list once

alvasw:
Previously, the onGetBlocksRequest method created an ArrayList and two LinkedList before creating the GetBlocksResponse. The first two lists were never used. PR #6947 introduced the second LinkedList creation. With this change, the GetBlocksRequestHandler only creates a single LinkedList.


ghost (approved):
ACK

alejandrogarcia83 (approved):
utACK

Issue #6996, opened 2024-01-06
Invalid PayoutTxPublished message closes trade leaving funds locked


ghost:
@pazza83 reports some trades are moved to `closed` having funds still locked in the multisig.

- Logs show payout tx paying a different address than was in the contract.  Bitcoin core rejected the tx due to signature not matching the script.  The funds remained locked.
- Bisq accepted the `PayoutTxPublished` P2P message and closed the trade.

### Suggested resolution

Bisq should validate the received P2P message, and if any payout address does not match the trade contract, display an error and leave the trade open.  Then the user will be able to get support from a mediator.

---

### Question / Alternate resolution

Bisq already closes a trade when it detects a payout tx in its wallet.  The additional path to close a trade upon receipt of a P2P message seems unnecessary -> would it make sense to remove the duplicate path?

ghost:
After investigating a further batch of 22 closed trades matching this scenario, I noticed the reason for the payout address being different.  The code called [getOrCreateAddressEntry(id, TRADE_PAYOUT)](https://github.com/bisq-network/bisq/blob/92283de42a34d3b26a2b92605d0c65467fc4ab36/core/src/main/java/bisq/core/trade/protocol/bisq_v1/tasks/seller/SellerSignAndFinalizePayoutTx.java#L65-L66) which creates a new address if one is not found.  The resulting new address would cause transaction failure because the buyer has signed their version of the payout to the addresses specified in the contract.  The examples from @pazza83 show that sometimes AddressEntryList is missing data, presumably due to a catastrophe causing them to restore a backup / PendingTrades but not AddressEntryList.  This is backed up by the following code comment:

https://github.com/bisq-network/bisq/blob/92283de42a34d3b26a2b92605d0c65467fc4ab36/core/src/main/java/bisq/core/trade/protocol/bisq_v1/tasks/seller/SellerSignAndFinalizePayoutTx.java#L75-L82

The code already attempts to handle this scenario by obtaining the signing key from the wallet:

https://github.com/bisq-network/bisq/blob/92283de42a34d3b26a2b92605d0c65467fc4ab36/core/src/main/java/bisq/core/trade/protocol/bisq_v1/tasks/seller/SellerSignAndFinalizePayoutTx.java#L83-L87


In addition to that it needs to use the payout address from the trade contract (instead of a fresh address), to ensure a successful payout:

    sellerPayoutAddressString = checkNotNull(trade.getContract()).getSellerPayoutAddressString();


[...]

PR #6998, opened 2024-01-09
Allow mediation of failed and closed trades

ghost:
#6948 results in some pending trades falsely moved to failed.  #6996 results in some pending trades falsely moved to closed.

@pazza83 has requested that a dispute ticket initiated from the peer (via the mediator) should open on a client if their trade is pending, closed or failed.  The closed or failed trade will be reverted to pending so that the mediation can be processed in a normal manner.

This helps address the non-responding peer issue, as the non-responding peer will get a support ticket notification.


alvasw (approved):
utACK

A unlikely edge case, we can ignore for now: The `FailedTradesManager` could've cleared the sensitive data of a trade (if user manually set it to a low number of days) before the other peer opens mediation.

alejandrogarcia83 (approved):
utACK

PR #7005, opened 2024-01-16
Add SEPA QR Code for buyer payment

ghost:
Implements request https://github.com/bisq-network/bisq/discussions/7001 Allowing to initiate Sepa Payment by scanning QR Code
The `Show QRC` setting is saved in preferences for subsequent use. 

qrc-1

qrc-2

qrc-3


hoschi-it:
Thanks for implementing!
I would suggest changing the text to "Show QR Code", as the shortage "QRC" is not widely known.

ghost:
> I would suggest changing the text to "Show QR Code", as the shortage "QRC" is not widely known.

Done: qrc-4

ghost:
It would be good if someone could confirm that the QR code scans ok with an actual SEPA bank app.

alvasw:
Hi @jmacxx,
sorry for the delay! I found someone to test the QR Code with a SEPA bank app - still waiting to hear back from him.

alvasw (changes requested):
Nit

The QR Code works! I received video recordings of users testing four SEPA banking apps.

#### Feedback
- We can show the QR Code by default because most SEPA banking apps support QR Codes.
- After scanning the QR Code users can't double check the data because it's hidden. I think we should show the BIC and IBAN while showing the QR Code (instead of hiding it).
- Two of the tested banking apps sent the image to the banking servers to read the QR Code. In both images the text "[...] trade ID or other text like 'bitcoin', 'BTC', or 'Bisq. [...]" was visible. Banks that do OCR processing could flag users' accounts. We should add a softbreak to the text, so that it's never near the QR Code.

ghost:
qrc-5

https://github.com/bisq-network/bisq/pull/7005/commits/6bc79e5be2e105934cbb841e3bbb089ca8e257bf Addressing your comments.  Note there is a bug with the scroll bar which affects buyer forms that have many vertical elements.  Which is why I prefer the toggle button to switch between right column fields and QRC, to limit the vertical size of the form.

hoschi-it:
> Two of the tested banking apps sent the image to the banking servers to read the QR Code. In both images the text "[...] trade ID or other text like 'bitcoin', 'BTC', or 'Bisq. [...]" was visible. Banks that do OCR processing could flag users' accounts. We should add a softbreak to the text, so that it's never near the QR Code.

Do you think, showing the QR code in full screen (on click or something) would be useful to mitigate that risk completely?

ghost:
> Do you think, showing the QR code in full screen (on click or something) would be useful to mitigate that risk completely?

:+1: 

Per your suggestion, perhaps instead of showing the small QR code (which has issues with limited screen real estate), simply a button to show it like this in a popup, :

qrc-7

pazza83:
Can we hide the text, and make the screen full size with just the QR in the centre to protect privacy

ghost:
> Can we hide the text, and make the screen full size with just the QR in the centre to protect privacy  :+1:

qrc-a1

qrc-a2

[Note: code & image updated after request from @pazza83 (below) to vertically center the QRC].

pazza83:
@jmacxx is it possible to add a few more line breaks at the top to give more 'padding / blank space' between the menu bar and the image. To make it easier for a user not to accidently include the Bisq menu bar in their screenshot

ghost:
Done, see https://github.com/bisq-network/bisq/pull/7005#issuecomment-1948990831 for update.

alvasw:
It looks a lot better! What do you think about switching to a QR code icon. I tried out two variations and prefer the one with the icon only. What do you think?

Icon with text:
qr_code_icon_and_text
Icon only:
qr_code_only_icon

ghost:
@alvasw I think this looks good.  

qrc-b2

---

Also added a one-time message explaining why the QR is shown so large.

qrc-b1

alejandrogarcia83 (approved):
utACK

PR #7023, opened 2024-02-01
GetTrades should not filter open failed trades

### LINKED ISSUE #7022: Open failed trade not returned by RPC method `GetTrades`
### Description

An open trade that has failed, but which has not been moved to failed trades, is not returned by `GetTrades` API method.
None of the category parameters (`OPEN`, `CLOSED` or `FAILED`) are returning this trade,
Once the trade is moved to failed trades, it appears in `FAILED` trade.

#### Version

v1.9.14

### Steps to reproduce

1. Have a failed trade
2. Keep it in open trades list (do not move it to failed trades)
3. Call API method `GetTrades` with category parameter `OPEN`, `CLOSED` and `FAILED`
4. Observe that the trade is not returned by any API call

### Expected behaviour

Trade should be listed under `OPEN` trades.

### Actual behaviour

Trade is not listed by calling `GetTrades` API method with  category parameter `OPEN`

#### Device or machine

Debian, Bisq v1.9.14

ghost:
Fixes #7022

### DIFF
--- a/core/src/main/java/bisq/core/trade/TradeManager.java
+++ b/core/src/main/java/bisq/core/trade/TradeManager.java
@@ -108,6 +108,7 @@ import javafx.collections.ObservableList;
 
 import org.bouncycastle.crypto.params.KeyParameter;
 
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -911,9 +912,7 @@ public class TradeManager implements PersistedDataHost, DecryptedDirectMessageLi
     }
 
     public List<Trade> getTrades() {
-        return getObservableList().stream()
-                .filter(t -> !t.hasFailed())
-                .collect(Collectors.toList());
+        return Collections.unmodifiableList(getObservableList());
     }
 
     private void removeTrade(Trade trade) {
--- a/core/src/main/java/bisq/core/api/model/TradeInfo.java
+++ b/core/src/main/java/bisq/core/api/model/TradeInfo.java
@@ -93,6 +93,8 @@ public class TradeInfo implements Payload {
     private final boolean isCompleted;
     private final String contractAsJson;
     private final ContractInfo contract;
+    private final boolean hasFailed;
+    private final String errorMessage;
     // Optional BSQ swap trade protocol details (post v1).
     private BsqSwapTradeInfo bsqSwapTradeInfo;
     private final String closingStatus;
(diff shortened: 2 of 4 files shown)


dutu:
Thanks @jmacxx, I've applied and tested the change on v1.9.14 and it works as intended.

However, I could not see an obvious way to differentiate a trade that has failed from a valid trade.
Would it be possible to expose to the API an additional trade field (such as `hasFailed`, `errorMessage` or `isMaybeInvalidTrade`)?

The idea is to be able to use `FailTrade()`, which can only be done if the trade is identified. 

Note: The issue with leaving a failed trade in the open trade list, is when the offer is made available again and a new trade is taken. The new trade will have the same id as previously failed trade (the id of the offer). I have previously encountered issues when calling a trade API method complaining the trade is in an invalid state (it seems that due to two  trades with same id, the failed trade was picked up when calling the method).

core/src/main/java/bisq/core/trade/TradeManager.java
```diff
@@ -911,9 +912,7 @@ public Optional<Trade> getTradeById(String tradeId) {
     }
 
     public List<Trade> getTrades() {
-        return getObservableList().stream()
-                .filter(t -> !t.hasFailed())
-                .collect(Collectors.toList());
+        return new ArrayList<>(getObservableList());
```
alvasw: We can avoid the redudant copying of the trade list by using `Collections.unmodifiableList(getObservableList())`.

alvasw (changes requested):
Nit

I know that someone else wrote the original code but we should try to clean up existing code when working on it. We can avoid the redudant copying of the trade list by using `Collections.unmodifiableList(getObservableList())`.

Apart from that I aggree with @dutu. API consumers should be able to tell whether a trade failed.

ghost:
> Would it be possible to expose to the API an additional trade field (such as hasFailed, errorMessage

Added `has_failed` and `error_message` properties to API `TradeInfo` record.

core/src/main/java/bisq/core/trade/TradeManager.java
```diff
@@ -108,6 +108,8 @@
 
 import org.bouncycastle.crypto.params.KeyParameter;
 
+import java.util.ArrayList;
```
dutu: not required anymore

alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #7025, opened 2024-02-03
Fix resource path URL decoding in tx validator tests and clean similar tests

stejbac:
Prevent tests calling `MakerTxValidatorSanityCheckTests::getValidBtcMakerFeeMempoolJsonResponseString` from failing when running the build from a working directory with spaces in its absolute path, due to the fact that the test resource paths returned by `Class::getResource` are URL encoded.

Also do some minor tidying of other tests which load resources, and replace manual `MockitoSession` setup & tear down in tests which still do that, now that `mockito-junit-jupiter` has been added to the build dependencies.


alvasw (approved):
ACK

alejandrogarcia83 (approved):
utACK

PR #7035, opened 2024-02-13
 Update data stores for v1.9.15

alejandrogarcia83:
light_1
full_1
light_2
full_2
light_3
full_3
light_4
full_4


alejandrogarcia83:
@alvasw @HenrikJannsen @jmacxx Could anyone of you ACK this PR? Thanks!

common/src/main/java/bisq/common/app/Version.java
```diff
@@ -45,7 +45,7 @@ public class Version {
      */
     public static final List<String> HISTORICAL_RESOURCE_FILE_VERSION_TAGS = Arrays.asList("1.4.0", "1.5.0", "1.5.2",
             "1.5.5", "1.5.7", "1.6.0", "1.6.3", "1.6.5", "1.7.2", "1.7.4", "1.8.0", "1.8.1", "1.8.3", "1.9.0", "1.9.3",
-            "1.9.5", "1.9.6", "1.9.7", "1.9.10", "1.9.11", "1.9.13");
+            "1.9.5", "1.9.6", "1.9.7", "1.9.10", "1.9.11", "1.9.13", "1.9.15");
```
ghost, line 48: `VERSION = "1.9.15"`
`HISTORICAL...TAGS` would have 1.9.14 (the prior version) appended.

ghost (changes requested):
Please also consider setting `PROPOSAL_412_ACTIVATION_DATE` which should be set to a few weeks after release date.  See https://github.com/bisq-network/proposals/issues/412#issuecomment-1492851657


> As both traders need to have the exact same distribution for the delayed payout transaction we would need to roll that change out with a activation blockheight sufficiently far in the future so that we can be sure to release and enforce the updated version.

HenrikJannsen:
I can confirm that my hashes on a light node are the same.

common/src/main/java/bisq/common/app/Version.java
alvasw: This is correct! 1.9.14 was a hotfix release without data store updates.

alvasw (approved):
utACK

DAO hashes match with my nodes too.

PR #7036, opened 2024-02-14
Add API for payment started with XMR auto conf params

### LINKED ISSUE #7033: Cannot provide XMR tx id and tx key with RPC method ConfirmPaymentStarted
### Description

When using RPC method `ConfirmPaymentStarted` for XMR trades, user is not able to provide transaction id and transaction key.

#### Version

v1.9.14

### Steps to reproduce

RPC method `ConfirmPaymentStarted()` takes only one parameter (the `trade_id`).

### Expected behaviour

`ConfirmPaymentStarted()` should allow two additional (optional) parameters for transaction id and transaction key, `tx_id` (or`tx_hash`) and `tx_key`.
This would be consistent with how GUI works and improve user experience.

### Actual behaviour

API user cannot provide transaction id and transaction key with RPC method `ConfirmPaymentStarted()`

#### Device or machine

Debian, Bisq v1.9.14

ghost:
Fixes #7033

_Basedon / follow-on from #7023._

Allows `tx-id` & `tx-key` to be specified for XMR auto-confirm.

### DIFF
--- a/cli/src/main/java/bisq/cli/CliMain.java
+++ b/cli/src/main/java/bisq/cli/CliMain.java
@@ -586,6 +586,19 @@ public class CliMain {
                     out.printf("trade %s payment started message sent%n", tradeId);
                     return;
                 }
+                case confirmpaymentstartedxmr: {
+                    var opts = new GetTradeOptionParser(args).parse();
+                    if (opts.isForHelp()) {
+                        out.println(client.getMethodHelp(method));
+                        return;
+                    }
+                    var tradeId = opts.getTradeId();
+                    var txId = opts.getTxId();
+                    var txKey = opts.getTxKey();
+                    client.confirmPaymentStartedXmr(tradeId, txId, txKey);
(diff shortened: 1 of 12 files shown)


dutu:
It works great, thank you!

alvasw (changes requested):
Nit

We can simplify the null and `isEmpty` handling with `Optional<String>`s. Moreover `Optional`s enforce us to handle both cases and avoid `NullPointerException`s.

dutu:
@jmacxx are we not going to have this change implemented?

ghost:
I had to fix merge conflicts with the current master branch.  Also, I don't know how to implement @alvasw suggestion above.

dutu:
From what I can see, you have implemented `null` and `isEmpty` checks in line with existing code done for other parameters. 

@alvasw, for consistency sake would it make more sense to change these checks based on your suggestions separately, when it's done for all parameters?

alejandrogarcia83 (approved):
utACK

PR #7043, opened 2024-02-20
Account for invalid tx BSQ burns in supply & market cap charts and optimise the latter

stejbac:
Use `Tx::getBurntBsq` instead of `Tx::getBurntFee`, so as not to exclude BSQ burned by invalid txs from the supply calculations. There are no invalid BSQ txs at present on mainchain, but accidentally burned BSQ should definitely count as a reduction in supply, so this fixes a bug.

Also tidy `DaoChartDataModel` & `PriceChartDataModel` a little and somewhat speed up the market cap calculations in the latter, by fixing a quadratic time bug in `getBsqMarketCapByInterval`.

--

Prior to the above changes, the following hotspots in the market cap chart display are revealed by JProfiler, when changing the date interval size selection 12 times to recompute the chart:
Screenshot from 2024-02-20 21-25-20

After the changes, the following hotspots are seen instead:
Screenshot from 2024-02-20 21-28-45

Thus, a roughly twofold speed improvement is seen.

It isn't clear to me how best to reduce the next biggest hotspot from the stream pipeline in `getAveragePriceFromDateFilter`. Interestingly, changing it to stream directly over the `TreeSet` of `TradeStatistics3` instances, instead of an `ObservableSet` wrapper, actually makes it slower. I think this is due to the native `TreeSet` spliterator being less performant for _serial_ streaming than the default `IteratorSpliterator` that winds up being used from the `ObservableSet` delegation (due to JavaFX not being aware of the stream API).

### DIFF
--- a/desktop/src/main/java/bisq/desktop/components/chart/ChartDataModel.java
+++ b/desktop/src/main/java/bisq/desktop/components/chart/ChartDataModel.java
@@ -23,7 +23,9 @@ import java.time.Instant;
 import java.time.temporal.TemporalAdjuster;
 
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.BinaryOperator;
+import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.ToLongFunction;
 import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


ghost (approved):
ACK

alejandrogarcia83 (approved):
utACK

Issue #4940, opened 2020-12-13
UI / Performance: Remove dependencies on AWT (system tray icon)


cd2357:
As per https://stackoverflow.com/a/11121207 having AWT dependencies in a JavaFX application can lead to more CPU/RAM usage than necessary, because two separate graphics stacks are loaded and active at the same time.

From what I can tell, Bisq mainly uses AWT for the system tray icon.

If that's the case, then with AWT we have:

* benefits: ability to show Bisq in the system tray
* drawbacks: a separate graphics stack (older, superseded by JavaFX) loaded in Java, with possible impact on RAM and CPU usage

To me, it seems the drawbacks outweigh the benefits.

Therefore, I suggest to remove the AWT dependencies from Bisq (and with that, the system tray icon).

cd2357:
Looks like we also use AWT for the webcam/QR code reader, via the class `java.awt.image.BufferedImage`.

As per https://stackoverflow.com/questions/30970005/bufferedimage-to-javafx-image there is a JavaFX equivalent to that.

ripcurlx:
> Looks like we also use AWT for the webcam/QR code reader, via the class `java.awt.image.BufferedImage`.
> 
> As per https://stackoverflow.com/questions/30970005/bufferedimage-to-javafx-image there is a JavaFX equivalent to that.

Besides that it also stopped working with Catalina as it doesn't work without notarization anymore to access the webcam. So I think it would be good to tackle the webcam part anyways.

ripcurlx:
Related to #4374.

ghost:
There remains outstanding issue of the library `com.github.sarxos.webcam` which uses awt in its public interface and so puts `java.awt.image.BufferedImage` and `java.awt.Dimension` into Bisq code.  To remove awt from the runtime stack the whole webcam library will need to change.

cd2357:
A good non-AWT webcam alternative might be [JavaCV](https://github.com/bytedeco/javacv) (see [sample](https://github.com/rladstaetter/javacv-webcam)).

ripcurlx:
And we really need to get rid of it now:

```
{
  "logFormatVersion": 1,
  "jobId": "b28d2a4f-7f18-4539-ac05-a2858b7cf1f2",
  "status": "Invalid",
  "statusSummary": "Archive contains critical validation errors",
  "statusCode": 4000,
  "archiveFilename": "Bisq-1.6.2.dmg",
[...]

PR #7050, opened 2024-03-01
Restore QR code scanner feature for mobile notification pairing

devinbileck:
This restores the functionality that was removed in b5beea58. This implementation utilizes the JavaCV library rather than the webcam-capture library for reasons discussed in #4940.

I have tested this on both Linux (Ubuntu 23.10) and macOS (Sonoma 14.4).


alejandrogarcia83 (changes requested):
NACK

The CI build fails on macOS (see https://github.com/bisq-network/bisq/actions/runs/8447335400/attempts/1#summary-23137520876),

devinbileck:
@alejandrogarcia83 I am not sure why that failed. It does not provide much detail.
I re-ran it on my fork and it passed: https://github.com/devinbileck/bisq/actions/runs/8447335253
Would you be able to re-run it, since I dont seem to have the ability to, in order to see if it was just a transient failure?

alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #7070, opened 2024-04-01
Increase maxItems at requesting  trade statistic items

HenrikJannsen:
Increase maxItems from 3000 to 15000. 3000 items are about 180.325 kB.

For nodes not being online for longer the repeated requests consumes quite some time. With 15k we can expect a 1 MB payload which is still acceptable.

The past 6 months had about 18 000 trade statistic items.


alejandrogarcia83 (approved):
utACK

PR #7077, opened 2024-04-15
Use BM accounting resource file from Devin's seednode

ghost:
There was an issue reported on Matrix where the BurningMan accounting report was missing some data.
This commit corrects that by using Devin's seednode datastore as supplied on April 6th, 2024.

---

Investigation and several full node resyncs from scratch revealed the following differences between the release's resource file vs. what was expected.  A difference of 0.67 distributed BTC missing from the accounting records for the months July/Aug 2023. 

boilingfrog_devin_BM_comparison

A deeper look into the exact txns missing can be seen by comparing the following data dumps.  I spot-checked them and all were valid fee payments / DPTs ([here's one example](https://mempool.space/tx/fb1e59ac33b261d4bcbd7d803b64b4a7e4cc07d32cd05f6acf1a1d6ab2ba3954)).  Reviewers are also encouraged to do a similar checking which is quite easy to do if you are a burningman as the payments should be in your wallet history.  (blockheight to date then find the tx in your wallet by timestamp).  

[burningmandump_frog.txt](https://github.com/bisq-network/bisq/files/14983730/burningmandump_frog.txt)
[burningmandump_devin.txt](https://github.com/bisq-network/bisq/files/14983732/burningmandump_devin.txt)

visual_diff_frog_devin_Aug2023


ghost:
@alejandrogarcia83 -> [BurningManAccountingStore_v3_devin.zip](https://github.com/bisq-network/bisq/files/15171136/BurningManAccountingStore_v3_devin.zip)

PR #7081, opened 2024-04-23
Improve export account reputation process for Bisq 2

### LINKED ISSUE #7066: Improve export account
### Description

When importing account age in Bisq 2, users need to export their account from Bisq 1.
We should make it more explicit with adding the profile ID field and show the json content which will be copied.
Ref: https://github.com/bisq-network/bisq2/issues/1834#issuecomment-2002847706

ghost:
Fixes #7066
Show the imported profile ID if available.  If no valid data found, the user can paste it in the edit box.

image

image

image

image


ghost:
PR patch uploaded to #7066

PR #7102, opened 2024-05-10
Update seednode files

runbtc:
I ran into some issues trying to deploy a seednode. This resolves those issues.


runbtc:
Not stale

alejandrogarcia83 (approved):
utACK

PR #7109, opened 2024-05-21
Disable ImageCaptureReaderTest

HenrikJannsen:
If tests are run via gradle I get an error:
`Loading library prism_es2 from resource failed: java.lang.UnsatisfiedLinkError: Can't load library: /Users/dev/.openjfx/cache/16/libprism_es2.dylib` If run from the IDE it works.
CI on windows also fails, other OS work.
Seems that there are some platform specific dependency issues...

@devinbileck @alvasw Any ideas how to fix that?

### DIFF
--- a/desktop/src/test/java/bisq/desktop/main/account/content/notifications/qr/ImageCaptureReaderTest.java
+++ b/desktop/src/test/java/bisq/desktop/main/account/content/notifications/qr/ImageCaptureReaderTest.java
@@ -28,6 +28,7 @@ import java.util.function.Consumer;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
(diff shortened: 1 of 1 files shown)


devinbileck:
Perhaps you could try using a newer version of the JavaFX Gradle Plugin.

We are using [0.0.10](https://github.com/bisq-network/bisq/blob/master/gradle/libs.versions.toml#L44), but the latest is [0.1.0](https://github.com/openjfx/javafx-gradle-plugin/releases/tag/0.1.0)

It seems a [similar issue](https://github.com/openjfx/javafx-gradle-plugin/issues/112) was fixed in [v0.0.11](https://github.com/openjfx/javafx-gradle-plugin/releases/tag/0.0.11) when they added M1 Mac support.

alejandrogarcia83 (approved):
utACK

PR #7122, opened 2024-05-26
Don't show ARS blue market price warning at startup

HenrikJannsen:
Implements https://github.com/bisq-network/bisq/releases/


alejandrogarcia83 (approved):
utACK

PR #7123, opened 2024-05-26
Only show XMR subaddress recommendation if the user has a XMR account setup

HenrikJannsen:
We have shown the popup each time the user goes to account.
Now we only show if if the user has an XMR account.


alejandrogarcia83 (approved):
utACK

PR #7129, opened 2024-05-30
Remove emzy nodes

devinbileck:
Emzy is planning on shutting down most of his nodes soon, only leaving his mempool explorer (including bisq) and one btc node.


alejandrogarcia83 (approved):
utACK

Thank you!

cbeams (commented):
@devinbileck, 135.181.215.237 should be left in the list and 136.243.53.40 removed.

135.181.215.237 is the only one of @Emzy's four nodes that still has `-peerbloomfilters=1` configured. The other three are causing quite a bit of churn with trying and failing connections, especially when it's all happening on the currently stormy Tor network. It also looks, like @mrosseel's btc2.vante.me node suffers the same problems and should be removed. I'm working on a comprehensive list of nodes that still function as we need them to.

cbeams:
Note that I've asked @Emzy to reconfigure `-peerbloomfilters=1` for all his nodes until we can get v1.9.16 out with an updated list. Have not heard back yet about that (but communications were only just now sent out).

devinbileck:
@cbeams I assume you mean `peerbloomfilters=1`?

cbeams:
yes, fixed, thank you.

PR #7131, opened 2024-05-31
Update resource files

HenrikJannsen:

PR #7133, opened 2024-06-01
Update data stores for v1.9.16

alejandrogarcia83:
light_1
full_1
light_2
full_2
light_3
full_3


HenrikJannsen (approved):
utACK
Same hashes as I got from my full node posted at https://github.com/bisq-network/bisq/pull/7131

PR #7137, opened 2024-06-04
Remove nodes

devinbileck:
No need to merge this right away as I will keep them running for a while longer.


devinbileck:
Not stale

alejandrogarcia83 (dismissed):
utACK

alejandrogarcia83:
Hi @devinbileck!
I'm sorry for the delay. Could you please rebase this PR against master again? I tried to do it myself but it breaks the GPG signature. Thank you!

HenrikJannsen:
We should also remove persisted entries like done in `BisqSetup.maybeUpgradeBsqExplorerUrl` with other retired nodes.

alejandrogarcia83 (approved):
utACK

PR #7147, opened 2024-06-07
Update resource files

HenrikJannsen:
DaoStateStore of current resources has mismatch of chain height and last blocks height.

Here are the hashes:
 
Screenshot 2024-06-07 at 14 31 10
Screenshot 2024-06-07 at 14 31 14
Screenshot 2024-06-07 at 14 31 18


alejandrogarcia83 (approved):
utACK

PR #7149, opened 2024-06-07
Add new fields to filter

HenrikJannsen:
Adds 2 new fields to filter to allow to provide new btc nodes or seed nodes to the network.
The data are written to the properties files from which is will be read at next start.


alejandrogarcia83 (approved):
utACK

PR #7150, opened 2024-06-08
[1] Refactor Proto interface

HenrikJannsen:
Refactor: Replace toProtoMessage().toByteArray() calls to serialize() or serializeForHash() methods.

Currenty serializeForHash just calls serialize. This will be change in later commits


alejandrogarcia83 (approved):
utACK

PR #7151, opened 2024-06-08
[2] Add exclude for hash annotation

HenrikJannsen:
Based on https://github.com/bisq-network/bisq/pull/7150

PR #7152, opened 2024-06-08
[3] Apply exclude for hash annotation to filter

HenrikJannsen:
Based on https://github.com/bisq-network/bisq/pull/7151


alejandrogarcia83 (dismissed):
utACK

alejandrogarcia83 (approved):
utACK

PR #7166, opened 2024-06-23
Reduce bundled javacv library size

devinbileck:
With the introduction of the javacv library when re-adding the QR code scanner feature for mobile notification pairing in #7050, the Bisq binary increased to nearly 1 GB from its previous size of around 300 MB.

When including the javacv-platform dependency, it pulls in binaries for all platforms. But it is possible to limit this by utilizing the gradle-javacpp platform plugin which by default will pull in binaries only for the current platform.

Reference: https://github.com/bytedeco/gradle-javacpp#the-platform-plugin

With this change, the included libs went from ~980 MB down to ~366 MB.


alejandrogarcia83 (approved):
utACK

PR #7167, opened 2024-06-23
Enable ImageCaptureReaderTest

devinbileck:
This test was disabled in #7109 due to encountering dependency issues. Using a newer version of the JavaFX Gradle Plugin should hopefully resolve those issues since a [similar issue] was fixed in v0.0.11 when they added M1 Mac support.

[similar issue]: https://github.com/openjfx/javafx-gradle-plugin/issues/112


devinbileck:
It seems `javafx-graphics-16-mac-aarch64.jar` and `javafx-base-16-mac-aarch64.jar` are unavailable. We would likely need to update javafx at least to 17. That is outside the scope of this. Closing for now.

PR #7185, opened 2024-06-29
Add missing snapshot creation in full DAO mode

HenrikJannsen:
If running a full DAO node but not having Full mode for DAO Monitor activated we did not create snapshots while parsing blocks. 
This was bad when doing a resync as only after parsing complete which can take very long in case of a resync from genesis there have been no blocks and state persisted. Activating Full mode for DAO Monitor would slow down parsing as at each block the dao state hash would get created.

With this PR we create snapshots if full DAO mode is activated.

However when running a full DAO node it is recommended to set `--useFullModeDaoMonitor` to true to ensure the linked list of dao state hashes is correct and we do not depend on data provided by seed nodes.


djing-chan (approved):
ACK

alejandrogarcia83 (approved):
utACK

PR #7196, opened 2024-07-06
Set -fx-font-smoothing-type to grey.

HenrikJannsen:
With ` -fx-font-smoothing-type: gray;`:
lcdtext=false

With ` -fx-font-smoothing-type: lcd;`, which is default set by moderna.css:
def

Images are taken from Bisq2 where the same will be applied.


alejandrogarcia83 (approved):
utACK

PR #7203, opened 2024-07-15
Check burning man receiver address validity

stejbac:
Set the burn cap of a candidate to zero if he has an invalid receiver address, that is, one that bitcoinj cannot parse. This prevents trade protocol failure when creating the DPT, by making such BM inactive and distributing their shares to the other BM. (Setting the burn cap to zero is a little more robust than simply filtering out such candidates, as `BurningManService` handles subsequent share redistribution better than `DelayedPayoutTxReceiverService` or `BtcFeeReceiverService`.

While this case should normally never occur, due to UI validation of the custom receiver address, there are at least two ways a BM could invalidate his own receiver address if so inclined:

 1. He could simply bypass the UI validation;
 2. He could manually create a compensation issuance tx with a change address type unrecognised by bitcoinj, such as P2TR, as the address field is pulled straight from the RPC JSON by each full DAO node.

Thus, it is necessary to check both change and custom addresses.

--

This PR additionally does a couple of minor optimisations to `BurningManService`, fixes nondeterministic receiver address selection in the edge case of multiple accepted compensation proposals by the same candidate in the same cycle, and cleans up redundant date checks for the activation of fixes for https://github.com/bisq-network/bisq/issues/6699 and https://github.com/bisq-network/proposals/issues/412.

None of the changes should alter the candidate burn shares or receiver addresses, except in possible future edge cases where the addresses would be nondeterministic or invalid, resulting in DPT creation/validation errors anyway (and thus failure to open the trade). So an activation date for these changes shouldn't be necessary.

### DIFF
--- a/core/src/main/java/bisq/core/dao/burningman/BtcFeeReceiverService.java
+++ b/core/src/main/java/bisq/core/dao/burningman/BtcFeeReceiverService.java
@@ -27,7 +27,6 @@ import javax.inject.Singleton;
 
 import com.google.common.annotations.VisibleForTesting;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
 import java.util.stream.Collectors;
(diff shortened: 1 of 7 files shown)


HenrikJannsen (dismissed):
utACK

stejbac:
There is another small change to `BtcAddressValidator` I wanted to push:

I noticed that bitcoinj allows any witness v1-v16 program (up to length 40 bytes) to be decoded from a Bech32 address, not just witness v0 (that is, p2wpkh & p2wsh). These give standard txs when used as outputs (and it doesn't look like bitcoinj would fail to validate or broadcast such txs either), though all of them apart from p2tr ScriptPubKeys would be anyone-can-spend and couldn't be spent in standard txs.

However, since Taproot was introduced, witness v1 programs and above are supposed to be encoded as Bech32m, not Bech32, so really any Segwit address that isn't witness v0 shouldn't validate unless/until our fork of bitcoinj parses them correctly as Bech32m (which is already done in the upstream version). Otherwise, there could be complications making such an upgrade.

HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #7206, opened 2024-07-21
Reset AutoConfirmSettings if it contains a retired address

HenrikJannsen:
If AutoConfirmSettings contain a retired address we reset to the default values


alejandrogarcia83 (approved):
utACK

PR #7221, opened 2024-07-28
Optimise/simplify some stream filtering for the REST API

stejbac:
Replace the streaming of `Map` entry sets to pick out a single entry by key equality, and instead do a lookup into the map. Also, optimise the date range filtering in `TradeStatisticsManager::getTradeStatisticsList` by using `RangeUtils::subSet` to avoid scanning the entire collection. (This method is applicable, as the trade statistics set is navigable and naturally sorted by date.)


HenrikJannsen (approved):
utACK

PR #7248, opened 2024-09-17
Update seednode burningman oracle keys

devinbileck:
Remove keys for removed seednodes and add keys for newer seednodes.


HenrikJannsen (approved):
utACK

PR #7251, opened 2024-09-19
Sort foreign alphabet payment methods above the untranslated ones.

cparke2:
By default, JAVA sorts foreign characters below English characters. However, the payment methods that are translated are also more likely to be used by someone native to that language, so it makes better sense to sort as two separate lists and put the foreign language ones on top. This commit does just that for the payment methods list on the Accounts page.


cparke2:
Application was built and run to test and verify the sorting changed as expected, and nothing broken.

See attached screenshot from Russian language mode with the change is applied, with an line drawn hand drawn to illustrate where the Russian ends and the English begins.

HenrikJannsen (approved):
utACK

PR #7252, opened 2024-09-19
New option in PhoneNumberValidator to enforce exact number of digits

cparke2:
Until now, national phone numbers worldwide just had to be between 5-12 digits long to pass validation. However, many countries have a specific number of of digits in all their phone numbers which is very well known (EX: all USA numbers are 10 digits long). So this change adds an optional parameter to specify how many digits must be in the phone number for it to validate. Uncommited changes will also use this new capability, provided this pull request is adopted.

### DIFF
--- a/core/src/main/resources/i18n/displayStrings_cs.properties
+++ b/core/src/main/resources/i18n/displayStrings_cs.properties
@@ -3519,6 +3519,7 @@ validation.phone.missingCountryCode=K ověření telefonního čísla je potřeb
 validation.phone.invalidCharacters=Telefonní číslo {0} obsahuje neplatné znaky
 validation.phone.insufficientDigits=V čísle {0} není dostatek číslic, aby mohlo být platné telefonní číslo
 validation.phone.tooManyDigits=V čísle {0} je příliš mnoho číslic, než aby mohlo být platné telefonní číslo
+validation.phone.incorrectLength=The field must contain {0} numbers
 validation.phone.invalidDialingCode=Telefonní předvolba země pro číslo {0} je pro zemi {1} neplatná. Správné předčíslí je {2}.
 validation.invalidAddressList=Seznam platných adres musí být oddělený čárkami
 validation.capitual.invalidFormat=Musíte zadat CAP kód ve formátu: CAP-XXXXXX (6 alfanumerických znaků)
--- a/desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
+++ b/desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
@@ -106,7 +106,11 @@ public class PhoneNumberValidator extends InputValidator {
                 normalizedPhoneNumber = "+" + getCallingCode() + pureNumber;
             }
 
+            //  If the 'requiredLength' was set, there's still one more check to apply on the normalizedNumber itself
             result = validateRequiredLength(normalizedPhoneNumber, normalizedCallingCode);
+            if (!result.isValid) {
+                normalizedPhoneNumber = null;
+            }
         }
         return result;
     }
--- a/desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
+++ b/desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
@@ -7,6 +7,8 @@ import lombok.Getter;
 
 import javax.annotation.Nullable;
 
+import java.util.Optional;
+
 /**
  * Performs lenient validation of international phone numbers, and transforms given
  * input numbers into E.164 international form.  The E.164 normalized phone number
@@ -44,7 +46,7 @@ public class PhoneNumberValidator extends InputValidator {
      * Required length of phone number (excluding country code).
      */
     @Getter
-    private int requiredLength;
+    private Optional<Integer> requiredLength;
 
     ///////////////////////////////////////////////////////////////////////////////////////////
     // Constructors
@@ -59,11 +61,7 @@ public class PhoneNumberValidator extends InputValidator {
         this.isoCountryCode = isoCountryCode;
         this.callingCode = CountryCallingCodes.getCallingCode(isoCountryCode);
         this.normalizedCallingCode = CountryCallingCodes.getNormalizedCallingCode(isoCountryCode);
-    }
-
-    public PhoneNumberValidator(String isoCountryCode, int requiredLength) {
-        this(isoCountryCode);
-        this.requiredLength = requiredLength;
+        this.requiredLength = Optional.ofNullable(PhoneNumberRequiredLengths.getRequiredLength(isoCountryCode));
     }
 
     ///////////////////////////////////////////////////////////////////////////////////////////
@@ -107,9 +105,11 @@ public class PhoneNumberValidator extends InputValidator {
             }
 
             //  If the 'requiredLength' was set, there's still one more check to apply on the normalizedNumber itself
-            result = validateRequiredLength(normalizedPhoneNumber, normalizedCallingCode);
-            if (!result.isValid) {
-                normalizedPhoneNumber = null;
+            if (requiredLength.isPresent()) {
+                result = validateRequiredLength(normalizedPhoneNumber, normalizedCallingCode);
+                if (!result.isValid) {
+                   normalizedPhoneNumber = null;
+                }
             }
         }
         return result;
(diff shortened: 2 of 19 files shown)


desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
```diff
@@ -161,4 +173,14 @@ private ValidationResult validateIncludedCountryDialingCode(String rawInput,
             return new ValidationResult(false, Res.get("validation.invalidInput", t.getMessage()));
         }
     }
+
+    private ValidationResult validateRequiredLength(String normalizedPhoneNumber, String normalizedCallingCode) {
+        try {
+            return ((0 == requiredLength) || ((normalizedPhoneNumber.length() - normalizedCallingCode.length() - 1) == requiredLength))
+                    ? new ValidationResult(true)
+                    : new ValidationResult(false, Res.get("validation.phone.incorrectLength", requiredLength));
```
cparke2, line 181: This is a new resource string currently missing from this pull request.  Do not merge until it is defined in every language!

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
```diff
@@ -49,13 +54,18 @@ public class PhoneNumberValidator extends InputValidator {
...
-    
+
     public PhoneNumberValidator(String isoCountryCode) {
         this.isoCountryCode = isoCountryCode;
         this.callingCode = CountryCallingCodes.getCallingCode(isoCountryCode);
         this.normalizedCallingCode = CountryCallingCodes.getNormalizedCallingCode(isoCountryCode);
     }
 
+    public PhoneNumberValidator(String isoCountryCode, int requiredLength) {
+        this(isoCountryCode);
```
cparke2: Another approach to calling this alternative constructor with a 'requiredLength' argument is to modify the **CountryCallingCodes.java** Map itself to incorporate **requiredLength** into each country's definition there along with the country code (it would become a Map of Maps or Pairs). Or make a separate Map file, as that is repurposing the additional map with additional information it was not originally intended for.  As here, if the requiredLength is not known (or missing if we setup an alternative Map), setting it to 0 will skip the requiredLength validation. If you want to go with that approach instead, I'm only setting the required length for US, Canada, and Russia that I'm sure about...

core/src/main/resources/i18n/displayStrings_cs.properties
```diff
@@ -3519,6 +3519,7 @@ validation.phone.missingCountryCode=K ověření telefonního čísla je potřeb
 validation.phone.invalidCharacters=Telefonní číslo {0} obsahuje neplatné znaky
 validation.phone.insufficientDigits=V čísle {0} není dostatek číslic, aby mohlo být platné telefonní číslo
 validation.phone.tooManyDigits=V čísle {0} je příliš mnoho číslic, než aby mohlo být platné telefonní číslo
+validation.phone.incorrectLength=The field must contain {0} numbers
```
HenrikJannsen, line 3522: No need to add i18n strings to the non-default file. If not present it falls back to the default anyway

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
```diff
@@ -40,6 +40,11 @@ public class PhoneNumberValidator extends InputValidator {
     @Nullable
     @Getter
     private String normalizedPhoneNumber;
+    /**
+     * Required length of phone number (excluding country code).
+     */
+    @Getter
+    private int requiredLength;
```
HenrikJannsen: As it is optional I would suggest to use an Optional<Integer> type instead.

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
```diff
@@ -95,6 +105,8 @@ public ValidationResult validate(String input) {
             } else {
                 normalizedPhoneNumber = "+" + getCallingCode() + pureNumber;
             }
+
+            result = validateRequiredLength(normalizedPhoneNumber, normalizedCallingCode);
```
HenrikJannsen: I think the check if the requiredLength is 0 (or better ifPresent when optional is used) should be done here, as we only validate if that is the case.

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
HenrikJannsen: Is this method used already in the app? I have not seen it in the PR. Or did you plan to apply it in another PR?

I would consider a map instead of a custom constructor better. I think a single map of country code to requiredLength should be enough, and if no entry present we ignore the check.

HenrikJannsen (changes requested):
See comments

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
cparke2: I have a pull request coming soon for you to add a new Bisq payment method that will use this feature, but as these are actually independent things, this pull request is a **_prerequisite_** as I need to know how you want this implemented in the app before I submit an actual usage of it. It's fine with me if you approve but don't merge this pull request until that final pull request which uses these changes is opened.

As for **displayString**, every other string among the **2,669** strings that (already) exist is in **_every_** language file, so it would be inconsistent to not add to all language files here.  Also, commented instructions at the top of the **displayStrings.properties** indicate that new strings should be added to all language files:
`# Please use in all language files the exact same order of the entries, that way a comparison is easier.
`
So on this matter, I'm going to ask you to please reconsider that decision.

However, accordingly, I will revise this pull request to add a new source module with a **country-code -> requiredLength** map as you suggest, as well as use the 'optional' flag since not all countries may have a consistent phone number length (or we don't know what it is).

HenrikJannsen: "As for displayString, every other string among the 2,669 strings that (already) exist is in every language file, so it would be inconsistent to not add to all language files here." - the translations would be handled via transifex. Though I am not sure how well it is maintained these days. If you like you can add all translations (using some AI translation tool). But to just add the english is redundant and give the wrong impression as it was translated (I think in transifex they see missing entries, but not if its in english).

HenrikJannsen (approved):
utACK

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
cparke2: > the translations would be handled via transifex. Though I am not sure how well it is maintained these days. 
> If you like you can add all translations (using some AI translation tool). 
> But to just add the english is redundant and give the wrong impression as it was translated (I think in transifex they see missing entries, but not if its in english).

This is off-topic, but a rough analysis of the language files which use foreign characters reveals the current level of translation in them is pretty low and there is already a ton of these redundant English strings in them:
```
displayStrings_cs.properties: translatedcount=1542, percent=57.77%
displayStrings_fa.properties: translatedcount=1556, percent=58.30%
displayStrings_ja.properties: translatedcount=2050, percent=76.81%
displayStrings_pl.properties: translatedcount=1364, percent=51.11%
displayStrings_ru.properties: translatedcount=1585, percent=59.23%
displayStrings_th.properties: translatedcount=1438, percent=53.88%
displayStrings_vi.properties: translatedcount=1445, percent=54.14%
displayStrings_zh-hans.properties: translatedcount=2016, percent=75.53%
displayStrings_zh-hant.properties: translatedcount=1991, percent=74.60%

totalcount=2669, overall average 62.39% translated
```
The above illustrates that there may be a follow-up issue here to improve the level of translation here in the 15 foreign languages currently supported.  Bisq is a a truly global project!

The above also shows that redundant English strings in the foreign language files is **_very_** prevalent, and it certainly appears to have been a common practice, and it has been done according to the instructions. I'm just continuing the status quo by adding more. As a developer, I really feel uncomfortable using a tool like Google Translate to translate my new string into 15 other languages when I have no ability to read and verify those translations. Like others before me, I'd prefer to just put the English in there as a placeholder until somebody, some day, with competency in that language undertakes the task that you mentioned of using a translation tool on the language file (and having the English strings in them already makes it much easier for such a person to use a translation tool like "transifex" that you mentioned to do it).

HenrikJannsen: Yes, I just told you that we use the translation tool Transifex and as far I am aware best practice is to only add the english string (default file) and let the translation be done by translators. The alternative path is to do it yourself with AI tools but of course as you said its not bulletproof (though also human translations had many mistakes). 
I am not familiar with the details of the workflow but I assume adding all languages in english adds the risk that the translation is missed. @ripcurlx might have more input here.

HenrikJannsen (approved):
utACK

desktop/src/main/java/bisq/desktop/util/validation/PhoneNumberValidator.java
HenrikJannsen: We are aware that the status quo is unfortunately that there are many translations missing. We lack a maintainer atm who organize and fix that. Anyone who wants to help here is welcome!

HenrikJannsen: As I saw now that we use other english entries, its probably best to stick with that...

PR #7255, opened 2024-09-22
Create payment method "Faster Payments System (SBP)" for Russian Ruble

cparke2:
This is the standard P2P payment method in Russia to perform funds transfers and payments between Russian bank accounts in Russian Rubles. There is no chargeback risk. Recipient bank account is located using telephone number and bank name, and sender receives recipients first name, middle name, and initial of last name to confirm the phone number entered was correct before sending. 

Adding this new payment method has been discussed at length on the GitHub 'growth' channel at:
   https://github.com/bisq-network/growth/issues/288


cparke2:
New payment method utilizes next available account **_payload ID of 41_** in 'protobuf'.

Tested and verified using an end-to-end trade through bitcoind -regtest and the Bob & Alice Bisq test Bisq users accounts. 

Testing also included Add/Delete/Edit account and Import/Export of accounts in Bisq GUI.

cparke2:
**Screenshots of the new payment method in the Bisq GUI, as test evidence:**


Screenshot from 2024-09-19 02-11-51
Screenshot from 2024-09-22 16-36-31
Screenshot from 2024-09-22 16-35-45
Screenshot from 2024-09-22 16-34-34
Screenshot from 2024-09-22 16-47-38
Screenshot from 2024-09-22 16-50-13
Screenshot from 2024-09-22 16-56-56

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -4295,6 +4297,14 @@ payment.amazonGiftCard.info=To pay with Amazon eGift Card, you will need to send
   - Amazon eGift Cards can only be redeemed on the Amazon website they were purchased on (e.g., a gift card purchased on amazon.it can only be redeemed on amazon.it)
 payment.mercadoPago.holderId=UserID linked to financial institution. Like phone number or email or CVU.
 
+payment.sbp.info.account=The Faster Payments System (SBP) is an inter-bank funds transfer service in RUSSIA that allows individuals \
```
HenrikJannsen, line 4300: Why use all cap for Russia?

HenrikJannsen (approved):
utACK

core/src/main/resources/i18n/displayStrings.properties
HenrikJannsen: I will merge it as it is, if you agree to use "Russia" instead just add it in a new PR...

cparke2: One big concern in the discussion thread was that the name of this payment method (in English) is almost identical to the existing "Faster Payments" payment method. Added "(SBP)" to help it stand out in the list for people looking for it as SBP.  Still, to emphasize to anyone clicking it intending for the other "Faster Payments" method, I used all caps so they'd understand quickly this is not what they're looking for.  

I certainly can remove the emphasis of RUSSIA if it seems like overkill. Or, add some other text in the pop-up to clearly and unambiguously differentiate the two payment methods (perhaps on both payment methods?). Really wanted to get some input or response from @pazza83 in the discussion thread on this, but he's mostly avoided my pings to give a comment so I was left on my own.

I am working on adding Russian translations, so certainly could change this however the team feels appropriate.

HenrikJannsen: Ah ok. Yes I agree that both methods using a similar name might be confusing. I would though not do it by all caps for Russia, but rather to change the name of the payment method to maybe: SBP-Faster Payments System or Faster Payments System Russia (SBP)

cparke2: Another thought is to change the display name of the existing "Faster Payments" to "Faster Payments (UK)".  I don't think that is a compatibility issue, as the Bisq1 accounts identify their associated payment method with an enumeration value, and once created, the display name is always a "custom name" so it won't change for anyone who already has it.

As for the new payment method from Russia, I was trying to avoid too many acronyms in the name, but "SBP" is definitely necessary in the English version or people might not realize that is what it is.  English transliteration of the name in Russian is very cheesy too.  Maybe to do that and still be consistent: "Faster Payments System-SBP (Russia)"?

There also is a strange set of payment methods prefixed with "India", so that is another approach that also could be used to differentiate the two systems.

I have no problem removing the all caps, that's fine.  The existing Faster Payments has no informational pop-up message at all, and further as both payment methods are locked to their respective currencies, so someone choosing the wrong one isn't going to get far...

HenrikJannsen: Yes the display string should not have any impact on anything. So Faster Payments (UK) is ok as well.
"Faster Payments System-SBP (Russia)" - sounds ok to me.

> ...prefixed...

Yes I think that helps ppl to find quickly their relevant method.

> ...no informational pop-up...

Yes we use popups only if needed. Too many anyway....

cparke2: Ok, so it's actually going to be (in English):

"Faster Payment System (UK)"
https://www.wearepay.uk/what-we-do/payment-systems/faster-payment-system/

"Faster Payments System-SBP (Russia)"
https://www.cbr.ru/eng/psystem/sfp/

I'll keep the SBP pop-up but remove the all caps RUSSIA since the naming convention already makes that pretty clear.

Separately, I am also adding some Russian translations currently missing.

The new pull requests should be ready in about a week. Any idea when the next release is scheduled?

HenrikJannsen: I don't know but I guess it will take a bit longer for the release. In doubt you can make smaller PRs so that we get in at least parts.

pazza83: Faster Payments System-SBP (Russia) works for me also

pazza83: Wiki has been added https://bisq.wiki/Faster_Payments_System_SBP

PR #7257, opened 2024-09-23
"Transfer Same Bank" & "Transfer Specific Banks" match case-insentitivity

cparke2:
Offers match these payment method types only if bank names entered in the maker and taker accounts are the same. However, if the maker entered, for example, "BANK OF AMERICA" as bank name in their Bisq account but the taker entered "Bank of America", the offer account mismatches without this fix to make the match case-insensitive.


HenrikJannsen (approved):
utACK

PR #7258, opened 2024-09-24
Improve exception msg and handling when 'Same Bank' BIC/SWIFT blank

cparke2:
Users in most countries (see **BankUtil.useValidation()** for the list of exceptions) can create a "Transfer Same Bank" payment account in Bisq with no BIC/SWIFT code. However, to create an offer using this account type, in most countries (see **BankUtil.isBankIdRequired()** for these exceptions) the account must have a value in the **bankId** field, or JAVA will throw a **_'null pointer exception'_**, and then the Bisq application will then become unhealthy and hang trying on exit. This fix adds a validation check for that condition and throws a more "friendly" exception which explains the problem, and it happens earlier to keep the app. healthy.

### DIFF
--- a/core/src/main/java/bisq/core/offer/OfferUtil.java
+++ b/core/src/main/java/bisq/core/offer/OfferUtil.java
@@ -30,6 +30,8 @@ import bisq.core.offer.bisq_v1.OfferPayload;
 import bisq.core.payment.CashByMailAccount;
 import bisq.core.payment.F2FAccount;
 import bisq.core.payment.PaymentAccount;
+import bisq.core.payment.SameBankAccount;
+import bisq.core.payment.SpecificBanksAccount;
 import bisq.core.payment.payload.PaymentMethod;
 import bisq.core.provider.fee.FeeService;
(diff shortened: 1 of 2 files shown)


cparke2:
**BEFORE:**
Screenshot from 2024-09-24 05-55-00

**AFTER:**
Screenshot from 2024-09-24 05-56-17

cparke2:
Another thing I wanted to do, but didn't want to "step on someone's toes", was just make the BIC/SWIFT required on the account form for creating the "Transfer Same Bank" payment method. Or, make it use **Bank Name** rather than BIC for bank matching ("Transfer Specific Banks" works differently already and uses Bank Name for all currencies to match offers).  So this pull request represents the minimum: just handle the invalid case more gracefully; hopefully that'll be enough info for the user to fix their problem.  

I also opened a discussion on these related account types which raises other issues with this payment method and it's sibling:
[ https://github.com/bisq-network/bisq/discussions/7256 ](https://github.com/bisq-network/bisq/discussions/7256)

HenrikJannsen (approved):
utACK

PR #7265, opened 2024-10-03
Create payment method "Faster Payments System (SBP)" for Russian Ruble

cparke2:
Removed all caps of "RUSSIA" in new account pop-up and renamed existing payment method "Faster Payments" which is a similarly named payment method to "Faster Payment System (UK)", as required follow-up to to pull request https://github.com/bisq-network/bisq/pull/7255 and discussed in issue https://github.com/bisq-network/growth/issues/288


HenrikJannsen (approved):
utACK

PR #7253, opened 2024-09-22
Language strings file cleanup; added missing, applied rename

cparke2:
The Bisq policy for displayStrings, as stated in the comments at the beginning of the primary (English) language properties file, is that all string keys should be present in **_all_** language files, even though fallthrough would otherwise result. In reviewing how well the language files actually comply with this rule, the following discrepancies were found and are fixed here:

1. **tradeFeedbackWindow.msg.part1/2/3** apparently was changed at some point in the past to **tradeFeedbackWindow.msg.feedback** and **tradeFeedbackWindow.msg.thanks**.  This change also involved removal of reference to a user survey in favor of redirecting users to a Bisq forums.  However, that change was never applied to any of the language files, which meant fallthrough has been happening and the translations were **_not_** displaying.

2. **filterWindow.addedSeedNodes** and **filterWindow.addedBtcNodes** were never added to language files. These and the other filterWindow strings are used by **desktop/src/main/java/bisq/desktop/main/overlays/windows/FilterWindow.java** file, although that window itself seems to be dead code at present (I cannot find it in the application or how it would be launched by the code). This is may be a separate issue for a different pull request at a later time, depending on whether FilterWindow is deemed obsolete or if it is actually a new feature still in development.

Prerequsite to PR https://github.com/bisq-network/bisq/pull/7268

PR #7268, opened 2024-10-11
Added 500+ string translations to Russian language file

cparke2:
Previously, about 1,585 out of 2,676 strings (59.23%) were translated into Russian, the rest remained in English. This meant a native Russian speaker with limited English ability might have great difficulty using Bisq application and might not understand or notice the way the system works and the many important pop-ups that can appear. Now there are about 2,103 strings in Russian (78.59%) and most of the messages the end user normally would see are displayed in Russian, making the application very usable.

Testing was done in Russian language mode using both connection to the actual Bisq P2P network with live data, as well as using Alice, Bob, & Mediator to walk through the trade process (with and without mediation and arbitration), in order to find which strings are important but missing a translation, as well as to verify that strings fit as well as possible in the space provided. In some cases, existing translations were changed (ex: "Fiat" was previously translated into "Natl. Currency", but fiat is also a borrowed word in Russian and was changed to transliteration, and also "Offer ID" had been translated "Identificator Offer", but that way too long for the space available so "ID Offer" in Russian is good enough).

Preceded by PR https://github.com/bisq-network/bisq/pull/7253


core/src/main/resources/i18n/displayStrings_ru.properties
```diff
@@ -669,16 +669,16 @@ portfolio.pending.autoConf.state.ERROR=An error at a service request occurred. N
...
-portfolio.pending.step2.confReached=Your trade has reached at least one blockchain confirmation.\n\n
+portfolio.pending.step2.confReached=Ваша сделка получила как минимум одно подтверждение в блокчейне.\n\n
 
-portfolio.pending.step2_buyer.refTextWarn=Important: when making the payment, leave the \"reason for payment\" field empty. DO NOT put the trade ID or any other text like 'bitcoin', 'BTC', or 'Bisq'. If you are required to fill in a reason, use your account name eg "Joe Bloggs", alternatively enter something non-descriptive like a dash (-). You are free to discuss via trader chat if an alternate \"reason for payment\" would be suitable to you both.
+portfolio.pending.step2_buyer.refTextWarn=Важно: при совершении платежа оставьте поле \«причина платежа\» пустым. НЕ вводите идентификатор сделки или любой другой текст, например ''биткоин'', ''bitcoin'', ''BTC'' или ''Bisq''. Если вам необходимо указать причину, используйте имя своей учетной записи, например, \«Джон Блоггс\», или введите что-то неописательное, например, тире (-). Вы можете свободно обсудить в чате трейдеров, подойдет ли вам обоим альтернативная \«причина платежа\».
```
HenrikJannsen: As here is not variable used (e.g. {0}) the duplicate single quote is not needed and would get displayed as duplicate quote. Only if a variable and a single quote occurs in a string we need to duplicate it.

HenrikJannsen (changes requested):

core/src/main/resources/i18n/displayStrings_ru.properties
cparke2: Good catch, I fixed that one plus all the other incorrect doubling up's (or missing doubling up's) in the default (English) and other language files.  However, French and Italian have a large number of incorrect apostrophes because of stress marks used in the language itself (see comment below for an example), so fixing those will need to be a separate pull request, as it's far outside the scope of this pull request to add Russian translations.

core/src/main/resources/i18n/displayStrings.properties
```diff
@@ -1314,9 +1314,9 @@ Traders can communicate within the application via secure chat on the open trade
 support.initialInfo=Please enter a description of your problem in the text field below. \
   Add as much information as possible to speed up dispute resolution time.\n\n\
   Here is a check list for information you should provide:\n\
-  \t● If you are the BTC buyer: Did you make the Fiat or Altcoin transfer? If so, did you click the 'payment started' \
+  \t● If you are the BTC buyer: Did you make the Fiat or Altcoin transfer? If so, did you click the ''payment started'' \
```
HenrikJannsen, line 1317: Here is no variable used so the single quote is correct. Otherwise it would be displayed as double quote.

HenrikJannsen (changes requested):

core/src/main/resources/i18n/displayStrings.properties
cparke2: It's a big, multi-line string.  The variable is 10 lines down on line 1327 in this case.  All these apostrophe fixes were identified using a script to analyze the properties files.

core/src/main/resources/i18n/displayStrings_fr.properties
```diff
@@ -3477,7 +3477,7 @@ validation.sortCodeChars={0} doit être composer de {1} caractères.
 validation.bankIdNumber={0} doit être composer de {1} chiffres.
 validation.accountNr=Le numéro du compte doit comporter {0} chiffres.
 validation.accountNrChars=Le numéro du compte doit comporter {0} caractères.
-validation.btc.invalidAddress=L''adresse n''est pas correcte. Veuillez vérifier le format de l''adresse.
+validation.btc.invalidAddress=L'adresse n'est pas correcte. Veuillez vérifier le format de l'adresse.
```
cparke2, line 3480: This is an example of the problems remaining in just French and Italian, as using an apostrophe for accent/stress is pervasive in these languages, and doubling-up for the Java properties file requirement is sometimes done incorrectly.

cparke2: I added a commit to fix a situation that I found where substitution parameters are missing (or were removed at some time) but the code was still trying to to do the replacements. That is another way these inconsistencies can happen (though certainly not the only way), and really the correct way to fix such these cases is to modify the code to not pass the (unused) replacements as I have done in this case (no guarantees that there aren't others). Really though, an even better solution for everyone (which should have been done at day-1) would be enhance Res.get() to perform the apostrophe double-up itself when replacement parameters are provided and make the language files consistent.

core/src/main/resources/i18n/displayStrings.properties
cparke2: For the string `support.initialInfo` you specifically questioned here:

**BEFORE:**
Screenshot from 2024-10-11 23-30-26(1)

**AFTER:**
Screenshot from 2024-10-12 01-20-25

This was an unusual one because it originally had both single apostrophe's and double apostrophe's in the same string! (and I had not changed anything about it in my original commit on this branch)

I think we're good.

HenrikJannsen (approved):
utACK

PR #7269, opened 2024-10-12
Added 500+ string translations to Russian language file

cparke2:
Removed **_non-printable_** characters that were discovered during a words analysis of the Russian translations file:
   ... (например, «С днем **\u200b\u200b**рождения, Сьюзен!») ...
While apparently harmless and invisible, these do not belong and a needle in the haystack later, so removing them now.


HenrikJannsen (approved):
utACK

PR #7277, opened 2024-11-05
Implement BitcoinCoreSeedListDiscovery method

alvasw:
- Update bitcoin core seed list to upstream version
- Use IpV4 and onion addresses from it

Our users are still protected because we connect to IpV4 addresses
through Tor. This improves connectivity because Bitcoin Core creates an
ephemeral onion service on each startup.


alejandrogarcia83 (approved):
utACK

PR #7280, opened 2024-11-05
Socks5MultiDiscovery: Don't create useless array

alvasw:
We create an array in the getPeers method and throw it away without
using it.


alejandrogarcia83 (dismissed):
utACK

alejandrogarcia83 (approved):
utACK

PR #7283, opened 2024-11-07
core: Implement transaction offset signing

alvasw:
The BisqTransactionSigner signs the inputs of a transaction after the
given offset using the LocalOffsetTransactionSigner. The
LocalOffsetTransactionSigner is identical to BitcoinJ's
LocalTransactionSigner. The only difference is that the
LocalOffsetTransactionSigner accepts an offset in its constructor. All
inputs below this offset are skipped during signing.

This is needed for Bisq's BSQ implementation because the mining fees of
BSQ transactions come from Bisq's BTC wallet. BitcoinJ's
LocalTransactionSigner expects all inputs to be part of the same wallet.


alejandrogarcia83 (approved):
utACK

PR #7285, opened 2024-11-07
BsqWalletV2: Implement send BSQ

alvasw:
The sendBsq method creates a BSQ transaction, signs the BSQ inputs, and
adds the requires BTC mining fees to the transaction.


alejandrogarcia83 (dismissed):
utACK

alejandrogarcia83 (approved):
utACK

PR #7289, opened 2024-11-10
gradle: Create bisq.java-integration-tests plugin

alvasw:
The bisq.java-integration-tests plugin creates an integrationTest source
set that depends on JUnit. The source set inherits the main source set's
compile classpath, runtime classpath, and the implementation's source
set configuration and runtimeOnly configuration.


alejandrogarcia83 (approved):
utACK

PR #7294, opened 2024-11-11
sendBsq: Throw InsufficientMoneyException when BSQ balance too low

alvasw:
We should throw an InsufficientMoneyException when the CoinSelection
gathered amount is lower than the amount the user wants to send.


alejandrogarcia83:
Hi @alvasw,
please resolve the merge conflict. Thank you!

alejandrogarcia83 (approved):
utACK

PR #7297, opened 2024-11-14
BtcNetworkConfig: Shuffle Bitcoin Core peers

alvasw:
We try to connect to the first 7 Bitcoin Core nodes always in the same
order. Only if connections to these nodes fail we look further into the
list. This change shuffles the node addresses before passing them to
BitcoinJ thus removing the bias from the first 7 prioritized nodes.


alejandrogarcia83 (approved):
utACK

PR #7304, opened 2024-11-18
Fix duplicates in BtcNode list

alvasw:
When the user uses our federated BTC nodes, we merge the hard-coded
nodes with the ones provided by the filter. The hard-coded node's
operator field is set to the node's operator and operator field of the
nodes from the filter is set to "Provided by filter". When the same BTC
node is in the hard-coded list and the filter, Bisq adds both to the
merged list because the operator field is different.

This change explicitly marks the onionAddress, hostName, address, and
port field to be used in the hashCode and equals implementation.


alejandrogarcia83 (dismissed):
utACK

alejandrogarcia83 (approved):
utACK

PR #7309, opened 2024-11-18
BtcNodes: Fix hostname parsing

alvasw:
The fromFullAddress method parsed hostnames as IPV4 addresses instead as
hostnames.


alejandrogarcia83 (approved):
utACK

PR #7310, opened 2024-11-18
NodeAddress: Fix broken IPV6 parsing

alvasw:
Bisq accepts IPV6 node addresses but the NodeAddress can't parse them.


alejandrogarcia83 (approved):
utACK

PR #7311, opened 2024-11-18
BtcNodeProvider: Fix BTC node banning

alvasw:
The FederatedBtcNodeProvider compared the list of banned nodes with each
BtcNode's hostname instead of checking a BtcNode's hostname, ip address,
and onion address.


alejandrogarcia83 (approved):
utACK

PR #7319, opened 2024-12-09
BitcoinjRegtestSetup: Implement wallet funding

alvasw:
The fundWallet method sends the given amount to the BitcoinJ wallet from
the Bitcoin Core miner wallet, mines one block, and waits until the
BitcoinJ wallet updates its wallet balance.


alejandrogarcia83 (approved):
utACK

PR #7322, opened 2024-12-09
RegtestWalletAppKit: Implement createNewBsqWallet method

alvasw:
The createNewBsqWallet method creates a new BitcoinJ wallet and
registers it with BitcoinJ's BlockChain and PeerGroup.


alejandrogarcia83 (approved):
utACK

PR #7323, opened 2024-12-09
BitcoinjBsqTests: Create per test BSQ wallet

alvasw:
At the moment, some tests are sharing the same BSQ wallets. This change
creates a separate wallet for indenpendent tests to improve the tests's
reliability.


alejandrogarcia83 (approved):
utACK

PR #7327, opened 2024-12-19
Add instruction to select JDK version at build time

suddenwhipvapor:
When multiple JDK versions are installed in Linux, Gradle will usually select the last one. When this is JDK22, for example used to build Bisq2. this will not yield a correct build for Bisq1.
Hence, I provided steps to fix this.


alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #7329, opened 2024-12-21
getoffers (api): Return all offers by default

alvasw:
The getoffers api doesn't return any offers until the user has created
his payment accounts. Afterward the api will only return offers that the
user can take. This behavior is confusing and frustrating for new
developers. The new behavior is to show all offers for a given market
(direction, currency) by default.

Changes:
- [api: Remove getoffers' --all option](https://github.com/bisq-network/bisq/commit/462144eb8f835ea70364f4622b49fa134f5a75ec)
- [getoffers: Include offers that can't be taken](https://github.com/bisq-network/bisq/commit/c3fbc859f137227b1a134571db2987df40abb582)
- [api: Include my offers in getoffers](https://github.com/bisq-network/bisq/commit/0ba23b5a502ae4e0fd593772ffa63d78312078ca)


alejandrogarcia83 (approved):
utACK

PR #7332, opened 2024-12-21
Add link to trading rules wiki article in Tac Window

suddenwhipvapor:
The trading rules, as explained on https://bisq.wiki/Trading_rules have a tidier layout, and allow for future adjustments in case they are needed, so it is useful to have them linked from the Tac.

I originally intended to make the word "rules" in bullet 6 into a link, but when I understood how the overlay contents are built, I rapidly changed my mind, so I thought about adding another sub-bullet above the dispute resolution link, by creating another HBox and have the two HBox's placed into a VBox, but that in turn proved unfeasible as the relative height of the hyperlinks to the preceding text were inconsistent no matter what I did.
So I finally settled for placing the two in the same HBox, and at the same time fixed the line-spacing with the above bullets, that was larger than the rest in the original version. I found no way of fixing the vertical alignment of the hyperlinks with the text, which was below the base level even in the original version.

Original:
old

My PR:
new

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -5,8 +5,8 @@
    ```sh
    git clone https://github.com/bisq-network/bisq
    # if you intend to do testing on the latest release, you can clone the respective branch selectively, without downloading the whole repository
-   # for the 1.9.18 release, you would do it like this:
-   git clone --recurse-submodules --branch release/v1.9.18 https://github.com/bisq-network/bisq
+   # for the 1.9.3 release, you would do it like this:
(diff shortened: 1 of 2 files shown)


docs/build.md
alejandrogarcia83, line 1: I think you accidentally committed this file to your PR. You are reverting your previous PR (https://github.com/bisq-network/bisq/pull/7327).

alejandrogarcia83 (changes requested):

HenrikJannsen:
@suddenwhipvapor Can you squash the 3 commits so the accidental commits disappear?

PR #7338, opened 2024-12-27
Show ARS Blue Rate popup during payment account creation

alvasw:
At the moment, we show the ARS Blue Rate popup when the user saves a single currency payment account (PR #7122). This misses many payment accounts that support ARS. The new behaviour is to show the ARS Blue Rate popup when ARS is selected during account generation (either pre-selected or later selected by the user).

Payment accounts supporting ARS:
- CashByMail
- CashDeposit
- F2F
- MoneyGram
- National bank
- OKPayAccount (deprecated)
- Same bank
- Specific bank
- Transferwise
- Uphold
- Western Union


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #7342, opened 2025-01-01
Updater: Remove JAR_HASH download

alvasw:
Recent changes broke reproducibility and the last release's JAR_HASH
file contains "Dummy file (needed for updater)". We need to keep
creating a dummy file because the updater fails without a dummy file.
This change fixes the bug.


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #7343, opened 2025-01-01
Fix crash due to broken filter provided nodes merge

alvasw:
The filter provided nodes were added to a read-only List and this lead
to crashes at start-up. We didn't catch the bug before because the lists
in the tests are mutable. Now the FederatedBtcNodeProvider.getNodes
method operates on Streams.


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #7348, opened 2025-01-04
fix: remove unnecessary null check, use correct marker class, add mis…

### LINKED ISSUE #7347: bug in Preferences.setRpcUser?
https://github.com/bisq-network/bisq/blob/b3f44d2b4cda28dd04cd0218cb6e99cbe7c48e63/core/src/main/java/bisq/core/user/Preferences.java#L780-L787

the if statement is rendered useless, which one is correct ?

thecockatiel:
…sing return keyword, and fix Preferences.setRpcUser

as discussed and confirmed on matrix chat (bisq v1 dev channel), these are minor bugs, so I made a PR

resolves #7347

### DIFF
--- a/core/src/main/java/bisq/core/offer/placeoffer/bisq_v1/tasks/CheckNumberOfUnconfirmedTransactions.java
+++ b/core/src/main/java/bisq/core/offer/placeoffer/bisq_v1/tasks/CheckNumberOfUnconfirmedTransactions.java
@@ -13,8 +13,10 @@ public class CheckNumberOfUnconfirmedTransactions extends Task<PlaceOfferModel> 
 
     @Override
     protected void run() {
-        if (model.getWalletService().isUnconfirmedTransactionsLimitHit() || model.getBsqWalletService().isUnconfirmedTransactionsLimitHit())
-            return failed(Res.get("shared.unconfirmedTransactionsLimitReached"));
-        complete();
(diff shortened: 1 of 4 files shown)


core/src/main/java/bisq/core/offer/placeoffer/bisq_v1/tasks/CheckNumberOfUnconfirmedTransactions.java
```diff
@@ -14,7 +14,7 @@ public CheckNumberOfUnconfirmedTransactions(TaskRunner<PlaceOfferModel> taskHand
     @Override
     protected void run() {
         if (model.getWalletService().isUnconfirmedTransactionsLimitHit() || model.getBsqWalletService().isUnconfirmedTransactionsLimitHit())
-            failed(Res.get("shared.unconfirmedTransactionsLimitReached"));
+            return failed(Res.get("shared.unconfirmedTransactionsLimitReached"));
```
HenrikJannsen: The curly brackets at the if are missing. The failed() method has void as return type so it cannot be used as return statement.

HenrikJannsen (changes requested):

HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #7355, opened 2025-01-09
RestApi: Fix parseBlockCompleteAfterBatchProcessing data race

alvasw:
The parseBlockCompleteAfterBatchProcessing field is set on the block
parsing thread and read by Jersey's HTTP threads. By default, Jersey
handles requests synchronous on multiple threads.


alejandrogarcia83 (approved):
utACK

PR #7357, opened 2025-01-14
Prepare activation to reduce min security deposit

HenrikJannsen:
Implements https://github.com/bisq-network/proposals/issues/469


alejandrogarcia83 (approved):
utACK

HenrikJannsen:
I don't know why the test failed, it is unrelated to the changes. I did not had time to look further into it.

Also the activation date with 1.3. 2025 might be too tight as discussed on Matrix. So it might need a follow up PR.

PR #7368, opened 2025-01-31
Ensure AutoConfirmSettings is reset if it contains my retired explorer

devinbileck:
My XMR explorer was removed in bisq-network/bisq#7137. I just recently shut it down as per https://github.com/bisq-network/roles/issues/109#issuecomment-2581327905. This change ensures AutoConfirmSettings is reset if it contains my explorer address.


alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #7369, opened 2025-02-02
Makefile: Add exec-commands to .PHONY target

alvasw:
Add bitcoind, seednode, seednode2, mediator, alice, bob, block, and blocks to .PHONY target.


alejandrogarcia83 (approved):
utACK

PR #7373, opened 2025-02-02
Update data stores for v1.9.19

alejandrogarcia83:
light_1
full_1
light_2
full_2
light_3
full_3


alvasw (approved):
DAO hashes match with my full nodes.

HenrikJannsen (approved):
utACK

Matches with my full node

PR #7374, opened 2025-02-03
fix: payment account form issues

### LINKED ISSUE #7367: extraData in PaymentAccount needs to be excluded in PaymentAccountForm
currently this field is causing the payment account form to fail completely
reason is its included in all forms (since it's part of PaymentAccount), but not handled at invokeSetterMethod:
https://github.com/bisq-network/bisq/blob/6db1c9e21b6ffa0ecf2ae09134d9b3168ee47786/core/src/main/java/bisq/core/api/model/PaymentAccountTypeAdapter.java#L261-L270

since this is a map, and form only supports string or int right now, I assume this is field needs to be excluded

thecockatiel:
Fixes #7367


alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #7379, opened 2025-02-05
AltCoinAccountsDataModel: Handle XMR detection NullPointerException

alvasw:
The PaymentAccount.getSingleTradeCurrency(...) method can return null
in certain cases.


alejandrogarcia83 (approved):
utACK

PR #7386, opened 2025-02-07
fix: some issues in EditOfferOptionParser

thecockatiel:
some small issues I found in EditOfferOptionParser

`triggerPriceOpt` is a bit more serious

PR #7387, opened 2025-02-07
cli: Option parser checks if empty twice

alvasw:
The OptionSet.hasArgument(...) method [1] already check if the argument is empty.

[1] https://github.com/jopt-simple/jopt-simple/blob/jopt-simple-5.0.4/src/main/java/joptsimple/OptionSet.java#L127

Relates to #7386.

### DIFF
--- a/cli/src/main/java/bisq/cli/opts/AbstractMethodOptionParser.java
+++ b/cli/src/main/java/bisq/cli/opts/AbstractMethodOptionParser.java
@@ -74,8 +74,7 @@ abstract class AbstractMethodOptionParser implements MethodOpts {
         }
     }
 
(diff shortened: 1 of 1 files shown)


thecockatiel:
will it return `false` if an option is passed like `--trade-id` but no value is given?
edit: sorry, I hadn't check the link yet
this is correct

thecockatiel:
but somehow if you use `=` and then specify no value, it returns true, and that's why the test is not passing
so the check is probably necessary

alvasw:
My bad, the check is needed because "--trigger-price=" is turned into "--trigger-price=\"\"" (empty string after equal sign) before parsing it.

PR #7388, opened 2025-02-11
chore: print RemoveAssetProposal instead of generic

thecockatiel:
small fix to `toString` to reflect name of class correctly


alvasw (approved):
utACK

HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #7390, opened 2025-02-12
fix: remove unnecessary coin usage

thecockatiel:
It seems to be unnecessary
if it's not, can you explain, so we can add a comment for it ?

### DIFF
--- a/core/src/main/java/bisq/core/dao/burningman/accounting/blockchain/AccountingTxOutput.java
+++ b/core/src/main/java/bisq/core/dao/burningman/accounting/blockchain/AccountingTxOutput.java
@@ -21,8 +21,6 @@ import bisq.core.dao.burningman.BurningManPresentationService;
 
 import bisq.common.proto.network.NetworkPayload;
 
-import org.bitcoinj.core.Coin;
-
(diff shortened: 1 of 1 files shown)


HenrikJannsen (dismissed):
utACK

alvasw (changes requested):
It looks like the developer wanted to handle unsigned integers correctly because the protobuf definition of `AccountingTxOutput.value` is a `uint32` (unsigned integer) [1]. Java doesn't natively support unsigned integer and therefore we need to use the datatype long. Before persisting the data we need to convert the long back to an unsigned integer. The conversion is wrong here but it doesn't cause any bugs in this case.

`2147483647` equals to `Coin.valueOf(2147483647).getValue()`. Could you please replace `2147483647` with `Integer.MAX_VALUE` and change the second argument to `"We only support integer values in protobuf storage for the amount."` to make the code more readable? Thank you!

[1] https://github.com/bisq-network/bisq/blob/631d0f9222e9c009b3f4fedc70d68b57228f730b/proto/src/main/proto/pb.proto#L2032

alejandrogarcia83 (approved):
utACK

PR #7393, opened 2025-02-13
Handle burning man selection height edge cases

alvasw:
If trader A's block height is 134 his burning man selection height is
120, and if trader B's block height is 135 his burning man selection
height is 130.

The current burning man selection height verification code doesn't
handle these edge-cases.

Changes:
- [DelayedPayoutTxReceiverService: Make bm snapshot grid size a constant](https://github.com/bisq-network/bisq/commit/353ff6d06f11d7f6eed7a4f1be52cfd0ad7e0c68)
- [Handle burning man selection height edge-cases](https://github.com/bisq-network/bisq/commit/96355d7a3cd0ad3cf5d6dfbe62c6cfd28e9a4e45)

PR #7395, opened 2025-02-14
Use taker's burning man selection height

alvasw:
When a taker takes a trade he sends his burning man selection height to
the maker. After validating the taker's burning man selection height, he
has to use the taker's burning man selection height to verify the
taker's delayed payout transaction because the maker's and taker's
burning man selection height could be different. The current
implementation uses the maker's burning man selection height. This can
cause failed trades.

Relates to https://github.com/bisq-network/bisq/pull/7393.


HenrikJannsen (dismissed):
utACK

alejandrogarcia83 (approved):
utACK

PR #7412, opened 2025-03-06
BisqExecutable: Fix startApplication() threading issue

alvasw:
The startApplication method is only called on the user thread if the
user didn't downgrade his app. The readAllPersisted method calls
startApplication on the user thread. However, the code path handling
downgrades doesn't call startApplication on the user thread.


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #7413, opened 2025-03-06
BurningManAccountingService: Use synchronized map for balanceModelByBurningManName

alvasw:
This field needs to be thread-safe because the UI thread and multiple
worker threads write to the variable.


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #7426, opened 2025-03-28
Add run command with specific JAVA_HOME in build.md

suddenwhipvapor:
Updated the doc to reflect a situation I had to fix on my own system, and suppose would be common for those who have multiple Java versions installed.


alvasw (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #7432, opened 2025-04-07
fix: use payment-account-id param name consistently

thecockatiel:
makes usage of `payment-account-id` parameter name more consistent
the correct name is `payment-account-id`, but it was sometimes referred to as `payment-account`

https://github.com/bisq-network/bisq/blob/ce2ffceccf5fc1c896f25276a7b71f7aa64ddef5/cli/src/main/java/bisq/cli/opts/OptLabel.java#L42

## Summary by CodeRabbit

- **Documentation**
  - Updated help texts and documentation to consistently use `--payment-account-id` instead of `--payment-account` for relevant CLI commands.

- **Chores**
  - Updated scripts to use the new `--payment-account-id` parameter name in CLI commands for offer creation and trading.

### DIFF
--- a/apitest/scripts/limit-order-simulation.sh
+++ b/apitest/scripts/limit-order-simulation.sh
@@ -115,7 +115,7 @@ done
 
 printdate "ALICE: Creating $DIRECTION $CURRENCY_CODE offer with payment acct $ALICE_ACCT_ID."
 CMD="$CLI_BASE --port=$ALICE_PORT createoffer"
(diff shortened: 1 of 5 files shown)


alvasw (approved):
ACK

- Grep for `payment-account` and `payment-account=` before change
- Grep for `payment-account` and `payment-account=` after change

alejandrogarcia83 (approved):
utACK

PR #7434, opened 2025-04-09
BlockchainExplorerSelection: Don't compare btc explorers with bsq explorer blacklist

alvasw:


core/src/main/java/bisq/core/user/BlockchainExplorerSelection.java
```diff
@@ -25,8 +25,7 @@ public void selectNodes() {
 
         // if no valid Bitcoin block explorer is set, select the 1st valid Bitcoin block explorer
         ArrayList<BlockChainExplorer> btcExplorers = preferences.getBlockChainExplorers();
-        if (preferences.getBlockChainExplorer() == null ||
-                preferences.getBlockChainExplorer().getName().matches(deprecatedExplorers)) {
```
HenrikJannsen, line 29: Even we dont use btc explorers in deprecated it could happen in future, no?

HenrikJannsen (approved):
utACK

core/src/main/java/bisq/core/user/BlockchainExplorerSelection.java
alvasw: Yes, that's correct.

PR #7437, opened 2025-04-10
fix: change protobuf definition to use repeated fields for consistency

thecockatiel:
This change looks quite large but it's very simple: It changes maps to repeated fields for deterministic serialization of maps
This is as we discussed it in the matrix chat
I have left the DaoState to still use hashmap before serializing to the new deterministic approach as we need to discuss it further

the reason for having separate definition for each entry is because protobuf does not have generics and using oneof for that purpose in our case seemed like a bad choice 

I have also [added a test](https://github.com/bisq-network/bisq/pull/7437/files#diff-c4afb7263827d76f64fa682c918dcd73a94d50b73bdf67619bfbc9048ff03757) to demonstrate the change and how serialization of bytes are affected
the test also serves as a tool to detect if somehow protobuf decides to do things differently in future, so we don't fail silently in case serialization changes
but the test needs improvement

one immediate issue that I realized right now is that if newer client sends data back to an old client, and old client tries to parse that into a hashmap (using `Collectors.toMap`), the order can/will be different, which can have complications

the other way around is not an issue though.

I also normalized a few protobuf field names to snake case so that they are not generated weirdly on python and possibly other languages, like the rest of the fields

## Summary by CodeRabbit

- **New Features**
  - Introduced an enhanced data serialization framework that streamlines the transformation of supplementary data across various messaging and account operations.
- **Bug Fixes**
  - Improved data ordering and conversion, ensuring consistent and reliable behavior in payload handling during network communications.
- **Tests**
  - Added comprehensive tests to validate backward compatibility and data integrity across updated message formats.


thecockatiel:
one immediate issue that I realized right now is that if newer client sends data back to an old client, and old client tries to parse that into a hashmap (using `Collectors.toMap`), the order can/will be different, which can have complications

the other way around is not an issue though.

alvasw:
This PR affects Alerts, Filters, and many DAO parameters. I didn't follow the matrix chat. What is the reason to enforce the deterministic behavior? Is it because of the signature?

If an old client receives data from a client with this change the order will be different and the signature checks fail.

If we need this to verify the signature on other platforms, it's less risky to add a second signature field that sorts the keys and values deterministically before generating and verifying the p2p network objects. In the next version we can check both signature in native Bisq clients and later deprecate the old signature format. We need to wait until the whole network has updated to the new version before we can deprecate the old signature format.

If you extract the proto name changes into a separate PR, we can merge it earlier.

@HenrikJannsen What do you think?

HenrikJannsen:
I would say we park that PR as not intended to be implemented as it cause too much effort, risks. Once a larger DAO refactoring effort is launched we can address it.

PR #7423, opened 2025-03-19
fix: return after calling failed

thecockatiel:
adds return after a few `failed()` calls, like the rest of tasks

PR #7445, opened 2025-04-29
MakerProcessesInputsForDepositTxRequest: Merge all error catching cases

alvasw:
We have `failed(...)` methods for `String`s and `Throwable`s. The `failed(Throwable t)` extracts the `Throwable`'s message and appends it to the `errorMessage`. Therefore, both catch clauses are identical. We can remove the other catch clause because the most outer catch clause is identical to the inner one.

Relates to #7423.

## Summary by CodeRabbit

- **Refactor**
  - Updated the handling of trade price verification to allow errors to propagate naturally, removing previous exception handling logic.


thecockatiel:
utACK

alejandrogarcia83 (approved):
utACK

PR #7456, opened 2025-05-28
Return correct matrix address for refundagent2

suddenwhipvapor:
refundagent2 pointed out the arbitration message reported a wrong matrix address for his account.
The method returning the address has `:matrix.org` hardcoded, so I had to add a condition check.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated Matrix chat links so that the "refundagent2" agent now directs users to a new domain. All other agents continue to use the original domain for their Matrix links.


alejandrogarcia83 (approved):
utACK

suddenwhipvapor:
@alejandrogarcia83 this still needs to be merged, was this by choice?

PR #7459, opened 2025-05-30
Remove ClearXchange references from payment methods UI

suddenwhipvapor:
As per https://bisq.community/t/zelle-clear-xchange-2022-shutdown/13014/1 I have removed the reference to ClearXchange, which was shutdown in 2022, from the Zelle line in the fiat account type dropdown, without refactoring the variable names as to avoid unnecessary complication.

## Summary by CodeRabbit

- **Style**
  - Updated the display name for the payment method previously shown as "Zelle (ClearXchange)" to "Zelle" across all supported languages.


alejandrogarcia83 (approved):
utACK

PR #7483, opened 2025-07-08
API: allow filtering BTC addresses with positive balance

helixx87:
This PR introduces a new `--only-funded` flag to the `getfundingaddresses` method, allowing users to filter the output to include only addresses with a positive balance. It improves usability by helping users focus on relevant, active addresses and reduces unnecessary output in large wallets. The default behavior remains unchanged when the flag is not used.

## Summary by CodeRabbit

* **New Features**
  * Added an optional command-line parameter to the `getfundingaddresses` command, allowing users to filter and display only funded addresses.
* **Documentation**
  * Updated help text to describe the new `--only-funded` option for the `getfundingaddresses` command.
* **Tests**
  * Adjusted tests to accommodate the new parameter for filtering funded addresses.

### DIFF
--- a/cli/src/main/java/bisq/cli/CliMain.java
+++ b/cli/src/main/java/bisq/cli/CliMain.java
@@ -67,6 +67,7 @@ import bisq.cli.opts.GetAddressBalanceOptionParser;
 import bisq.cli.opts.GetAvgBsqPriceOptionParser;
 import bisq.cli.opts.GetBTCMarketPriceOptionParser;
 import bisq.cli.opts.GetBalanceOptionParser;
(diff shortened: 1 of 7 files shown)


alvasw (approved):
ACK

- Tests
	- Test cli without specifying --only-funded flag
	- Test API call with --only-funded flag set to true
	- Test API call with --only-funded flag set to false (explicitly)
- Balance filtering happens on client-side and only cli users benefit from this feature. I moved the change to the server-side (follow up PR).

alejandrogarcia83 (approved):
utACK

Issue #7448, opened 2025-05-12
Bisq does not allow multiple BTC nodes in the `--btcNodes` command line parameter


Akira45-0:
### Description

Bisq does not allow multiple BTC nodes in the `--btcNodes` command line parameter

#### Version
1.9.19

### Steps to reproduce

Run Bisq from command line with more than one server in the `btcNodes` parameter, for example:
`Bisq --btcNodes=2oalsctcn76axnrnaqjddiiu5qhrc7hv3raik2lyfxb7eoktk4vw6sad.onion:8333,ybryiy2k4p4pery4qseap4iu2rxput2akuvpvczwvg4eyfafcdsvyqid.onion:8333`

### Expected behaviour

Bisq should start and use the servers passed through the btcNodes command line parameter

### Actual behaviour

Bisq fails to start. You receive the following error message:
`error: problem parsing option 'btcNodes': Found multiple arguments for option btcNodes, but you asked for only one`

#### Additional info

I would like to point out that in line 212 of the Config.java file we can see:
`public final String btcNodes;`
instead of:
`public final List<String> btcNodes;`
which may be the cause of this bug.

https://github.com/bisq-network/bisq/blob/631d0f9222e9c009b3f4fedc70d68b57228f730b/common/src/main/java/bisq/common/config/Config.java#L212

mustardcaviar:
Hi, I'd like to work on this issue.  Could you please assign it to me?

PR #7484, opened 2025-07-09
Fix: Allow multiple btcNodes in command-line argument

mustardcaviar:
Fixes issue #7448

Bisq would not load if multiple nodes were provided via the `--btcNodes` command-line argument because the value was being parsed as a single `String`

To solve that: 
- Updated `Config.java` to correctly parse the `--btcNodes` argument as a `List<String>`

- Updated `BitcoinModule.java` to handle the new list type during dependency injection

- Updated `Preferences.java` to correctly process the list of nodes from the configuration

## Summary by CodeRabbit

* **New Features**
  * Enhanced BTC nodes configuration to support specifying multiple nodes instead of a single entry.

* **Bug Fixes**
  * Improved handling of BTC node options to allow for multiple entries, ensuring more flexible and robust configuration.

### DIFF
--- a/common/src/main/java/bisq/common/config/Config.java
+++ b/common/src/main/java/bisq/common/config/Config.java
@@ -209,7 +209,7 @@ public class Config {
     public final int msgThrottlePer10Sec;
     public final int sendMsgThrottleTrigger;
     public final int sendMsgThrottleSleep;
-    public final String btcNodes;
+    public final List<String> btcNodes;
     public final boolean useTorForBtc;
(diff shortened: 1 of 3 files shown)


alvasw (approved):
ACK

Notes:
- Setup second bitcoin core node and connect to first bitcoin core node
- Patch WalletsSetup to bypass localhost auto connection
- Start bisq instance that connects to both bitcoin core nodes

alejandrogarcia83 (approved):
utACK

PR #7515, opened 2025-10-04
api: Move BTC positive balance filtering to API-level

alvasw:
PR https://github.com/bisq-network/bisq/pull/7483 added the --only-funded flag to the getfundingaddresses method.
This flag was added to bisq-cli and the filtering happened client-side.
This change moves the filtering to the server-side, so that all API
users can use the 'only-funded' option.

## Summary by CodeRabbit

* New Features
  * Added --only-funded option to getfundingaddresses, allowing display of only addresses with a positive balance.
* API
  * gRPC now accepts an only_funded flag in the funding addresses request to filter results to funded addresses.
* Documentation
  * Updated help text for getfundingaddresses to describe the new filtering option.
* Tests
  * Updated tests to cover the new only-funded behavior.


alvasw:
Tests:
- Setup
	- Generate few new addresses
	- Fund some of the generated addresses
- Test API call without specifying --only-funded flag
- Test API call with --only-funded flag set to true
- Test API call with --only-funded flag set to false

alejandrogarcia83 (dismissed):
utACK

PR #7527, opened 2025-10-23
Update suddenwhipvapor node onion

suddenwhipvapor:
A human error led to backing up the onion private key for node o7cl into the file for izdw, so the real private key for izdw was lost.
When restoring the node, I had to generate a new onon.

## Summary by CodeRabbit

* **Chores**
  * Updated Bitcoin node address configuration to maintain network connectivity and service reliability.


alvasw (approved):
utACK

PR #7556, opened 2025-12-02
Fix bug with applying snapshot

HenrikJannsen:
This fixes a critical problem when chain forks happen. It caused an inconsistent dao state from a resync of the seed nodes and incorrect dao state hashes.

Cause of the problem:
We accessed the dao data from the store which gets cleared after initial application of the persisted data.

Now we load the data from disk and use that. This adds asynchronous handling for the parsing after the application of the snapshot.

It also fixes the comparison of previous applied snapshots, which is now only set when called from a resync, not from the initial application of the persisted data.


alvasw:
ACK

- At startup still null - (checked in debugger)
	- Only affects clients, doesn't matter for seednodes because we are already at least 20 blocks behind.

PR #7564, opened 2025-12-16
Add bitcoind v29.2 support

alvasw:
- [Migrate to new bitcoind RPC implementation](https://github.com/bisq-network/bisq/commit/0a4dba6cc75db70524a9008872aeff278bd6f6cf)
  - The new bitcoind RPC implementation supports Bitcoin Core 29.2.
- [core: Remove unused jackson.databind dependency](https://github.com/bisq-network/bisq/commit/7353adb7dd1a71114885b47b82813f9d74b7864d)

## Summary by CodeRabbit

* **Refactor**
  * Reworked Bitcoin RPC integration to use native bitcoind response types and a new client, updating block/transaction handling and DTO mappings.

* **Dependencies**
  * Added bitcoind core and JSON‑RPC libraries; removed explicit jackson-databind mapping and added verification metadata entries.

* **Tests**
  * Updated tests to new response types and removed several legacy test fixtures and a client test suite.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


core/src/main/java/bisq/core/dao/node/full/rpc/DtoPubKeyScript.java
```diff
@@ -0,0 +1,45 @@
...
+        // However they are exported into the DAO classes (and therefore a component of the DAO state hash)
+        // so we must generate address from the hex script using BitcoinJ.
+        // (n.b. the DAO only ever uses/expects one address)
+        try {
+            String address = new Script(Hex.decode(hex))
+                    .getToAddress(Config.baseCurrencyNetworkParameters()).toString();
+            this.addresses = List.of(address);
+            this.reqSigs = 1;
+        } catch (Exception ex) {
+            // certain scripts e.g. OP_RETURN do not resolve to an address
```
HenrikJannsen, line 54: Shouldn't we set addresses to an empty list in that case to avoid potential NPEs?

core/src/main/java/bisq/core/dao/node/full/RpcService.java
```diff
@@ -430,26 +442,27 @@ private static RawTx getTxFromRawTransaction(RawDtoTransaction rawDtoTx,
         // it until the fork activates, which is determined by block height.
         boolean allowSegwit = blockHeight >= getActivateHardFork2Height();
 
-        final List<TxInput> txInputs = rawDtoTx.getVIn()
+        final List<TxInput> txInputs = rawDtoTx.getVin()
                 .stream()
-                .filter(rawInput -> rawInput != null && rawInput.getVOut() != null && rawInput.getTxId() != null)
+                .filter(rawInput -> rawInput != null && rawInput.getTxId() != null)
```
HenrikJannsen, line 447: Is the null check with rawInput.getVout() not needed anymore?

core/src/main/java/bisq/core/dao/node/full/rpc/DtoPubKeyScript.java
```diff
@@ -0,0 +1,45 @@
...
+
+        // >> addresses are not provided by bitcoin RPC from v22 onwards. <<
+        // However they are exported into the DAO classes (and therefore a component of the DAO state hash)
+        // so we must generate address from the hex script using BitcoinJ.
+        // (n.b. the DAO only ever uses/expects one address)
+        try {
+            String address = new Script(Hex.decode(hex))
+                    .getToAddress(Config.baseCurrencyNetworkParameters()).toString();
+            this.addresses = List.of(address);
+            this.reqSigs = 1;
```
HenrikJannsen: As reqSigs is hard coded 1 and not used beside in PubKeyScript protobuf (and therefor has impact on dao hash), I think its best to remove it from here and only set it hardcode as final value in PubKeyScript with a comment explaining its a required value for backward compatibility. I guess the value would be in all DAO txs 1 in any case.

HenrikJannsen (changes requested):

core/src/main/java/bisq/core/dao/node/full/rpc/DtoPubKeyScript.java
alvasw: I applied this change.

Notes:
- How do empty list get serialized to protobuf wireformat?
	- The default value for repeated fields is empty (generally an empty list in the appropriate language). (https://protobuf.dev/programming-guides/proto3/#default)
		- Write PubKeyScript test case -> Yes, we always set it to same value.

core/src/main/java/bisq/core/dao/node/full/RpcService.java
alvasw: No, it is not needed because every bitcoin transaction has at least one input and the new bitcoind rpc implementation enforces it.

core/src/main/java/bisq/core/dao/node/full/rpc/DtoPubKeyScript.java
alvasw: No, we can't hardcode it to 1 because we set it to null and fallback to 0 in certain dao transactions. One example are transactions of type "COMP_REQ_OP_RETURN_OUTPUT" with scriptType "ScriptType.NULL_DATA(name=nulldata)". In this case reqSigs is always 0 and addresses are null. This is wrong but we can't change it without breaking the dao hashes.

core/src/main/java/bisq/core/dao/state/model/blockchain/PubKeyScript.java
```diff
@@ -39,15 +37,14 @@
 public class PubKeyScript implements PersistablePayload, ImmutableDaoStateModel {
```
HenrikJannsen, line 35: Maybe remove @AllArgsConstructor and use a explicit constructor as it makes more visible that the addresses when this constr. is called can be null, but it seems the new code expect a non-nullable address. Not sure if an empty list can be used or if it would carry risks.

core/src/main/java/bisq/core/dao/state/model/blockchain/ScriptType.java
```diff
@@ -66,8 +66,7 @@ public static ScriptType forName(String name) {
                 }
             }
         }
-        throw new IllegalArgumentException("Expected the argument to be a valid 'bitcoind' script type, "
-                + "but was invalid/unsupported instead. Received scriptType=" + name);
+        return UNDEFINED;
```
HenrikJannsen: The protobuf is not 100% mirroring the java class:
`ScriptType.UNDEFINED` vs. `PB_ERROR_SCRIPT_TYPES`

In 
```
  public static ScriptType fromProto(protobuf.ScriptType scriptType) {
        return ProtoUtil.enumFromProto(ScriptType.class, scriptType.name());
    }
```

we convert an unknown type (like `anchor`) to UNDEFINED. But in would throw an `IllegalArgumentException` as no `UNDEFINED` exist on the protobuf enum.
```
    public protobuf.ScriptType toProtoMessage() {
        return protobuf.ScriptType.valueOf(name());
    }
```

With this change we move the exception  escalation from `forName` to `toProtoMessage`. 

To minimize risks for DAO inconsistency, I think its better to stick with the old implementation as far as possible.

HenrikJannsen (changes requested):

core/src/main/java/bisq/core/dao/state/model/blockchain/PubKeyScript.java
alvasw: I applied this change.

core/src/main/java/bisq/core/dao/state/model/blockchain/ScriptType.java
alvasw: I checked the P2A spec (https://github.com/bitcoin/bips/blob/master/bip-0433.mediawiki) and on-chain transaction with anchor outputs. Bitcoin Core v27 returns `"type": "witness_unknown"` and Bitcoin Core v29 `"type": "anchor"`.

Now the code returns `witness_unknown` for anchor outputs.

_____

What are anchor transactions?
- "keyless" meaning no signature or witness data required to spend
- P2A is native segwit output script (v1)

HenrikJannsen (approved):
ACK

HenrikJannsen:
@alvasw I cannot run it on M3 mac anymore. It seems I get a Intel javafx binary.

alvasw:
I spoke with @HenrikJannsen in private and his issue was unrelated to this PR. @alejandrogarcia83 we can finally merge this PR.

HenrikJannsen:
Yes, my issue was that I required to use a intel JDK on my M3 notebook (which is still weird as before I could compile and run with aarch64 jdk). I could now resycn from resources and got the same hashes as the lite node.

alejandrogarcia83 (approved):
utACK

PR #7571, opened 2025-12-23
Update binary build instructions, remove windows support

suddenwhipvapor:
Instructions to build Bisq from source were fixed to reflect currently working method (running bisq-desktop from repo root, and running ./gradlew :desktop:run were not functional anymore).
Also removed references to Windows, since it is unlikely the existing directions would still work, and also that any contributor would be able to support building from source on Windows.

## Summary by CodeRabbit

* **Documentation**
  * Updated build instructions and release reference to the latest release.
  * Streamlined build and runtime invocation guidance (explicit Java environment setup and clean build step).
  * Corrected executable/run paths for all platforms and cleaned up formatting in docs.

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -5,8 +5,8 @@
    ```sh
    git clone https://github.com/bisq-network/bisq
    # if you intend to do testing on the latest release, you can clone the respective branch selectively, without downloading the whole repository
-   # for the 1.9.18 release, you would do it like this:
-   git clone --recurse-submodules --branch release/v1.9.19 https://github.com/bisq-network/bisq
+   # for the 1.9.21 release, you would do it like this:
+   git clone --recurse-submodules --branch release/v1.9.21 https://github.com/bisq-network/bisq
    cd bisq
    ```
 
@@ -15,13 +15,8 @@
    On macOS and Linux, execute:
    ```sh
    ./gradlew build
+   ./gradlew installDist
    ```
-
-   On Windows:
-   ```cmd
-   gradlew.bat build
-   ```
-
    If you prefer to skip tests to speed up the building process, just append _-x test_ to the previous commands.
 
 ### Important notes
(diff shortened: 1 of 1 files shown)


suddenwhipvapor:
Is this not a useful PR?

docs/build.md
```diff
@@ -15,13 +15,8 @@
    On macOS and Linux, execute:
    ```sh
    ./gradlew build
+   ./gradlew installDist
```
alvasw: This is not needed because `./gradlew build` calls the `installDist` task (https://github.com/bisq-network/bisq/blob/80f5ed863cbc9ed88131556ac7905817c764450c/build-logic/commons/src/main/groovy/bisq.application.gradle#L6).

alvasw (changes requested):
Thank you for opening this PR! I didn’t realize the scripts were broken, and it seems like only end-users are currently using them to launch Bisq. I’ve fixed the scripts in PR #7597, added proper Windows support, and moved the scripts into the scripts directory. Could you please update the instructions to reflect these changes?

Users can run any script within the scripts directory in a shell. This applies to macOS, Linux, and Windows. The command for the desktop application would be:
```
scripts/desktop
```

It’s better to switch to the correct JDK version (if multiple versions are installed) rather than setting the JAVA_HOME variable. You can do this using the following commands:

For Debian-based OSes:
```
sudo update-java-alternatives --set zulu-11-amd64
```
For RedHat-based OSes:
```
sudo alternatives --config java
```

suddenwhipvapor:
Hi alva, hope you're well!
Thank you for the comment, here's my follow-up:
- I see your PR is still to be merged, I will go ahead with the `scripts/desktop` and `scripts/desktop.bat` reference
- I use  `JAVA_HOME` in my setup because I run both bisq1 and bisq2 from source and that's a simple way of selecting the required java version without forcing it system-wide, if I used `update-java-alternatives` then I (and the other users who run from source both 1 and 2) would still have to use `JAVA_HOME` to select Java22 to launch Bisq2... so with your approval I would rather leave it the way it is 😅 
- I will remove the `installDist` line
- I won't be able squash commits since I only use the web interface
- I hope I won't make it worse by editing my existing PR 😄

PR #7580, opened 2026-01-27
UX: Show disk space warning only once

viresinnumer1s:
This PR fixes an issue that causes Bisq to constantly nag the user about low disk space, instead of just warning the user once.

Changes:
 - Added a bool to store if the storage warning has been shown yet
 - When checkFreeDiskSpace() is ran, it will check for low disk space, and, if disk space is low, will warn user and set bool to true.
 - When re-ran, if disk space is low and the user has already been warned, then it will do nothing.

I wrote this PR as anyone who has experienced this knows how annoying it is to get the error pop up randomly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed repeated disk space warnings. The low disk space alert now displays only once per application session.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqSetup.java
+++ b/core/src/main/java/bisq/core/app/BisqSetup.java
@@ -557,16 +557,21 @@ public class BisqSetup {
             }
         });
     }
-
+    
+    boolean hasShownStorageWarning = false;
     private void checkFreeDiskSpace() {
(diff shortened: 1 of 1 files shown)


HenrikJannsen (approved):
utACK

PR #7593, opened 2026-02-24
Delay shutdown when payout transaction broadcasts are pending

alvasw:
A support team member reported an issue where sellers confirm receipt of payment, but the payout transaction is not broadcasted because they close Bisq before the transaction is broadcasted.

To address this, we now check for pending payout transaction broadcasts during shutdown and delay the shutdown if necessary.

## Summary by CodeRabbit

* **New Features**
  * Shutdown now waits briefly (up to 2s) to ensure pending payment-received messages are published before the app stops, reducing risk of lost notifications during shutdown.
  * Release flow now tracks payment-received messages to ensure they are sent even on failures.

* **Tests**
  * Added tests covering shutdown wait behavior and payment-message handling.


HenrikJannsen (approved):
utACK

alejandrogarcia83 (approved):
utACK

PR #7597, opened 2026-02-28
Fix broken Bisq launch scripts

alvasw:
The previous approach of copying app start scripts from build directories to the root and adjusting them no longer works. This commit introduces wrapper scripts that invoke the custom `<app>:startBisqApp` Gradle task.

## Summary by CodeRabbit

* **New Features**
  * Added cross‑platform launcher scripts to start apps and services from the project root.

* **Chores**
  * Standardized startup/configuration across modules and updated usage docs to reference script-based entry points.
  * Replaced custom packaging/startup tweaks with default behavior to simplify builds.
  * Restored previously ignored patterns so matching paths are now tracked.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,7 +22,6 @@ build
 *.java.hsw
 *.~ava
 /bundles
-/bisq-*
 /lib
 /xchange
 desktop.ini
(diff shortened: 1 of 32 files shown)


alejandrogarcia83 (approved):
utACK

PR #7598, opened 2026-02-28
Move seednode JVM arguments to global configuration

alvasw:
The AppStartPlugin currently overrides heap settings for the seednode,
and the bisq seednode service unit overrides both heap settings and other JVM
arguments.

These overrides are not applied in the dev setup, which can lead to
inconsistencies and potentially hide bugs during local testing.

This change moves the seednode JVM arguments to the global configuration
to ensure consistent behavior across environments.

## Summary by CodeRabbit

* **Chores**
  * Reorganized JVM heap size configuration for seednode. Memory initialization and maximum heap settings (`4096M`), along with out-of-memory error handling and network stack preferences, have been consolidated into the build configuration.


alejandrogarcia83 (approved):
utACK

Issue #3010, opened 2019-07-26
Allow the user to configure Tor before connecting to it


anon88391:
Currently when installing Bisq for the first time and running it the software connects to the Tor network without letting the user configure it. Only after the initial connection is established the user is allowed to configure Tor.
This is poses a significant risk to some users' privacy, anonymity and security. In some countries with widespread censorship just connecting to the Tor network can cause negative consequences from the government and put the user in a very bad light forever. It is therefore very important to let the user configure the Tor connection before it's established so that they can take necessary countermeasures against detection.
Connecting to Tor before having the ability to configure it goes against Tor's philosophy. Tor Browser doesn't connect to Tor without letting the user configure the connection first. 
Please fix this and allow the user to configure Tor before connecting to it. Thank you.

mrosseel:
do you have any examples of how other desktop apps handle this, particularly UI-wise ?

ghost:
@anon88391 
I suppose your concerns are about the /etc/tor/torrc proper tuning.
Could you give some examples of some tuning(s) you think about.

freimair:
Means of providing a torrc config file or even single config parameters have been introduced back in 2018 via https://github.com/bisq-network/bisq/pull/1893. Please give more details so we can enhance the software.

anon88391:
@mrosseel 
A desktop instant messenger app called Ricochet utilizes Tor network to establish connections between user, althought it hasn't been updated in a while. When running Ricochet for the first time it asks the user to configure Tor connection settings before establishing any connection.
Whonix, very good software that is a virtual machine used to route other virtual machines through Tor also does this. It doesn't establish any connections before the user configures the connection settings.
@HarryMacfinned 
Yes indeed, by saying "configure Tor connection settings" I'm actually referring to the torrc settings, where the user can change them on a high-level (bridges, proxy settings, etc).
@freimair 
That is not what I meant to address. I'm aware that Bisq has means of providing a torrc config file, however the issue stems from the fact that when launching Bisq for the first time, the user is never asked to configure the connection settings (set up necessary bridges or proxy settings) before the first connection is established. Bisq first establishes Tor connections and only then allows the user to change the settings, which is a significant risk for some users.

freimair:
so what you are saying is you like to see
- the Tor connection settings dialog pop up on first start of Bisq?
- then, maybe transition some torrc configs to the GUI?
- and of course, the popup has to be reworked for usability in its new role?

anon88391:
[...]

PR #7600, opened 2026-03-02
Fix bridge address initialization order in TorNetworkSettingsWindow

5andr0:
### Problem

Opening the Tor Network Settings window corrupts the stored bridge address list,
causing Tor to start without bridges on the next restart — even if the window is
closed without making any changes. The text field still displays the correct bridge
address, making the bug invisible to users.

The root cause is an initialization ordering issue in
`TorNetworkSettingsWindow.addContent()`. Two fields are loaded from preferences
*after* `applyToggleSelection()` is called:

```java
applyToggleSelection();   // calls setBridgeAddressesByCustomBridges() here
                          // but customBridges is still "" (class field default)
selectedTorTransportOrdinal = ...  // too late
customBridges = preferences.getCustomBridges();  // too late
```

When the bridge option is `CUSTOM`, `applyToggleSelection()` calls
`setBridgeAddressesByCustomBridges()`, which splits the empty string `""` and
immediately force-persists `[""]` to disk via `Preferences.setBridgeAddresses()` →
`forcePersistNow()`. The `customBridges` display string is loaded correctly on the
next line, so the UI looks fine, but `bridgeAddresses` — the field actually read by
`BridgeAddressProvider.getBridgeAddresses()` at startup and passed to the Tor process
— is now `[""]`.

At the next startup, `NewTor.getTor()` receives `[""]` and Tor starts
without `UseBridges 1` in its torrc. The startup log prints `[StartTor] INFO
bisq.network.p2p.network.NewTor: Using bridges:` (blank).

The same issue affects `PROVIDED` bridges: if the stored transport type is not OBFS4
(the class field default for `selectedTorTransportOrdinal`), `applyToggleSelection()`
overwrites `bridgeAddresses` with OBFS4 bridges before the real transport ordinal is
loaded.

### How to reproduce (failure)

1. Set a custom bridge in Tor Network Settings, click Shut Down, confirm bridges work
   on restart (`Using bridges: <address>` in log)
2. Open Tor Network Settings again
3. Make no changes, click Shut down
4. Restart Bisq
5. Log shows `Using bridges:` (empty) — Tor connects without bridges

### How to work around it (before this fix)

After opening the window, trigger any listener before closing:
- Click **No bridges**, then click **Custom bridges** again, or
- Click in the bridge text field and make any edit (e.g. add and delete a space)

Either action fires a listener that was registered *after* `applyToggleSelection()`,
at which point `customBridges` is already correctly loaded, so
`setBridgeAddressesByCustomBridges()` re-persists the correct value.

### Fix

Load `selectedTorTransportOrdinal` and `customBridges` from preferences *before*
calling `applyToggleSelection()`, so the method always operates on the persisted
values rather than class field defaults.

### Disclaimer

I have no time to compile and test this patch. Can somebody test it please?

Regarding this topic, I also second this: [Allow the user to configure Tor before connecting to it #3010
](https://github.com/bisq-network/bisq/issues/3010)

## Summary by CodeRabbit

* **Refactor**
  * Improved initialization ordering in Tor network settings to ensure persisted values are loaded at the appropriate stage of the setup process.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/overlays/windows/TorNetworkSettingsWindow.java
+++ b/desktop/src/main/java/bisq/desktop/main/overlays/windows/TorNetworkSettingsWindow.java
@@ -295,6 +295,8 @@ public class TorNetworkSettingsWindow extends Overlay<TorNetworkSettingsWindow> 
 
         // init persisted values
         selectedBridgeOption = BridgeOption.values()[preferences.getBridgeOptionOrdinal()];
+        selectedTorTransportOrdinal = Transport.values()[preferences.getTorTransportOrdinal()];
+        customBridges = preferences.getCustomBridges();
         switch (selectedBridgeOption) {
             case PROVIDED:
                 toggleGroup.selectToggle(providedBridgesRadioButton);
@@ -309,10 +311,7 @@ public class TorNetworkSettingsWindow extends Overlay<TorNetworkSettingsWindow> 
         }
         applyToggleSelection();
 
-        selectedTorTransportOrdinal = Transport.values()[preferences.getTorTransportOrdinal()];
         transportTypeComboBox.getSelectionModel().select(selectedTorTransportOrdinal);
-
-        customBridges = preferences.getCustomBridges();
         bridgeEntriesTextArea.setText(customBridges);
 
         toggleGroup.selectedToggleProperty().addListener((observable, oldValue, newValue) -> {


wodoro:
> ### How to reproduce (failure)
> 
>     1. Set a custom bridge in Tor Network Settings, click Shut Down, confirm bridges work
>        on restart (`Using bridges: <address>` in log)
> 
>     2. Open Tor Network Settings again
> 
>     3. Make no changes, click Shut down
> 
>     4. Restart Bisq
> 
>     5. Log shows `Using bridges:` (empty) — Tor connects without bridges

Could not reproduce
Bridges are working as expected on v1.10.1 by following the steps exactly

wodoro:
Managed to reproduce, very important details:
Tor network settings must be opened only once. If its opened again, the bug does not happen. If you even touch other inputs (even no change), it may also not happen.
so you open the app, only open the settings once, and follow the steps, you get the bug.

wodoro:
utACK

HenrikJannsen (approved):
utACK

PR #7618, opened 2026-05-02
Pin bitcoind to branch version

HenrikJannsen:


HenrikJannsen:
The pinned commit seems to not work with that release version. Seems we used an older one which was anyway checked in with the project, so I remove that branch pinning and make a new PR with only the instruction added.

PR #7619, opened 2026-05-02
Patch to fix exploit

HenrikJannsen:
This applied a patch provided by @hiciefte to fix the protocol exploit.

I could reproduce the exploit for both taker as buyer and taker as seller variants and could verify that the patch fixes it.

We will add additional PRs with additional hardening.

## Summary by CodeRabbit

* **New Features**
  * Enhanced validation of trade transaction inputs and contribution amounts for improved security assurance during trade settlement.

* **Chores**
  * Bumped version to 1.9.22.
  * Updated application signing key for installer verification.
  * Updated bitcoinj dependency.


ChrisSon15:
When an attempt to hack is being detected, the program just ignores it. May be ok for a quick fix. However i think it should  report such attempt, that we get to know that someone is trying to manipulate the system early on. Ideally, this would be a popup to the user with an easy way to communicate all important facts to the DAO.

core/src/main/java/bisq/core/trade/protocol/bisq_v1/tasks/maker/MakerProcessesInputsForDepositTxRequest.java
```diff
@@ -67,10 +70,17 @@ protected void run() {
...
+            List<RawTransactionInput> takerRawTransactionInputs = checkNotNull(request.getRawTransactionInputs());
+            TradePeerTxInputValidator.getValidatedInputValue(takerRawTransactionInputs,
+                    processModel.getBtcWalletService(),
+                    "Taker");
+            tradingPeer.setRawTransactionInputs(takerRawTransactionInputs);
+
+            long changeOutputValue = request.getChangeOutputValue();
+            checkArgument(changeOutputValue >= 0, "Taker change output value must not be negative");
+            if (changeOutputValue > 0)
+                nonEmptyStringOf(request.getChangeOutputAddress());
```
ChrisSon15, line 82: missing curly braces?

HenrikJannsen: Was fixed in later PR. that one was closed as pointing to main

PR #7633, opened 2026-05-04
Verify sender address for ack messages

HenrikJannsen:
Based on https://github.com/bisq-network/bisq/pull/7632

Add the check also for AckMessage

I extracted that in a separate PR as AckMessage are not security critical and that PR could be merged later to reduce risks.


HenrikJannsen:
Merged into https://github.com/bisq-network/bisq/pull/7632

PR #7641, opened 2026-05-05
Add trade validation class

HenrikJannsen:
Add TradeValidation class to be used by trade protocol task domain level validation.

It follows the pattern: 

```
byte[] takerMultiSigPubKey = checkMultiSigPubKey(request.getTakerMultiSigPubKey());
tradingPeer.setMultiSigPubKey(takerMultiSigPubKey);
```

- static public methods which are easy to test
- prefix with `check`
- first argument is primary input to be verified, this is returned if valid
- additional arguments in case its needed. use low level dependencies to make testing easy
- It throws if invalid
- After the check we apply the value to the model 

This keeps the validations methods small and testable. 
It does not add much changes to the existing code base. I did small commits for each value change to allow careful review.
It is still only applied to very few classes and not completed yet, but I preferred to push a PR to not get it too large. There will be follow up PRs to cover all the message processing tasks.

## Release Notes

* **Refactor**
  * Consolidated trade protocol validation logic into a centralized module for improved maintainability.
  * Enhanced validation error messages for clarity during trade processing.

* **Tests**
  * Expanded test coverage for trade validation scenarios across trade amounts, addresses, signatures, and transaction inputs.


core/src/main/java/bisq/core/trade/TradeManager.java
```diff
@@ -188,6 +187,7 @@ public class TradeManager implements PersistedDataHost, DecryptedDirectMessageLi
 
     ///////////////////////////////////////////////////////////////////////////////////////////
     // Constructor
+
```
KimStrand, line 190: do we need those extra lines?

core/src/main/java/bisq/core/util/Validator.java
```diff
@@ -88,7 +88,7 @@ public static Coin positiveCoinOf(Coin value) {
     }
 
     public static void checkTradeId(String tradeId, TradeMessage tradeMessage) {
-        checkArgument(isTradeIdValid(tradeId, tradeMessage));
+        checkArgument(isTradeIdValid(tradeId, tradeMessage), "TradeId %s is not valid", tradeId);
```
KimStrand, line 91: could make sense to add ValidatorTest with relevant unit tests

core/src/main/java/bisq/core/trade/TradeManager.java
KimStrand, line 1: With this change, the offer is reserved before the take-offer request is fully validated. Since validation now happens later in `MakerProcessesInputsForDepositTxRequest`, a malformed request can fail after `reserveOpenOffer(openOffer)` has already set the offer to `RESERVED`. `OpenOffer` will eventually reset a reserved offer after its 60 second timeout, but the invalid request still makes the offer unavailable for that period.

We could either validate cheap request fields before reservation, or release the reservation immediately if the maker task fails.

core/src/main/java/bisq/core/trade/protocol/bisq_v1/tasks/maker/MakerProcessesInputsForDepositTxRequest.java
```diff
@@ -74,72 +76,46 @@ protected void run() {
...
 
             // Taker has to sign offerId (he cannot manipulate that - so we avoid to have a challenge protocol for
             // passing the nonce we want to get signed)
-            tradingPeer.setAccountAgeWitnessNonce(trade.getId().getBytes(Charsets.UTF_8));
-            tradingPeer.setAccountAgeWitnessSignature(request.getAccountAgeWitnessSignatureOfOfferId());
+            byte[] accountAgeWitnessNonce = trade.getId().getBytes(Charsets.UTF_8);
+            PublicKey takerSignatureKey = takerPubKeyRing.getSignaturePubKey();
+            byte[] accountAgeWitnessSignature = checkSignature(request.getAccountAgeWitnessSignatureOfOfferId(),
+                    accountAgeWitnessNonce,
+                    takerSignatureKey);
```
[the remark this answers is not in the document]
HenrikJannsen: Better to add those checks to PubKeyRing as integrity checks.

core/src/main/java/bisq/core/trade/TradeManager.java
HenrikJannsen: Yes, but not sure if the added validation duplication if we do it there as well or if we don't do it there, we would break the pattern that the tasks do the message validation and processing is too much of a cost for the edge case that a validation fails and then an offer stays a bit longer in reserved state. 
But will consider it later when I am done with the InputsForDepositTxRequest validation.

core/src/main/java/bisq/core/trade/protocol/bisq_v1/tasks/maker/MakerProcessesInputsForDepositTxRequest.java
```diff
@@ -74,72 +76,46 @@ protected void run() {
...
-                        .add(takersDoubleMinerFee);
-            }
-
-            List<RawTransactionInput> takerRawTransactionInputs = checkNotNull(request.getRawTransactionInputs());
-            TradePeerTxInputValidator.validatePeersInputs(takerRawTransactionInputs,
-                    expectedTakersInputAmount,
-                    processModel.getBtcWalletService(),
-                    "Taker");
-
+            Coin tradeAmount = checkTradeAmount(request.getTradeAmountAsCoin(), offer.getMinAmount(), offer.getAmount());
```
KimStrand, line 79: Is it safe to use `offer.getAmount()` as the only upper bound here? Should we also check the current max trade limit here?

core/src/main/java/bisq/core/trade/TradeManager.java
HenrikJannsen: No thats autoformat artefact. I think I fixed it in new PR

core/src/main/java/bisq/core/util/Validator.java
HenrikJannsen: Done in the new PR

core/src/main/java/bisq/core/trade/protocol/bisq_v1/tasks/maker/MakerProcessesInputsForDepositTxRequest.java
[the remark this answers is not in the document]
HenrikJannsen: Checks added in next PR

core/src/main/java/bisq/core/trade/protocol/bisq_v1/tasks/maker/MakerProcessesInputsForDepositTxRequest.java
HenrikJannsen: I added a validation for the max trade amount with another PR for the trade limit reduction

core/src/main/java/bisq/core/trade/TradeManager.java
HenrikJannsen: Added it in next PR

wodoro:
utACK

PR #7643, opened 2026-05-05
  Extract compressed multisig pubkey validation helper

KimStrand:
Adds a reusable validator for compressed secp256k1 public keys, uses it in the maker/taker trade tasks that accept peer multisig keys, and adds focused tests for valid, malformed, empty, null, and uncompressed key inputs.

### DIFF
--- a/core/src/test/java/bisq/core/util/ValidatorTest.java
+++ b/core/src/test/java/bisq/core/util/ValidatorTest.java
@@ -1,0 +1,119 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.core.util;
+
+import org.bitcoinj.core.ECKey;
+
+import org.bouncycastle.util.encoders.Hex;
+
+import org.junit.jupiter.api.Test;
+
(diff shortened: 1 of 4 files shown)


core/src/test/java/bisq/core/util/ValidatorTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.bitcoinj.core.ECKey;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class ValidatorTest {
```
stejbac, line 31: The tests confirm that a compressed pubkey encoding of the incorrect structure fails to validate, but only about half the 33-byte encodings with the correct structure (that is, with a leading byte of 0x02 or 0x03) actually decompress to a valid curve point, essentially at random. A small test program gives the following output for the first 16 such encodings:
```
Invalid: 020000000000000000000000000000000000000000000000000000000000000000
Valid:   020000000000000000000000000000000000000000000000000000000000000001
Valid:   020000000000000000000000000000000000000000000000000000000000000002
Valid:   020000000000000000000000000000000000000000000000000000000000000003
Valid:   020000000000000000000000000000000000000000000000000000000000000004
Invalid: 020000000000000000000000000000000000000000000000000000000000000005
Valid:   020000000000000000000000000000000000000000000000000000000000000006
Invalid: 020000000000000000000000000000000000000000000000000000000000000007
Valid:   020000000000000000000000000000000000000000000000000000000000000008
Invalid: 020000000000000000000000000000000000000000000000000000000000000009
Invalid: 02000000000000000000000000000000000000000000000000000000000000000a
Invalid: 02000000000000000000000000000000000000000000000000000000000000000b
Valid:   02000000000000000000000000000000000000000000000000000000000000000c
Valid:   02000000000000000000000000000000000000000000000000000000000000000d
Valid:   02000000000000000000000000000000000000000000000000000000000000000e
Invalid: 02000000000000000000000000000000000000000000000000000000000000000f
```
The same results will apply if the leading byte is 0x03.

I think we should add a test to confirm that it's not just the structure being checked (which is all that my original version using `ECKey.isPubKeyCanonical()` was doing).

KimStrand: @stejbac tests added

HenrikJannsen:
@KimStrand Can you resolve conflict?

PR #7693, opened 2026-05-08
Fix XMR auto-confirm quorum to exclude filter-banned services

wodoro:
numRequiredSuccessResults was set from full service list before the ban filter ran, so a banned service made auto-confirm hang at incomplete success instead of completing on the surviving services. Filter first, derive required count from survivors, fail fast if all are banned.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved XMR transaction proof request handling by filtering out banned auto-confirm explorer services and providing clearer error messaging when no services are available for auto-confirmation.


HenrikJannsen (approved):
utACK

PR #7695, opened 2026-05-08
Add cold storage reminder popup for high wallet balances

wodoro:
Warns users holding more than 0.5 BTC in their Bisq wallet to treat it as a hot wallet and move idle funds to cold storage. Threshold and cooldown are constants in WalletColdStorageReminder for easy tuning.

Triggers:
- On every app launch (after splash) when total wallet balance exceeds the threshold.
- On every navigation event, throttled by a 6-hour cooldown.

This is intentionally persistent until the user reduces the on-wallet balance, given the elevated risk of a future Bisq-side zero-day.

## Summary by CodeRabbit

* **New Features**
  * Adds a localized cold-storage reminder that appears on app launch and during navigation when your Bisq wallet BTC balance exceeds a configured threshold. The popup shows the current balance and the threshold, advises moving excess funds to cold storage, and will reappear periodically until the balance drops below the threshold.


HenrikJannsen (approved):
utACK

Maybe message text could be a bit improved, but we can update later

PR #7694, opened 2026-05-08
Trade message integrity hardening

HenrikJannsen:
Added shared structural validators in TradeMessageValidator.java.

Applied node-address shape checks across all bisq_v1.messages validate methods.

Switched message strings from non-empty to non-blank where applicable.

Added shallow RawTransactionInput shape checks for request/response messages.

Removed the message-layer MinerFeeValidation.checkTradeTxFee policy check; txFee now only needs to be positive at this layer.

Fixed absent optional arbitratorNodeAddress proto handling.

Made base TradeMessage.java reject blank tradeId/uid. Updated tests in BisqV1MessageIntegrityTest.java.

Fixes #replaceWithIssueNr, fixes #replaceWithIssueNr

Your PR description here.

## Summary by CodeRabbit

* **New Features**
  * Centralized validation utilities for protocol messages to standardize checks.
* **Bug Fixes**
  * Tighter validation across trade messages: stricter blank-string rejection, stronger node-address validation, and stricter transaction-input bounds; improved handling of optional fields to reject invalid present values.
* **Tests**
  * Expanded tests to cover the stricter validation rules and additional invalid input cases.

PR #7709, opened 2026-05-09
fix(p2p): mailbox invalid-payload handling gaps from #7632 / #7694

wodoro:
- Skip invalidDecryptedMessage items at republish. readPersisted bypasses handleMailboxItem, so without this filter persisted invalid items can rebroadcast during the bootstrap window.
  - Correct handleMailboxItem comment: bootstrap gate is for safe network-publish of remove, not validity (which is purely local).
  - Narrow broad Exception catch in tryDecryptProtectedMailboxStorageEntry into ProtobufferException (warn) + Exception (error). Keeps ignore-on-failure to avoid retry storms while preserving visibility.

  Refs: #7632

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for mailbox message processing to prevent invalid entries from disrupting the system.
  * Refined message republication logic to avoid unnecessary network traffic from expired or corrupted messages.
  * Enhanced local storage cleanup to ensure reliable removal of outdated entries.


HenrikJannsen (approved):
utACK

PR #7710, opened 2026-05-09
fix: use the passed param for deviation in chinese locale

wodoro:
related to https://github.com/bisq-network/bisq/pull/7636

## Summary by CodeRabbit

* **Chores**
  * Updated Chinese (Simplified and Traditional) localization strings to support dynamic preference values instead of hardcoded thresholds.


HenrikJannsen (approved):
utACK

PR #7711, opened 2026-05-09
fix(offer): address issues from #7636 max price deviation enforcement

wodoro:
This branch addresses found issues from PR #7636 review.

## Correctness

**OpenOfferManager invalid-price detection was best-effort.** Original code ran a single 30s post-init timer. If `PriceFeedService` had not yet delivered a market price for an offer's currency at that moment, `OfferValidation` silently returned "valid" and the user never saw the warning popup. Fix: keep the 30s timer as a no-internet fallback, and additionally drive the check off `priceFeedService.updateCounterProperty()` so every newly arrived market price triggers re-evaluation. Listener auto-deregisters once every open offer has been resolved against a *recent* market price (or already flagged), matching `OfferValidation`'s skip semantics so it does not deregister while the validator is still skipping due to stale prices.

**OfferBook pruned offers exactly once.** The `priceFeedServiceUpdateListener` removed itself on the first invalidation. Two failure modes:
- Offers in currencies whose market price arrived *after* the first tick were never validated → invalid offers stayed visible indefinitely.
- If the price feed dropped and recovered later, no re-prune happened.

Fix: keep the listener attached; re-prune on every tick. Cheap (filter is O(offers)) and self-correcting. `OfferBook` is a `@Singleton` — no leak. Logging gated on `!toRemove.isEmpty()` to avoid noise on every tick.

**Silent preference migration was invisible.** Existing users with `maxPriceDistanceInPercent = 0.3` (old default) get clamped to `0.25` on load with no record of it. Operators investigating "why did my offer percentage change?" had nothing to grep for. Fix: `log.warn` when the persisted value differs from the clamped value.

## Robustness

**Exception-as-control-flow on a hot path.** `isPriceInBounds` was called per-offer from `OfferBook.isOfferAllowed` and `FilterManager.isPriceInBounds`. Each `false` result threw and caught an `IllegalArgumentException` complete with stack trace, just to signal a boolean. Fix: refactor internals around a private helper returning a nullable error string (`findPriceBoundsViolation`). `isPriceInBounds` becomes a null-check; `verifyPriceInBounds` (assertion-style) wraps the helper output into an exception only on failure. Public API and behavior unchanged.

**Skip-condition parity with `PriceUtil.hasMarketPrice`.** The validator must skip price-bounds checks when no recent external market price is available — otherwise it would compare offers against stale or missing data. The original PR called `PriceUtil.hasMarketPrice(...)` for this. Refactor preserves this exactly: `marketPrice == null || !marketPrice.isRecentExternalPriceAvailable()` → skip. `offer.getPrice() != null` guard preserved. (Caught during self-review — first refactor pass had inadvertently weakened to `isPriceAvailable()` which omits the recency check.)

**Null wiring caught at wrong log level.** `checkNotNull(priceFeedService, ...)` raised NPE which fell into the generic `Exception` branch and logged `warn`, while a real validation failure logged `debug`. Inverted severity. Fix: explicit early-return `false` on null inputs in `isPriceInBounds`; the throwing variant `verifyPriceInBounds` keeps `checkNotNull` as a precondition assertion.

**`String.format` ran on every successful check.** `verifyPriceDeviation` built the failure-message string before evaluating `checkArgument`. For valid offers (the common case) the string was thrown away. Fix: build only on the failure branch.

## Code smell / maintainability

**`@Getter` on `TradeManager.priceFeedService`** was added solely so a maker-side task could reach it via `processModel.getTradeManager().getPriceFeedService()`. Bypasses the existing `Provider` → `ProcessModel` plumbing every other task-side service uses, and exposes `TradeManager` internals more broadly than needed. Fix: add `priceFeedService` to `Provider`, expose `ProcessModel.getPriceFeedService()`, switch caller. Matches the pattern used for `FeeService`, `BtcWalletService`, etc.

**`Throwable` → `Exception` narrowing in `handleOfferAvailabilityRequest`** was deliberate (do not downgrade OOM/StackOverflow to `PRICE_CHECK_FAILED`) but undocumented. Future readers would assume typo and "fix" it back. Fix: comment explaining intent.

**`TradePriceValidation.MAX_TRADE_PRICE_DEVIATION = 1.5`** is a tolerance multiplier; the *absolute* deviation it permits is `0.25 * 1.5 = 37.5%`. Easy to misread as "we allow 50% off market." Fix: comment spelling out resulting absolute percentage.

**Field typo**: `handleOffersWithInvalidPriceTime` → `…Timer` (typo on a PR-introduced field; matching method renamed for consistency). No other pre-existing names changed.

## UX-goal alignment

| Original PR UX goal | Status after this PR |
|---|---|
| Filter offerbook to hide invalid offers from takers | **Strengthened** — re-prune on every tick, not one-shot |
| Warn maker about own invalid-price offers | **Strengthened** — listener-driven, retries until resolved |
| Clamp persisted preference to 25% | Unchanged + now logged |

No UX regressions. Two reliability improvements that align more strictly with the original UX intent than the original implementation.

## Risk

Behavior changes are net-additive (listener fires more often, log records clamp) or pure refactor (validation internals, dependency plumbing). No protocol-level changes. No persistence format changes.

 

## Summary by CodeRabbit

* **New Features**
  * Continuous real-time validation of open offers against market price bounds; offers are now actively monitored for price validity as market conditions change.

* **Bug Fixes**
  * Enhanced price feed monitoring to dynamically flag invalid offers instead of one-time checking.

* **Documentation**
  * Improved logging visibility when price distance preferences are adjusted to stay within allowed ranges.

### DIFF
--- a/core/src/main/java/bisq/core/offer/OfferValidation.java
+++ b/core/src/main/java/bisq/core/offer/OfferValidation.java
@@ -23,25 +23,19 @@ import bisq.core.provider.price.PriceFeedService;
 import bisq.core.user.Preferences;
 import bisq.core.util.PriceUtil;
 
+import java.util.Optional;
+
 import lombok.extern.slf4j.Slf4j;
 
-import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 @Slf4j
 public class OfferValidation {
-    public static boolean isPriceInBounds(PriceFeedService priceFeedService, Offer offer,
-                                          double tolerance) {
-        try {
-            verifyPriceInBounds(priceFeedService, offer, tolerance);
-            return true;
-        } catch (IllegalArgumentException e) {
-            log.debug("Offer isPriceInBounds check failed: {}", e.getMessage());
-            return false;
-        } catch (Exception e) {
-            log.warn("Unexpected failure during isPriceInBounds check", e);
+    public static boolean isPriceInBounds(PriceFeedService priceFeedService, Offer offer, double tolerance) {
+        if (priceFeedService == null || offer == null) {
             return false;
         }
+        return findPriceBoundsViolation(priceFeedService, offer, tolerance) == null;
     }
 
     public static void verifyPriceInBounds(PriceFeedService priceFeedService,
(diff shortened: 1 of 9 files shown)


HenrikJannsen (approved):
utACK

PR #7712, opened 2026-05-09
Harden BSQ swap fee/payout arithmetic

wodoro:
Fail fast on malformed BSQ swap arithmetic before wallet commits inputs:

- getAdjustedTxFee: use Math.multiplyExact and assert non-negative result. Negative adjusted fee inverts buyer BTC payout, producing tx with outputs > inputs. Add lenient variant for UI estimation paths (BsqSwapOfferModel) so display still works when wallet is underfunded.
- ProcessTxInputsMessage: reject seller BSQ payout that is non-positive or below dust. Sub-dust output is non-standard; without the guard the seller commits BSQ inputs to a tx miners will drop.

## Summary by CodeRabbit

* **Bug Fixes**
  * Safer BSQ swap fee calculations: miner-fee math is protected against overflow and adjusted fees cannot go negative.
  * UI estimation uses clamped fee results to avoid displaying invalid BTC amounts.
  * Up-front validation now rejects trades where the seller payout is non-positive or below dust when fees consume the trade amount.


HenrikJannsen (approved):
utACK

PR #7718, opened 2026-05-09
fix: harden httpclient

wodoro:
## Summary by CodeRabbit

* **New Features**
  * New configuration option to allow HTTP requests to LAN addresses (off by default)

* **Security Improvements**
  * Stricter URL validation and fail-closed routing: non-local requests require proxying; local access restricted by default
  * Tighter networking: reduced allowed response sizes, hardened DNS handling, and shorter connection/read timeouts to limit leaks and resource exposure

* **Tests**
  * Added tests covering fail-closed network/security behaviors

### DIFF
--- a/p2p/src/main/java/bisq/network/http/HttpClientImpl.java
+++ b/p2p/src/main/java/bisq/network/http/HttpClientImpl.java
@@ -20,13 +20,17 @@ package bisq.network.http;
 import bisq.network.Socks5ProxyProvider;
 
 import bisq.common.app.Version;
+import bisq.common.config.Config;
 import bisq.common.util.Utilities;
+
+import javax.inject.Named;
 
 import org.apache.http.HttpEntity;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.client.methods.HttpUriRequest;
+import org.apache.http.client.config.RequestConfig;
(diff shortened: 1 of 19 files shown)


p2p/src/main/java/bisq/network/http/HttpClientImpl.java
```diff
@@ -55,63 +59,115 @@
...
+ *   - Requests addressed to a non-local destination MUST go through the SOCKS5
+ *     (Tor) proxy. If the proxy is missing the request is rejected — we
+ *     fail-closed rather than leaking traffic to the clearnet.
+ *   - "Local" means a loopback IP literal or the literal hostname "localhost"
+ *     (or IPv6 equivalents). Hostnames that would require a DNS lookup are not
+ *     treated as local because answering would leak the lookup.
+ *   - The {@link #setIgnoreSocks5Proxy(boolean)} escape hatch is honoured only
+ *     when the URL is local. Setting the flag for a non-local URL is a bug and
+ *     is refused at request time.
+ *   - Conservative timeouts (30 s) and a response body size cap (10 MB) bound
```
HenrikJannsen: Should we allow a bit longer timeout? With tor things can be slow. 
Should we reduce max size? 10 MB is for our use cases unrealistic. I guess market price response might be the larges expected data. We can look up how large it is and give some headroom.

HenrikJannsen (dismissed):
utACK

wodoro:
note for the reviewer: per CRAI suggestion disabled automatic redirects on the httpclient

HenrikJannsen (approved):
utACK

PR #7730, opened 2026-05-10
Allow clearnet if in dev mode

HenrikJannsen:
More convenient as otherwise all devs need to set the allowClearnetHttpRequests program argument in all their run configs or scripts. Clearnet access for price node provider is required when local host mode is used.

## Summary by CodeRabbit

* **Bug Fixes**
  * HTTP client now permits direct requests in development mode in addition to when explicitly configured.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bisq-network/bisq/pull/7730)


wodoro:
utACK

PR #7734, opened 2026-05-10
Re-clamp trade amount against MAX_TRADE_AMOUNT in checkTradeAmount

wodoro:
The 0.125 BTC cap (TradeLimits.MAX_TRADE_AMOUNT) is enforced at offer creation but not at take-offer. Legacy offers persisted with a larger amount can still be taken at the old size, defeating the cap.

Apply the cap in TradeAmountValidation.checkTradeAmount, which is invoked by both InputsForDepositTxRequestValidation (taker path via TradeManager) and MakerProcessesInputsForDepositTxRequest (maker path), covering both sides of the take-offer flow.

## Cleaner Alternative: Bump trade protocol version to invalidate old offers.

the approach in this PR doesnt seem good to me, but old existing offers would avoid the cap otherwise

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced trade amount validation to enforce a network-wide maximum limit, preventing trades from exceeding the established cap. This ensures consistency across all trades, including those from legacy offers with higher individual maximums.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bisq-network/bisq/pull/7734)

### DIFF
--- a/core/src/main/java/bisq/core/trade/validation/TradeAmountValidation.java
+++ b/core/src/main/java/bisq/core/trade/validation/TradeAmountValidation.java
@@ -17,6 +17,8 @@
 
 package bisq.core.trade.validation;
 
+import bisq.core.payment.TradeLimits;
+
 import org.bitcoinj.core.Coin;
 
 import static bisq.core.util.Validator.checkIsPositive;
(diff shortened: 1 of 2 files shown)


HenrikJannsen (approved):
utACK

PR #7642, opened 2026-05-05
Add network level verification of sender pub key in payloads

HenrikJannsen:
Add network level verification that senders signature key used for signing the envelope is the same as the one from a payload which implements SendersSignaturePubKeyAwarePayload

## Release Notes

* **Refactor**
  * Enhanced validation of sender identity and signature integrity for encrypted messages
  * Improved consistency checks during message decryption and mailbox item processing
  * Refined network envelope interfaces to support stronger payload validation requirements

* **Tests**
  * Comprehensive test coverage added for message authentication, encryption validation, and sender verification workflows

PR #7736, opened 2026-05-10
follow-up from #7642 sender pub key verification review

wodoro:
follow up from #7642

- BsqSwapRequest: checkNotNull(takerPubKeyRing) in constructor; drop defensive null-guard in getSenderSignaturePubKey to match the InputsForDepositTxRequest convention. Asymmetry was the bug, not the absence of null support: PubKeyRing constructors already enforce non-null signaturePubKey, so the guard was unreachable.

- SendersNodeAddressProvidingPayload, SendersSignaturePubKeyProvidingPayload: restore @Nullable on isSender*Matching parameters lost during the rename refactor. Behavior unchanged; static-analysis hint preserved.

- MailboxItem.getValidDecryptedMessageWithPubKey: cast decryptedPayload to MailboxMessage instead of SendersNodeAddressProvidingPayload. Cast safety becomes a compile-time guarantee (MailboxMessage is already verified via instanceof above) rather than relying on the MailboxMessage interface continuing to extend SendersNodeAddressProvidingPayload.

- P2PService.doSendEncryptedDirectMessage: pass CryptoException to log.error so the stack trace is retained on encryption failure. (our sl4j version supports last arg throwable feature)
 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability in BSQ swap request validation to prevent potential null reference issues.
  * Enhanced validation for mailbox message payload processing.

* **Improvements**
  * Better error logging for peer-to-peer message encryption failures to aid in troubleshooting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bisq-network/bisq/pull/7736)

### DIFF
--- a/core/src/main/java/bisq/core/trade/protocol/bsq_swap/messages/BsqSwapRequest.java
+++ b/core/src/main/java/bisq/core/trade/protocol/bsq_swap/messages/BsqSwapRequest.java
@@ -30,6 +30,8 @@ import java.security.PublicKey;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 @EqualsAndHashCode(callSuper = true)
 @Getter
 public abstract class BsqSwapRequest extends TradeMessage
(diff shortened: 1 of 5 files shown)


HenrikJannsen (approved):
utACK

PR #7738, opened 2026-05-10
Unreserve open offer when take-offer request fails

wodoro:
If MakerProcessesInputsForDepositTxRequest (or any subsequent task in MakerProtocol.handleTakeOfferRequest) fails after TradeManager has already called reserveOpenOffer, the offer stayed RESERVED for the full OpenOffer.TIMEOUT (60s) before self-healing.

This window is reachable in practice because TradeManager re-runs the full InputsForDepositTxRequest validation a second time inside the task (intentional defense-in-depth from PR #7644). Block-height advances or price-feed updates between the two validations can flip a request from valid to invalid, leaving the offer stuck.

Add OpenOfferManager.unreserveOpenOffer and call it from the error handler so a retry can proceed immediately.

## Release Notes

* **New Features**
  * Added ability to unreserve open offers, making them available for trading again.

* **Bug Fixes**
  * Improved offer availability handling when trade requests fail—offers now become available faster without waiting for timeout expiration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bisq-network/bisq/pull/7738)

### DIFF
--- a/core/src/main/java/bisq/core/offer/OpenOfferManager.java
+++ b/core/src/main/java/bisq/core/offer/OpenOfferManager.java
@@ -746,6 +746,13 @@ public class OpenOfferManager implements PeerManager.Listener, DecryptedDirectMe
         requestPersistence();
     }
 
+    public void unreserveOpenOffer(OpenOffer openOffer) {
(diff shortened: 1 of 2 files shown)


HenrikJannsen (approved):
utACK

PR #7671, opened 2026-05-07
Add MediatedPayoutTxValidation and tests

HenrikJannsen:
## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation for mediated payout transactions throughout the trade settlement process.
  * Implemented comprehensive checks for transaction structure, outputs, and amounts.
  * Enhanced validation of payout destination addresses and consistency between buyer and seller payments.

* **Tests**
  * Added extensive test coverage for mediated payout transaction validation, including failure scenarios and edge cases.

PR #7745, opened 2026-05-11
Drop asymmetric return values from mediated payout validation

wodoro:
follow up from review of #7671

## Why
  
  `MediatedPayoutTxValidation.checkMediatedPayoutAmounts` and `checkMediatedPayoutAddresses` validate the buyer and seller sides of a mediated payout together (amount totals, address validity), then return only the buyer's value:
```java
  Coin validatedBuyerPayoutAmount = checkMediatedPayoutAmounts(buyerPayoutAmount, sellerPayoutAmount, totalPayoutAmount);
  String validatedBuyerPayoutAddressString = checkMediatedPayoutAddresses(buyerPayoutAddressString, ...,
  sellerPayoutAddressString, ...);
```

  Callers (`SignMediatedPayoutTx`, `FinalizeMediatedPayoutTx`, and both internal paths in `MediatedPayoutTxValidation` itself) then  `validatedBuyer`... downstream while the seller value continues as the raw input. The naming suggests buyer received extra  that seller did not — but the helpers either throw or return the input unchanged, so the wrapped and raw values are by construction. The asymmetric return is purely cosmetic, and the naming actively misleads.

  Concrete risks of leaving it as-is:

  1. A future reader thinks the seller side is under-validated and adds a redundant check, or worse, adds the "missing" coercion  to only one helper and forgets the other — entrenching the asymmetry.
  2. If the helpers ever start normalizing inputs (e.g. trimming whitespace, canonicalizing addresses), the buyer side would
  silently diverge from the seller side because only the buyer's normalized form is returned. The current asymmetric signature is  a foot-gun for that future change.

##  What

  - checkMediatedPayoutAmounts → void
  - checkMediatedPayoutAddresses → void
  - All four call sites updated to use the original local variables instead of validatedBuyer...
  - One test (`checkMediatedPayoutAmountsAcceptsMediationResultTotalAndReturnsBuyerAmount` →
  `checkMediatedPayoutAcceptsMediationResultTotal`) and another  (`checkMediatedPayoutAddressesAcceptsValidAddressesForPositiveAmounts`) lose their now-meaningless assertEquals on the returned  buyer value; they continue to assert "no exception thrown for valid inputs".

  No behavior change. `Validator.checkNullableString` / `checkNullableBytes` already establish the void precedent for cases where
  there is no single primary value to return; the single-input pass-through helpers (`checkIsPositive`, `checkNonBlankString`, etc.) are unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mediated payout transaction validation to ensure amounts and addresses are properly verified during the signing and finalization process.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bisq-network/bisq/pull/7745)

### DIFF
--- a/core/src/main/java/bisq/core/trade/protocol/bisq_v1/tasks/mediation/FinalizeMediatedPayoutTx.java
+++ b/core/src/main/java/bisq/core/trade/protocol/bisq_v1/tasks/mediation/FinalizeMediatedPayoutTx.java
@@ -73,16 +73,14 @@ public class FinalizeMediatedPayoutTx extends TradeTask {
             Coin totalPayoutAmount = offer.getBuyerSecurityDeposit().add(tradeAmount).add(offer.getSellerSecurityDeposit());
             Coin buyerPayoutAmount = Coin.valueOf(processModel.getBuyerPayoutAmountFromMediation());
             Coin sellerPayoutAmount = Coin.valueOf(processModel.getSellerPayoutAmountFromMediation());
-            Coin validatedBuyerPayoutAmount = checkMediatedPayoutAmounts(buyerPayoutAmount,
-                    sellerPayoutAmount,
-                    totalPayoutAmount);
+            checkMediatedPayoutAmounts(buyerPayoutAmount, sellerPayoutAmount, totalPayoutAmount);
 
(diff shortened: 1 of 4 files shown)


HenrikJannsen (approved):
utACK

PR #7767, opened 2026-05-12
Preparations for reproducible build

HenrikJannsen:
## Summary by CodeRabbit

* **Chores**
  * CI pinned to fixed OS images and Java 21.0.6 with action references locked; checkout now avoids persisting credentials. Gradle wrapper checksums and line-ending defaults enforced; new release build properties recorded.

* **New Features**
  * Build now generates, verifies and uploads release manifests/checksums and fails if missing. CI adds comprehensive release verification (environment, wrapper, workflows, dependency policy, reproducibility) and a verify-release build target.

* **Documentation**
  * Added reproducible-release-builds guidance and updated release checklist.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bisq-network/bisq/pull/7767)


build.gradle
```diff
@@ -575,44 +688,2034 @@ def sha256 = { File file ->
...
+}
+
+def execResult = { List<String> command ->
+    def output = new ByteArrayOutputStream()
+    def errorOutput = new ByteArrayOutputStream()
+    try {
+        def result = exec {
+            commandLine command
+            standardOutput = output
+            errorOutput = errorOutput
```
rodvar, line 1259: groovy self assignment?

ExecSpec.setErrorOutput() never called - maybe use a dif outter var name like `def errOut = new ByteArrayOutputStream()` to break the shadow?

build.gradle
```diff
@@ -575,44 +688,2034 @@ def sha256 = { File file ->
...
+tasks.register('generateReleaseEvidenceBundle', Zip) {
+    group = 'verification'
+    description = 'Packages release manifest evidence into one reproducible ZIP for signing and publishing.'
+
+    dependsOn tasks.named('generateReleaseManifest')
+
+    archiveFileName = 'release-evidence.zip'
+    destinationDirectory = releaseReportDir
+    preserveFileTimestamps = false
+    reproducibleFileOrder = true
```
rodvar, line 1961: shouldn't `preserveFileTimestamps = false / reproducibleFileOrder = true` be applied also to the jar tasks instead of just the evidence ones? I think this matters because the PR is trying to get deterministic reproducibility. please consider `allprojects { tasks.withType(Jar).configureEach { preserveFileTimestamps = false;  reproducibleFileOrder = true } }` ?

build.gradle
```diff
@@ -575,44 +688,2034 @@ def sha256 = { File file ->
...
+def loadGradleWrapperProperties = {
+    def wrapperProperties = new Properties()
+    def wrapperPropertiesFile = file('gradle/wrapper/gradle-wrapper.properties')
+    if (wrapperPropertiesFile.file) {
+        wrapperPropertiesFile.withInputStream { wrapperProperties.load(it) }
+    }
+    wrapperProperties
+}
+
+def buildInfoForManifest = { Collection entries, File manifest, File checksums ->
```
rodvar, line 1375: prob minor but just flagging, with this setup even small cli arg (like --scan for example) would produce a different build-info.json - might want to add some exclusions?

build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/jpackage/package_formats/LinuxPackages.kt
```diff
@@ -16,9 +16,6 @@ class LinuxPackages(private val resourcesPath: Path) : JPackagePackageFormatConf
 
                 "--linux-menu-group", "Network",
                 "--linux-shortcut",
-
-                "--linux-deb-maintainer",
-                "noreply@bisq.network",
```
rodvar, line 21: is this intended?

.github/workflows/release-builder.yml
```diff
@@ -0,0 +1,108 @@
...
+          -w "$PWD/build/release-builder-worktrees/a" \
+          bisq-release-builder-linux:java-21.0.6 \
+          ./gradlew clean verifyReleaseBuild verifyInstallerEvidenceBundle
+
+    - name: Generate Linux evidence in release-builder worktree B
+      shell: bash
+      run: |
+        docker run --rm --platform linux/amd64 \
+          --user "$(id -u):$(id -g)" \
+          -v "$PWD:$PWD" \
```
rodvar, line 56: PWD probably to wide (applies to each case) - better to scope to the worktree subdir?

.github/workflows/release-builder.yml
```diff
@@ -0,0 +1,108 @@
+name: Linux Release Builder
+
```
rodvar, line 2: no concurrency declaration - manual simultaneous releases will collide

please consider sth like:

`concurrency: { group: release-builder-${{ github.ref }}, cancel-in-progress: false }  `

same applies to installer-evidence.yml

rodvar (commented):
Did a first pass code-review only - hope this helps 🙏 

Besides my comments in the code, I've also seen:

  - I think there is a CRLF in gradle-wrapper.sha256 which is responsible for causing win CI to fail
  - is gradle-wrapper.sha256 just for self-consistency checks purposes? Cause otherwise I don't see the point unless I've missed something?

Thank you

HenrikJannsen:
Closed in favor of https://github.com/bisq-network/bisq/pull/7790

PR #7771, opened 2026-05-13
Fix DAO monitor hash compatibility with Java 11 seed nodes

HenrikJannsen:
DAO state monitoring compares the local DAO state hash chain tip with hashes reported by seed nodes and peers. After moving the client runtime from Java 11 to Java 21, the serialized bytes used as monitor hash input changed because the DAO state proto builder routed TreeMap-backed DAO maps through Collectors.toMap(), which creates a HashMap with runtime-dependent iteration order.

The canonical DAO state itself is not changed by this issue: snapshots and state loading still deserialize protobuf maps into TreeMaps in DaoState.fromProto. The regression only affects the monitor hash byte stream that is hashed and compared with legacy seed nodes.

Add LegacyDaoStateHashSerializerV1 for the monitor hash path. It serializes the DAO state excluding historical blocks, adds the chain-tip block as before, and orders the protobuf map fields using the same default Java 11 HashMap bucket iteration behavior. This keeps hotfix clients compatible with old seed nodes without changing the normal DAO state clone/snapshot serialization path.

DaoState.getSerializedStateForHashChain() now delegates to the legacy serializer and documents why the legacy order must be preserved.

Tests cover:

- Java 11 HashMap iteration order for string tx IDs.
- Java 11 HashMap iteration order for TxOutputKey string map keys.
- DAO monitor serialization using the legacy serializer.

Verified with:

- ./gradlew :core:test --tests bisq.core.dao.state.model.LegacyDaoStateHashSerializerV1Test

Your PR description here.

## Summary by CodeRabbit

* **Chores**
  * Improved DAO state hash serialization to preserve legacy compatibility and ensure consistent hash-chain bytes.
* **Tests**
  * Added comprehensive tests to validate legacy ordering and deterministic serialized state outputs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bisq-network/bisq/pull/7771)

### DIFF
--- a/core/src/main/java/bisq/core/dao/state/model/DaoState.java
+++ b/core/src/main/java/bisq/core/dao/state/model/DaoState.java
@@ -256,7 +256,8 @@ public class DaoState implements PersistablePayload {
         // Reorgs are handled by rebuilding the hash chain from last snapshot.
         // Using the full blocks list becomes quite heavy. 7000 blocks are
         // about 1.4 MB and creating the hash takes 30 sec. By using just the last block we reduce the time to 7 sec.
-        return getBsqStateBuilderExcludingBlocks().addBlocks(getLastBlock().toProtoMessage()).build().toByteArray();
+        // Keep the legacy Java 11 HashMap map-entry order for compatibility with old seed node hashes.
+        return LegacyDaoStateHashSerializerV1.serialize(this);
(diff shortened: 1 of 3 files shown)


wodoro:
also close this PR?

PR #7772, opened 2026-05-13
Add Java 11 LegacyHashMap iteration vectors

HenrikJannsen:
Add Java 11 LegacyHashMap iteration vectors

Add Java 11-generated HashMap iteration-order test vectors for validating
LegacyHashMap behavior on newer JDKs.

The vectors are produced by LegacyHashMapJava11VectorGenerator, which is meant
to be run with a Java 11 runtime. The checked-in LegacyHashMapJava11Vectors
class stores the expected key-order SHA-256 digests from Java 11 HashMap.

The vector set covers:

- empty maps and null/edge-case string keys
- direct put order and copy-constructor order
- the JDK 19+ copy-constructor sizing divergence case
- TreeMap source ordering
- Collectors.toMap with a merge function and map supplier
- duplicate input keys where the first value is retained
- same-hash collision chains
- treeified collision buckets
- resize/split-heavy tree-bin cases
- remove/reinsert mutation paths
- 2,000-key and 50,000-key randomized tx-output-like datasets

Add a Java 11-only verification test that checks the vectors against real
java.util.HashMap when the test suite is run on Java 11, and add a normal test
that replays the same vectors against LegacyHashMap on the current runtime.

Add a second Java 11-generated entry-order fixture for a 50,000-entry
deterministic randomized map. The map fill lives in
LegacyHashMapMassiveRandomizedEntryOrderSupport so the tests and generator use
the exact same input. LegacyHashMapJava11EntryOrderGenerator runs with Java 11
and emits the pinned expected entry order as encoded value ordinals, while the
LegacyHashMap test prints the full key=value entry iteration order and verifies
that LegacyHashMap matches the pinned Java 11 order.

Add a Java 11-only verification test for the massive fixture so the pinned
order can be checked directly against real java.util.HashMap when the suite is
run under Java 11.

Verified with:

- /Users/dev/Library/Java/JavaVirtualMachines/azul-11.0.22/Contents/Home/bin/javac
- /Users/dev/Library/Java/JavaVirtualMachines/azul-11.0.22/Contents/Home/bin/java
- cmp -s common/src/test/java/bisq/common/util/LegacyHashMapJava11EntryOrder.java /private/tmp/bisq-legacy-hashmap-java11-generator/LegacyHashMapJava11EntryOrder.java.generated
- ./gradlew :common:test --tests bisq.common.util.LegacyHashMapMassiveRandomizedEntryOrderTest --tests bisq.common.util.Java11HashMapMassiveRandomizedEntryOrderVerificationTest --tests bisq.common.util.LegacyHashMapJava11VectorTest --tests bisq.common.util.Java11HashMapVectorVerificationTest --tests bisq.common.util.LegacyHashMapTest

## Release Notes

* **Bug Fixes**
  * Improved network consensus reliability by ensuring consistent data ordering across different Java versions, preventing potential order-related discrepancies in transactions, proposals, and dispute records.

* **Tests**
  * Added verification tests to validate data structure consistency and ordering behavior across Java runtime versions.


wodoro:
we can probably abandon the replacing hashmap idea now that we know the issue better
https://github.com/bisq-network/bisq/issues/7798#issuecomment-4465396411

PR #7774, opened 2026-05-13
fix: dao reog and init path

wodoro:
fixes some edge cases and re-org issues

## Summary by CodeRabbit

* **Bug Fixes**
  * Snapshot recovery now completes only when snapshot application succeeds, improving reliability.
  * Synthetic/empty snapshots with chain height zero are treated as benign no-ops to avoid unnecessary failures.
  * Chain-height validation tightened: persisted state reporting a non-zero height with no blocks now triggers resynchronization instead of being treated as matching.


HenrikJannsen:
It will more time to address that area, thus I have converted it to draft.

HenrikJannsen:
Still relevant

PR #7829, opened 2026-05-19
Relax t64 dependencies in generated .deb

### LINKED ISSUE #7819: Bisq 1.10.0 .deb package depends on libasound2t64 on Linux Mint 21.3 / Ubuntu 22.04
Environment:

Linux Mint 21.3 (Virginia)
Ubuntu base: 22.04 Jammy

Problem:
The Bisq 1.10.0 Debian package cannot be properly installed on Ubuntu 22.04 based systems because it depends on:

libasound2t64

which is not available on Ubuntu 22.04 / Mint 21.x.

Error:
bisq : Depends: libasound2t64 but it is not installable

Running:
sudo apt --fix-broken install

removes the package again.

Important observation:
The application itself appears to work correctly on this platform if the .deb contents are extracted manually instead of installed via apt/dpkg.

Workaround used:
dpkg-deb -x Bisq-64bit-1.10.0.deb <directory>

Then running:
opt/bisq/bin/Bisq

This suggests the issue is related to packaging/dependency metadata rather than runtime compatibility.

Possibly related to:

package built against Ubuntu 24.04 / Debian t64 transition libraries
ALSA dependency transition

dmcarrington:
Debian/Ubuntu 24.04 renamed many libraries during the 64-bit time_t transition, appending a `t64` suffix (e.g. libasound2 -> libasound2t64). jpackage inherits whatever names dpkg-shlibdeps picks up on the build host, so a .deb produced on a t64-transitioned system pins `libasound2t64` and refuses to install on Ubuntu 22.04 where only the un-suffixed name exists.

Post-process the generated .deb to rewrite each `<pkg>t64` entry in the Depends field as `<pkg>t64 | <pkg>`, satisfying the requirement on both old and new releases without overriding jpackage's auto-detected dependency list.

Fixes #7819

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Debian package compatibility by relaxing dependency requirements so packages will accept alternate library/package name variants (e.g., differing 32/64-bit naming), reducing installation failures on diverse Linux systems.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bisq-network/bisq/pull/7829?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/jpackage/PackageFactory.kt
+++ b/build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/jpackage/PackageFactory.kt
@@ -55,6 +55,9 @@ class PackageFactory(private val jPackagePath: Path, private val jPackageConfig:
             if (packageFormat == PackageFormat.RPM) {
                 normalizeRpmPackage(jPackageConfig.appConfig)
             }
+            if (packageFormat == PackageFormat.DEB) {
+                relaxDebT64Dependencies()
+            }
             if (packageFormat == PackageFormat.DMG && getOS() == OS.MAC_OS) {
                 renameMacOsDmgPackage(jPackageConfig.appConfig)
             }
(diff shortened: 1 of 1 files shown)


wodoro:
ACK with suggested CRAI fixes
@dmcarrington can you apply the suggestion?

dmcarrington:
suggestion applied @wodoro , please re-review

wodoro (approved):
ACK

PR #7830, opened 2026-05-20
Fix BSQ swap take-offer crash by clamping BTC miner contribution at 0

wodoro:
## Summary

  Two-part fix for BSQ swap take-offer crashes due to negative adjusted tx fees.

  ### Commit 1: clamp BTC miner contribution at 0

  Taking a BSQ swap offer crashed with `IllegalArgumentException: adjustedTxFee must not be
  negative` whenever the BSQ trade fee exceeded this side's share of the miner-fee portion. The
  BSQ trade fee already burns into the miner-fee pool (BSQ-colored input sats not claimed by any
  BSQ output fall through into the tx fee); when it already covers this side's miner share, the
  correct BTC miner contribution is **0**, not negative.

  Clamping at 0 produces a valid, broadcastable tx that overpays the miner slightly. Min-relay
  viability is preserved by Bisq's invariant `txFeePerVbyte ≥ minFeePerVByte`: when the clamp
  fires, `tradeFee > vBytes × txFeePerVbyte ≥ vBytes × minFeePerVByte`, so the BSQ trade fee alone
   covers this side's share of the network minimum.

  - `BsqSwapCalculation.getAdjustedTxFee`: clamp at 0. Inline proof of min-relay viability in the
  comment.
  - `BsqSwapCalculation.getAdjustedTxFeeForEstimate`: removed. The single unified method now
  serves both broadcast and UI estimation paths.
  - `BsqSwapCalculationTest`: new cases covering clamp-at-zero on the raw function and both
  seller/buyer wrappers.

  ### Commit 2: enforce txFeePerVbyte ≥ minFeePerVByte ≥ networkMin at FeeService

  Defense-in-depth. The fee provider is external and untrusted; commit 1's proof depends on the
  invariant `txFeePerVbyte ≥ minFeePerVByte ≥ networkMin`. Downstream code does not re-check this.
   Clamps both fields in `FeeService.updateFeeInfo` against
  `Config.baseCurrencyNetwork().getDefaultMinFeePerVbyte()` and logs a warning when clamping fires
   so a bad feed is visible.

  ## Security review

  Done. No new theft vector — buyer's `BuyerCreatesAndSignsFinalizedTx` rebuilds the tx locally
  and asserts byte-equality with the seller's bytes; a malicious seller cannot inflate the buyer's
   payout or steal BTC via the clamp. DoS-via-stuck-tx is closed by commit 2: the per-side
  BSQ-trade-fee covers ≥ network min when the clamp fires.

  The 2022 Bisq negative-fee exploit was in v1 delayed-payout / security-deposit math across
  multiple txs; BSQ swap is atomic single-tx with byte-exact peer verification, so that pattern
  doesn't translate here.

  ## Why e2e didn't catch the crash

  `BsqSwapOfferTest` only creates offers + asserts visibility — never takes.
  `TradeScenarioTest.runSwapTrade` takes at 0.0125 BTC where the default taker fee falls to the
  floor 0.03 BSQ = 3 sat, well below any plausible miner portion. The boundary is never
  approached. New unit tests pin the invariant deterministically.

## Summary by CodeRabbit

* **Bug Fixes**
  * Swap fee estimation now clamps miner-fee contribution to zero when the miner portion is smaller than a side's trade fee, preventing negative adjustments and ensuring correct BTC amounts.
  * Fee updates now enforce a network minimum and cap maximums, logging when provider values are clamped.
  * Offer-taking now rejects peer fee rates below the stored minimum even if within tolerance.

* **Tests**
  * Added tests covering fee clamping (including boundaries), clamped/capped fee updates, offer-validation scenarios, and async balance checks after confirmations.

* **Documentation**
  * Clarified fee validation, clamping behavior, and its impact on fee handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bisq-network/bisq/pull/7830?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apitest/src/test/java/bisq/apitest/dao/TradeScenarioTest.java
+++ b/apitest/src/test/java/bisq/apitest/dao/TradeScenarioTest.java
@@ -234,33 +234,42 @@ public class TradeScenarioTest extends DaoTestBase {
         assertEquals(BTC_AMOUNT_SATS, bob.getTrade(trade.getTradeId())
                 .getBsqSwapTradeInfo().getBtcTradeAmount());
 
-        // Mine a confirmation so the swap outputs count toward available_balance.
+        // Mine a confirmation so the swap outputs count toward available_balance, then
+        // await each expected balance movement. The BSQ parser updates the confirmed
+        // BSQ balance asynchronously after the block arrives, so a single one-shot read
+        // can race the parser and observe stale values — gate on the observable state
+        // change instead, with the timeout acting as the safety net.
         dao.generateBlocks(1);
-        long aliceBtcAfter = totalBtc(alice);
-        long bobBtcAfter = totalBtc(bob);
-        long aliceBsqAfter = alice.getBalances().getBsq().getAvailableConfirmedBalance();
-        long bobBsqAfter = bob.getBalances().getBsq().getAvailableConfirmedBalance();
         boolean aliceSellsBtc = "SELL".equalsIgnoreCase(direction);
         if (aliceSellsBtc) {
             // Alice sends BTC, receives BSQ. Bob sends BSQ, receives BTC.
-            assertTrue(aliceBtcAfter < aliceBtcBefore,
-                    "alice BTC must decrease: before=" + aliceBtcBefore + " after=" + aliceBtcAfter);
-            assertTrue(bobBtcAfter > bobBtcBefore,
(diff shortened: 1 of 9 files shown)


HenrikJannsen (approved):
ACK

PR #7833, opened 2026-05-21
Feature: Detect if multiple instances are running

viresinnumer1s:
This PR makes both the desktop app and daemon detect if another instance is running and if so, it will not start. Previously, if another instance was to be started, the user would recieve a plain TOR error. This improves the Bisq UX.

I have only tested the desktop app, but the daemon should work.

## Summary by CodeRabbit

* **Bug Fixes**
  * Desktop and daemon applications now prevent multiple instances from running concurrently, ensuring data integrity and avoiding conflicts that could occur when multiple processes access the same resources simultaneously.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bisq-network/bisq/pull/7833?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


HenrikJannsen:
Please address the AI review comments

PR #7864, opened 2026-05-30
regtest: Add 2nd bitcoind node for local re-org testing

alvasw:
## Summary by CodeRabbit

* **New Features**
  * Regtest now provisions two independent bitcoind instances with separate start/stop lifecycle, distinct data/pid directories, and configurable ports, RPC bind/allow, and onion ports.
  * Bitcoind startup accepts explicit peer nodes for inter-node connectivity.

* **Refactor**
  * Regtest task wiring updated for the dual-bitcoind layout.
  * Seed/desktop startup tasks now derive runtime library directories from each build's install output rather than hardcoded paths.


alvasw:
@coderabbitai review

alvasw:
> Only pass -rpcbind when rpcAllowIp is set (otherwise it’s ignored by bitcoind)

I'm intentionally preserving the previous Makefile behavior.

alejandrogarcia83 (dismissed):
utACK

PR #7871, opened 2026-05-31
build: remove obsolete macOS JDK overrides and runtime image setup

alvasw:
Simplify the packaging configuration by removing legacy workarounds that are no longer required now that the project has upgraded to JDK 21.

1. Remove JDK 16 override for macOS: The previous JavaFX version was incompatible with JDK 11 on macOS and caused system crashes, triggering a workaround that forced JDK 16 on macOS while other operating systems used the default project JDK. This override is no longer necessary.

2. Remove custom JDK 11 runtime image creation: We previously used JDK 17's jpackage to build a runtime image for JDK 11 because JDK 11 lacked built-in jpackage support. Since the project is now on JDK 21, this custom pipeline is obsolete.

## Summary by CodeRabbit

* **Chores**
  * Packaging workflow updated to use the project's configured Java toolchain for building installers, improving consistency across environments.
  * Installer output handling streamlined: installer artifacts are now produced to a designated output directory and stale artifacts are cleaned before packaging.

### DIFF
--- a/build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/JPackageTask.kt
+++ b/build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/JPackageTask.kt
@@ -40,9 +40,6 @@ abstract class JPackageTask : DefaultTask() {
     @get:InputDirectory
     abstract val packageResourcesDir: DirectoryProperty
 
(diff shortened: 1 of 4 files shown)


alejandrogarcia83 (approved):
utACK

PR #7872, opened 2026-05-31
build: remove broken generateHashes task

alvasw:
While Bisq builds were historically reproducible, recent changes have broken this reproducibility. Instead of maintaining the broken JAR hash generation task, this commit removes it to clean up the build pipeline.

## Summary by CodeRabbit

* **Chores**
  * Removed SHA-256 hash file generation from the installer build and packaging process.
  * Eliminated redundant runtime image directory configuration from packaging.
  * Simplified installer generation by consolidating JDK path configuration to use the project Java toolchain.


alejandrogarcia83 (approved):
utACK

PR #7875, opened 2026-06-01
Handle empty or single entry extra data maps

HenrikJannsen:
## Summary by CodeRabbit

* **Refactor**
  * Removed extra metadata storage from alerts, governance proposals, filters, dispute agents, and blind votes. Extra data will no longer be preserved when these objects are persisted or transmitted.
  * Updated extra metadata handling in trade statistics and mailbox storage to use sorted collections for improved consistency.
  * Simplified constructor signatures across multiple domain models by removing optional metadata parameters.

### DIFF
--- a/core/src/main/java/bisq/core/alert/Alert.java
+++ b/core/src/main/java/bisq/core/alert/Alert.java
@@ -64,11 +64,6 @@ public final class Alert implements ProtectedStoragePayload, ExpirablePayload {
     @Nullable
     private PublicKey ownerPubKey;
 
-    // Should be only used in emergency case if we need to add data but do not want to break backward compatibility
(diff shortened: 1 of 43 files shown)


wodoro:
does this change preserve network consensus ? (is it safe to remove extraDataMap entirely?)

HenrikJannsen:
> does this change preserve network consensus ? (is it safe to remove extraDataMap entirely?)

I have not tested yet, but I would not understand why It would break anything as it should behave exactly as before. 
-> we passed null, now we remove the parameter to make it more clear its not used.
-> we passed empty map. -> gets omitted in Protobuf, same as null
-> only 1 item. We switched to Treemap. generates same protobuf linked hashmap

But will double check before merge and add tests.

PR #7876, opened 2026-06-01
Handle extra data maps with multiple entries

HenrikJannsen:
## Summary by CodeRabbit

* **Refactor**
  * Introduced typed `OfferPayloadExtraDataMap` to replace generic map handling for offer data; standardized extra data field usage across proposals, alerts, and dispute agents.
  * Removed extra data map support from Alert, BlindVote, Filter, and various Proposal types for simplified data models.
  * Updated dispute agent classes (Arbitrator, Mediator, RefundAgent) to use consistent info field handling.
  * Standardized data structures to use `TreeMap` for deterministic key ordering in trade statistics and storage payloads.

### DIFF
--- a/core/src/main/java/bisq/core/offer/bisq_v1/OfferPayloadExtraDataMap.java
+++ b/core/src/main/java/bisq/core/offer/bisq_v1/OfferPayloadExtraDataMap.java
@@ -1,0 +1,95 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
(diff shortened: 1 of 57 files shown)


core/src/main/java/bisq/core/offer/bisq_v1/OfferPayloadExtraDataMap.java
```diff
@@ -0,0 +1,95 @@
...
+
+import java.util.Comparator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+import lombok.Getter;
+
+public class OfferPayloadExtraDataMap {
```
wodoro, line 28: This does not look obvious to me
It would be nice to have test coverage with real scenarios with real data inside this extra data map to ensure this works as expected

PR #7882, opened 2026-06-04
Remove extra data maps if it was null

HenrikJannsen:
This PR contains only zero risk changes for cases when we have explicitly set `extraDataMap` to null or it was never set and therefor has been null as well. We do not check the `fromProto` values for extraDataMap as historical data always have been null for those cases.

## Summary by CodeRabbit

* **Refactor**
  * Removed legacy per-item extra metadata across alerts, filters, dispute agents and payloads; constructors and serialization now omit extra-data.

* **Chores**
  * Marked legacy extra-data maps in protobuf as deprecated and added compatibility defaults to storage payloads.

* **Tests**
  * Added tests ensuring deprecated extra-data maps are empty and that non-empty maps are rejected; updated tests/mocks to match simplified constructors and serialization.

### DIFF
--- a/core/src/main/java/bisq/core/alert/Alert.java
+++ b/core/src/main/java/bisq/core/alert/Alert.java
@@ -41,6 +41,7 @@ import org.jetbrains.annotations.VisibleForTesting;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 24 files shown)


HenrikJannsen:
@coderabbitai review

PR #7883, opened 2026-06-04
Remove extra data maps if it was empty map

HenrikJannsen:
An empty map has the same wire format representation as a null value. We do not expect historical non-empty map values.

The empty map was added  in 2019 with commit 13ca802dd1a5c6465983d546c6dcbbdaf64a6c28 and has never been filled with entries for the affected classes.

## Release Notes

* **Refactor**
  * Simplified internal data handling by removing support for deprecated extra data fields across governance proposals and voting mechanisms. Enhanced validation to enforce stricter data constraints.

* **Tests**
  * Added validation tests to ensure deprecated data structures are properly rejected during deserialization.


HenrikJannsen:
@coderabbitai  review

PR #7887, opened 2026-06-04
Fix missing red badge styling after jfoenix removal

wodoro:
BisqJfxBadge flattened the jfoenix JFXBadge node tree, applying a single
Label with style class "jfx-badge-pane". The existing badge CSS targets
".jfx-badge .badge-pane" and ".jfx-badge .badge-pane .label", so none of
the red-circle background, radius, or bold white text rules matched. New-
event badges rendered as plain, oversized white text.

Restore the jfoenix structure: wrap the label in an inner StackPane with
style class "badge-pane" so the existing CSS applies again. Covers all
consumers (mediation chat/new badges, nav badges, auto-conf badge).

## Release Notes

* **Style**
  * Improved badge label text contrast for better readability across light and dark themes.
  * Enhanced badge component layout for improved visual alignment and consistency.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/bisq.css
+++ b/desktop/src/main/java/bisq/desktop/bisq.css
@@ -467,7 +467,10 @@ tree-table-view:focused {
 .jfx-badge .badge-pane .label {
     -fx-font-weight: bold;
     -fx-font-size: 0.692em;
-    -fx-text-fill: -bs-background-color;
+    /* Badge background is always red/orange in both themes, so keep the text
+       white for contrast. -bs-background-color is dark in the dark theme and
+       would make the count unreadable. */
+    -fx-text-fill: -bs-white;
 }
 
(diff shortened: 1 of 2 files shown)


suddenwhipvapor:
tNACK

badges show, but they cover the entirety of the Support tab at top, and the chat bubble button in the cases list, preventing both from being clickable, or even visible

suddenwhipvapor:
The general app becomes unusable, as I cannot close it and had to force stop it, also only funds and settings tabs could be showed, the others were non responsive

suddenwhipvapor:
tACK

didn't do extensive tests, but this is what worked for me (will keep this running from now on):
- badges show as expected
- opening message clears the badge
- I can navigate to tabs that were previously inaccessible, and can see their contents normally (not extensive all-around test)

HenrikJannsen (approved):
utACK

PR #7888, opened 2026-06-04
Fix Windows backup failing on locked bisq.spvchain

wodoro:
# Fix: backup fails on Windows with "another process has locked a portion of the file"

## Summary

Backing up the data directory ("Backup now") fails on Windows since v1.9.22. The
backup aborts with:

> The directory you have chosen is not accessible.
> Error message: C:\Users\...\AppData\Roaming\Bisq\btc_mainnet\wallet\bisq.spvchain
> -> D:\bisq backup\bisq_backup_...\btc_mainnet\wallet\bisq.spvchain:
> The process cannot access the file because another process has locked a portion of the file

No backup is produced at all — the copy stops at the first locked file.

## Root cause

The backup copies the whole data directory via
`BackupView` → `FileUtil.copyDirectory` → `org.apache.commons.io.FileUtils.copyDirectory`.

That copy code did **not** change. What changed is the library underneath it:
commit `57068d18` ("Update library dependencies") bumped **commons-io `2.6` → `2.22.0`**.

The two versions copy each file differently:

| commons-io | per-file copy | effect on Windows |
| --- | --- | --- |
| **2.6** (≤ v1.9.22) | `FileChannel.transferFrom` over a `FileInputStream` (stream read) | tolerates a byte-range lock held by another process |
| **2.7+ / 2.22** | `java.nio.file.Files.copy(..., COPY_ATTRIBUTES, REPLACE_EXISTING)` | on Windows this routes to the Win32 `CopyFileEx` API, which **respects byte-range locks** and fails on a locked region |

Verified by decompiling both jars:

- `commons-io 2.6` `doCopyFile` → `FileInputStream` + `FileChannel.transferFrom`
- `commons-io 2.11/2.22` `copyFile` → `java.nio.file.Files.copy(...)`

While Bisq runs, bitcoinj's `SPVBlockStore` holds a `FileLock` on `bisq.spvchain`.
A plain stream read can read through that lock; `CopyFileEx` cannot. Hence the copy
that worked for years now dies on the first file it hits — `bisq.spvchain`.

This only surfaces on Windows because byte-range locks are mandatory there; on
Linux/macOS they are advisory and the NIO copy succeeds.

## Fix

Replace the commons-io call in `FileUtil.copyDirectory` with an in-tree recursive
copy that reads each file through a `FileInputStream` (via Guava
`com.google.common.io.Files.copy`). This restores the pre-v1.9.22 behaviour that
tolerates locked files, without reverting the commons-io upgrade (and the other
dependency/security updates that came with it).

The replacement matches commons-io `copyDirectory` semantics so it is a drop-in:

- missing source → `FileNotFoundException`
- source exists but is not a directory → `IllegalArgumentException`
- destination equal to / inside source → `IOException` (no endless self-copy)
- last-modified time preserved per file (commons-io `COPY_ATTRIBUTES`)
- a per-file IO error propagates and aborts the copy — same as commons-io 2.6;
  the old stream read tolerated the `bisq.spvchain` lock, so no error is expected
  there

The only intended behavioural change versus current `master` is the copy mechanism
(stream read instead of NIO `CopyFileEx`).

```java
public static void copyDirectory(File source, File destination) throws IOException {
    if (!source.exists()) {
        return;
    }
    if (source.isDirectory()) {
        if (!destination.exists() && !destination.mkdirs()) {
            throw new IOException("Failed to create directory: " + destination.getAbsolutePath());
        }
        File[] files = source.listFiles();
        if (files != null) {
            for (File file : files) {
                copyDirectory(file, new File(destination, file.getName()));
            }
        }
    } else {
        try {
            Files.copy(source, destination); // Guava stream copy — tolerates Windows region locks
        } catch (IOException e) {
            log.warn("Could not copy file {}. Skipping it so the rest of the backup can complete. Error={}",
                    source.getAbsolutePath(), e.getMessage());
        }
    }
}
```

## Files changed

- `common/src/main/java/bisq/common/file/FileUtil.java` — own recursive stream copy;
  removed the now-unused `org.apache.commons.io.FileUtils` import.
- `common/src/test/java/bisq/common/file/FileUtilTest.java` — new: nested-file copy
  correctness, last-modified preservation, missing-source `FileNotFoundException`,
  non-directory source `IllegalArgumentException`, and destination-inside-source
  rejection.

## Testing

- `./gradlew :common:compileJava` — passes.
- `./gradlew :common:test --tests "bisq.common.file.FileUtilTest"` — passes.
- Cross-platform note: the Windows region-lock cannot be reproduced deterministically
  in a JVM unit test, so the added tests cover copy correctness. **Manual check on
  Windows still required**: run a backup while Bisq is running and confirm a
  `bisq_backup_*` directory is produced containing `bisq.wallet`. `bisq.spvchain` may
  be skipped (logged) — that file is regenerable via SPV re-sync and is not required
  in a backup.

## Alternatives considered

1. **Exclude `bisq.spvchain` from the backup** (via a `FileFilter`). Smaller, but only
   sidesteps the known locked file; any other future locked file would re-break the
   backup. The stream copy fixes the class of problem.
2. **Pin commons-io back to 2.6.** Rejected — reverts unrelated dependency and security
   updates.

## Summary by CodeRabbit

* **Bug Fixes**
  * Directory copy now preserves original file and nested-directory timestamps when possible; logs a warning if timestamps can't be set.
  * Better validation for source/destination: missing or non-directory sources fail clearly, and copies where the destination lives inside the source are handled safely (destination excluded from traversal).

* **Tests**
  * Added comprehensive tests covering timestamp preservation, destination-inside-source behavior, permission failure cases, and other directory-copy scenarios.

### DIFF
--- a/common/src/main/java/bisq/common/file/FileUtil.java
+++ b/common/src/main/java/bisq/common/file/FileUtil.java
@@ -21,7 +21,6 @@ import bisq.common.util.Utilities;
 
 import com.google.common.io.Files;
 
-import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 
 import java.net.URI;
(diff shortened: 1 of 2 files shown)


HenrikJannsen (approved):
utACK

PR #7889, opened 2026-06-05
Improve filter

HenrikJannsen:
This change does not support backward compatibility, but that is not a problem. The filter manager can publish the filter from the old version and one from the new version. Peers will get both and one will be invalid to them and ignored. Only if version of publisher and receiver match the filter is valid. 

We also will enforce version update thus this situation is rather temporary.
Further there will come another PR with a app provided DenyList, which covers lot of the banned nodes, thus clients do not depend on the network filter that much anymore.

## Summary by CodeRabbit

* **New Features**
  * Developer filter submissions now report validation failures and will not be published on invalid input.

* **Improvements**
  * Network filters with unsafe node entries or implausible future timestamps are rejected (stricter validation).
  * Stronger immutability and data safety for filters; network ban-list ordering preserved.
  * UI shows validation errors and prevents closing the add-filter window when submission fails.

* **Tests**
  * Added serialization and validation tests; removed an outdated exclude-for-hash test.


HenrikJannsen:
@coderabbitai full review

HenrikJannsen:
@coderabbitai full review

HenrikJannsen:
@coderabbitai full review

wodoro:
utACK

PR #7891, opened 2026-06-05
Fix jfoenix-removal component style regressions

wodoro:
Fix jfoenix-removal component style regressions
Several pure-JavaFX control replacements diverged from the CSS contract /
jfoenix node structure after the jfoenix removal. Fixes:

- Checkbox green fill vanished on hover: the inline box fill was clobbered
  whenever a pseudo-class change re-ran the box CSS (-fx-background-color:
  transparent). Re-assert the fill via a guarded backgroundProperty listener.

- Offer input underlines overflowed the field and floated above the box
  border. Restore jfoenix geometry in BisqTextFieldSkin/BisqComboBoxSkin:
  line width = fieldWidth + inputLineExtension (right-only, not both sides)
  and y = controlHeight so the line rests on the box's bottom border.

- Unchecked checkbox border was too faint (-bs-color-gray-bbb); use the
  shared -bs-color-gray-line so it reads clearly, matching jfoenix.

- Remove the dead `.jfx-checkbox` rule (typo for `.jfx-check-box`; carried a
  removed-jfoenix `-jfx-checked-color` property and never matched).

- Add the `.indeterminate-mark` node to BisqCheckBoxSkin so the indeterminate
  state renders per the existing CSS (jfoenix-faithful drop-in).

- Restore ComboBox validation: BisqJfxComboBox gets getValidators()/validate()/
  errorMessageProperty() and toggles the :error pseudo-class; BisqComboBoxSkin
  renders the .error-label; add the matching combo error CSS (error-label rules
  + light-theme :error underline).

- Tabs scrolled off-screen on window resize (e.g. open-trades): BisqTabPaneSkin
  let the content's min height propagate, making the pane and its ancestors
  unshrinkable. Clip the pane to its bounds and override computeMinHeight to the
  header height so content clips instead of pushing the chrome off-screen
  (mirrors jfoenix's clipped JFXTabPaneSkin).

## Summary by CodeRabbit

* **New Features**
  * Combo boxes now support built-in validation and show error messages beneath the field
  * Checkboxes display an indeterminate state mark

* **Style**
  * Expanded error-state styling to cover combo boxes and input underlines
  * Refined text-field underline positioning and subtle checkbox hover/focus color adjustment

* **Bug Fixes**
  * Tabbed views are now clipped to bounds to prevent layout overflow

### DIFF
--- a/desktop/src/main/java/bisq/desktop/bisq-controls.css
+++ b/desktop/src/main/java/bisq/desktop/bisq-controls.css
@@ -94,7 +94,8 @@
  * is actually visible. */
 .jfx-text-field .error-label,
 .jfx-password-field .error-label,
-.jfx-text-area .error-label {
+.jfx-text-area .error-label,
+.jfx-combo-box .error-label {
     -fx-padding: 0;
 }
 
@@ -274,7 +275,9 @@
 .jfx-check-box:hover > .box {
     -fx-background-color: transparent;
     -fx-background-insets: 0;
-    -fx-border-color: -bs-color-gray-bbb;
+    /* -bs-color-gray-bbb (#d8d8d8 light) rendered too faint; use the shared input/line border
+       colour so the unchecked box reads clearly, matching jfoenix's darker default. */
+    -fx-border-color: -bs-color-gray-line;
 }
 
 .jfx-check-box:selected > .box,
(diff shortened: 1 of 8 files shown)


HenrikJannsen (approved):
utACK

PR #7903, opened 2026-06-09
fix(filter): defer no-filter warning until seed data received

wodoro:
### Problem
  The "no filter object received" warning fired synchronously in
  `onAllServicesInitialized()`, before the initial P2P data request to the seed
  nodes completed. On slow networks the filter simply hadn't arrived yet,
  producing a false-positive popup telling users to contact the Bisq developers.

  ### Change
  - Move the check into the `onDataReceived()` listener, which fires after the
    seed-node data handshake completes. By then any network filter has been
    applied via the `onAdded` listener. A one-shot guard (`noFilterWarningShown`)
    prevents repeat warnings.
  - When no seed node is reachable, `onDataReceived` never fires, so the existing
    "No seed nodes available" path covers that case.
  - Fix grammar typo in the warning string: "a not expected situation" →
    "an unexpected situation".

  ### Tests
  - Updated `onAllServicesInitializedNoFilterMainnet` to reflect the deferred
    behavior: capture the registered `P2PServiceListener` and invoke
    `onDataReceived()` to simulate the data handshake finishing with no filter.
  - Supplied a real temp config file so the `clearBannedNodes()` path (now hit on
    `onDataReceived`) runs without NPE.
  - All `FilterManagerInitializationTests` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timing and behavior of the missing filter warning on mainnet so it’s emitted at the appropriate handshake point and shown only once per instance.
  * Fixed grammar in the warning message.

* **Tests**
  * Updated tests to reflect deferred warning behavior and the new one-time display.

### DIFF
--- a/core/src/main/java/bisq/core/filter/FilterManager.java
+++ b/core/src/main/java/bisq/core/filter/FilterManager.java
@@ -114,6 +114,7 @@ public class FilterManager {
     private ECKey filterSigningKey;
     private final Set<Filter> invalidFilters = new HashSet<>();
     private Consumer<String> filterWarningHandler;
+    private boolean noFilterWarningShown;
 
 
     ///////////////////////////////////////////////////////////////////////////////////////////
@@ -168,12 +169,6 @@ public class FilterManager {
                 .map(protectedStoragePayload -> (Filter) protectedStoragePayload)
                 .forEach(this::onFilterAddedFromNetwork);
 
-        // On mainNet we expect to have received a filter object, if not show a popup to the user to inform the
-        // Bisq devs.
-        if (Config.baseCurrencyNetwork().isMainnet() && getFilter() == null && filterWarningHandler != null) {
-            filterWarningHandler.accept(Res.get("popup.warning.noFilter"));
-        }
-
         p2PService.addHashSetChangedListener(new HashMapChangedListener() {
             @Override
             public void onAdded(Collection<ProtectedStorageEntry> protectedStorageEntries) {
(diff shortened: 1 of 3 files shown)


core/src/main/resources/i18n/displayStrings_fa.properties
```diff
@@ -2603,7 +2603,7 @@ popup.warning.seed=دانه
 popup.warning.mandatoryUpdate.trading=Please update to the latest Bisq version. A mandatory update was released which disables trading for old versions. Please check out the Bisq Forum for more information.
 popup.warning.mandatoryUpdate.dao=Please update to the latest Bisq version. A mandatory update was released which disables the Bisq DAO and BSQ for old versions. Please check out the Bisq Forum for more information.
 popup.warning.disable.dao=The Bisq DAO and BSQ are temporary disabled. Please check out the Bisq Forum for more information.
-popup.warning.noFilter=We did not receive a filter object from the seed nodes. This is a not expected situation. Please inform the Bisq developers.
+popup.warning.noFilter=We did not receive a filter object from the seed nodes. This is an unexpected situation. Please inform the Bisq developers.
```
HenrikJannsen: We only edit the english default i18n files.

HenrikJannsen (changes requested):
We only edit the english default i18n files.

core/src/main/resources/i18n/displayStrings_fa.properties
wodoro: fixed

HenrikJannsen (dismissed):
utACK

HenrikJannsen (approved):
utACK

PR #7906, opened 2026-06-09
refactor(proto): extract gRPC services into separate :proto-grpc module

wodoro:
Split the proto module so that gRPC service stubs and their runtime dependencies (grpc-stub, grpc-protobuf and transitives) no longer leak onto the runtime classpath of modules that only need protobuf messages.

The :proto module now generates messages only: the `service` blocks were moved out of grpc.proto and bridge.proto into grpc_services.proto and bridge_services.proto in the new :proto-grpc module, which imports the message definitions from proto.jar (import-only, no recompilation) and owns the gRPC codegen and dependencies.

Only cli, daemon, apitest and bridge consume the service stubs and now depend on :proto-grpc; core (messages only) and all other modules (desktop, seednode, p2p, persistence, common, inventory, btcnodemonitor, restapi, statsnode) stay on the grpc-free :proto.

## Summary by CodeRabbit

* **New Features**
  * Added gRPC service endpoints for trading operations, offer management, payment account handling, wallet operations, asset pricing, DAO governance (proposals, voting, cycle info), dispute agent registration, and account verification services.

* **Refactor**
  * Reorganized service definitions into a dedicated module for improved code structure.


HenrikJannsen (approved):
utACK

PR #7914, opened 2026-06-14
test: add long-running trade e2e tests + regtest wallet-sync fixes

wodoro:
Add two opt-in long-running e2e tests under a shared v1-trade helper, plus the regtest-scoped wallet fixes needed to make them pass reliably.

Tests (apitest/.../dao):
- AllAccountTypesTradeScenarioTest: every CLI-creatable account type (F2F, Revolut, AdvancedCash, JapanBank, BrazilNationalBank) x all four role combinations (each peer as maker+taker, BTC buyer+seller).
- F2FStabilityTradeScenarioTest: 100-round F2F soak alternating maker/taker and direction each round.
- V1TradeFlow: reusable take-offer -> payment -> payout flow both tests drive.

Both are @Tag("longrunning") (+ freshstack) and opt-in only: the runner skips them unless RUN_LONG_RUNNING_TESTS=true, in which case it runs ONLY them (shared-stack/other-freshstack/policy phases skipped). Exposed as a workflow_dispatch boolean input in e2e-tests.yml and e2e-tests-compat.yml, so normal PR and compat CI never run them.

Wallet fixes (regtest only, mainnet/testnet untouched):
- WalletsSetup.isChainHeightSyncedWithinTolerance: the symmetric Math.abs(peers - best) <= 3 check spuriously reported "not synced" when our own downloaded chain ran ahead of the height the PeerGroup last tracked for the peer (routine under burst mining), making makers NACK offer-availability with "chain is not synced" despite being fully at the tip. On regtest, treat at-or-ahead as synced (best >= peers - 3, peers > 0); only a genuine lag is unsynced. At mainnet/testnet block cadence the tracking lag never reaches the tolerance, so the original symmetric check is kept there.
- WalletConfig: force bitcoinj bloom-filter false-positive rate to 1.0 on regtest so a freshly broadcast wallet tx (trade deposit/payout) is never omitted from the confirming block's merkleblock (the bloom-filter race that left deposits buried on-chain but unconfirmed in the wallet). No effect off regtest.

Supporting:
- New GetWalletSyncStatus gRPC (proto -> CoreWalletsService -> GrpcWalletsService -> cli) exposing the bitcoinj wallet's chain-sync flag + best height.
- DaoTestUtils.confirmTxAfterFilterPropagation and a take-retry/maker-sync gate in V1TradeFlow as additional robustness.

## Release Notes

* **New Features**
  * Added a wallet synchronization status API that reports whether chain height is in tolerance and provides the current best-chain height.
  * Added an opt-in “long-running-only” mode for E2E workflows to run only long-running soak/coverage tests.

* **Tests**
  * Added new long-running trade soak scenarios covering additional payment account types and F2F stability.
  * Improved E2E trade test support with clearer tx confirmation ordering and more robust maker/taker sync handling.

* **Bug Fixes**
  * Improved regtest chain height synchronization behavior for more reliable readiness checks.

### DIFF
--- a/.github/workflows/e2e-tests-compat.yml
+++ b/.github/workflows/e2e-tests-compat.yml
@@ -13,6 +13,10 @@ on:
         description: "Git tag/ref for the release peer (default: latest tag)"
         required: false
         default: ""
+      run_long_running_tests:
+        description: "Run ONLY the long-running soak/coverage tests (@Tag(\"longrunning\")); skips the normal suite"
+        type: boolean
+        default: false
 
 jobs:
   e2e-tests-compat:
@@ -135,6 +139,9 @@ jobs:
           # recognize and exits on. It tests a single-node master feature, not
           # cross-version compatibility.
           RUN_POLICY_E2E_TESTS: "false"
+          # Default false → normal compat suite (long-running tests excluded). Set true
+          # via workflow_dispatch to run ONLY the long-running tests against the release peer.
+          RUN_LONG_RUNNING_TESTS: ${{ inputs.run_long_running_tests }}
         run: apitest/docker/run-e2e-tests.sh --skip-build --skip-docker-build
 
       - name: Upload container logs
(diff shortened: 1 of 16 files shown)


HenrikJannsen (approved):
utACK

PR #7924, opened 2026-06-18
feat: implement single-instance lock to prevent concurrent runs

wodoro:
follow up to #7833

## New Features
- Prevents multiple instances of the application from running simultaneously by enforcing a single-instance lock for the selected data directory.
- If another instance is already running, the app shows (or prints, if no UI is available) a localized “already running” message with helpful context to close the current instance first.

## Improvements
- Improves restart handover by releasing the instance lock during shutdown, enabling clean subsequent launches.

### DIFF
--- a/common/src/main/java/bisq/common/app/InstanceLock.java
+++ b/common/src/main/java/bisq/common/app/InstanceLock.java
@@ -1,0 +1,154 @@
+/*
+ * This file is part of Bisq.
+ *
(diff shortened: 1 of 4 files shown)


HenrikJannsen (approved):
ACK

HenrikJannsen:
@wodoro Can you implment that for Bisq 2 as well?

wodoro:
sure

PR #7929, opened 2026-06-19
fix: prevent duplicate & ensure teardown of overlays

wodoro:
## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate overlays from appearing when opened multiple times.
  * Improved overlay hide behavior to avoid interaction-blocking modals in edge cases.
  * Updated overlay teardown so hide animations don’t allow unintended re-showing mid-animation.
  * Added resilience so animation setup/playback failures won’t leave overlays stuck.


HenrikJannsen (approved):
utACK

PR #7939, opened 2026-07-02
fix: broken mediation on master

wodoro:
Commit 8b432c6e94 moved DisputeValidation ahead of the side effects in onOpenNewDisputeMessage and made every failure fail-closed. Two problems:

1. testIfDisputeTriesReplay assumed the dispute under test was already in the list (pre-1.10.3 order validated after disputeList.add). Run before the add, the first legitimate dispute for a trade hit a null trade-id set and threw "more then 2 disputes", so the agent silently dropped it and mediation never progressed.

2. Fail-closing on advisory data checks (donation address, node address) silently discarded legitimate disputes and stopped surfacing the agent popup those cases rely on, while still exposing a popup-flood vector for forged payloads.

Fixes:
- testIfDisputeTriesReplay now counts the dispute under test via an idempotent uid injection, so the <= 2 limit holds regardless of whether the caller validates before or after adding to the list.
- Split ingest validation: integrity/forgery checks (contract data, replay) stay fail-closed and drop silently (log only, no popup); advisory checks (node/donation address) store and forward the dispute and alert the agent via validationExceptions, matching pre-1.10.3 behavior and the sibling peerOpenedDisputeForTrade handler.

Tests: DisputeValidationTest replay regressions; DisputeManagerAuthTest hard-drop vs advisory-store paths.

### DIFF
--- a/core/src/main/java/bisq/core/support/dispute/DisputeManager.java
+++ b/core/src/main/java/bisq/core/support/dispute/DisputeManager.java
@@ -415,18 +415,33 @@ public abstract class DisputeManager<T extends DisputeList<Dispute>> extends Sup
             return;
         }
 
+        // Integrity checks: a mismatch here means the payload was tampered with or replayed. The opener
+        // is already authenticated, so a genuine dispute cannot fail these. Fail closed and drop silently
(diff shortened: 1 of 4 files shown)


HenrikJannsen (approved):
utACK

PR #7941, opened 2026-07-04
Temporarily deactivate addSeedNode in Filter

HenrikJannsen:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated how network-provided node filters are applied: previously stored “filter provided” node lists are now cleared instead of being retained.
  * As a result, stale or unwanted node entries no longer persist after new filter information is received.
* **Configuration**
  * Command-line/config options for “filter provided” BTC/seed nodes are now ignored.

### DIFF
--- a/common/src/main/java/bisq/common/config/Config.java
+++ b/common/src/main/java/bisq/common/config/Config.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.io.UncheckedIOException;
 
+import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


wodoro (commented):
I guess we should also take precaution and block the consumption too? just in case there's any saved?
https://github.com/bisq-network/bisq/blob/b9fd8d8c767e37ba4022f032e53f82573c610032/common/src/main/java/bisq/common/config/Config.java#L969-L972

wodoro (approved):

alejandrogarcia83 (approved):
utACK

PR #7942, opened 2026-07-04
Temporarily deactivate usage of btcFeeReceiverAddresses from filter.

HenrikJannsen:
It is not used at the moment and will be handled in a new filter implementation in master.

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted BTC fee receiver selection so it no longer includes addresses from the configured filter, while still using other supported address sources.
  * This helps ensure the displayed receiver list reflects the expected set of valid addresses.


alejandrogarcia83 (approved):
utACK

PR #7950, opened 2026-07-07
Fix bug in opening dispute

HenrikJannsen:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dispute validation so only the expected trader can open a dispute.
  * Refined replay detection so legitimate first disputes are accepted while duplicate/replay attempts are blocked without overly strict null handling.
  * Enhanced dispute validation error logging to provide clearer diagnostics when validation fails.
* **Tests**
  * Expanded dispute authentication/validation test coverage, including new cases for replay failure behavior and first-dispute acceptance/persistence.

### DIFF
--- a/core/src/main/java/bisq/core/support/dispute/DisputeManager.java
+++ b/core/src/main/java/bisq/core/support/dispute/DisputeManager.java
@@ -406,8 +406,6 @@ public abstract class DisputeManager<T extends DisputeList<Dispute>> extends Sup
 
         // Disputes from clients < 1.2.0 always have support type ARBITRATION in dispute as the field didn't exist before
         dispute.setSupportType(openNewDisputeMessage.getSupportType());
-        // disputes from clients < 1.6.0 have state not set as the field didn't exist before
-        dispute.setState(Dispute.State.NEW);    // this can be removed a few months after 1.6.0 release
 
         if (!isDisputeOpenerSignaturePubKeyValid(dispute,
                 senderSignaturePubKey,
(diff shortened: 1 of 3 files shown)


core/src/main/java/bisq/core/support/dispute/DisputeManager.java
```diff
@@ -418,13 +418,10 @@ protected void onOpenNewDisputeMessage(OpenNewDisputeMessage openNewDisputeMessa
         try {
             DisputeValidation.validateDisputeData(dispute, btcWalletService);
             DisputeValidation.validateNodeAddresses(dispute, config);
-            DisputeValidation.validateSenderNodeAddress(dispute, openNewDisputeMessage.getSenderNodeAddress());
+            DisputeValidation.validateDisputeOpenerIsTrader(dispute, openNewDisputeMessage.getSenderNodeAddress());
             DisputeValidation.testIfDisputeTriesReplay(dispute, disputeList.getList());
-            if (dispute.isUsingLegacyBurningMan()) {
```
wodoro, line 423: instead of removing, how about rejecting the dispute if its legacy burningman? since it should not be happening anymore

HenrikJannsen: There are edge cases where fee receiver selection falls back to legacy BM. Not having all that in my head, but I think its not worth the risk as long we are not aware of a abuse scenario. Actually considering that there might be edge cases where we still use a legacy BM, I would rather revert the removal as then we have at least some protection if that case still occurs.

HenrikJannsen:
Replaced by PR to master: https://github.com/bisq-network/bisq/pull/7952

Issue #7243, opened 2024-09-14
Start Payment Button not working


WhatThaHecks:
### Description

The button "Payment Started" is not working properly. I've accepted a BTC offer and made the payment. Once I click on the "payment Started" button nothing whappens. I can see a spinning wheel, but instead of moving to the next step, I receive a "Please send confirmation again" message. It doesn't seems to be a conectivity problem once I've the green status on my app.

#### Version

Bisq v1.9.17

### Expected behaviour

After clicking on the Payment Started button I was expecting the trade proccess moving to the 3rd step "Wait until payment arrived"

### Actual behaviour

Nothing happens and I receive a message of "Please send confirmation again".

### Screenshots

![image](https://github.com/user-attachments/assets/f793e590-4eed-4daf-9723-8c6895bea6db)

![sendagain](https://github.com/user-attachments/assets/7caee541-020b-415b-9b13-a0f72d5031b4)

![image](https://github.com/user-attachments/assets/661740cf-72e0-4c00-94e0-096a3ee683eb)

#### Device or machine

Windowns 11 Desktop

pazza83:
Hi @WhatThaHecks thanks for opening this issue.

Unfortunately this has been an issue for a while. There are a handful cases a month where traders need to open mediation to get assistance for the issue.

Did you open mediation to get this resolved?

Issue #6715, opened 2023-05-30
Clicking the payment started button when buying btc results in an error message: "Message of payment sent but not yet received by peer"


Iyd4raf:
Clicking the payment started button when buying btc results in an error message: "Message of payment sent but not yet received by peer"

#### Version

1.9.10

### Steps to reproduce

1. Accept a BTC buy order
2. Click the payment started button once you have sent payment by bank transfer
3. Two different error messages appear. Either this message (paraphrased) is displayed next to the payment started button: "Sending confirmation failed. Please try again". Or the message shown below in the screenshot appears instead: "Message of payment sent but not yet received by peer".
4. The transaction is stuck on this stage because I cannot proceed

### Screenshots

![Capture](https://github.com/bisq-network/bisq/assets/43270055/a6567abe-bb37-4228-a667-111de43028bb)

#### Device or machine

windows 7

pazza83:
Maybe the trade peer is offline or has trouble with their Bisq.

Once the trade window time is up you can open mediation to get support from a mediator.

https://bisq.wiki/Dispute_resolution

Iyd4raf:
Thanks, this has been resolved - it seems that the peer was offline

Issue #3372, opened 2019-10-08
"Sending message failed" when trying to confirm my payment has started


jacopocer:
Hi there.
My trade is stuck: I (BTC buyer) cannot proceed confirming my payment has started. When I clicked on "Payment started" I got "Sending message failed". I was asked to go with mediator, but he/she is not doing anything about it.
The seller wrote to me in the direct chat that he/she got the payment, the trade is just stuck because of the problem of Bisq.
<img width="577" alt="Schermata 2019-10-07 alle 22 43 00" src="https://user-images.githubusercontent.com/19292930/66369798-1650a800-e99e-11e9-80f6-d044ccb7f545.png">
<img width="829" alt="Schermata 2019-10-07 alle 22 33 51" src="https://user-images.githubusercontent.com/19292930/66369799-1650a800-e99e-11e9-90b3-74f2f7f8a079.png">

Has anyone some ideas on how to resolve this issue?

calkob:
See issue ##3371 for similar problem.  I went to mediation yesterday with the same issue which didn't resolve it.  when i accepted the mediation it just ave me another error.  i then went to Arbitration which worked and i received my BTC, as Arb has a third key.  something is going wrong in that the seller node is getting banned. 

  You may need to go to Arb to fix.  As you may know you can now chat with the seller to inform them of the actions you are taking.

good luck

ripcurlx:
@jacopocer Was your issue solved in the end? We did publish in our latest release (v1.2.7) the option to refresh the trade status which should fix problems as described by you. Please re-open this issue if your problem wasn't resolved by now.

PR #7951, opened 2026-07-07
Re-send payment started message at startup when it was never ACKed

fenlark:
The buyer re-sends CounterCurrencyTransferStartedMessage on a timer until the seller ACKs it. The timer only lives in memory, so a restart kills it. The STARTUP rule in BuyerProtocol#onInitialized re-ran the send task only for the STORED_IN_MAILBOX and SEND_FAILED states. It skipped BUYER_SENT_FIAT_PAYMENT_INITIATED_MSG (sent, but no ACK yet). Such a trade never re-sent the message. The buyer stayed stuck at "Please send confirmation again". The seller never learned the payment was started. This often ended in needless mediation.

Fix: include BUYER_SENT_FIAT_PAYMENT_INITIATED_MSG in the STARTUP re-send condition. Re-sending is safe: the message has a deterministic uid, so re-sends deduplicate, and the seller side tolerates repeated delivery.

Also skip the send when the persisted message state is already ACKNOWLEDGED. That happens when the ACK arrives while no task is listening (right after a restart). In that case only apply the BUYER_SAW_ARRIVED state transition.

Fixes the buyer side of #7243. Also reported in #6715 and #3372.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved buyer recovery after restart for counter-currency “payment started” messages, ensuring resend eligibility remains active until the seller acknowledgment is received.
  * Avoided duplicate handling when the “payment started” message was already acknowledged before task execution.
  * Tightened trade state transition rules to support valid resumption paths while blocking invalid regressions.
* **Tests**
  * Added unit tests covering startup resend conditions, acknowledgment behavior after restarts, and allowed/blocked trade-state transitions for the buyer flow.

### DIFF
--- a/core/src/main/java/bisq/core/trade/protocol/bisq_v1/BuyerProtocol.java
+++ b/core/src/main/java/bisq/core/trade/protocol/bisq_v1/BuyerProtocol.java
@@ -42,6 +42,8 @@ import bisq.network.p2p.NodeAddress;
 import bisq.common.handlers.ErrorMessageHandler;
 import bisq.common.handlers.ResultHandler;
 
+import com.google.common.annotations.VisibleForTesting;
+
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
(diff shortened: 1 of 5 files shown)


HenrikJannsen (approved):
utACK

PR #7953, opened 2026-07-08
Fix installer evidence artifact lookup path

KimStrand:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved installer generation by ignoring temporary packaging output directories so they aren’t treated as installer artifacts.
  * Updated which generated installer files are used when creating the installer manifest and structure report, ensuring reports reflect the correct package outputs.


HenrikJannsen (dismissed):
utACK

HenrikJannsen (approved):
utACK

Issue #7945, opened 2026-07-06
Buyer phase gate rejects DepositTxAndDelayedPayoutTxMessage after deposit confirmation (seller side accepts it) - payment account never arrives, trade permanently stuck (re-report of #6530)


fenlark:
# Buyer phase gate rejects DepositTxAndDelayedPayoutTxMessage after deposit confirmation (seller side accepts it) - payment account never arrives, trade permanently stuck (re-report of #6530)

### Description

The buyer's handler for `DepositTxAndDelayedPayoutTxMessage` only accepts the message in phases TAKER_FEE_PUBLISHED or DEPOSIT_PUBLISHED, while the seller's symmetric handler also accepts DEPOSIT_CONFIRMED:

```java
// BuyerProtocol.java:108 - handle(DepositTxAndDelayedPayoutTxMessage)
expect(anyPhase(Trade.Phase.TAKER_FEE_PUBLISHED, Trade.Phase.DEPOSIT_PUBLISHED)

// SellerProtocol.java:92 - handle(ShareBuyerPaymentAccountMessage)
expect(anyPhase(Trade.Phase.TAKER_FEE_PUBLISHED, Trade.Phase.DEPOSIT_PUBLISHED, Trade.Phase.DEPOSIT_CONFIRMED)
```

The deposit confirms via the wallet listener independently of messaging, so a `DepositTxAndDelayedPayoutTxMessage` processed after the first confirmation is silently rejected by the buyer. The seller's payment account payload never gets stored, both sides keep a hash-only contract, and the buyer's trade UI turns into a blank pane - the trade is permanently stuck. Same defect as #6530 (2023, closed by the stale bot). Suggested fix is a one-liner: add DEPOSIT_CONFIRMED to the buyer gate, like the seller side (details under Additional info).

#### Version

v1.10.2. All line references below are also unchanged on current master (b9fd8d8c).

### Steps to reproduce

I hit this on a live trade; the exact delivery timing is hard to force on demand, but the structural race reproduces on regtest:

1. Take an offer as buyer.
2. Kill the buyer client right after TAKER_FEE_PUBLISHED.
3. Let the seller publish the deposit tx and mine a block.
4. Restart the buyer.

[...]

PR #7954, opened 2026-07-08
Accept deposit-tx message in DEPOSIT_CONFIRMED

fenlark:
The buyer's handler for DepositTxAndDelayedPayoutTxMessage accepts the message only in the TAKER_FEE_PUBLISHED and DEPOSIT_PUBLISHED phases. The deposit can confirm independently through the wallet listener, moving the trade to DEPOSIT_CONFIRMED before the (mailbox) message is processed. The phase gate then rejects it: no ACK is sent and the mailbox entry is not removed, so it is redelivered and re-rejected on every restart until its TTL. As a result the seller's payment account payload is never stored, both sides keep a hash-only contract, and the trade is permanently stuck.

The symmetric seller handler for ShareBuyerPaymentAccountMessage already accepts DEPOSIT_CONFIRMED. Add the same phase to the buyer gate. The existing preCondition still prevents reprocessing: it proceeds only while the payment account payload is missing and otherwise just resends the ACK.

The accepted phases move to a @VisibleForTesting constant, and a regression test pins the list (including DEPOSIT_CONFIRMED), mirroring STARTUP_RESEND_PAYMENT_STARTED_STATES.

Part of #7945

## Summary by CodeRabbit

* **Bug Fixes**
  * Buyers can now continue the trade flow after the deposit is confirmed, not just during earlier deposit-related steps.
  * Trade messages are handled consistently across all supported stages, reducing unexpected rejections.
  * Added coverage to verify message acceptance and rejection at the correct points in the trade process.

### DIFF
--- a/core/src/main/java/bisq/core/trade/protocol/bisq_v1/BuyerProtocol.java
+++ b/core/src/main/java/bisq/core/trade/protocol/bisq_v1/BuyerProtocol.java
@@ -69,6 +69,22 @@ public abstract class BuyerProtocol extends DisputeProtocol {
             Trade.State.BUYER_SEND_FAILED_FIAT_PAYMENT_INITIATED_MSG
     };
 
+    // Phases in which the buyer accepts the DepositTxAndDelayedPayoutTxMessage.
+    // DEPOSIT_CONFIRMED is included because the deposit can confirm through the
+    // wallet listener before the (mailbox) message is processed. A message
+    // arriving after the first confirmation must still be accepted, otherwise the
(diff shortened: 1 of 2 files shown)


wodoro:
utACK

wodoro (approved):

wodoro:
what was changed?

fenlark:
> what was changed?

Rebase on latest master only, the patch is byte-identical.

HenrikJannsen (approved):
utACK

HenrikJannsen:
Failed test is not PR related and fixed on master.

PR #7955, opened 2026-07-08
Render buyer step 2 without seller payment account

fenlark:
When the seller's PaymentAccountPayload never arrives (for example the mailbox DepositTxAndDelayedPayoutTxMessage is missed), BuyerStep2View dereferenced it in addContent() via
paymentAccountPayload.showRefTextWarning(), throwing an NPE inside the TradeStepView constructor. TradeSubView swallows that exception and leaves the whole trade step pane blank, so the buyer cannot reach the chat, the dispute button or the mediation-result popup and the trade cannot be resolved in the app.

Guard the null case: show a short warning that the seller's payment details are missing and to contact the mediator, and let addContent() finish so the pane and its controls render normally. The 'Pay from BSQ wallet' button is likewise skipped when the payload is null, since its click handler would otherwise dereference it.

Part of #7945

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the buyer payment step to handle missing seller payment details safely.
  * Added a clear warning instructing buyers to contact the seller in trade chat and open mediation if the trade cannot be recovered.
  * Prevented payment confirmation and BSQ wallet actions when required payment information is unavailable.
  * Prevented errors when payment account data is missing.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/steps/buyer/BuyerStep2View.java
+++ b/desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/steps/buyer/BuyerStep2View.java
@@ -440,7 +440,9 @@ public class BuyerStep2View extends TradeStepView {
                 gridRow = SbpForm.addFormForBuyer(gridPane, gridRow, paymentAccountPayload);
                 break;
             default:
-                log.error("Not supported PaymentMethod: " + paymentMethodId);
+                if (paymentAccountPayload != null) {
+                    log.error("Not supported PaymentMethod: " + paymentMethodId);
+                }
         }
 
         Trade trade = model.getTrade();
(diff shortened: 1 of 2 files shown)


desktop/src/main/java/bisq/desktop/main/portfolio/pendingtrades/steps/buyer/BuyerStep2View.java
```diff
@@ -484,7 +495,7 @@ protected void addContent() {
             gridPane.setMinHeight(600); // make the scrollpane parent node activate its scrollbar
         }
 
-        if (trade.getOffer().getCurrencyCode().equals("BSQ")) {
+        if (paymentAccountPayload != null && trade.getOffer().getCurrencyCode().equals("BSQ")) {
```
wodoro, line 499: Nice catch on the NPE, a blank pane with no chat or dispute button is definitely worse than a warning.

One thing this opens up though: in master the NPE fires at paymentAccountPayload.showRefTextWarning() (line 464) before confirmButton is created at line 477, so the button never renders in the null-payload case. Now that addContent() completes, the button renders at line 488 and is enabled, since the only disable logic is confirmButton.setDisable(!trade.confirmPermitted()) in updateDisputeState() and BuyerTrade.confirmPermitted() only looks at the dispute state.

Clicking it doesn't throw. Every instanceof in onPaymentStarted() is false on null, so it falls through to showConfirmPaymentStartedPopup(), and PendingTradesDataModel.onPaymentStarted has no payload guard either. The trade moves to FIAT_SENT. So the buyer signals a payment they cannot have made, since they were never shown where to send it, and the trade lands past the phase list that #7954 widened, where a late DepositTxAndDelayedPayoutTxMessage would still have been accepted.

Suggest disabling the button when the payload is null. It needs to be in two places, since updateDisputeState() re-enables it on any dispute state change and would undo a one-shot disable in addContent():
```
confirmButton.setDisable(model.dataModel.getSellersPaymentAccountPayload() == null
        || !trade.confirmPermitted());
```
Two smaller things:

- if (!paymentMethodId.isEmpty()) is a proxy for the real condition. if (paymentAccountPayload != null) is equivalent here and states the intent directly.
- The new string says to contact the mediator, but TradeStepView.activate() only wires the step-info button to openSupportTicket() once isTradePeriodOver(); before that it opens tntioning chat in the string so the advice matches what the button actually does at this point.

Separately, and probably out of scope here: CoreTradesService.confirmPaymentStarted only checks isDepositConfirmed(), so bisq-cli confirmpaymentstarted reaches FIAT_SENT with no payload regardless of any UI fix. Worth a follow-up under #7945.

fenlark: All three applied: the button is now disabled when the payload is null, both in addContent and in updateDisputeState so a dispute state change does not re-enable it. The proxy condition is replaced with the explicit null check. The warning now points to the trade chat first and to mediation once the trade period is over, matching what the step info button actually opens.

HenrikJannsen (approved):
utACK

Issue #1118, opened 2017-12-26
GBYTE has 9 decimal places, not 8


TheButterZone:
When I type .xxxxxxxxx in a create buy bitcoin offer, the 9th number keeps getting truncated out.

And BTC has 8, not 4, anything past 4 keeps getting truncated out of the BTC to buy, minimum amount of BTC, security deposit fields.

ripcurlx:
@TheButterZone Is there a reason why you need this precision? Atm for buying BTC with any altcoin we show 8 decimals as you stated above.

ripcurlx:
I'm closing this issue for now. Feel free to reopen if this is a pressing issue.

PR #7960, opened 2026-07-10
Round altcoin trade amounts to coin precision

fenlark:
## Problem
Bisq represents every altcoin amount at 8 decimals
(`Altcoin.SMALLEST_UNIT_EXPONENT`), regardless of how many decimals the
coin actually supports on its own chain. For a coin with fewer than 8
decimals (for example USDC, which has 6), the volume shown as the amount
to send carries digits the sender cannot transfer, so the exact amount
stated in the trade can never be paid.

This is the opposite of #1118 (a coin with more than 8 decimals, where
capping at 8 is fine because 8 decimals are still sendable). Here the
coin has fewer than 8 decimals, so the extra digits are unsendable.

## Change
- Add a per-asset precision to the asset model (`Asset.getPrecision()`,
  default 8), overridden on the coins that use fewer than 8 decimals,
  exposed via `CurrencyUtil.getCryptoPrecision(code)` (cached next to
  the existing `isCryptoCurrency` cache, since the registry lookup is a
  linear scan called per offer row).
- Round the altcoin trade volume to that precision on the payment path:
  `Offer`, `Trade`, `Contract` and the create/take amount models.

The internal 8-decimal storage exponent is unchanged; only the volume
the sender is asked to transfer is rounded, and precision 8 is an exact
no-op. Rounding is half up in exact integer arithmetic with an overflow
guard. A volume below half of one on-chain unit rounds to zero and is
to be rejected by callers as an amount below the coin's minimum, never
inflated to a whole unit, which would silently change the agreed rate;
the create path already validates the rounded volume and min volume as
non-zero before an offer can be placed.

## Notes
- Volume is not part of the signed contract (only `tradeAmount` and
  `tradePrice` are), so the wire format is unaffected. Rollout is not
  behaviour-identical though: both peers compute the trade volume
  locally, so an old and a new client can display send amounts
  differing by up to half a chain unit until both sides are upgraded.
- Fiat is unaffected. Fiat volumes are already rounded to whole currency
  units (`getRoundedFiatVolume`), which is payable for every currency,
  so this PR does not touch the fiat path.
- Derived analytics stay raw: `TradeStatistics3` and the market-data
  JSON exports (`OfferForJson`, `TradeStatisticsForJson`) are untouched.
  The rule is: the payment instruction is rounded to chain precision,
  derived analytics are not.
- A few obscure listed coins stay at the default 8 for now: four whose
  exact precision needs confirming (XDR0 - XDR, ZOD - Krypton, AMIT -
  Amitycoin, DONU - Donu) and two indivisible whole-unit coins (ASK -
  Askcoin, SF - Siafund).

## Tests
`VolumeUtilTest` covers rounding a volume to a coin's precision, that
precision 8 stays bit-exact above 2^53, that a volume below half of one
on-chain unit rounds to zero (and zero stays zero at every precision),
and that the rounding arithmetic fails loudly on overflow.

## New Features
- Added asset-specific cryptocurrency precision, with a default of 8 decimal places.
- Added precision-aware volume calculations and exact half-up rounding to valid on-chain units.
- Added explicit precision support for coins and ERC-20 tokens.
- Added validation to reject trade volumes that round to zero.

## Bug Fixes
- Improved cryptocurrency volume accuracy across varying decimal places.
- Prevented invalid fractional amounts and standardized rounding behavior.
- Take-off controls now flag and prevent amounts that cannot produce a transferable volume.

### DIFF
--- a/assets/src/main/java/bisq/asset/AbstractAsset.java
+++ b/assets/src/main/java/bisq/asset/AbstractAsset.java
@@ -33,11 +33,20 @@ public abstract class AbstractAsset implements Asset {
 
     private final String name;
     private final String tickerSymbol;
+    private final int precision;
     private final AddressValidator addressValidator;
 
     public AbstractAsset(String name, String tickerSymbol, AddressValidator addressValidator) {
+        this(name, tickerSymbol, DEFAULT_PRECISION, addressValidator);
+    }
+
+    public AbstractAsset(String name, String tickerSymbol, int precision, AddressValidator addressValidator) {
         this.name = notBlank(name);
         this.tickerSymbol = notBlank(tickerSymbol);
+        if (precision < 0) {
+            throw new IllegalArgumentException("precision must be non-negative: " + precision);
+        }
+        this.precision = precision;
         this.addressValidator = Objects.requireNonNull(addressValidator);
     }
 
@@ -49,6 +58,11 @@ public abstract class AbstractAsset implements Asset {
     @Override
     public final String getTickerSymbol() {
         return tickerSymbol;
+    }
+
+    @Override
+    public final int getPrecision() {
+        return precision;
     }
 
     @Override
--- a/core/src/main/java/bisq/core/util/VolumeUtil.java
+++ b/core/src/main/java/bisq/core/util/VolumeUtil.java
@@ -26,6 +26,8 @@ import org.bitcoinj.core.Monetary;
 import org.bitcoinj.utils.Fiat;
 import org.bitcoinj.utils.MonetaryFormat;
 
+import com.google.common.math.LongMath;
+
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
 
@@ -60,6 +62,31 @@ public class VolumeUtil {
         // Smallest allowed volume is factor (e.g. 10 EUR or 1 EUR,...)
         roundedVolume = Math.max(factor, roundedVolume);
         return Volume.parse(String.valueOf(roundedVolume), volumeByAmount.getCurrencyCode());
+    }
+
+    /**
+     * Round an altcoin volume to the coin's own precision so the resulting amount can
+     * actually be transferred on that coin's chain. Altcoin volumes are stored at
+     * 10^-8 precision, but the coin may support fewer decimals; one on-chain unit is
+     * 10^(8 - precision) internal units. Unlike {@link #getAdjustedFiatVolume} this does
+     * NOT coarsen to whole coins - it rounds to the finest unit the coin can represent.
+     *
+     * @param volumeByAmount    The volume generated from an amount.
+     * @param precision         The number of decimals the coin supports on its chain.
+     * @return The adjusted altcoin volume.
+     */
+    public static Volume getAdjustedAltcoinVolume(Volume volumeByAmount, int precision) {
+        int cappedPrecision = Math.min(precision, Altcoin.SMALLEST_UNIT_EXPONENT);
+        long step = LongMath.pow(10, Altcoin.SMALLEST_UNIT_EXPONENT - cappedPrecision);
+        long value = volumeByAmount.getValue();
+        // Round to the nearest multiple of step (half up) using exact integer arithmetic. We
+        // avoid floating point so large volumes stay bit-exact and precision 8 is a true no-op
+        // (a long -> double cast would lose precision above 2^53).
+        long roundedVolume = ((value + step / 2) / step) * step;
+        // Never round down to zero: a volume below one on-chain unit (e.g. a fraction of a
+        // whole coin on a 0-decimal coin) is raised to exactly one unit, so it stays sendable.
+        roundedVolume = Math.max(step, roundedVolume);
+        return new Volume(Altcoin.valueOf(volumeByAmount.getCurrencyCode(), roundedVolume));
     }
 
 
--- a/core/src/main/java/bisq/core/util/VolumeUtil.java
+++ b/core/src/main/java/bisq/core/util/VolumeUtil.java
@@ -26,6 +26,8 @@ import org.bitcoinj.core.Monetary;
 import org.bitcoinj.utils.Fiat;
 import org.bitcoinj.utils.MonetaryFormat;
 
+import com.google.common.math.LongMath;
+
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
 
@@ -60,6 +62,32 @@ public class VolumeUtil {
         // Smallest allowed volume is factor (e.g. 10 EUR or 1 EUR,...)
         roundedVolume = Math.max(factor, roundedVolume);
         return Volume.parse(String.valueOf(roundedVolume), volumeByAmount.getCurrencyCode());
+    }
+
+    /**
+     * Round an altcoin volume to the coin's own precision so the resulting amount can
+     * actually be transferred on that coin's chain. Altcoin volumes are stored at
+     * 10^-8 precision, but the coin may support fewer decimals; one on-chain unit is
+     * 10^(8 - precision) internal units. Unlike {@link #getAdjustedFiatVolume} this does
+     * NOT coarsen to whole coins - it rounds to the finest unit the coin can represent.
+     *
+     * @param volumeByAmount    The volume generated from an amount.
+     * @param precision         The number of decimals the coin supports on its chain.
+     * @return The adjusted altcoin volume; zero when the volume is below half of one on-chain
+     *         unit, which callers must reject as an amount below the coin's minimum.
+     */
+    public static Volume getAdjustedAltcoinVolume(Volume volumeByAmount, int precision) {
+        int cappedPrecision = Math.min(precision, Altcoin.SMALLEST_UNIT_EXPONENT);
+        long step = LongMath.pow(10, Altcoin.SMALLEST_UNIT_EXPONENT - cappedPrecision);
+        long value = volumeByAmount.getValue();
+        // Round to the nearest multiple of step (half up) using exact integer arithmetic. We
+        // avoid floating point so large volumes stay bit-exact and precision 8 is a true no-op
+        // (a long -> double cast would lose precision above 2^53). checkedAdd guards the one
+        // overflow-capable step, failing loudly instead of producing a silently wrong volume.
+        // A volume below half of one on-chain unit rounds to zero; inflating it to one unit
+        // here would silently change the agreed rate, so rejection is left to the callers.
+        long roundedVolume = (LongMath.checkedAdd(value, step / 2) / step) * step;
+        return new Volume(Altcoin.valueOf(volumeByAmount.getCurrencyCode(), roundedVolume));
     }
 
 
--- a/assets/src/main/java/bisq/asset/Asset.java
+++ b/assets/src/main/java/bisq/asset/Asset.java
@@ -38,9 +38,28 @@ package bisq.asset;
  */
 public interface Asset {
 
+    /**
+     * Default number of decimal places for an asset. This is also Bisq's internal
+     * maximum precision for altcoin amounts, so assets whose chain supports more
+     * decimals are still handled at this precision, while assets supporting fewer
+     * should report their real value (see {@link #getPrecision()}).
+     */
+    int DEFAULT_PRECISION = 8;
+
     String getName();
 
     String getTickerSymbol();
+
+    /**
+     * The number of decimal places this asset supports on its own chain, i.e. the
+     * number of digits after the decimal point in its smallest transferable unit.
+     * Defaults to {@value #DEFAULT_PRECISION}. Assets whose chain supports fewer
+     * decimals (e.g. USD Coin with 6) should report their real value so that trade
+     * amounts are rounded to a figure the sender can actually transfer.
+     */
+    default int getPrecision() {
+        return DEFAULT_PRECISION;
+    }
 
     AddressValidationResult validateAddress(String address);
 }
--- a/assets/src/main/java/bisq/asset/Coin.java
+++ b/assets/src/main/java/bisq/asset/Coin.java
@@ -44,8 +44,17 @@ public abstract class Coin extends AbstractAsset {
         this(name, tickerSymbol, addressValidator, Network.MAINNET);
     }
 
+    public Coin(String name, String tickerSymbol, int precision, AddressValidator addressValidator) {
+        this(name, tickerSymbol, precision, addressValidator, Network.MAINNET);
+    }
+
     public Coin(String name, String tickerSymbol, AddressValidator addressValidator, Network network) {
         super(name, tickerSymbol, addressValidator);
+        this.network = network;
+    }
+
+    public Coin(String name, String tickerSymbol, int precision, AddressValidator addressValidator, Network network) {
+        super(name, tickerSymbol, precision, addressValidator);
         this.network = network;
     }
 
--- a/assets/src/main/java/bisq/asset/Erc20Token.java
+++ b/assets/src/main/java/bisq/asset/Erc20Token.java
@@ -30,4 +30,8 @@ public abstract class Erc20Token extends Token {
     public Erc20Token(String name, String tickerSymbol) {
         super(name, tickerSymbol, new EtherAddressValidator());
     }
+
+    public Erc20Token(String name, String tickerSymbol, int precision) {
+        super(name, tickerSymbol, precision, new EtherAddressValidator());
+    }
 }
--- a/assets/src/main/java/bisq/asset/Token.java
+++ b/assets/src/main/java/bisq/asset/Token.java
@@ -34,4 +34,8 @@ public abstract class Token extends AbstractAsset {
     public Token(String name, String tickerSymbol, AddressValidator addressValidator) {
         super(name, tickerSymbol, addressValidator);
     }
+
+    public Token(String name, String tickerSymbol, int precision, AddressValidator addressValidator) {
+        super(name, tickerSymbol, precision, addressValidator);
+    }
 }
--- a/assets/src/main/java/bisq/asset/coins/Animecoin.java
+++ b/assets/src/main/java/bisq/asset/coins/Animecoin.java
@@ -23,7 +23,7 @@ import bisq.asset.NetworkParametersAdapter;
 
 public class Animecoin extends Coin {
     public Animecoin() {
-        super("Animecoin", "ANI", new Base58AddressValidator(new AnimecoinMainNetParams()));
+        super("Animecoin", "ANI", 5, new Base58AddressValidator(new AnimecoinMainNetParams()));
     }
 
     public static class AnimecoinMainNetParams extends NetworkParametersAdapter {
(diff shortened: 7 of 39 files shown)


assets/src/main/java/bisq/asset/AbstractAsset.java
```diff
@@ -33,11 +33,17 @@ public abstract class AbstractAsset implements Asset {
...
 
     public AbstractAsset(String name, String tickerSymbol, AddressValidator addressValidator) {
+        this(name, tickerSymbol, DEFAULT_PRECISION, addressValidator);
+    }
+
+    public AbstractAsset(String name, String tickerSymbol, int precision, AddressValidator addressValidator) {
         this.name = notBlank(name);
         this.tickerSymbol = notBlank(tickerSymbol);
+        this.precision = precision;
         this.addressValidator = Objects.requireNonNull(addressValidator);
```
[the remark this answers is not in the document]
fenlark: Added the non-negative check in AbstractAsset.

assets/src/main/java/bisq/asset/coins/Amitycoin.java
```diff
@@ -23,6 +23,6 @@
 public class Amitycoin extends Coin {
 
     public Amitycoin() {
-        super("Amitycoin", "AMIT", new RegexAddressValidator("^amit[1-9A-Za-z^OIl]{94}"));
+        super("Amitycoin", "AMIT", 4, new RegexAddressValidator("^amit[1-9A-Za-z^OIl]{94}"));
```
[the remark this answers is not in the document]
fenlark: These came from each coin's chain config (CryptoNoteConfig.h and the amount.h COIN constant), but AMIT and DONU are genuinely disputed. Since over-declaring precision would make amounts unsendable (the bug this PR fixes), I moved both to the default 8 rather than assert a disputed value in the unsafe direction. ANI and CLOAK matched the config, so they stay.

core/src/main/java/bisq/core/util/VolumeUtil.java
```diff
@@ -62,6 +64,31 @@ public static Volume getAdjustedFiatVolume(Volume volumeByAmount, int factor) {
...
+     * @return The adjusted altcoin volume.
+     */
+    public static Volume getAdjustedAltcoinVolume(Volume volumeByAmount, int precision) {
+        int cappedPrecision = Math.min(precision, Altcoin.SMALLEST_UNIT_EXPONENT);
+        long step = LongMath.pow(10, Altcoin.SMALLEST_UNIT_EXPONENT - cappedPrecision);
+        long value = volumeByAmount.getValue();
+        // Round to the nearest multiple of step (half up) using exact integer arithmetic. We
+        // avoid floating point so large volumes stay bit-exact and precision 8 is a true no-op
+        // (a long -> double cast would lose precision above 2^53).
+        long roundedVolume = ((value + step / 2) / step) * step;
```
wodoro: nit:
((value + step / 2) / step) * step has no overflow guard. Unreachable with real trade limits, but TradeStatistics3.isValid() bounds amount and only requires price > 0 (TradeStatistics3.java:469), so a crafted peer payload is the one untrusted path into this arithmetic. The failure mode is a silent wrong value (Math.max turns the negative into step) rather than an exception. LongMath.checkedAdd costs nothing.

wodoro (changes requested):
1) The "no-op for every coin except the sub-8-decimal ones" claim is not quite true.

    roundedVolume = Math.max(step, roundedVolume);
    
      At precision 8, step == 1, so a volume of 0 becomes 1. Every default-precision altcoin, BSQ included, now returns a non-zero
      volume where it previously returned zero. Callers currently guard on non-zero amount so this is unreachable today, but it silently
      invalidates the stated invariant, and zero-volume is exactly the kind of sentinel that gets relied on later. Fix is small:
    
      ```
      if (value > 0) {
          roundedVolume = Math.max(step, roundedVolume);
      }
      ```
  
2) Separately, the clamp silently changes the agreed rate.
    ```
      assertEquals(100_000_000L,
              VolumeUtil.getAdjustedAltcoinVolume(altcoin("SF", 40_000_000L), 0).getValue(),
              "A sub-unit volume on a 0-decimal coin must round up to one whole unit, not zero.");
    ```
    
      0.4 SF becoming 1 SF is a 150% increase, not a rounding error. If a 0-decimal coin is ever given a real precision (the description mentions ASK and SF as candidates) that becomes live behaviour, and the next reviewer sees a passing test and assumes it was considered. For altcoins the honest response to "this volume rounds to zero" is to reject the amount as below the coin's minimum, not to inflate it.

3) The JSON export question should be answered before merge, not left open.

    Right now the same trade is published two ways: the app shows the rounded volume, while TradeStatisticsForJson.java:93 computes
  its own getTradePrice().getVolumeByAmount(getTradeAmount()) and publishes the raw 8 decimal value.

      From first principles these are two different concepts that the PR conflates:
    
      - Payment instruction (what the altcoin sender must transfer): must be representable on-chain. This is Offer, Trade, Contract and
        the create/take models. Rounding belongs here.
      - Economic volume (price times amount, for charts, averages, market data): has no on-chain constraint.
        TradeStatistics3.getTradeVolume() recomputes from stored price and amount, so after upgrade every historical trade's volume
        shifts and old and new clients report different volumes for the same past trade.

      To be fair on, it seems outside BSQ and USD the only readers of TradeStatistics3.getTradeVolume() are the market trades chart, a tooltip and a list cell. AveragePriceUtil and both DAO economy dashboards filter to BSQ and USD first.
      So this is a consistency and scope argument, not a safety one.
    
      Recommendation: drop the TradeStatistics3 change and leave the JSON exports raw. That keeps the diff on the payment path and makes the rule stateable in one sentence: the payment instruction is rounded to chain precision, derived analytics are not. If you keep the TradeStatistics3 change then the JSON exports have to be rounded too. Shipping the two halves inconsistent is the worst option.

nits:
1. The peer-compatibility claim is overstated. "No consensus or peer-compatibility impact" is true for the wire format, since volume is not in the signed contract (only tradeAmount and tradePrice are). It is not true for behaviour: both peers compute
Contract.getTradeVolume() locally, so during rollout an old maker and a new taker display different amounts to send. Bounded by half a chain unit, so low severity, but the description asserts the opposite.

2. Agreeing with the CodeRabbit nit on caching getCryptoPrecision. findAsset is a linear stream scan over 122 assets and it is now called per offer row on UI-driven recalculation. CurrencyUtil.java:66-72 already memoizes isCryptoCurrency with the comment "calls to isFiatCurrency and isCryptoCurrency are very frequent", and the new lookup sits right next to it, uncached.

core/src/main/java/bisq/core/util/VolumeUtil.java
fenlark: All three applied, plus the nits.

1+2) The clamp is gone: rounding is pure half-up, zero stays zero, and a sub-half-unit volume rounds to zero instead of being inflated. Rejection already exists on the create path (the view model requires both the rounded volume and min volume to be non-zero before place is enabled), and with the lowest declared precision being 2 a zero rounding needs a price orders of magnitude off market, so the zero contract is just documented in the javadoc and locked by tests.

3) TradeStatistics3 change dropped. The rule is as you stated it - the payment instruction is rounded to chain precision, derived analytics are not.

Nits: precision lookup cached next to the isCryptoCurrency cache, the rounding add guarded with LongMath.checkedAdd (loud failure, covered by a test), and the description's peer-compatibility paragraph is reworded.

wodoro: One small nit:

`computeIfAbsent` throws NPE on a null key, where the uncached version returned `DEFAULT_PRECISION`. Unreachable today since every call site sits behind `isCryptoCurrency`, but both other cached methods in this file guard it (`isFiatCurrency:203`, `isCryptoCurrency:245`) for the same reason, and the method already falls back to the default for unknown codes:

```java
if (currencyCode == null) {
    return Asset.DEFAULT_PRECISION;
}
```

wodoro (dismissed):
utACK

core/src/main/java/bisq/core/util/VolumeUtil.java
fenlark: Applied

HenrikJannsen:
Codex review result:

here codex review to 7960:

# Review: `ff88a2f15f487bd540c177dfe5e88011fbddb456`

## Scope

- Reviewed commit: `ff88a2f15f487bd540c177dfe5e88011fbddb456` (`Round altcoin trade amounts to coin precision`)
- Parent: `47e87f219e1c68e101be22861deda009d668f845`
- Branch: `respect-altcoin-precision-in-trade-amount`
- Scope covered: asset precision metadata, precision lookup, rounding arithmetic, offer creation/taking, persisted trades and contracts, desktop and API validation paths, protocol compatibility, and focused tests.

## Commit summary

The commit adds per-asset decimal precision, rounds derived altcoin payment volumes to that precision, and applies the calculation in offers, take-offer models, contracts, trades, and desktop models. It also adds unit tests for the integer rounding helper.

The underlying goal is sound: a payment obligation must be representable in the asset's smallest transferable unit. The implementation is not safe to merge yet because activation is not versioned, zero-valued obligations are not rejected at core trust boundaries, and at least one listed indivisible asset is missing its precision metadata.

## Findings

### P1 — Gate the new payment semantics behind a new trade protocol version

Locations:

- `core/src/main/java/bisq/core/trade/model/bisq_v1/Trade.java:776`
- `core/src/main/java/bisq/core/trade/model/bisq_v1/Contract.java:477`
- `common/src/main/java/bisq/common/app/Version.java:111`

The added branches unconditionally change the counter-currency payment amount derived from an offer's signed `tradeAmount` and `tradePrice`, but `Version.TRADE_PROTOCOL_VERSION` remains `4`. `OfferFilterService` accepts offers whose protocol version equals the local version, so pre-change and post-change clients continue to trade with one another as protocol version 4.

The rounded volume is not persisted as a signed contract field; clients derive it locally from amount and price. Consequently, a pre-change peer can derive `0.12345678 USDC` while a post-change peer derives `0.123457 USDC` from the same contract. The same unconditional logic also changes the displayed payment obligation of already active persisted v4 trades as soon as one party upgrades. The contract JSON can still match because it signs amount and price, not the computed volume, allowing the protocol to proceed despite different settlement expectations.

Suggested fix:

- Increment `TRADE_PROTOCOL_VERSION` (or introduce an equivalent mandatory capability).
- Apply chain-precision rounding only to offers/contracts carrying the new version.
- Preserve legacy volume derivation for already active trades using the old version.
- Add mixed-version and persisted-trade tests. `OpenOfferManager` already has migration logic for persisted open offers when the protocol version increases.

### P1 — Reject a rounded-zero payment in core offer and take validation

Locations:

- `core/src/main/java/bisq/core/util/VolumeUtil.java:76`
- `core/src/main/java/bisq/core/offer/bisq_v1/TakeOfferModel.java:197`
- `core/src/main/java/bisq/core/offer/placeoffer/bisq_v1/tasks/ValidateOffer.java:72`

`getAdjustedAltcoinVolume` intentionally returns zero below half of one on-chain unit and documents that callers must reject it, but none of the core callers performs that rejection. `ValidateOffer` validates only the BTC amount and positive price; `TakeOfferModel` calculates the rounded volume after its input validation and does not assert that the result is positive. The desktop take button likewise validates the BTC amount, not the resulting counter-currency volume. The daemon/API create and take paths therefore bypass the maker UI's incidental non-zero volume check, and a crafted network offer reaches the same taker path.

For example, a `0.0001 BTC` trade at `0.025 BTC/TRTL` yields `0.004 TRTL`, which this commit rounds to `0.00 TRTL`. Such an offer can pass the current core checks and proceed into a BTC escrow trade with a zero altcoin payment obligation.

Suggested fix:

- At offer creation/placement, reject when either the rounded minimum or maximum volume is zero.
- At take time, reject the rounded volume for the selected trade amount before funding or starting the trade protocol.
- Enforce this in core/domain validation, including received peer offers and API paths, not only in desktop validation.
- Add tests around the exact below-half, half-unit, range-offer, API, and peer-offer boundaries.

### P1 — Declare Siafund's zero-decimal precision

Locations:

- `core/src/main/java/bisq/core/locale/CurrencyUtil.java:441`
- `assets/src/main/java/bisq/asset/coins/Siafund.java:27`
- `core/src/test/java/bisq/core/util/VolumeUtilTest.java:68`

The new unit test uses `SF` as the zero-decimal example, but production metadata was not updated: `Siafund` still calls the legacy constructor and therefore inherits `Asset.DEFAULT_PRECISION == 8`. `CurrencyUtil.getCryptoPrecision("SF")` consequently returns 8 and performs no rounding for Siafund trades. Siafund is registered and included in Bisq's main crypto-currency list.

This is not merely a display convention: the Sia consensus documentation describes the 10,000 Siafunds as indivisible units: https://github.com/NebulousLabs/Sia/blob/master/doc/Consensus.md#currency.

Suggested fix:

- Construct `Siafund` with precision `0`.
- Add registry-backed assertions such as `CurrencyUtil.getCryptoPrecision("SF") == 0`, rather than testing only the helper with a manually supplied precision.
- Add precision assertions for every asset assigned a non-default value so metadata omissions cannot pass the suite.

## Risks and unresolved domain questions

- The payment amount is derived from mutable application metadata rather than persisted in the signed contract. Any later precision correction can retroactively change how historical or active trades are displayed unless every change is protocol-versioned. Persisting/negotiating the settlement precision or rounded amount would make this invariant more robust.
- No authoritative trade specification currently defines the rounding mode, activation boundary, treatment of zero, or whether trade statistics represent the quoted unrounded amount versus the actual settled amount. This protocol-visible rule should be specified under `docs/specifications/trade/`.
- The individual precision constants added for the other assets need authoritative-source verification. Existing asset tests cover registration and address validation but do not assert monetary precision.
- The gRPC `OfferInfo.volume` comments still describe altcoin volume as rounded to two decimal places; the public API documentation should be updated to describe per-asset chain precision.

## Verification performed

- `git diff --check 47e87f219e1c68e101be22861deda009d668f845 ff88a2f15f487bd540c177dfe5e88011fbddb456` — passed.
- `./gradlew --no-daemon --max-workers=2 :core:test --tests bisq.core.util.VolumeUtilTest :assets:test --console=plain` — passed.
- `./gradlew --no-daemon --max-workers=2 :core:test --tests bisq.core.offer.OfferTest --tests bisq.core.trade.model.bisq_v1.ContractTest --tests bisq.core.offer.bisq_v1.OfferPayloadCanonicalEncoderTest --console=plain` — passed.

These passing tests do not exercise the three failures above.

## Overall assessment

Do not merge as-is. The arithmetic helper is reasonable for positive values, but the change modifies settlement semantics without protocol isolation, permits zero-payment trades through core/API paths, and omits production precision metadata for Siafund. Address those findings, add end-to-end/version-boundary tests, and document the settlement invariant before release.

wodoro:
I do not think a TRADE_PROTOCOL_VERSION bump is the right instrument, for four reasons.

First, it cannot fix what it is meant to fix. The volume is not in the signed contract (Contract carries trade_amount and trade_price only), so both peers derive it locally. Bumping the version does not make an old client round. OfferFilterService.hasSameProtocolVersion uses strict equality, so the only effect is that old and new clients stop seeing each other's offers at all, on every market including fiat. That is a hard order book split for a rounding change affecting twelve altcoins. The last bump was 1.7.0, for an actual protocol change.

Second, the divergence is bounded by half of one on-chain unit, which is by construction the smallest amount the coin can transfer. For USDC and USDT-E that is 0.0000005 units. There is no automated check this can break: the only automated counter-currency verification is the XMR tx proof, and XMR is unaffected (12 decimals, capped at 8, exact no-op).

Third, there is precedent for this exact class of change without a bump. getRoundedFiatVolume and getAdjustedVolumeForHalCash are the same shape, locally derived, unsigned, computed independently by both peers, and both shipped without touching the protocol version. The HalCash adjustment moves the displayed fiat obligation by up to 5 EUR, two orders of magnitude more than anything here.

Fourth, if a clean activation boundary is wanted anyway, the codebase already has a cheaper mechanism: Contract.contractVersion plus a date-gated activation, exactly as requiresDisputeAgentPubKeyVersion does. That gives both peers the same signed switch, grandfathers in-flight trades, and does not segregate the order book. Filter.requireUpdateToNewVersionForTrading is the operational lever if a hard cutover is ever needed.

On the specific concern about in-flight v4 trades: real, but small. The change is display only, no persisted or signed field changes, so nothing historic is rewritten (TradeStatistics3 and the JSON exports are deliberately untouched). An in-flight trade can shift by at most half of one chain unit when one side upgrades.
in 2019, and ANI, CLOAK, EMC, LCP, MILE, UPX, WRKZ have no recorded trades). But USDC and USDT-E are active, 187 trades between them in 2026 so far and around 26 BTC lifetime. The affected volume is real, which argues for merging rather than for the gate.

On rejecting a rounded-zero volume

Agreed, and this is independent of the versioning question. getAdjustedAltcoinVolume documents that callers must reject zero and no core caller does. need to add the rejection at offer placement (both min and max volume) and at take time, in core rather than desktop so the API paths are covered, with boundary tests.

On Siafund

The test using SF as the zero-decimal example while Siafund still inherits the default 8 is a genuine inconsistency and need to fix it, but by changing the test rather than by setting SF to 0. At precision 0, half of one unit is a whole Siafund, which is not dust and would silently move real value. Indivisible assets need their own answer (reject them, or adjust the BTC amount to land on a whole-unit volume), not the same half-up rounding, and that is out of scope here. Same reasoning for ASK. needs adding registry-backed precision assertions for every asset that does get a non-default value, so a metadata omission cannot pass the suite.

fenlark:
@coderabbitai resume

fenlark:
Zero volume rejection: added at the three trust boundaries in core. ValidateOffer rejects a zero volume and min volume at placement, covering the desktop, API and cloned offers. TakeOfferModel rejects the derived volume on the API take path. The maker rejects the trade volume of the take request after the taker's validated price is applied, since a maker trade has no price before that point. These checks also cover market-priced offers whose price has risen far enough since placement and offers edited after placement, which do not re-run placement validation. The desktop take screen disables the take button and shows a validation message. Boundary tests cover the half-unit thresholds at precision 0 and 2, the helper contracts, and a maker trade whose price pushes the volume below half a unit.

Protocol version: agree with wodoro's analysis, no bump. The spec commit records the bounded divergence argument.

Siafund: kept at the default precision and the zero-decimal test example uses a synthetic ticker instead. CurrencyUtilTest now derives the set of sub-8 assets from the registry and asserts their declared precisions, so a metadata omission or an unreviewed addition fails the suite; SF and ASK are asserted through the registry as deliberately default.

Also updated the stale gRPC volume comment (it described 2 decimal places) and added docs/specifications/trade/altcoin-volume-precision.md with the rounding rule, the zero rejection boundaries, the open question on indivisible assets, and BSQ documented as a deliberate exception (its on-chain unit is 2 decimals, but BSQ trading uses the swap protocol and its own 2-decimal handling).

A follow-up candidate outside this PR: the create-offer screen shows the raw ValidateOffer failure popup for a fixed-price offer whose volume rounds to zero; a friendly validation message like the take screen's could be added there.

HenrikJannsen (approved):
utACK

PR #7979, opened 2026-07-28
Remove JDK toolchain resolver

alvasw:
Since upgrading to JDK 21, the toolchain resolver is no longer needed. It was originally introduced because most OS distributions did not include JDK 11 in their default package repositories, but JDK 21 is widely available.

## Summary by CodeRabbit

* **Build Improvements**
  * Updated the project’s build configuration to use standard Gradle plugin management.
  * Replaced the previous custom Java toolchain resolution setup with a new build module.
  * Existing project modules and dependency version settings remain unchanged.


alejandrogarcia83 (approved):
utACK

PR #7980, opened 2026-07-28
Enforce consistent locale and encoding for reproducible builds

alvasw:
If a build is executed on machines with different native locales or
character sets, the resulting binary artifacts can differ. Tools like
Javadoc or code generators might output different date formats,
alphabetical sorting, or localized text.

This pins the Gradle JVM's system properties to UTF-8 and en_US,
aligning with the JVM guidelines from reproducible-builds.org.
This ensures our build artifacts remain bit-for-bit identical
regardless of the host environment (CI or local developer machine).

## Summary by CodeRabbit

* **Build Improvements**
  * Standardized character encoding and language settings across builds for more consistent behavior across environments.
  * Simplified build configuration by removing custom Java runtime resolution and relying on the standard build tooling configuration.
* **Maintenance**
  * Streamlined project setup and reduced custom build components, helping improve build maintainability and consistency.


alejandrogarcia83 (dismissed):
utACK

PR #7981, opened 2026-07-28
common: replace SignalHandler with ShutdownHook

alvasw:
The current shutdown implementation traps INT and TERM signals
using the internal JDK classes Signal and SignalHandler.

These APIs are part of the internal JDK and are strongly
encapsulated starting with JDK 17, making them inaccessible on
modern JVMs. In addition, this mechanism only works on Unix-like
systems (macOS and Linux) where these signals are available.

Replace the signal-based shutdown handling with a JVM
ShutdownHook. This provides a supported and portable way to
perform graceful shutdown across JVM versions and platforms.

## Summary by CodeRabbit

- **Build & Compatibility**
  - Improved build consistency by standardizing UTF-8 encoding and English (US) locale settings.
  - Updated build configuration to use centralized plugin management.
  - Added support for managing the Tor binary as part of the build.

- **Bug Fixes**
  - Improved graceful application shutdown across supported environments, including a controlled wait for cleanup to complete.


alejandrogarcia83 (dismissed):
utACK

PR #7991, opened 2026-08-02
Add dust limit validation in WithdrawalView to prevent IllegalArgumentException

### LINKED ISSUE #5532: [v1.6.5] Handle send amounts below dust more gracefully
### Description

We shouldn't get to the point showing a IllegalArgumentException when an amount below the dust limit is entered. 

#### Version

v1.6.5

### Steps to reproduce

Send funds from wallet and try to send an amount below dust limit 

### Expected behaviour

An validation error is displayed and `Withdraw selected` can't be pressed.

### Actual behaviour

Withraw selected can be triggered and an `java.lang.IllegalArgumentException: The amount is too low (dust limit).` is surfaced.

### Screenshots

Bildschirmfoto 2021-05-26 um 11 01 20

mmustafasenoglu:
## Summary

Fixes #5532

When a user enters an amount below the dust limit in the withdrawal screen, the "Withdraw selected" button could still be clicked, resulting in an unhandled `IllegalArgumentException: The amount is too low (dust limit).` being shown to the user.

## Problem

In `BtcWalletService.sendFundsForMultipleAddresses()`, the dust check uses `Preconditions.checkArgument()` which throws `IllegalArgumentException`. However, `WithdrawalView.sendFunds()` only catches `Wallet.DustySendRequested`, not `IllegalArgumentException`, so the raw exception message is displayed to the user in a popup.

## Fix

Added a dust limit validation check in `WithdrawalView.areInputsValid()` using `Restrictions.isAboveDust()` to catch the invalid amount early and display a proper formatted error message using the existing `validation.amountBelowDust` i18n key.

This prevents the `IllegalArgumentException` from being thrown at all, since the validation now catches the dust violation before the transaction is attempted.

## Changes

- `desktop/src/main/java/bisq/desktop/main/funds/withdrawal/WithdrawalView.java`: Added dust limit check in `areInputsValid()` method (+6 lines)

## Summary by CodeRabbit

* **Bug Fixes**
  * Withdrawals now warn users when the sender amount is too small to meet the minimum non-dust output requirement.
  * Invalid low-value withdrawals are blocked before further address and balance validation.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/funds/withdrawal/WithdrawalView.java
+++ b/desktop/src/main/java/bisq/desktop/main/funds/withdrawal/WithdrawalView.java
@@ -618,6 +618,12 @@ public class WithdrawalView extends ActivatableView<VBox, Void> {
             return false;
         }
 
+        if (!Restrictions.isAboveDust(sendersAmount)) {
+            new Popup().warning(Res.get("validation.amountBelowDust",
+                    formatter.formatCoinWithCode(Restrictions.getMinNonDustOutput()))).show();
+            return false;
+        }
+
         if (!btcAddressValidator.validate(withdrawToTextField.getText()).isValid) {
             new Popup().warning(Res.get("validation.btc.invalidAddress")).show();
             return false;


mmustafasenoglu:
Thanks for the thorough review. You're right that in fee-excluded mode the sender amount could exceed dust while the receiver output doesn't. However, this PR's primary goal is to surface the dust limit error to the user via a friendly message (rather than letting it bubble up as an unhandled IllegalArgumentException from BtcWalletService). The current check on sendersAmount catches the most common case where both amounts are the same (fee-included mode).

Moving the validation to receiverAmount/netValue would require accessing the fee calculation earlier in the flow, which is a larger change. I'd suggest merging this as-is for the immediate UX improvement, and opening a follow-up issue to tighten the validation to netValue if needed. Happy to address any feedback from maintainers.

PR #8000, opened 2026-08-03
Delete .github/workflows/stale.yml

alvasw:
We receive more requests than we can handle. The stale-bot marks most items as stale and closes them before we’ve had a chance to process them. Until we’re fast enough to process every request, let’s temporarily disable it.

## Summary by CodeRabbit

* **Chores**
  * Removed automated stale issue and pull-request management.
  * Scheduled and manual stale-processing runs are no longer performed.


HenrikJannsen:
Replaced by https://github.com/bisq-network/bisq/pull/8004

PR #8008, opened 2026-08-09
Keep historical BTC fees address as fee receiver

### LINKED ISSUE #7987: Offers deactivated with "Fee receiver not a recognised address" after 2026-07-30 filter update
### Description

Since 2026-07-30, open offers whose maker fee tx pays the legacy Burning Man BTC fees address 38bZBj5peYS3Husdz7AH3gEUiUbYRD951t are being deactivated with "Fee receiver not a recognised address" (NACK_UNKNOWN_FEE_RECEIVER). The offers were created during the July fee routing transition, validated fine for days, and only started failing after a filter update received on 2026-07-30. Multiple users report the same on the forum: https://bisq.community/t/offer-deactivated-fee-receiver-not-a-recognised-address/13892

#### Version

v1.10.4

### Related tickets

#7534 (same error string, Nov 2025), #7905 (filter based fee receiver routing)

### Cause

A stock client only pays the maker fee to an address emitted by BtcFeeReceiverService.getAddress(), so these offers paid 38bZ... because the network data active at creation time designated it. Validation however is time-of-check: MempoolService.getAllBtcFeeReceivers() is rebuilt on every validation from the current filter receivers, the BM address list legacy address, DAO donation addresses and BM candidates that ever burned BSQ, and none of these contains 38bZ... any more, so TxValidator.checkFeeAddressBTC() rejects fee txs that were valid when created (the BLOCK_TOLERANCE leniency only covers offers created before block 600000, ~Oct 2019). TriggerPriceService then deactivates the maker's own offers, and desktop takers are blocked at the next step of the take offer screen (popup.warning.makerTxInvalid), so the offers are effectively untakeable. Makers who cancel forfeit the paid maker fee through no fault of their own.

### Suggested fix

1. Immediate, data only: re-add the address to the filter's btc_fee_receiver_addresses. The weighted syntax minimum is 0.0001, so 38bZBj5peYS3Husdz7AH3gEUiUbYRD951t#0.0001 makes it recognised again while routing only ~0.01% of new fees to it. Note the parser rejects configs mixing weighted and plain entries, so if the current filter carries plain entries the whole list needs to be republished in weighted form (a plain 38bZ... entry would instead route 1/N of all new fees to it). Affected makers can then simply re-enable their offers (fee validation status is transient and resets on activation).

2. Structural: retired fee receivers should enter a validate-only state instead of dropping out of the recognised set entirely. Selection never emits them for new offers, but mempool validation keeps accepting fee txs made while the address was still active, e.g. by giving each receiver an active height range. A routing change then cannot invalidate already published offers, and the retired entry loses relevance on its own once no offer referencing it remains. Can submit a PR for this if preferred.

fenlark:
Fixes #7987.

The mempool validation list is built from current network data only, so
the 2026-07-30 filter update retroactively invalidated fee txs which
had paid the legacy burning man BTC fees address while it was the
network-designated receiver. Affected offers get deactivated with "Fee
receiver not a recognised address" and takers are blocked from taking
them.

This adds the historical address permanently to getAllBtcFeeReceivers,
mirroring how getAllDonationAddresses keeps the past DPT addresses. The
fee receiver selector never emits it for new fee txs, so no new fees
route there; the address only stays valid for validation of existing
ones. Includes a test.

This fixes validation from the next release on. Already deactivated
offers can then simply be re-enabled by their makers (the fee
validation status is transient and resets on activation). A faster
remedy for currently affected users would be a filter update re-listing
the address, as described in #7987.

## Summary by CodeRabbit

* **Bug Fixes**
  * Historical Bitcoin fee payments are now consistently recognized, including when configured fee addresses target a different network.
  * Improved compatibility with transactions sent to the legacy Bitcoin fee address.

* **Tests**
  * Added coverage to verify recognition of the historical fee receiver address.

### DIFF
--- a/core/src/main/java/bisq/core/dao/burningman/BurningManPresentationService.java
+++ b/core/src/main/java/bisq/core/dao/burningman/BurningManPresentationService.java
@@ -67,7 +67,7 @@ public class BurningManPresentationService implements DaoStateListener {
     private static final long BURN_TARGET_BOOST_AMOUNT = 10000000;
     public static final String LEGACY_BURNING_MAN_DPT_NAME = "Legacy Burningman (DPT)";
     public static final String LEGACY_BURNING_MAN_BTC_FEES_NAME = "Legacy Burningman (BTC fees)";
-    static final String LEGACY_BURNING_MAN_BTC_FEES_ADDRESS = "38bZBj5peYS3Husdz7AH3gEUiUbYRD951t";
+    public static final String LEGACY_BURNING_MAN_BTC_FEES_ADDRESS = "38bZBj5peYS3Husdz7AH3gEUiUbYRD951t";
     // Those are the opReturn data used by legacy BM for burning BTC received from DPT.
     // For regtest testing burn bsq and use the pre-image `dpt` which has the hash 14af04ea7e34bd7378b034ddf90da53b7c27a277.
     // The opReturn data gets additionally prefixed with 1701
--- a/core/src/main/java/bisq/core/provider/mempool/MempoolService.java
+++ b/core/src/main/java/bisq/core/provider/mempool/MempoolService.java
@@ -267,6 +267,9 @@ public class MempoolService {
         List<String> btcFeeReceivers = new ArrayList<>();
         btcFeeReceivers.addAll(daoFacade.getAllDonationAddresses());
         addBurningManAddressListLegacyAddress(btcFeeReceivers);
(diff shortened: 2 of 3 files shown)


fenlark:
Closing: the legacy fee address stays disabled by maintainer decision, re-enabling would mean collecting funds back from past contributors.
Noting for the record that this would also have needed the address in bm-addresses-v000x.json, e.g. as a cappedBurnAmountShare=0 entry like the refund agents, so it is not eligible for DPT payouts.

PR #8012, opened 2026-08-09
Show update popup only once per session

### LINKED ISSUE #7936: Update warning opens two times in a row
After closing this warning it immediately reappears:

Image

fenlark:
Fixes #7936

The software update alert can be delivered to a running node more
than once. When its P2P storage entry expires or gets removed,
AlertManager clears alertMessageProperty or falls back to another
alert (onAlertRemoved), and a later re-receipt sets the property
again with an alert for an update that was already shown. The
alertMessageProperty listener in BisqSetup then fires again and
queues another update popup, so the user gets the same update
window again right after closing the first one, as described in
the report.

The normal message alert branch of displayAlertIfPresent already
guards against repeats by comparing user.getDisplayedAlert(). The
update branch has no such check, and the persisted displayedAlert
cannot serve as the guard there because the update popup should
show again in a new session as a reminder (canShowPopup already
handles the dont-show-again preference).

This change tracks the alert versions shown in the current session
and skips repeated popups for them, following the same approach as
the disk space warning fix in #7580. The session guard uses the
same version key as the persisted dont-show-again preference, so
it also covers a re-published alert for the same version. Opening
the popup via the footer link (openNewVersionPopup) is not
affected and an alert for a new version still shows.

Deduplicating inside AlertManager instead would change the
property semantics covered by AlertManagerTest; BisqSetup is the
only consumer and the message alert branch already handles repeats
there, so the guard lives next to it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate update alerts from appearing repeatedly during the same session.
  * Update alerts can still be reopened manually through the footer link.
  * Newer software versions continue to display their update alerts.
  * Respected the “don’t show again” preference for automatically displayed alerts.

### DIFF
--- a/core/src/main/java/bisq/core/app/BisqSetup.java
+++ b/core/src/main/java/bisq/core/app/BisqSetup.java
@@ -88,6 +88,7 @@ import java.io.FileWriter;
 import java.io.IOException;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


HenrikJannsen (approved):
utACK

PR #8022, opened 2026-08-11
Discard superseded trade statistics async results

### LINKED ISSUE #8018: TradesChartsViewModel: async refilter and chart results are not guarded against a newer request
On the market trade statistics screen the asynchronous refilter and chart calculations are not guarded against a newer request, so an older completion can overwrite state produced by a newer one.

`TradesChartsViewModel.applyAsyncTradeStatisticsForCurrency` runs the refilter on a background thread and, on completion, does `tradeStatisticsByCurrency.setAll(list)` on the user thread. `applyAsyncChartData` publishes `priceItems`, `volumeItems` and `volumeInUsdItems` the same way. Neither checks whether the request that started the work is still the current one.

These are started from `onSetTradeCurrency`, `setTickUnit`, the trade-statistics set-change listener and `activate`. If the user switches currency or tick unit while a previous run is still in flight, the older run can complete last and replace the visible data with the previous currency's or interval's result until the next refresh corrects it.

This is pre-existing and not specific to any recent change; it was noted during review of #8017, which is scoped to memory pressure and does not touch this path.

A fix would give the view model a monotonically increasing request id, captured when each request starts and threaded through both async calls, and checked before the refilter result is applied with `setAll` and before the chart items are published. Stale completions would be dropped.

fenlark:
Fixes #8018

The trade statistics screen runs asynchronous refilter and chart
calculations in the view model and builds the table rows in the
view, from several triggers: a currency change, a tick unit change,
the trade statistics set change and activate. An older calculation
could complete after a newer request and overwrite the model data,
the chart items or the table with an outdated result.

Each refilter, chart calculation and table build now takes a
generation when it starts on the user thread and applies its result
only if the generation is still current, so superseded results are
dropped. The refilter, the chart and the table build use separate
generations: the refilter and the chart are independent view model
pipelines (a tick unit change bumps only the chart), and the table
build lives in the view, so a single shared counter cannot span
them without cross-invalidating an in-flight build.

The table build also snapshots the observable list and the current
sort comparator on the user thread and sorts off it. Only the pure
per-item column comparators run on the background thread, no scene
graph is read there, which removes the earlier off-thread bind of
the sorted list comparator to the live table property.

This is stacked on #8020.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved responsiveness and stability when updating trade charts and tables.
  * Prevented outdated trade data, currency filters, and chart calculations from replacing newer results.
  * Improved behavior when leaving and returning to the trade charts view.
  * Ensured rapidly changing currency selections display the latest available results.
  * Preserved error handling while allowing superseded updates to complete safely.
  * Fixed chart updates when USD price data is not yet available.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/market/trades/ChartCalculations.java
+++ b/desktop/src/main/java/bisq/desktop/main/market/trades/ChartCalculations.java
@@ -119,7 +119,9 @@ public class ChartCalculations {
             // Generate date range and create sets for all ticks
             List<Pair<Date, Set<TradeStatistics3>>> itemsPerInterval = getItemsPerInterval(tradeStatisticsByCurrency, tickUnit);
 
-            Map<Long, Long> usdAveragePriceMap = usdAveragePriceMapsPerTickUnit.get(tickUnit);
+            // getOrDefault: the map is empty when the chart runs before the USD-average maps are
+            // ready; a missing tick entry then falls back to the previous/zero average below.
+            Map<Long, Long> usdAveragePriceMap = usdAveragePriceMapsPerTickUnit.getOrDefault(tickUnit, Map.of());
             AtomicLong averageUsdPrice = new AtomicLong(0);
 
             // create CandleData for defined time interval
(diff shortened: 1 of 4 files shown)


HenrikJannsen (approved):
utACK

HenrikJannsen:
Added via merge from the security branch:  https://github.com/bisq-network/bisq/commit/6bf412601de914ec87913f85edfa285e8a1cb7cc

PR #8024, opened 2026-08-11
Avoid materializing a Price per trade statistic

fenlark:
onAllServicesInitialized seeds PriceFeedService with the newest price
per currency by calling getTradePrice() on every trade statistic.
getTradePrice() lazily builds and caches a Price on the element, so the
loop pinned a Price (plus its Monetary) on every row for the whole
session, although only the latest price per currency is used.

The statistics set is sorted by date, so this iterates it newest-first
and keeps the first price seen per currency with computeIfAbsent. The
resulting map is identical (the set is a TreeSet with a total order, so
descending-first-wins selects the same element per currency as the
previous ascending forEach with last-write-wins), but getTradePrice()
runs once per currency instead of once per element.

Measured on a full mainnet store: retained Price instances at startup
dropped from about 410k to a few hundred, roughly 19 MB less, before
the Market -> Trades screen is opened. No behavior change; the seed map is
identical and is refreshed by the live price feed as before.

putIfAbsent and a toMap collector were avoided on purpose: both evaluate
getTradePrice() for every element, which is exactly the materialization
this change removes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved initial market price selection by using the latest available trade statistics for each currency.
  * Prevented older trade data from incorrectly overriding newer market prices.

### DIFF
--- a/core/src/main/java/bisq/core/trade/statistics/TradeStatisticsManager.java
+++ b/core/src/main/java/bisq/core/trade/statistics/TradeStatisticsManager.java
@@ -118,10 +118,14 @@ public class TradeStatisticsManager {
                 .filter(TradeStatistics3::isValid)
                 .forEach(observableTradeStatisticsSet::add);
 
-        // get the most recent price for each ccy and notify priceFeedService
-        // (this relies on the trade statistics set being sorted by date)
+        // Seed the most recent price per currency. The set is date-sorted, so iterating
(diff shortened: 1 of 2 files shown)


HenrikJannsen (approved):
utACK

HenrikJannsen:
Added via merge from the security branch: https://github.com/bisq-network/bisq/commit/1697b04e6f0952058d4a91f3d298d5dbdc473a33

PR #8060, opened 2026-09-02
Skip the republish retry before bootstrap

fenlark:
An API edit that re-enables a deactivated offer is reachable before the
P2P network is bootstrapped: the daemon reads the persisted open offers
and starts the gRPC server before BisqSetup begins the P2P setup, and
editoffer requires neither a wallet nor a bootstrapped network. addOffer
then reports "the P2P network is not bootstrapped yet" through the error
handler, which stops and re-arms the retry timer unconditionally. That
republishes every open offer every 10 seconds, logging an error per
offer, for as long as the bootstrap takes, and forever if it never
completes.

Retrying cannot change the outcome there, since only the bootstrap can,
and onBootstrapComplete republishes all open offers anyway. The armed
timer also suppresses the second republish round scheduled 30 seconds
after that one, which is armed only while no retry timer is set.

Skip the retry while the offer book is not bootstrapped and report the
failure as a warning naming the offer, matching maybeRefreshOffer, whose
error handler for the same condition is a plain log.warn with no retry.
A failure on a bootstrapped node keeps the error log and the retry.

P2PService.isBootstrapped is assigned once, inside applyIsBootstrapped,
and is never reset, and the flag is set before the P2P listeners are
notified. So onBootstrapComplete always runs with the flag true and
republishes every open offer, and the reconnect path
(onAllConnectionsLost / restart) runs only after the bootstrap, so it
keeps its retry unchanged.

Only testMaybeRepublishOfferSkipsRetryWhileNotBootstrapped is a
regression test: it fails on master with two addOffer calls instead of
one. testMaybeRepublishOfferRetriesWhenBootstrapped passes on master and
is there to keep the guard narrow. The log level itself is not asserted,
as the repo has no log capture helper.

No specification change: the contract is already stated in
docs/specifications/offer/offer-edit-and-removal.md, which says that a
publish failure before the bootstrap does not fail the edit and that the
offer is published by the republish that follows the bootstrap. The
retry schedule and the log level are implementation.

Follow-up to the review of #8021.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved offer republishing after failed publication attempts.
  - Avoided unnecessary retry attempts while the peer-to-peer network is not bootstrapped.
  - Continued processing subsequent offers when network availability changes.
  - Preserved retry behavior once the network is bootstrapped.
  - Completed offer processing cleanly when publication cannot proceed due to unavailable network readiness.

- **Tests**
  - Added coverage for republishing before and after network initialization, including continued offer processing when retries are skipped.

### DIFF
--- a/core/src/main/java/bisq/core/offer/OpenOfferManager.java
+++ b/core/src/main/java/bisq/core/offer/OpenOfferManager.java
@@ -1248,6 +1248,22 @@ public class OpenOfferManager implements PeerManager.Listener, DecryptedDirectMe
                 },
                 errorMessage -> {
                     if (!stopped) {
+                        if (!offerBookService.isBootstrapped()) {
+                            // Retrying cannot change the outcome before the bootstrap, and
+                            // onBootstrapComplete republishes all open offers anyway. An armed
+                            // retry timer would also suppress the republish round which is
+                            // scheduled after the bootstrap only while no retry timer is set.
+                            log.warn("Adding offer to P2P network failed. We do not retry while the " +
+                                            "P2P network is not bootstrapped yet, as all open offers get " +
+                                            "republished after the bootstrap. offerId={}, errorMessage={}",
+                                    openOffer.getId(), errorMessage);
+
+                            if (completeHandler != null) {
+                                completeHandler.run();
+                            }
+                            return;
+                        }
+
                         log.error("Adding offer to P2P network failed. " + errorMessage);
                         stopRetryRepublishOffersTimer();
                         retryRepublishOffersTimer = UserThread.runAfter(OpenOfferManager.this::republishOffers,
--- a/core/src/test/java/bisq/core/offer/OpenOfferManagerTest.java
+++ b/core/src/test/java/bisq/core/offer/OpenOfferManagerTest.java
@@ -575,4 +575,115 @@ public class OpenOfferManagerTest {
         assertFalse(errorHandled.get());
     }
 
+    @Test
+    public void testMaybeRepublishOfferSkipsRetryWhileNotBootstrapped() {
+        P2PService p2PService = mock(P2PService.class);
+        OfferBookService offerBookService = mock(OfferBookService.class);
+        when(p2PService.getPeerManager()).thenReturn(mock(PeerManager.class));
+        OpenOfferManager manager = new OpenOfferManager(coreContext,
+                null,
+                null,
+                null,
+                p2PService,
+                null,
+                null,
+                null,
+                offerBookService,
+                null,
+                null,
+                null,
+                null,
+                null,
+                null,
+                null,
+                null,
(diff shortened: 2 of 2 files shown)


majesticsky (commented):
Reviewed by running the tests rather than reading the diff, on `c1d595e` against current master.

What I checked:

* `:core:test --tests bisq.core.offer.OpenOfferManagerTest`: 12 tests, 0 failures; both new tests are present.
* Reverting only the `OpenOfferManager` change makes `testMaybeRepublishOfferSkipsRetryWhileNotBootstrapped` fail with `TooManyActualInvocations` (`addOffer` called twice), confirming the regression test.
* `OfferBookService.isBootstrapped()` delegates directly to `p2PService.isBootstrapped()`, matching the condition behind the reported error.
* `onBootstrapComplete` only arms `retryRepublishOffersTimer` when it is `null`, so the second republish round is indeed suppressed.
* The early return does not need to stop the timer: nothing can arm it before bootstrap.

Two non-blocking remarks:

* The guard is based on network state rather than the specific error. This also suppresses retries for the two earlier `addOffer` exits before bootstrap. Seems harmless since `onBootstrapComplete` republishes all open offers, but it is slightly broader than necessary.
* `completeHandler` keeps `processListForRepublishOffers` walking the list, but this isn't covered by the new tests. Both call `maybeRepublishOffer(openOffer)` with `null`, so dropping it later could stop list processing while the tests remain green.

fenlark:
@majesticsky Thanks for review.

First remark: intentional. The two earlier exits in addOffer (mandatory update filter, price out of bounds) cannot be fixed by a retry before the bootstrap either, and onBootstrapComplete republishes all open offers once it is there. maybeRefreshOffer handles the same condition the same way.

Second remark: agreed, added ed2f5b861f with a test where the retry fires after the network is no longer bootstrapped with two open offers in the list. It fails if the early return stops calling completeHandler.

HenrikJannsen:
Can you rebase on master?

majesticsky:
Both remarks settled from my side.

On the first — agreed. The two earlier exits in `addOffer` are in the same position: retrying before bootstrap cannot change their outcome, and `onBootstrapComplete` republishes everything once it arrives. Nothing to fix.

On the second — verified the new test catches the defect. `:core:test --tests OpenOfferManagerTest` on `ed2f5b8`: 13/13 green. Removing `completeHandler.run()` made exactly `testRepublishOffersKeepsWalkingTheListWhileNotBootstrapped` fail; the other 12 stayed green. The two older tests pass a null handler, so they would not catch the stopped walk.

The rebase preserves the same patches (`8eca917` / `bd0723e` vs. `c1d595e` / `ed2f5b8`), so no re-check needed.

Looks good to me.

HenrikJannsen (approved):
utACK

PR #8061, opened 2026-09-02
Reject payment started without seller account

fenlark:
CoreTradesService.confirmPaymentStarted checks only that the trade has
not failed and that its deposit tx is confirmed. Neither says anything
about the seller's payment account payload, which reaches the buyer
with the DepositTxAndDelayedPayoutTxMessage and can be absent on a
trade whose deposit the buyer confirmed from the wallet alone.

None of the five tasks in BuyerProtocol.onPaymentStarted reads that
payload: ApplyFilter treats it as nullable, the fee verification tasks
are no-ops, and the payout tasks use the contract only for addresses
and multisig keys. So the call succeeds, the trade moves to FIAT_SENT
and the payment started message goes to the seller for a payment the
buyer had no details to make.

FIAT_SENT is also past every phase in which the buyer still accepts
the DepositTxAndDelayedPayoutTxMessage, so the confirm closes the last
window in which the missing payload could still arrive.

Reject the call while the contract carries no seller payment account
payload. The check sits before the counter currency tx proof is
stored, so a rejected confirmpaymentstartedxmr leaves no proof data on
the trade.

The desktop client refuses the same state since #7955, which disables
the confirm button on the same contract payload.

Both rejection tests fail on master: without the guard the call reaches
onPaymentStarted and stores the tx proof. The third test covers the
normal path, so the guard cannot be widened unnoticed.

A precondition in BuyerProtocol would not work for this: FluentProtocol
returns before the task runner is built, so the gRPC caller would still
get an OK. A first task in onPaymentStarted would cover both clients,
but it runs after the phase is set and its failure writes an error
message on the trade, so it is a protocol change rather than the guard
this follow-up was about.

No specification change: this is a client-side precondition, not a rule
of the trade protocol, and neither of the two merged PRs for this issue
carried one.

Part of #7945

## Summary by CodeRabbit

- **Bug Fixes**
  - Payment-start confirmation is now rejected when required trade or seller payment details are missing.
  - Prevents payment-start actions from occurring when confirmation data is incomplete.
  - Valid payment-start confirmations continue to be processed successfully.

### DIFF
--- a/core/src/main/java/bisq/core/api/CoreTradesService.java
+++ b/core/src/main/java/bisq/core/api/CoreTradesService.java
@@ -202,6 +202,19 @@ class CoreTradesService {
                                 trade.getId(),
                                 trade.getDepositTxId()));
             }
+
+            // Without the seller's payment account payload the buyer has no details to pay
+            // to, so a payment started message would tell the seller a payment happened that
+            // cannot have happened. The desktop client disables its confirm button on the
+            // same condition.
+            if (trade.getContract() == null
+                    || trade.getContract().getSellerPaymentAccountPayload() == null) {
+                throw new FailedPreconditionException(
+                        format("cannot send a payment started message for trade '%s'%n"
+                                        + "until the seller's payment account details have been received",
+                                trade.getId()));
+            }
+
             // pass along counter currency tx proof info if provided
             if (txId != null && txKey != null && !txId.isEmpty() && !txKey.isEmpty()) {
                 trade.setCounterCurrencyTxId(txId);
--- a/core/src/test/java/bisq/core/api/CoreTradesServiceConfirmPaymentStartedTest.java
+++ b/core/src/test/java/bisq/core/api/CoreTradesServiceConfirmPaymentStartedTest.java
@@ -1,0 +1,116 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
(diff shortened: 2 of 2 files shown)


majesticsky (commented):
Reviewed by running the tests and checking the claims against master, on `d19bf16`.

Tests: merged into current master: 3 tests, 0 failures. With only the new test file on current master, 2/3 fail as expected (`Expected FailedPreconditionException`, nothing thrown); the normal-path test stays green.

Code checks: `confirmPaymentStarted` previously checked only `hasFailed()` and `isDepositConfirmed()`, and the new guard runs before `counterCurrencyTxId` is stored. `onPaymentStarted` matches the five described tasks. `DEPOSIT_TX_AND_DELAYED_PAYOUT_TX_MSG_PHASES` ends at `DEPOSIT_CONFIRMED`, with `FIAT_SENT` following it. The desktop disables the button under the same condition, and `BuyerTrade.confirmPermitted()` only covers arbitrated trades, so the guard is not redundant.

Follow-up: the same `FluentProtocol.executeTasks` behavior means `!trade.confirmPermitted()` cannot be enforced as a `BuyerProtocol` precondition either. I checked that with a throwaway test using a real `BuyerProtocol` rather than a mock: on a trade in arbitration `confirmPaymentStarted` throws nothing and the trade state is never written, while the same test with `confirmPermitted()` true proceeds. Same pattern as what this PR fixes, potentially worth a follow-up PR.

majesticsky:
If nobody else is on it, I am happy to do the `confirmPermitted()` follow-up as a separate PR.

HenrikJannsen (approved):
utACK

PR #8062, opened 2026-09-05
Initialize trusted block providers only once

### LINKED ISSUE #7973: Use an explicit initialized flag in TrustedBsqBlockProviderRepository
Address review comment: https://github.com/bisq-network/bisq/pull/7966#discussion_r3653946833

majesticsky:
Fixes #7973 — the review comment it points to:
https://github.com/bisq-network/bisq/pull/7966#discussion_r3653946833

`getTrustedBsqBlockProviders()` used the emptiness of the provider set as its lazy initialization guard. An empty set is a legitimate outcome: regtest bundles no providers, the resource file can be absent, and the deny list may filter out every entry. In those cases, `lazyInitialize()` ran again on every call, re-reading the resource on a path that `DaoBlockSignatureVerifier` walks per block.

Populating the set in place also handed callers a live reference to it, and left it briefly empty between `clear()` and `addAll()`. Every caller reaches this singleton on the user thread today, so that window is not observable in practice, but nothing in the class enforces it.

Track initialization in an explicit flag, synchronize the accessor, and publish the providers as an immutable set, so the guard is atomic and callers cannot mutate shared state.

Two new tests cover this; both fail without the production change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved trusted block provider loading so initialization occurs reliably and only once, including when no providers are available.
  - Prevented accidental modification of the trusted provider list after it has been loaded.
  - Improved consistency when accessing trusted block provider information from multiple operations.

### DIFF
--- a/core/src/main/java/bisq/core/dao/node/block_provider/TrustedBsqBlockProviderRepository.java
+++ b/core/src/main/java/bisq/core/dao/node/block_provider/TrustedBsqBlockProviderRepository.java
@@ -33,7 +33,6 @@ import java.io.InputStreamReader;
 import java.io.UncheckedIOException;
 
 import java.util.Collection;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


majesticsky:
CI is green on my fork for this commit — ubuntu-24.04, macos-15 and
windows-2025 on Java 21.0.6:
https://github.com/majesticsky/bisq/actions/runs/33975169843

majesticsky:
Anything needed from my side to get the checks running here? The fork run linked above is on "dcf9916f", the commit this PR points at.

HenrikJannsen (approved):
utACK

PR #8069, opened 2026-09-21
Approve the Gradle 9.0.0 wrapper files

fenlark:
The "Verify Gradle wrapper checksums" step in build.yml runs
`sha256sum -c gradle/wrapper/gradle-wrapper.sha256` before anything
else. The update to Gradle 9.0.0 (406a4cc8f5, merged with the v1.10.8
integration in #8067) replaced gradle-wrapper.jar and
gradle-wrapper.properties but did not update that approved list, so the
step fails on every workflow run since 2026-09-16, on master as well as
on every pull request, and the test jobs never start.

This records the checksums of the current files and changes nothing
else.

How the files were checked:
- gradle-wrapper.jar is byte for byte the wrapper jar published for
  Gradle 8.9 (`gradle-8.9-wrapper.jar.sha256` at services.gradle.org).
  This is what the wrapper task writes when it is run by Gradle 8.9 with
  a newer target version. Running `./gradlew wrapper` once more under
  9.0.0 would replace it with the 9.0.0 jar (76805e32...) and then need
  another update of this list; left to a maintainer.
- gradle-wrapper.properties is the file from 406a4cc8f5 (Gradle 9.0.0,
  distribution checksum, validateDistributionUrl=true).
- gradlew and gradlew.bat are unchanged and still match.

## Summary by CodeRabbit

* **Chores**
  * Updated Gradle wrapper integrity checksums to reflect the current wrapper files.


HenrikJannsen:
The hash of the jar is different to the one in https://github.com/bisq-network/bisq/pull/8070

HenrikJannsen:
Replaced by https://github.com/bisq-network/bisq/pull/8070

